PR #1274, opened 2020-09-07
497: Migrate to node 12

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/497

### Changes:
* Update node to v12.18
* As a consequence, update `realm` to version 6.1 in `meditrak-app`
* Use [core-js v3](https://github.com/zloirock/core-js). This should enable us to use all the hot ES features, even the ones not supported in Node.js yet (eg `Object.fromEntries`!)
* Fix a couple of sort methods which used to return a boolean instead of a number. This seemed to work fine in node v10 (perhaps as a coicidence?) but not in v12
* Silence winston logs while running `data-api` tests
* Use `.flat()` in one data builder to server as a proof of concept that previously unsupported syntax can be used now


packages/utils/src/period/periodExtremes.js
```diff
@@ -22,24 +22,23 @@ export const getMostAncientPeriod = periods => getPeriodExtreme(periods, false);
 
 /**
  * @param {string[]} periods An array of periods
- * @param {boolean} gettingMostRecent Should the function return the most recent period?
+ * @param {boolean} descending Whether to use descending order or not
```
edmofro, line 25: I find the old variable name much easier to think about!

packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfDataValues/tableOfPercentagesOfValueCounts.js
```diff
@@ -3,18 +3,16 @@
...
 
 class TableOfPercentagesOfValueCounts extends TableOfDataValuesBuilder {
   buildDataElementCodes() {
-    const dataElementCodes = flatten(
-      flatten(this.config.cells).map(cell =>
-        cell.numerator.dataValues.concat(cell.denominator.dataValues),
-      ),
-    );
+    const dataElementCodes = this.config.cells
+      .flat()
```
edmofro, line 13: Hooray!

packages/meditrak-app/package.json
```diff
@@ -109,7 +109,7 @@
     "react-navigation": "^1.5.8",
     "react-navigation-redux-helpers": "^1.0.3",
     "react-redux": "^5.0.7",
-    "realm": "^2.29.1",
+    "realm": "^6.1.0",
```
edmofro, line 112: Did you look at release notes for realm to check if there were breaking changes we need to be aware of? Most of their changes over the last years have been related to their internal sync module, which we don't use, but it'd be worth looking to see if there's anything we do need to change.

edmofro (approved):
lgtm!

We'll also need to update the ec2 instances to use node 12. Although you've updated the CI/CD machine, the packages eventually get built and run on whatever version the ec2 instance is running.

packages/utils/src/period/periodExtremes.js
kael89: I was very confused by the old variable name, that' why I changed it... Using a boolean variable may be the problem in the first place, I will try a different approach

packages/meditrak-app/package.json
kael89: Great point, I will have a look! I haven't been able to try the app locally yet because of some issues in my environment, but will also do

packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfDataValues/tableOfPercentagesOfValueCounts.js
kael89: Yay! 🎉 I created [1240: Use modern ES syntax](https://github.com/beyondessential/tupaia-backlog/issues/1240) to do the same in the rest of the codebase

kael89:
> We'll also need to update the ec2 instances to use node 12. Although you've updated the CI/CD machine, the packages eventually get built and run on whatever version the ec2 instance is running.

About that, I guess that the steps are:
1. Manually switch to v12 in the feature branch and test this
2. Manually switch to v12 in production just before the release

Don't know if there is a better way to do that

babel.config.json
```diff
@@ -1,38 +1,8 @@
...
-    [
-      "@babel/plugin-proposal-pipeline-operator",
-      {
-        "proposal": "minimal"
-      }
-    ],
-    "@babel/plugin-proposal-nullish-coalescing-operator",
-    "@babel/plugin-proposal-do-expressions",
-    "@babel/plugin-proposal-function-bind"
+    "@babel/plugin-proposal-export-default-from"
```
kael89, line 5: It seems that [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env) automatically loads Stage 4 proposals! 🎉

kael89:
Closing in favour of https://github.com/beyondessential/tupaia/pull/1336

PR #1275, opened 2020-09-07
776 store date picker in url

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/776


packages/web-frontend/src/historyNavigation/utils.js
```diff
@@ -87,6 +89,29 @@ export const isLocationEqual = (a, b) => {
...
+  return {
+    startDate: moment(startDate, GRANULARITY_CONFIG[periodGranularity].urlFormat),
+    endDate: moment(endDate || startDate, GRANULARITY_CONFIG[periodGranularity].urlFormat),
+  };
+};
+
+export const convertObjectToUrlPeriodString = ({ startDate, endDate }, periodGranularity) => {
+  if (GRANULARITIES_WITH_ONE_DATE.includes(periodGranularity)) {
+    if (!startDate.isSame(endDate)) {
+      console.error(
```
Alastair-L, line 104: How should we handle this case?

packages/web-frontend/src/actions.js
```diff
@@ -138,6 +138,8 @@ export const SET_PROJECT = 'SET_PROJECT';
 export const FETCH_PROJECTS_ERROR = 'FETCH_PROJECTS_ERROR';
 export const REQUEST_PROJECT_ACCESS = 'REQUEST_PROJECT_ACCESS';
 export const UPDATE_HISTORY_LOCATION = 'UPDATE_HISTORY_LOCATION';
+export const UPDATE_MEASURE_CONFIG_ONCE_HIERARCHY_LOADS =
```
Alastair-L, line 141: This feels really hacky to me...
The reason that I thought creating a new action is necessary:
- We need to wait for the measure hierarchy to be populated before we try to update it
- I only know how to do that in sagas

packages/web-frontend/src/historyNavigation/historyMiddleware.js
```diff
@@ -10,35 +10,38 @@
...
 
 import {
-  SET_PROJECT,
-  SET_ORG_UNIT,
-  SET_DASHBOARD_GROUP,
-  OPEN_ENLARGED_DIALOG,
-  CLOSE_ENLARGED_DIALOG,
-  SET_MEASURE,
   CLEAR_MEASURE,
+  CLOSE_ENLARGED_DIALOG,
```
Alastair-L, line 14: Organize imports

packages/web-frontend/src/historyNavigation/utils.js
```diff
@@ -87,6 +92,37 @@ export const isLocationEqual = (a, b) => {
   return true;
 };
 
+export const convertUrlPeriodStringToObject = (
```
Alastair-L, line 95: I ended up putting this logic here because it is specific to the url.

packages/web-frontend/src/utils/periodGranularities.js
```diff
@@ -27,34 +27,39 @@ const CONFIG = {
     chartFormat: 'Do MMMM YYYY',
     rangeFormat: 'Do MMMM YYYY',
     pickerFormat: 'D',
+    urlFormat: 'Do_MMM_YYYY',
```
Alastair-L, line 30: These formats should probably get chosen by someone else. I've just put place holders in.

packages/web-frontend/src/sagas.js
```diff
@@ -775,25 +749,18 @@ function* fetchCurrentMeasureInfo() {
...
-        measureHierarchy,
-        selectedMeasureId,
-        selectCurrentProject(state),
-      );
-
-      if (newMeasure !== selectedMeasureId) {
-        yield put(setMeasure(newMeasure));
-      }
-    } else {
+    const isHierarchyPopulated = !!measureHierarchy.length;
```
Alastair-L, line 778: This whole commit is unrelated, and could/should probably be reverted to save regression. What do you think?

packages/web-frontend/src/historyNavigation/historyMiddleware.js
kael89: Nice!

packages/web-frontend/src/utils/periodGranularities.js
kael89: Fair enough! Let's add some information in the card about how the urls will look like, and then we can all decide

packages/web-frontend/src/sagas.js
kael89: Before I review this change, could you give me some context about it? Is it intended as pure refactoring?

kael89:
@Alastair-L I will continue reviewing after https://github.com/beyondessential/tupaia/pull/1275#discussion_r485242554 is answered

packages/web-frontend/src/sagas.js
```diff
@@ -745,28 +739,8 @@ function* watchMeasureChange() {
...
-function* fetchMeasureInfoForMeasurePeriodChange() {
-  const state = yield select();
-  const currentMeasureId = selectCurrentMeasureId(state);
-
-  yield fetchMeasureInfo(currentMeasureId);
-}
-
-function* watchMeasurePeriodChange() {
-  yield takeLatest(UPDATE_MEASURE_CONFIG, fetchMeasureInfoForMeasurePeriodChange);
-}
```
Alastair-L: Should have mentioned earlier, but this section is not part of the refactor (should be with later commits)

packages/web-frontend/src/sagas.js
Alastair-L: Yes, it should be purely refactoring (plus a small performance loss? from using selectors (because of the overhead and lack of recalculation in real life))

Alastair-L: The motivation was that I found the existing code very confusing.

kael89: Fair enough! I think it is a valuable refactor. Remember to leave regression testing notes in the card

Alastair-L: Will do 👍

packages/web-frontend/src/historyNavigation/utils.js
kael89: As discussed, while this may indicate an error (start and end date being different for a one-date granularity), it's not our concern here. So we could as well as remove the check and error logging, and leave this concern to places where this state of data does matter

packages/web-frontend/src/actions.js
kael89: I'll admit my ignorance regarding sagas that must run in sync, @rohan-bes any suggestions for a clean approach here?

packages/web-frontend/src/actions.js
```diff
@@ -554,6 +556,18 @@ export function updateMeasureConfig(measureId, measureConfig) {
   };
 }
 
+/**
+ * Updates measure config for current measure in measureBar once the hierarchy is populated
+ *
+ * @param {object} measureConfig
+ */
+export function updateCurrentMeasureConfigOnceHierarchyLoads(periodString) {
```
kael89, line 564:
```suggestion
export function updateMeasureDateRangeOnceHierarchyLoads(periodString) {
```

(can also keep `Current` in the name if you thing it's required)

packages/web-frontend/src/actions.js
kael89: But if we keep this, I suggest renaming it to `UPDATE_MEASURE_DATE_RANGE_ONCE_HIERARCHY_LOADS` to make our intentions clearer

kael89 (changes requested):
Looks good overall! A couple of renames requested, and waiting for feedback on `UPDATE_MEASURE_CONFIG_ONCE_HIERARCHY_LOADS `

Could you add unit tests for the `periodString <=> dateRange` conversion utils?

Also, I hope that import reordering will not cause massive conflicts in other PRs. If it does it should be easy to solve those by also ordering the imports in the other PR, and then comparing. At any case it may worth letting people know in the `dev` channel after this is merged into `dev`

packages/web-frontend/src/historyNavigation/utils.js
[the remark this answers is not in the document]
Alastair-L: `DateRange` is a good name 👍

packages/web-frontend/src/actions.js
rohan-bes: Didn't get a chance to properly look into this one unfortunately, but I can't immediately see a better way of doing it. I'm cool with this going in as it is for now, we always have the option to refactor later.

kael89: Thanks @rohan-bes !

packages/web-frontend/src/tests/historyNavigation/UrlPeriodStringToDateRangeConversions.test.js
```diff
@@ -0,0 +1,56 @@
...
+};
+
+describe('convertUrlPeriodStringToDateRange', () => {
+  it('should decode strings with only one date', () => {
+    const { startDate, endDate } = convertUrlPeriodStringToDateRange(
+      SINGLE_PERIOD_STRING,
+      'one_day_at_a_time',
+    );
+    const { startDate: expectedStartDate, endDate: expectedEndDate } = SINGLE_DATE_RANGE;
+    expect(startDate.isSame(expectedStartDate)).toBe(true);
```
kael89, line 29: Using `isSame` is clever and clean! 💯

Can I suggest an optional change: it took me a bit to realise that `isSame` comes from `moment` and not from jest (although at a closer look, the syntax does not imply that). You can add a `assertMomentDatesAreEqual ` helper:

```js
const assertMomentDatesAreEqual = (moment1, moment2) => expect(moment1.isSame(moment2)).toBe(true)`
```

kael89 (changes requested):
Nice tests! One requested and one optional change.

By the way, we still need to decide on the date formats... Given that your code is easily configurable (+ we have tests), I happy for you to merge it in `url-refactor` and then change it if a different format is required. Could you ask a question in the card about the format?

packages/web-frontend/src/tests/historyNavigation/UrlPeriodStringToDateRangeConversions.test.js
Alastair-L: Clean!

packages/web-frontend/src/tests/historyNavigation/UrlPeriodStringToDateRangeConversions.test.js
```diff
@@ -0,0 +1,56 @@
+import moment from 'moment';
+import {
+  convertUrlPeriodStringToDateRange,
+  convertDateRangeToUrlPeriodString,
+} from '../../historyNavigation/utils';
+
+const SINGLE_PERIOD_STRING = moment('2020-01-02').format('Do_MMM_YYYY');
```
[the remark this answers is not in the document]
Alastair-L: Good point.

Alastair-L:
Fyi, I want to wait until `url-refactor` is merged to dev before this card is tested, as the priority is getting the testing on that finished I believe, which Bindu is almost finished.

kael89 (approved):

Alastair-L:
^ Force pushed due to the squash and merge of the url-refactor 😢

kael89:
> ^ Force pushed due to the squash and merge of the url-refactor 😢

No worries at this case!

kael89:
@rohan-bes the same `OneTimeLogin` flaky test failed in this build... 🤔

packages/web-frontend/src/tests/components/dateRangePicker.test.js
```diff
@@ -23,8 +27,8 @@ describe('dateRangePicker', () => {
       render(<DateRangePicker granularity={key} />);
 
       const labelText = screen.getByLabelText('active-date');
-      const startDate = MIN_MOMENT_DATE.format(value.rangeFormat);
-      const endDate = MAX_MOMENT_DATE.startOf(value.momentUnit).format(value.rangeFormat);
+      const startDate = formatMomentAsString(MIN_MOMENT_DATE, key, value.rangeFormat);
```
Alastair-L, line 30: Is it ok to use this util as the util itself is not being tested here?

packages/web-frontend/src/utils/periodGranularities.js
```diff
@@ -122,8 +122,11 @@ export function roundStartEndDates(granularity, startDate = moment(), endDate =
 
 export const formatMomentAsString = (date, granularity, format) =>
   granularity === 'week' || granularity === 'one_week_at_a_time'
-    ? date.startOf('W').format(format)
-    : date.format(format);
+    ? date
+        .clone()
+        .startOf('W')
+        .format(format)
+    : date.clone().format(format);
```
Alastair-L, line 129: Hurray for unit tests!

packages/web-frontend/src/tests/components/dateRangePicker.test.js
kael89: It's actually a better practice to use the end value directly

packages/web-frontend/src/utils/periodGranularities.js
kael89: Oh man it would be so hard to catch this if not for unit tests... Mutability is one of the key points of moment.js criticism

packages/web-frontend/src/tests/components/dateRangePicker.test.js
Alastair-L: Should I create a map up the top?

```javascript
MIN_MOMENT_STRINGS = {
  [GRANULARITIES.SINGLE_DAY]: '20 Jan 2020',
  ...
};
```
Then 
```javascript
const startDate = MIN_MOMENT_STRINGS[key];
```

packages/web-frontend/src/utils/periodGranularities.js
Alastair-L: Yeah I should have checked that `.startOf` didn't mutate the original, but I just assumed it wouldn't 😬. Three cheers for https://github.com/beyondessential/tupaia-backlog/issues/1303

packages/web-frontend/src/tests/components/dateRangePicker.test.js
kael89: Yeap sounds good!

packages/web-frontend/src/utils/periodGranularities.js
```diff
@@ -115,6 +120,14 @@ export function roundStartEndDates(granularity, startDate = moment(), endDate =
   };
 }
 
+export const formatMomentAsString = (date, granularity, format) =>
```
kael89, line 123: I find `{{typeA}}to{{typeB}}`/`{{formatA}}to{{formatB}}` names quite intuitive for this sort of functionality:

```suggestion
export const momentToDateString = (date, granularity, format) =>
```

packages/web-frontend/src/utils/periodGranularities.js
```diff
@@ -115,6 +120,14 @@ export function roundStartEndDates(granularity, startDate = moment(), endDate =
...
+  granularity === 'week' || granularity === 'one_week_at_a_time'
```
kael89, line 124: Let's use the symbolic constants defined in this file instead of `week`/`one_week_at_at_time`

kael89 (changes requested):
Looks good overall! A couple of changes and it's ready to go

kael89:
@Alastair-L you'll also need to resolve conflicts

Alastair-L:
Thanks for the review @kael89. I'll get another PR up and then come back to this.

packages/web-frontend/src/utils/periodGranularities.js
Alastair-L: Me too now that I see it!

packages/web-frontend/src/tests/components/dateRangePicker.test.js
```diff
@@ -12,23 +12,67 @@ import { DateRangePicker } from '../../components/DateRangePicker';
...
+  [GRANULARITIES.MONTH]: 'Mar 2018',
+  [GRANULARITIES.SINGLE_MONTH]: 'Mar 2018',
+  [GRANULARITIES.QUARTER]: 'Q1 2018',
+  [GRANULARITIES.SINGLE_QUARTER]: 'Q1 2018',
+  [GRANULARITIES.YEAR]: '2018',
+  [GRANULARITIES.SINGLE_YEAR]: '2018',
+};
+
 describe('dateRangePicker', () => {
+  it('Has a MIN_DATE_PICKER_DATE consistent with tests', () => {
```
Alastair-L, line 65: This is more so that if it changes and this test fails, the error is easier to diagnose, as opposed to testing any functionality.

kael89: Good idea!

kael89 (approved):
Looks good!

Alastair-L:
Can I merge this, as it was testing passed here: https://github.com/beyondessential/tupaia/pull/1275#issuecomment-696518759, and since I have only merged dev and made those small changes? 

I will give it a smoke test after it builds in dev.

What do you think?

kael89:
@Alastair-L sounds good!

PR #1279, opened 2020-09-08
1242 support defaults arithmetic indicator

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1242

### DIFF
--- a/packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
+++ b/packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
@@ -30,6 +30,16 @@ describe('buildArithmetic', () => {
         { formula: 'A + B', aggregation: { A: 'MOST_RECENT' } },
         /B.* has no aggregation defined/,
       ],
+      [
+        'a data element not referenced in the formula but has a default',
+        { formula: 'A + B', aggregation: { A: 'MOST_RECENT', B: 'SUM' }, defaultValues: { C: 10 } },
+        /C.* is in defaultValues but not referenced in the formula/,
+      ],
+      [
+        'a default value must be a number',
+        { formula: 'A + 1', aggregation: { A: 'SUM' }, defaultValues: { A: '10' } },
+        /Default value for 'A' is not a number: '10'/,
+      ],
     ];
 
     it.each(testData)('%s', async (_, config, expectedError) =>
@@ -193,6 +203,17 @@ describe('buildArithmetic', () => {
           { organisationUnit: 'PG', period: '2019', value: 8.8 },
           { organisationUnit: 'PG', period: '2020', value: 65 },
         ],
+      ],
+      [
+        "defaults don't replace 0s",
+        {
+          formula: 'Z_To_2019 * 2',
+          aggregation: getAggregationForCodes(['Z_To_2019']),
+          defaultValues: {
+            Z_To_2019: 10,
+          },
+        },
+        [{ organisationUnit: 'TO', period: '2019', value: 0 }],
       ],
     ];
 
--- a/packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
+++ b/packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/camelcase */
 /**
  * Tupaia
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
@@ -162,6 +163,36 @@ describe('buildArithmetic', () => {
           aggregation: getAggregationForCodes(['C_ToPg_2019', 'D_To_20192020', 'E_Pg_20192020']),
         },
         [],
+      ],
+      [
+        'some combos include an undefined data element, but all defaults defined',
+        {
+          formula: 'C_ToPg_2019 + E_Pg_20192020',
+          aggregation: getAggregationForCodes(['C_ToPg_2019', 'E_Pg_20192020']),
+          defaultValues: {
+            C_ToPg_2019: 10,
+            E_Pg_20192020: 100,
+          },
+        },
+        [
+          { organisationUnit: 'TO', period: '2019', value: 103 },
+          { organisationUnit: 'PG', period: '2019', value: 8.8 },
+          { organisationUnit: 'PG', period: '2020', value: 65 },
+        ],
+      ],
+      [
+        'some combos include an undefined data element, but not all defaults defined',
+        {
+          formula: 'C_ToPg_2019 + E_Pg_20192020',
+          aggregation: getAggregationForCodes(['C_ToPg_2019', 'E_Pg_20192020']),
+          defaultValues: {
+            C_ToPg_2019: 10,
+          },
+        },
+        [
+          { organisationUnit: 'PG', period: '2019', value: 8.8 },
+          { organisationUnit: 'PG', period: '2020', value: 65 },
+        ],
       ],
     ];
 
--- a/packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
+++ b/packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
@@ -38,7 +38,7 @@ describe('buildArithmetic', () => {
       [
         'a default value must be a number',
         { formula: 'A + 1', aggregation: { A: 'SUM' }, defaultValues: { A: '10' } },
-        /Value for 'A' is not a number: '10'/,
+        /Value 'A' is not a number: '10'/,
       ],
     ];
 
--- a/packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
+++ b/packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
@@ -38,7 +38,7 @@ describe('buildArithmetic', () => {
       [
         'a default value must be a number',
         { formula: 'A + 1', aggregation: { A: 'SUM' }, defaultValues: { A: '10' } },
-        /Default value for 'A' is not a number: '10'/,
+        /Value for 'A' is not a number: '10'/,
       ],
     ];
 
--- a/packages/indicators/src/builders/buildArithmetic.ts
+++ b/packages/indicators/src/builders/buildArithmetic.ts
@@ -11,7 +11,7 @@ import {
   hasContent,
   isAString,
   isPlainObject,
-  assertAllValuesAreNumbers,
+  allValuesAreNumbers,
   constructIsEmptyOr,
 } from '@tupaia/utils';
 import { getAggregationsByCode, fetchAnalytics, validateConfig } from './helpers';
@@ -53,7 +53,7 @@ const configValidators = {
   aggregation: [hasContent, isPlainObject, assertAggregationIsDefinedForCodesInFormula],
   defaultValues: [
     constructIsEmptyOr(assertAllDefaultsAreCodesInFormula),
-    constructIsEmptyOr(assertAllValuesAreNumbers),
+    constructIsEmptyOr(allValuesAreNumbers),
   ],
 };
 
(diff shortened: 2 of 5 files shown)


packages/indicators/src/builders/buildArithmetic.ts
```diff
@@ -45,8 +49,20 @@ const fetchAnalyticClusters = async (
   const checkClusterIncludesAllElements = (cluster: AnalyticCluster) =>
     allElements.every(member => member in cluster.dataValues);
 
+  const replaceAnalyticValuesWithDefaults = (cluster: AnalyticCluster) => {
+    const returnDataValues = cluster.dataValues;
+    Object.keys(defaultValues).forEach(code => {
+      returnDataValues[code] = returnDataValues[code] || defaultValues[code];
```
edmofro, line 55: This will also replace 0's - be good to add a test case for this

packages/indicators/src/builders/buildArithmetic.ts
```diff
@@ -45,8 +49,20 @@ const fetchAnalyticClusters = async (
   const checkClusterIncludesAllElements = (cluster: AnalyticCluster) =>
     allElements.every(member => member in cluster.dataValues);
 
+  const replaceAnalyticValuesWithDefaults = (cluster: AnalyticCluster) => {
+    const returnDataValues = cluster.dataValues;
```
edmofro, line 53: Not great practice to mutate a parameter - even by assigning it to this variable, you're still mutating the original (so actually, the `return` becomes redundant)
```suggestion
    const returnDataValues = {...cluster.dataValues};
```

edmofro (changes requested):
Nice one - a couple of small things to fix up

packages/indicators/src/builders/buildArithmetic.ts
```diff
@@ -62,7 +78,12 @@ export const buildArithmetic: Builder = async input => {
   const { aggregator, config: configInput, fetchOptions } = input;
   const config = await validateConfig<ArithmeticConfig>(configInput, configValidators);
```
kael89, line 79: @Alastair-L  @edmofro Do you think it's worth adding validation for the `defaultValues` field? It should be

* either "not present", or
* an object with all keys included in the formula as data elements, and numeric values

By having a look at the code, it seems that wrong defatulValues keys will not result in unwanted behaviour:
```js
{
  formula: 'A + B',
  defaultValues: { A: 0, B: 0, C: 0 } // 'C' will eventually be ignored in `runArithmetic`
}
```

But may still be worth it given the existing validation is pretty thorough

packages/indicators/src/builders/buildArithmetic.ts
Alastair-L: 😬 Good catch, I didn't notice that

packages/indicators/src/builders/buildArithmetic.ts
Alastair-L: Good idea, I wasn't sure exactly how to do that, see below for my attempt.

packages/indicators/src/builders/buildArithmetic.ts
Alastair-L: Now does not replace 0s, there is a test case for this.

packages/indicators/src/builders/buildArithmetic.ts
```diff
@@ -27,14 +30,36 @@ const assertAggregationIsDefinedForCodesInFormula = (
...
+    if (typeof value !== 'number') {
+      throw new Error(`Default value for '${key}' is not a number: '${value}'`);
+    }
+  });
+};
+
 const configValidators = {
   formula: [hasContent, isAString],
   aggregation: [hasContent, isPlainObject, assertAggregationIsDefinedForCodesInFormula],
+  defaultValues: [assertAllDefaultsAreCodesInFormula, assertAllValuesAreNumbers],
```
Alastair-L, line 56: Is there a cleaner way to write these validators?

packages/indicators/src/builders/buildArithmetic.ts
```diff
@@ -45,8 +70,20 @@ const fetchAnalyticClusters = async (
   const checkClusterIncludesAllElements = (cluster: AnalyticCluster) =>
     allElements.every(member => member in cluster.dataValues);
 
+  const replaceAnalyticValuesWithDefaults = (cluster: AnalyticCluster) => {
+    const returnDataValues = { ...cluster.dataValues };
+    Object.keys(defaultValues).forEach(code => {
+      returnDataValues[code] = returnDataValues[code] ?? defaultValues[code];
```
Alastair-L, line 76: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

Alastair-L:
Wasn't sure who to request, so just requested both of you 😄

packages/indicators/src/builders/buildArithmetic.ts
edmofro: Nice :+1:

packages/indicators/src/builders/buildArithmetic.ts
```diff
@@ -30,9 +30,30 @@ const assertAggregationIsDefinedForCodesInFormula = (
...
+) => {
+  const variables = getVariables(formula);
+  Object.keys(defaultValues).forEach(code => {
+    if (!variables.includes(code)) {
+      throw new Error(`'${code}' is in defaultValues but not referenced in the formula`);
+    }
+  });
+};
+
+const assertAllValuesAreNumbers = (defaultValues: DefaultValuesSpecs = {}) => {
```
edmofro, line 45:
```suggestion
const assertAllDefaultValuesAreNumbers = (defaultValues: DefaultValuesSpecs = {}) => {
```

edmofro (approved):
Tiny nit, then lgtm! Preapproving.

packages/indicators/src/builders/buildArithmetic.ts
Alastair-L: I have gone with a different approach to keep that validator general. I would appreciate it if you could re-review 🙏

kael89: Good on you for being proactive! I would suggest a similar change 😆

packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
```diff
@@ -29,6 +30,16 @@ describe('buildArithmetic', () => {
...
       ],
+      [
+        'a data element not referenced in the formula but has a default',
+        { formula: 'A + B', aggregation: { A: 'MOST_RECENT', B: 'SUM' }, defaultValues: { C: 10 } },
+        /C.* is in defaultValues but not referenced in the formula/,
+      ],
+      [
+        'a default value must be a number',
+        { formula: 'A + 1', aggregation: { A: 'SUM' }, defaultValues: { A: '10' } },
+        /Value for 'A' is not a number: '10'/,
```
kael89, line 41: You need to update this error message for the tests to pass

edmofro (approved):
Love it!

packages/indicators/src/builders/buildArithmetic.ts
kael89: This is a great syntax!

packages/indicators/src/builders/buildArithmetic.ts
```diff
@@ -27,14 +36,31 @@ const assertAggregationIsDefinedForCodesInFormula = (
...
+    }
+  });
+};
+
 const configValidators = {
   formula: [hasContent, isAString],
   aggregation: [hasContent, isPlainObject, assertAggregationIsDefinedForCodesInFormula],
+  defaultValues: [
+    constructIsEmptyOr(assertAllDefaultsAreCodesInFormula),
+    constructIsEmptyOr(assertAllValuesAreNumbers),
```
kael89, line 56: That's really nice!

packages/utils/src/validation/validatorFunctions.js
```diff
@@ -87,6 +87,14 @@ export const isPlainObject = value => {
   }
 };
 
+export const assertAllValuesAreNumbers = object => {
+  Object.entries(object).forEach(([key, value]) => {
+    if (typeof value !== 'number') {
+      throw new Error(`Value '${key}' is not a number: '${value}'`);
```
kael89, line 93: For consistency with the rest of the validator functions, although we will change this at some point according to https://github.com/beyondessential/tupaia-backlog/issues/1201:

```suggestion
      throw new ValidationError(`Value '${key}' is not a number: '${value}'`);
```

packages/indicators/src/builders/buildArithmetic.ts
Alastair-L: 😆, perhaps you are finally rubbing off on me.

packages/utils/src/validation/validatorFunctions.js
```diff
@@ -87,6 +87,14 @@ export const isPlainObject = value => {
   }
 };
 
+export const assertAllValuesAreNumbers = object => {
```
kael89, line 90:
```suggestion
export const allValuesAreNumbers= object => {
```

`assert` is a more "correct" prefix in theory, but omitting it makes config declarations very readable. I think I prefer omitting it, but the actual reason I suggest this change is for consistency with existing validator function names.

packages/utils/src/validation/validatorFunctions.js
```diff
@@ -150,9 +158,9 @@ export const constructRecordExistsWithId = (modelOrDatabase, recordType) => asyn
   }
 };
 
-export const constructIsEmptyOr = validatorFunction => value => {
+export const constructIsEmptyOr = validatorFunction => (value, ...args) => {
```
kael89, line 161: This is clever! I suggest making it slightly more readable by making it explicit:

```suggestion
export const constructIsEmptyOr = validatorFunction => (value, object, key) => {
```

and similarly in line 163

kael89 (changes requested):
I love the new validation implementation! A few minor changes requested to maintain consistency with existing patterns.

Kudos for keeping code coverage at 100%!

packages/utils/src/validation/validatorFunctions.js
Alastair-L: That makes sense.

Alastair-L:
Incidentally, what do you think about:
```javascript
export const constructAllValuesAreType = type => object => {
  Object.entries(object).forEach(([key, value]) => {
    if (typeof value !== type) {
      throw new ValidationError(`Value '${key}' is not a number: '${value}'`);
    }
  });
};
```
and then used like:
`constructAllValuesAreType('number')`.

I think it's a bit overkill but interested to hear what you think.

packages/utils/src/validation/validatorFunctions.js
Alastair-L: This is more readable! I didn't realize validator functions had a common signature.

kael89:
> Incidentally, what do you think about:
> 
> ```js
> export const constructAllValuesAreType = type => object => {
> 

I think it's great! I would potentially implement this as

```js
const constructAllValuesAreType  = type => /* ... */
export const allValuesAreNumbers = constructAllValuesAreType('number');
```

so that if someone wants to check any fancy type in the future (eg `Moment`), we can export the underlying `constructAllValuesAreType` function.

`constructAllValuesAreType('number')` is still legit, but pehraps it makes sense to keep validators as readable as possible. I'll leave the final decision to you!

Only note: I'd suggest constructAllValuesAre**Of**Type (or construct~All~ValuesAreOfType to make it a bit shorter)

packages/utils/src/validation/validatorFunctions.js
kael89: It may not be obvious, but their function signature is a result of how they are called in `ObjectValidator.runValidators()`

packages/utils/src/tests/validation/validationFunctions.test.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { expect } from 'chai';
+
+import { allValuesAreNumbers } from '../../validation';
+
+describe('validationFunctions', () => {
```
kael89, line 10: Bonus points!!

kael89 (approved):
Fabuloso!!

PR #1280, opened 2020-09-08
Upgrade dev dependencies

kael89:
### Changes:
* Upgrade `winston` to v3.3.3 (minor version update)
* Upgrade `dotenv` to v8.2.0 (major version update - no breaking changes according to its [Changelog](https://github.com/motdotla/dotenv/blob/master/CHANGELOG.md))
* Moved common `devDependencies` in the root `package.json`
* Updated `devDependencies`, either automatically using `yarn upgrade` (minor/patch version updates), or manually when a new major version was available. Major version changes were done for packages that are clearly dev dependencies


packages/ui-components/package.json
```diff
@@ -38,18 +38,6 @@
...
-    "@storybook/addon-essentials": "^6.0.10",
-    "@storybook/react": "^6.0.10",
-    "@testing-library/jest-dom": "^5.11.0",
-    "@testing-library/react": "^10.4.5",
-    "@testing-library/user-event": "^12.0.11",
-    "babel-loader": "^8.1.0",
-    "babel-preset-react-app": "^9.1.2",
-    "cross-env": "^7.0.2",
-    "eslint-plugin-react-hooks": "^4.0.8",
-    "faker": "^4.1.0",
```
kael89, line 52: `faker` is already defined this package's `dependencies`

packages/meditrak-app/package.json
```diff
@@ -124,8 +124,6 @@
     "@babel/preset-flow": "^7.0.0",
     "@babel/runtime": "^7.4.2",
     "@react-native-community/eslint-config": "^0.0.3",
-    "eslint-plugin-import": "^2.18.0",
-    "eslint-plugin-jsx-a11y": "^6.2.3",
```
kael89, line 128: Our `eslint` config package imports those internally

packages/psss/package.json
```diff
@@ -39,7 +39,6 @@
     "react-hook-form": "^6.0.0",
     "react-redux": "^7.1.0",
     "react-router-dom": "^5.1.2",
-    "react-scripts": "3.4.1",
```
kael89, line 42: This is a dev dependency in reality

edmofro (approved):
I agree with all of the calls you've made here! Let me know if you'd like me to review line by line, which I haven't done (have looked at every file broadly, but not done all of the matching that would be required for a truly thorough review)

PR #1285, opened 2020-09-08
785 json editor

tcaiger:
### Changes:

- Add Json Editor plugin

There are a lot of possible configuration options and themes available. I have gone with the defaults apart from setting history which allows users to undo and redo changes.

---

### Screenshots:

Screen Shot 2020-09-09 at 10 12 45 AM
Screen Shot 2020-09-09 at 10 12 33 AM


packages/admin-panel/src/widgets/JsonEditor.js
```diff
@@ -4,40 +4,51 @@
...
-        const object = JSON.parse(value);
-        editorValue = JSON.stringify(object, null, 2);
-      } catch (e) {
-        editorValue = value;
-      }
-      break;
-    default:
-      editorValue = value.toString();
+  if (!value) {
+    return null;
```
edmofro, line 32: Is there a case where we want to allow creating a new json object, where previously the column was null?

packages/admin-panel/src/widgets/JsonEditor.js
```diff
@@ -4,40 +4,51 @@
...
-    />
+    <Container>
+      <Label gutterBottom>{label}</Label>
+      {/* Use json editor plugin. For configuration options @see https://github.com/vankop/jsoneditor-react */}
+      <Editor
+        history
+        value={editorValue}
+        onChange={json => onChange(inputKey, JSON.stringify(json))}
+      />
+      {secondaryLabel && <HelperText>{secondaryLabel}</HelperText>}
```
edmofro, line 50: I know these are below the fields in all other cases, but with the size of the json editor it might get missed, and/or miss a bit of context? What do you think?

edmofro (approved):
All looks good! Some UX questions, but very cool stuff overall and code is approved :+1:

packages/admin-panel/src/widgets/JsonEditor.js
tcaiger: No, it is in case the component gets mounted with no data in which case the json editor will throw an error. Another way to handle this would be to set value to an empty object which now I think about it might be better in case there is a null value from the db.

edmofro: Yeah I reckon that's best :+1:

packages/admin-panel/src/widgets/JsonEditor.js
tcaiger: hmm I tried it and it looks a bit funny. I think we leave them at the bottom for now

packages/admin-panel/src/widgets/JsonEditor.js
tcaiger: hmmm so the json-editor handles it's state internally so the value prop is actually just the default value. When the component first renders, the data hasn't come back form the server yet and is null, then when the server responds the editor will not update with the new data. The editor needs to be updated by user input. The null check fixes this issue but I guess there could be cases where this is a problem.

tcaiger: Looking through the database, all the empty values for json fields are `{}` which works fine. I'm not sure if that is a default somewhere or not.

edmofro: We use that as a default in most cases - sounds like maybe all, so that's good. There are some json strings stored in the db as text fields, and I'm not sure but we might also use the json editor to handle them, and I'm not sure whether they all have the same default. Can just keep it as is for now and log the bug if it comes up haha

PR #1287, opened 2020-09-09
Reinstate private field

kael89:
### Issue #:
[No issue]

I had removed the `private` field when open sourcing our codebase. This was a wrong interpretation of the field: it actually refers to whether the package can be published in the npm registry, not whether it is "Open source".

We really only need this for `meditrak-app`, since the package must be `private` for the `workspaces.nohoist` field in `meditrak-app/package.json` to work. But I thought we should keep using `private` in all `package.json`, since we wouldn't want our monorepo ending up in npm anyway.

### DIFF
--- a/packages/access-policy/package.json
+++ b/packages/access-policy/package.json
@@ -1,5 +1,6 @@
 {
   "name": "@tupaia/access-policy",
+  "private": true,
   "version": "3.0.0",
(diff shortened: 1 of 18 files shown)


Alastair-L (approved):
LGTM!

PR #1290, opened 2020-09-09
939 psss link to tupaia

tcaiger:
### Changes:

- Add preview tupaia button
- fix date picker regression bug
- fix storybook regression bug

---

### Screenshots:

Screen Shot 2020-09-09 at 6 05 08 PM

Screen Shot 2020-09-09 at 6 12 56 PM


packages/psss/src/components/Header.js
```diff
@@ -129,14 +142,19 @@ export const Header = ({ Title, back, ExportModal }) => {
...
-            <>
-              <LightOutlinedButton onClick={() => setIsModalOpen(true)} startIcon={<SaveAlt />}>
-                Export Data
-              </LightOutlinedButton>
-              <ExportModal isOpen={isModalOpen} handleClose={() => setIsModalOpen(false)} />
-            </>
-          )}
+          <div>
+            <TupaiaButton href="https://tupaia.org" startIcon={<TupaiaIcon />}>
+              Preview Tupaia
```
edmofro, line 147: Not sure about the wording "Preview". @enunan? It's more like "Visit" or "See" or "View"

edmofro (approved):
Just the wording question, but all the code looks good :-)

packages/psss/src/components/Header.js
tcaiger: Also another good question that @Klaus-S rasied. Should we open Tupaia.org in a new tab or just take them to it in the existing tab? It's Easy enough to implement either way.

edmofro: New tab IMO

PR #1291, opened 2020-09-09
1251 create weather service

IgorNadj:
### Changes:
- Add WeatherService
- Add weather-api package
- Add geography column casting to Entity model

---

### Screenshots:

Screen Shot 2020-09-09 at 6 21 10 pm


IgorNadj:
Marking as WIP because the database changes have broken tests. I spent a bit of time looking into it but not sure how to proceed.

Note to self:
- https://github.com/beyondessential/tupaia/blob/dev/packages/meditrak-server/src/tests/questionHooks.test.js#L203
- first two tests here, somehow the first test fails only if the second test is run? Error is 
  ```
{"name":"error","length":147,"severity":"ERROR","code":"XX000","hint":"\"{\"\" <-- parse error at position 2 within geometry","file":"lwgeom_pg.c","line":"174","routine":"pg_parser_errhint","level":"error"}
   ```
- Execution path is questionHooks.test.js -> ... -> packages/database Entity.js #updatePointCoordinatesFormatted.
- I have tried running the same SQL directly against the database but I can't reproduce the error.

edmofro:
Regarding https://github.com/beyondessential/tupaia/pull/1291#issuecomment-690901153, I'll also need to solve this for the refactor I'm working on. I imagine it might be because the dummy entity records that are generated and saved to the db may have invalid geo data. Would really appreciate if you can solve that problem here, and I can copy the fix over on my branch.

A possible solution: adding either `null` or some static geo data against those fields in `upsertDummyRecord.js`, e.g. (+'s show new lines)

```
const CUSTOM_DUMMY_VALUES = {
  [ENTITY]: {
    type: 'facility', // default testing entity should be facility
    country_code: 'DL', // use demo land by default in testing
+  point: null,
+  region: null,
+  bounds: null,
  },
  [SURVEY_RESPONSE]: {
    timezone: 'Africa/Abidjan', // equivalent to GMT
  },
};
```

This is speculation, not confident this will be the issue.

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
```diff
@@ -0,0 +1,123 @@
+import { expect } from 'chai';
+import { DateSanitiser } from '../../../services/weather/DateSanitiser';
+import { mockNow, resetMocks } from './testutil';
+
+describe('DateSanitiser', () => {
+  beforeEach(() => {
+    mockNow(1549360800 * 1000); // (2019-02-05 10:00 UTC)
```
edmofro, line 7: Mocking now certainly makes the tests nice and easy to follow!

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
```diff
@@ -0,0 +1,123 @@
...
+  describe('limits on latest data', () => {
+    /*
+     * Today's data is incomplete, e.g. we don't know the total rainfall for the day until
+     * the day is over. So we cannot return data for today, instead we return data for yesterday.
+     *
+     * There may be a possibility of using the forecast API to get today's forecast max temp / precip,
+     * but this will be covered by #1250
+     *
+     * Note for latest available dates:
+     *  - for WeatherBit API, end date today means at 00:00:01 today, which does not cover any weather data for today
```
edmofro, line 64: Right! This answers my earlier question. That's a helpful comment, thanks.

I don't think the date sanitiser should be "deceptive", if you know what I mean. Sure, it may constrain the endpoints so that they fall within the accepted range, but I don't think it should "shift" the date range that is being requested. I like that you've got a separate and clear code path for requesting yesterday's data in the service, when no date range is provided. However, if a client chooses to select a date range that explicitly requests today, I think we should just return no results, or throw an error.

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
```diff
@@ -0,0 +1,123 @@
...
+
+    it('both start and end dates are before the earliest available data', () => {
+      const { startDate: actualStartDate, endDate: actualEndDate } = new DateSanitiser().sanitise(
+        '2017-02-10',
+        '2017-02-20',
+      );
+
+      expect(actualStartDate).to.equal('2018-02-05'); // earliest date
+      expect(actualEndDate).to.equal('2018-02-06'); // earliest + 1 day
+    });
```
edmofro, line 32: I think this case should return no data. If you've requested dates that all fall outside of the supported range, you get no data - that seems fine. On the other hand, getting data from a single date that you didn't choose seems like unexpected behaviour that would be hard to predict without knowing the internals

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
```diff
@@ -0,0 +1,123 @@
...
+
+    it('only start date is before the earliest available data', () => {
+      const { startDate: actualStartDate, endDate: actualEndDate } = new DateSanitiser().sanitise(
+        '2017-02-10',
+        '2019-01-01',
+      );
+
+      expect(actualStartDate).to.equal('2018-02-05'); // earliest date
+      expect(actualEndDate).to.equal('2019-01-01');
+    });
```
edmofro, line 42: I like this case!

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
```diff
@@ -0,0 +1,123 @@
...
+
+    it('only start date is before earliest, end date is already earliest', () => {
+      const { startDate: actualStartDate, endDate: actualEndDate } = new DateSanitiser().sanitise(
+        '2017-02-10',
+        '2018-02-06',
```
edmofro, line 47: From the comment in the `it`, I thought this would be 
```suggestion
        '2018-02-05',
```
I.e., the end date is the earliest date

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
```diff
@@ -0,0 +1,123 @@
...
+    it('only end date is after latest, start date is already latest', () => {
+      const { startDate: actualStartDate, endDate: actualEndDate } = new DateSanitiser().sanitise(
+        '2019-02-04',
+        '2019-07-20',
+      );
+
+      expect(actualStartDate).to.equal('2019-02-04'); // (same as input) latest - 1 day
+      expect(actualEndDate).to.equal('2019-02-05'); // latest date
+    });
+  });
```
edmofro, line 96: All the same comments as with `startDate`

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
```diff
@@ -0,0 +1,123 @@
...
+
+    it('does not push end date beyond the latest available data limit', () => {
+      const { startDate: actualStartDate, endDate: actualEndDate } = new DateSanitiser().sanitise(
+        '2019-02-05',
+        '2019-02-05',
+      );
+
+      expect(actualStartDate).to.equal('2019-02-04'); // pushed back
+      expect(actualEndDate).to.equal('2019-02-05'); // latest date
+    });
```
edmofro, line 121: Not a fan of this case

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
```diff
@@ -0,0 +1,123 @@
...
+     */
+    it('gives a midnight-midnight range when a single date is provided', () => {
+      const { startDate: actualStartDate, endDate: actualEndDate } = new DateSanitiser().sanitise(
+        '2019-01-05',
+        '2019-01-05',
+      );
+
+      expect(actualStartDate).to.equal('2019-01-05');
+      expect(actualEndDate).to.equal('2019-01-06'); // pushed out
+    });
```
edmofro, line 111: This case seems good - but also seems like what we should be doing with all end dates, whether it's a single date case or not! Our other apis are inclusive, and the end client (web-config-server) should be able to expect predictability across data services.

packages/data-broker/src/services/weather/DateSanitiser.js
```diff
@@ -0,0 +1,70 @@
...
+        endDate = moment(startDate)
+          .add(1, 'day')
+          .format(DATE_FORMAT);
+      } else {
+        // edge case, on the limit, can't go beyond it
+        startDate = moment(endDate)
+          .subtract(1, 'day')
+          .format(DATE_FORMAT);
+      }
+    }
```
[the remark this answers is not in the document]
edmofro: This is now covered in the tests section.

packages/data-broker/src/services/weather/ApiResultTranslator.js
```diff
@@ -0,0 +1,98 @@
...
+    };
+  }
+
+  _apiResultToEvents(apiResult, entity) {
+    const events = [];
+
+    for (const entry of apiResult.data) {
+      const event = {
+        organisationUnit: entity.code,
+        period: entry.datetime.replace('-', '').replace('-', ''),
```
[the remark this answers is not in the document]
edmofro: Hmm actually...our events generally include times as well as dates. We could add a time of 23:59, considering the data is for the day just been?

packages/data-broker/src/tests/services/weather/WeatherService.test.js
```diff
@@ -0,0 +1,234 @@
...
+
+      const mockApi = createWeatherApiStub(mockApiResponse);
+
+      const service = new WeatherService(mockModels, mockApi);
+
+      const actual = await service.pull(
+        [
+          {
+            model: {},
+            id: 'SOME_DATA_GROUP_ID', // data group doesn't matter, it will return the data elements it knows about
```
edmofro, line 95: Which is maybe not a greeaat thing...

packages/data-broker/src/tests/services/weather/WeatherService.test.js
```diff
@@ -0,0 +1,234 @@
...
+        getMockOptionsArg(),
+      );
+
+      expect(actual.results).to.deep.equal([
+        {
+          organisationUnit: 'MELB',
+          PRECIP: 23.6,
+          MAX_TEMP: 29.8,
+          MIN_TEMP: 24,
+          period: '20200820',
```
edmofro, line 112: I think the format of events is more like
```suggestion
          event: 'someId',
          orgUnit: 'MELB',
          orgUnitName: 'Melbourne',
          dataValues: {
            PRECIP: 23.6,
            MAX_TEMP: 29.8,
            MIN_TEMP: 24,
          },
          eventDate: '2020-08-20T23:59:59',
```

packages/data-broker/src/tests/services/weather/WeatherService.test.js
```diff
@@ -0,0 +1,234 @@
...
+          MAX_TEMP: 6,
+          MIN_TEMP: 7,
+          period: '20200821',
+        },
+      ]);
+    });
+  });
+
+  describe('gets current weather', () => {
+    it("calls the api with yesterday's date when no dates are provided (simple)", async () => {
```
edmofro, line 126: Have you run eslint and prettier? Would expect these to be single quotes

edmofro (changes requested):
Really nicely put together! Architecture is great, the tests are super helpful (including to understand everything), and the code is all very clear and well written 🙌 

I have many specific points, which are all relatively small, and am happy to talk through any of them. Some of the main points are:
- I think the responsibility for knowing that we should fetch yesterday's data should be pushed forward, maybe even into the front-end - so this PR should end up with no mention of the word "current". I don't mind the "no dates specified -> yesterday's data", but I think it should be called yesterday rather than current, and I don't think it should return data outside of the explicitly specified dates
- We should +1 the end date on every request, given that clients generally fetch inclusively, but the weather api is inclusive for start date, but exclusive for end date
- The events section is formatted differently to other events. Are we actually using the event fetch format? It's not 100% necessary to implement both.

packages/database/src/migrations/20200820022356-AddEntityTimezone-modifies-schema.js
```diff
@@ -0,0 +1,28 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function(db) {
+  await db.runSql(`ALTER TABLE entity ADD COLUMN timezone VARCHAR(255) DEFAULT NULL;`);
```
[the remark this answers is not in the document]
IgorNadj: Ah I didn't know that about postgres, good to know, thanks.

packages/database/src/migrations/20200820022356-AddServiceTypeWeather-modifies-schema.js
```diff
@@ -0,0 +1,28 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function(db) {
+  await db.runSql(`ALTER TYPE public.service_type ADD VALUE IF NOT EXISTS 'weather';`);
```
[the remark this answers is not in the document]
IgorNadj: It seemed to work, but I have changed it anyway.

packages/data-broker/src/tests/services/weather/WeatherService.test.js
IgorNadj: Yeah? I don't really understand data groups fully yet, what do you recommend instead?

packages/data-broker/src/tests/services/weather/WeatherService.test.js
IgorNadj: Yeah my IDE changed that from `'abc \' def'` to `"abc ' def"`. Does that happen in yours as well? Or I can find how to turn it off.

edmofro: Oh I missed the `'` in there - this is fine!

packages/data-broker/src/tests/services/weather/WeatherService.test.js
edmofro: I'd go with retrieving the data elements attached to the data group via the join table. May be a precedent for this in one of the other services

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+   * @private
+   */
+  async _getHistoricWeatherDataForEntities(entities, dateRangeByEntityCode) {
+    // Run requests in parallel for performance
+    const getDataForEntity = async entity => {
+      const apiResult = await this.api.historicDaily(
+        entity.pointLatLon().lat,
+        entity.pointLatLon().lon,
+        dateRangeByEntityCode[entity.code].startDate,
+        dateRangeByEntityCode[entity.code].endDate,
```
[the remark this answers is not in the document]
IgorNadj: Ah yeah that's what I had before, but even with the same request dates, each entity may have a different local start/end date because of different timezones.

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Fetch API data and return in format of events/analytics
+   *
+   * @param EntityType[] entities
+   * @param string type
+   * @param string[] dataElementCodes
+   * @returns {Promise<Object.<results: Object[], metadata: Object>>}
+   * @private
+   */
+  async _getCurrentWeather(entities, apiResultTranslator) {
```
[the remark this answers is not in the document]
IgorNadj: Discussed, this will pull down today's forecast weather so will be accurate

packages/database/src/migrations/20200820022356-AddEntityTimezone-modifies-schema.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
```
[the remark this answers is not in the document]
IgorNadj: Yeah good idea, created https://github.com/beyondessential/tupaia-backlog/issues/1285

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
IgorNadj: Discussed, using forecast api and different data elements for current day, so DateSanitiser no longer needs to be sneaky

packages/data-broker/src/services/weather/WeatherService.js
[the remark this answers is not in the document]
edmofro: Oh of course! That makes total sense

packages/data-broker/src/tests/services/weather/WeatherService.test.js
IgorNadj: Sounds good. Q) Do you think `eventDate: '2020-08-20T23:59:59',` makes sense having the time component? The most specific time we can get for the event is at the day level, max temp / min temp are reached at different times, so it might be confusing having a time component in there at 23:59:59. Thoughts?

edmofro: That's a good point about them being reached at different times during the day. The very end of the day, i.e. 23:59:59 , still makes sense to me though - it's at that point that we know we can lock in the max and min. And I do think we should have a time component for consistency, in case a downstream function expects it. Valid questions to raise though, and @kael89 (who has put a lot of thought into the difference between event and analytic formats) may have additional comments

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
IgorNadj: Sorted by other comments

packages/data-broker/src/tests/services/weather/testutil.js
```diff
@@ -0,0 +1,13 @@
+import sinon from 'sinon';
+
+export const mockNow = whenIsNow => {
+  if (!whenIsNow) {
+    whenIsNow = 1549360800000; // 2019-02-05T10:00:00.000Z
+  }
+
+  sinon.useFakeTimers(whenIsNow);
+};
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/database/src/migrations/20200820022356-AddEntityTimezone-modifies-schema.js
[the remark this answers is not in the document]
IgorNadj: Updated

packages/database/src/migrations/20200820022356-AddServiceTypeWeather-modifies-schema.js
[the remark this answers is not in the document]
IgorNadj: Updated

packages/database/src/migrations/20200910004911-AddWeatherDataElements-modifies-data.js
```diff
@@ -0,0 +1,35 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function(db) {
+  for (const code of ['PRECIP', 'MAX_TEMP', 'MIN_TEMP']) {
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/weather-api/src/WeatherApi.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * @param string endpoint
+   * @param {} params
+   * @returns {Promise<{}>}
+   * @private
+   */
+  async _fetch(endpoint, params) {
+    params.key = this.apiKey;
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+    this.api = api;
+    this.dateSanitiser = new DateSanitiser();
+  }
+
+  /**
+   * Note, if no period specified will return current weather
+   * @inheritDoc
+   */
+  async pull(dataSources, type, options = {}) {
+    const dataElementCodes = this._dataElementCodes(type, dataSources);
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  /**
+   * Get the entities we want weather information about.
+   *
+   * @param string[] code
+   * @returns {EntityType[]}
+   * @private
+   */
+  async _getWeatherEntitiesByCode(code) {
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+    const entities = await this.models.entity.find({ code: code });
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Note, if no period specified will return current weather
+   * @inheritDoc
+   */
+  async pull(dataSources, type, options = {}) {
+    const dataElementCodes = this._dataElementCodes(type, dataSources);
+
+    const entities = await this._getWeatherEntitiesByCode(options.organisationUnitCodes);
+
+    const requestType = this._getRequestType(type);
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+   * @param Object.<entityCode: string: Object.<startDate: string, endDate: string> dateRangeByEntityCode
+   * @returns {Promise<Object.<entityCode: string: apiResult: Object>>}
+   * @private
+   */
+  async _getHistoricWeatherDataForEntities(entities, dateRangeByEntityCode) {
+    // Run requests in parallel for performance
+    const getDataForEntity = async entity => {
+      const apiResult = await this.api.historicDaily(
+        entity.pointLatLon().lat,
+        entity.pointLatLon().lon,
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+        dateRangeByEntityCode[entity.code].startDate,
+        dateRangeByEntityCode[entity.code].endDate,
+      );
+      return {
+        entityCode: entity.code,
+        apiResult: apiResult,
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/data-broker/src/services/weather/DateSanitiser.js
```diff
@@ -0,0 +1,70 @@
...
+
+    // single day requested
+    if (startDate === endDate) {
+      if (startDate <= latestStartDate) {
+        // happy path, push end date out
+        endDate = moment(startDate)
+          .add(1, 'day')
+          .format(DATE_FORMAT);
+      } else {
+        // edge case, on the limit, can't go beyond it
```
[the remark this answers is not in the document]
IgorNadj: Covered by change to using forecast data

packages/data-broker/src/services/weather/ApiResultTranslator.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+export class ApiResultTranslator {
+  /**
+   * @param entities
+   * @param string requestType - one of 'events' | 'analytics'
+   * @param dataElementCodes
+   */
+  constructor(entities, requestType, dataElementCodes) {
+    this.entities = entities;
+    this.requestType = requestType;
```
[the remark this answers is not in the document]
IgorNadj: Updated

packages/data-broker/src/services/weather/ApiResultTranslator.js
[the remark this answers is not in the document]
IgorNadj: Covered in discussion below

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
IgorNadj: Agreed, updated

packages/data-broker/src/tests/services/weather/WeatherService.test.js
kael89: I agree with Edwin's points:

✔️ we should have a time for consistency. Possibly we should also include a timezone - I'm gonna check this now
✔️ we should use the most accurate time to our knowledge. At this use case, the last minute of the day makes sense, since the event reflects a summary of the day

We haven't been using time till now, but you never know in the future. It's also to be consistent for predictable field parsing etc

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
IgorNadj: Updated

packages/data-broker/src/tests/services/weather/DateSanitiser.test.js
IgorNadj: Updated

packages/data-broker/src/tests/services/weather/WeatherService.test.js
IgorNadj: Updated

packages/data-broker/src/tests/services/weather/WeatherService.test.js
kael89: I confirmed that no timezone info is included in event dates

IgorNadj: Updated PR with 23:59:59

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,236 @@
...
+  _dataElementCodes(type, dataSources) {
+    let dataElementCodes = null;
+    if (type === this.dataSourceTypes.DATA_ELEMENT) {
+      // single data element requested
+      dataElementCodes = [dataSources[0].code]; // FIXME: bad
+    } else if (type === this.dataSourceTypes.DATA_GROUP) {
+      // all data elements requested
+      dataElementCodes = SUPPORTED_DATA_ELEMENT_CODES;
+    }
+    return dataElementCodes;
```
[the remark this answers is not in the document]
IgorNadj: Updated code to match your suggestion.

Removed SUPPORTED_DATA_ELEMENT_CODES, using group instead

packages/data-broker/src/services/weather/WeatherService.js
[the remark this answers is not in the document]
IgorNadj: Updated, uses forecast api instead

IgorNadj:
@edmofro Updated PR with fixes as comments above, plus:

- "Current weather" now uses the forecast api with data elements WTHR_FORECAST_*
- Timezones no longer needed, removed from Entity

Last TODO:
- Fix tests around database changes

packages/data-broker/src/services/weather/ApiResultTranslator.js
```diff
@@ -51,24 +58,33 @@ export class ApiResultTranslator {
...
-  _apiResultToEvents(apiResult, entity) {
+  apiResultToEvents(apiResult, entity) {
     const events = [];
 
     for (const entry of apiResult.data) {
+      const eventDate = `${entry.datetime}T23:59:59`; // time is 23:59:59 to represent the complete day
       const event = {
-        organisationUnit: entity.code,
-        period: entry.datetime.replace('-', '').replace('-', ''),
+        event: `weather_${entity.code}_${entry.datetime}`,
```
edmofro, line 78: Nice choice of event id :-)

IgorNadj:
Have fixed tests. Removing WIP. 

Ready to re-review @edmofro :) I've left the original commits as is so the changes should just be the new commits.

packages/database/src/modelClasses/Entity.js
```diff
@@ -131,19 +131,78 @@ class EntityType extends DatabaseType {
...
 
   types = ENTITY_TYPES;
 
   isOrganisationUnitType = type => Object.values(ORG_UNIT_ENTITY_TYPES).includes(type);
 
+  getColumnSpecs = tableAlias => {
+    return this.buildColumnSpecs(tableAlias, false);
+  };
+
+  buildColumnSpecs = tableAlias => {
```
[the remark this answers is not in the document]
IgorNadj: Happy to change, but thought I'd leave this because it's going to be changed in your upcoming PR

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,267 @@
...
+    });
+
+    const metadata = [];
+    for (const dataElement of dataElements) {
+      metadata.push({
+        code: dataElement.code,
+        name: dataElement.name,
+      });
+    }
+    return Promise.resolve(metadata);
```
edmofro, line 89:
```suggestion
    return metadata;
```
I don't understand the need for the Promise.resolve, but I'm probably just missing something

packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -0,0 +1,267 @@
...
+
+  /**
+   * Fetch API data and return in format of events/analytics
+   *
+   * @param EntityType[] entities
+   * @param string type
+   * @param string[] dataElementCodes
+   * @param options
+   * @returns {Promise<Object.<results: Object[], metadata: Object>>}
+   * @private
```
edmofro, line 207: jsdoc out of date

edmofro (approved):
This is fantastic 🙌 

I have a few tiny suggestions, but it's approved either way. Thank you for seeing this through all of the initial requested changes :-)

edmofro (approved):
🎉  🚀  🙌

PR #1295, opened 2020-09-10
Add unit test for survey responses permission handling

EMcQ-BES:
### Changes:

- Added assertSurveyResponsesPermission.test.js


packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
```diff
@@ -0,0 +1,144 @@
...
+    const adminPermissionGroup = await findOrCreateDummyRecord(models.permissionGroup, {
+      name: 'Admin',
+    });
+    const donorPermissionGroup = await findOrCreateDummyRecord(models.permissionGroup, {
+      name: 'Donor',
+    });
+    const vanuatuCountry = await findOrCreateDummyRecord(models.country, { code: 'VU' });
+    const laosCountry = await findOrCreateDummyRecord(models.country, { code: 'LA' });
+    const vanuatuEntity = await findOrCreateDummyRecord(models.entity, {
+      id: generateTestId(),
```
chris-bes, line 40:
```suggestion
```
Hmm I think you don't have to pass the id here. Any functions for creating/updating dummy records will generate the test id by default. Same for creating dummy surveys/survey responses below

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
```diff
@@ -0,0 +1,144 @@
...
+      },
+    ]);
+    surveyResponses = surveyResponseModels.map(srm => srm.surveyResponse);
+  });
+
+  describe('filterSurveyResponsesByPermissions()', async () => {
+    it('Should filter any survey responses that users do not have access to to the survey of', async () => {
+      //Remove the permission of VU to have insufficient permissions to some surveys
+      const policy = {
+        // VU: ['Admin', 'Donor'],
```
chris-bes, line 102: I did the same in my code and Edwin suggested me to changed this to

```suggestion
        VU: [/*'Admin'*/ 'Donor'],
```
Which I think makes it a better test case because it can check the case of having at least some permissions to the survey country but not sufficient for the permission group of the survey.

Same goes to other test cases

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
```diff
@@ -0,0 +1,144 @@
...
+      const accessPolicy = new AccessPolicy(policy);
+      const result = await filterSurveyResponsesByPermissions(
+        accessPolicy,
+        surveyResponses,
+        models,
+      );
+
+      // Should have dropped all but [laos/Admin] survey response
+      expect(result.length).to.equal(1);
+      expect(result[0].id).to.equal(laosAdminResponseId);
```
chris-bes, line 114:
```suggestion
      expect(results.map(r => r.id)).to.deep.equal([laosAdminResponseId]);
```

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
```diff
@@ -0,0 +1,144 @@
...
+        surveyCode: 'TEST_SURVEY_2',
+        entityCode: laosEntity.code,
+        submission_time: '2020-01-31T09:00:00Z',
+        answers: [],
+      },
+    ]);
+    surveyResponses = surveyResponseModels.map(srm => srm.surveyResponse);
+  });
+
+  describe('filterSurveyResponsesByPermissions()', async () => {
```
chris-bes, line 98: Maybe add another test case that users have access to all the passed in survey responses and this function doesn't filter any thing?

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
```diff
@@ -0,0 +1,144 @@
...
+    const laosCountry = await findOrCreateDummyRecord(models.country, { code: 'LA' });
+    const vanuatuEntity = await findOrCreateDummyRecord(models.entity, {
+      id: generateTestId(),
+      code: 'VU_TEST',
+      country_code: vanuatuCountry.code,
+    });
+    const laosEntity = await findOrCreateDummyRecord(models.entity, {
+      id: generateTestId(),
+      code: 'LA_TEST',
+      country_code: laosCountry.code,
```
chris-bes, line 47:
```suggestion
      country_code: 'LA',
```

Probably no need to fetch the country

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
```diff
@@ -0,0 +1,144 @@
...
+      const policy = {
+        // VU: ['Admin', 'Donor'],
+        // LA: ['Admin', 'Donor'],
+        DL: ['Admin'],
+      };
+      const accessPolicy = new AccessPolicy(policy);
+
+      // Should only have permission for [Laos/Admin]
+      expect(() => assertSurveyResponsePermissions(accessPolicy, surveyResponses[0], models)).to
+        .throw;
```
chris-bes, line 141:
```suggestion
      expect(() => assertSurveyResponsePermissions(accessPolicy, models, surveyResponses[0], )).to
        .throw;
```

I think the parameters are in the wrong order? `assertSurveyResponsePermissions` function signature is
`export const assertSurveyResponsePermissions = async (accessPolicy, models, surveyResponse)`

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
```diff
@@ -0,0 +1,144 @@
...
+  describe('assertSurveyResponsePermissions()', async () => {
+    it('Sufficient permissions: Should return true if users have access to the survey permission group in the entity country', async () => {
+      const accessPolicy = new AccessPolicy(DEFAULT_POLICY);
+      const result = await assertSurveyResponsePermissions(
+        accessPolicy,
+        surveyResponses[0],
+        models,
+      );
+
+      expect(result).to.true;
```
chris-bes, line 127: I wonder how could this return true as I think the parameters passed in to `assertSurveyResponsePermissions` are in the wrong order? The function signature is `export const assertSurveyResponsePermissions = async (accessPolicy, models, surveyResponse)`

chris-bes (changes requested):
Nice work! I just have a couple minor suggestions

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
chris-bes: Also if users have BES Admin access to any countries, it should have access to all the survey responses without any restriction. So would be great if we can add test cases for that scenario too (I missed this in my code too, just added it recently)

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
EMcQ-BES: You're totally right. I think I had some local changes while I was first running these. I've fixed it up.

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
EMcQ-BES: Added test cases for all permissions and for BES Admin permissions.

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
EMcQ-BES: Changed

packages/meditrak-server/src/tests/routes/GETSurveyResponses/assertSurveyResponsesPermission.test.js
EMcQ-BES: That situation is already being tested in this test.

VU has no permissions.
LA has permissions for one of the surveyResponses but not the other.

I thought that was a reasonable spread to show the filtering working correctly.

chris-bes (approved):
Looks good!

edmofro:
@EMcQ-BES this is getting a bit stale - how does it fit together with the other permissions work you've been doing? Can it be merged?

EMcQ-BES:
This whole thing needs to be redone as an endpoint test so I'm just going to close this PR since this is outdated.

PR #1299, opened 2020-09-10
Add CD3b_014b_C19_9 to TO_CD_Validation_CD3

nick-beyondessential:
[165](https://github.com/beyondessential/tupaia-backlog/issues/165) Configure CD3b Validation Report so answers to the new question "CD3b_014b_C19_9" show up in Result Status column

In This PR:
simple migration to add CD3b_014b_C19_9 to additional data config of TO_CD_Validation_CD3 report
(This is a drilldown report so unable to to use admin panel


packages/database/src/migrations/20200910043451-AddCovidResultsToCD3bValidationReport-modifies-data.js
```diff
@@ -0,0 +1,67 @@
...
+const newAdditionalData = [
+  'CD3b_008_A30_9',
+  'CD3b_009_A74_9',
+  'CD3b_010_A54_9',
+  'CD3b_011_A53_9',
+  'CD3b_012_Z21',
+  'CD3b_013_B16_9',
+  'CD3b_014_A01_0',
+  'CD3b_014a_A39_9',
+  'CD3b_014b_C19_9',
```
nick-beyondessential, line 40: Is this ok way of updating? Maybe better to do the jsonb equivalent of push and pop value?

nick-beyondessential, line 40: Is this an ok way of adding a value? Maybe there is a better jsonb push/pop equivalent?
Seemed easier than querying for existing value to modify for simple task..

kael89: An alternative could be fetching the config and storing it in a variable, then pushing the desired element at the desired position, and then stringify it and save it in the DB.

But what you've done here is totally fine! Again, the answer depends on whether you expect other changes to happen in the same part of the config in parallel, by another migration.

kael89 (approved):
Looks good!

packages/database/src/migrations/20200910043451-AddCovidResultsToCD3bValidationReport-modifies-data.js
nick-beyondessential: Thanks @kael89 I was looking around for away of inserting the array element natively kinda like jsonb_insert but not not having luck, can also use concat method inside the jsonb_set but starts getting tricky with deep nested json.
Fetching and pushing seemed excessive in this case.

kael89: Yeah, sometimes it's easier to just do

```js
  const { rows } = await db.runSql(
    `SELECT * FROM "dashboardReport" WHERE id = "TO_CD_Validation_CD3"`,
  );
  const { dataBuilderConfig } = rows[0];
  dataBuilderConfig.columns.CD3b_007.additionalData.push('CD3b_014b_C19_9');
  await db.runSql(
    `UPDATE "dashboardReport" SET "dataBuilderConfig" = '${JSON.stringify(
      dataBuilderConfig,
    )}' WHERE id = 'TO_CD_Validation_CD3'`,
  );
```

This is just an example, I haven't tested the actual code above but I have used a similar approach in other cases

PR #1315, opened 2020-09-14
Add Laos eoc historic weather dashboard

IgorNadj:
### Changes:

- Add Laos eoc historic weather dashboard

---

### Screenshots:
See backlog ticket

### DIFF
--- a/packages/database/src/migrations/20200914010424-AddLaosEocHistoricWeatherDashboard-modifies-data.js
+++ b/packages/database/src/migrations/20200914010424-AddLaosEocHistoricWeatherDashboard-modifies-data.js
@@ -1,0 +1,111 @@
+'use strict';
+
+import { insertObject } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const DASHBOARD_GROUP_CODE = 'LAOS_EOC_City_Weather';
+
+const DASHBOARD_REPORT = {
+  id: 'LA_EOC_Daily_Rainfall_And_Temperature',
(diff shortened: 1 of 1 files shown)


rohan-bes (approved):
Dashboard config looks really nice here! Good job.

I took a look at the dashboard screenshot in the ticket, and while it looks really nice, it suffers from an issue of there being no clarity as to which yAxis corresponds to which line. This dashboard is by no means the only dashboard that has this issue in Tupaia, however this one to me was particularly unclear as the values of each scale are so similar (ie. It seems equally plausible that the yearly temperature fluctuated between 40 and 20 degrees with 35mm peak daily rainfall, as if the yearly temperature fluctuated between 30 and 10 degrees with 55mm peak daily rainfall).

I took a look into how easy it would be to implement labels per yAxis, however it seems that it would overlap with some existing (broken) functionality so I think it's definitely out of the scope for this PR. I've raised a ticket to address this: https://github.com/beyondessential/tupaia-backlog/issues/1262

I'm approving this PR, however I really do this that we ought to implement the multiple yAxis labelling feature (or something similar) soon and apply that to this dashboard. How do you feel @IgorNadj?

IgorNadj:
@rohan-bes do you mean it's hard to see immediately which y axis is for rainfall and which is for temperature? If so yeah that's something I've noticed not just here but with these kinds of graphs out in the wild over the past few years... maybe a question for data people. I always struggle with these, maybe data people don't struggle so much?

rohan-bes:
> @rohan-bes do you mean it's hard to see immediately which y axis is for rainfall and which is for temperature? If so yeah that's something I've noticed not just here but with these kinds of graphs out in the wild over the past few years... maybe a question for data people. I always struggle with these, maybe data people don't struggle so much?

Yep, that's exactly the issue. In the ticket I raised there's a link to another ticket raised by @geofish76 (https://github.com/beyondessential/tupaia-backlog/issues/1079) who experienced the same confusion with another graph in Tupaia. I think labelling the axis is probably the easiest solution, though there may be better options out there...

IgorNadj:
Closing, will create a new PR because branch name has changed

IgorNadj:
New PR: https://github.com/beyondessential/tupaia/pull/1476

PR #1324, opened 2020-09-15
916 covid 19 laos dashboard

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/916

Did a small refactor of the recent work by @chris-bes to add the ability to use CHECK_CONDITION on a cell in a table. Thanks to him for making it easily extensible!

### Screenshots:
On card: https://github.com/beyondessential/tupaia-backlog/issues/916#issuecomment-692478900


packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfDataValues/getValuesByCell.js
```diff
@@ -68,7 +68,12 @@ export const getPercentageCountOfValuesByCell = (cells, results) => {
 export const getCalculatedValuesByCell = (cells, results) => {
   const calculatedValuesByCell = {};
   cells.forEach(cell => {
-    calculatedValuesByCell[cell.key] = calculateArithmeticOperationForAnalytics(results, cell);
+    if (typeof cell === 'string') {
```
Alastair-L, line 71: I use `typof cell === 'string'` a few times throughout this PR. I find it easier to read than `!cell.someProperty`, which is used some other places in the code. Do you agree or is there a reason that `typeof` is avoided?

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return sum;
+};
+
+const ARITHMETIC_OPERATORS = {
+  DIVIDE: divideValues,
+  SUBTRACT: subtractValues,
+};
+
+const performArithmeticOperation = (analytics, arithmeticConfig) => {
```
Alastair-L, line 34: This looks like it could be a candidate for `@beyondessential/arithmetic`! But out of scope for this PR.

kael89: We can probably use indicators now!

Alastair-L: True! Should I add an indicator builder for CHECK_CONDITION? Probably would have been simpler, I just didn't think of it.

kael89: We probably need to add support for an `analyticCount` builder (or similar), see the discussion starting in https://github.com/beyondessential/tupaia-backlog/issues/261#issuecomment-686868714.

It will take some time to properly implement this indicator. Until this happens, I suggest we keep your work here, and refactor those builders at a later stage, at which stage we should do an easy regression test

PR #1325, opened 2020-09-15
Fixes various entities in various tables for Samoa

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1273

### DIFF
--- a/packages/database/src/migrations/20200915040216-FixSamoaEntities-modifies-data.js
+++ b/packages/database/src/migrations/20200915040216-FixSamoaEntities-modifies-data.js
@@ -1,0 +1,144 @@
+'use strict';
+import { generateId } from '../utilities/generateId';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20200915040216-FixSamoaEntities-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+      NULL,
+      'WS',
+      NULL,
+      '{"dhis": {"isDataRegional": true}}',
+      '{}'
+    );
+
+    update survey_response set entity_id='${newFacilityEntityId}' where entity_id='${motootuaDistrictEntityId}';
+
+    delete from entity where id='${motootuaDistrictEntityId}';
```
edmofro, line 60:
```suggestion
    delete from entity where id in ('${motootuaDistrictEntityId}', '${tuamasagaDistrictEntityId}');
```
And can delete the line below. Similar suggestion for the 2x geo area lines

edmofro (changes requested):
Nice migration! It looks good, I have a small suggestion which is not actually necessary.

I've also commented on the original issue with an update to the approach I think we should take.

edmofro (approved):
She's a beauty

PR #1328, opened 2020-09-15
693 to fana dm htn use total pop

nick-beyondessential:
https://github.com/beyondessential/tupaia-backlog/issues/693
**Tonga Fanafana Ola Community Health (DM and HTN Prevalence) Dashboard Edit** 

This edit is to use total population for denominator as not all facities collect DM & HTN data.

In This PR:
- Allow `percentagesPerPeriod` dataBuilder to accept for each series denominator a config for `organisationUnitCode` and/or `includeAllResults` (which defaults to false)
- `organisationUnitCode` allows a code to overwrite query organisationUnitCode
- `includeAllResults` when set to true for denominator passes `includeAllDenominatorValues` = `true` to the `getMatchedNumeratorsAndDenominators` util which signals that the denominator values should be summed across all results for period rather than those matched with denominator results.
- a migration setting report `TO_CH_DM_HTN_Prevalence` databuilder config denominators to us `TO` (Tonga) and to use `includeAllResults` set to `true`


rohan-bes (commented):
Might weight to hear back about the data before finishing this review. It always saddens me a little bit when databuilders get more and more complex config, but I'm not sure if creating a new databuilder would be the right way to go either...

I'm not exactly sure what the denominator should be here, the total population of tonga? Perhaps we could consider creating an indicator for this?

edmofro:
@nick-beyondessential @rohan-bes can we get this one moving again?

packages/database/src/migrations/20200915121622-UpdateToFanaDmHtmPrevUseTotalPop-modifies-data.js
```diff
@@ -0,0 +1,161 @@
...
+      defaultValues: {
+        POP14: 0,
+        POP16AND18: 0,
+        POP20AND22: 0,
+        POP24AND26: 0,
+        POP28AND30: 0,
+        POP3234AND36: 0,
+      },
+    },
+  },
```
rohan-bes, line 47: Woohoo! New indicators 😄

packages/database/src/migrations/20200915121622-UpdateToFanaDmHtmPrevUseTotalPop-modifies-data.js
```diff
@@ -0,0 +1,161 @@
...
+  }));
+
+  const newConfig = { ...oldConfig, series: newSeries };
+
+  return db.runSql(`
+    update "dashboardReport"
+    set "dataBuilderConfig" = '${JSON.stringify(newConfig)}'
+    where "id" = '${reportId}';
+  `);
+};
```
rohan-bes, line 134: There are a couple more dashboards that could use these indicators btw:
TO_RH_Descriptive_FP01_02
TO_CH_DM_HTN_Incidence

Think we could update those too?

*edit*
Infact, are those dashboards suffering from the same miscalculation that `TO_CH_DM_HTN_Prevalence` has been?

packages/web-config-server/src/apiV1/utils/getMatchedNumeratorsAndDenominators.js
```diff
@@ -23,11 +24,32 @@ export const getMatchedNumeratorsAndDenominators = (
...
+    const periodValues = {};
+    Object.keys(results).forEach(resultKey => {
+      const resultPeriod = results[resultKey].period;
+      if (!periodValues[resultPeriod]) {
+        periodValues[resultPeriod] = results[resultKey].value;
+      } else {
+        periodValues[resultPeriod] += results[resultKey].value;
+      }
+    });
+    return periodValues;
```
rohan-bes, line 37: I was really hoping we'd be able to avoid having this logic here, but I understand that it's a bit tricky.

I think ideally what we'd be doing here is:
1. Specify a different aggregation type for just the denominator that adds up the values within the country (`SUM_PER_PERIOD_PER_ORG_GROUP with { dataSourceEntityType: 'facility', aggregationEntityType: 'country' }`)
2. Pass in the `denominator.organisationUnitCode` into the `getMatchedNumeratorsAndDenominators()` and have it correctly match all numerators to that organisationUnitCode.

This way we can leave the `sumValuePerPeriod` logic to the aggregator.

However, I can see in this databuilder that the aggregation types are hardcoded unfortunately... It would be possible to make them overridable via config but that is a tad ugly. There are only 4 dashboards that use this databuilder, so it's feasible that we could change them to specify their aggregation type for numerator/denominator (which would allow us to just have 1 databuilder instead of two).

What are your thoughts on this? Happy to be pushed back on if you feel for practical reasons it's exceeding the scope of this bug fix

packages/database/src/migrations/20200915121622-UpdateToFanaDmHtmPrevUseTotalPop-modifies-data.js
nick-beyondessential: Yes I think that makes sense.

packages/web-config-server/src/apiV1/utils/getMatchedNumeratorsAndDenominators.js
nick-beyondessential: I think I have to try it out before i have a strong opinion. Seems reasonable option to reduce the databuilder. Thanks for the suggestion. If it works out to be too obtuse then I consider pushing back :)

edmofro:
@nick-beyondessential what is still needed to get this finished up? Has been sitting open for a while now!

edmofro:
@nick-beyondessential is there still work needed on this?

edmofro:
@nick-beyondessential this has gone very stale...is there a blocker on it?

nick-beyondessential:
@edmofro updating it now, will close this PR in favour of fresh branch

PShoe:
@nick-beyondessential  close this one when you have a sec, and maybe add a link to the new PR?

edmofro:
@nick-beyondessential could you please close this down if it's been replaced by a new PR?

PR #1331, opened 2020-09-16
Add SUM_UNTIL_THIS_DAY aggregation

kael89:
### Issue #:
Addresses https://github.com/beyondessential/tupaia-backlog/issues/261#issuecomment-692453832

### Changes:
* Add a new `SUM_UNTIL_THIS_DAY` aggregation type
* Use it to fix the bug


packages/database/src/migrations/migrationData/20200826215113-UseIndicatorsInStriveVisualisations.js
```diff
@@ -22,8 +22,8 @@ export const indicators = {
     config: {
       formula: 'MRDT_POS_CONSULT_WEEKLY / SSWT1072',
       aggregation: {
-        MRDT_POS_CONSULT_WEEKLY: 'SUM',
-        SSWT1072: ['FINAL_EACH_WEEK', 'SUM'],
+        MRDT_POS_CONSULT_WEEKLY: 'SUM_UNTIL_THIS_DAY',
+        SSWT1072: ['FINAL_EACH_WEEK', 'SUM_UNTIL_THIS_DAY'],
```
kael89, line 26: This migration has not run in `dev`, so it's simpler to update it and re-deploy the feature instance

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.js
```diff
@@ -3,27 +3,46 @@
...
     // If there are no matching response elements already being returned, add it
-    if (indexOfEquivalentResponseElement < 0) {
-      summedAnalytics.push(responseElement);
+    if (i < 0) {
+      summedAnalytics.push({ ...analytic });
     } else {
-      summedAnalytics[indexOfEquivalentResponseElement].value += responseElement.value;
+      summedAnalytics[i].value += analytic.value;
     }
   });
```
kael89, line 26: I just changed the names here... I could also refactor the logic to make it a bit more performant (using a `${dataElement}_${organisationUnit}` key). Given the scope of my changes + how critical this functionality is, I thought I should apply a "don't fix it if it's not broken" approach in this specific case

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.js
```diff
@@ -3,27 +3,46 @@
...
+        analytic.organisationUnit === otherAnalytic.organisationUnit,
     );
     // If there are no matching response elements already being returned, add it
-    if (indexOfEquivalentResponseElement < 0) {
-      summedAnalytics.push(responseElement);
+    if (i < 0) {
+      summedAnalytics.push({ ...analytic });
     } else {
-      summedAnalytics[indexOfEquivalentResponseElement].value += responseElement.value;
+      summedAnalytics[i].value += analytic.value;
```
kael89, line 24: It may not be so obvious by the existing code but we were actually mutating the input here! That's why I now `spread` the analytic in line 22.

This may not have come up as an issue before because more probably we don't reuse the mutated input. But it did came up as an issue during tests, where the same `analytics` input is passed in multiple test cases.

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
```diff
@@ -0,0 +1,127 @@
...
+    ['BCD01', 'TO', '201801', 0.1],
+    ['BCD01', 'TO', '201901', 0.01],
+    // BCD01 - PG
+    ['BCD01', 'PG', '201601', 2],
+    ['BCD01', 'PG', '201801', 0.2],
+    ['BCD01', 'PG', '201901', 0.02],
+    // BCD02 - TO
+    ['BCD02', 'TO', '201801', 4],
+    ['BCD02', 'TO', '202001', 0.44],
+  ]);
```
kael89, line 32: This is a new analytics definition style (`arrayAnalytics`). It's not used in actual code but I find it very convenient for tests.

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
```diff
@@ -0,0 +1,127 @@
...
+  it('single analytic', () => {
+    expect(sumAcrossPeriods([ANALYTICS[0]])).to.have.same.deep.members([ANALYTICS[0]]);
+  });
+
+  it('multiple analytics', () => {
+    expect(sumAcrossPeriods(ANALYTICS)).to.have.same.deep.members(
+      arrayToAnalytics([
+        ['BCD01', 'TO', '201601', 1.11],
+        ['BCD01', 'PG', '201601', 2.22],
+        ['BCD02', 'TO', '201801', 4.44],
```
kael89, line 59: The tests revealed a flaw in the current functionality!! We should use the max period among the summed analytics. It should be:

```suggestion
        ['BCD01', 'TO', '201901', 1.11],
        ['BCD01', 'PG', '201901', 2.22],
        ['BCD02', 'TO', '202001', 4.44],
```

I decided to keep this PR regression-free and not fix this now

packages/aggregator/src/aggregationTypes.js
```diff
@@ -8,6 +8,7 @@ export const AGGREGATION_TYPES = {
   MOST_RECENT: 'MOST_RECENT',
   MOST_RECENT_PER_ORG_GROUP: 'MOST_RECENT_PER_ORG_GROUP',
   SUM: 'SUM',
+  SUM_UNTIL_THIS_DAY: 'SUM_UNTIL_THIS_DAY',
```
Alastair-L, line 11: Maybe `SUM_UNTIL_CURRENT_DAY`? "This" makes me think that we can specify which day, whereas it's always the current day.

packages/database/src/migrations/migrationData/20200826215113-UseIndicatorsInStriveVisualisations.js
Alastair-L: Makes sense.

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
Alastair-L: Good catch unit tests!

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
Alastair-L: I've seen you use it before in the indicator package I think. Very clean!

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.js
Alastair-L: Good catch unit tests!

packages/aggregator/src/aggregationTypes.js
Alastair-L: Also, I don't believe that the `UNTIL` represents what is going on at the moment. If there is data AFTER today's date, that data will still be included in the sum, so perhaps a rename here as well?

Alastair-L (commented):
I was initially very confused when I saw `SUM_UNTIL_THIS_DAY`, as the operation this performs is identical to `SUM` except for changing the period at the end. Perhaps, in the spirit of modularity, we could instead add a `CHANGE_PERIODS_TO_CURRENT_DAY` and combine that with `SUM`. That same aggregation type could be used with `SUM_MOST_RECENT_PER_FACILITY` and many other aggregations. 

Additionally, I think it would be easier to read the config
```javascript
    config: {
      formula: 'MRDT_POS_CONSULT_WEEKLY / SSWT1072',
      aggregation: {
        MRDT_POS_CONSULT_WEEKLY: ['SUM', 'CHANGE_PERIODS_TO_CURRENT_DAY'],
        SSWT1072: ['FINAL_EACH_WEEK', 'SUM', 'CHANGE_PERIODS_TO_CURRENT_DAY'],
}
```

The disadvantage of course is an extra key, and redoing the work that you have already done.

Overall sorry to be nitpicky but if you disagree with the above suggestion I would like to see a rename of `SUM_UNTIL_THIS_DAY`. What do you think?

kael89:
That was a good review @Alastair-L ! My thoughts:

1. I definitely agree that `SUM_UNTIL_CURRENT_DAY` is better than `SUM_UNTIL_THIS_DAY`! Happy to change that
2. I also thought that we could use something like `CHANGE_PERIODS_TO_CURRENT_DAY`. But I felt it would be a bit hacky: aggregation should reflect a calculation which makes sense, and not an arbitrary data manipulation. What we want to do in this case is _indeed_ sum until the current date.
3. My point 2 is in contradiction to something you very cleverly spotted: this isn't totally accurate since we allow future dates to sneak in! But on the other hand, I thought that this should be dealt with on a more general level, in other words we should generally exclude "future" data from our visualizations by default. So that shouldn't be this specific aggregator's concern

Based on all the above, I suggest
a. Rename the aggregator to `SUM_UNTIL_CURRENT_DAY`
b. Create card for addressing the "future data" issue in a generic way

What do you think?

kael89:
> That same aggregation type could be used with `SUM_MOST_RECENT_PER_FACILITY` and many other aggregations.

This is also a good point. It comes down to the generic "add specific aggregation types" or "use aggregation config" dilemma. In other words, in our data visualization/indicator config, do we prefer using:

**Option 1**
```js
const aggregations = [{ type: 'SUM', groupBy: 'facility', useCurrent: true, periodType: 'DAY' }];
```

or 

**Option 2**
```js
// too long name! not very extensible
const aggregation = ['SUM_MOST_RECENT_PER_FACILITY_UNTIL_CURRENT_DAY'];
```

or

**Option 3**
```js
// perhaps a good compromise?
const aggregation = ['SUM_MOST_RECENT_PER_FACILITY', 'SUM_UNTIL_CURRENT_DAY'];
```

I advocate for Option 3, because it extends on the current practice. It is juust a bit suboptimal on a performance level, but I guess we're talking about microseconds. Option 1 is very extensible, but inconsistent with how we generally define aggregations at the moment: perhaps we should revisit this when data builders are refactored. I certainly think that Option 2 is not sustainable.

The good thing is that under the hood, all approaches use the same code. As we have discussed in the past, this is just a matter of how we expose the functionality to clients.

Alastair-L (approved):
I would be happy enough with that! Preapproving pending the name change.

### a)
I agree with this statement wholeheartedly
> Aggregation should reflect a calculation which makes sense, and not an arbitrary data manipulation.

I am approving as it is fine as is, but I have one final suggestion which makes much more intuitive sense but will be possibly harder to implement. There are certain operations (like summing) that destroy the period associated with the data (because the data is now an aggregate of many periods). It is arbitrary to prefer one period over another after such an operation. Although this would introduce a lot of regression, what do you think about returning a `{ period: null }` from such operations instead of an arbitrary period from the input.

The indicator package could accept that null period as an 'any' period.


### b) 
I don't know about addressing future data generically. It could be perfectly valid for future data to be part of our visualizations (e.g. stock: how much stock **will be** at a particular facility on a day based on usage and deliveries planned). However, I do understand that it isn't currently a requirement so doesn't need to be supported by this card.

Alastair-L:
I didn't see your comment before I posted mine, I totally agree, and didn't think about the fact that it could be used like that! Much happier now.

kael89:
**a)** See https://github.com/beyondessential/tupaia-backlog/issues/255#issuecomment-692519178 -> Suggested Solution 1 (which was rejected) - I believe that it's perfectly valid to use the **max + more granular** period across aggregated data as the final period. It reflects reality, if you think about it (at least when summing past data!)

**b)** Good point, but I still think that using future data would already be wrong for the vast majority of our existing data visualizations. This is why I feel we shouldn't make this `sum`'s problem, but somehow deal with it in a generic aggregation level. I would expect future data to be excluded by default: if you want them in, you have to use a config explicitly/a specific aggregation type.

Alastair-L:
Sounds good 🎉.

edmofro:
Just to chime in on a couple of things:
- I really like the name change, nice suggestion @Alastair-L!
- I agree with @Alastair-L that the aggregation should ensure that no data beyond the current date is summed.
  - _Use case for future data_ We actually have a current use case for future data: rainfall, for which there is both historical observation data and future forecast data. If I did a `SUM_UNTIL_CURRENT_DAY` on rainfall, I would expect just the total rainfall up until now, and not the forecast data as well. I actually agree that this should still be the responsibility of something higher up to request analytics for the correct dates, but still, it is worth thinking about
  - _Expectations of the unit_ More compelling for me is looking at it without any concrete use case at all: IMO the aggregator should operate as a predictable utility, and I would predict `SUM_UNTIL_CURRENT_DAY` to cut off at the current date.
- Where I disagree with @Alastair-L  and am glad with the conclusion you reached instead is that the period is "destroyed" through summing, and that choosing any period is arbitrary. I think the period reached through summing is the final period in the range used during data fetch (or the max period in the data, if we don't know which date range the data was fetched for). I am glad we aren't using `CHANGE_PERIODS_TO_CURRENT_DAY` ;-)
- I agree with the current compromise between bundling all config into a long aggregation type, and having overly complex configs - though as @kael89 says, this may be reviewed during the data builder refactor

kael89:
Card created: [1290: Handle future data in aggregation](https://github.com/beyondessential/tupaia-backlog/issues/1290)

Let me re-iterate that: I definitely agree that `SUM_UNTIL_CURRENT_DAY` should exclude future data - I just believe this is a generic issue we may have, that this aggregation type makes more obvious because of its name. As such, I believe we should possibly tackle it at a general level, which exceeds the scope of this PR.

Let's consider that the current `SUM` aggregation has been returning inconsistent/wrong periods in the results since possibly ever (see my comments in the unit tests)

edmofro:
Good point on the inconsistent period in `SUM`! I believe that's a separate issue to whether future periods should be included though. All aggregation types except the two `MOST_RECENT` ones seem fine to return future data if it exists IMO, and those two could be renamed to `LATEST` and then they'd also be fine to do that. From my perspective, the future data issue is contained to this one new aggregation type

kael89:
It is a separate issue, I just wanted to point out that perhaps this concern is not worth the time investment right now 😉

PR #1335, opened 2020-09-16
932 refactor to fix lint errors

tcaiger:
### Issue #: towards [932](https://github.com/beyondessential/tupaia-backlog/issues/932)

I know that there is a lint package update coming soon thanks to Kostas but I wanted to get the lint errors fixed now and the rules are the same in the new package so it has to be done anyway.

### Changes:

- fix lint errors
- refactor layout component to make a bit drier

### DIFF
--- a/packages/psss/.eslintrc
+++ b/packages/psss/.eslintrc
@@ -11,6 +11,7 @@
     "jest/globals": true
   },
   "rules": {
+    "react/jsx-props-no-spreading": "off",
     "object-shorthand": ["error", "always"],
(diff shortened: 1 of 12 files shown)


packages/psss/.eslintrc
```diff
@@ -11,6 +11,7 @@
     "jest/globals": true
   },
   "rules": {
+    "react/jsx-props-no-spreading": "off",
```
kael89, line 14: This will be part of our new style guide! So good call adding it 🙌

kael89 (approved):
I love your attention to linting!

PR #1336, opened 2020-09-16
497: Migrate to node v12.18.3

kael89:
### Issue #:
Implements https://github.com/beyondessential/tupaia-backlog/issues/497 🎉 

But... There is a small problem... A local dev build of `meditrak-app` **does not** work in Windows 10 using the native OS API. It **does** work in MacOS. This is a [documented compatibility issue](https://github.com/expo/expo-cli/issues/1074) that `metro-config` has with Node `^12.10`. Here are some possible solutions, in my order of preference (✔ means I have confirmed it works)

1. Use WLS in Windows 10 (this is just my theory, haven't read this anywhere)
2. If `NODE.ENV === 'DEV'`,  use a (`postinstall`?) script in `meditrak-app` to apply the following solution: https://stackoverflow.com/questions/58120990/how-to-resolve-the-error-on-react-native-start/58199866#58199866 (✔ - have done it manually though)
3. Use Node version `12.9` instead of `12.18.3` 😞 ✔
4. Do not update node to version 12 at this stage. Thanks to babel + corejs-3, we can still use the modern ES syntax ✔
5. Update metro-config to `^0.56.4` => must update `react-native` to `^0.60`

I tried **5** for a bit and it seems a big pain, since react-native 0.60 introduces a lot of breaking changes. We should probably do this at some stage though!


packages/meditrak-app/app/database/types/User.js
```diff
@@ -7,9 +7,13 @@ import { Object as RealmObject } from 'realm';
...
-  constructor(...args) {
-    super(...args);
-    this.accessPolicy = new AccessPolicy(this.accessPolicyData);
+  accessPolicySingleton = null;
+
+  get accessPolicy() {
+    if (!this.accessPolicySingleton) {
+      this.accessPolicySingleton = new AccessPolicy(this.accessPolicyData);
+    }
+    return this.accessPolicySingleton;
```
kael89, line 16: Most of the changes in this PR are already reviewed in https://github.com/beyondessential/tupaia/pull/1274. It just occurred to me that I could have placed reviewed and non-reviewed changes in different commits 🤔

To assist the reviewer, I will point out which parts are new - so this is **1/3**

packages/meditrak-app/package.json
```diff
@@ -71,7 +71,9 @@
     "bump-patch": "npm-bump patch; ios/version-sync.sh",
     "bump-minor": "npm-bump minor; ios/version-sync.sh",
     "bugsnag-sourcemaps-upload": "scripts/uploadSourcemaps.sh",
-    "start-dev": "react-native run-android"
+    "start-dev": "react-native run-android",
+    "log:android": "react-native log-android",
+    "log:ios": "react-native log-ios"
```
kael89, line 76: New change 2/3 (just added those scripts)

packages/meditrak-app/package.json
```diff
@@ -110,7 +112,7 @@
     "react-navigation": "^1.5.8",
     "react-navigation-redux-helpers": "^1.0.3",
     "react-redux": "^5.0.7",
-    "realm": "^2.29.1",
+    "realm": "^6.1.1",
```
kael89, line 115: New change 3/3 - used `6.1.1` instead of the reviewed `6.1.0`

edmofro:
My favourite solution is your least favourite - updating react native seems the "right" way to go, even if it is a lot of work

edmofro (approved):
Three new changes all approved

edmofro:
So to move forward:
- Add a postinstall script to apply that fix
- Add a new card for updating react-native, and removing the postinstall script

kael89:
> My favourite solution is your least favourite - updating react native seems the "right" way to go, even if it is a lot of work

I agree with you! I should rephrase this to: "my favourite solution for now"

kael89:
Created [1289: Update react-native to the latest version](https://github.com/beyondessential/tupaia-backlog/issues/1289)

kael89:
@edmofro I have created an APK locally using Node 12, and @BinduMadhaviTammireddy did a quick smoke test in the app - all looks good. I am not familiar with the actual app deployment process though - is there something else we'd need to change/test regarding this update, to make sure that the production build of the app will work?

Another pending task before merging this to `dev` is figuring out a solution for building a local `meditrak-app` app in Windows 10 - @Alastair-L is currently investigating if it works in WSL

edmofro:
@kael89 another pending task is to ensure the iOS build also still works. There are three phases, though I'm happy if we only do the first two as the third is quite tricky (and from experience, if the first two pass, the third will also):
- Run the app in iOS simulator in dev mode, smoke test on your computer
- "Archive" (build) the app for deployment on the app store (though don't actually deploy) - this process is generally tricky, and where I expect us to meet issues!
- Test a production version on a phone through TestFlight (a fair bit of admin to do this - not totally necessary)

edmofro:
In terms of the apk deployment process though - you've done all that's required 🎉

Alastair-L:
Can confirm that I am able to build a debug version and see the output logs: 
image

Alastair-L:
Also able to build release:
image

https://drive.google.com/file/d/1rviCcLDsugCxhIOuRpv484X9Livxe-rk/view?usp=sharing

kael89:
Thanks amazing, thanks heaps @Alastair-L !! 🎉

PR #1339, opened 2020-09-16
1052: School Indicators EiE Map Overlay: Update to functioning hand washing facilities

chris-bes:
### Changes:

- Added migration to update Laos Schools Functioning Hand Washing facilities overlay


kael89 (approved):
Looks good

PR #1340, opened 2020-09-16
1053: School Indicators EiE Map Overlay: Update to water supply sources

chris-bes:
### Changes:

- Added migration to update the responses of access to water supply overlay


kael89 (approved):
Looks good!

PR #1341, opened 2020-09-16
932 refactor components and containers

tcaiger:
### Changes:

There is a rough existing convention in the PSSS package that any react components that are connected to redux are considered _containers_ and placed in the `containers` directory while all other react components are placed in the `components` directory. These changes are solidifying that convention and putting all react components in the appropriate directory.


edmofro (approved):
Always improving, good call

PR #1344, opened 2020-09-17
932 psss design qa

tcaiger:
### Issue #: [732](https://github.com/beyondessential/tupaia-backlog/issues/932)

This PR is a bit of a catch all of design feedback on the PSSS front end build and so it is a mix of style tweaks and a few bug fixes. Unfortunately the changes are a bit piecemeal which might make it a bit hard to review. The issue card has a list of most of the issues and might be helpful for understanding the changes.

### Changes:

- Tweak table styles
- Tweak button styles
- Make 'Review and Submit' weekly reports panel button work
- Make table row clickable

---

### Screenshots:

Screen Shot 2020-09-18 at 10 06 35 AM


packages/ui-components/src/components/Table/TablePaginator.js
```diff
@@ -61,6 +71,10 @@ export const TablePaginator = React.memo(
       [onChangeRowsPerPage],
     );
 
+    if (count <= rowsPerPage) {
+      return null;
+    }
```
rohan-bes, line 76: I think I'm missing something here. Does moving this check after creating the hooks achieve something?

packages/ui-components/src/components/Button.js
```diff
@@ -56,7 +60,7 @@ Button.defaultProps = {
 };
 
 export const LightPrimaryButton = styled(Button)`
-  background-color: ${props => props.theme.palette.primary.light};
+  background-color: #dceffb;
```
rohan-bes, line 63: Did we wanna update the value in the `theme` to this? Or perhaps have a new key for it?

rohan-bes (approved):
Hehe yeah like you said a little tricky to review this one with so much going on. Comments are fairly minor things, so I'm approving anyway and feel free to address them if you wish 😄

packages/ui-components/src/components/Table/TablePaginator.js
tcaiger: yeah it just ensures that the hooks are not called conditionally which is one of the rules of hooks. It can cause weird bugs with how they store state internally if you do. It was just bought to my attention by the updated linting rules that Kosas added

packages/ui-components/src/components/Button.js
tcaiger: Unfortunately I wasn't able to use a theme variable because the colour for the button is not in the palette of the design. Designers just can't stick to the rules ay 😜  I'll check with Dave to see if he wants to change the theme but for now it can probably just stay as a one off

PR #1012, opened 2020-07-21
Indicators

edmofro:
### Issues

[254: Design and create 'indicator' table in db](https://github.com/beyondessential/tupaia-backlog/issues/254)
[255: Design config for 'arithmetic' indicator builder](https://github.com/beyondessential/tupaia-backlog/issues/255)
[256: Design indicator package interface, build scaffolding](https://github.com/beyondessential/tupaia-backlog/issues/256)
[257: Implement 'arithmetic' indicator builder](https://github.com/beyondessential/tupaia-backlog/issues/257)
[259: Add indicator package as a service for data-broker](https://github.com/beyondessential/tupaia-backlog/issues/259)
[261: Use indicator for febrile cases and mrdt positive cases](https://github.com/beyondessential/tupaia-backlog/issues/261)
[1242: Support default values in arithmetic indicator calculation](https://github.com/beyondessential/tupaia-backlog/issues/1242)
[1290: Handle future data in aggregation](https://github.com/beyondessential/tupaia-backlog/issues/1290)

PR #1345, opened 2020-09-17
1290: Exclude future dates

kael89:
### Issue #:
Addresses https://github.com/beyondessential/tupaia-backlog/issues/1290 - at least in part

### DIFF
--- a/packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
+++ b/packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
@@ -19,16 +19,17 @@ describe('sumAcrossPeriods', () => {
 
   const ANALYTICS = arrayToAnalytics([
     // BCD01 - TO
-    ['BCD01', 'TO', '201601', 1],
+    ['BCD01', 'TO', '201601', 0.01],
     ['BCD01', 'TO', '201801', 0.1],
-    ['BCD01', 'TO', '201901', 0.01],
+    ['BCD01', 'TO', '201901', 1],
     // BCD01 - PG
-    ['BCD01', 'PG', '201601', 2],
+    ['BCD01', 'PG', '201601', 0.02],
     ['BCD01', 'PG', '201801', 0.2],
-    ['BCD01', 'PG', '201901', 0.02],
+    ['BCD01', 'PG', '201901', 2],
     // BCD02 - TO
-    ['BCD02', 'TO', '201801', 4],
-    ['BCD02', 'TO', '202001', 0.44],
+    ['BCD02', 'TO', '201801', 0.04],
+    ['BCD02', 'TO', '202001', 0.4],
+    ['BCD02', 'TO', '210001', 4.0],
   ]);
 
   const CURRENT_DATE_STUB = '2020-01-31T00:00:00Z';
@@ -87,6 +88,12 @@ describe('sumAcrossPeriods', () => {
     });
 
     describe('useCurrent', () => {
+      it('defaults to `false`', () => {
+        expect(sumAcrossPeriods(ANALYTICS, { periodOptions: {} })).to.deep.equal(
+          sumAcrossPeriods(ANALYTICS, { periodOptions: { useCurrent: false } }),
+        );
+      });
+
       it('does nothing if not set', () => {
         expect(
           sumAcrossPeriods(ANALYTICS, { periodOptions: { useCurrent: false } }),
@@ -119,6 +126,42 @@ describe('sumAcrossPeriods', () => {
             ['BCD01', 'TO', CURRENT_PERIOD_STUBS.DAY, 1.11],
             ['BCD01', 'PG', CURRENT_PERIOD_STUBS.DAY, 2.22],
             ['BCD02', 'TO', CURRENT_PERIOD_STUBS.DAY, 4.44],
+          ]),
+        );
+      });
+    });
+
+    describe('excludeFuture', () => {
+      it('defaults to `false`', () => {
+        expect(sumAcrossPeriods(ANALYTICS, { periodOptions: { periodOptions: {} } })).to.deep.equal(
+          sumAcrossPeriods(ANALYTICS, {
+            periodOptions: { periodOptions: { excludeFuture: false } },
+          }),
+        );
+      });
+
+      it('does nothing if not set', () => {
+        expect(
+          sumAcrossPeriods(ANALYTICS, {
+            periodOptions: { periodOptions: { excludeFuture: false } },
+          }),
+        ).to.have.same.deep.members(
+          arrayToAnalytics([
+            ['BCD01', 'TO', '201601', 1.11],
+            ['BCD01', 'PG', '201601', 2.22],
+            ['BCD02', 'TO', '201801', 4.44],
+          ]),
+        );
+      });
+
+      it('filters future periods', () => {
+        expect(
+          sumAcrossPeriods(ANALYTICS, { periodOptions: { excludeFuture: true } }),
+        ).to.have.same.deep.members(
+          arrayToAnalytics([
+            ['BCD01', 'TO', '201601', 1.11],
+            ['BCD01', 'PG', '201601', 2.22],
+            ['BCD02', 'TO', '201801', 0.44],
           ]),
         );
       });
--- a/packages/utils/src/period/period.js
+++ b/packages/utils/src/period/period.js
@@ -125,12 +125,18 @@ const checkNonNumericPeriod = (period, potentialType) => {
   return isValidFormat;
 };
 
-/**
- *
- * @param {any} period
- * @returns {boolean}
- */
 export const isValidPeriod = period => typeof period === 'string' && !!periodToType(period);
+
+/**
+ * @returns {number} > 0 if periodA > periodB, 0 if periodA = periodB, < 0 if periodA < periodB
+ */
+export const comparePeriods = (periodA, periodB) => {
+  const dayPeriodA = convertToPeriod(periodA, DAY);
+  const dayPeriodB = convertToPeriod(periodB, DAY);
+  return parseInt(dayPeriodA, 10) - parseInt(dayPeriodB, 10);
+};
+
+export const isFuturePeriod = period => comparePeriods(period, getCurrentPeriod(DAY)) > 0;
 
 export const parsePeriodType = periodTypeString => {
   const error = new Error(`Period type must be one of ${Object.values(PERIOD_TYPES)}`);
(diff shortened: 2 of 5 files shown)


packages/utils/src/period/period.js
```diff
@@ -125,12 +125,18 @@ const checkNonNumericPeriod = (period, potentialType) => {
...
+export const isValidPeriod = period => typeof period === 'string' && !!periodToType(period);
+
 /**
- *
- * @param {any} period
- * @returns {boolean}
+ * @returns {number} > 0 if periodA > periodB, 0 if periodA = periodB, < 0 if periodA < periodB
  */
-export const isValidPeriod = period => typeof period === 'string' && !!periodToType(period);
+export const comparePeriods = (periodA, periodB) => {
```
kael89, line 133: I believe there would be lots of places where we could use that, this justifies the attempt for rigourous test cases. We also don't want to risk nasty bugs because of wrong comparisons (eg being different period types!)

But I won't refactor out of scope code in this PR

packages/utils/src/tests/period/period.test.js
```diff
@@ -7,19 +7,21 @@ import { expect } from 'chai';
...
+  isValidPeriod,
   momentToPeriod,
-  periodToDisplayString,
   parsePeriodType,
+  PERIOD_TYPES,
+  periodToDisplayString,
   periodToTimestamp,
-  isValidPeriod,
-  isCoarserPeriod,
+  periodToType,
```
kael89, line 24: Just sorted the existing imports + added `comparePeriods`, `isFuturePeriod`

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
```diff
@@ -19,16 +19,17 @@ describe('sumAcrossPeriods', () => {
...
+    ['BCD01', 'PG', '201601', 0.02],
     ['BCD01', 'PG', '201801', 0.2],
-    ['BCD01', 'PG', '201901', 0.02],
+    ['BCD01', 'PG', '201901', 2],
     // BCD02 - TO
-    ['BCD02', 'TO', '201801', 4],
-    ['BCD02', 'TO', '202001', 0.44],
+    ['BCD02', 'TO', '201801', 0.04],
+    ['BCD02', 'TO', '202001', 0.4],
+    ['BCD02', 'TO', '210001', 4.0],
```
Alastair-L, line 32: Clever values for this test! I'll use this in future!

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
```diff
@@ -123,5 +130,41 @@ describe('sumAcrossPeriods', () => {
         );
       });
     });
+
+    describe('excludeFuture', () => {
+      it('defaults to `false`', () => {
+        expect(sumAcrossPeriods(ANALYTICS, { periodOptions: { periodOptions: {} } })).to.deep.equal(
```
Alastair-L, line 136:
```suggestion
        expect(sumAcrossPeriods(ANALYTICS, { periodOptions: {} })).to.deep.equal(
```
And similarly below.

packages/utils/src/tests/period/period.test.js
```diff
@@ -108,6 +108,177 @@ context('periodTypes', () => {
...
+    it('present', () => {
+      expect(isFuturePeriod('2020')).to.be.false;
+      expect(isFuturePeriod('2020Q4')).to.be.false;
+      expect(isFuturePeriod('202012')).to.be.false;
+      expect(isFuturePeriod('20201231')).to.be.false;
+    });
+
+    it('present - week period type', () => {
+      // We need to match the last date of a week
+      clock = sinon.useFakeTimers({ now: new Date('2020-12-27T00:00:00Z') });
```
edmofro, line 268: See https://github.com/beyondessential/tupaia/pull/1291/files#diff-341d943defdc3255425fddec1741973b for Igor's util - small difference in format, perhaps you'd like to comment over there if it is better to specify `now` within an object

edmofro (approved):
Very nicely done!

packages/utils/src/period/period.js
Alastair-L: Makes sense!

packages/utils/src/period/period.js
```diff
@@ -125,12 +125,18 @@ const checkNonNumericPeriod = (period, potentialType) => {
   return isValidFormat;
 };
 
+export const isValidPeriod = period => typeof period === 'string' && !!periodToType(period);
```
Alastair-L, line 128: Clever!

packages/utils/src/period/period.js
```diff
@@ -125,12 +125,18 @@ const checkNonNumericPeriod = (period, potentialType) => {
...
+ * @returns {number} > 0 if periodA > periodB, 0 if periodA = periodB, < 0 if periodA < periodB
  */
-export const isValidPeriod = period => typeof period === 'string' && !!periodToType(period);
+export const comparePeriods = (periodA, periodB) => {
+  const dayPeriodA = convertToPeriod(periodA, DAY);
+  const dayPeriodB = convertToPeriod(periodB, DAY);
+  return parseInt(dayPeriodA, 10) - parseInt(dayPeriodB, 10);
+};
+
+export const isFuturePeriod = period => comparePeriods(period, getCurrentPeriod(DAY)) > 0;
```
Alastair-L, line 139: Nice!

packages/utils/src/tests/period/period.test.js
kael89: Nice, I didn't know the other syntax! I guess they are equivalent - the object param just allows you to provide more options

packages/utils/src/tests/period/period.test.js
```diff
@@ -108,6 +108,177 @@ context('periodTypes', () => {
     });
   });
 
+  describe('comparePeriods', () => {
```
Alastair-L, line 111: I know this is a pain, but why not add a test case which requires that it be a certain number rather than just >0? In future, code may rely on this functionality.

packages/utils/src/tests/period/period.test.js
kael89: I'll use the shorter syntax here too

Alastair-L (approved):
Very nice implementation! Preapproving, but there are a couple of comments about the tests.

Alastair-L:
Sorry @kael89, didn't notice Edwin's review! Feel free to ignore mine :)

packages/utils/src/tests/period/period.test.js
kael89: If code relies on that, it is an implementation detail - the specs of the function are the ones defined in the JsDoc, and the ones we need to assert in the test. The specs were based on `compareFunction` which can be passed in [Array.prototype.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort). 

Overly specific assertions make for fragile tests

kael89: What I didn't assert is this

> compareFunction(a, b) must always return the same value when given a specific pair of elements a and b as its two arguments. If inconsistent results are returned, then the sort order is undefined.

i.e don't care if `compareFunction(a, b)` returns 3 or 5, as long as it is greater than zero - but it must return the same value every time. Happy to add a test case for those!

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
kael89: Great catch!! 👀

packages/utils/src/tests/period/period.test.js
kael89: Actually, I changed my mind - of course the same output should be returned with the same input, the implementation is clearly deterministic so need to add a test case for that

kael89:
> Sorry @kael89, didn't notice Edwin's review! Feel free to ignore mine :)

Yeah it was first come, first served 😆

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
kael89: Fixed in https://github.com/beyondessential/tupaia/pull/1012/commits/7d3fbebb19a1fc5677f439cf6a24d0a89055eb13 - I couldn't find another instance of that issue, but please let me know if I'm wrong

packages/utils/src/tests/period/period.test.js
Alastair-L: That makes sense! In my head `comparePeriods` was more like 'return the number of days between these two periods', but he docstring makes the true specs clear.

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
```diff
@@ -123,5 +130,41 @@ describe('sumAcrossPeriods', () => {
...
+          sumAcrossPeriods(ANALYTICS, {
+            periodOptions: { periodOptions: { excludeFuture: false } },
```
Alastair-L, line 138: Just here

packages/aggregator/src/tests/analytics/aggregateAnalytics/aggregations/sumAcrossPeriods.test.js
```diff
@@ -123,5 +130,41 @@ describe('sumAcrossPeriods', () => {
...
+          }),
+        );
+      });
+
+      it('does nothing if not set', () => {
+        expect(
+          sumAcrossPeriods(ANALYTICS, {
+            periodOptions: { periodOptions: { excludeFuture: false } },
```
Alastair-L, line 146: and here

kael89: Oh cool, thanks!

packages/utils/src/tests/period/period.test.js
kael89: What you say is correct in this specific implementation, but it just happened that I implemented it this way - a different implementation with other integer values could work as well

What we want to achieve here is a classic C-inherited way of allowing a three-state return value instead of a binary `true`/`false` - the fact that one can convert entities to numbers and then substract them has historically made developers to define those states as `> 0`, `= 0`, `< 0`

Alastair-L: That makes sense. I'm sure you're used to it, but it's kind of strange when you think about it the first time.

PR #1346, opened 2020-09-18
1006 admin panel optional remember me

tcaiger:
### Changes:

- add remember me checkbox to login form
- add remember me field to redux
- conditionally persist user data based on remember me value

The same feature needs to be added to PSSS so once these changes are approved, a similar PR will be made for that.

---

### Screenshots:
Screen Shot 2020-09-18 at 2 18 09 PM


rohan-bes (approved):
Looks really clean 😄 Nice one!

PR #1353, opened 2020-09-18
794 fix optionset import

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/794

I'm sure theres a way to do this same functionality using `hasContent` (which I removed from this file) but I couldn't figure out how. This code works but I'm happy to do it a cleaner way if there's an obvious solution

### DIFF
--- a/packages/meditrak-server/src/database/models/Option.js
+++ b/packages/meditrak-server/src/database/models/Option.js
@@ -4,20 +4,19 @@
  **/
 
 import { DatabaseModel, DatabaseType, TYPES } from '@tupaia/database';
+import { hasContent } from '@tupaia/utils';
 
 class OptionType extends DatabaseType {
   static databaseType = TYPES.OPTION;
 
   static fieldValidators = new Map()
     .set('value', [
-       async (value) => {
-        const checkIsEmpty = value => value === undefined || value === null || value.length === 0;
-          if (checkIsEmpty(value)) {
-            return 'Value cannot be empty'; 
-          }
-          else {           
-           return null;
-          };
+      value => {
+        try {
+          return hasContent(value) && null;
+        } catch (error) {
+          return error.message;
+        }
       },
       async (value, model) => {
         const foundConflict = await findFieldConflict('value', value, model);
--- a/packages/meditrak-server/src/database/models/Option.js
+++ b/packages/meditrak-server/src/database/models/Option.js
@@ -4,14 +4,23 @@
  **/
 
 import { DatabaseModel, DatabaseType, TYPES } from '@tupaia/database';
-import { hasContent } from '@tupaia/utils';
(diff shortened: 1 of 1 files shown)


packages/meditrak-server/src/database/models/Option.js
```diff
@@ -4,14 +4,21 @@
...
-      hasContent,
+       async (value) => {
+        const checkIsEmpty = value => value === undefined || value === null || value.length === 0;
+          if (checkIsEmpty(value)) {
+            return 'Value cannot be empty'; 
+          }
+          else {           
+           return null;
+          };
+      },
```
tcaiger, line 21:
```suggestion
       value => {
        try {
          return hasContent(value) && null;
        } catch (error) {
          return error.message;
        }
      },
```

tcaiger: We could refactor the fix to use a try/catch block which would have the advantage of re-using the `hasContent` method from utils but on the other hand breaks the pattern of the other validators.

tcaiger (approved):
Nice work nailing down the bug! I was confused about why the original bug was happening and so I had a look at some other places the hasContent helper is being used and realised that elsewhere it is used inside a try/catch block and that is why it doesn't work properly here.

I have suggested adding an alternative solution but I would also be happy to stick with your proposed solution as I think they both work and are both equally good fixes in this case!

I'm happy for you to decide and so am marking as approved.

packages/meditrak-server/src/database/models/Option.js
```diff
@@ -4,14 +4,21 @@
...
-import { hasContent } from '@tupaia/utils';
 
 class OptionType extends DatabaseType {
   static databaseType = TYPES.OPTION;
 
   static fieldValidators = new Map()
     .set('value', [
-      hasContent,
+       async (value) => {
+        const checkIsEmpty = value => value === undefined || value === null || value.length === 0;
```
tcaiger, line 14: Also you should be seeing a couple of lint errors in this function if your linter is working correctly so it would be good to look into that when you get a chance. 

eg.
Screen Shot 2020-09-21 at 11 55 43 AM

tcaiger (approved):

avaek:
Hey @tcaiger - thanks for that review, your suggested change is perfect & I've implemented that now (if you just want to quickly double check I implemented it correctly)

tcaiger:
> Hey @tcaiger - thanks for that review, your suggested change is perfect & I've implemented that now (if you just want to quickly double check I implemented it correctly)

Yes looks good!

edmofro:
updated base branch to today's hot fix, we can merge in there, test those two fixes, then release

@avaek can you coordinate getting both tested?

PR #1356, opened 2020-09-20
1118 manage profile on admin panel & psss

tcaiger:
### Changes:

- Add profile pages and routes
- Add profile form
- Add react hook form and form validation to form
- Add profile form handling
- tweak server to return more user fields

---

### Screenshots:
Screen Shot 2020-09-21 at 4 10 56 PM


packages/admin-panel/src/App.js
```diff
@@ -8,7 +8,9 @@ import { BrowserRouter as Router, Route, Switch, Redirect } from 'react-router-d
 import { TabsToolbar } from '@tupaia/ui-components';
 import { Navbar, Footer } from './widgets';
 import { ROUTES } from './routes';
-import { PrivateRoute, LoginView } from './authentication';
+import { PROFILE_ROUTES } from './profileRoutes';
+import { PrivateRoute } from './authentication';
+import { LoginPage } from './pages/LoginPage';
```
edmofro, line 13:
```suggestion
import { PrivateRoute } from './authentication';
import { PROFILE_ROUTES } from './profileRoutes';
import { LoginPage } from './pages/LoginPage';
```

packages/admin-panel/src/authentication/actions.js
```diff
@@ -63,3 +64,17 @@ export const loginError = errorMessage => ({
 export const logout = () => ({
   type: LOGOUT,
 });
+
+// Profile
+export const updateProfile = (id, payload) => async (dispatch, getState, { api }) => {
+  try {
+    await api.put(`user/${id}`, null, payload);
```
edmofro, line 71: We should be able to use a `/me` endpoint here, which can then have built in permission checking that the same user is making the request as is being edited

packages/admin-panel/src/pages/ChangePasswordPage.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Tupaia MediTrak
```
edmofro, line 2: This file's not really part of this PR, but will hand wave it coz I think that's the next PR in the reviews channel

packages/meditrak-server/src/routes/authenticate.js
```diff
@@ -57,6 +57,10 @@ const getAuthorizationObject = async ({ accessPolicy, refreshToken, user, permis
   const userDetails = {
     id: user.id,
     name: user.fullName,
+    firstName: user.first_name,
+    lastName: user.last_name,
+    position: user.position,
+    employer: user.employer,
```
edmofro, line 63: This adds the extra fields to every client doing authentication. I think I'd prefer to add a `GET /me` endpoint that just returns more info about the user attached to the auth header

packages/admin-panel/src/pages/ProfilePage.js
```diff
@@ -0,0 +1,151 @@
...
+const ProfilePageComponent = React.memo(({ user, onUpdateProfile, getHeaderEl }) => {
+  const [errorMessage, setErrorMessage] = useState(null);
+  const [successMessage, setSuccessMessage] = useState(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const { handleSubmit, register, errors } = useForm();
+  const HeaderPortal = usePortalWithCallback(<Header title={user.name} />, getHeaderEl);
+
+  const onSubmit = handleSubmit(async ({ firstName, lastName, role, employer }) => {
+    setIsLoading(true);
+    setErrorMessage(null);
```
edmofro, line 46:
```suggestion
    setErrorMessage(null);
    setSuccessMessage(null);
```

packages/admin-panel/src/pages/ProfilePage.js
```diff
@@ -0,0 +1,151 @@
...
+        first_name: firstName,
+        last_name: lastName,
+        position: role,
+        employer: employer,
+      });
+      setIsLoading(false);
+      setSuccessMessage('Profile successfully updated.');
+    } catch (error) {
+      setErrorMessage(error.message);
+      setIsLoading(false);
```
edmofro, line 58: This is super nitpicky, but just for consistency...

```suggestion
      setIsLoading(false);
      setSuccessMessage('Profile successfully updated.');
    } catch (error) {
      setIsLoading(false);
      setErrorMessage(error.message);
```

packages/admin-panel/src/pages/ProfilePage.js
```diff
@@ -0,0 +1,151 @@
...
+
+const SuccessMessage = styled.p`
+  color: ${props => props.theme.palette.success.main};
+`;
+
+const ProfilePageComponent = React.memo(({ user, onUpdateProfile, getHeaderEl }) => {
+  const [errorMessage, setErrorMessage] = useState(null);
+  const [successMessage, setSuccessMessage] = useState(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const { handleSubmit, register, errors } = useForm();
```
edmofro, line 41: With `noValidate` on the form, will `errors` ever have content?

edmofro: Oh yep, presumably the "required" fields will still have that validated

edmofro (changes requested):
Front end all looks awesome. Just got some api change suggestions

packages/admin-panel/src/pages/ProfilePage.js
tcaiger: Good catch

packages/admin-panel/src/pages/ProfilePage.js
tcaiger: yes that's correct, it's just doing required validation in this instance

packages/meditrak-server/src/routes/authenticate.js
tcaiger: hmmm I feel like there is no real downside to including a few extra fields on each response and it saves doing an extra request every time the user logs in. Plus if we keep using `name` in this endpoint and then have another endpoint with `firstName` and `lastName`, we would have to be careful not to let them get out of sync in an app. eg. user details have been changed in AdminPanel and PSSS user has not logged back out but has been re-authenticated in the background.

packages/meditrak-server/src/routes/editUser.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Tupaia MediTrak
+ * Copyright (c) 2017 Beyond Essential Systems Pty Ltd
+ */
```
edmofro, line 4: Even inside meditrak-server we now use the generic mono-repo header for new files

(I only notice because of your updates to the others)

packages/meditrak-server/src/routes/editUser.js
```diff
@@ -0,0 +1,25 @@
...
+  if (!user) {
+    throw new DatabaseError('User not found');
+  }
+
+  try {
+    req.params = {
+      id: userId,
+      resource: 'user',
+    };
+    await editRecord(req, res);
```
edmofro, line 21: This will run into conflicts once @EMcQ-BES has finished the permissions work

packages/meditrak-server/src/routes/authenticate.js
edmofro: Fair enough, I'm happy with that :-)

edmofro (approved):
Approved, just a copyright header to change over before merging

PR #1358, opened 2020-09-21
Add lint scripts

kael89:
### Issue #:
Towards https://github.com/beyondessential/tupaia-backlog/issues/967

### DIFF
--- a/packages/web-config-server/package.json
+++ b/packages/web-config-server/package.json
@@ -5,8 +5,6 @@
   "description": "Config server for tupaia.org",
   "main": "dist",
   "scripts": {
-    "lint": "eslint --ext .js  --ignore-pattern __tests__/ .",
-    "lint:all": "yarn run lint \"src/**/*.{js,jsx}\"",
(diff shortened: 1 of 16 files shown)


packages/web-config-server/package.json
```diff
@@ -5,15 +5,15 @@
   "description": "Config server for tupaia.org",
   "main": "dist",
   "scripts": {
-    "lint": "eslint --ext .js  --ignore-pattern __tests__/ .",
```
tcaiger, line 8: So we used to ignore tests and now we include them. Just double checking that this is intentional?

tcaiger (approved):
Looks good to me. Just left a question that is a double check but not something that needs to be addressed for approval.

packages/web-config-server/package.json
kael89: Yeap! Tests should be linted too (I think people excluded them in the past just for convenience 😆 )

We can still disable some rules in the tests folder (eg enforcing camelCase, since some times we use snake_case for test data), I'll do that if required in a next PR

PR #1370, opened 2020-09-22
1218 admin panel change password

tcaiger:
#662 # Issue #: [1218](https://github.com/beyondessential/tupaia-backlog/issues/1218)

### Changes:

- Add change password page and form
- Add password strength bar

---

### Screenshots:

Screen Shot 2020-09-22 at 4 44 10 PM
Screen Shot 2020-09-22 at 4 44 30 PM
Screen Shot 2020-09-22 at 4 45 05 PM
Screen Shot 2020-09-22 at 4 45 23 PM

### Lazy Loading Test

**When doing a production build Webpack notices the lazy loaded package and creates a seperate chunk for it**
Screen Shot 2020-09-25 at 10 30 00 AM

**With lazy loading**
The original chunk loads (0.95MB) on the first page visited and then the second chunk loads only when the change password page is visited.
after

**Without lazy loading**
Only one chunk loads with everything bundled together (1.3MB)
before


packages/meditrak-server/src/routes/changePassword.js
```diff
@@ -45,7 +45,7 @@ export async function changePassword(req, res, next) {
       password,
     };
 
-    editRecord(req, res);
+    await editRecord(req, res);
```
edmofro, line 48: Good catch!

packages/admin-panel/src/pages/ChangePasswordPage.js
```diff
@@ -1,23 +1,156 @@
...
+import { connect } from 'react-redux';
+import { useForm } from 'react-hook-form';
+import { Button, TextField, SmallAlert, PasswordStrengthBarFallback } from '@tupaia/ui-components';
 import { usePortalWithCallback } from '../utilities';
 import { Header } from '../widgets';
+import { updatePassword, getUser } from '../authentication';
+
+// Lazy load the component as it uses zxcvbn which is a large library.
+// For more about lazy loading components @see: https://reactjs.org/docs/code-splitting.html#reactlazy
+const PasswordStrengthBar = lazy(() => import('../widgets/PasswordStrengthBar'));
```
edmofro, line 19: Really cool! Nice to see that lazy loading is easy in react :-)

packages/admin-panel/src/pages/ChangePasswordPage.js
```diff
@@ -1,23 +1,156 @@
...
+          type="password"
+          error={!!errors.passwordConfirm}
+          helperText={errors.passwordConfirm && errors.passwordConfirm.message}
+          inputRef={register({
+            required: 'Required',
+            minLength: { value: 9, message: 'Password must be over 8 characters long.' },
+            validate: value => value === getValues('password') || 'Passwords do not match.',
+          })}
+        />
+        <Suspense fallback={<PasswordStrengthBarFallback />}>
```
edmofro, line 116: Cool!!

packages/ui-components/src/components/PasswordStrengthBar.js
```diff
@@ -0,0 +1,81 @@
...
+`;
+
+const HelperText = styled(Typography)`
+  font-size: 0.9rem;
+  line-height: 1rem;
+  color: ${props => props.theme.palette.text.tertiary};
+  margin-top: 1rem;
+`;
+
+export const PasswordStrengthBarFallback = styled.div`
```
edmofro, line 27: So this is where you could have something that looked like a fuzzy version of a strength bar, if you wanted to make it fancy?

packages/admin-panel/src/pages/ChangePasswordPage.js
edmofro: I'm interested to know more about the code splitting. I imagine:
- because we just build a single bundle for ui-components, the javascript that represents our customised PasswordStrengthBar component has already been downloaded at this point; but
- it hasn't loaded in the dependencies of that until that part of our ui-components bundle is actually run, so the zxcvbn gets lazily loaded

Is that right?

Would be nice to code split the ui-components bundle itself at some point!

edmofro (changes requested):
This has some really awesome stuff in it, nice work!

packages/admin-panel/src/pages/ChangePasswordPage.js
```diff
@@ -1,23 +1,156 @@
...
+      setIsLoading(false);
+      setSuccessMessage('Password successfully updated.');
+      event.target.reset();
+    } catch (error) {
+      setErrorMessage(error.message);
+      setIsLoading(false);
+    }
+  });
+
+  const password = watch('password');
```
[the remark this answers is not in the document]
tcaiger: yeah it is pretty handy for this exact use case but I would hope not to use it on many forms as it has a small performance impact. No more than making the field a 'controlled' field though, so I can live with that

packages/admin-panel/src/pages/ChangePasswordPage.js
```diff
@@ -1,23 +1,156 @@
...
+          name="passwordConfirm"
+          placeholder="Enter your password"
+          required
+          type="password"
+          error={!!errors.passwordConfirm}
+          helperText={errors.passwordConfirm && errors.passwordConfirm.message}
+          inputRef={register({
+            required: 'Required',
+            minLength: { value: 9, message: 'Password must be over 8 characters long.' },
+            validate: value => value === getValues('password') || 'Passwords do not match.',
```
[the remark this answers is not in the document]
tcaiger: good catch

packages/ui-components/src/components/PasswordStrengthBar.js
tcaiger: yep it's basically just a loader. Obviously it doesn't need to be here - it could be made directly in the suspense component but I wanted to make it easy for future implementations.

packages/admin-panel/src/pages/ChangePasswordPage.js
tcaiger: thanks for the question. It made me realise that the lazy loading wasn't working properly and triggered me to fix it. As we discussed the best solution for this use case was to bring the PasswordStrengthBar into the admin-panel package so that it could be properly lazy loaded. Yes it could be good to split the ui-components at some point. I have created a task in the backlog for that and added some notes from my investigation. @see https://github.com/beyondessential/tupaia-backlog/issues/1347

tcaiger:
@edmofro This is ready for another review. I have updated the PasswordStrengthBar component to be in AdminPanel. I did some better testing to make sure it is lazy loading properly and left screenshots and comments about that in the PR description above.

edmofro (approved):

PR #1371, opened 2020-09-22
1166 filter out blank values in admin panel data element config

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1166

_Note:_ I get a "no-unused-vars" eslint error on `key` in `.filter(([key, value])` but I couldn't figure out how to get that filtering to work without including it so I just submitted the PR without resolving it. If its an egregious error I'm happy to try fix it but I'm not exactly sure how to make it work without it


packages/admin-panel/src/pages/resources/DataSourcesPage.js
```diff
@@ -20,12 +20,15 @@ const localStyles = {
...
 const DataSourceConfigView = row => {
-  const entries = Object.entries(row.value).map(([key, value]) => (
-    <React.Fragment key={key}>
-      <dt style={localStyles.config.dt}>{key}:</dt>
-      <dd>{value.toString()}</dd>
-    </React.Fragment>
-  ));
+  const blankString = '';
+  const entries = Object.entries(row.value)
+    .filter(([key, value]) => value !== blankString)
```
tcaiger, line 25:
```suggestion
.filter(rowValue => rowValue.value !== blankString)
```

tcaiger (approved):
Nice work - the fix looks good to me! It's not really a big deal but I left a suggestion if you want to fix that lint error.

PR #1376, opened 2020-09-23
1271 utils replace stubs

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1271


kael89:
@billli0 A lot of changes here!! Ideally we'd like to break our PRs into much smaller chunks (10, 20 files max), but this is a special case as most of the changes are moving stuff around / re-organising things

I highly recommend going through the commits one-by-one since they are generally self-contained. I will also add notes to assist your review. Cheers!

kael89:
We can now also see the code coverage in this package - which is not too bad!

image

kael89:
@billli0 apart from fixing the skipped tests, I restructured a lot of parameterized tests. A few points for us to keep in mind for future conversions:

* Some times using parameterized tests makes the tests more complicated to read eg. when there are only a couple of test cases. We should make a judgement call for each case. The ultimate goal is, we want future developers to feel thankful for whoever wrote the tests, because the tests are very easy to read 😃
* I found it more convenient to define the test data in a `const testData = ` variable before `it.each`, so let's standardise this syntax. The advantage is that test data and code are separated
* Each test case should have a unique, meaningful description. We want to avoid test results like this:
image

and instead see something like the following:
image

This was very evident in the `period.test.js` parameterized tests
* If we cannot find a way to distinguish parameterized test cases, we can use indexes:
```js
'object input (i)',
'object input (ii)'
```

but only if we can't use anything more specific:
```js
// the descriptions below provide useful information about the test input
'object input - single key',
'object input - multiple keys'
```
* Generally we want to stick to the following structure for parameterized tests. There are exceptions, again this a decision to be made for each case:
```js
const testData = [
  ['description', inputParams, expected]
  // ...
]

// Example, for function accepting multiple params
const testData = [
    [
      'should convert a date range in 2 consecutive months',
      ['20180528', '20180602'], // input: [start, end]
      ['20180528', '20180529', '20180530', '20180531', '20180601', '20180602'], // expected value is an array
    ],
]
```

packages/utils/src/__tests__/period/convertDateRangeToPeriods.test.js
```diff
@@ -7,139 +7,81 @@ import moment from 'moment';
...
+      ['201602', '201603', '20160401'],
+    ],
+    ['should handle last day of month', ['20160701', '20160831'], ['201607', '201608']],
+  ];
+
+  it.each(testData)('%s', (_, [startDate, endDate], expected) => {
+    const startMoment = moment.utc(startDate);
+    const endMoment = moment.utc(endDate);
+
+    expect(convertDateRangeToPeriods(startMoment, endMoment)).toStrictEqual(expected);
```
billli0, line 85: Nice Kostas! It makes sense to use `.toStrictEqual` . I kept thinking to convert `.to.have.ordered.members([...])` instead of thinking in the big picture.

packages/utils/src/__tests__/object.test.js
```diff
@@ -65,7 +65,7 @@ describe('object', () => {
         {},
         ['one', 'five'],
       ],
-    ])('%s', (name, object, options, expected) => {
+    ])('%s', (_, object, options, expected) => {
```
kael89: This is a detail, but handy whatsover: VSCode shows a warning for unused parameters, which is turned off for parameters with names using just underscore(s) (`_`, `__` etc). I don't if this is the case for WebStorm

packages/utils/src/__tests__/object.test.js
```diff
@@ -263,23 +270,24 @@ describe('object', () => {
...
-      [
-        'should create a set out of an object dictionary',
-        { id1: object1, id2: object2 },
-        'id',
-        expectedResult,
-      ],
-    ])('%s', (_, objectCollection, property, expected) => {
-      expect(reduceToSet(objectCollection, property)).toStrictEqual(expected);
+    const object1 = { id: 'id1', value: 10 };
+    const object2 = { id: 'id2', value: 20 };
```
kael89, line 263: We use the same `object1`, `object2` variables in lots of test suites here, but it's a long file and those are simple values so it's better to just redefine them in this case

At some point we could break this file into smaller ones, as you will see I did with `period.test.js` in later commits

packages/utils/src/__tests__/object.test.js
```diff
@@ -18,49 +18,42 @@ describe('object', () => {
     const testData = [
       [
         'should sort the keys of an object containing string values',
-        { fourth: 'd', third: 'c', second: 'b' },
-        { asc: true },
+        [{ fourth: 'd', third: 'c', second: 'b' }, { asc: true }],
```
kael89, line 21: This is an example of grouping all input params together. In other words:
```js
[[input, options], expected]
```

instead of 
```js
[input, options, expected]
```

Both seem perfectly valid to me! But I slightly prefer the first because the separation between input and output is more obvious

packages/utils/src/__tests__/object.test.js
```diff
@@ -208,26 +201,26 @@ describe('object', () => {
...
 
     describe('key mappers', () => {
-      const testData = [
-        ['string', [object1, object2], 'id', 'value', ['id1', 'id2']],
-        ['function', [object1, object2], object => object.value / 100, 'id', ['0.1', '0.2']],
-      ];
+      it('string', () => {
+        const result = reduceToDictionary([object1, object2], 'id', 'value');
+        expect(Object.keys(result)).toStrictEqual(['id1', 'id2']);
+      });
```
kael89, line 195: And here is an example where I felt that parameterized tests actually made the test cases harder to read, so I reverted back the changes

packages/utils/src/__tests__/WorkBookParser.test.js
```diff
@@ -58,7 +57,7 @@ describe('WorkBookParser', () => {
 
     it('should map the input to itself if no mapper is provided', async () => {
       const parser = new WorkBookParser();
-      await expect(parser.parse(WORK_BOOK)).resolves.toStrictEqual(SHEETS);
+      return expect(parser.parse(WORK_BOOK)).resolves.toStrictEqual(SHEETS);
```
kael89, line 60: `await` would be also correct instead of `return` above - I used `return` just for consistency

packages/utils/src/__tests__/validation/validatorFunctions.test.js
```diff
@@ -5,20 +5,24 @@
...
 
-describe('validationFunctions', () => {
+describe('validatorFunctions', () => {
   describe('allValuesAreNumbers', () => {
-    it.each([
-      ['fails if a not given an object', null, ''],
-      ['fails if a value is not a number', { hi: 'hello' }, "Value 'hi' is not a number: 'hello'"],
-    ])('%s', (_, object, expectedError) => {
-      expect(() => allValuesAreNumbers(object)).toThrowError(expectedError);
+    it('fails if a not given an object', () => {
```
kael89, line 10: Reverted to non-parameterized tests in this commit - seems more readable to me in this case

packages/utils/src/__tests__/period/period/period.test.js
```diff
@@ -0,0 +1,246 @@
...
+  momentToPeriod,
+  parsePeriodType,
+  PERIOD_TYPES,
+  periodToTimestamp,
+  periodToType,
+} from '../../../period/period';
+import { testConvertToPeriod } from './testConvertToPeriod';
+import { testGetPeriodsInRange } from './testGetPeriodsInRange';
+import { testPeriodToDisplayString } from './testPeriodToDisplayString';
+import { testComparePeriods } from './testComparePeriods';
```
kael89, line 25: This commit seems large but don't worry - I just extracted the longer test suites from `period.test.js` to the above files! It' s a pattern we commonly use to have more manageable and readable test files

The convention we use for those file names is:
`test{{TestedUnit}}`/`test{{TestedFunctionality}}`

packages/utils/src/__tests__/period/period/period.test.js
```diff
@@ -0,0 +1,246 @@
...
+      ['week', WEEK, '2020W52'],
+      ['day', DAY, '20201227'],
+    ];
+
+    it.each(testData)('%s', (_, periodType, expected) => {
+      expect(getCurrentPeriod(periodType)).toBe(expected);
+    });
+  });
+
+  describe('periodToDisplayString', testPeriodToDisplayString);
```
kael89, line 203: And here you can see an example of how we include the extracted tests in the main file. Those are the only changes in this commit, everything else is the same!

packages/utils/src/__tests__/object.test.js
billli0: Yeah the first one is easier to read. I'll stick to the first format in the future conversions.

packages/utils/src/__tests__/object.test.js
```diff
@@ -109,18 +99,20 @@ describe('object', () => {
           [beta, alpha, alphabet, alpha],
           [alpha, alpha, alphabet, beta],
         ],
-      ])('check sorted string values', (input, expected) => {
+      ];
+
+      testData.forEach(([input, expected]) => {
```
billli0, line 97: Good strategy using `testData.foreach `

billli0:
@kael89 Could you change the merge target branch from `1271-access-policy-use-jest-framework-only` to `1271-utils-use-framework-only`?

kael89:
Oh great spot @billli0 ! I was wondering why all those commits were there, even after rebasing

PR #1386, opened 2020-09-24
967: Update style guide - Part 1

kael89:
### Issue #:

**Still TODO:** (in future PRs)
* Use the new TypeScript style guide in `indicators`
* Run the new lint rules in `meditrak-app`, `meditrak-server`, `web-config-server`, `web-frontend`

**Note:** Only autofix rules were applied (whitespace/comment changes, or code safely refactored by ESLint). This is to minimise regression risk. Fixing more ESLint errors is possible, but it requires human intervention and I suggest doing it as a separate task


packages/admin-panel/src/pages/resources/SocialFeedPage.js
```diff
@@ -28,10 +28,7 @@ const FIELDS = [
     Header: 'Creation date',
     source: 'creation_date',
     type: 'tooltip',
-    accessor: row =>
-      moment(row.creation_date)
-        .local()
-        .toString(),
+    accessor: row => moment(row.creation_date).local().toString(),
```
kael89, line 31: Prettier 2 (which we now use) uses an, according to them, [Improved method chain breaking heuristic](https://prettier.io/blog/2020/03/21/2.0.0.html#improved-method-chain-breaking-heuristic-6685httpsgithubcomprettierprettierpull6685-by-mmkalhttpsgithubcommmkal). They only break chained methods in new lines if they are deemed to be complicated.

Is this new style better than the previous one? It's a matter of taste and I'm not 100% sure, but I think I like it

tcaiger: yeah I personally prefer this new style as it reads more like a sentence

tcaiger (approved):
I went through it file by file and everything looks accurate to me. I guess no surprise that eslint is reliable. Nice work - if we had to do all that refactoring manually, it would have taken forever!

kael89:
Merging in `dev` as dev-related

PR #1387, opened 2020-09-24
Optimise ui components bundle w/ code splitting

tcaiger:
ui-components library exports all its components in one file and is used by importing components as named imports. eg `import { NavBar } from '@tupaia/ui-components';`

This is nice and tidy on the import side of things but the problem is that the whole ui-components library is included in the app bundle when a single component is imported. 

The library could be a file for each component which would allow apps consuming the library to import one component at a time without bloating the bundle.

**Notes**
- This may not be an issue as AdminPanel and PSSS use pretty much all of the components anyway
- The current export doesn't allow lazy loading of components
- Exports need to be default exports to work with react lazy loading

The new imports would look like this:
`import { NavBar } from '@tupaia/ui-components/lib/NavBar';`

**Refactoring needed:**
- rename all imports from ui-components in psss and admin-panel
- maybe convert exports to defaults?


tcaiger:
Closing and have created an issue in the back log 

https://github.com/beyondessential/tupaia-backlog/issues/1347

PR #1388, opened 2020-09-24
1249: Creating a custom sort order for map overlay groups

chris-bes:
### Changes:

1. Added `sort_order` column to `map_overlay_group_relation` table
2. Created a root `World` (root) map overlay group so that top level overlay groups can connect to them and define `sort_order` (See: https://github.com/beyondessential/tupaia-backlog/issues/1109#issuecomment-698657929). This World map overlay group will not be shown in the UI.
3. Migrated `sortOrder` column from `mapOverlay` table to map_overlay_group_relation table. 
4. Deleted some orphan overlay relations. 
5. Added notifier to delete corresponding relations when overlays or overlay groups get deleted.


packages/database/src/modelClasses/MapOverlay.js
```diff
@@ -12,7 +12,20 @@ class MapOverlayType extends DatabaseType {
...
 
 export class MapOverlayModel extends DatabaseModel {
+  notifiers = [onChangeDeleteRelation];
+
   get DatabaseTypeClass() {
     return MapOverlayType;
   }
 }
+
+const onChangeDeleteRelation = async ({ type: changeType, record }, models) => {
```
chris-bes, line 22: This notifier is for removing any orphan relations in case devs forget when they remove mapOverlays/mapOverlayGroups (context: https://github.com/beyondessential/tupaia/pull/1074#discussion_r466104407).

packages/database/src/migrations/20200924035244-DeleteOrphanMapOverlayRelations-modifies-data.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
chris-bes, line 1: Some orphan relations that I've forgot to deleted in my migrations a while ago that I found out while working on this task

packages/database/src/migrations/20200925011104-CreateRootMapOverlay-modifies-data.js
```diff
@@ -0,0 +1,38 @@
...
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const WORLD_MAP_OVERLAY_GROUP = {
+  id: generateId(),
+  name: 'World',
+  code: 'World_Group',
+};
```
chris-bes, line 23: Create a world overlay group (context: https://github.com/beyondessential/tupaia-backlog/issues/1109#issuecomment-698657929). 

Thanks @kael89 for your input! I totally agree with your suggestion now. I initially thought it was a bit hacky when we discussed but when the implementation comes together I think it's great! I didn't find that I have to do anything dodgy to make this work

kael89: Good to know! I also had a bit suspicion that this may be "hacky", but the more about it the more it made sense. Eventually, I believe that our "hack" alert was triggered by the fact we will create a group that will not be displayed - but perhaps this is a view, not data model, issue. We could indeed display the root overlay group in the UI, if we chose to!

kael89: That's nice! I'm not sure about the name though... "World" brings entities to my mind. Perhaps we should name it something else, like "Root"? What do you think?

To keep things simple, the code can be the same as the name (eg "Root", not "Root_Group")

chris-bes: Ah yep, I was also picky between World and Root, but not sure why I went with World. Happy to change it to Root!

packages/database/src/migrations/20200924001437-AddSortOrderToMapOverlayGroupRelationTable-modifies-schema.js
```diff
@@ -0,0 +1,33 @@
...
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function(db) {
+  await db.runSql(`
+    ALTER TABLE map_overlay_group_relation 
+    ADD COLUMN sort_order real;
```
kael89, line 20: I know that we currently use `float4` (= `real`) for map overlay sort order, but we really shouldn't: we should use an integer type.

Based on the [Postgres docs on integer types](https://www.postgresql.org/docs/9.6/datatype-numeric.html), I believe we should go for `integer`. It will consume the same amount of memory with a `float4` type (4 bytes), but it represents the data model more accurately (+arithmetic/conversion may be more optimised, although that is not a concern).

kael89 (changes requested):
Great functionality and as always, really clean migrations! I'm glad this came together with not much additional complexity. A few implementation details to address and it should be ready to go!

packages/database/src/migrations/20200924001437-AddSortOrderToMapOverlayGroupRelationTable-modifies-schema.js
chris-bes: Thanks Kostas! The reason I kept them as `real` data type is some of the existing sortOrders of our Overlays are in decimal (I'm not sure why)...

However, I'm happy to convert all of them to integers so that we can use integer type in this case!

kael89: Yeah I'm pretty sure that was for convenience (infinite floats exist between two floats, while finite between two integers...) But this seems a big hack to me 😛 

Like, if you have sort order of 1 and 2, you can sneak in 1.5 there, then 1.6, then 1.55 etc...

packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -91,14 +93,25 @@ const checkIfGroupedMapOverlaysAreEmpty = nestedMapOverlayGroups => {
...
 
+const allRelationsHaveSortOrder = relations =>
+  relations.every(m => m.sort_order !== undefined && m.sort_order !== null);
+
+const getSortFunction = relations => {
+  const childIdToSortOrder = reduceToDictionary(relations, 'child_id', 'sort_order');
+  const sortBySortOrder = (m1, m2) => childIdToSortOrder[m1.id] - childIdToSortOrder[m2.id];
+
+  //If all the relations have sort_order, sort by the sort_order. Otherwise by default sort by name.
+  return allRelationsHaveSortOrder(relations) ? sortBySortOrder : getSortByKey('name');
```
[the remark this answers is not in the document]
chris-bes: I think my reasoning for either sort all by name or by orders (if all relations have sort order) is to encourage having sort orders for all map overlay items. In case that a dev missed an item when writing migration, I think it will be easier to spot. (I can't think of any cases that we actually want to sort only a portion of the items...?)

But if we want to be flexible then I think your suggested approach is certainly better.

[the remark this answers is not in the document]
kael89: @chris-bes That also makes sense! Let's leave a comment in the original PR so that business will now, and let's keep the existing functionality 🙌

packages/database/src/migrations/20200923065724-ConvertOverlaySortOrderToInteger-modifies-data.js
```diff
@@ -0,0 +1,59 @@
...
+    await db.runSql(`
+      UPDATE "mapOverlay"
+      SET "sortOrder" = ${index}
+      WHERE id = '${overlayId}'
+    `);
+  }
+};
+
+exports.up = async function (db) {
+  await reorderGroupOverlays(db, INTERNATIONAL_HEALTH_REGULATIONS_OVERLAYS_ORDERED);
```
kael89, line 48: Nice migration!

kael89 (approved):
All looks nice, great work!!

edmofro:
Looks like this is ready for merge @chris-bes! 🎉

PR #1390, opened 2020-09-24
Fix retry fetch location meditrak app

Alastair-L:
### Issue #:
https://github.com/beyondessential/meditrak/issues/411

Quite a few of the changes are from formatting on save


packages/meditrak-app/app/assessment/specificQuestions/GeolocateQuestion.js
```diff
@@ -57,7 +57,10 @@ class GeolocateQuestionComponent extends React.Component {
     if (isFindingLocation) {
       this.updateAnswer();
 
-      if (userLocation.errorMessage) {
+      if (
+        userLocation.errorMessage &&
+        this.props.answer.errorMessage !== userLocation.errorMessage
```
Alastair-L, line 62: This is for when the retry button has just been clicked, so that the error isn't immediately set again before "refreshWatchingUserLocation" has a chance to take effect.

packages/meditrak-app/app/assessment/specificQuestions/GeolocateQuestion.js
```diff
@@ -72,8 +75,8 @@ class GeolocateQuestionComponent extends React.Component {
       isFindingLocation: true,
       errorMessage: '',
     });
-
-    this.updateAnswer();
+    this.props.onRefresh();
+    this.props.onChangeAnswer({});
```
Alastair-L, line 79: This is the crux of the fix, clearing the errors on refresh.

packages/meditrak-app/app/assessment/specificQuestions/GeolocateQuestion.js
```diff
@@ -129,12 +135,12 @@ class GeolocateQuestionComponent extends React.Component {
...
-      latitude !== answer.latitude &&
-      longitude !== answer.longitude &&
-      accuracy !== answer.accuracy
-    ) {
+    const locationHasChanged =
+      latitude !== answer.latitude ||
+      longitude !== answer.longitude ||
+      accuracy !== answer.accuracy;
+
+    if ((latitude !== null || errorMessage) && locationHasChanged) {
```
Alastair-L, line 143: I'm pretty sure that we should consider the location to have changed if it only changes in one dimension?

packages/meditrak-app/app/utilities/userLocation/actions.js
```diff
@@ -46,6 +47,16 @@ export const watchUserLocation = () => async (dispatch, getState) => {
...
 
+export const refreshWatchingUserLocation = () => dispatch => {
+  dispatch({ type: CLEAR_LOCATION_DATA });
+
+  Geolocation.getCurrentPosition(
+    position => dispatch(onGeolocationPosition(position.coords)),
+    error => dispatch(onGeolocationError(error.message)),
+    GEOLOCATION_OPTIONS,
+  );
+};
```
edmofro, line 58: Is this necessary? The watch should still continue to update things when the location changes, I would have thought.

If the watch doesn't continue updating, then we also need to add a new watch here, similar to in `watchUserLocation`, so that the coordinates automatically update as you wander around

edmofro (changes requested):
Just one question

packages/meditrak-app/app/utilities/userLocation/actions.js
Alastair-L: You are totally right @edmofro, I must admit I have a very shallow understanding of what is going on here.

edmofro (approved):
lgtm :-)

Alastair-L:
https://github.com/beyondessential/meditrak/issues/411#issuecomment-700346279. This must have been why I added that `getCurrentPosition` call. Without it, once the watcher times out once, it cannot time out again.

In my mind this is what is happening (which is supported by my tests):
image

Does this make sense @edmofro? Is there a better way to go about it?

edmofro:
Right - that does make sense @Alastair-L. It's not what I expected, but I can understand it working that way :-)

edmofro (approved):

PR #1398, opened 2020-09-27
1271 auth convert mocha to jest

billli0:
### Issue #:
[1271- Use a single test framework : **auth module only**](https://github.com/beyondessential/tupaia-backlog/issues/1271)

Current issue: 

- [x] Jest did not exit one second after the test run has completed. (Resolved)


packages/auth/jest.config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "clearMocks": true,
```
kael89, line 2: In a following PR we can define this as a reusable config in the root level (`jest-internal-js.config.json`)

packages/auth/src/__tests__/buildAccessPolicy/testAccessPolicyHandler.js
```diff
@@ -0,0 +1,141 @@
...
+    await upsertDummyRecord(Demo.models.userEntityPermission, {
+      user_id: user.id,
+      entity_id: ottawa.id,
+      permission_group_id: Demo.publicPermission.id,
+    });
+
+    accessPolicy = await buildAccessPolicy(Demo.models, user.id);
+  });
+
+  describe('check permissions', () => {
```
kael89, line 95: Great parameterized tests in this file! 🎉

packages/auth/src/__tests__/buildAccessPolicy/testAccessPolicyHandler.js
```diff
@@ -0,0 +1,141 @@
...
+  // integration test
+  describe('can be interpreted by our access policy parser', () => {
+    let parser;
+
+    beforeAll(() => {
+      parser = new AccessPolicyParser(accessPolicy);
+    });
+
+    describe('individual checks', () => {
+      const testData = [
```
kael89, line 117: Since each test case doesn't provide much information here, perhaps it's better to use a plain `forEach` (instead of i`t.each`) and convert this block to an `it` block

packages/auth/src/__tests__/buildAccessPolicy/testAccessPolicyHandler.js
```diff
@@ -0,0 +1,141 @@
...
+        ['iii', [['CA_OT_OT', 'Public'], true]],
+        ['iv', [['CA_OT_TO', 'Public'], false]],
+        ['v', [['CA', 'Public'], false]],
+      ];
+      it.each(testData)('%s', (_, [[entity, permissionGroup], expected]) => {
+        expect(parser.allows(entity, permissionGroup)).toBe(expected);
+      });
+    });
+
+    describe('groups of entities', () => {
```
kael89, line 129: Similarly here, let's use a `forEach`

packages/auth/src/__tests__/AccessPolicyBuilder.test.js
```diff
@@ -0,0 +1,100 @@
...
+        [true, [spyOnBuildLegacyAccessPolicy, spyOnBuildAccessPolicy]],
+      ],
+    ];
+
+    it.each(testData)(
+      '%s',
+      async (_, [useLegacyFormat, [policyToBeCalled, policyNotToBeCalled]]) => {
+        const builder = new AccessPolicyBuilder(models);
+        builder.getPolicyForUser(userId, useLegacyFormat).then(_ => {
+          expect(policyToBeCalled).toHaveBeenCalledWith(models, userId);
```
kael89, line 53:
```suggestion
          expect(policyToBeCalled).toHaveBeenCalledOnceWith(models, userId);
```

packages/auth/src/__tests__/authenticator/testAuthenticatePassword.js
```diff
@@ -0,0 +1,181 @@
...
+          },
+          'Email address not yet verified',
+        ],
+      ],
+    ];
+
+    it.each(testData)('%s', async (_, [arg, expectedError]) => {
+      await authenticator
+        .authenticatePassword(arg)
+        .catch(e => expect(e.message).toBe(expectedError));
```
kael89, line 90: `toBeRejectedWith` can also be used here

packages/auth/src/__tests__/authenticator/testAuthenticatePassword.js
```diff
@@ -0,0 +1,181 @@
...
+            emailAddress: 'unverified@test.com',
+            password: 'validPassword',
+            deviceName: 'validDevice',
+          },
+          'Email address not yet verified',
+        ],
+      ],
+    ];
+
+    it.each(testData)('%s', async (_, [arg, expectedError]) => {
```
kael89, line 87:
```suggestion
    it.each(testData)('%s', async (_, input, expectedError) => {
```

packages/auth/src/__tests__/authenticator/testAuthenticatePassword.js
```diff
@@ -0,0 +1,181 @@
...
+  verifiedUser,
+  refreshToken,
+} from './Authenticator.fixtures';
+
+export const testAuthenticatePassword = () => {
+  const authenticator = new Authenticator(models, AccessPolicyBuilderStub);
+
+  describe('throws an error with invalid arguments', () => {
+    const testData = [
+      ['null', [undefined, "Cannot read property 'emailAddress' of undefined"]],
```
kael89, line 20:
```suggestion
      ['undefined', [undefined, "Cannot read property 'emailAddress' of undefined"]],
```

packages/auth/src/__tests__/authenticator/testAuthenticatePassword.js
```diff
@@ -0,0 +1,181 @@
...
+  accessPolicy,
+  MEDITRAK_DEVICE_DETAILS,
+  verifiedUser,
+  refreshToken,
+} from './Authenticator.fixtures';
+
+export const testAuthenticatePassword = () => {
+  const authenticator = new Authenticator(models, AccessPolicyBuilderStub);
+
+  describe('throws an error with invalid arguments', () => {
```
kael89, line 18: Great work on parameterizing the tests here! 🎉

kael89 (changes requested):
A lot of hard work was put in this PR! A few changes requested, which is expected since in this package we came across new test patterns.

Based on our experience till now, I enhanced the "Implementation per package" list in https://github.com/beyondessential/tupaia-backlog/issues/1271#issuecomment-693722114 with the following TODO steps after the migration script has run:

* Replace `then` with `await` syntax
* Replace `toHaveBeenCalledWith` with `toHaveBeenCalledOnceWith`, if the original code used Sinon's `calledOnceWithExactly` (I see you have already found this 👏 )
* Use `toBeRejectedWith` when asserting async function errors
* Delete mock reset calls, since `clearMocks: true` in `jest.config.json` takes care of that. An exception is when we want to reset mocks during the same `it` case

Those steps must be applied in this PR too.

Finally, I like your approach to encapsulate the common context of `buildAccessPolicy` tests under `Demo`. I have a few comments about that so that it can match existing patterns in our codebase, but let's discuss them in a second round of review.

Good luck with the changes! 🙌

packages/auth/src/__tests__/AccessPolicyBuilder.test.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  describe('handles caching and cache invalidation', () => {
+    it('avoids rebuilding the policy for the same user', async () => {
+      try {
+        const builder = new AccessPolicyBuilder(models);
+        await builder.getPolicyForUser(userId);
+        await builder.getPolicyForUser(userId);
+        expect(spyOnBuildAccessPolicy).toHaveBeenCalledTimes(1);
+      } catch (e) {}
```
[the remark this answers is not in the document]
billli0: image(`buildAccessPolicy.js line 18`)

The reason is that `.getPolicyForUser (line 64, 65)`  will trigger the code above. It will run `models.useEntitiesPermission.find` but the model we defined in `line 12` has no `find` method..
image

 I don't know how mocha can get pass but it will cause error in jest. With `try..catch..`, `.toHaveBeenCalledTimes` can be 'expected'.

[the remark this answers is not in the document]
kael89: Could you try adding an `expect(false).toBe(true)` statement and tell me if the test will fail?

packages/auth/src/__tests__/AccessPolicyBuilder.test.js
```diff
@@ -0,0 +1,100 @@
...
+      ],
+      [
+        'builds a legacy access policy when useLegacyFormat is set to true',
+        [true, [spyOnBuildLegacyAccessPolicy, spyOnBuildAccessPolicy]],
+      ],
+    ];
+
+    it.each(testData)(
+      '%s',
+      async (_, [useLegacyFormat, [policyToBeCalled, policyNotToBeCalled]]) => {
```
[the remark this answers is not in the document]
billli0: image

The test expectations are with different orders. Perhaps I should use a dissimilar name between these.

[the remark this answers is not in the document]
billli0: > `policyToBeCalled` is `spyOnBuildAccessPolicy` for all test data
> `policyNotToBeCalled` is `policyNotToBeCalled` for all test data
> 
> Thus, we don't have to pass them as parameters but can use them directly:

image

The test expectations are with different orders. Perhaps I should use a dissimilar name between these.

packages/auth/src/__tests__/AccessPolicyBuilder.test.js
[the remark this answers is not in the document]
billli0: image(`buildAccessPolicy.js line 18`)

The reason is that `.getPolicyForUser (line 64, 65)`  will trigger the code above. It will run `models.useEntitiesPermission.find` but the model we defined in `line 12` has no `find` method..
image

 I don't know how mocha can get pass but it will cause error in jest. With `try..catch..`, `.toHaveBeenCalledTimes` can work for the test purpose.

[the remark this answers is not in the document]
billli0: > What is the reason for catching an error and ignoring it here? This could be the test failing, which is something we would definitely want to be thrown up
> 
> Similarly for other cases below

image(`buildAccessPolicy.js line 18`)

The reason is that `.getPolicyForUser (line 64, 65)`  will trigger the code above. It will run `models.useEntitiesPermission.find` but the model we defined in `line 12` has no `find` method..
image

 I don't know how mocha can get pass but it will cause error in jest. With `try..catch..`, `.toHaveBeenCalledTimes` can work for the test purpose.

packages/auth/src/__tests__/AccessPolicyBuilder.test.js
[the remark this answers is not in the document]
kael89: Ooh I see now! No worries, I believe the rename I suggested will make this easier to spot (since now they both start with the same `spyOn` words) 👍

packages/auth/jest.config.json
```diff
@@ -0,0 +1,8 @@
...
+  "transform": { "^.+\\.jsx?$": "../../jestTransformer.js" },
+  "testMatch": ["<rootDir>/src/__tests__/authenticator/Authenticator.test.js"],
```
kael89, line 4: We'll also need to update the above

billli0: > We'll also need to update the above

Will do

packages/auth/src/__tests__/AccessPolicyBuilder.test.js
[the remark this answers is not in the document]
billli0: > Could you try adding an `expect(false).toBe(true)` statement and tell me if the test will fail?

`expect(false).toBe(true); ` didn't work with `try..catch..` But I change it like this now: 
image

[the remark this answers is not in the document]
kael89: So the tests were correctly passin in `mocha` because `buildAccessPolicy` and `buildLegacyAccessPolicy`, that call the `models.userEntityPermission.find` method, were stubbed (lines 24, 25 of the existing `AccessPolicyBuilder.test.js`). 

`spyOn` in lines 24 and 25 in your tests will not be enough here: we need to replace those functions with a jest mock (`jest.mock('../buildAccessPolicy')` may work)

packages/auth/src/__tests__/buildLegacyAccessPolicy.test.js
```diff
@@ -54,10 +53,14 @@ describe('buildLegacyAccessPolicy', () => {
     );
   });
 
+  afterAll(async () => {
+    await clearTestData(getTestDatabase());
+  });
```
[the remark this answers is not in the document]
billli0: > Are you sure this is required? Although it doesn't harm having this here, there wasn't an `after` case in the existing tests

Yes you are right. It is deleted.

packages/auth/package.json
```diff
@@ -20,5 +20,17 @@
...
+    "overrides": [
+      {
+        "files": [
+          "**/__tests__/**"
+        ],
+        "env": {
+          "jest": true
+        }
+      }
+    ]
```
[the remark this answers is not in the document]
billli0: > I believe the following should be enough after the recent style guide update:

Awesome!

packages/auth/src/__tests__/buildLegacyAccessPolicy.test.js
```diff
@@ -104,28 +107,43 @@ describe('buildLegacyAccessPolicy', () => {
...
-
-    it('should have Tonga admin, donor, and public access, as interpreted by legacy parser', () => {
-      expect(hasAccess(accessPolicy, 'surveys', ['TO'], 'Public')).to.equal(true);
-      expect(hasAccess(accessPolicy, 'surveys', ['TO'], 'Donor')).to.equal(true);
-      expect(hasAccess(accessPolicy, 'surveys', ['TO'], 'Admin')).to.equal(true);
+    const testData = [
+      [
+        'should have Demo Land public access, as interpreted by legacy parser',
+        [
+          [['surveys', ['DL'], 'Public'], true],
```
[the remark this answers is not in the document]
billli0: > 'surveys' is common in all test cases here, so it could be explicitly specified when `hasAccess` is called in line 131
> 
> Similarly for the other test data here

Good idea!

packages/auth/jest.config.json
billli0: > In a following PR we can define this as a reusable config in the root level (`jest-internal-js.config.json`)

I'm working on this now :sunglasses:

packages/auth/src/__tests__/jestScaffolding.js
```diff
@@ -5,11 +5,8 @@
...
-  done();
-});
-
-afterAll(async done => {
-  await clearTestData(getTestDatabase());
-  done();
+afterAll(async () => {
+  const database = getTestDatabase();
+  await clearTestData(database);
+  await database.closeConnections();
```
kael89, line 11: It seems we reached the same conclusion in parallel! 😆 https://github.com/beyondessential/tupaia/pull/1433/files

packages/auth/jest.config.json
```diff
@@ -2,7 +2,7 @@
   "clearMocks": true,
   "transform": { "^.+\\.jsx?$": "../../jestTransformer.js" },
   "testMatch": ["<rootDir>/src/__tests__/**/**.test.js"],
-  "setupFilesAfterEnv": ["../../jestScaffolding.js","./jest.setup.js"],
+  "setupFilesAfterEnv": ["../../jestScaffolding.js","./src/__tests__/jestScaffolding.js"],
```
kael89, line 5: I like the `jest.setup.js` file you originally used, we could even consider renaming `jestScaffolding`.js under the root to `jest.setup.base.js` or similar... So I suggest we keep it!

kael89 (approved):

PR #1399, opened 2020-09-27
Move jest transformer to root

billli0:
### Issue #:
JestTransformer.js is used for jest test in many packages.  

### Changes:
Move jest transformer to root so that we won't see it everywhere.


kael89 (approved):
Really nice! You can now update https://github.com/beyondessential/tupaia/pull/1398 by merging dev and using the common `JestTransformer`

PR #1403, opened 2020-09-28
1055: District, Provincial and National Level Dashboard: ICT Facilities bar graph

chris-bes:
### Changes:

- Added composeDataByDataPoint builder.
- Added ICT Facilities bar graph


packages/web-config-server/src/apiV1/dataBuilders/generic/compose/composeDataByDataPoint.js
```diff
@@ -0,0 +1,71 @@
...
+ * 
+ * Will be converted and returned:
+ * { 
+ *    data: 
+ *        [
+ *          { name: dataPoint1, builder1: 'ABC', builder2: 'GHI' }, 
+ *          { name: dataPoint2, builder1: 'DEF', builder2: 'JKL' }, 
+ *        ]
+ *    }
+ */
```
chris-bes, line 46: This is just a reverted version of `composeData` dataBuilder. For this report, we have to return the data like:
```
[
   {
		Pre-School: 0.444,
		Primary: 0.222,
		Secondary: 0.555,
		name: 'Electricity'
   },
   {
		Pre-School: 0.333,
		Primary: 0.444,
		Secondary: 0.555,
		name: 'Functioning smart TV'
	},
]
```

However, it is quite difficult to achieve calculating the percentages for each type of school (stored in attributes column). I guess it is possible, but I can't think of any proper way than adding more config to `percentagesOfValueCounts` data builder which just makes the data builder more complex.

Hence, I went with creating another version of the `composeData` dataBuilder which group the data by the dataPoint (see example documentation in the code)

chris-bes: Not sure if this is a good solution but happy for any other suggestions!

packages/web-config-server/src/apiV1/dataBuilders/generic/compose/composeDataByDataPoint.js
```diff
@@ -0,0 +1,71 @@
...
+ *            dataBuilder: <some databuilder>,
+ *            dataBuilderConfig: <some config>,
+ *        }
+ *    }
+ * }
+ * 
+ * Data returned from fetchComposedData():
+ * {
+ *     builder1: { 
+ *        data: [{name: dataPoint1: value: '111'}, {name: dataPoint2: value: '222'}}
```
Alastair-L, line 31:
```suggestion
 *        data: [{name: dataPoint1: value: '111'}, {name: dataPoint2: value: '222'}]
```

packages/web-config-server/src/apiV1/dataBuilders/generic/compose/composeDataByDataPoint.js
```diff
@@ -0,0 +1,71 @@
...
+ *     },
+ *     builder2: {
+ *        data: [{name: dataPoint1: value: '333'}, {name: dataPoint2: value: '444'}},
```
Alastair-L, line 34:
```suggestion
 *        data: [{name: dataPoint1: value: '333'}, {name: dataPoint2: value: '444'}],
```

packages/web-config-server/src/apiV1/dataBuilders/generic/compose/composeDataByDataPoint.js
```diff
@@ -0,0 +1,71 @@
...
+ * { 
+ *    data: 
+ *        [
+ *          { name: dataPoint1, builder1: '111', builder2: '333' }, 
+ *          { name: dataPoint2, builder1: '222', builder2: '444' }, 
+ *        ]
+ *    }
+ */
+
+export const composeDataByDataPoint = async (config, aggregator, dhisApi) => {
```
Alastair-L, line 48: To keep this consistent with other names, may I suggest `composeDataByDataClass`? Although that is already defined as a helper, so happy to keep it as is, or see if you can somehow replace `dataPoint` with `dataClass`.

packages/web-config-server/src/apiV1/dataBuilders/generic/compose/composeDataByDataPoint.js
```diff
@@ -0,0 +1,71 @@
...
+  const responses = await fetchComposedData(config, aggregator, dhisApi);
+  const dataObject = {};
+
+  Object.entries(responses).forEach(([responseName, { data: responseData }]) => {
+    if (!responseData || responseData.length === 0) return;
+    responseData.forEach(responseObject => {
+      const { name: dataPointName, value } = responseObject;
```
Alastair-L, line 55:
```suggestion
      const { name: dataClassName, value } = responseObject;
```
?

packages/web-config-server/src/apiV1/dataBuilders/generic/compose/composeDataByDataPoint.js
```diff
@@ -0,0 +1,71 @@
...
+
+      if (!dataObject[dataPointName]) {
+        dataObject[dataPointName] = {};
+        dataObject[dataPointName].name = dataPointName;
+      }
+
+      if (responseObject[`${dataPointName}_metadata`]) {
+        dataObject[dataPointName][`${responseName}_metadata`] = responseObject[`${dataPointName}_metadata`];
+      }
```
Alastair-L, line 64: Nice feature 🎉

Alastair-L (approved):
Nice work! I agree that creating a new dataBuilder is the cleanest way to do this. I love that it works at multiple levels out of the box 🎉

packages/web-config-server/src/apiV1/dataBuilders/generic/compose/composeDataByDataPoint.js
chris-bes: I agree with you, is `composeDataPerDataClass` okay?

Alastair-L: Yep, that sounds good (although heaven help anyone trying to tell the difference between that and `composeDataByDataClass` 😂. I think it's the best that you could do though with the existing code).

packages/web-config-server/src/apiV1/dataBuilders/generic/compose/composeDataPerDataClass.js
```diff
@@ -0,0 +1,71 @@
...
+ * { 
+ *    data: 
+ *        [
+ *          { name: dataClass1, builder1: '111', builder2: '333' }, 
+ *          { name: dataClass2, builder1: '222', builder2: '444' }, 
+ *        ]
+ * }
+ */
+
+export const composeDataPerDataClass = async (config, aggregator, dhisApi) => {
```
chris-bes, line 48: @edmofro Do you care about functional data builders (which doesn't extend from DataBuilder class)?

edmofro: Sometimes! In this case, there will be no conflict, but if the functional data builder either:
- Use a model through direct import, or 
- Called another functional data builder that relied on models, and didn't pass it in (in this case `fetchComposedData` will receive the models correctly through `config`)

PR #1407, opened 2020-09-29
1159 tile picker

tcaiger:
### Changes:

- Add new tile picker
- Add tile_sets column to project table which holds a comma separated list of tile set keys
- Add tile set options as a constant along with other mapbox config

---

### Screenshots:

Screenshots can be seen [here](https://github.com/beyondessential/tupaia-backlog/issues/1159)


packages/database/src/migrations/20200929223403-AddTileSetsToProject-modifies-scheme.js
```diff
@@ -0,0 +1,32 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function () {
+  const db = new TupaiaDatabase();
```
rohan-bes, line 20: Is there a reason we are doing the `new TupaiaDatabase()` here instead of doing `exports.up = async function (db) { ... }` ?

Does it trigger some functions?

rohan-bes (approved):
This is a really cool looking feature!! Nice work 😄 

Just a couple of comments but nothing blocking approval

packages/database/src/migrations/20200929223403-AddTileSetsToProject-modifies-scheme.js
tcaiger: To be honest, this is my first time writing a migration in Tupaia so I just looked up another example for reference and copied the syntax. I went back and looked at some more examples now and it seems that different ones are done different ways. I have updated it to the way you mentioned though as it makes more sense to me.

packages/database/src/migrations/20200929223403-AddTileSetsToProject-modifies-scheme.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
```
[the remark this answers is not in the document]
tcaiger: hehe that is definitely a typo. Well picked up, I must have made an error when running the `create-migrate` script. I have renamed it as you suggested

PR #1408, opened 2020-09-29
1271 aggregator convert mocha to jest

billli0:
### Issue #:
[1271- Use a single test framework : **aggregator module only**](https://github.com/beyondessential/tupaia-backlog/issues/1271)


packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/entityAggregations.test.js
```diff
@@ -25,18 +24,22 @@ const BASE_TEST_ANALYTICS = [
...
-        { dataElement: 'element1', organisationUnit: 'org2', period: '20200102', value: 2 },
-        { dataElement: 'element1', organisationUnit: 'org3', period: '20200103', value: 3 },
-      ]);
+      expect(sumPerOrgGroup(BASE_TEST_ANALYTICS, {})).toEqual(
+        expect.arrayContaining([
+          { dataElement: 'element1', organisationUnit: 'org1', period: '20200101', value: 1 },
+          { dataElement: 'element1', organisationUnit: 'org2', period: '20200102', value: 2 },
+          { dataElement: 'element1', organisationUnit: 'org3', period: '20200103', value: 3 },
+        ]),
+      );
```
kael89, line 33: In the test cases of this file, we indeed cannot use `.toEqual` because the order of the results is not guaranteed. While the existing code is correct, it can be made even simpler be using `toIncludeSameMembers` from [jest-extended](https://github.com/jest-community/jest-extended#toincludeanymembersmembers) (already installed in our monorepo)

```suggestion
      expect(sumPerOrgGroup(BASE_TEST_ANALYTICS, {})).toIncludeSameMembers([
        { dataElement: 'element1', organisationUnit: 'org1', period: '20200101', value: 1 },
        { dataElement: 'element1', organisationUnit: 'org2', period: '20200102', value: 2 },
        { dataElement: 'element1', organisationUnit: 'org3', period: '20200103', value: 3 },
      ]);
```

Similarly for the other test cases in this file

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/sumPreviousPerPeriod.test.js
```diff
@@ -0,0 +1,176 @@
...
+const { DAY } = PERIOD_TYPES;
+const { YEAR } = PERIOD_TYPES;
+
+describe('sumPreviousPerPeriod()', () => {
+  const testData = [
+    [
+      'should sum across periods',
+      [
+        [
+          [
```
kael89, line 18: Unfortunately I believe that the deeply nested data here will make the test cases harder to read in a parameterized structure... I suggest using the simple `it` blocks (no parameterization).

As in the previous comment, `.toIncludeSameMembers` can be used for the match

packages/aggregator/src/__tests__/analytics/periodFromAnalytics.test.js
```diff
@@ -0,0 +1,109 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { periodFromAnalytics } from '../../analytics/periodFromAnalytics';
+
+describe('periodFromAnalytics()', () => {
```
kael89, line 8: Same here, I believe that the non-parameterized version will be easier to read... Sorry for the back and forth!

I guess it's a bit of a judgement call in those cases. I was also probably over emphatic about parameterized tests early on in this project, in practice it seems they are a useful tool but not a "silver bullet". I appreciate that you took the time to go this way though!

The ideal use case would be when using simple input/output combinations. If you are in doubt, I suggest converting one test block to parameterized and see how it looks like. It may take a bit of extra time experimenting but I think it's worth it in the long run!

packages/aggregator/src/__tests__/analytics/filterAnalytics.test.js
```diff
@@ -0,0 +1,112 @@
...
+import { filterAnalytics } from '../../analytics/filterAnalytics';
+
+const analytics = [{ value: 1 }, { value: 2 }, { value: 3 }];
+
+describe('filterAnalytics()', () => {
+  describe('empty filter', () => {
+    const testData = [
+      ['i', []],
+      ['ii', [undefined]],
+      ['iii', [{}]],
```
kael89, line 15: We can use descriptive names here. In order: 'empty array', 'undefined', 'empty object'

packages/aggregator/src/__tests__/analytics/filterAnalytics.test.js
```diff
@@ -0,0 +1,112 @@
...
+                { value: { operator: 'unknownFilter', value: 0 } },
+                [{ value: 1 }, { value: 2 }, { value: 3 }],
+              ],
+            ],
+          ],
+        ],
+      ],
+    ];
+    testData.forEach(([name, testSuites]) => {
+      describe(name, () => {
```
kael89, line 103: While creating a dynamic `describe` here is clever, I believe it will be actually easier to read if we would break this into two describe blocks with hard coded descriptions, even if this means we have to duplicate the `it.each` logic below

That way the array will have less nesting and may potentially achieve a good balance between file-level and line-level readability! (Direct matchers are readable in line-level, parameterized are in file-level)

kael89 (changes requested):
Looks great overall, just some changes here and there and this should be ready too!

----

By the way, it's a bit weird that, while all tests pass, I get this error which indicates a database connectivity issue:

image

I would assume that some of the mocks which involve DB mocking may be incomplete. We can have a better look at this after the changes here are addressed.

billli0:
> Looks great overall, just some changes here and there and this should be ready too!
> 
> By the way, it's a bit weird that, while all tests pass, I get this error which indicates a database connectivity issue:
> 
> image
> 
> I would assume that some of the mocks which involve DB mocking may be incomplete. We can have a better look at this after the changes here are addressed.

Thanks Kostas for your comments!:raised_hands: I can see there is a few mistakes that I've made in code..:persevere: Not sure how it passes...Thanks again to finding them out. 

Rgarding this error, I solved this by adding SSL config in Postgre setting. I'm not quite sure whether enable SSL connection is part of the test requirement, we can investigate later.:sunglasses:

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/entityAggregations.test.js
billli0: I can understand your confusion about `.toEqual(expect.arrayContaining())` here, which is exactly the same as `.toIncludesSameMembers`. Definely I will change it to `.toIncludesSameMembers()` because I agree it is hard to read.

packages/aggregator/src/__tests__/analytics/periodFromAnalytics.test.js
billli0: Don't feel sorry Kostas. Thanks for pointing it out. This was done before you mentioned compromising and I forgot to make it right. I should check all code again later. Anyway, hopefully now I 'extend' your judgement mostly :sunglasses:

kael89: Thanks for that, also keen to listen to your judgement on this matter! Feel free to take decisions based on our taste in this subject for the following work. If it doesn't feel right, we don't have to use parameterized tests 🙌

packages/aggregator/src/__tests__/analytics/filterAnalytics.test.js
billli0: Totally agree!

packages/aggregator/src/__tests__/analytics/filterAnalytics.test.js
```diff
@@ -0,0 +1,113 @@
...
+      ['undefined', [undefined]],
+      ['empty object', [{}]],
+    ];
+
+    it.each(testData)('%s', (_, [filter]) => {
+      expect(() => filterAnalytics(analytics, filter)).not.toThrow();
+    });
+  });
+
+  describe('filter type', () => {
```
kael89, line 23: It's a bit subjective but I believe that the tests here are very readable now! 🎉

packages/aggregator/src/__tests__/createAggregator.test.js
```diff
@@ -1,9 +1,9 @@
-import { expect } from 'chai';
-
 import { DataBroker } from '@tupaia/data-broker';
 import { Aggregator } from '../Aggregator';
 import { createAggregator } from '../createAggregator';
 
+jest.mock('@tupaia/data-broker');
```
kael89, line 5: This fixed the database related issues... `DataBroker[14]` calls `getModels()`, which creates a database instance. We don't need the actual database, or even `DataBroker` here so it's safe (and actually a good practice) to mock it.

Those issues should exist in `mocha` too, but for some reason it seems to be more forgiving with open handles than `jest`.

kael89 (approved):

packages/aggregator/src/__tests__/createAggregator.test.js
billli0: Thanks for the clear explaination. I've learned how you mock modules with the best strategy!

PR #1409, opened 2020-09-29
Add tonga covid dashboards and overlays to Samoa

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1254
### Changes:
1. Added the required dashboard report to the Samoan dashboard
2. Added 'WS' to the country codes of some overlays
3. For others, the group name needed to be changed, so all the overlays needed to be duplicated. Possibly could have something like `$countryName` in the title and replace it differently for each country, but this seemed simpler given that time was/is a factor.


packages/database/src/migrations/20200910024211-CreateNewCovidDashboardSamoa-modifies-data.js
```diff
@@ -0,0 +1,123 @@
...
+      set "countryCodes" = "countryCodes" || '{WS}'
+      where id = '${id}';
+    `);
+  });
+
+  // 3. Copy overlays so that the group name can be changed to Samoa without having duplicate groups
+  const overlaysToCopy = (await getOverlaysForGroup(db, 'COVID19_Tonga')).rows;
+
+  const mapOverlayGroupId = generateId();
+  insertObject(db, 'map_overlay_group', {
```
chris-bes, line 60: I think we need `await` here?

packages/database/src/migrations/20200910024211-CreateNewCovidDashboardSamoa-modifies-data.js
```diff
@@ -0,0 +1,123 @@
...
+    name: 'COVID-19 Samoa',
+    code: 'COVID19_Samoa',
+  });
+
+  overlaysToCopy.forEach(overlayObject => {
+    const id = `Samoa_${overlayObject.id}`;
+
+    // eslint-disable-next-line no-param-reassign
+    delete overlayObject.linkedMeasures;
+    insertObject(db, 'mapOverlay', {
```
chris-bes, line 71: Same that we might need `await` here. And if so we might also need to convert the forEach to for loop

chris-bes: And we might also need to do the same in the `exports.down` code too

Alastair-L:
Thanks Chris. Unsure why it was working... (unless I was just imagining it)

chris-bes (approved):
Looks great!

PR #1413, opened 2020-09-29
824: Script to create config for full-scale report testing

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/824

script run


packages/web-frontend/config/webpack.config.dev.js
```diff
@@ -161,6 +161,7 @@ module.exports = {
           // It enables caching results in ./node_modules/.cache/babel-loader/
           // directory for faster rebuilds.
           cacheDirectory: true,
+          presets: ['react-app'],
```
kael89, line 164: In the past, we relied on `.babelrc` configuration to be picked up by `babel-loader` in our webpack configuration. This causes some issues with the new `cypress:generate-config` script:

`babel-node cypress/scripts/generateConfig --config-file '../../babel.config.json'`

In this script, we just want to use the specified `config-file` and not include `.babelrc`. Unfortunately, `.babelrc` is picked up automatically by babel.

A solution would be using a `babelrc.json` file, which doesn't get picked up automatically. But then this file isn't picked up by babel-loader 🤷 

My theory is that this happens because `.babelrc` is the (legacy, but still supported) babel configuration file, and `babelrc.json` is a new config file introduced in Babel 7. Apparently `babel-loader` hasn't caught up with this change.

edmofro: Well great work on finding a solution!

packages/web-frontend/cypress/scripts/generateDashboardReportConfig.js
```diff
@@ -0,0 +1,62 @@
...
+      reportIdToGroup[reportId] = dashboardGroup;
+    });
+  });
+  return reportIdToGroup;
+};
+
+/**
+ * We generate the least amount of config required to test each report once.
+ * 1. For each report, use the first dashboard group that includes it
+ * 2. For each dashboard group, use the first project that includes it
```
edmofro, line 31: Nice!

edmofro: Just to be persnickety, it's not actually "the first" dashboard group, but rather the last based on the order the db returns them

edmofro (changes requested):
Two minor changes requested - but this is really cool man! Exciting stuff :-)

packages/web-frontend/.env.example
```diff
@@ -1 +1,6 @@
+DB_DISABLE_SSL=
+DB_NAME=
+DB_PASSWORD=
+DB_URL=
+DB_USER=
```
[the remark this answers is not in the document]
kael89: Good call!

packages/web-frontend/cypress/scripts/generateDashboardReportConfig.js
kael89: Oh good point! By the way, we may want to change this strategy to include more combinations, or we can support multiple strategies based on a script option.... But for now the "minimum" combinations strategy will work

scripts/node/utilities.js
```diff
@@ -5,20 +5,26 @@
...
+          winston.format.colorize({ all: true }),
+          winston.format.printf(({ message }) => message),
+        ),
+      }),
+    ],
+  });
 
-module.exports = {
-  logger: winston,
+const isWinstonConfigured = false;
```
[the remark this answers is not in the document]
kael89: Oopsy!

edmofro (approved):

kael89:
Merging straight into `dev` as dev-related

PR #1414, opened 2020-09-29
fix countevent groupby not grouped by datvalues

nick-beyondessential:
https://github.com/beyondessential/tupaia-backlog/issues/1348
**Tonga HPU Number of NCD risk factor screening events (by setting) not showing data**

A recent change to groupby logic in countEvents datbulder rendered a couple of existing dashboards broken. The fix was to be more specific on the getDataElementCodes groupby logic.


IgorNadj:
Ah sorry about that, I thought I tested the others. Good fix

IgorNadj (approved):

PR #1415, opened 2020-09-30
Convert user and permissions GET filtering to pure SQL

EMcQ-BES:
### Changes:

Converted permissions filtering for user and permissions GET functions to use pure SQL implementation.

### DIFF
--- a/packages/access-policy/src/index.js
+++ b/packages/access-policy/src/index.js
@@ -115,7 +115,7 @@ export class AccessPolicy {
    *
    * @returns entities[] The entity objects
    */
-  getEntitiesByPermission(permissionGroup) {
+  getEntitiesAllowed(permissionGroup) {
     return Object.keys(this.policy).filter(e => this.allows(e, permissionGroup));
   }
 }
--- a/packages/access-policy/src/index.js
+++ b/packages/access-policy/src/index.js
@@ -107,4 +107,15 @@ export class AccessPolicy {
     }
     return this.cachedPermissionGroupSets[cacheKey];
   }
+
+  /**
+   * Return entities the user has access to the given permission group for
+   *
+   * @param {string} [permissionsGroup]
+   *
+   * @returns entities[] The entity objects
+   */
+  getEntitiesByPermission(permissionGroup) {
+    return Object.keys(this.policy).filter(e => this.allows(e, permissionGroup));
+  }
 }
--- a/packages/meditrak-server/src/routes/accessRequests/GETAccessRequests.js
+++ b/packages/meditrak-server/src/routes/accessRequests/GETAccessRequests.js
@@ -3,18 +3,16 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-import { QUERY_CONJUNCTIONS } from '@tupaia/database';
 import { GETHandler } from '../GETHandler';
 import {
   assertAnyPermissions,
   assertBESAdminAccess,
   assertTupaiaAdminPanelAccess,
-  hasBESAdminAccess,
 } from '../../permissions';
-import { TUPAIA_ADMIN_PANEL_PERMISSION_GROUP } from '../../permissions/constants';
-import { assertAccessRequestPermissions } from './assertAccessRequestPermissions';
-
-const { RAW } = QUERY_CONJUNCTIONS;
+import {
+  assertAccessRequestPermissions,
+  createAccessRequestDBFilter,
+} from './assertAccessRequestPermissions';
 
 /**
  * Handles endpoints:
@@ -46,20 +44,12 @@ export class GETAccessRequests extends GETHandler {
   }
 
   async findRecords(criteria, options) {
-    const dbConditions = criteria;
-    if (!hasBESAdminAccess(this.accessPolicy)) {
-      // If we don't have BES Admin access, add a filter to the SQL query
-      const countryList = this.accessPolicy.getEntitiesByPermission(
-        TUPAIA_ADMIN_PANEL_PERMISSION_GROUP,
-      );
-      countryList.push('DL'); // If we have admin panel anywhere, we can also view Demo Land
-      const entities = await this.models.entity.find({
-        code: countryList,
-      });
-      const entityIds = entities.map(e => e.id);
-      dbConditions.entity_id = entityIds;
-    }
-    const accessRequests = await super.findRecords(criteria, options);
+    const dbConditions = await createAccessRequestDBFilter(
+      this.accessPolicy,
+      this.models,
+      criteria,
+    );
+    const accessRequests = await super.findRecords(dbConditions, options);
 
     if (!accessRequests.length) {
       throw new Error('Your permissions do not allow access to any of the requested resources');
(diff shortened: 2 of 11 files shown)


packages/access-policy/src/index.js
```diff
@@ -107,4 +107,15 @@ export class AccessPolicy {
...
   }
+
+  /**
+   * Return entities the user has access to the given permission group for
+   *
+   * @param {string} [permissionsGroup]
+   *
+   * @returns entities[] The entity objects
+   */
+  getEntitiesByPermission(permissionGroup) {
```
edmofro, line 118:
```suggestion
  getEntitiesAllowed(permissionGroup) {
```
For consistency with all the 'allow' terminology above, and we also have a bit of a convention that `getXByY` returns `{ y1: [x1, x3], y2: [x2]}` rather than taking `y1` and returning `[x1, x3]`
Please also add this to the readme of access policy - it's used by a bunch of internal packages, so the docs can come in handy to show what is publicly accessible

packages/meditrak-server/src/routes/accessRequests/GETAccessRequests.js
```diff
@@ -3,16 +3,18 @@
...
   assertAnyPermissions,
   assertBESAdminAccess,
   assertTupaiaAdminPanelAccess,
+  hasBESAdminAccess,
 } from '../../permissions';
-import {
-  assertAccessRequestPermissions,
-  filterAccessRequestsByPermissions,
-} from './assertAccessRequestPermissions';
+import { TUPAIA_ADMIN_PANEL_PERMISSION_GROUP } from '../../permissions/constants';
```
edmofro, line 14: Instead of 'reaching in', we could export the permissions constants from the root of the permissions folder

packages/meditrak-server/src/routes/accessRequests/GETAccessRequests.js
```diff
@@ -44,15 +46,22 @@ export class GETAccessRequests extends GETHandler {
   }
 
   async findRecords(criteria, options) {
+    const dbConditions = criteria;
+    if (!hasBESAdminAccess(this.accessPolicy)) {
+      // If we don't have BES Admin access, add a filter to the SQL query
+      const countryList = this.accessPolicy.getEntitiesByPermission(
```
edmofro, line 52:
```suggestion
      const accessibleCountryCodes = this.accessPolicy.getEntitiesByPermission(
```

edmofro (changes requested):
Looking good!

packages/access-policy/src/index.js
EMcQ-BES: Renamed and added to readme

packages/meditrak-server/src/routes/accessRequests/GETAccessRequests.js
EMcQ-BES: Looks like this was actually already the case, I was just following a pattern from an outdated file.

packages/meditrak-server/src/routes/accessRequests/GETAccessRequests.js
EMcQ-BES: Renamed. I've also moved the creation of SQL filters into the permissions files to make this read a bit cleaner.

packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
```diff
@@ -3,45 +3,35 @@
...
   models,
   userEntityPermission,
 ) => {
-  const results = await filterUserEntityPermissionsByPermissions(
-    accessPolicy,
-    [userEntityPermission],
-    models,
-  );
-  if (results.length === 0) {
+  const entity = await models.entity.findById(userEntityPermission.entity_id);
```
chris-bes, line 13: I've looked into the problem reported by Klaus this morning and I think it is related to this line of code. This might not work for the Admin Panel because it doesn't request for `entity_id` column. So userEntityPermission record does not have `entity_id` value 😢 

permission

packages/meditrak-server/src/routes/accessRequests/assertAccessRequestPermissions.js
```diff
@@ -12,3 +12,22 @@ export const assertAccessRequestPermissions = async (accessPolicy, models, acces
...
+export const createAccessRequestDBFilter = async (accessPolicy, models, criteria) => {
+  if (hasBESAdminAccess(accessPolicy)) {
+    return criteria;
+  }
+  // If we don't have BES Admin access, add a filter to the SQL query
+  const dbConditions = criteria;
+  const accessibleCountryCodes = accessPolicy.getEntitiesAllowed(
+    TUPAIA_ADMIN_PANEL_PERMISSION_GROUP,
+  );
+  accessibleCountryCodes.push('DL'); // If we have admin panel anywhere, we can also view Demo Land
```
edmofro, line 25: This looks like it might return demo land info for people who have no admin panel access at all?

packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
```diff
@@ -16,3 +16,22 @@ export const assertUserEntityPermissionPermissions = async (
   }
   return true;
 };
+
+export const createUserEntityPermissionDBFilter = async (accessPolicy, models, criteria) => {
```
edmofro, line 20: Is this _exactly_ the same as the access request one? Can they use a common helper?

packages/meditrak-server/src/routes/accessRequests/assertAccessRequestPermissions.js
EMcQ-BES: You wouldn't get to here without any admin panel access because of how we gate route handlers with `assertUserHasAccess`

edmofro: Right, that makes sense. However, I think we should have that logic here as well - coupling logic in different files could be dangerous, especially when this is a nicely named util that someone might pick up to use somewhere else.

packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
EMcQ-BES: Good catch! I've swapped out the permission checks to accept the `id` rather than the object. Then we look up the model before running the checks so we know exactly what we're dealing with.

I've swapped this over for the user and permission endpoints that this PR is touching, I'll fix up the rest alongside the rest of the SQL-ized work in another PR.

packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
EMcQ-BES: I've abstracted this out into a `createAdminPanelDBFilter` utility

packages/meditrak-server/src/routes/accessRequests/assertAccessRequestPermissions.js
EMcQ-BES: Added a length check to ensure we have __some__ permissions before adding demo land

edmofro (approved):
Looks great!

PR #1418, opened 2020-09-30
Minor jest test updates

kael89:
### Issue #:
[No issue]

We forgot to update `checkValueSatisfiesCondition.test.js` to use the new parameterized test structure


billli0 (approved):
Awesome Kostas! I should have a look at `access-policy` with the updated structure standard.

kael89:
@billli0 I ended up destructuring the test cases... I wasn't sure if a big array of data is readable enough, especially since readability is often subjective. I got mixed feedback so I guess it's safer to use a compromise!

billli0:
> @billli0 I ended up destructuring the test cases... I wasn't sure if a big array of data is readable enough, especially since readability is often subjective. I got mixed feedback so I guess it's safer to use a compromise!

I agree. The structure looks better now.

billli0 (approved):

PR #1419, opened 2020-09-30
Add a base jest config in root

billli0:
Add a `jest.config.base.json` in root. It is resuable in sub pacakges while jest testing. 

If any pacakges need extension jest config, you can easily add a custom jest config under sub pacakge.

### DIFF
--- a/jest.config.base.json
+++ b/jest.config.base.json
@@ -1,0 +1,7 @@
+{
+  "clearMocks": true,
+  "transform": { "^.+\\.jsx?$": "../../jestTransformer.js" },
+  "testMatch": ["<rootDir>/src/__tests__/**/**.test.js"],
+  "transformIgnorePatterns": ["/node_modules/"],
+  "testEnvironment": "node"
+}


kael89 (approved):
Nice! The next step is updating existing test scripts to use this config, and delete the local configs, where this is possible

billli0:
We can do it after converting all the mocha to jest. For now we need to change `jest.config.json` to `jest.config.js`, because `json` is not extendable. However, a better approach is to do it from the bigger picture using [jest project config](https://jestjs.io/docs/en/configuration#projects-arraystring--projectconfig)

billli0:
Dependency: [1271: Use a single test framework](https://github.com/beyondessential/tupaia-backlog/issues/1271)

kael89:
@billli0 question: isn't there a way to use the base `jest` config just in the packages that use jest? I would assume that #1271 is not a blocker

billli0:
Yes, but we only can copy and paste the jest.base.json to the subpackage to make it works. 
Another thing is, with JSON format, it can not be extended by the sub packages' `jest.config.json`. We can do it either to use JS format or using [Jest `projects` config](https://jestjs.io/docs/en/configuration#projects-arraystring--projectconfig).
Overall, adding a jest.base.json could be a temporary solution just for copy and paste usage. Give a thumb up if that sounds good.

billli0:
I realise this `jest.base.json` can be further used. 😐

PR #1420, opened 2020-09-30
Added CreateUserAccount endpoint and added permissions checker

chris-bes:
### Changes:

- Added generic CreateHandler and CreateUserAccount handler

More context: https://github.com/beyondessential/tupaia-backlog/issues/1028#issuecomment-701270414


packages/meditrak-server/src/routes/createUser.js
```diff
@@ -24,8 +24,9 @@ const PERMISSION_GROUPS = {
 
 const DEMO_LAND_NAME = 'Demo Land';
 
+//Used for registering user account
 export const createUser = async (req, res) => {
-  await req.assertPermissions(assertBESAdminAccess);
+  await req.assertPermissions(allowNoPermissions);
```
chris-bes, line 29: This endpoint is actually for registering user account from Tupaia webpage so I think this should actually be for "anyone"

edmofro (commented):
I see @EMcQ-BES has picked this up for review, but my question after a brief look is whether the two create user handlers could share more code? Perhaps the one servicing tupaia.org could extend the other, but put more restrictions on what data is allowed to be sent through (so they can't specify their own email verified status, for example)?

packages/meditrak-server/src/routes/createUser.js
```diff
@@ -24,8 +24,9 @@ const PERMISSION_GROUPS = {
 
 const DEMO_LAND_NAME = 'Demo Land';
 
+//Used for registering user account
 export const createUser = async (req, res) => {
```
EMcQ-BES, line 28: Should we convert this to a CreateHandler to keep consistent?

packages/meditrak-server/src/app/addRoutesToApp.js
```diff
@@ -153,6 +154,7 @@ export function addRoutesToApp(app) {
   app.post('(/v[0-9]+)/import/users', upload.single('users'), importUsers);
   app.post('(/v[0-9]+)/import/optionSets', upload.single('optionSets'), importOptionSets);
   app.post('(/v[0-9]+)?/user', createUser);
+  app.post('(/v[0-9]+)?/userAccount', createUserAccount);
```
EMcQ-BES, line 157: Might be nice to have a comment explaining the difference between these end points. Would be confusing if you're reading this without context.

EMcQ-BES (approved):
Love it! I appreciate the separate verbiage to make a distinction between the endpoints.

PR #1425, opened 2020-10-01
Fix webpack

kael89:


edmofro (approved):
Complicated stuff! It looks fine to my untrained eye

PR #1427, opened 2020-10-01
Pre push validation

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/350

#### Branch name is valid, no `.only` tests found

The branch can be pushed normally.
image

#### Branch name is invalid

Husky prevents the user from pushing the branch:

Too long name
image

Invalid ending
image

Includes uppercase char:
> ❌ Branch name cannot contain uppercase characters

Includes invalid char (/, \, ., &, ?). Eg for `&`
> ❌ Invalid character in branch name: '&'

#### Branch name is valid, `.only` tests found

The user sees a list of the `.only` tests found, and they get a warning. They can still push their branch, since people may  want to disable tests between commits. If the `.only` tests are committed, the dev build will fail in the "validate" step (existing behaviour)
image
image


codeship-steps.yml
```diff
@@ -7,15 +7,9 @@
...
-          steps:
-            - name: validate meditrak-server code
-              service: validation
-              command: './packages/devops/scripts/ci/validateCode.sh meditrak-server'
-            - name: validate web-config-server code
-              service: validation
-              command: './packages/devops/scripts/ci/validateCode.sh web-config-server'
+        - name: validate tests
+          service: validation
+          command: './packages/devops/scripts/ci/validateTests.sh'
```
kael89, line 12: Currently our build continues to run even if those steps (which are quickly executed) fail... I believe we should stop the build as soon as `validation` fails

packages/meditrak-server/package.json
```diff
@@ -60,8 +60,7 @@
     "prestart": "npm run -s build",
     "start": "node dist",
     "start-dev": "../../scripts/bash/backendStartDev.sh 9999",
-    "start-verbose": "LOG_LEVEL=debug npm run start-dev",
-    "validate": "babel-node scripts/validate/index.js --config-file '../../babel.config.json'"
```
kael89, line 64: We could still allow per-package specific validation in the future, although the more generic and centrally defined a validation is, the better, I suppose

scripts/node/validateTests.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+const PACKAGES = ['meditrak-server', 'web-config-server'];
+const MOCHA_TEST_BLOCKS = ['it', 'describe', 'context', 'specify'];
+const STD_ERROR_CODE = 1;
+
+const getTestDirectory = packageName =>
+  path.resolve(`${__dirname}/../../packages/${packageName}/src/tests`);
+
+const findExclusiveTests = async packageName => {
```
kael89, line 23: This file is based on the now deleted `validateNoExclusiveTestsExist.js`, with a few changes... Unfortunately it's not easy to see in the diff here.

Also, it should use the `Scripts` class (defined in the same folder), but I didn't do this for ease of reviewing (and to avoid spending too much time)... Happy to do that in a following review.

Finally, we could consider using a dedicate cli creation library. [yargs](https://www.npmjs.com/package/yargs), [commander](https://www.npmjs.com/package/commander) seem popular choices. I'm not familiar with any of those, but they can probably assist as in creating elegant node scripts in the future!

packages/devops/scripts/ci/validateTests.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+DIR=`dirname "$0"`
+ROOT="${DIR}/../../../../"
+
+node ${ROOT}/scripts/node/validateTests
```
kael89, line 6: So the validation scripts are a bit scattered around now... This one, defined in `devops` invokes another one defined in the root.

It will make things simpler to move all those in one location, probably `validate...` ones (which are not only devops specific) under the root? Or we could create a `scripts` package, which can also include reusable scripts (eg `build` for internal packages). This would be similar to how `react-scripts` is used by `create-react-app` projects.

packages/devops/scripts/ci/utils.sh
```diff
@@ -0,0 +1,22 @@
+ANSI_COLOR_GREEN="\e[32m"
+ANSI_COLOR_RED="\e[31m"
+ANSI_COLOR_RESET="\e[0m"
+ANSI_COLOR_YELLOW="\e[33m"
+
+function log_with_color() {
+  text=$1
+  color=$2
+  echo -e "$2$1$ANSI_COLOR_RESET"
+}
```
rohan-bes, line 10: Oooh this is very handy 😄

packages/devops/scripts/ci/validateBranchName.sh
```diff
@@ -1,11 +1,32 @@
...
-    if [[ "$CI_BRANCH" == *$reserved_ending ]]; then
-        echo "Invalid branch name, must not end in '$reserved_ending'"
+    if [[ "$branch_name" == *$reserved_ending ]]; then
+        log_error "❌ Invalid branch name ending: '$reserved_ending'"
         exit 1;
     fi
 done
+
+log_success "✔ Branch name is valid!"
+exit 0;
```
rohan-bes, line 32: I think there are a couple more restrictions on branch names:
- No uppercase letters
- Characters which may interfere with URL namespaces must not be included (eg. `/ \ . & ?` etc.)
- A couple more namespace reserved endings: `aggregation`, `psss`?)

https://docs.beyondessential.com.au/books/software-development/page/git-branching-strategy-and-naming-conventions

rohan-bes (changes requested):
Nice one! This is great to be part of our pipeline 😸 

Just a couple changes for branch name validation 👍

packages/devops/scripts/ci/validateBranchName.sh
kael89: Oh great call! I will update the functionality accordingly

packages/devops/scripts/ci/utils.sh
kael89: Yeah color adds a new dimension to logs 🌈

packages/devops/scripts/ci/validateBranchName.sh
kael89: Do you have an idea of how I could find a comprehensive list of

> Characters which may interfere with URL namespaces

If not, I believe using `/ \ . & ?` would be good enough for now - we can always update the list if need be

packages/devops/scripts/ci/validateBranchName.sh
```diff
@@ -1,11 +1,43 @@
 #!/bin/bash
 
-reserved_endings=(api config export mobile admin www)
+DIR=`dirname "$0"`
+. ${DIR}/utils.sh
 
-for reserved_ending in ${reserved_endings[@]}
+MAX_LENGTH=56
+RESERVED_CHARS=('/' '\' '.' '&' '?')
```
kael89, line 7: @rohan-bes Some of those characters (eg `?`) are not disallowed from git in the first place, so a branch name wouldn't include them. do you think it's worth removing them from this list?

I'm fine either way

rohan-bes: Yeah I'm fine either way too 👍

kael89: I'll leave it as is then, to match our doco

packages/devops/scripts/ci/validateBranchName.sh
rohan-bes: I think I got that list just from googling which characters would need to be escaped for a URL originally hehe :P

I think that's fine for now, if anyone encounters others they can just update the script. It's not common for anyone to include those special chars in their branch names anyway

rohan-bes: @kael89 is it possible to check if the branch has upper case letters? If I recall upper case letters causes failure re-deploying to feature branches

kael89: Sorry, my bad - it should definitely be possible! I just missed that 👍

kael89: @rohan-bes ready to re-review

rohan-bes (approved):

rohan-bes:
Looks really good, awesome work!!

kael89:
@edmofro has mentioned an issue with GitHub Desktop, where the actual error is not displayed correctly :

Screen Shot 2020-10-05 at 5 14 48 PM

It seems that this is an [open GitHub Desktop issue](https://github.com/desktop/desktop/issues/5439). Until we figure out how we deal with this, would you be happy if I remove the pre-push check @edmofro but keep the script in the root `package.json`? That way people can still use the script manually for now, although it probably loses much of its value this way.

Another reason for this is that we may want to also invoke the functionality included in [Lint package.json](https://github.com/beyondessential/tupaia-backlog/issues/1252) in the `validate` script.

Finally, we can leverage some of the refactoring done in this PR.

edmofro:
Yep that sounds good thanks @kael89

kael89:
@edmofro ready now, do you want to test this? I have verified it with both local branches and the CI/CD pipeline build, but of course I am the author so perhaps another dev should take a look.

edmofro:
Yep was able to see the validation output via command line, but github desktop can still push successfully

Nice one @kael89!

PR #1428, opened 2020-10-01
Stacked bar graphs by school type laos

Alastair-L:
### Issue (epic):
https://github.com/beyondessential/tupaia-backlog/issues/1351
### Changes:
Added a new countMeasureValues dataBuilder.

Screenshots on each individual card


kael89:
After a better discussion with @Alastair-L , I believe his idea of using an indicator makes a lot of sense in this use case, because it will allows us to capture a new piece of domain-specific information

packages/database/src/migrations/20200925020630-AddStackedBarGraphsBySchoolTypeLaosSchools-modifies-data.js
```diff
@@ -0,0 +1,215 @@
...
+  chartConfig: {
+    $all: {
+      stackId: 1,
+    },
+  },
+  renderLegendForOneItem: true,
+  presentationOptions: { hideAverage: true },
+});
+
+const createReport = (id, name, dataElement, dataElementToString) => ({
```
Alastair-L, line 172: I like this 'functional' config rather than `{...BASE_X_CONFIG, variable}` over and over. What do you think?

packages/web-config-server/src/apiV1/dataBuilders/generic/count/countMeasureValues.js
```diff
@@ -0,0 +1,64 @@
...
+    const {
+      measureBuilder,
+      convertToPercentage,
+      measureBuilderConfig,
+      valueKey = 'value',
+    } = this.config;
+
+    const { aggregator, dhisApi, query, entity } = this;
+
+    const { data: measureData } = await getMeasureBuilder(measureBuilder)(
```
chris-bes, line 16: Just my personal opinion, my gut feeling is I don't quite prefer having a data builder that calls a measure builder like this 😢 . A reason I can think of is if we have a major refactor on measureBuilder class we might miss this one? 

If you agree, and if this is just to reuse the `getStringFromBinaryData` logic and count for `Multiple`, do you think you can reuse the `countCalculatedValuePerOrgUnit` data builder that I did for https://github.com/beyondessential/tupaia/pull/1429. (It has access to `calculateOperationForAnalytics.js` logic which can combine the string for you.

Alastair-L: Good idea. I don't necessarily mind using measure builders, but I agree that it would be cleaner to use the same dataBuilder for both.

chris-bes (approved):
Looks great to me! Hope it wasn't too much work converting it 😄

Alastair-L:
Totally worth being able to cut a databuilder 👍

Alastair-L:
Using a normal merge for convenience as it is being merged into another feature branch.

PR #1430, opened 2020-10-02
1116 user profile photo upload

tcaiger:
### Changes:

- Add profile image field to profile page form

---

### Screenshots:
Screenshots can be found [here](https://github.com/beyondessential/tupaia-backlog/issues/1116)


packages/admin-panel/public/manifest.json
```diff
@@ -4,7 +4,7 @@
   "icons": [
     {
       "src": "favicon.ico",
-      "sizes": "192x192",
```
tcaiger, line 7: fix existing console warning

packages/admin-panel/src/pages/ProfilePage.js
```diff
@@ -35,41 +48,118 @@ const SuccessMessage = styled(SmallAlert)`
...
+    fileId: null,
+    data: null,
+  },
+};
+
+const initReducer = user => {
+  return { ...initialState, profileImage: { data: user.profileImage, fileId: null } };
+};
+
+function reducer(state, action) {
```
tcaiger, line 74: Using a local reducer keeps the ProfilePage component nice and encapsulated but also allows us to manage the complex(ish) state management outside of the component

packages/admin-panel/src/pages/ProfilePage.js
```diff
@@ -35,41 +48,118 @@ const SuccessMessage = styled(SmallAlert)`
...
+    }
+    case 'fileChange': {
+      return {
+        ...state,
+        profileImage: action.payload,
+        status: STATUS.IDLE,
+      };
+    }
+    case 'deleteProfile': {
+      return { ...state, profileImage: { fileId: '-', data: null }, confirmModalIsOpen: false };
```
tcaiger, line 92: We need to set a file id to something so that the backend will pick up the profileImage field as an updatedField and save the change.

packages/admin-panel/src/pages/ProfilePage.js
edmofro: It is cool! I haven't seen React.useReducer before, thanks for the intro

A suggestion here would be to keep the action/reducer flow contained, and out of the component, by adding an extra layer. Something like `connectProfilePageReducer()` which does
```
const connectProfilePageReducer = () => {
  const [
    state,
    dispatch,
  ] = React.useReducer(reducer, user, initReducer);
  const actions = {
    onUpdateStart: () => dispatch({ type: 'loading' }),
    onUpdateSuccess: () => dispatch({ type: 'success', payload: 'Profile successfully updated.' }),
    onUpdateError: error => dispatch({ type: 'error', payload: error.message }),
    onFileChangeStart: () => dispatch({ type: 'disable' }),
    onFileChangeSuccess: (user, fileName) => dispatch({
      type: 'fileChange',
      payload: {
        fileId: `${user.id}-${fileName}`,
        data: Base64,
      },
    }),
    ... etc.
  };
  return [
    state,
    actions,
  ];
};
```
All of the reducer stuff could live in a neighbouring file rather than the same file as the component. This would mean the component itself would read a bit more fluently, and not risk people getting confused by the action types or whether dispatch means it's going to redux. The downside is it means you have one extra connection to make if you're trying to understand the whole flow.

Over to you!

edmofro (changes requested):
Cool stuff in here!

tcaiger:
I think that is a good point that it is a bit confusing having the 'dispatch' calls in the middle of the component. I tried a couple of different things to refactor it. One good idea I had was to pull the reducer and state out together into a custom hook and then return the state and methods together. However half way through I realised that in this instance `confirm modal` should actually be part of the profile field component and once I pulled it out I felt the reducer was no longer needed. With less state to manage, I was able to go back to the useState api and I think it is much better now. What do you think?

tcaiger:
> I think that is a good point that it is a bit confusing having the 'dispatch' calls in the middle of the component. I tried a couple of different things to refactor it. One good idea I had was to pull the reducer and state out together into a custom hook and then return the state and methods together. However half way through I realised that in this instance `confirm modal` should actually be part of the profile field component and once I pulled it out I felt the reducer was no longer needed. With less state to manage, I was able to go back to the useState api and I think it is much better now. What do you think?

edmofro (approved):
Yep, much easier to follow with that confirm modal pulled out!

tcaiger:
**Testing notes**

Branch `1116-user-profile-photo-upload`

Test cases:
- user can upload photo
- user can remove photo
- confirm modal opens before deleting
- photo displays in the profile menu when you open it after saving

note: the first letter of your name will still show on the profile button even if there is a photo saved.

PR #1431, opened 2020-10-04
Speed up full build and cache

edmofro:
From ~18 minutes down to ~40 seconds for a full run

@kael89 I'm sorry about the number of times you've had to re-review this same section of code. My motivations here was that I realised that the cache invalidation logic for handling each individual case was becoming quite complex, and that if I could get the time for a full run down by enough, we could just re-run the full build whenever something changed, simplifying things greatly.

### DIFF
--- a/packages/database/src/TupaiaDatabase.js
+++ b/packages/database/src/TupaiaDatabase.js
@@ -268,11 +268,15 @@ export class TupaiaDatabase {
   async createMany(recordType, records) {
     // generate ids for any records that don't have them
     const sanitizedRecords = records.map(r => (r.id ? r : { id: this.generateId(), ...r }));
-    await this.query({
-      recordType,
-      queryMethod: QUERY_METHODS.INSERT,
-      queryMethodParameter: sanitizedRecords,
-    });
+    const batchSize = this.maxBindingsPerQuery;
+    for (let i = 0; i < sanitizedRecords.length; i += batchSize) {
+      const batchOfRecords = sanitizedRecords.slice(i, i + batchSize);
+      await this.query({
+        recordType,
+        queryMethod: QUERY_METHODS.INSERT,
+        queryMethodParameter: batchOfRecords,
+      });
+    }
     return sanitizedRecords;
   }
 
--- a/packages/database/src/TupaiaDatabase.js
+++ b/packages/database/src/TupaiaDatabase.js
@@ -11,6 +11,10 @@ import { Multilock } from '@tupaia/utils';
 import { getConnectionConfig } from './getConnectionConfig';
 import { DatabaseChangeChannel } from './DatabaseChangeChannel';
 import { generateId } from './utilities/generateId';
+import {
+  runDatabaseFunctionInBatches,
+  MAX_BINDINGS_PER_QUERY,
+} from './utilities/runDatabaseFunctionInBatches';
 
 const QUERY_METHODS = {
   COUNT: 'count',
@@ -42,9 +46,6 @@ export const JOIN_TYPES = {
 // no math here, just hand-tuned to be as low as possible while
 // keeping all the tests passing
 const HANDLER_DEBOUNCE_DURATION = 250;
-
-// some part of knex or node-pg struggles with too many bindings, so we batch them in some places
-const MAX_BINDINGS_PER_QUERY = 2500; // errors occurred at around 5000 bindings in testing
 
 export class TupaiaDatabase {
   constructor(transactingConnection) {
@@ -268,15 +269,13 @@ export class TupaiaDatabase {
   async createMany(recordType, records) {
     // generate ids for any records that don't have them
     const sanitizedRecords = records.map(r => (r.id ? r : { id: this.generateId(), ...r }));
-    const batchSize = this.maxBindingsPerQuery;
-    for (let i = 0; i < sanitizedRecords.length; i += batchSize) {
-      const batchOfRecords = sanitizedRecords.slice(i, i + batchSize);
-      await this.query({
+    await runDatabaseFunctionInBatches(sanitizedRecords, async batchOfRecords =>
+      this.query({
         recordType,
         queryMethod: QUERY_METHODS.INSERT,
         queryMethodParameter: batchOfRecords,
-      });
-    }
+      }),
+    );
     return sanitizedRecords;
   }
 
@@ -447,15 +446,10 @@ export class TupaiaDatabase {
   }
 
   async executeSqlInBatches(arrayToBeBatched, generateSql) {
-    const records = [];
-    const batchSize = MAX_BINDINGS_PER_QUERY;
-    for (let i = 0; i < arrayToBeBatched.length; i += batchSize) {
-      const batch = arrayToBeBatched.slice(i, i + batchSize);
+    return runDatabaseFunctionInBatches(arrayToBeBatched, async batch => {
       const [sql, substitutions] = generateSql(batch);
-      const batchOfRecords = await this.executeSql(sql, substitutions);
-      records.push(...batchOfRecords);
-    }
-    return records;
+      return this.executeSql(sql, substitutions);
+    });
   }
 }
 
--- a/packages/database/src/cachers/EntityHierarchyCacher.js
+++ b/packages/database/src/cachers/EntityHierarchyCacher.js
@@ -17,10 +17,8 @@ export class EntityHierarchyCacher {
   async buildAndCacheAll() {
     // projects are the root entities of every full tree, so start with them
     const projects = await this.models.project.all();
-    // iterate through projects in serial, as each one is quite resource intensive
-    for (const project of projects) {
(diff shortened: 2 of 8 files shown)


packages/database/src/migrations/20201002193222-AddEntityParentIdIndex-modifies-schema.js
```diff
@@ -0,0 +1,27 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.addIndex('entity', 'entity_parent_id_key', ['parent_id']);
```
edmofro, line 18: This was about half of the speedup!

packages/database/src/tests/cachers/EntityHierarchyCacher.fixtures.js
```diff
@@ -3,19 +3,18 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-// canonical hierarchy is a - k, with each entity having the previous letter as its parent
+// canonical hierarchy splits into two children at each of two generations
+//          a
+//    aa         ab
+// aaa  aab   aba  abb
```
edmofro, line 9: The single path structure was not as suitable for tests to come later

packages/database/src/tests/cachers/EntityHierarchyCacher.test.js
```diff
@@ -17,31 +17,42 @@ describe('EntityHierarchyCacher', () => {
...
+      entity_hierarchy_id: hierarchyId,
+    });
+    expect(
+      relationsForProject.map(r => ({
+        ancestor_id: r.ancestor_id,
+        descendant_id: r.descendant_id,
+        generational_distance: r.generational_distance,
+      })),
+    ).to.deep.equalInAnyOrder(relations);
+  };
```
edmofro, line 37: Getting a bit ahead of this PR, but this helps with tests to come later

packages/database/src/migrations/20201002193222-AddEntityParentIdIndex-modifies-schema.js
kael89: Nice!

packages/database/src/cachers/EntityHierarchyCacher.js
```diff
@@ -40,105 +34,115 @@ export class EntityHierarchyCacher {
...
    * Stores the generation of ancestor/descendant info in the database
    * @param {string} hierarchyId
-   * @param {Entity[]} entityIds
-   * @param {Entity[]} ancestorIds In order of generational distance, with immediate parent at index 0
+   * @param {Entity[]} childIdsToAncestorIds  Ids of the child entities as keys, with the ids of their
+   *                                          ancestors in order of generational distance, with immediate
+   *                                          parent at index 0
    */
-  async cacheGeneration(hierarchyId, entityIds, ancestorIds) {
+  async cacheGeneration(hierarchyId, childIdsToAncestorIds) {
```
kael89, line 134: Based on a previous comment:

```suggestion
  async cacheGeneration(hierarchyId, childIdToAncestorIds) {
```

packages/database/src/cachers/EntityHierarchyCacher.js
```diff
@@ -40,105 +34,115 @@ export class EntityHierarchyCacher {
...
+    return numberChildrenCached === childCount;
+  }
+
+  async fetchChildIdsToAncestorIds(hierarchyId, parentIdsToAncestorIds, useEntityRelationLinks) {
+    const parentIds = Object.keys(parentIdsToAncestorIds);
+    const relations = useEntityRelationLinks
+      ? await this.getRelationsViaEntityRelation(hierarchyId, parentIds)
+      : await this.getRelationsCanonically(parentIds);
+    const childIdToParentId = reduceToDictionary(relations, 'child_id', 'parent_id');
+    const childIdsToAncestorIds = Object.fromEntries(
```
kael89, line 86: I don't know if you agree with this change? I believe it's a good indication of the `childId => ancestorIds` relationship in this object.

By the way, there should really be a universal naming guide for variables, similar to code style guides 🤓 Lots of the patterns and structures are often repeated!

```suggestion
    const childIdToAncestorIds = Object.fromEntries(
```

kael89 (changes requested):
That's an amazing performance optimisation!! I am glad we can keep the code simple and quick 🎉

A few minor changes requested, also a question: If I'm not mistaken, we currently build the cache once when the app runs. I guess we'll also add cache invalidation when an entity/entity_relation is updated, as is implied in the PR description?

packages/database/src/TupaiaDatabase.js
```diff
@@ -268,11 +268,15 @@ export class TupaiaDatabase {
...
     // generate ids for any records that don't have them
     const sanitizedRecords = records.map(r => (r.id ? r : { id: this.generateId(), ...r }));
-    await this.query({
-      recordType,
-      queryMethod: QUERY_METHODS.INSERT,
-      queryMethodParameter: sanitizedRecords,
-    });
+    const batchSize = this.maxBindingsPerQuery;
+    for (let i = 0; i < sanitizedRecords.length; i += batchSize) {
+      const batchOfRecords = sanitizedRecords.slice(i, i + batchSize);
```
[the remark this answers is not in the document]
edmofro: True! I think you're remembering `executeSqlInBatches`, which does that for raw sql. It would make sense to extract the common util. Cheers for the suggestion :+1:

packages/database/src/cachers/EntityHierarchyCacher.js
```diff
@@ -40,105 +34,115 @@ export class EntityHierarchyCacher {
...
     }
 
-    // keep recursing through the hierarchy
-    await this.fetchAndCacheDescendants(hierarchyId, childIdsToAncestorIds);
+    // if there is another generation, keep recursing through the hierarchy
+    await this.fetchAndCacheDescendants(
+      hierarchyId,
+      childIdsToAncestorIds,
+      entityRelationChildCount === 0,
+    );
```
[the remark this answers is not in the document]
edmofro: Good catch!

packages/database/src/cachers/EntityHierarchyCacher.js
```diff
@@ -40,105 +34,115 @@ export class EntityHierarchyCacher {
...
-      parent_id: entityId,
+    return this.models.entityRelation.find({
+      parent_id: parentIds,
       entity_hierarchy_id: hierarchyId,
     });
-    return hierarchyLinks.map(l => l.child_id);
+  }
+
+  async getRelationsCanonically(entityId) {
+    const canonicalTypes = Object.values(ORG_UNIT_ENTITY_TYPES);
```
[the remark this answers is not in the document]
edmofro: Good point - I've taken another approach to make it DRY

packages/database/src/cachers/EntityHierarchyCacher.js
edmofro: Yeah I tend to agree with you on the singular for childId, it matches the naming conventions we've used elsewhere. 

Perhaps a good dev chat item - naming conventions! I've noticed some of our common tupaia conventions may not be natural to devs coming from other companies. A good example is that others tend to use 
```
async function getXByY(y) { 
  const x = await db.fetchX();
  return x.filter(x => x.y === y); 
}
```
whereas we would expect the same name to do something like
```
async function getXByY() { 
  const x = await db.fetchX();
  return groupBy(x, y); 
}
```

packages/database/src/cachers/EntityHierarchyCacher.js
```diff
@@ -40,105 +34,115 @@ export class EntityHierarchyCacher {
...
-    const hierarchyLinks = await this.models.entityRelation.find({
-      parent_id: entityId,
+    return this.models.entityRelation.find({
+      parent_id: parentIds,
       entity_hierarchy_id: hierarchyId,
     });
-    return hierarchyLinks.map(l => l.child_id);
+  }
+
+  async getRelationsCanonically(entityId) {
```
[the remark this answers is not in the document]
edmofro: Kind of just a lucky coincidence that this worked with me thinking it was a single id! Thanks database magic

packages/database/src/cachers/EntityHierarchyCacher.js
kael89: Agree! Sounds like a valuable documentation, it's not something that changes every day

packages/database/src/TupaiaDatabase.js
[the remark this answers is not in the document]
edmofro: This made a few places much cleaner!

kael89 (approved):
Looks good!

PR #1432, opened 2020-10-04
1271 data broker convert mocha to jest

billli0:
### Issue #:
[1271- Use a single test framework : **data-broker module only**](https://github.com/beyondessential/tupaia-backlog/issues/1271)

Current issues: 

- [x] Jest did not exit one second after the test run has completed. (Resolved)
- [x] A few tests couldn't resolve relating to stub. (Resolved)


packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
```diff
@@ -3,73 +3,52 @@
...
     it('throws if the data sources belong to multiple services', async () => {
       const data = [{ value: 1 }, { value: 2 }];
       const dataBroker = new DataBroker();
       return expect(
-        dataBroker.push({ code: ['TEST_01', 'OTHER_01'], type: 'dataElement' }),
-        data,
-      ).to.be.rejectedWith('Cannot push data belonging to different services');
+        dataBroker.push({ code: ['TEST_01', 'OTHER_01'], type: 'dataElement' }, data),
+      ).toBeRejectedWith('Cannot push data belonging to different services');
     });
```
billli0, line 36: I think the `data` should be one of the params in `.push` instead of `expect`

packages/data-broker/src/__tests__/DataBroker/DataBroker.stubs.js
```diff
@@ -44,9 +41,10 @@ export const createServiceStub = serviceData => {
       default:
         throw new Error(`Invalid data source type: ${type}`);
     }
-  });
-
-  return sinon.createStubInstance(Service, { pull });
+  };
+  const { Service } = jest.createMockFromModule('../../services/Service');
+  Service.prototype.pull = pull;
+  return new Service();
```
billli0, line 47: I am not sure why it doesn't work using mockImplementation. I think it is because `Service` is an abstract class.

kael89: I don't think this is related since there are no actual abstract classes with the in JavaScript, only in TypeScript. The `abstract` JsDoc is for developer reference only, it indicates that this class should not be instantiated directly - however, this is still possible.

kael89 (commented):
Everything looks great!! Very clean job. I have just implemented a few minor changes in https://github.com/beyondessential/tupaia/pull/1485

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
kael89: Oh yeah great catch!! 👀

kael89 (approved):
Feel free to merge after the tests pass!

PR #1433, opened 2020-10-05
Auth test tweaks

kael89:
* Fix the hanging connection issue
* Fix `toBeRejectedWith` to work with async class methods
* Minor tweaks


packages/utils/src/testUtilities/addCustomJestMatchers.js
```diff
@@ -25,13 +26,11 @@ class JestMatcherFactory {
     this.extendApi = extendApi;
   }
 
-  isAsync = config => config.matcher.constructor.name === 'AsyncFunction';
```
kael89, line 28: This method was not working correctly for `async` class methods...

It seems there is not a 100% correct way to say if a method is async in code, so I used an `isAsync` option instead

billli0: Nice approach! It happened to me a few times before.. I can feel you.

billli0 (approved):
Thanks Kostas for checking in! There is a lot of work correcting the test matcher. Nice job!

PR #1435, opened 2020-10-05
allow custom sort_order of options

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1269

Specifically requesting @edmofro 's review because I don't think this is actually necessary
- I didn't realise this until I was doing some final tests locally, but it's possible to edit the `sort_order` of options in an existing option set just by moving the rows in the import file around
- When this file is reimported it just updates the `sort_order` based on the ordering of the rows in Excel, so effectively the way to set a custom `sort_order` is just by how the rows are positioned in Excel
- Since the only use case I can think of for setting a custom `sort_order` was updating existing option sets, this PR kinda seems useless and unnecessary

So basically, the purpose of this review is to decide whether or not to discard this PR (personally I think it's pointless and I just made a mistake in creating an issue for this in the first place)
- I guess I can write some documentation around option sets so nobody gets confused about this in future


edmofro (approved):
Na this is great! It might be a little redundant, but given that you thought it should be possible, other people might think the same, and supporting the functionality means we don't have to stress about people not finding the documentation and getting confused

Nice work @avaek! All the code looks great too 💯

PR #1436, opened 2020-10-05
1374 fix default date is wrong in map overlays

IgorNadj:
### Changes:
- Measures use correct default date
- Prevent set dates to the same dates (this prevents multiple requests)

### DIFF
--- a/packages/web-frontend/src/components/DateRangePicker/DateRangePicker.js
+++ b/packages/web-frontend/src/components/DateRangePicker/DateRangePicker.js
@@ -132,7 +132,10 @@ export const DateRangePicker = ({
   );
 
   useEffect(() => {
-    onSetDates(roundedCurrentStartDate, roundedCurrentEndDate);
+    // Prevent set dates to the same dates
+    if (!(initialStartDate && initialEndDate)) {
+      onSetDates(roundedCurrentStartDate, roundedCurrentEndDate);
+    }
   }, []);
 
   // Number of periods to move may be negative if changing to the previous period
--- a/packages/web-frontend/src/containers/MeasureBar/Control.js
+++ b/packages/web-frontend/src/containers/MeasureBar/Control.js
@@ -109,6 +109,7 @@ const ExpandedContent = styled.div`
 export const Control = ({
   emptyMessage,
   selectedMeasure,
+  defaultDates,
   isMeasureLoading,
   onUpdateMeasurePeriod,
   children,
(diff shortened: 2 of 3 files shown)


packages/web-frontend/src/components/DateRangePicker/DateRangePicker.js
```diff
@@ -132,7 +132,10 @@ export const DateRangePicker = ({
   );
 
   useEffect(() => {
-    onSetDates(roundedCurrentStartDate, roundedCurrentEndDate);
+    // Prevent set dates to the same dates
```
tcaiger, line 135: Yes good call adding this check to get rid of the extra request!

packages/web-frontend/src/containers/MeasureBar/Control.js
```diff
@@ -183,6 +188,10 @@ Control.propTypes = {
     startDate: PropTypes.shape({}),
     endDate: PropTypes.shape({}),
   }),
+  defaultDates: PropTypes.shape({
+    startDate: PropTypes.string,
+    endDate: PropTypes.string,
+  }).isRequired,
```
tcaiger, line 194: is default date required?

packages/web-frontend/src/containers/MeasureBar/Control.js
```diff
@@ -129,6 +130,10 @@ export const Control = ({
     onUpdateMeasurePeriod(moment(startDate).startOf(period), moment(endDate).endOf(period));
   };
 
+  let { startDate, endDate } = selectedMeasure;
+  if (!startDate && defaultDates.startDate) startDate = defaultDates.startDate;
```
tcaiger, line 134: In the prop types it says that default dates is required. If so then maybe this check is not necessary? However if it is possible for there to be one or the other of startDate and endDate then I can see why the check is needed.

tcaiger (approved):
Nice work Igor. I see that you split out the refactoring into another task which is fair enough given the complexity of this. I left one small question but I am happy to approve this anyway. My only other comment is that the DateRangePicker is used on Dashbaord Reports as well as map overlays so I would recommend adding a line to the testing notes for some regression testing on them as well.

packages/web-frontend/src/containers/MeasureBar/Control.js
IgorNadj: Yeah for map overlays we are saying it is required so that we always have initial dates, and so DateRangePicker is guaranteed to use the correct default dates.

IgorNadj: Added a comment to clarify

packages/web-frontend/src/containers/MeasureBar/Control.js
IgorNadj: Good point, removed

IgorNadj:
Accidentally merged before testing completed. Reverted.

PR #1437, opened 2020-10-05
Refactor test data

kael89:
The `Demo.initialise` functionality is clever, but it has the drawback of depending on state + things happening in a specific order. Demo must first be initialised (as it happens in `buildAccessPolicy.test.js`), only then the consuming test files (eg `testAsPublicUser.js`) can use the contents of the initialised Demo object. This can be a bit confusing for someone who reads the tests for the first time and looks eg at `testAsPublicUser.js`, wondering where and when Demo gets initialised.

Although it's an acceptable pattern, I decided to change it to a stateless pattern as  we do in other similar cases across the codebase


packages/auth/src/__tests__/buildAccessPolicy/helpers.js
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import {
+  findOrCreateDummyRecord,
+  getTestDatabase,
+  ModelRegistry,
+  upsertDummyRecord,
+} from '@tupaia/database';
+
+export const setUp = async () => {
```
kael89, line 13: I could also name this file `setup.js` as you have done, but we generally use the convention of a `helpers.js` so I ended up using this.

Other conventions you may see across the codebase:

* `${testName}.fixtures.js` for test fixtures (=> test data)
* `${testName}.stubs.js` for stubs

billli0: > I could also name this file `setup.js` as you have done, but we generally use the convention of a `helpers.js` file for similar functionality, so I ended up using this.
> 
> Other conventions you may see across the codebase:
> 
> * `${testName}.fixtures.js` for test fixtures (=> test data)
> * `${testName}.stubs.js` for stubs

Roger that. So nice you list it all! Thanks Kostas!

kael89: Perhaps it would be good to document this kind of information... Although the problem with documentation is that it often becomes obsolete even before you finish writing it!

billli0:
> The `Demo.initialise` functionality is clever, but it has the drawback of depending on state + things happening in a specific order. Demo must first be initialised (as it happens in `buildAccessPolicy.test.js`), only then the consuming test files (eg `testAsPublicUser.js`) can use the contents of the initialised Demo object. This can be a bit confusing for someone who reads the tests for the first time and looks eg at `testAsPublicUser.js`, wondering where and when Demo gets initialised.
> 
> Although it's an acceptable pattern, I decided to change it to a stateless pattern as we do in other similar cases across the codebase

billli0 (commented):
Yes you are right. Now the tests are more readable. I create `Demo.initialise` for reusing code. Obviously now 'setup()' > 'initialise()'!

billli0 (approved):
Yes you are right. Now the tests are more readable. I create Demo.initialise for reusing code. Obviously now 'setup()' > 'initialise()'!

PR #1445, opened 2020-10-06
Update e2e test documentation

kael89:


packages/web-frontend/cypress/README.md
```diff
@@ -1,22 +1,57 @@
...
+1. If you haven't already, follow the instructions in [Tupaia Monorepo setup](https://docs.beyondessential.com.au/books/software-development/page/tupaia-monorepo-setup) to setup this project.
+
+   ⚠️: In **Step 2. Install node dependencies** you need to run the commands under a Windows terminal, see the note in that section.
+
+2. Add a `cypress.env.json` file under the root folder of `web-frontend` (see [cypress.env.example.json](https://github.com/beyondessential/tupaia/tree/dev/packages/web-frontend/cypress.env.example.json) for a template)
+3. Use a **[LOCAL]** `.env` file for `web-frontend`, so that it points to a local `web-config-server`
+4. The tests depend on `.json` configuration files that must be placed under `cypress/config`. To generate the default config, run `yarn cypress:generate-config`. You can also use custom config by manually populating those files, see the `*.example.jsonc` files for more information.
```
sepidehdd, line 13: can you please also add under which directory we need to run this command

kael89: Good idea! I updated this part to use `yarn workspace ...` instead, which can be run from any location inside the repository.

Let me know if you think this should be more obvious!

sepidehdd (approved):

PR #1449, opened 2020-10-06
1208 dashboard loading and error state

tcaiger:
### Changes:

- Add Alert component
- Add loading state to date picker
- Add info and error alert to Enlarged Dialog component
- Add info and error alert to dashboards

---

### Screenshots:
Screenshots here: https://github.com/beyondessential/tupaia-backlog/issues/1208


packages/web-frontend/config/webpack.config.dev.js
```diff
@@ -87,6 +87,7 @@ module.exports = {
       // Support React Native Web
       // https://www.smashingmagazine.com/2016/08/a-glimpse-into-the-future-with-react-native-for-web/
       'react-native': 'react-native-web',
+      '@material-ui/styles': path.resolve(__dirname, '..', 'node_modules', '@material-ui/styles'),
```
tcaiger, line 90: This prevents the package being loaded more than once

packages/web-frontend/src/actions.js
```diff
@@ -686,12 +686,12 @@ export function fetchDashboardSuccess(dashboardConfig) {
 /**
  * Changes state to communicate error to user appropriately.
  *
- * @param {object} error  response from saga on failed fetch
+ * @param {string} errorMessage  response from saga on failed fetch
```
tcaiger, line 689: Fix existing bug. The error that gets passed from the saga is a string

packages/web-frontend/src/sagas.js
```diff
@@ -606,14 +606,16 @@ function* fetchViewData(parameters, errorHandler) {
...
-    } else {
-      console.log(`Failed to handle error: ${error.message}`);
+    let errorMessage = error.message;
+
+    if (error.response) {
+      const json = yield error.response.json();
+      errorMessage = json.error;
     }
+
+    yield put(errorHandler(errorMessage, infoViewKey));
```
tcaiger, line 618: Always use the designated error handler for the saga. Otherwise another error handler from the saga could handle the error.

packages/web-frontend/src/components/Alert.js
```diff
@@ -0,0 +1,93 @@
...
+  severity: PropTypes.string,
+  variant: PropTypes.string,
+  iconMapping: PropTypes.object,
+};
+
+Alert.defaultProps = {
+  severity: 'success',
+  variant: 'standard',
+  iconMapping: {
+    success: <CheckCircle fontSize="inherit" />,
```
IgorNadj, line 89: How come this is configurable? Seems ok to leave it up to the component to decide this.

packages/web-frontend/src/components/Alert.js
```diff
@@ -0,0 +1,93 @@
...
+`;
+
+export const AlertLink = styled.a`
+  font-size: 0.875rem;
+  text-decoration: underline;
+  color: inherit;
+  font-weight: 500;
+`;
+
+export const Alert = ({ variant, severity, iconMapping, ...props }) => (
```
IgorNadj, line 75: Storybook story for this would be great

packages/web-frontend/config/webpack.config.dev.js
IgorNadj: Nice, add as a comment

packages/web-frontend/src/sagas.js
IgorNadj: I'm not a fan of ignoring errorFunctions here. errorFunctions should be checked first and handled globally. E.g. if the user is not logged in, that should be a single UI element saying you're logged out, please log in again, instead of each component on the screen having an error message saying "please log in". In that case, a simple "something went wrong" message on each component, with maybe a retry button would be fine.

It's only if the global error function is not hit that we can assume this is a user error, e.g. wrong dates, and the container that made the request can display the specific error.

```
if (error.errorFunction) {	   
  // global error (e.g. not logged in), handle globally, plus display simple error message in each component
  yield put(error.errorFunction(error, infoViewKey));	
  if (errorHandler) {
    errorHandler('Failed to load data');
  }
} else {
  // user error e.g. bad dates
  if (errorHandler) {
    const json = ...
    ...
    errorHandler(errorMessage, infoViewKey);
  }
}

```

With 500 errors:
Sometimes the server falls over and gives a 500. Usually users cannot do anything about this e.g. server is offline, or there is a bug in the code, apart from try again. 

Generally speaking, an API that returns 500 should not return any information about the error. The server fell over, and it shouldn't have. Instead it should be logged and devs alerted and go fix the error. 

It seems from the screenshot before ("Internal server error: Start period must be equal to or earlier than the end period") that some of our APIs are returning 500 with user errors, this is a problem... these should be 400 Bad Request. If that's the case then that would be a separate ticket to change these.

IgorNadj (changes requested):

packages/web-frontend/src/components/Alert.js
tcaiger: This is to set our own defaults over the top of the material-ui ones to match the designs. The component will also be able to customise it when it's being used if needed.

packages/web-frontend/src/components/Alert.js
tcaiger: Yes good call. I have added it.

IgorNadj:
LGTM! Nice one

IgorNadj (approved):

packages/web-frontend/src/sagas.js
```diff
@@ -606,11 +606,17 @@ function* fetchViewData(parameters, errorHandler) {
...
+    let errorMessage = error.message;
+
+    if (error.response) {
+      const json = yield error.response.json();
+      errorMessage = json.error;
+    }
+
     if (error.errorFunction) {
-      yield put(error.errorFunction(error, infoViewKey));
+      yield put(error.errorFunction(errorMessage, infoViewKey));
```
IgorNadj, line 619: Ah this should probably stay as errorFunction(error, infoViewKey) no?

And you removed errorHandler()? I have approved anyway, assuming it works as you want it to

tcaiger: yep it is working as we discussed. On line 609 the errorHandler gets passed into the request call and then the request handler returns the correct errorHandler back in the error response. If there is a server error, then a generic errorHandler gets passed back otherwise the specified error handler gets passed. I did a bit of digging and testing and realised that this behaviour is already set up in the saga request handler.

tcaiger: > Ah this should probably stay as errorFunction(error, infoViewKey) no?
> 
> And you removed errorHandler()? I have approved anyway, assuming it works as you want it to

No here I dod want errorMessage as this will give me the full error message from the server if there is a user error message present. Thanks for the review, I'm much happier with the code and my understanding than I was before :)

PR #1451, opened 2020-10-07
1019: Laos Schools: MediTrak Functionality: Add 'Select School Type'

chris-bes:
### Changes:

- Added select school type to meditrak app.


packages/meditrak-app/app/assessment/Question.js
```diff
@@ -39,6 +40,7 @@ const QUESTION_TYPES = {
   Number: NumberQuestion,
   Photo: PhotoQuestion,
   Radio: RadioQuestion,
+  Selection: SelectionQuestion,
```
chris-bes, line 43: I don't think there we support any question type like the mock up (https://app.zenhub.com/files/232001970/fcaeb48f-54bc-4a24-ac37-2089a0829ae9/download). The closest we have is Radio question. So I created a new type of question above

packages/meditrak-app/app/database/types/Entity.js
```diff
@@ -42,14 +44,17 @@ Entity.schema = {
     parent: { type: 'Entity', optional: true },
     code: { type: 'string', optional: true },
     type: { type: 'string', default: 'Entity not properly synchronised' },
+    attributesType: { type: 'string', optional: true },
```
chris-bes, line 47: I'm aware that it's better to sync the whole `attributes` column. But as I understand it requires creating a new schema for it (in order to perform filtering) as well as link it to Entity in order to update whenever entity is sync. 

Maybe, if we want to go with syncing the whole attributes column, what I can do is in Attributes schema, have an entity_id property that points to the Entity it is attached to so that when syncing we can find the corresponding attributes and update?

packages/meditrak-app/app/assessment/Question.js
chris-bes: As discussed with Edwin, we will use the existing Radio question.

packages/meditrak-app/app/entityMenu/actions.js
```diff
@@ -34,6 +34,20 @@ const getEntityFilters = (state, database, questionId) => {
...
 };
 
+const getEntityAttributeFilters = (state, questionId) => {
+  const question = getQuestion(state, questionId);
+  const { attributes } = question.config.entity;
+
+  return entity =>
+    attributes
+      ? Object.entries(attributes).every(([key, config]) => {
+          const attributeValue = getAnswerForQuestion(state, config.questionId);
```
edmofro, line 44: There's no requirement for how to deal with it, but I'm interested in what happens when there is no answer to the specified question? Do we get all entities or none?

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processEntityConfig.js
```diff
@@ -29,14 +31,24 @@ export const processEntityConfig = async (models, config) => {
   return replaceQuestionCodesWithIds(models, processedConfig);
 };
 
+const containJsonField = fieldKey =>
+  ENTITY_CREATION_JSON_FIELD_LIST.find(field => fieldKey.startsWith(field));
```
edmofro, line 35:
```suggestion
// JSON subfields will be in the format field.subfield, e.g. 'attributes.type'
const isJsonSubfield = fieldKey =>
  ENTITY_CREATION_JSON_FIELD_LIST.find(field => fieldKey.startsWith(field));
```

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processEntityConfig.js
```diff
@@ -29,14 +31,24 @@ export const processEntityConfig = async (models, config) => {
...
   const resultConfig = {};
 
   Object.keys(config).forEach(fieldKey => {
     const resultKey = ENTITY_CREATION_FIELD_TRANSLATION[fieldKey];
+    const questionCode = config[fieldKey];
+
     if (resultKey) {
-      const questionCode = config[fieldKey];
       resultConfig[resultKey] = questionCode;
+    } else if (containJsonField(fieldKey)) {
```
edmofro, line 46:
```suggestion
    } else if (isJsonSubfield(fieldKey)) {
```

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processEntityConfig.js
```diff
@@ -29,14 +31,24 @@ export const processEntityConfig = async (models, config) => {
...
+      const [jsonKey, jsonFieldKey] = splitStringOn(fieldKey, '.');
+      if (!resultConfig[jsonKey]) {
+        resultConfig[jsonKey] = {};
+      }
+      resultConfig[jsonKey][jsonFieldKey] = questionCode;
```
edmofro, line 51:
```suggestion
      const [primaryFieldKey, subfieldKey] = splitStringOn(fieldKey, '.');
      if (!resultConfig[primaryFieldKey]) {
        resultConfig[primaryFieldKey] = {};
      }
      resultConfig[primaryFieldKey][subfieldKey] = questionCode;
```

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processEntityConfig.js
```diff
@@ -50,6 +62,15 @@ const replaceQuestionCodesWithIds = async (models, config) => {
...
       newValue = { questionId };
+    } else if (ENTITY_CREATION_JSON_FIELD_LIST.includes(fieldKey)) {
+      for (let i = 0; i < Object.entries(value).length; i++) {
+        const [jsonFieldKey, jsonFieldValue] = Object.entries(value)[i];
+        const { id: questionId } = await models.question.findOne({ code: jsonFieldValue });
+        newValue = {
+          ...newValue,
+          [jsonFieldKey]: { questionId },
+        };
+      }
```
edmofro, line 73:
```suggestion
    } else if (ENTITY_CREATION_JSON_FIELD_LIST.includes(fieldKey)) {
      for (let i = 0; i < Object.entries(value).length; i++) {
        const [subfieldKey, questionCode] = Object.entries(value)[i];
        const { id: questionId } = await models.question.findOne({ code: questionCode });
        newValue = {
          ...newValue,
          [subfieldKey]: { questionId },
        };
      }
```

edmofro (approved):
The suggestions aren't necessary, just me working through what was going on by playing with the code. Looks good, I really like the power this gives us - nice work @chris-bes!

packages/meditrak-app/app/entityMenu/actions.js
chris-bes: That's a good point. In my opinion the right behaviour should be getting all entities as it is equivalent to no filter is applied? I can update it

edmofro (approved):

PR #1452, opened 2020-10-07
Add permissions checking to user entity permissions DELETE endpoint

EMcQ-BES:
### Changes:

- Added Generic DeleteHandler
- Added DeleteUserEntityPermissionsHandler
- Added some extra error messages

### DIFF
--- a/packages/meditrak-server/src/app/addRoutesToApp.js
+++ b/packages/meditrak-server/src/app/addRoutesToApp.js
@@ -179,6 +179,7 @@ export function addRoutesToApp(app) {
    * DELETE routes
    */
   app.delete('(/v[0-9]+)/userEntityPermissions/:recordId', deleteUserEntityPermissions);
+  // TODO: Remove the generic handlers once all DELETE endpoints have specific handlers
   app.delete('(/v[0-9]+)/:parentResource/:parentRecordId/:resource/:recordId', deleteRecord);
   app.delete('(/v[0-9]+)/:resource/:recordId', deleteRecord);
 
--- a/packages/meditrak-server/src/app/addRoutesToApp.js
+++ b/packages/meditrak-server/src/app/addRoutesToApp.js
@@ -18,6 +18,7 @@ const {
   authenticate,
   countChanges,
   deleteRecord,
+  deleteUserEntityPermissions,
   editRecord,
   editAccessRequests,
   editUserAccounts,
@@ -177,6 +178,7 @@ export function addRoutesToApp(app) {
   /**
    * DELETE routes
    */
+  app.delete('(/v[0-9]+)/userEntityPermissions/:recordId', deleteUserEntityPermissions);
   app.delete('(/v[0-9]+)/:parentResource/:parentRecordId/:resource/:recordId', deleteRecord);
   app.delete('(/v[0-9]+)/:resource/:recordId', deleteRecord);
 
--- a/packages/meditrak-server/src/routes/CRUDHandler/CRUDHandler.js
+++ b/packages/meditrak-server/src/routes/CRUDHandler/CRUDHandler.js
@@ -3,6 +3,7 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
+import { singularise } from '@tupaia/utils';
 import { RouteHandler } from '../RouteHandler';
 import { resourceToRecordType } from '../../utilities';
 import { extractResourceFromEndpoint } from './helpers';
@@ -21,5 +22,6 @@ export class CRUDHandler extends RouteHandler {
     this.parentResource = params.parentResource;
     this.parentRecordType = resourceToRecordType(this.parentResource);
     this.parentRecordId = params.parentRecordId;
+    this.model = this.models[singularise(this.resource)];
   }
 }
--- a/packages/meditrak-server/src/routes/CRUDHandler/CRUDHandler.js
+++ b/packages/meditrak-server/src/routes/CRUDHandler/CRUDHandler.js
@@ -22,6 +22,6 @@ export class CRUDHandler extends RouteHandler {
     this.parentResource = params.parentResource;
     this.parentRecordType = resourceToRecordType(this.parentResource);
     this.parentRecordId = params.parentRecordId;
-    this.model = this.models[singularise(this.resource)];
+    this.resourceModel = this.models[singularise(this.resource)];
   }
 }
--- a/packages/meditrak-server/src/routes/DeleteHandler/DeleteHandler.js
+++ b/packages/meditrak-server/src/routes/DeleteHandler/DeleteHandler.js
@@ -7,8 +7,6 @@ import {
   constructRecordExistsWithId,
   DatabaseError,
   respond,
-  singularise,
-  ValidationError,
 } from '@tupaia/utils';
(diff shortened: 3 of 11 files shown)


packages/meditrak-server/src/app/addRoutesToApp.js
```diff
@@ -177,6 +178,7 @@ export function addRoutesToApp(app) {
   /**
    * DELETE routes
    */
+  app.delete('(/v[0-9]+)/userEntityPermissions/:recordId', deleteUserEntityPermissions);
   app.delete('(/v[0-9]+)/:parentResource/:parentRecordId/:resource/:recordId', deleteRecord);
   app.delete('(/v[0-9]+)/:resource/:recordId', deleteRecord);
```
edmofro, line 183:
```suggestion
  // todo: delete the generic delete handlers once all have been specifically listed
  app.delete('(/v[0-9]+)/:parentResource/:parentRecordId/:resource/:recordId', deleteRecord);
  app.delete('(/v[0-9]+)/:resource/:recordId', deleteRecord);
```

packages/meditrak-server/src/routes/accessRequests/assertAccessRequestPermissions.js
```diff
@@ -8,6 +8,10 @@ import { getAdminPanelAllowedEntityIds } from '../utilities';
 
 export const assertAccessRequestPermissions = async (accessPolicy, models, accessRequestId) => {
   const accessRequest = await models.accessRequest.findById(accessRequestId);
+  if (accessRequest === null) {
```
edmofro, line 11:
```suggestion
  if (!accessRequest) {
```
So long as we don't expect that another falsey value will throw us off (like, we actually want `0` to make it through, for example), just using a `!` is more standard (I'm also wanting to avoid a future regression risk if we decided `findById` should return `undefined` rather than `null` if none was found)

packages/meditrak-server/src/routes/DeleteHandler/DeleteHandler.js
```diff
@@ -0,0 +1,52 @@
...
+import { CRUDHandler } from '../CRUDHandler';
+
+/**
+ * Responds to DELETE requests for a resource
+ */
+
+export class DeleteHandler extends CRUDHandler {
+  constructor(req, res) {
+    super(req, res);
+    this.model = this.models[singularise(this.resource)];
```
edmofro, line 22: Might be useful in other CRUDHandlers?

packages/meditrak-server/src/routes/DeleteHandler/DeleteHandler.js
```diff
@@ -0,0 +1,52 @@
...
+  async handleRequest() {
+    await this.validate();
+    await this.deleteRecord();
+    respond(this.res, { message: `Successfully deleted ${this.resource}` });
+  }
+
+  async validate() {
+    if (!this.model.isDeletableViaApi) {
+      throw new ValidationError(`${this.resource} is not a valid DELETE endpoint`);
+    }
```
edmofro, line 34: We can delete this, and get rid of `isDeletableViaApi` now that we're listing them explicitly. It's the weird delete equivalent of `EDITABLE_RECORD_TYPES` in `editRecord.js`.

If deleting `isDeletableViaApi` means there is nothing of substance left in the meditrak-server version of the model, and it was extending from a model in `@tupaia/database`, we can also delete the model entirely from meditrak-server.

packages/meditrak-server/src/routes/userAccounts/assertUserAccountPermissions.js
```diff
@@ -10,6 +10,10 @@ const { RAW } = QUERY_CONJUNCTIONS;
 
 export const assertUserAccountPermissions = async (accessPolicy, models, userAccountId) => {
   const userAccount = await models.user.findById(userAccountId);
+  if (userAccount === null) {
```
edmofro, line 13:
```suggestion
  if (!userAccount) {
```

packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
```diff
@@ -12,6 +12,10 @@ export const assertUserEntityPermissionPermissions = async (
   userEntityPermissionId,
 ) => {
   const userEntityPermission = await models.userEntityPermission.findById(userEntityPermissionId);
+  if (userEntityPermission === null) {
```
edmofro, line 15:
```suggestion
  if (!userEntityPermission) {
```

packages/meditrak-server/src/routes/userEntityPermissions/DeleteUserEntityPermissions.js
```diff
@@ -0,0 +1,24 @@
...
+export class DeleteUserEntityPermissions extends DeleteHandler {
+  async assertUserHasAccess() {
+    // Check Permissions
+    const userEntityPermissionChecker = accessPolicy =>
+      assertUserEntityPermissionPermissions(accessPolicy, this.models, this.recordId);
+    await this.assertPermissions(
+      assertAnyPermissions([assertBESAdminAccess, userEntityPermissionChecker]),
+    );
+  }
+}
```
edmofro, line 24: Beautiful, so simple!

edmofro (changes requested):
This makes me feel good about the overall design we've got going here - nice work!

packages/meditrak-server/src/routes/DeleteHandler/DeleteHandler.js
EMcQ-BES: Moved it up a layer into the CRUDHandler

packages/meditrak-server/src/routes/DeleteHandler/DeleteHandler.js
EMcQ-BES: Just something to note - Removing this check leaves our `validate` function with just a check for whether the entity exists. I think realistically most of the endpoints will also check that condition in the permissions gate, but worth keeping in the generic layer since conceptually that should be a generic test and "all delete handlers check this during the permissions gate" isn't necessarily always a valid assumption in our design.

edmofro: Good note, and agreed :+1:

packages/meditrak-server/src/routes/CRUDHandler/CRUDHandler.js
```diff
@@ -21,5 +22,6 @@ export class CRUDHandler extends RouteHandler {
     this.parentResource = params.parentResource;
     this.parentRecordType = resourceToRecordType(this.parentResource);
     this.parentRecordId = params.parentRecordId;
+    this.model = this.models[singularise(this.resource)];
```
edmofro, line 25:
```suggestion
    this.resourceModel = this.models[singularise(this.resource)];
```
What do you reckon? Seeing it in the context of the fields above, this seems to be less ambiguous

edmofro (approved):
One more comment but approved either way :-)

packages/meditrak-server/src/routes/CRUDHandler/CRUDHandler.js
EMcQ-BES: Good call, having `models` and `model` at the same level was probably asking for mixups.

PR #1453, opened 2020-10-07
1114 psss profile picture and avatar colors

tcaiger:
### Changes:

- Add new avatar component with a derived colour value based on user first letter
- Replace material ui avatar with new component through out ui components

---

### Screenshots:
https://github.com/beyondessential/tupaia-backlog/issues/1114


packages/ui-components/src/components/ProfileButton.js
```diff
@@ -117,14 +113,16 @@ export const ProfileButton = ({ user, MenuOptions, className }) => {
         <StyledButton
           onClick={event => setAnchorEl(anchorEl ? null : event.currentTarget)}
           className={className}
-          endIcon={<Avatar>{userInitial}</Avatar>}
+          endIcon={<StyledAvatar initial={userInitial}>{userInitial}</StyledAvatar>}
```
edmofro, line 116: No `src` here?

packages/ui-components/src/components/UserMessage/UserMessageFooter.js
```diff
@@ -40,19 +40,25 @@ const FooterAvatar = styled(Avatar)`
...
+  return (
+    <FooterContainer>
+      <FooterInner>
+        <FooterText>
+          Last updated on:{' '}
+          {format(dateTime, `${DAY_MONTH_YEAR_DATE_FORMAT} - ${AM_PM_DATE_FORMAT}`)}
+        </FooterText>
+        <FlexSpaceBetween>
+          <FooterUser>{user.name}</FooterUser>
+          <FooterAvatar initial={userInitial} src={user.avatar}>
```
edmofro, line 54: The first instance of an avatar uses `user.profileImage`, while this uses `user.avatar` - is there a reason for that?

edmofro (approved):
One optional change: it seems we _always_ want the user's initial to show up inside, so we could pass in `initial` just as a prop, and do the nesting inside the Avatar component, rather than passing in `{userInitial}` as the child as well?

packages/ui-components/src/components/ProfileButton.js
tcaiger: I didn't put it in because I preferred how the initial looked. But on reflection, I think that is silly and the user can decide.

packages/ui-components/src/components/UserMessage/UserMessageFooter.js
tcaiger: Good pick up. The user message was developed a while ago before the profileImage existed.

PR #1455, opened 2020-10-08
1120 fix spectrum scale for no data

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1120
### Changes:
Tiny refactor of map overlay legends in general, will need a bit of a regression test (not needed for the fix, maybe I shouldn't have?)

Also got rid of the little empty box for radius map overlays.
Also made the map control always render on the far right, even when loading the overlay.


packages/web-frontend/src/components/Marker/markerColors.js
```diff
@@ -65,6 +65,7 @@ export function resolveSpectrumColour(scaleType, scaleColorScheme, value, min, m
 }
 
 const normaliseToPercentage = (value, min = 0, max = 1) => {
+  if (min === max) return 0.5;
```
edmofro, line 68: I would have thought 
```suggestion
  if (min === max) return 1;
```
was more appropriate? If you graphed the min tending towards the max, the percentage would get closer and closer to 1, and I wouldn't expect it to then jump back to 0.5. You may have a good reason for 0.5 though!

packages/web-frontend/src/containers/MeasureLegend/LegendEntry.js
```diff
@@ -0,0 +1,63 @@
...
+      <div>{label}</div>
+    </Key>
+  );
+};
+
+const mapLegendStateToProps = () => {
+  return state => ({
+    hiddenMeasures: state.map.measureInfo.hiddenMeasures,
+  });
+};
```
tcaiger, line 36: const mapStateToProps = () => state => ({
  hiddenMeasures: state.map.measureInfo.hiddenMeasures,
});

packages/web-frontend/src/containers/MeasureLegend/LegendEntry.js
```diff
@@ -0,0 +1,63 @@
...
+const mapLegendDispatchToProps = () => {
+  return dispatch => ({
+    onClick: (key, value, hide) =>
+      dispatch({
+        key,
+        value,
+        type: hide ? 'HIDE_MAP_MEASURE' : 'UNHIDE_MAP_MEASURE',
+      }),
+  });
+};
```
tcaiger, line 47: const mapDispatchToProps = () => dispatch => ({
  onClick: (key, value, hide) =>
    dispatch({
      key,
      value,
      type: hide ? 'HIDE_MAP_MEASURE' : 'UNHIDE_MAP_MEASURE',
    }),
});

packages/web-frontend/src/containers/MeasureLegend/LegendEntry.js
tcaiger: Might as well use arrow function 'implicit return' here

tcaiger: And might as well just use the conventional 'mapStateToProps'. We don't need to say legend since this is now the only mapState in the file. This is one of the benefits of having one component for file as you have so nicely done!

packages/web-frontend/src/containers/MeasureLegend/LegendEntry.js
```diff
@@ -0,0 +1,63 @@
...
+  onClick: PropTypes.func.isRequired,
+  hiddenMeasures: PropTypes.object.isRequired,
+  unClickable: PropTypes.bool,
+};
+
+LegendEntry.defaultProps = {
+  unClickable: false,
+};
+
+export default connect(mapLegendStateToProps, mapLegendDispatchToProps)(LegendEntry);
```
tcaiger, line 63:
```suggestion
export default connect(mapStateToProps, mapDispatchToProps)(LegendEntry);
```

packages/web-frontend/src/containers/MeasureLegend/LegendEntry.js
```diff
@@ -0,0 +1,63 @@
...
+    onClick: (key, value, hide) =>
+      dispatch({
+        key,
+        value,
+        type: hide ? 'HIDE_MAP_MEASURE' : 'UNHIDE_MAP_MEASURE',
+      }),
+  });
+};
+
+LegendEntry.propTypes = {
```
tcaiger, line 49: I believe it is more conventional to put the propTypes before the mapState functions as they belong to the 'unconnected' component. It is not really consistently applied in web-frontend but if you look at `SearchBar/index.js` you will see a good example.

packages/web-frontend/src/containers/MeasureLegend/MultiLegend.js
```diff
@@ -0,0 +1,111 @@
...
+      return MarkerLegend;
+  }
+};
+
+// returning <LegendOuterFrame /> keeps the map control on the right hand side
+const renderEmptyMultiLegend = () => <LegendOuterFrame />;
+
+const MultiLegend = ({ measureOptions, isMeasureLoading }) => {
+  if (isMeasureLoading) {
+    return renderEmptyMultiLegend();
```
tcaiger, line 62:
```suggestion
    return <LegendOuterFrame />;
```

packages/web-frontend/src/containers/MeasureLegend/MultiLegend.js
```diff
@@ -0,0 +1,111 @@
...
+    default:
+      return MarkerLegend;
+  }
+};
+
+// returning <LegendOuterFrame /> keeps the map control on the right hand side
+const renderEmptyMultiLegend = () => <LegendOuterFrame />;
+
+const MultiLegend = ({ measureOptions, isMeasureLoading }) => {
+  if (isMeasureLoading) {
```
tcaiger, line 61:
```suggestion
const displayedLegends = measureOptions.filter(({ type }) => type !== MEASURE_TYPE_RADIUS);

  if (isMeasureLoading || !displayedLegends) {
    return renderEmptyMultiLegend();
  }
```

tcaiger: It would be great if possible to handle both empty legend cases at once at the top. measureOptions.filter should always be safe to call because measureOptions is a required field in propTypes. If it causes errors, then you could update measureOptions to have a defaultPropType = [] or something;

packages/web-frontend/src/containers/MeasureLegend/MultiLegend.js
```diff
@@ -0,0 +1,111 @@
...
+        hasRadiusLayer={hasRadiusLayer}
+        hasColorLayer={hasColorLayer}
+        measureOptions={mo}
+      />
+    );
+  });
+
+  // returning <LegendOuterFrame /> keeps the map control on the right hand side
+  if (legends.length === 0) return renderEmptyMultiLegend();
+
```
tcaiger, line 88:
```suggestion

```

tcaiger: assuming the above suggestion works

packages/web-frontend/src/containers/MeasureLegend/labels.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
 import React from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
```
tcaiger, line 8: It would be great to move `LabelLeft`and `LabelRight`to SpectrumLegend.js since that is the only place they are used. As a general rule it is good to have constants and components defined as close as possible to their usage.

Then this file only contains the `NoDataLabel` component and so the file can be renamed to `NoDataLabel.js`

packages/web-frontend/src/containers/MeasureLegend/common.js
```diff
@@ -0,0 +1,14 @@
...
+
+import styled from 'styled-components';
+
+export const LegendContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  align-items: center;
+  justify-content: center;
+`;
```
tcaiger, line 14: These are entirely optional suggestions. 

Option 1
You could remove this re-usable component and use [material-ui box component](https://material-ui.com/components/box/) inline. eg 
```
<Box display="flex" alignItems="center" justifyContent="center" flexWrap="wrap">
`

**Review by tcaiger — CHANGES_REQUESTED** [review]
Great work - I for one love to see some refactoring so yes you should have. Sometimes it would be worth making a seperate PR for the refactoring and the actual change but I think in this instance the change is simple enough to bundle them together. 

Apologies but I have been pretty pedantic with my review 😬  I thought that since it is a refactor we might as well clean up as much as possible. Feel free to disagree with anything though. And send me a slack or reply if you have any questions about my suggestions.

File: packages/web-frontend/src/containers/MeasureLegend/MarkerLegend.js [hunk 11]
```diff
@@ -0,0 +1,147 @@
...
+  }
+  if (hasRadiusLayer) {
+    // show rings, as we are indicating radius colors
+    return getMarkerForOption(LEGEND_RADIUS_ICON, color);
+  }
+  // color is the only measure here - show pins
+  return getMarkerForOption(DEFAULT_ICON, color);
+}
+
+export const MarkerLegend = ({ measureOptions, hasIconLayer, hasRadiusLayer, hasColorLayer }) => {
```
tcaiger, line 84:
```suggestion
export const MarkerLegend = React.memo(({ measureOptions, hasIconLayer, hasRadiusLayer, hasColorLayer }) => {
```

packages/web-frontend/src/containers/MeasureLegend/MarkerLegend.js
```diff
@@ -0,0 +1,147 @@
...
+    );
+  }
+
+  return (
+    <LegendContainer>
+      {keys}
+      {nullKey}
+    </LegendContainer>
+  );
+};
```
tcaiger, line 140:
```suggestion
});
```

packages/web-frontend/src/containers/MeasureLegend/MarkerLegend.js
tcaiger: I would recommend wrapping computation heavy components such as this in [React.memo](https://reactjs.org/docs/react-api.html#reactmemo) for a small performance boost. I think it would be worth it for the LegendEntry, MultiLegend and SpectrumLegend components too.

packages/web-frontend/src/components/Marker/markerColors.js
Alastair-L: > If you graphed the min tending towards the max, the percentage would get higher and higher and probably be at least 1

I'm not sure what you mean by this? To my mind, as the min tends to the max the percentage will get lower and lower (as it is the relative position of the value that we are interested in)

image

There is however, an inconsistency as above, so we can do:
```
if value < min: 0
if value > max: 1
if min = max = value: 0.5
```
(but that isn't a use case at the moment, because of how scales are calculated)

packages/web-frontend/src/containers/MeasureLegend/LegendEntry.js
Alastair-L: Much nicer!

packages/web-frontend/src/components/Marker/markerColors.js
edmofro: We know the value can't be outside the bounds of the min and max (that's kinda the definition, right?), so as the interval gets smaller, the percentage of that interval any number inside takes up becomes larger.

packages/web-frontend/src/containers/MeasureLegend/MarkerLegend.js
Alastair-L: Hurrah! This is great!

packages/web-frontend/src/containers/MeasureLegend/MultiLegend.js
Alastair-L: Always love a diff with more minus than plus! And measureOptions is defaulted to [] in mapStateToProps, so that should be no issue :).

packages/web-frontend/src/components/Marker/markerColors.js
Alastair-L: The value can be outside the bounds, and is in some current use cases (covidau).

Also, I think it generally depends on if it is the min or the max that is changing. If it is the min that increases, the result will decrease, no?
image
image

packages/web-frontend/src/containers/MeasureLegend/LegendEntry.js
Alastair-L: That actually makes a lot of sense!

packages/web-frontend/src/components/Marker/markerColors.js
edmofro: Oh really?? I thought the min and max were derived from the actual values. 

And yes, true about the min increasing.

Where do the min and max come from?

Alastair-L: `getSpectrumScaleValues` in `src/utils/measures`. There is various logic there to modify the min and max (although you are right, the default is to derive them from the actual values).

packages/web-frontend/src/containers/MeasureLegend/labels.js
Alastair-L: I decided to go one further and use LegendEntry. However, this causes some (I believe positive) change to the ui:

What do you think?
(Note: the difference in background color is not related)
## Black:

### Old:
image
image

New:
image
image

## Grey:
### Old (no change on click):
image

### New:
image
image

packages/web-frontend/src/containers/MeasureLegend/common.js
Alastair-L: As a css newbie, I would much prefer to have it all spirited away in another file like `LegendContainer.js` than have to read it inline (but this is possibly just a personal thing which will go away with experience).

packages/web-frontend/src/components/Marker/markerColors.js
edmofro: Right! Well, your suggestion sounds great then:
if value < min: 0
if value > max: 1
if min = max = value: 0.5

Alastair-L: Sounds good, thanks Edwin 👍

Alastair-L:
> Sometimes it would be worth making a separate PR for the refactoring and the actual change 

Will definitely keep that in mind for next time 👍 

> I have been pretty pedantic with my review 😬

Thanks for that! I'm very new to redux (well, programming in general) so it's good to learn some stuff from the UI master 😄. Plus it's a good time to fix this stuff when it will be tested anyway.

tcaiger (approved):
Looks great to me! Thanks for all the effort refactoring. All future devs thank you in advance :)

PR #1459, opened 2020-10-08
1276: Reporting Package POC

rohan-bes:
### Issue #https://github.com/beyondessential/tupaia-backlog/issues/1276:

Basic proof of concept for the reporting package.

TODO:
- Unit tests!!
- Design structure for fetch and output steps
- Add exception handling for useful error messages (eg. `error in transform[7] - 'select': ...` )


packages/reports/src/aggregator/Aggregator.ts
```diff
@@ -0,0 +1,56 @@
+import { Aggregator as BaseAggregator } from '@tupaia/aggregator';
```
rohan-bes, line 1: This is largely copied from `Aggregator` in `web-config-server`. Still deciding how best to handle this duplicate code. There's a separate card for cleanup of this package where we can decide that: https://github.com/beyondessential/tupaia-backlog/issues/1283

packages/reports/src/reportBuilder/functions/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  value,
+  add,
+  eq,
+  neq,
+  gt,
+  exists,
+  convertToPeriod,
+  periodToTimestamp,
+  periodToDisplayString,
+};
```
rohan-bes, line 14: I think the idea with exporting collections of functions like this is that we can easily group them by functionality/scope. For now we've just got one general scope but I think this may change in the future

packages/reports/src/reportBuilder/reportBuilder.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  constructor(report, aggregator: Aggregator, query: FetchReportQuery) {
+    this.report = report;
+    this.aggregator = aggregator;
+    this.query = query;
+  }
+
+  build = async (): Promise<BuildReport> => {
+    const data: FetchResponse = await fetch(this.report.config.fetch, this.aggregator, this.query);
+    const builtTransform = buildTransform(this.report.config.transform);
```
rohan-bes, line 37: So this pattern of building a transform function from the config is one I quite like and would like to generalise it to `fetch` and `output` when they are a bit more fleshed out.

The idea is that each step of the process is responsible for building a function from the config it's given and in that process we validate that config. Typescript forces us to validate the config anyway, and by doing it as a separate step to running the transform, we can validate the user's config at the time of saving it.

packages/reports/reports.hjson
```diff
@@ -0,0 +1,212 @@
+{
```
edmofro, line 1: Is the `.hjson` intentional? What does it mean??

packages/reports/reports.hjson
```diff
@@ -0,0 +1,212 @@
...
+                "isDataRegional": true
+              }
+            ]
+          }
+        }
+      ],
+      "transform": [
+        {
+          "transform": "select",
+          "$.dataElement": "$.value",
```
edmofro, line 35: Clever! So `$` is the row, and `$.` is `row.`?

packages/reports/reports.hjson
```diff
@@ -0,0 +1,212 @@
...
+          "...": ["period", "organisationUnit"]
+        },
+        {
+          "transform": "sort",
+          "by": "$.period",
```
edmofro, line 40: Do we need to specify a custom sort type for periods?

packages/reports/reports.hjson
```diff
@@ -0,0 +1,212 @@
...
+          "transform": "aggregate",
+          "organisationUnit": "group", 
+          "...": "last"
+        },
+        {
+          "transform": "aggregate",
+          "organisationUnit": "drop",
+          "period": "drop",
+          "...": "sum"
+        },
```
edmofro, line 52: This is really powerful, and I love it, but also the sort of thing we'll probably pack together for the sake of data stewards into a saved block

packages/reports/reports.hjson
```diff
@@ -0,0 +1,212 @@
...
+              }
+            ]
+          }
+        }
+      ],
+      "transform": [
+        {
+            "transform": "select",
+            "$.dataElement": "$.value",
+            "'period'": "$.period"
```
edmofro, line 109: I don't understand this line...

packages/reports/reports.hjson
```diff
@@ -0,0 +1,212 @@
...
+        },
+        {
+          "transform": "aggregate",
+          "period": "group",
```
edmofro, line 113:
```suggestion
          "period": "combine",
```
I feel this could be easier for data stewards to understand. What do you think? Perhaps `group` is more related to the SQL groupBy?

packages/reports/reports.hjson
```diff
@@ -0,0 +1,212 @@
...
+      "transform": [
+        {
+            "transform": "select",
+            "$.dataElement": "$.value",
+            "'period'": "convertToPeriod($.period, 'QUARTER')",
+            "'organisationUnit'": "$.organisationUnit"
+        },
+        {
+            "transform": "filter",
+            "where": "not (exists($.RHS4UNFPA809) and gt(1, $.RHS4UNFPA809))"
```
edmofro, line 153: oh cool...!

packages/reports/reports.hjson
```diff
@@ -0,0 +1,212 @@
...
+        },
+        {
+            "transform": "aggregate",
+            "organisationUnit": "group",
+            "period": "group",
+            "...": "last"
+        },
+        {
+            "transform": "select",
+            "'organisationUnit'": "'COUNTRY'",
```
edmofro, line 163: Is this doing something special? Or is it literally just setting `organisationUnit: 'COUNTRY'`?

packages/reports/reports.hjson
edmofro: Another option `match`

packages/reports/src/reportBuilder/transform/functions/where.ts
```diff
@@ -0,0 +1,48 @@
...
+const buildParams = (params: unknown): WhereParams => {
+  if (typeof params !== 'object' || params === null) {
+    throw new Error(`Expected params object but got ${params}`);
+  }
+
+  if (!('where' in params)) {
+    return {};
+  }
+
+  const whereClaus: unknown = params.where;
```
edmofro, line 34:
```suggestion
  const whereClause: unknown = params.where;
```

packages/reports/reports.hjson
rohan-bes: Yep I totally agree. It's basically just an existing aggregation type, so it could be something like:
```
{
          "transform": "finalValuePerPeriod",
          "period": "YEAR"
}
```
or even just:
```
 "finalValuePerYear"
```

packages/reports/reports.hjson
rohan-bes: Exactly, `mathjs` lets you add objects to the expression scope, and you can reference them with standard JS notation.
eg. `$.BCD1 or $['BCD1']`

packages/reports/reports.hjson
rohan-bes: It's just a json format that allows me to add comments which makes it really easy to play around with the report transforms. Won't be in the final product 👍 

@kael89 also recommended jsonc for this, which I might switch to as prettier doesn't accept hjson so it doesn't auto format the file 😢

kael89: 👍 for `.jsonc`, also it's fine if we commit those as templates/documentation in cases it makes sense, eg [dashboardReports.example.jsonc](https://github.com/beyondessential/tupaia/blob/dev/packages/web-frontend/cypress/config/dashboardReports.example.jsonc)

packages/reports/reports.hjson
rohan-bes: Yup, you're right, but using `'` is just what I've had to do currently to get mathjs to parse the expression currently. Ideally we'd check the string and see if it contains a `$` (or something else that requires parsing) and determine whether to parse it at all. That way we wouldn't need to wrap literals in `'` in these cases.

packages/reports/reports.hjson
rohan-bes: I think I slightly prefer `combine` to `match`. Honestly I've struggled over the right term for this one for a bit, nothing that I've come up with quite captures the right idea. Yeah I chose `group` intially as I was thinking about SQL `groupBy` at the time, but I don't think it's the best option...

I've also considered something like this:
```
{
    "transform": "aggregate",
    "per": ["period", "organisationUnit"],
    "BCD1": "sum",
    "...": "count"
}
``` 

Our current aggregation names all follow a convention of `<aggregationType>Per[ period | organisationUnit ]` eg. `sumPerPeriod`. I think it's a good way of describing the operation but I don't know the best way to re-use that naming convention through config here...

packages/reports/reports.hjson
rohan-bes: Yeah! It does boolean logic by itself, and we can even use our custom defined functions with it quite gracefully 😄

I've been using `mathjs` so far just to get things running, but I'm kinda tempted to use it for the core expression parsing for this config given how well it's been working. Not definite in that stance though, and I think it requires us to consider a few things, but even if we don't I think using a similar design for our own expression parsing libraries could be really neat

packages/reports/reports.hjson
rohan-bes: Nope, you're right it's just setting it to the string `'COUNTRY'`. What this should really be is something like:
`"'organisationUnit'": "getCountryCode($.organisationUnitCode)"`

But I haven't got anything like that implemented as of yet.

PR #1294, opened 2020-09-10
995 laos eoc weather entities

IgorNadj:
### Changes:

- Add entity type 'city'
- Import 10 most populous cities of Laos

PR #1465, opened 2020-10-08
Update Laos EOC entities

IgorNadj:
### Issue #: 995

Update entity names

---

This is the same PR as https://github.com/beyondessential/tupaia/pull/1294 but with different spellings.

### DIFF
--- a/packages/database/src/migrations/20201009061046-UpdateLaosEocEntities-modifies-data.js
+++ b/packages/database/src/migrations/20201009061046-UpdateLaosEocEntities-modifies-data.js
@@ -1,0 +1,257 @@
+'use strict';
+
+import { arrayToDbString, generateId, insertObject } from '../utilities';
+import { insertEntity } from '../utilities/migration';
+import { EntityModel } from '../modelClasses';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+/*
+ * Each province has a capital and each capital has a point center lat/lon.
+ * See: https://www.wikiwand.com/en/Administrative_divisions_of_Laos
+ *
+ * Each province already has bounds set in the database.
+ */
+const PROVINCES = {
+  Attapeu: {
+    capital: {
+      name: 'Xaisettha',
+      lat: 14.82,
+      lon: 106.820556,
+    },
+  },
+  Bokeo: {
+    capital: {
+      name: 'Houayxay',
+      lat: 20.263056,
+      lon: 100.433611,
+    },
+  },
+  Bolikhamsai: {
+    capital: {
+      name: 'Paksan',
+      lat: 18.396389,
+      lon: 103.655833,
+    },
+  },
+  Champasak: {
+    capital: {
+      name: 'Pakse',
+      lat: 15.116667,
+      lon: 105.783333,
+    },
+  },
+  Houaphanh: {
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20201009061046-UpdateLaosEocEntities-modifies-data.js
```diff
@@ -0,0 +1,257 @@
...
+    'INSERT INTO entity (id, code, parent_id, name, type, point, country_code, bounds, metadata) VALUES \n';
+
+  const rowsSql = [];
+
+  for (const province of Object.values(PROVINCES)) {
+    let rowSql = '(';
+
+    const point = { type: 'Point', coordinates: [province.capital.lon, province.capital.lat] };
+
+    rowSql +=
```
Alastair-L, line 212: Is performance such an issue/priority here? Don't get me wrong, I love building the sql like this!

However, given that it is just a migration perhaps it will be more readable as a single `db.runSql` inside the loop, similar to how you've done `entity_relation`? (I assume that you can't insert geoJson with `insertObject`, otherwise that would be ideal).

If it's too messy, you could extract it to a `insertEntity` function.

packages/database/src/migrations/20201009061046-UpdateLaosEocEntities-modifies-data.js
```diff
@@ -0,0 +1,257 @@
...
+  for (const province of Object.values(PROVINCES)) {
+    await insertObject(db, 'entity_relation', {
+      id: generateId(),
+      parent_id: parentId,
+      child_id: province.capital.id,
+      entity_hierarchy_id: entityHierarchyId,
+    });
+  }
+
+  return null;
```
Alastair-L, line 246: Is there a reason to return null here? Should I start doing it? Otherwise perhaps this is unneeded?

packages/database/src/migrations/20201009061046-UpdateLaosEocEntities-modifies-data.js
```diff
@@ -0,0 +1,257 @@
...
+  sql += rowsSql.join(',\n');
+
+  sql += ';';
+
+  await db.runSql(sql);
+
+  /*
+   * Entity relation
+   */
+  for (const province of Object.values(PROVINCES)) {
```
Alastair-L, line 237: clean!

packages/database/src/migrations/20201009061046-UpdateLaosEocEntities-modifies-data.js
```diff
@@ -0,0 +1,257 @@
...
+  Xaisomboun: {
+    capital: {
+      name: 'Anouvong',
+      lat: 18.9057,
+      lon: 103.092,
+    },
+  },
+};
+
+// Pre-calculate entity codes (needed for both up() and down()), and id
```
Alastair-L, line 156: Perhaps this could go in `up`? Since you are deleting by regex you don't need it in `down` and personally I wouldn't expect any code to be run at top level when reading a migration. Definitely just an opinion though.

Alastair-L (approved):
Approving as all of my comments are a matter of opinion. Nice work!

packages/database/src/migrations/20201009061046-UpdateLaosEocEntities-modifies-data.js
IgorNadj: Ah yes I would love to use insertEntity but it doesn't play nice with inserting geojson fields :(

packages/database/src/migrations/20201009061046-UpdateLaosEocEntities-modifies-data.js
IgorNadj: Hmm, I just started copying other people doing it, I don't think it matters at all

PR #1467, opened 2020-10-08
OS-agnostic refreshDatabase script

kael89:
### Issue #:
[No issue]

The inspiration for this occurred while working on a new [Tupaia Monorepo setup](https://docs.beyondessential.com.au/books/software-development/page/tupaia-monorepo-setup) documentation, to assist our testers setting up tupaia.org locally so that they can run e2e tests. I had realized for some time that the `refreshDb.sh` script does not work in Windows installations of PostgreSQL.

In fact, I had created a personal batch script (Windows cmd) a while ago to overcome this, but for some reason I assumed that no one else would have this problem, since if they had I would know... Well, turns out I was wrong 🙂

Then I thought it was high about time I commit my batch script in the codebase and make this operation Windows friendly. And then it occurred to me... Why bother maintaining two scripts when `node` works on both OS out of the box?

I believe that moving forward we should convert all our dev-related scripts to be `node` scripts. I used [yargs](https://www.npmjs.com/package/yargs) here, which seems to be an excellent CLI helper package.

### Screenshots

My windows machine has a native PostgreSQL installation. If I run the script under WSL:

error wsl

If I run it under Windows, it works as the `refreshDb.sh` used to in MacOS machines.

An example of the auto-generated help output by yargs:

```bash
yarn refresh-database --help
```

output


README.md
```diff
@@ -46,6 +46,25 @@ if you prefer to prebuild internal dependencies, add `--skip-internal` to the ab
 
 Most packages will require a .env file. `.env.example` files indicate the required variables per package.
 
+### Local database
```
Alastair-L, line 49: Perhaps this could fit in the database package, with a link from the root readme?

README.md
```diff
@@ -46,6 +46,25 @@ if you prefer to prebuild internal dependencies, add `--skip-internal` to the ab
...
+Set up PostgreSQL on your machine and create the database 'tupaia', with credentials matching those in
+the .env file.
+
+The project requires importing an initial database dump. Grab the latest database dump from dev server using
+
+```bash
+yarn dump-database ~/path/to/key.pem
+```
+
+(pem key can be obtained in LastPass - Tupaia Main Server). This will create a db dump and pull it to your current directory. You also need to be added into the Security Groups of the server.
```
Alastair-L, line 60: I was under the impression that we removed all references to lastpass, is that not true?

scripts/node/refreshDatabase.js
```diff
@@ -0,0 +1,56 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+const Script = require('./Script');
+
+class RefreshDatabaseScript extends Script {
+  config = {
+    command: '* <dumpPath>',
+    options: {
+      migrate: {
```
kael89, line 14: By the way this option doesn't work on Windows, because we use UNIX paths in the `migrate` command (`'../../babel.config.json'). This is also true in current `dev`, so this is not a regression.

I guess that the solution would be to encapsulate the `migrate` command under a new node script

scripts/node/refreshDatabase.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  refreshDatabase() {
+    this.logInfo('Refreshing the tupaia database...');
+
+    this.verifyPsql();
+    this.exec('psql -U postgres -c "DROP DATABASE IF EXISTS tupaia"');
+    this.exec('psql -U postgres -c "CREATE DATABASE tupaia WITH OWNER tupaia"');
+    this.exec('psql -U postgres -d tupaia -c "CREATE EXTENSION postgis"');
+    this.exec('psql -U postgres -c "ALTER USER tupaia WITH SUPERUSER"');
```
Alastair-L, line 37: You've probably thought about it but these are in the opposite order to in the bash script. (I don't normally run either of these 😬)

README.md
kael89: That's correct, although in some cases it's not possible without a further restructure of either the documentation or the code. There is a task for that, see https://github.com/beyondessential/tupaia-backlog/issues/1202

In this specific case, we could avoid those references by moving this documentation to a Tupaia resource, eg our BookStack. It would be easy to cover this process in a "Step 5" of the new [Tupaia Monorepo setup guide](https://docs.beyondessential.com.au/books/software-development/page/tupaia-monorepo-setup), but those scripts will still exist in the codebase, just be undocumented/non operational for people that don't have access to our server.

Thoughts?

Alastair-L: I think that's a great idea! Perhaps also add a comment at the start of the script itself mentioning that as well, in case anyone stumbles across it.

scripts/node/refreshDatabase.js
kael89: The first one will be required only in the initial setup of the database, but doesn't harm running in each consequent case. So it's fine you don't run it 🙂

The order here doesn't matter from a functionality perspective, but I believe that the new order makes the command flow more readable.

I guess you do have to have to run `ALTER USER...`, since if not you'll get a bunch of errors? Unless if you have made `tupaia` a SUPERUSER in your local db 😉

Alastair-L: Yeah thanks, that makes sense!

And yeah, I never knew that I should make tupaia not a super user 😁...

kael89: Well, it may not create an issue ever, but it's good to replicate the actual DB as closely as possible. Eg it may be that a specific command you run succeeds in your local DB, but not in our CI/CD pipeline, because it depends on admin rights.

kael89: And those are usually the frustrating category of issues to solve 😆

Alastair-L: Yeah thanks, I never knew!

That's the hidden (or not so hidden perhaps) advantage of these utility scripts. Making sure that everyone's process is consistent to minimize the 'works on my machine' problem.

kael89:
This package can also help making our scripts OS-agnostic: [shelljs](https://www.npmjs.com/package/shelljs)

README.md
edmofro: I'd vote for keeping some documentation in place, and just have flags that mark the pathway for BES Internal, or Open Source Community within the readme. Personally, I don't see it as that bad having a reference to Lastpass in the readme, so long as there's a clear alternative for someone external to follow (in this case, it might be running refresh database using testDataDump). If anything, it makes things more transparent, and explains otherwise unanswered questions, like "what's this dump-database script about". 

Also happy for the BES Internal pathway to be a link to bookstack if others prefer that (and it would make more sense if it's a long set of steps or contains secrets), but again I still think that should be kept in the readme and flagged as BES Internal.

What do you guys think?

scripts/node/testAll.js
```diff
@@ -65,22 +68,29 @@ class TestAllScript extends Script {
...
-    if (this.options.silentMode) {
+
+    let passesTests = true;
+    try {
+      this.exec(`yarn workspace @tupaia/${packageName} test`, {
+        printOutput: !this.args.silent,
+      });
+    } catch (error) {
+      passesTests = false;
+    }
```
edmofro, line 82: This change to a try/catch seems unrelated? Was it broken before?

edmofro (approved):
This all looks awesome to me! Even more powerful/succinct way of defining new scripts 🙌

scripts/node/testAll.js
kael89: Good point, I should have left a comment! This is because I switched from `child_process.spawnSync` to `child_process.execSync`, which throws an error instead of returning an error code and the `stderr` output. I've found that this behaviour is more JS-friendly, from a programming perspective

edmofro: Oh right! Makes sense, thanks for the explanation

README.md
kael89: I believe there are merits in both approaches. I'd say we defer this decision for when https://github.com/beyondessential/tupaia-backlog/issues/1202 is implemented. Even if we remove this reference to LastPass, we only "fake" more open source compatibility - people will still have a hard time figuring out how to setup the whole thing.

So let's keep this here for now, and discuss a more generic strategy of making our project approachable (eg exposing a docker container for local dev etc) when the time comes.

kael89: I guess that we can still replace this with a link to our BookStack and an "BES internal" label? I wouldn't worry about the script because as @edmofro pointed out, it does work as long as a dev has a DB dump, which may obtained from a different source than our servers.

README.md
kael89: Now that I replaced the documentation with a link, we can probably leave this here and don't link just a link from the database README? What do you think? Given that most people will look at the main README for the basics (setting up the DB etc)

README.md
```diff
@@ -48,22 +48,7 @@ Most packages will require a .env file. `.env.example` files indicate the requir
...
-```
-
-(pem key can be obtained in LastPass - Tupaia Main Server). This will create a db dump and pull it to your current directory. You also need to be added into the Security Groups of the server.
-
-After pulling the latest database dump from the `dev` server, run:
-
-```bash
-yarn refresh-database dump.sql
-```
+🔑 **BES internal:** [Tupaia monorepo setup](https://docs.beyondessential.com.au/books/software-development/page/tupaia-monorepo-setup#bkmrk-step-4.-install-post) - steps 4 and 5
```
edmofro, line 51: Love the emoji!

edmofro (approved):

README.md
Alastair-L: Sounds good 👍, I agree

PR #1468, opened 2020-10-09
1294 laos eoc observed weather map overlays

IgorNadj:
### Changes:
* Add Laos EOC map overlays
* Add one decimal place value formatter
* Update storybook

---

### Screenshots:
https://github.com/beyondessential/tupaia-backlog/issues/1294#issuecomment-705892635


rohan-bes (approved):
Nice work! Looks really clean 😄

IgorNadj:
@rohan-bes do you mind looking at the last 3 commits please:
- Fix Reset to Default dates not working when switching measures
- Fix date picker reset using wrong default dates
- Fix tests

Basically the change is that instead of reseting to the initial start/end date passed into the component, we always reset to the default date range. 

The issue I was solving was one where it would reset to the wrong dates, and I chose this implementation because it makes more sense to me. 

Instead of relying on initial start/end date props passed into the component to figure out the dates used for the reset, we just use the default dates. To me, start/end dates are the current date, and completely unrelated to the default dates. If a developer wants to use different dates for the reset, they should change the default dates.

rohan-bes:
> @rohan-bes do you mind looking at the last 3 commits please:
> 
> * Fix Reset to Default dates not working when switching measures
> * Fix date picker reset using wrong default dates
> * Fix tests
> 
> Basically the change is that instead of reseting to the initial start/end date passed into the component, we always reset to the default date range.
> 
> The issue I was solving was one where it would reset to the wrong dates, and I chose this implementation because it makes more sense to me.
> 
> Instead of relying on initial start/end date props passed into the component to figure out the dates used for the reset, we just use the default dates. To me, start/end dates are the current date, and completely unrelated to the default dates. If a developer wants to use different dates for the reset, they should change the default dates.

Yeah I agree with this reasoning, looks good to me 👍

PR #1471, opened 2020-10-09
1295 laos eoc forecast weather map overlays

IgorNadj:
### Changes:
- Add Laos EOC map overlays for forecast data
- Change WeatherService to always require start/end dates (and so make sure that forecast weather requests that specify start/end dates are handled correctly)
---

### Screenshots:
https://github.com/beyondessential/tupaia-backlog/issues/1295#issuecomment-705946095

### DIFF
--- a/packages/data-broker/src/services/weather/WeatherService.js
+++ b/packages/data-broker/src/services/weather/WeatherService.js
@@ -183,7 +183,7 @@ export class WeatherService extends Service {
     const getDataForEntity = async entity => {
       const { lat, lon } = entity.pointLatLon();
 
-      // Maximum forecast is 16 days, we request all of it and filter it dow to the dates we need.
+      // Maximum forecast is 16 days, we request all of it and filter it down to the dates we need.
       // Performance looks fine requesting 16 days.
       const days = 16;
 
--- a/packages/data-broker/src/services/weather/WeatherService.js
+++ b/packages/data-broker/src/services/weather/WeatherService.js
@@ -1,4 +1,4 @@
-import moment from 'moment';
+import { ValidationError } from '@tupaia/utils';
 import { Service } from '../Service';
 import { ApiResultTranslator } from './ApiResultTranslator';
 import { DateSanitiser } from './DateSanitiser';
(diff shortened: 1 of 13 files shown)


packages/data-broker/src/services/weather/WeatherService.js
```diff
@@ -149,39 +143,55 @@ export class WeatherService extends Service {
...
+        data: filteredData,
+      };
+    };
+
     // Run requests in parallel for performance
     const getDataForEntity = async entity => {
       const { lat, lon } = entity.pointLatLon();
 
-      const days = 1; // just need today
+      // Maximum forecast is 16 days, we request all of it and filter it dow to the dates we need.
```
edmofro, line 186:
```suggestion
      // Maximum forecast is 16 days, we request all of it and filter it down to the dates we need.
```

packages/data-broker/src/tests/services/weather/WeatherService.test.js
```diff
@@ -136,10 +136,30 @@ describe('WeatherService', () => {
...
     });
+
+    it('throws when no dates are provided', async () => {
+      const mockModels = await createMockModelsStubWithMockEntity();
+
+      const mockApi = createWeatherApiStubWithMockResponse();
+
+      const service = new WeatherService(mockModels, mockApi);
+
+      const functionCall = async () =>
```
edmofro, line 147:
```suggestion
      const pullWithUndefinedDates = async () =>
```
With `functionCall` I have to think a bit harder about "ok, but what function call?"

edmofro (approved):
Nice! A very clean implementation, shows all of your work is paying off :-)

I've made a couple of very minor comments since there was nothing else to pick up, feel free to take them or leave them.

edmofro:
@IgorNadj when you're back next week let me know if any of those changes need re-review. Because of the force pushes I'm not sure exactly which commits to look at for the differences

IgorNadj:
> @IgorNadj when you're back next week let me know if any of those changes need re-review. Because of the force pushes I'm not sure exactly which commits to look at for the differences

Yep no worries, it's just rebases, no actual changes. If something came back from testing that I needed to fix I would add a new commit and request re-review.

PR #1472, opened 2020-10-09
Add trigger to disallow updating immutable table

edmofro:
I was thinking of doing this for `entity_relation`, but thought I would go down the proper route (see earlier PR). However, this is still very useful for ancestor_descendant_relation (and any other derived tables in future)

### DIFF
--- a/packages/database/src/runPostMigration.js
+++ b/packages/database/src/runPostMigration.js
@@ -16,6 +16,24 @@ const EXCLUDED_TABLES_FROM_TRIGGER_CREATION = [
   "'ancestor_descendant_relation'",
 ];
 
+// tables that should only have records created and deleted, and will throw an error if an update is
+// attempted
+const IMMUTABLE_TABLES = ["'ancestor_descendant_relation'"];
+
+const SELECT_TABLES_WITHOUT_TRIGGERS = `
+  SELECT t.table_name
+    FROM information_schema.tables t
+    LEFT JOIN information_schema.table_privileges privileges
+      ON t.table_name = privileges.table_name
+    WHERE NOT EXISTS (
+      SELECT * from pg_trigger p
+      WHERE concat(lower(t.table_name), '_trigger') = p.tgname
+    )
+    AND privileges.privilege_type = 'TRIGGER'
+    AND t.table_schema = 'public'
+    AND t.table_type != 'VIEW'
+`;
+
 export const runPostMigration = async driver => {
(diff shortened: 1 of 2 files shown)


packages/database/src/runPostMigration.js
```diff
@@ -27,37 +45,52 @@ export const runPostMigration = async driver => {
...
+        FOR EACH ROW EXECUTE PROCEDURE notification();
+      `),
+    ),
+  );
+
+  // Find any table in the database that doesn't have a corresponding notification trigger with
+  // the name {TABLE_NAME}_immutable_trigger (eg survey_response_immutable_trigger).
+  const { rows: tablesWithoutImmutableTriggerResults } = await driver.runSql(`
+    ${SELECT_TABLES_WITHOUT_TRIGGERS}
+    AND t.table_name IN (${IMMUTABLE_TABLES.join(',')});
```
rohan-bes, line 70: Super trivial, but perhaps it's a bit nicer to use the `arrayToDbString` function here? Saves having to remember to wrap the immutable table names in quotes

rohan-bes (approved):
Looks really good 😄 

Just want to confirm, so the idea is that we will make all tables without existing trigger functions to be immutable, on top of the explicitly declared immutable ones?

Just one comment but it's up to you if you want to do it 👍

packages/database/src/runPostMigration.js
edmofro: Oh yep, that'd be a nice improvement :+1:

edmofro:
Actually it's not the case that tables that don't have change notifiers are immutable - they're two distinct groups. But some tables we do want to make immutable - and now we have the ability to bake that in at the db error level :-)

rohan-bes:
> Just want to confirm, so the idea is that we will make all tables without existing trigger functions to be immutable, on top of the explicitly declared immutable ones?

> Actually it's not the case that tables that don't have change notifiers are immutable - they're two distinct groups. But some tables we do want to make immutable - and now we have the ability to bake that in at the db error level :-)

Omg I misread the `AND` as an `OR` the code here: 
` ${SELECT_TABLES_WITHOUT_TRIGGERS} AND t.table_name IN (${arrayToDbString(IMMUTABLE_TABLES)});`

Sorry for the confusion 😝

rohan-bes (approved):
Looks good, nice catch on the bug where the immutability trigger name wasn't matching `<tableName>_trigger` 👍

PR #1474, opened 2020-10-09
Fix database test issue

billli0:
### Issue #:
The database package test fails sometimes but it will pass when rebuilding the test.

image

### Changes:
Add database close command after test finished. I rebuilt the tests twice to make sure it is solved. We can investigate if it happens again.


IgorNadj:
Nice one!

IgorNadj (approved):

PR #1481, opened 2020-10-12
1388 psss profile photo

tcaiger:
### Changes:

- Add profile image field to PSSS

---

### Screenshots:
See issue https://github.com/beyondessential/tupaia-backlog/issues/1388

### DIFF
--- a/packages/psss/src/views/Tabs/ProfileTabView.js
+++ b/packages/psss/src/views/Tabs/ProfileTabView.js
@@ -8,9 +8,11 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { connect } from 'react-redux';
 import { useForm } from 'react-hook-form';
-import { Button, SmallAlert, TextField } from '@tupaia/ui-components';
+import MuiDivider from '@material-ui/core/Divider';
+import { Button, ProfileImageField, SmallAlert, TextField } from '@tupaia/ui-components';
 import { Main } from '../../components';
 import { updateProfile, getCurrentUser } from '../../store';
+import { createBase64Image } from '../../utils/createBase64Image';
 
 const Container = styled.section`
   padding-top: 1rem;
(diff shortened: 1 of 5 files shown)


packages/psss/src/views/Tabs/ProfileTabView.js
```diff
@@ -34,38 +36,82 @@ const SuccessMessage = styled(SmallAlert)`
...
+      fileId: `${user.id}-${fileName}`,
+      data: base64,
+    });
+    setStatus(STATUS.IDLE);
+  };
+
+  const handleFileDelete = () => {
+    // We need to set a file id to something so that the backend will pick up the profileImage field
+    // as an updatedField and save the change.
+    setProfileImage({ fileId: '-', data: null });
```
edmofro, line 95:
```suggestion
    setProfileImage({ fileId: '', data: null });
```
Does that work too?

edmofro (approved):
LGTM :-)

packages/psss/src/views/Tabs/ProfileTabView.js
tcaiger: no unfortunately not. In the submit handler, I use the fileId to determine whether or not the profileImage is updated. See line 70. I could tweak that if you think it is confusing.

packages/psss/src/views/Tabs/ProfileTabView.js
```diff
@@ -34,38 +36,82 @@ const SuccessMessage = styled(SmallAlert)`
...
-    setIsLoading(true);
+    setStatus(STATUS.LOADING);
     setErrorMessage(null);
     try {
       await onUpdateProfile({
         first_name: firstName,
         last_name: lastName,
         position: role,
         employer,
+        profileImage: profileImage.fileId && profileImage,
```
edmofro, line 70: Pointing me to this line made me notice that it's camel case, with the first two fields in snake case

packages/psss/src/views/Tabs/ProfileTabView.js
edmofro: Yeah I do find it a little confusing - could we check if `profileImage.data !== user.profileImage`?

PR #1485, opened 2020-10-13
Data broker test tweaks

kael89:
Your original PR looks awesome! I thought instead of requesting minor changes, I could as well do them myself, since some of them required some experimentation too


billli0 (approved):
These are very good changes and I've learned that `createJestMockInstance` can be used in this way! Thanks Kostas!

PR #1487, opened 2020-10-13
Fix a race condition in entity hierarchy building

edmofro:
If a rebuild was currently underway when another rebuild was scheduled, the one that was underway would resolve the newly added one too early (and other related issues with the shared instance fields). This fixes that issue.


nick-beyondessential (approved):
Looks good

PR #1491, opened 2020-10-13
Batch deletes of subtrees in entity hierarchy cacher

edmofro:
To avoid running out of database connections

### DIFF
--- a/packages/database/src/cachers/EntityHierarchyCacher.js
+++ b/packages/database/src/cachers/EntityHierarchyCacher.js
@@ -32,8 +32,7 @@ export class EntityHierarchyCacher {
     // rebuild all hierarchies
     const hierarchies = await this.models.entityHierarchy.all();
     const hierarchyTasks = hierarchies.map(async ({ id: hierarchyId }) => {
-      await this.deleteSubtree(hierarchyId, entityId);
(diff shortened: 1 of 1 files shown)


packages/database/src/cachers/EntityHierarchyCacher.js
```diff
@@ -82,22 +73,35 @@ export class EntityHierarchyCacher {
...
+    const deleteTasks = Object.entries(
+      subtreesForDelete,
+    ).map(async ([hierarchyId, rootEntityIds]) =>
+      this.deleteSubtrees(hierarchyId, [...rootEntityIds]),
+    );
+    await Promise.all(deleteTasks);
 
-    // run rebuild
+    // get the unique set of hierarchies to be rebuilt, then run the rebuild
+    const hierarchiesForRebuild = [...new Set(jobs.map(j => j.hierarchyId))];
```
kael89, line 92:
```suggestion
    const hierarchiesForRebuild = Object.keys(subtreesForDelete);
```

kael89 (approved):
Looks nice, just one optional suggestion but approving!

packages/database/src/cachers/EntityHierarchyCacher.js
edmofro: Hey, nice!

PR #1497, opened 2020-10-14
1297 laos eoc aggregated map overlays

IgorNadj:
### Changes:

- Add aggregate forecast map overlays
- Fix: data not being returned when aggregated
- Update measure builder to take data element from query
- Allow disable date picker in map overlays

---

### DIFF
--- a/packages/database/src/migrations/20201014220808-AddLaosEocAggregateForecastMapOverlays-modifies-data.js
+++ b/packages/database/src/migrations/20201014220808-AddLaosEocAggregateForecastMapOverlays-modifies-data.js
@@ -1,0 +1,129 @@
+'use strict';
+
+import { cloneDeep } from 'lodash'
+import { generateId, insertObject } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+
+const MAP_OVERLAY_GROUP_CODE = 'LA_EOC_Weather_Observations_And_Forecast';
+
+const commonMapOverlayConfig = {
+  userGroup: 'Laos EOC User',
+  dataElementCode: 'WTHR_FORECAST_PRECIP',
+  measureBuilderConfig: {
+    entityAggregation: {
+      dataSourceEntityType: 'city',
+    },
+  },
+  measureBuilder: 'sumAllPerOrgUnit',
+  presentationOptions: {
+    scaleType: 'neutral',
+    valueType: 'oneDecimalPlace',
+    displayType: 'spectrum',
+    measureLevel: 'City',
+    periodGranularity: 'day',
+    timePeriodEditable: false,
+  },
+  countryCodes: '{LA}',
+  projectCodes: '{laos_eoc}',
+};
+
+const FORECAST_MAP_OVERLAY_7_DAY = cloneDeep(commonMapOverlayConfig);
+FORECAST_MAP_OVERLAY_7_DAY.id = 'LA_EOC_Forecast_Rainfall_7_Day_Total';
+FORECAST_MAP_OVERLAY_7_DAY.name = 'Forecast rainfall - 7-day total (mm)';
+FORECAST_MAP_OVERLAY_7_DAY.presentationOptions = {
+  ...commonMapOverlayConfig.presentationOptions,
+  datePickerLimits: {
+    'start': { 'unit': 'day', 'offset': 0 },
+    'end': { 'unit': 'day', 'offset': 6 }, // including today is 7 days
+  },
+  defaultTimePeriod: {
+    'start': { 'unit': 'day', 'offset': 0 },
+    'end': { 'unit': 'day', 'offset': 6 }, // including today is 7 days
+  },
+}
+
+const FORECAST_MAP_OVERLAY_14_DAY = cloneDeep(commonMapOverlayConfig);
+FORECAST_MAP_OVERLAY_14_DAY.id = 'LA_EOC_Forecast_Rainfall_14_Day_Total';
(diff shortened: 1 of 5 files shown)


packages/database/src/migrations/20201014220808-AddLaosEocAggregateForecastMapOverlays-modifies-data.js
```diff
@@ -0,0 +1,129 @@
...
+    },
+  },
+  measureBuilder: 'sumAllPerOrgUnit',
+  presentationOptions: {
+    scaleType: 'neutral',
+    valueType: 'oneDecimalPlace',
+    displayType: 'spectrum',
+    measureLevel: 'City',
+    periodGranularity: 'day',
+    timePeriodEditable: false,
```
IgorNadj, line 38: This prevents the user from changing the dates

packages/web-frontend/src/containers/MeasureBar/index.js
```diff
@@ -149,10 +149,16 @@ export class MeasureBar extends Component {
       currentMeasure.datePickerLimits,
     );
 
+    const timePeriodEditable =
+      currentMeasure.timePeriodEditable !== undefined ? currentMeasure.timePeriodEditable : true; // defaults to true
```
kael89, line 153: This is equivalent to your code but a bit simpler to read:
```suggestion
  const { timePeriodEditable = true } = currentMeasure;
```

packages/web-config-server/src/apiV1/measureBuilders/DataPerOrgUnitBuilder.js
```diff
@@ -78,7 +78,10 @@ export class DataPerOrgUnitBuilder extends DataBuilder {
         throw new Error('The base data builder should return a single element array');
       }
 
-      return { organisationUnitCode, [dataElementCode]: data[0][dataElementCode] };
+      const value =
+        data[0][dataElementCode] !== undefined ? data[0][dataElementCode] : data[0].value; // key is sometimes changed to "value" by aggregator
```
kael89, line 82: This is an optional change: I believe it makes the `dataElementCode` vs `'value'` logic a bit clearer, but feel free to ignore it if you think otherwise!

```suggestion
      const [dataItem] = data;
      // key is sometimes changed to "value" by aggregator
      const value = dataElementCode in dataItem ? dataItem[dataElementCode] : dataItem.value;
```

Also, the fact that aggregator sometimes uses `'value'` and sometimes `dataElementCode` is a smell for an underlying problematic logic. However, this is probably out of scope for this task: according to my experience, consistent single and multi-value data handling is a bit hard without significant refactoring of existing data/measure builders.

packages/database/src/migrations/20201014220808-AddLaosEocAggregateForecastMapOverlays-modifies-data.js
```diff
@@ -0,0 +1,129 @@
...
+  },
+  countryCodes: '{LA}',
+  projectCodes: '{laos_eoc}',
+};
+
+const FORECAST_MAP_OVERLAY_7_DAY = cloneDeep(commonMapOverlayConfig);
```
kael89, line 44: I love `lodash.cloneDeep`! It may bit a bit heavy for front-end apps (since I believe it depends on multiple additional lodash modules), but for it's perfect for the back-end 👌

I can only imagine how useful it will be in other similar migrations!

kael89: I just realised using it will make other migrations much simpler!

packages/database/src/migrations/20201014220808-AddLaosEocAggregateForecastMapOverlays-modifies-data.js
```diff
@@ -0,0 +1,129 @@
...
+FORECAST_MAP_OVERLAY_7_DAY.id = 'LA_EOC_Forecast_Rainfall_7_Day_Total';
+FORECAST_MAP_OVERLAY_7_DAY.name = 'Forecast rainfall - 7-day total (mm)';
+FORECAST_MAP_OVERLAY_7_DAY.presentationOptions = {
+  ...commonMapOverlayConfig.presentationOptions,
+  datePickerLimits: {
+    'start': { 'unit': 'day', 'offset': 0 },
+    'end': { 'unit': 'day', 'offset': 6 }, // including today is 7 days
+  },
```
kael89, line 52: Nice and simple!

kael89 (changes requested):
Great work man! It's nice that this ended up being simple in the end. Just a few suggestions, some of them are optional

packages/web-frontend/src/containers/MeasureBar/index.js
IgorNadj: Ah, didn't know that one, nice. Updated.

packages/web-config-server/src/apiV1/measureBuilders/DataPerOrgUnitBuilder.js
IgorNadj: Yeah I saw some comments saying why it's needed, happy to leave for someone else further down the road.

Good thinking with the simplifying this part, updated.

packages/database/src/migrations/20201014220808-AddLaosEocAggregateForecastMapOverlays-modifies-data.js
kael89: Oh sorry man I forgot to comment this: could we also rename this variable to `canSelectDateRange` or similar?

I find the `is...`, `.are...`, `can...` prefixes very useful for boolean options/vars

kael89 (approved):
Just one additional change (sorry, somehow it got lost during my initial review!) but pre-approving since it's a simple one, to make your life easier 😃

packages/database/src/migrations/20201014220808-AddLaosEocAggregateForecastMapOverlays-modifies-data.js
IgorNadj: Good point! Updated

PR #1498, opened 2020-10-14
1420: Parameterise report e2e tests

kael89:
### Issue #:
[1420: Parameterise report e2e tests](https://github.com/beyondessential/tupaia-backlog/issues/1420) - only the first requirement (**organisation unit** parameters)

### Changes:

* Did a major refactor of the report snapshot capturing logic. Instead of using the UI to navigate to a report, we use a direct URL now.

  **Pros:**
  ✔ Much easier testing logic!
  ✔ Works for all types of reports, both expandable and not
  ✔The url can be used to identify a specific test setup, since it contains most information about a report's selected parameters. Example: `/covidau/AU/COVID-19?report=COVID_AU_Total_Cases_Each_State_Per_Day`

  **Cons:**
❌We loose the benefit of using a more realistic user path, which would be clicking around portal components to get to a report.

  It then occurred to me: the scope of those tests are not dashboard navigation, but data accuracy! We can add a new test suite for testing the report navigation. If we do, we'll probably do it once per report type and not once per the 300+ reports that we want to test here.

* Added specific logger levels and colors

---

### Screenshots:

`yarn cypress:generate-config` prints (using my slightly outdated local DB):

And creates the following file:

```jsonc
// dashboardReports.json

[
  "/covidau/AU/COVID-19?report=COVID_AU_Total_Cases_Each_State_Per_Day",
  "/covidau/AU/COVID-19?report=COVID_New_Cases_By_State",
  "/covidau/AU/COVID-19?report=COVID_Tests_Per_Capita",
  "/covidau/AU/COVID-19?report=COVID_Total_Cases_By_State",
  "/covidau/AU_Australian Capital Territory/COVID-19?report=COVID_Compose_Cumulative_Deaths_Vs_Cases",
  "/covidau/AU_Australian Capital Territory/COVID-19?report=COVID_Compose_Daily_Deaths_Vs_Cases",
   /* ...etc */
]
```

### DIFF
--- a/packages/web-frontend/cypress/config/orgUnitMap.json
+++ b/packages/web-frontend/cypress/config/orgUnitMap.json
@@ -1,0 +1,175 @@
+{
+  "covidau": {
+    "project": "AU"
+  },
+  "disaster": {
+    "project": "disaster"
+  },
+  "explore": {
+    "project": "explore"
+  },
+  "fanafana": {
+    "project": "TO"
+  },
+  "imms": {
+    "project": "imms"
+  },
+  "laos_eoc": {
+    "project": "LA"
+  },
+  "strive": {
+    "project": "PG"
+  },
+  "unfpa": {
+    "project": "unfpa"
+  },
+  "wish": {
+    "project": "FJ"
+  },
+  "AS": {
+    "country": "AS"
+  },
+  "AU": {
+    "country": "AU",
+    "district": "AU_Australian Capital Territory",
+    "sub_district": "AU_Australian Capital Territory_Unincorporated"
+  },
+  "CI": {
+    "country": "CI",
+    "district": "CI_Abidjan",
+    "facility": "CI_ABIDJAN01"
+  },
+  "CK": {
+    "country": "CK",
+    "district": "CK_Northern Islands",
+    "sub_district": "CK_Northern Islands_Manihiki",
+    "facility": "CK_aituH"
+  },
+  "DL": {
+    "country": "DL",
+    "district": "DL_North",
+    "sub_district": "DL_North_Hufflepuff",
+    "facility": "DL_1"
+  },
+  "DL_3": {
+    "facility": "DL_3"
+  },
+  "FJ": {
+    "country": "FJ",
+    "district": "FJ_Central",
+    "sub_district": "FJ_PROV_Ba",
+    "facility": "FJ_001",
+    "village": "FJ_DAM"
+  },
+  "FM": {
+    "country": "FM",
+    "district": "FM_Etten",
+    "facility": "FM_001"
+  },
+  "GU": {
+    "country": "GU"
+  },
+  "KH": {
+    "country": "KH",
+    "district": "KH_Batheay",
+    "facility": "KH_7 Makara_hc"
+  },
+  "KI": {
+    "country": "KI",
+    "district": "KI_Gilbert Islands",
+    "sub_district": "KI_Gilbert Islands_Abaiang",
+    "facility": "KI_ABRD14"
+  },
+  "LA": {
+    "country": "LA",
+    "district": "LA_Attapue",
+    "sub_district": "LA_Anouvong District",
+    "facility": "LA_ATPHO",
+    "village": "LA_vil_1001002",
+    "school": "LA_sch_1001001"
+  },
+  "MH": {
+    "country": "MH",
+    "district": "MH_Ailuk",
+    "facility": "MH_001"
+  },
+  "MP": {
+    "country": "MP"
+  },
+  "NC": {
+    "country": "NC"
+  },
+  "NR": {
+    "country": "NR",
+    "district": "NR_Denigomodu District",
+    "facility": "NR_NaoeroPHC"
+  },
+  "NU": {
+    "country": "NU"
+  },
+  "PF": {
+    "country": "PF"
+  },
+  "PG": {
+    "country": "PG",
+    "district": "PG_Madang",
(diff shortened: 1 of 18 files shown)


packages/web-frontend/cypress/config/orgUnitMap.json
```diff
@@ -0,0 +1,175 @@
+{
+  "covidau": {
```
kael89, line 2: _**Note:** the notes below apply only to configuration generated automatically using the `generateConfig.js` script_

A report may exist in multiple dashboard groups, which in turn may be defined for different organisation unit codes/levels and projects. Consider the following example:

1. We want to test the report `26`
2. This report exists under multiple dashboard groups, each one using a different org unit code: `CK`, `DL`, `FJ` etc. For each of those codes, the group is also available in multiple org unit levels: `country`, `district`, `sub_district` etc
3. Finally, each of those groups is defined in multiple projects `explore`, `disaster`, `imms` etc

❌One way around that would be submitting test data for each org unit that possibly meets a dashboard group's criteria. This way we will be sure that no matter which one is selected,  the rerport won't be empty. We are talking about thousands of org units 😱
✔Another way would be defining our preference for each possible combination of the above parameters. This is where `orgUnitMap.json` comes into play

For example, by defining
```json
{
  "TO": {
    "country": "TO",
    "district": "TO_Eua",
    "facility": "TO_CPMS",
    "village": "TO_Eua_Angaha"
  }
```

we know that if a dashboard group is selected that has `{ organisationUnitCode: 'TO', level: 'facility' }`, the report will be tested under `TO_CPMS`. Thus, submitting data for a single Tongan facility would be enough for testing groups with those specifications.

Our config generation script will select the first dashboard group whose specs meet an org unit map entry. If it cannot find any, the report will be skipped.

kael89: Another way around this problem would be changing the dashboard groups in the e2e test server to use DemoLand (or similar), where we just define a single/ a few org units for each level. But I intuitively think that for e2e tests we shouldn't interfere much with how the portal is actually set up.

scripts/node/utilities.js
```diff
@@ -7,27 +7,42 @@ const { exec } = require('child_process');
...
-const configureWinston = () =>
-  winston.configure({
+const LOG_LEVEL_CONFIG = {
+  error: 'red',
+  warn: 'yellow',
+  success: 'green',
+  info: 'cyan',
+  verbose: 'dim white',
+  debug: 'gray',
+};
```
kael89, line 17: Those are the same with the [default winston levels](https://github.com/winstonjs/winston#logging-levels) (=> npm levels), with the addition of `success` and the removal of `http`. The colors are a bit different to match common terminal script conventions.

packages/web-frontend/cypress/config/orgUnitMap.json
kael89: Finally, we may want to adjust the specific org unit map we use to make our test data submission even leaner (remove unused countries, specify priority of a country over another), but you get the idea.

edmofro: We could also add a generation script for this config file :-)

packages/web-frontend/cypress/scripts/generateDashboardReportConfig.js
```diff
@@ -3,60 +3,129 @@
...
+  NO_ORG_UNIT_MAP_ENTRY: 'noOrgUnitMapEntry',
+};
+
+const WARNING_TYPE_TO_MESSAGE = {
+  [WARNING_TYPES.NO_GROUP]: 'not attached to any dashboard group',
+  [WARNING_TYPES.NO_PROJECT]: 'not attached to any projects',
+  [WARNING_TYPES.NO_ORG_UNIT_MAP_ENTRY]: `no valid org unit map entry found in '${ORG_UNIT_MAP_PATH}'`,
+};
+
+const logWarningsForSkippedReports = (logger, skippedReportsByWarnType) => {
```
edmofro, line 25: This is a really nice way of outputting the errors!

packages/web-frontend/cypress/scripts/generateDashboardReportConfig.js
```diff
@@ -3,60 +3,129 @@
...
-const getDashboardReportIdToGroup = async database => {
+const selectUrlParams = dashboardGroups => {
+  for (const dashboardGroup of dashboardGroups) {
+    const {
+      organisationUnitCode: dashboardOrgUnitCode,
+      organisationLevel: dashboardLevel,
+    } = dashboardGroup;
+
+    const level = snake(dashboardLevel);
+    const orgUnitCode = orgUnitMap?.[dashboardOrgUnitCode]?.[level];
```
edmofro, line 54: Great use of `?.`

edmofro (approved):
All looks great to me!

packages/web-frontend/cypress/config/orgUnitMap.json
kael89: Definitely! I actually used a script locally (and made a few tweaks). I then thought we may want to set the org unit map "in stone", so that our testers now exactly which org units to use for submitting test data.

I suggest adjusting this behaviour after having a clearer idea of how test data will be submitted in practice, what do you think? We may even allow multiple automatically generated configs (random org unit, exhaustive report testing, etc)

edmofro: Yep I agree 👍

PR #1500, opened 2020-10-15
Invalidate soft cache on change to entity relation

edmofro:
The soft cache was being correctly invalidated when an entity was modified, but some of the cached functions also depend on ancestor_descendant_relation. This adds invalidation when entity_relation changes, so that the soft cache doesn't end up with a stale version of the hierarchy.

### DIFF
--- a/packages/database/src/DatabaseModel.js
+++ b/packages/database/src/DatabaseModel.js
@@ -31,13 +31,11 @@ export class DatabaseModel {
       });
 
       // if this model has caching that depends on other models, also add invalidation for them
-      if (this.cacheDependencies) {
-        this.cacheDependencies.forEach(databaseType => {
-          this.database.addChangeHandlerForCollection(databaseType, () => {
-            this.cache = {};
-          });
+      this.cacheDependencies.forEach(databaseType => {
+        this.database.addChangeHandlerForCollection(databaseType, () => {
+          this.cache = {};
         });
-      }
+      });
 
       // invalidate cached schema for this model on any change to db schema
       this.database.addSchemaChangeHandler(() => {
@@ -45,6 +43,11 @@ export class DatabaseModel {
         this.fieldNames = null;
       });
     }
+  }
+
+  // can be overridden by any subclass that needs cache invalidation when a related table changes
+  get cacheDependencies() {
+    return [];
   }
 
   // functionArguments should receive the 'arguments' object
--- a/packages/database/src/DatabaseModel.js
+++ b/packages/database/src/DatabaseModel.js
@@ -18,17 +18,29 @@ export class DatabaseModel {
     this.schema = null;
 
     this.cache = {};
+    this.cachedFunctionInvalidationCancellers = {};
 
     // If this model uses the singleton database, it is probably long running, so be sure to
     // invalidate the cache any time a change is detected. Non-singleton models are those created
     // during transactions, so are short lived and unlikely to need cache invalidation - thus we
     // avoid making an additional connection to pubsub and just leave their cache untouched
     if (this.database.isSingleton) {
+      // fully reset cache on any change to this model's records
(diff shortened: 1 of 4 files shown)


packages/database/src/migrations/20201009035426-AddImmutableTableTrigger-modifies-schema.js
```diff
@@ -31,7 +31,7 @@ exports.up = function (db) {
 
 exports.down = function (db) {
   return db.runSql(`
-    DROP FUNCTION IF EXISTS immutable_table();
+    DROP FUNCTION IF EXISTS immutable_table() CASCADE;
```
edmofro, line 34: Unrelated, just found when I was trying to migrate down to replay the migrations

packages/database/src/runPostMigration.js
```diff
@@ -21,14 +21,14 @@ const EXCLUDED_TABLES_FROM_TRIGGER_CREATION = [
 // attempted
 const IMMUTABLE_TABLES = ['ancestor_descendant_relation'];
 
-const SELECT_TABLES_WITHOUT_TRIGGERS = `
+const getSelectForTablesWithoutTriggers = triggerSuffix => `
```
edmofro, line 24: Changes in this file are unrelated, just found when I was trying to add both trigger types to ancestor_descendant_relation (ended up reverting that actual change)

packages/database/src/DatabaseModel.js
```diff
@@ -18,17 +18,29 @@ export class DatabaseModel {
...
     if (this.database.isSingleton) {
+      // fully reset cache on any change to this model's records
       this.database.addChangeHandlerForCollection(this.DatabaseTypeClass.databaseType, () => {
-        this.cache = {}; // invalidate cache on any change
+        this.cache = {};
       });
+
+      // if this model has caching that depends on other models, also add invalidation for them
+      if (this.cacheDependencies) {
+        this.cacheDependencies.forEach(databaseType => {
```
kael89, line 35: A minor suggestion, I see you used a getter for this property in `Entity.js`. Could we then define a base value for this property in `DatabaseModel.js`, i.e.

```js
  get cacheDependencies() {
    return [];
  }
```

or `this.cacheDependencies = [];`

In that case, we can avoid checking `if (this.cacheDependencies)`

packages/database/src/runPostMigration.js
```diff
@@ -21,14 +21,14 @@ const EXCLUDED_TABLES_FROM_TRIGGER_CREATION = [
...
   SELECT t.table_name
     FROM information_schema.tables t
     LEFT JOIN information_schema.table_privileges privileges
       ON t.table_name = privileges.table_name
     WHERE NOT EXISTS (
       SELECT * from pg_trigger p
-      WHERE p.tgname LIKE concat(lower(t.table_name), '%', '_trigger')
+      WHERE p.tgname = concat(lower(t.table_name), '${triggerSuffix}')
```
kael89, line 31: I guess that '%' was not required here? (Probably not, just double checking

kael89 (approved):
This is an important change done in a simple and elegant way, nice! Approving with comments

packages/database/src/runPostMigration.js
edmofro: Yeah no longer required. We were previously checking for `entity_trigger` and `entity_immutable_trigger` using a single `LIKE` comparison, with the `%` a wildcard for `_immutable`. Now we explicitly check for each

packages/database/src/DatabaseModel.js
edmofro: Good idea :-)

packages/database/src/runPostMigration.js
kael89: I see - thanks for the explanation! 🙌

PR #1502, opened 2020-10-15
1203-db-migration-delete-answer-with-9998

billli0:
### Issue #:
[Migrate 9998 to be no data](https://github.com/beyondessential/tupaia-backlog/issues/1203)

### Changes:
Delete all answers that have a value of 9998

### DIFF
--- a/packages/database/src/migrations/20201015015257-DeleteAnswerWith9998-modifies-data.js
+++ b/packages/database/src/migrations/20201015015257-DeleteAnswerWith9998-modifies-data.js
@@ -1,0 +1,29 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function(db) {
+  return db.runSql(`
+    DELETE FROM "answer" WHERE text = '9998';
+  `)
+};
+
+exports.down = function(db) {
+  return null;
+};
+
+exports._meta = {
+  "version": 1
+};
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20201015015257-DeleteAnswerWith9998-modifies-data.js
```diff
@@ -0,0 +1,29 @@
...
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function(db) {
+  return db.runSql(`
+    DELETE FROM "answer" WHERE text = '9998';
```
Alastair-L, line 19: My only concern here is that this could be too broad. 

I would feel a lot more comfortable if we also asserted that the survey was `Reproductive Health Combined`.
Something like: 
```suggestion
    DELETE FROM "answer" a
    using survey_response sr
    join survey s on s.id = sr.survey_id 
    where a.survey_response_id = sr.id 
    and a.text = '9998'
    and s.name = 'Reproductive Health Combined';
```

Alastair-L (changes requested):
Hey Biao! Looks good! Just a bit concerned with its current form because the issue is not system-wide (if we saw a 9998 in another project we wouldn't assume it to be no data straight away). And it's possible, however unlikely, that between now and this running in prod another 9998 response is submitted which we don't want to delete.

Do you agree?

packages/database/src/migrations/20201015015257-DeleteAnswerWith9998-modifies-data.js
billli0: Thanks Alastair, that's a good point. I haven't thought about it! This is also a nice query writing.

Alastair-L (approved):
LGTM!

Also tagging @kael89 (sorry) to just double check that this is what he had in mind as well.

kael89 (approved):
Looks good to me too! Doubly approved then 🙂

PR #1484, opened 2020-10-13
Entity hierarchy performance improvements

edmofro:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/478

Replacing https://github.com/beyondessential/tupaia/pull/1248 due to long branch name

PR #1504, opened 2020-10-15
1293 fix survey issue with displaying surveys that use new child permissions

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1293

Specifically requesting @edmofro 's review since we worked together on this
- Interestingly, the line `this.models.permissionGroup.addChangeHandler(() => this.resetCaches());` only seems to work when written exactly like that (which is how you wrote it originally but when we pair programmed you also refined that line to `this.models.permissionGroup.addChangeHandler(this.resetCaches);`, which didn't 100% work)

I also requested @edmofro 's review because this will also probably have conflicts with his `entity-hierarchy-cache-batch-deletes` branch (as they both make changes to the same file) - I'm not sure if this affects testing / releasing this branch at all


edmofro (approved):

packages/auth/src/AccessPolicyBuilder.js
```diff
@@ -14,6 +14,10 @@ export class AccessPolicyBuilder {
     this.setupCacheInvalidation();
   }
 
+  resetCaches() {
```
edmofro, line 17: @avaek this will now need to be updated from dev and conflicts resolved, since I have merged https://github.com/beyondessential/tupaia/pull/1484

PR #1505, opened 2020-10-15
Revert "1203-db-migration-delete-answer-with-9998"

billli0:
Reverts beyondessential/tupaia#1502 because it needs to be tested.


Alastair-L (approved):
Approving, including merging to dev without testing after Kostas gives approval.

kael89:
No need to reverse this if we can test it directly in dev. I suspect this could be possible - I'll check with the testers

PR #1506, opened 2020-10-16
Permissions convert gets to sql

EMcQ-BES:
### Changes:

Converting the following GET endpoints to use pure SQL when filtering:
* dashboardGroups
* dashboardReports
* mapOverlays
* surveyGroups

Updated the unit tests to run directly against the GET endpoints using TestableApp


packages/meditrak-server/src/tests/routes/GETDashboardGroups/assertDashboardGroupsPermissions.test.js
```diff
@@ -132,214 +130,98 @@ describe('Permissions checker for GETDashboardGroups', async () => {
...
 
-    it('Sufficient permissions: Should return project level dashboard groups that users have access to any of their child countries', async () => {
-      const accessPolicy = new AccessPolicy(DEFAULT_POLICY);
-      const results = await filterDashboardGroupsByPermissions(accessPolicy, models, [
-        nationalDashboardGroup2,
-        projectLevelDashboardGroup,
-      ]);
+  describe('GET /dashboardGroups/:id', async () => {
+    it('Sufficient permissions: Should return a requested dashboard group if users have access to the country of the dashboard group', async () => {
+      await prepareStubAndAuthenticate(app, DEFAULT_POLICY);
```
edmofro, line 139: Awesome work figuring this out!

packages/meditrak-server/src/tests/routes/GETDashboardGroups/assertDashboardGroupsPermissions.test.js
```diff
@@ -132,214 +130,98 @@ describe('Permissions checker for GETDashboardGroups', async () => {
...
+        districtDashboardGroup1.id,
+        nationalDashboardGroup1.id,
+        nationalDashboardGroup2.id,
+        projectLevelDashboardGroup.id,
+      ]);
     });
 
-    it('Insufficient permissions: Should throw an exception if dashboard groups are project level and users do not have access to any of their child countries', async () => {
-      //Remove Admin permission of TO, SB, KI, VU to have insufficient permissions to access the project level dashboard groups.
+    it('Insufficient permissions: Should throw an exception if we have permissions for no dashboard groups', async () => {
```
edmofro, line 217: I'm starting to question this as a general policy. What if the reason there are no results is that we specified an invalid search term, rather than a permissions issue? I guess here it does work, but we should either apply that on a case by case basis, or return an empty array if we want a general rule across endpoints. 

What do you think @EMcQ-BES?

packages/meditrak-server/src/routes/GETMapOverlays/assertMapOverlaysPermissions.js
```diff
@@ -2,72 +2,64 @@
...
+
+  // Generate lists of country codes we have access to per permission group
+  allPermissionGroups.forEach(pg => {
+    countryCodesByPermissionGroup[pg] = accessPolicy.getEntitiesAllowed(pg);
+  });
 
-  return true;
+  // Look up the json country codes by userGroup, compare them to the countryCodes from the mapOverlay
+  dbConditions[RAW] = {
+    sql: `("countryCodes" && array(select trim('"' from json_array_elements(?::json->"userGroup")::text)))
```
edmofro, line 48: SQL is impressive once again! Could you add a bit of whitespace to ease the reading experience?

packages/meditrak-server/src/routes/GETSurveyGroups/assertSurveyGroupsPermissions.js
```diff
@@ -38,16 +41,41 @@ export const filterSurveyGroupsByPermissions = async (accessPolicy, models, surv
...
+export const createSurveyGroupDBFilter = async (accessPolicy, models, criteria) => {
+  if (hasBESAdminAccess(accessPolicy)) {
+    return criteria;
+  }
+  const dbConditions = criteria;
+  const allPermissionGroupsNames = accessPolicy.getPermissionGroups();
+  const countryIdsByPermissionGroupId = {};
+
+  // Generate lists of country ids we have access to per permission group id
+  for (const pg of allPermissionGroupsNames) {
```
edmofro, line 66:
```suggestion
  for (const permissionGroup of allPermissionGroupsNames) {
```

packages/meditrak-server/src/routes/GETSurveyGroups/assertSurveyGroupsPermissions.js
```diff
@@ -38,16 +41,41 @@ export const filterSurveyGroupsByPermissions = async (accessPolicy, models, surv
...
+    if (permissionGroup) {
+      const countryNames = accessPolicy.getEntitiesAllowed(pg);
+      const countryList = await models.country.find({ code: countryNames });
+      countryIdsByPermissionGroupId[permissionGroup.id] = countryList.map(e => e.id);
+    }
+  }
+
+  // Check we have permissions to at least one of the surveys in this group
+  dbConditions[RAW] = {
+    sql: `(select count(*) from survey where survey.survey_group_id = survey_group.id and country_ids && array(select trim('"' from json_array_elements(?::json->permission_group_id)::text))) > 0`,
```
edmofro, line 77: Again just padding it out with indentation and new lines

Have you done any performance comparisons against dev at all? I wonder whether adding a select for every row adds much of a hit.

edmofro (changes requested):
This stuff takes a load of thinking, amazing work getting through all that dense sql.

Very minor note: we use hyphens rather than underscores in branch names

packages/meditrak-server/src/routes/GETDashboardGroups/assertDashboardGroupsPermissions.js
```diff
@@ -2,59 +2,60 @@
...
-
-  if (filteredDashboardGroups.length !== dashboardGroups.length) {
-    throw new Error('You do not have permissions for the requested dashboard group(s)');
+export const createDashboardGroupDBFilter = async (accessPolicy, models, criteria) => {
+  if (hasBESAdminAccess(accessPolicy)) {
+    return criteria;
   }
-
-  return true;
+  const dbConditions = criteria;
```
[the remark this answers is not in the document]
edmofro: This applies in several files

packages/meditrak-server/src/routes/GETMapOverlays/assertMapOverlaysPermissions.js
```diff
@@ -2,72 +2,64 @@
...
-        if (hasAccessToMapOverlay) {
-          break;
-        }
-      }
+export const assertMapOverlaysPermissions = async (accessPolicy, models, mapOverlayId) => {
+  const mapOverlay = await models.mapOverlay.findById(mapOverlayId);
 
-      accessCache[cacheKey] = hasAccessToMapOverlay;
-    }
+  const countryCodes = mapOverlay.countryCodes || [];
```
edmofro, line 14: Man `countryCodes` is a poorly named column now huh...could you add a card for renaming to reflect the fact that it also contains project codes?

edmofro: "now" being since the introduction of project level map overlays - nothing to do with your work.

packages/meditrak-server/src/tests/routes/GETDashboardGroups/assertDashboardGroupsPermissions.test.js
```diff
@@ -132,214 +130,98 @@ describe('Permissions checker for GETDashboardGroups', async () => {
...
+      await prepareStubAndAuthenticate(app, policy);
+      const { body: results } = await app.get(`dashboardGroups`);
 
-      expect(() =>
-        assertDashboardGroupsPermissions(accessPolicy, models, [
-          facilityDashboardGroup1,
-          nationalDashboardGroup2,
-        ]),
-      ).to.throw;
+      expect(results.map(r => r.id)).to.deep.equal([
```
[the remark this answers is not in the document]
EMcQ-BES: > Test for presence and absence of dummy data, rather than full array match

I think that's just swapping one assumption about the database state for another. Since this is running as a full request, it includes pagination so we couldn't guarantee that the test data would exist in the first X (100?) entries.

I'll take a look into alternatives.

packages/meditrak-server/src/tests/routes/GETDashboardGroups/assertDashboardGroupsPermissions.test.js
EMcQ-BES: I guess this made more sense when we were running these as filters since we could know for certain the permissions step caused the empty array. I think returning an empty array is fair.

Do we still want to throw an error when requesting a single record we don't have permission for? We can tell for certain on a single record that it's a permissions error, but we may prefer to keep consistency between those responses.

edmofro: Let's keep the error on single record. They've specified an id, so it's quite clear that they're trying to do something they shouldn't be able to, as you say.

packages/meditrak-server/src/routes/GETMapOverlays/assertMapOverlaysPermissions.js
EMcQ-BES: https://github.com/beyondessential/tupaia-backlog/issues/1489

packages/meditrak-server/src/tests/routes/GETDashboardGroups/assertDashboardGroupsPermissions.test.js
```diff
@@ -34,15 +31,16 @@ describe('Permissions checker for GETDashboardGroups', async () => {
...
-  const models = getModels();
+  const app = new TestableApp();
+  const { models } = app;
   let facilityDashboardGroup1;
   let districtDashboardGroup1;
   let nationalDashboardGroup1;
   let nationalDashboardGroup2;
   let projectLevelDashboardGroup;
 
   before(async () => {
```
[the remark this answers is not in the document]
EMcQ-BES: I've double checked and all this data does appear to get cleared out automatically.

packages/meditrak-server/src/tests/routes/GETDashboardGroups/assertDashboardGroupsPermissions.test.js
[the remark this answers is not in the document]
EMcQ-BES: I've updated these tests with the addition of filters to only request the test data. We discussed in person the need to add some unit tests for that filter function so I've written a separate ticket for that work: https://github.com/beyondessential/tupaia-backlog/issues/1490

edmofro (approved):
Fantastic stuff :-)

PR #1507, opened 2020-10-16
1399 remove hello from country modal

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1399

Basically when you create a new country the modal says "Hello" (the default EditModal.js message) and this change fixes that happening here


edmofro (approved):
Noice!

PR #1508, opened 2020-10-16
fix @tupaia/database test fail

billli0:
### Issue #:
[Fix flaky test (@tupaia/database)](https://github.com/beyondessential/tupaia-backlog/issues/1424)

### Changes:
Change Promise.all to forEach
Add jestTimeout in @tupaia/auth and @tupaia/data-api because it didn't pass occasionally
Also remove a reduntant jest.mock

### DIFF
--- a/packages/aggregator/src/__tests__/createAggregator.test.js
+++ b/packages/aggregator/src/__tests__/createAggregator.test.js
@@ -2,8 +2,6 @@ import { DataBroker } from '@tupaia/data-broker';
 import { Aggregator } from '../Aggregator';
 import { createAggregator } from '../createAggregator';
 
-jest.mock('@tupaia/data-broker');
-
 const CustomAggregatorClass = class CustomAggregator {
   constructor(dataBroker) {
     this.dataBroker = dataBroker;
--- a/packages/aggregator/src/__tests__/createAggregator.test.js
+++ b/packages/aggregator/src/__tests__/createAggregator.test.js
@@ -2,6 +2,8 @@ import { DataBroker } from '@tupaia/data-broker';
 import { Aggregator } from '../Aggregator';
 import { createAggregator } from '../createAggregator';
 
+jest.mock('@tupaia/data-broker');
+
 const CustomAggregatorClass = class CustomAggregator {
   constructor(dataBroker) {
     this.dataBroker = dataBroker;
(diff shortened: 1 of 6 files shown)


packages/auth/jest.config.json
```diff
@@ -4,5 +4,6 @@
   "testMatch": ["<rootDir>/src/__tests__/**/**.test.js"],
   "setupFilesAfterEnv": ["../../jestScaffolding.js", "./jest.setup.js"],
   "transformIgnorePatterns": ["/node_modules/"],
-  "testEnvironment": "node"
+  "testEnvironment": "node",
+  "testTimeout": 400000
```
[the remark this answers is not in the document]
billli0: I believe so. I've seen once the `auth` failed in codeship in someone's test. Maybe we can make it lower. I've tried 50000 but it failed in my local so I give it one more '0'.

packages/aggregator/src/__tests__/createAggregator.test.js
```diff
@@ -2,8 +2,6 @@ import { DataBroker } from '@tupaia/data-broker';
 import { Aggregator } from '../Aggregator';
 import { createAggregator } from '../createAggregator';
 
-jest.mock('@tupaia/data-broker');
```
kael89, line 5: I tried removing this locally and I got database connection errors... Which makes sense since `new DataBroker` calls `getModels`, which creates a new DB connection.

By the way, I don't have a `.env` file for `aggregator` locally, but it shouldn't be required.

billli0: That's weird. I don't have `.env` file for aggregator as well but test pass without this mock. I guess it uses the `.env` file in `database` directly.

packages/auth/jest.config.json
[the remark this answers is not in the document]
kael89: Unit tests should run fairly quickly (if not, people will run them less often😛).

Do you remember which test case it was? I would be tempted to try making the tests quicker, rather than increasing the test timeout threshold. We can utilize mocking for this, or breaking the test into different cases

packages/aggregator/src/__tests__/createAggregator.test.js
kael89: I don't think so, I also have a .`env` in my database packages... 🤷‍♀️

kael89 (approved):
Approved, no testing required for this since it's dev related, so you can merge straight into dev

PR #1359, opened 2020-09-21
1271 utils use jest framwork only

billli0:
### Issue #:
[1271- Use a single test framework : **Utils module only**](https://github.com/beyondessential/tupaia-backlog/issues/1271)

This branch is a draft PR for tracking the changes.

PR #1510, opened 2020-10-18
982: UNFPA: Addition of functionality to MediTrak - in survey analytics

chris-bes:
### Changes:

Some notes on development:

A new question type is added called CalculatedQuestion, which supports 2 types of calculations:
- Arithmetic (required for this task, see the task description). Eg: (QuestionA + QuestionB + QuestionC) / 3
- Condition (required for #1359). Eg: if  (QuestionA + QuestionB + QuestionC) >= 3 then the generated answer is Yes, otherwise No.

For calculate **arithmetic**, the config cell in the excel should have:

1. `type: arithmetic` (mandatory)
Type of the Calculated Question
2. `formula: SchCVD016a + SchCVD016b + SchCVD016c` (mandatory)
Formula for calculating arithmetic
3. `valueTranslation: SchCVD016a.Yes=3,SchCVD016a.No=0,SchCVD016b.Yes=2,SchCVD016b.No=0,SchCVD016c.Yes=1,SchCVD016c.No=0
` (optional)
An intepretation of the answers to the appropriate scores which will be used in the formula for calculating arithmetic. If the questions used for calculating arithmetic are pure numeric and have no different weighting, then we don't need `valueTranslation`
4. `defaultValues: SchCVD024=0,SchCVD007=0,SchFF004=0` (optional)
Default values if there is no answer for a question. If `defaultValues` is not specified, the overall default will be 0.
5. `text: Section score = $result` (optional, if not provided, it will show just the arithmetic result)
This is to show a representative text for the Calculated Question (if the question is not hidden). Eg: 'Section score = 23' (23 is the arithmetic result, so the text in the config should be `Section score = $result`).

*Note: If you want to show more complex text like Numerator / Denominator = Result (similar to the mock up), you will need to have intermediate Calculated Questions in the survey.
Eg: If your formula is: ((code1 + code2 + code3) / (code4 + code5 + code6)) * 100
And you want to show Numerator / Denominator = Result.
You will need:
- An intermediate Calculated Question for calculating Numerator (code: NumeratorCode, formula: code1 + code2 + code3) 
- An intermediate Calculated Question for calculating Denominator (code: DenominatorCode , formula: code4 + code5 + code6).
(The intermediate questions can be hidden). 
- The config for the final Calculated Question will be: 
`formula: (NumeratorCode / DenominatorCode) * 100`
`text: NumeratorCode / DenominatorCode = $result`.

The translated config for the question in the database (all question codes are replaced by question ids) will be:
```
{
   "calculated":{
      "type":"arithmetic",
      "formula":"5ed7558b61f76a6ba5003114 + 5ecb422561f76a44ae0398eb + 5ec5bf8961f76a18ad00d4f9",
      "defaultValues":{
         "5ed7558b61f76a6ba5003114":"0",
         "5ecb422561f76a44ae0398eb":"0",
         "5ec5bf8961f76a18ad00d4f9":"0"
      },
      "valueTranslation":{
         "5ed7558b61f76a6ba5003114":{
            "Yes":"3",
            "No":"0"
         },
         "5ecb422561f76a44ae0398eb":{
            "Yes":"2",
            "No":"0"
         },
         "5ec5bf8961f76a18ad00d4f9":{
            "Yes":"1",
            "No":"0"
         }
      }
   }
}
```
For calculating **condition** (see use case in #1359), the config cell in the excel should have:
1. `type: condition` (mandatory)
Type of the Calculated Question
2. `conditions: Yes.RHS6UNFPA1353Calc.operator=GE,Yes.RHS6UNFPA1353Calc.operand=3,No.RHS6UNFPA1353Calc.operator=LT,No.RHS6UNFPA1353Calc.operand=3`
operator can be GE(>=), GT(>), LE(<=),LT(<),... or anything else we want to add

Translated config in the database will be:
```
{
   "calculated":{
      "type":"condition",
      "conditions":{
         "Yes":{
            "5f8ae5415da1e604eb0009ea":{
               "operator":">=",
               "operand":"3"
            }
         },
         "No":{
            "5f8ae5415da1e604eb0009ea":{
               "operator":"<",
               "operand":"3"
            }
         }
      }
   }
}
```
Things that are out of scope at the moment (discussed with Edwin):
How should we handle when the survey scores are updated?
- If all the calculated answers are submitted (this is the case at the moment), and we eventually update the survey scores, we can use Question Hooks to re calculate the existing answers.
- If the calculated answers are not submitted, we can somehow share the config between the Calculated Question and Indicator to create visualisation.


packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -3,8 +3,17 @@
...
+import { getVariables, runArithmetic } from '@beyondessential/arithmetic';
 import { checkAnswerPreconditionsAreMet } from './helpers';
 
+const CONDITION_TYPE = {
+  '=': (value, filterValue) => value === filterValue,
+  '>': (value, filterValue) => value > filterValue,
+  '<': (value, filterValue) => value < filterValue,
+  '>=': (value, filterValue) => value >= filterValue,
+  '<=': (value, filterValue) => value <= filterValue,
+};
```
edmofro, line 15: @kael89 do you know where the `analyticCount` indicator development is up to? We were thinking of having a common set of comparison functions that parsed a string and returned a boolean result, in some published package right?

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -109,3 +118,80 @@ export const getQuestion = (state, questionId) => state.assessment.questions[que
...
+      const checkConditionMethod = CONDITION_TYPE[operator];
+      return checkConditionMethod ? checkConditionMethod(answer, operand) : false;
+    });
+  });
+
+  let calculatedResult;
+
+  if (result) {
+    [calculatedResult] = result;
+  }
```
edmofro, line 179: I find the part where you have to extract the key a bit awkward, can we change to this:
```suggestion
  const checkAllConditionsMet = conditionsToCheck => 
    Object.entries(conditionsToCheck).every(([questionId, { operator, operand }]) => {
      const answer = getAnswerForQuestion(state, questionId);

      if (answer === undefined) {
        return false;
      }

      const checkConditionMethod = CONDITION_TYPE[operator];
      return checkConditionMethod ? checkConditionMethod(answer, operand) : false;
    };
  const calculatedResult = Object.keys(conditions).find(displayValue => checkAllConditionsMet(conditions[displayValue]));
```

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -109,3 +118,80 @@ export const getQuestion = (state, questionId) => state.assessment.questions[que
...
+  const result = Object.entries(conditions).find(([displayValue, valueConditions]) => {
+    return Object.entries(valueConditions).every(([questionId, { operator, operand }]) => {
+      const answer = getAnswerForQuestion(state, questionId);
+
+      if (answer === undefined) {
+        return false;
+      }
+
+      const checkConditionMethod = CONDITION_TYPE[operator];
+      return checkConditionMethod ? checkConditionMethod(answer, operand) : false;
```
edmofro, line 171: Perhaps we should also throw an error telling the user to update the app if the operator isn't supported in the current version

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -109,3 +118,80 @@ export const getQuestion = (state, questionId) => state.assessment.questions[que
...
 export const getAnswerForQuestion = (state, questionId) => getAnswers(state)[questionId];
+
+const getArithmeticResult = (state, config) => {
+  const { formula, valueTranslation, defaultValues = {}, text = '' } = config.calculated;
+  const values = {};
+  const variables = getVariables(formula);
+  let translatedText = text;
+
+  variables.forEach(questionIdVariable => {
+    let answer = getAnswerForQuestion(state, questionIdVariable) || 0;
```
edmofro, line 129:
```suggestion
    let answer = getAnswerForQuestion(state, questionIdVariable);
```
You've got logic below that relies on `undefined` being left as is

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -109,3 +118,80 @@ export const getQuestion = (state, questionId) => state.assessment.questions[que
...
+      valueTranslation[questionIdVariable][answer] !== undefined
+    ) {
+      answer = valueTranslation[questionIdVariable][answer];
+    }
+
+    if (answer === undefined) {
+      answer = defaultValues[questionIdVariable];
+    }
+
+    values[questionIdVariable] = answer || 0;
```
edmofro, line 143:
```suggestion
    values[questionIdVariable] = answer !== undefined ? answer : 0;
```
In case we want the answer to be `false`...which seems unlikely in an arithmetic question now I think about it

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -109,3 +118,80 @@ export const getQuestion = (state, questionId) => state.assessment.questions[que
...
+
+    translatedText = translatedText.replace(questionIdVariable, answer);
```
edmofro, line 145: Can we do this in a separate loop please? Doing the 4 units of work 50 times is the same as doing 2 units of work 50 times plus another 2 units of work 50 times - i.e. splitting the contents of an iteration apart doesn't really slow things down (there's a tiny bit of overhead setting up the loop, but that's negligible). I think in this case the logic is distinct enough to justify a separate block. I'd also try to aim for `const` rather than `let` where possible (e.g. by making use of a function that contains the conditional logic with returns)

packages/meditrak-app/app/assessment/specificQuestions/CalculatedQuestion.js
```diff
@@ -0,0 +1,63 @@
...
+
+  updateAnswer(props) {
+    const { answer, onChangeAnswer, calculatedResult } = props;
+    if (answer !== calculatedResult) {
+      onChangeAnswer(calculatedResult);
+    }
+  }
+
+  render() {
+    const { translatedText, answer } = this.props;
```
edmofro, line 26:
```suggestion
    const { answerDisplayValue, answer } = this.props;
```
Particularly within the presentational component, I think we can be more specific with the name of this variable

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processCalculatedConfig.js
```diff
@@ -0,0 +1,195 @@
...
+  GE: '>=',
+  LT: '<',
+  LE: '<=',
+  EQ: '=',
+};
+
+export const processCalculatedConfig = async (models, config) => {
+  const calculatedConfig = await processCalculatedConfigByType(models, config);
+  return {
+    type: config.type,
```
edmofro, line 20:
```suggestion
export const processCalculatedConfig = async (models, { type, ...config }) => {
  const calculatedConfig = await processCalculatedConfigByType(models, type, config);
  return {
    type,
```

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processCalculatedConfig.js
```diff
@@ -0,0 +1,195 @@
...
+    ...calculatedConfig,
+  };
+};
+
+const processCalculatedConfigByType = async (models, config) => {
+  switch (config.type) {
```
edmofro, line 26:
```suggestion
const processCalculatedConfigByType = async (models, type, config) => {
  switch (type) {
```

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processCalculatedConfig.js
```diff
@@ -0,0 +1,195 @@
...
+    case 'arithmetic':
+      return translateArithmeticConfig(models, config);
+    case 'condition':
+      return translateConditionConfig(models, config);
+    default:
+      throw new Error(`Invalid calculated type: ${config.type}`);
+  }
+};
+
+const translateArithmeticConfig = async (models, config) => {
```
edmofro, line 36:
```suggestion
const translateArithmeticConfig = async (models, config) => {
  const { formula, defaultValues, valueTranslation, text } = config;
```

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processCalculatedConfig.js
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  if (config.valueTranslation) {
+    translatedConfig = {
+      ...translatedConfig,
+      valueTranslation: await translateValueTranslation(models, config.valueTranslation),
+    };
+  }
+
+  if (config.text) {
```
edmofro, line 57:
```suggestion
  if (config.answerDisplayText) {
```
Or `answerLabel`. I worry that `text` gets confused with the `question.text`

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processCalculatedConfig.js
```diff
@@ -0,0 +1,195 @@
...
+ * will be converted to:
+ * formula: (5ed7558b61f76a6ba5003114 / 5ec5bf8961f76a18ad00d4f9) * 100
+ * @param {*} models
+ * @param {*} formula
+ * @param {*} codes
+ */
+const translateFormula = async (models, formula, codes) => {
+  let translatedFormula = formula;
+
+  for (let i = 0; i < codes.length; i++) {
```
edmofro, line 78:
```suggestion
  for (const code of codes) {
```
We now allow this in our eslint rules :-)

Can do this in a few places if you like. Nothing wrong with using an index though!

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processCalculatedConfig.js
```diff
@@ -0,0 +1,195 @@
...
+};
+
+const translateConditionConfig = async (models, config) => {
+  return {
+    conditions: await translateConditions(models, config.conditions),
+  };
+};
+
+/**
+ * conditions: Yes.SchCVD016a.operator=GE,Yes.SchCVD016a.operand=5,No.SchCVD016a.operator=LT,No.SchCVD016a.operand=5
```
edmofro, line 163: I wonder if we should try to align this with the conditional formula text being developed for the analyticCount indicator @kael89

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processCalculatedConfig.js
```diff
@@ -0,0 +1,195 @@
...
+    }
+    if (!translatedConditions[finalValue][questionId]) {
+      translatedConditions[finalValue][questionId] = {};
+    }
+    translatedConditions[finalValue][questionId][conditionKey] =
+      conditionKey === 'operator' ? OPERATOR_TRANSLATIONS[targetValue] : targetValue;
+  }
+
+  return translatedConditions;
+};
```
edmofro, line 195: This file is super clearly put together - nice work!

packages/meditrak-server/src/routes/importSurveys/Validator/BaseValidator.js
```diff
@@ -45,4 +45,18 @@ export class BaseValidator {
       return true;
     };
   }
+
+  assertPointingToPrecedingQuestion(
```
edmofro, line 49: Thanks for making this a util

edmofro (changes requested):
Very nice work @chris-bes - this has a lot of moving parts and you've coordinated them all really nicely! I've gone in with some quite specific comments, given that overall it's looking beautiful :-)

packages/meditrak-app/app/assessment/selectors.js
kael89: That's correct, we also thought of using math.js for empowering that behaviour in indicators. Options:

A. Publish a package that exposes a common set of comparison options with custom code
B. Publish a package that exposes a common set of comparison options, wrapping math.js
C. Do not publish a package but use math.js directly

Selecting either A, B or C will most likely depend on the quality and functionality of math.js (which I am unaware of)

And then
i. Refactor data builders to use either the published package/math.js directly, or
ii. Leave this to be done in the reporting package work

edmofro: @chris-bes could you take a look at math.js and see what you think? With the aim of dropping it into this PR wherever @tupaia/arithmetic is currently used, as well as where the condition checking is done

packages/meditrak-server/src/routes/importSurveys/Validator/JsonFieldValidator.js
```diff
@@ -20,6 +18,12 @@ export class JsonFieldValidator extends BaseValidator {
     return true;
   }
 
+  getConfig(rowIndex) {
+    const fieldName = this.getFieldName();
+    const field = this.getField(rowIndex, fieldName);
+    return this.parseField(field);
+  }
+
```
chris-bes, line 26: I added this function (so that ConfigValidator can have access to the whole config in case a property is dependent on another property and want to validate something between that) but ended up not using it. However I still keep it cause I think it can be useful later on

packages/meditrak-server/src/routes/utilities/excel.js
```diff
@@ -32,6 +32,13 @@ export const splitStringOn = (string, splitCharacter) => {
   return string ? string.split(splitCharacter).map(segment => segment.trim()) : [];
 };
 
+export const splitStringOnFirstOccurrence = (string, splitCharacter) => {
+  let [first, ...second] = string ? string.split(splitCharacter) : [];
+  second = second.join(splitCharacter);
+  const result = [first, second];
+  return result.map(segment => segment.trim());
+};
+
```
chris-bes, line 41: Added this function and also use it in `convertCellToJson` so that `:` can be used in the value as well (I had to use `=` before). 
Eg: value: a:2,b:3,c:4

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/ConfigValidator.js
```diff
@@ -32,8 +33,10 @@ export class ConfigValidator extends BaseValidator {
         return EntityConfigValidator;
       case CODE_GENERATOR:
         return CodeGeneratorConfigValidator;
-      case CALCULATED:
-        return CalculatedConfigValidator;
+      case ARITHMETIC:
+        return ArithmeticConfigValidator;
+      case CONDITION:
+        return ConditionConfigValidator;
```
chris-bes, line 39: The logic of Calculated Question is getting a bit complex and I felt that there were a lot of switch cases to process things differently between `arithmetic` and `condition` type (and there's little common between them). So I decided to split CalculatedQuestion into ArithmeticQuestion and ConditionQuestion (which I remember you did suggest before).

edmofro:
@chris-bes is this ready for re-review?

chris-bes:
@edmofro yep, forgot to press the button 😓

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -109,3 +110,85 @@ export const getQuestion = (state, questionId) => state.assessment.questions[que
...
+    const answer = getArithmeticAnswer(questionId);
+    translatedAnswerDisplayText = translatedAnswerDisplayText.replace(questionId, answer);
+  });
+
+  expressionParser.setScope(values);
+
+  const result = !isNaN(expressionParser.evaluate(formula))
+    ? Math.round(expressionParser.evaluate(formula) * 1000) / 1000 // Round to 3 decimal places
+    : 0;
+  translatedAnswerDisplayText = translatedAnswerDisplayText.replace('$result', result);
```
edmofro, line 158: I'd like to try to group the logical concepts together more, so rather than setting up the display text and then translating it in a couple of places, a single concept is handled in a single place. In this instance, you could either pull all of your code for the display text down here just as it is, or even change it up a bit as follows
```suggestion
  // could be inline, or could be a util
  const replaceVariablesInText = (rawText, variables) => {
    let text = rawText;
    Object.entries(variables).forEach((variableName, value) => {
      text = text.replace(variableName, result);
    });
    return text;
  });
  const translatedAnswerDisplayText = replaceVariablesInText(answerDisplayText, { ...values, $result: result });
```

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -109,3 +110,85 @@ export const getQuestion = (state, questionId) => state.assessment.questions[que
...
+  } = config;
+  const expressionParser = new BooleanExpressionParser();
+  const checkConditionMet = ({ formula, defaultValues }) => {
+    const values = {};
+    const variables = expressionParser.getVariables(formula);
+
+    variables.forEach(questionIdVariable => {
+      const questionId = questionIdVariable.replace(/^\$/, ''); // Remove the first $ prefix
+      const answer = getAnswerForQuestion(state, questionId);
+      const value = answer || defaultValues[questionId];
```
edmofro, line 179: What if it had been answered "false" but the default value was "true"?

Oh, I guess we use strings rather than booleans, so even "false" will be "truthy"

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -109,3 +110,85 @@ export const getQuestion = (state, questionId) => state.assessment.questions[que
...
+      values[questionIdVariable] = value;
+    });
+
+    expressionParser.setScope(values);
+    return expressionParser.evaluate(formula);
+  };
+
+  const result = Object.keys(conditions).find(displayValue =>
```
edmofro, line 187: Are the keys of the conditions really a `displayValue`? I would have thought they were a question id...

That aside, I think we can use `Object.values` here

packages/meditrak-server/src/routes/utilities/translateExpression.js
```diff
@@ -0,0 +1,28 @@
...
+ * ($5ed7558b61f76a6ba5003114 / $5ec5bf8961f76a18ad00d4f9) * 100
+ * @param {*} models
+ * @param {*} formula
+ * @param {*} codes
+ */
+export const translateExpression = async (models, formula, codes) => {
+  let translatedFormula = formula;
+
+  for (const code of codes) {
+    const { id: questionId } = await models.question.findOne({ code });
```
edmofro, line 22: We've found this more efficient in the past (reducing number of individual db calls, compute in memory is cheap)

```suggestion

  const questions = await models.question.find({ code: codes });
  const questionCodeToId = reduceToDictionary(questions, 'code', 'id');
  for (const code of codes) {
    const questionId = questionCodeToId[code];
```

packages/meditrak-app/app/assessment/specificQuestions/CalculatedQuestion.js
```diff
@@ -0,0 +1,72 @@
+import React, { PureComponent } from 'react';
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+import { StyleSheet, View } from 'react-native';
+import { Text } from '../../widgets';
+import { getLineHeight, THEME_FONT_SIZE_ONE } from '../../globalStyles';
+import { getArithmeticResult, getConditionResult } from '../selectors';
+
+export class CalculatedQuestionComponent extends PureComponent {
```
edmofro, line 9: Does this need the `export`?

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processArithmeticConfig.js
```diff
@@ -0,0 +1,108 @@
...
+ *     5ec5bf8961f76a18ad00d4f9:0
+ * },
+ * @param {*} models
+ * @param {*} defaultValuesConfig
+ */
+const translateDefaultValues = async (models, defaultValuesConfig) => {
+  const defaultValues = splitStringOnComma(defaultValuesConfig);
+  const translatedDefaultValues = {};
+
+  for (const defaultValue of defaultValues) {
```
edmofro, line 56: Same comment before about fetching all questions at once from the db applies here and a couple of places below. Could be made into a util, using your same util code without the $ prepended.

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processArithmeticConfig.js
```diff
@@ -0,0 +1,108 @@
...
+const translateAnswerDisplayText = async (models, text, codes) => {
+  let translatedText = text;
+
+  for (const code of codes) {
+    const { id: questionId } = await models.question.findOne({ code });
+    translatedText = translatedText.replace(code, questionId);
+  }
+
+  return translatedText;
+};
```
edmofro, line 108: This whole file was really beautifully laid out, well named, nice comments... ⭐

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processConditionConfig.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+const translateConditions = async (models, conditionsConfig, defaultValuesConfig = '') => {
+  const conditions = splitStringOnComma(conditionsConfig);
+  const translatedConditions = {};
+  const expressionParser = new BooleanExpressionParser();
+
+  // Translate the expressions
+  for (const condition of conditions) {
+    const [targetValue, expression] = splitStringOn(condition, ':');
+    const codes = expressionParser.getVariables(expression);
```
edmofro, line 51: What if the data element code starts with a number? Do any currently?

edmofro: Answered that myself. A few do, the ones representing universal drug codes

Screen Shot 2020-11-13 at 3 34 35 PM

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/ArithmeticConfigValidator.js
```diff
@@ -0,0 +1,87 @@
...
+        this.assertPointingToPrecedingQuestion(
+          questionCode,
+          rowIndex,
+          `Code '${questionCode}' does not reference a preceding question`,
+        );
+      }
+
+      return true;
+    };
+  }
```
edmofro, line 86: This is all really nice, making powerful use of other code

edmofro (changes requested):
Great work @chris-bes :-)

packages/meditrak-app/app/assessment/selectors.js
chris-bes: The conditions config is something like:
```
conditions: {
    Yes: {
      formula: '$5fa0cdcd5da1e614f30001fd >= 3',
      defaultValues: {
         '5fa0cdcd5da1e614f30001fd': '0',
      },
    },
    No: {
      formula: '$5fa0cdcd5da1e614f30001fd < 3',
      defaultValues: {
        '5fa0cdcd5da1e614f30001fd': '0',
      },
    },
  }
```

so the keys are the values which would be generated if the associated condition is satisfied.

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processConditionConfig.js
chris-bes: Yeah, so in the excel config, users will have to prefix $ to the variables then...

packages/database/src/DatabaseModel.js
```diff
@@ -166,6 +166,20 @@ export class DatabaseModel {
...
+      throw new Error(`${this.databaseType} table does not have code or id column`);
+    }
+    const records = await this.find({ code: codes });
+    return reduceToDictionary(records, 'code', 'id');
+  }
+
+  async checkFieldNamesExist(fields) {
+    const fieldNames = await this.fetchFieldNames();
+    return fields.every(field => fieldNames.includes(field));
+  }
```
edmofro, line 181: Both great utils!

packages/meditrak-server/src/routes/utilities/translateExpression.js
```diff
@@ -5,24 +5,27 @@
...
-export const translateExpression = async (models, formula, codes) => {
-  let translatedFormula = formula;
+export const translateExpression = async (models, rawExpression, codes) => {
+  const questionCodeToId = await models.question.findCodeToId(codes);
+  let expression = rawExpression;
 
   for (const code of codes) {
-    const { id: questionId } = await models.question.findOne({ code });
+    const questionId = questionCodeToId[code];
     // UUID often starts with a number which is invalid when evaluating using mathjs
```
edmofro, line 26:
```suggestion
    // Retain the $ prefix, as UUID often starts with a number (breaks evaluation in mathjs)
```
Now that the code is also prefixed with a `$`, the comment could be misleading

edmofro (approved):
Great work @chris-bes. It gets a little confusing when the codes do and do not have a `$` prefix, which makes me curse mathjs a bit, but I think you've done as good a job as possible of making it clear what is happening when.

chris-bes:
> Great work @chris-bes. It gets a little confusing when the codes do and do not have a `$` prefix, which makes me curse mathjs a bit, but I think you've done as good a job as possible of making it clear what is happening when.

I understand your frustration 😢 . Have created a ticket for forking mathjs :https://github.com/beyondessential/tupaia-backlog/issues/1594. Hopefully we can work on it soon

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/ArithmeticConfigValidator.js
```diff
@@ -21,28 +21,65 @@ export class ArithmeticConfigValidator extends JsonFieldValidator {
...
       answerDisplayText: [constructIsNotPresentOr(hasContent)],
     };
   }
 
+  /**
+   * When any questions used in the formula are non numeric, users need to provide 'valueTranslation'
+   * to translate the raw answers into numbers so that arithmetic calculation can be done.
+   * @param {*} rowIndex
+   */
+  hasContentIfQuestionsInFormulaAreNonNumeric(rowIndex) {
```
chris-bes, line 49: I've fixed the bug that Sepi found (https://github.com/beyondessential/tupaia-backlog/issues/982#issuecomment-731864359). If the formula in Arithmetic question contains any questions that are non numeric, users must provide `valueTranslation` to convert them into numeric so that the arithmetic calculation can be run.

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/ArithmeticConfigValidator.js
```diff
@@ -127,4 +129,21 @@ export class ArithmeticConfigValidator extends JsonFieldValidator {
       return true;
     };
   }
+
+  valueTranslationAreNumeric = () => {
```
chris-bes, line 133: Also added validation to check if all value translations are numeric

edmofro, line 133: `valueTranslationsAreNumeric`

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/ArithmeticConfigValidator.js
edmofro: Clever fix!

edmofro (approved):

chris-bes:
Added some more validations when importing a survey:
**Arithmetic Question:**
1. `valueTranslation`
- Validate if non numeric questions have corresponding value translations.
2. `defaultValues`
- Validate if defaultValues property is not empty if the formula has any optional questions.
- Validate if optional questions have corresponding default values.
- Validate if all default values are numeric.

**Condition Question:**
1. `defaultValues`
- Validate if defaultValues property is not empty if the formula has any optional questions.
- Validate if optional questions have corresponding default values.

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/CalculatedConfigValidator.js
```diff
@@ -0,0 +1,58 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { ValidationError } from '@tupaia/utils';
+import { JsonFieldValidator } from '../JsonFieldValidator';
+import { getExpressionQuestionCodes } from '../../../utilities';
+import { isEmpty, convertCellToJson } from '../../utilities';
+
+/**
+ * Base class for questions that automatically generate answers by evaluating expressions.
```
edmofro, line 12: Cool pattern, and good call making it more DRY

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/ConditionConfigValidator.js
```diff
@@ -3,27 +3,101 @@
...
+   */
+  constructDefaultValuesProvidedForAllOptionalQuestions(rowIndex) {
+    return (value, object) => {
+      const { conditions } = object;
+      const defaultValues = splitStringOnComma(value);
+      const resultValueToExpression = {};
+      const resultValueToCodes = {};
+
+      for (const condition of conditions) {
+        const [targetValue, expression] = splitStringOn(condition, ':');
```
edmofro, line 74: Love this syntax

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/ConditionConfigValidator.js
```diff
@@ -3,27 +3,101 @@
...
+  /**
+   * Validate if, for any optional questions, there are corresponding default values so that if optional questions are not answered, default values can be populated.
+   * @param {*} rowIndex
+   */
+  constructDefaultValuesProvidedForAllOptionalQuestions(rowIndex) {
+    return (value, object) => {
+      const { conditions } = object;
+      const defaultValues = splitStringOnComma(value);
+      const resultValueToExpression = {};
+      const resultValueToCodes = {};
```
edmofro, line 71: I would sit each of these declarations directly above the for loop that populates them

packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/ConditionConfigValidator.js
```diff
@@ -3,27 +3,101 @@
...
+      rowIndex,
+    );
 
-    return {
-      conditions: [constructIsNotPresentOr(conditionsPointToOtherQuestions)],
-      defaultValues: [constructIsNotPresentOr(defaultValuesPointToOtherQuestions)],
+    return [
+      existIfQuestionsInFormulaAreOptional,
+      constructIsNotPresentOr(defaultValuesPointToOtherQuestions),
+      constructIsNotPresentOr(defaultValuesProvidedForAllOptionalQuestions),
```
edmofro, line 34: I believe the third one makes the first one redundant?

edmofro: Oh right - nope, because it's checking for not present or.

edmofro (approved):
Plenty of validation now!

I find it hard to visualise the shape of each thing being parsed along the way. In situations like this, when there are varying structures like `x.b:10` and `y:m,n,o` etc., it can be really helpful to have comments showing the structure of the thing you're dealing with before (and sometimes after) each transformation. Still, this all looks nice as far as I can tell without intimately knowing the string formats being dealt with, so have approved!

edmofro (approved):

PR #1511, opened 2020-10-19
Fix terminal color output for MacOS

kael89:
### Issue #:
[No issue]

### Changes:
* Replaced `\e` with `\033` as an terminal escape character, since the latter doesn't work in MacOS
* Used the [shell-format](https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format) VSCode extension to format existing shell scripts.
* Indented scripts to consistently use 4 spaces


packages/devops/scripts/ci/deployBackend.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 PACKAGE=$1
-DIR=`dirname "$0"`
+DIR=$(dirname "$0")
```
kael89, line 3: `shell-format` prefers `$()` over ``. It's probably fine either way, but being consistent is probably beneficial.

edmofro (approved):
Nice idea adding linting for bash scripts :-)

PR #1513, opened 2020-10-19
1311 remove redundant link from flutracking project

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1311

I also removed the horizontal rule above the link because I think it looks cleaner - examples below:

_Without the horizontal rule (the approach I took):_

Screen Shot 2020-10-19 at 2 41 16 PM

_With the horizontal rule left in:_
- I think this looks worse but can add it back in if it should be kept

Screen Shot 2020-10-19 at 2 40 48 PM


IgorNadj (approved):
LGMT!

PR #1515, opened 2020-10-19
273 limit length of shortName for orgunit creation after entity import

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/273


edmofro (approved):
That was an oversight, thanks for fixing!

PR #1517, opened 2020-10-20
Fix refresh-database script for MacOS users

kael89:
### Issue #:
[No issue]

### DIFF
--- a/scripts/node/refreshDatabase.js
+++ b/scripts/node/refreshDatabase.js
@@ -31,13 +31,29 @@ class RefreshDatabaseScript extends Script {
     this.logInfo('Refreshing the tupaia database...');
 
     this.verifyPsql();
-    this.exec('psql -U postgres -c "DROP DATABASE IF EXISTS tupaia"');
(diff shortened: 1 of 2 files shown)


scripts/node/refreshDatabase.js
```diff
@@ -31,14 +31,30 @@ class RefreshDatabaseScript extends Script {
...
   }
 
+  execDbCommand = (dbCommand, { user, db } = {}) => {
+    const parts = ['psql'];
+    if (user) {
+      parts.push('-U', user);
+    } else if (this.isOsWindows()) {
+      // `psql` in Windows requires a user to be specified
+      parts.push('-U', 'postgres');
+    }
```
kael89, line 49: This is the actual fix: instead of using `psql -U postgres` for both platforms, we now use it only in Windows, where it is required (and a "posgres" user is created during the installation of PostgreSQL)

edmofro: Oh nice! I got this error and fixed it by modifying the `postgres` user to allow log in, but it's nicer if people don't have to go through that step (and there may be a reason that is turned off by default on mac)

edmofro (approved):

PR #1518, opened 2020-10-20
967: Update style guide part 3/3

kael89:
### Issue #:
**Edit:** Please also see the [967: Consolidate eslint configuration PR](https://github.com/beyondessential/tupaia/pull/1521) after this one. I separated those the config consolidation PR is more contained, but has less straight-forward changes than the ones here.

3rd and last part of the style guide update! Another PR may follow for consolidating `.eslintrc` configs, but the bulk of the remaining changes are in this PR.

* Lint `meditrak-app`
* Lint `indicators`
* Minor fixes in the `migrations` folder. I decided to enable `prettier` for them but not autofix any formatting errors there, because that would lead to a PR of 600+ changed files. What we could also do is: I can run the autofix for migrations in the presence of another dev, and then they can directly approve the resulting PR.


.eslintrc
```diff
@@ -2,10 +2,12 @@
   "extends": "@beyondessential/js",
   "parser": "babel-eslint",
   "rules": {
+    "global-require": "off",
```
edmofro, line 5: I can see that this is deprecated in eslint 7, is that the reason for turning it off? I believe we're in a good habit with this across the board anyway, so I'm not worried, just interested in the reason.

.vscode/tupaia-packages.code-snippets
```diff
@@ -15,7 +15,6 @@
       "/**",
       " * @abstract",
       " */",
-      "// eslint-disable-next-line class-methods-use-this",
```
edmofro, line 18: Good find!

.eslintrc
kael89: Good point, I should have left a comment!

We generally use ES6 modules in this repo, so `require` is mostly used in scripts like the webpack/create-react-app ones. In those instances, there are justified cases where we want to conditionally require a dependency in the body of the script.

Also, the [dynamic import proposal](https://github.com/tc39/proposal-dynamic-import) is in Stage 4, and I expect that once this is included in ES it may change a bit how we handle imports. In conclusion, and taking into account that this rule is deprecated, I thought we may as well as turn it off for this repository. I believe we should still keep it in the base style guide, since other repos may rely heavily in CommonJS modules.

packages/indicators/src/types.ts
```diff
@@ -22,7 +22,7 @@ export interface AnalyticCluster {
   dataValues: Record<Analytic['dataElement'], Analytic['value']>;
 }
 
-type TypeFields = Record<string, string | number | {}>;
+type TypeFields = Record<string, string | number | Record<string, unknown>>;
```
edmofro, line 25: After a looking at a few examples of this change, I prefer the old way...

kael89: I preferred the old way too (in terms of syntax), but it's actually inaccurate...

https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md#default-options

> Avoid the Object and {} types, as they mean "any non-nullish value".
This is a point of confusion for many developers, who think it means "any object type".

edmofro: Oh good to know, I definitely thought it meant any object type so it's worth the change

packages/meditrak-app/app/sync/ChangeQueue.js
```diff
@@ -52,12 +52,12 @@ export class ChangeQueue {
     const allChanges = this.database.objects('Change').sorted('timestamp');
     const changesWithinThreshold = [];
     let size = 0;
-    for (let changeId in allChanges) {
+    for (const changeId in allChanges) {
       const change = allChanges[changeId];
```
edmofro, line 56:
```suggestion
    for (const change of allChanges) {
```
This is a substantive rather than formatting change, but I think it is appropriate

edmofro (approved):
Nice one K :-)

packages/meditrak-app/app/sync/ChangeQueue.js
kael89: Good point! Will do

edmofro (approved):

PR #1519, opened 2020-10-20
Fix default dates enlarged dialog

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/648

### Changes:
This line here (I think) was the catalyst for it breaking:
image

However, that change just exposed the fact that upon opening the enlargedDialog, the startDate and endDate are null (and were explicitly set that way in the reducer. 

Edit: will create a new branch name (it can't end with `export`)

### DIFF
--- a/packages/web-frontend/src/actions.js
+++ b/packages/web-frontend/src/actions.js
@@ -1139,10 +1139,12 @@ export function closeEnlargedDialog() {
   };
 }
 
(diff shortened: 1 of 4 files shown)


Alastair-L:
Closed in favor of: https://github.com/beyondessential/tupaia/pull/1520

PR #1520, opened 2020-10-20
Fix default dates enlarged dialog

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/648

### Changes:
~This line here (I think) was the catalyst for it breaking:~ Igor pointed out that the bug report is from June, which rules that out...
image

However, that change just exposed the fact that upon opening the enlargedDialog, the startDate and endDate are null (and were explicitly set that way in the reducer.


IgorNadj:
Yeah I'm not sure if the start/end dates being null was intentional, but we are moving towards never having null dates, see https://github.com/beyondessential/tupaia-backlog/issues/1381, so I'm happy with adding dates in generally

IgorNadj (approved):
Code changes look good, nice one. 

The behaviour here is pretty complicated, if it's using the right default dates in all cases (check the dashboards with default date offsets), and it's not sending multiple requests, should be good.

Alastair-L:
Closing as it has been incidentally fixed elsewhere

PR #1521, opened 2020-10-21
967: Consolidate eslint configuration

kael89:
### Issue #:

We haven't discussed this in the card, but I found out that jest uses a single [eslint configuration file](https://github.com/facebook/jest/blob/master/.eslintrc.js). I have spent some time lately working with ESLint configurations, and I personally find this pattern very valuable.

**Advantages:**
* Single source of truth
* Encourages consistent rules across the codebase
* Discourages overriding rules just for convenience

**Disadvantages:**
* Possibly a bit harder to read for people new to ESLint configs
* Can become messy if not properly maintained. Hopefully the commented sections in `.eslintrc` help to avoid this

 We can mitigate the disadvantages by delegating the task for creating/maintaining ESLint config for existing and new packages to devs that feel familiar with it.


edmofro (approved):
Yeah I like this change. The devDependencies is the weirdest bit, but we can deal with that - I think it's worth it to discourage changing configs on a package basis

PR #1522, opened 2020-10-21
885 Remove Mugil facility from STRIVE

IgorNadj:
### Changes:

- Remove Mugil from STRIVE

---


packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
```diff
@@ -0,0 +1,163 @@
...
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const getHierarchyId = async (db) => {
+  const results = await db.runSql(`SELECT id FROM entity_hierarchy WHERE name = 'strive';`);
+  if (results.rows.length > 0) {
+    return results.rows[0].id;
+  }
+  throw new Error(`Strive hierarchy not found`);
```
kael89, line 24: ❤️ those debugging messages!

packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
```diff
@@ -0,0 +1,163 @@
...
+      parent_id: entityIds.PG,
+      child_id: districtId,
+      entity_hierarchy_id: hierarchyId
+    });
+  }
+
+  // 1.2. generate list of facilities to insert
+  const facilityIdsToInsert = [];
+  for (const districtId of districtIds) {
+    const facilityIds = await getEntityIdsByParentId(db, districtId);
```
kael89, line 66: Nice generalisation in `getEntityIdsByParentId`

packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
```diff
@@ -0,0 +1,163 @@
...
+      '${hierarchyId}' as entity_hierarchy_id 
+    FROM entity 
+    WHERE id IN (${arrayToDbString(facilityIdsToInsert)});
+  `);
+
+  // 1.3. insert villages for each facility inserted
+  await db.runSql(`
+    INSERT INTO entity_relation (id, child_id, parent_id, entity_hierarchy_id)
+    SELECT
+      generate_object_id() as id,
```
kael89, line 92: First time I see this! It seems that it is a Postgres function that is available to us. I've been always using `generateId` defined in `@tupaia/database/src/utilities/generateId.js`, @edmofro thoughts?

It would be convenient to indeed be able to use this!

packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
```diff
@@ -0,0 +1,163 @@
...
+      generate_object_id() as id, 
+      entity.id as child_id, 
+      entity.parent_id as parent_id, 
+      '${hierarchyId}' as entity_hierarchy_id 
+    FROM entity 
+    WHERE id IN (${arrayToDbString(facilityIdsToInsert)});
+  `);
+
+  // 1.3. insert villages for each facility inserted
+  await db.runSql(`
```
kael89, line 89: I haven't done any development work in entity hierarchies but I've done reviewing, and my impression is that we just need to define entity relations for the level we want to customise: in this case, it would be defining entity relations for Strive facilities.

This means that we don't need to create entity relations for villages, since we will retrieve those via the standard hierarchy: they will be attached to their parent facilities via `parent_id` in the `entity` table. Similarly, we wouldn't need to create entity relations for districts because no customisation is needed for them. @edmofro could you please confirm this?

packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
```diff
@@ -0,0 +1,163 @@
...
+  for (const districtId of districtIds) {
+    const facilityIds = await getEntityIdsByParentId(db, districtId);
+    for (const facilityId of facilityIds) {
+      if (facilityId === entityIds.Mugil) {
+        continue;
+      }
+      facilityIdsToInsert.push(facilityId);
+    }
+  }
+  facilityIdsToInsert.push(entityIds.Sausi);
```
kael89, line 74: I don't think we need to treat `Sausi` in a special manner here, since its id would have been inserted in `facilityIdsToInsert` as part of the above process

packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
```diff
@@ -0,0 +1,163 @@
...
+
+    if (dataBuilderConfig.dataBuilders) {
+      // composed config, need to update nested configs
+      for (const nestedConfigKey of Object.keys(dataBuilderConfig.dataBuilders)) {
+        updateConfig(dataBuilderConfig.dataBuilders[nestedConfigKey].dataBuilderConfig);
+      }
+    } else {
+      // simple case
+      updateConfig(dataBuilderConfig);
+    }
```
kael89, line 125: Nice!!

kael89 (commented):
Great work overall! I like the disciplined way you structured your migration. My review is mainly comments and questions, holding approval till those are answered.

One item that is directly actionable is removing the special code for `Sausi` (if my point is correct)

IgorNadj:
Kostas Karvounis  9:36 AM
Hey man! About https://github.com/beyondessential/tupaia/pull/1522, we forgot to follow up with Edwin on Thursday (edited) 
9:38
There are two points there:
The id one is probably a detail. Safest option though is to use generateId (which is what we usually do in migrations). I like your usage of generate_object_id() , but we could skip it this time till Edwin gets back to us
About the entity hierarchy, I suggest simplifying the alternative hierarchy as suggested in the comment and see if it works. In other words, you can try to just insert entries for the facility level. You can create a new/temporary branch locally for that

packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
edmofro: Yep, it is really cool actually, Kurt introduced it [here](https://github.com/beyondessential/meditrak-server/commit/78820d90349a9a60de59dd8ecc494e12c281aadf). I have always been slightly nervous that it will produce ids differently in some minor way, so I tend to use the js version in migrations - perhaps one of you could do a brief assessment of whether there are differences in the ids produced, and check the original gist for comments on the appropriateness of using it? https://gist.github.com/jamarparris/6100413

IgorNadj: After a quick read through, having the client generate the id prevents conflicts because the machine id is part of it. The db function doesn't have a machine id so uses a random number. 

But in either case, the possibility for conflict is statistically impossible, so the two methods look to me to be equally valid.

edmofro: Great, thanks @IgorNadj!

packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
IgorNadj: I double checked and you are completely right.

Not sure why it didn't work for me like that the first time, but changing to just insert facility level (without Mugil).

packages/database/src/migrations/20201016001619-RemoveMugilFromStrive-modifies-data.js
IgorNadj: Updated

IgorNadj:
@kael89 updated, please have a look again, thanks! 🚀

kael89 (approved):
Looks good, great work! (I focused on the parts that seemed different to the previous state of your PR)

edmofro:
@IgorNadj now that https://github.com/beyondessential/tupaia-backlog/issues/885 has had some of the issues resolved, can this move forward?

IgorNadj:
> @IgorNadj now that [beyondessential/tupaia-backlog#885](https://github.com/beyondessential/tupaia-backlog/issues/885) has had some of the issues resolved, can this move forward?

@edmofro yep, just figuring out who is going to pick this up next

IgorNadj:
Force pushed to rebase on dev. Not much changed, added https://github.com/beyondessential/tupaia/pull/1522/commits/ebefc4ea14d8172ef3630f6422273e646afd4eb2

edmofro:
Is this one ready for re-review @IgorNadj? Might be worth putting it in the reviews channel?

IgorNadj:
@edmofro this is up for testing, I skipped re-review because of it just being a small addition of a missing param.

edmofro:
Hmm seems worth re-review given the last approval was 6th Nov - something might have changed from underneath it!

edmofro (approved):
Beautiful migration, glad to get the chance to read it!

PR #1526, opened 2020-10-22
1526 Prevent change question's data service on survey reimport

IgorNadj:
### Changes:

- See https://github.com/beyondessential/tupaia-backlog/issues/1333


edmofro (changes requested):
This is a good change, and all of the code here is approved. However, it doesn't fix https://github.com/beyondessential/tupaia-backlog/issues/1333 completely. We want to avoid the situation where a BCD survey is reimported, and although 99% of the data elements use the `dhis` service, the data element `BCD1` uses `tupaia`. On reimport, we want to leave that as is. 

I believe the fix is `const dataElement = await models.dataSource.updateOrCreate(` -> `const dataElement = await models.dataSource.findOrCreate(`, so that the `service_type` only gets set if it doesn't already exist.

IgorNadj:
Yeah interesting, I considered that but I saw some validation on import which enforces dataGroup.service_type === dataElement.serviceType, so I thought it could never get into that state. I guess the dataElement's service type could be changed manually after the import.

edmofro:
Yep :+1:

IgorNadj:
@edmofro Do we want to allow mixed service dataGroups (some dataElements use dhis, some use tupaia)? If so, I think we should remove the [validation in DataElementDataGroup](https://github.com/beyondessential/tupaia/blob/dev/packages/database/src/modelClasses/DataElementDataGroup.js#L13). This validation seems to be incorrect if we allow the behaviour.

We can fix this bug without making this change, but it would be a little bit hacky. Because it deletes the DataElementDataGroup links and recreates them on every survey import[1], the validation runs every time, including for re-imports where the dataElement and dataGroup already exist. If the existing dataElement and dataGroup use different services, it throws the validation error and import fails. 

The hacky fix would be to skip the validation check on re-imports where dataElement and dataGroup already exist. Alternatively the cleaner fix would be to remove the validation.

[1] I believe it has to to be able to do a clean import. I don't think it's worth doing a conditional delete here.

---

My preference would be to have the validation be correct, to keep the data in a valid state, which means if it is wrong removing it. However, if there are other implications with removing this that I'm unaware of, let me know and I will do the quick fix.

edmofro:
The absolutely correct rules are a bit hard to capture in code, but go like this:
- If it is event based data, we only support pushing all data elements in a group to the same data service. It is event based if:
  - It is set to can_repeat = true
  - A specific survey response is for an entity that is not a canonical organisation unit (i.e. is not a country, district, subdistrict, facility, or village)
- If it is aggregate data, we support pushing data elements via different data services

The right thing to do seems to be:
- Remove the validation here
- Add validation into DHIS2 EventPusher that makes sure all individual answers being pushed relate to data elements that are for DHIS2, and either
  - Throws an error, which means we still need to configure those to all be for a single service (currently that's a fine setup); or
  - Filters those that aren't for the service out of the event being pushed (or never puts them in in the first place?), which would add support for pushing some data of a survey response to an event, and leaving some for another data service to take care of
- Make sure that aggregate data is added to the dhis sync queue depending on the data service attached to the data element attached to the question, rather than just assuming the data service attached to the data group applies to all the content within

IgorNadj:
Discussed above.

The option of throwing or filtering - we want to throw. Conceptually a survey response is a whole event. Filtering would water it down into a partial event (the other part being in some other service), so we don't want to do that. We want to throw.

IgorNadj:
@edmofro Updated, with notes and screenshots in the backlog ticket description.

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
```diff
@@ -72,21 +80,44 @@ const updateOrCreateDataElementInGroup = async (models, dataElementCode, dataGro
...
       code: dataElementCode,
-    },
-    { service_type: serviceType },
-  );
-  dataElement.config = { ...dataElement.config, ...config };
-  dataElement.sanitizeConfig();
-  await dataElement.save();
+      service_type: serviceType,
+    });
+    dataElement.config = { ...dataElement.config, ...config };
```
edmofro, line 93:
```suggestion
    });
    dataElement.config = config;
```
The spreading of existing config isn't required for newly created data elements

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
```diff
@@ -72,21 +80,44 @@ const updateOrCreateDataElementInGroup = async (models, dataElementCode, dataGro
...
+      data_element_id: dataElement.id,
+      data_group_id: dataGroup.id,
+    });
+  } else {
+    /*
+     * Link the existing dataElement to a potentially new dataGroup
+     *
+     * We use separate find and create instead of findOrCreate because we may have an
+     * "invalid" state, where the dataElement and dataGroup have different services, and
+     * we want to allow this to continue as is.
```
edmofro, line 107: Even with your well placed comment I can't understand why we can't use `findOrCreate`. Could you explain a bit more? Is there validation that runs on `findOrCreate` that doesn't run on `create` or something?

edmofro (approved):
I must say, your code is extremely clear to read through.

Just two minor comments, but this is pretty much good to go so I'm preapproving.

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
IgorNadj: Seems like an old comment that does not apply anymore, updated

IgorNadj:
> I must say, your code is extremely clear to read through.

Thanks :) it's more for my benefit I think, I had a bit of a hard time wrapping my head around the business logic here, I was thinking this could really use some tests but it seems like a behemoth task, including breaking up importSurveys into testable bits.

> Just two minor comments, but this is pretty much good to go so I'm preapproving.
👍

edmofro (approved):

edmofro:
Yes it did @IgorNadj, I thought that was also well laid out and helped me to get to grips with this coming back in to it.

PR #1528, opened 2020-10-23
1305 project permission flow

tcaiger:
### Issue #: [1305](https://github.com/beyondessential/tupaia-backlog/issues/1305)

Improve flow of user trying to access projects they don't have access to.

### Changes:

- If a user goes to a url they don’t have permission for, and is signed in => show a “request access” modal
- If user goes to a url they don’t have permission for, and is not signed in => show the login modal
- After the user is signs in, remember the url that they were entered and handle it accordingly

See comments inline for more description of the changes.

I have added another card to handle urls that are not correctly entered [here](https://github.com/beyondessential/tupaia-backlog/issues/1474)

---

### Screenshots:
See issue for screenshots


packages/web-frontend/src/historyNavigation/historyMiddleware.js
```diff
@@ -12,92 +12,25 @@
...
   addPopStateListener,
   attemptPushHistory,
   clearLocation,
-  getInitialLocation,
   setLocationComponent,
 } from './historyNavigation';
-import { PROJECTS_WITH_LANDING_PAGES, PROJECT_LANDING } from '../containers/OverlayDiv/constants';
-import { convertDateRangeToUrlPeriodString, decodeLocation } from './utils';
-
-export const reactToInitialState = store => {
```
tcaiger, line 47: Move `reactToInitialState` and `reactToLocationChange` to sagas.js because they involve complex interaction with the store. This also allows us to re-use this behaviour to react to a location change in other flows of the app (eg. after logging in)

packages/web-frontend/src/projects/sagas.js
```diff
@@ -29,14 +26,6 @@ function* fetchProjectData() {
   }
 }
 
-function* watchFetchInitialDataAndFetchProjects() {
```
tcaiger, line 32: Moved these 2 watcher effects to the initial data and login success sagas

packages/web-frontend/src/reducers.js
```diff
@@ -587,7 +587,7 @@ function measureBar(
 function global(
   state = {
     isSidePanelExpanded: false,
-    overlay: !isMobile() && LANDING,
+    overlay: null,
```
tcaiger, line 590: We don't always want to show the landing by default in case a user enters a nested url

packages/web-frontend/src/selectors/projectSelectors.js
```diff
@@ -6,15 +6,16 @@
...
 import { selectLocation } from './utils';
 import { TILE_SETS } from '../constants';
 
 const selectAllProjects = state => state.project.projects;
 
-export const selectCurrentProjectCode = createSelector([selectLocation], location =>
-  getLocationComponentValue(location, URL_COMPONENTS.PROJECT),
+export const selectCurrentProjectCode = createSelector(
+  [selectLocation],
+  location => getLocationComponentValue(location, URL_COMPONENTS.PROJECT) || DEFAULT_PROJECT_CODE,
```
tcaiger, line 18: Add fallback so that this selector always returns a valid project

packages/web-frontend/src/historyNavigation/historyMiddleware.js
kael89: Good call!

packages/web-frontend/src/sagas.js
```diff
@@ -117,9 +130,124 @@ import {
...
+
+function* handleInvalidPermission({ projectCode }) {
+  const state = yield select();
+  const { isUserLoggedIn } = state.authentication;
+
+  if (isUserLoggedIn) {
+    // show project access dialog
+    const project = selectProjectByCode(state, projectCode);
+
+    if ('code' in project) {
```
kael89, line 162: Probably this is a bit closer to the actual check (that a project exists?) Since we don't need `project.code` in specific

```suggestion
    if (project) {
```

packages/web-frontend/src/sagas.js
```diff
@@ -117,9 +130,124 @@ import {
...
+    // handle 404s
+    console.log('project does not exist - 404'); // Todo: handle 404s
+    return;
+  }
+  // show login dialog
+  yield put(setOverlayComponent(LANDING));
+}
+
+function* setComponentsIfUpdated(components, previousComponents, otherComponents) {
+  for (let i = 0; i < components.length; i++) {
```
kael89, line 177: I guess you can also use a `forEach` here

packages/web-frontend/src/sagas.js
```diff
@@ -117,9 +130,124 @@ import {
...
+  const previousComponents = decodeLocation(previousLocation);
+  yield call(
+    setComponentsIfUpdated,
+    [
+      { key: URL_COMPONENTS.PROJECT, setter: setProject },
+      { key: URL_COMPONENTS.ORG_UNIT, setter: setOrgUnit },
+      { key: URL_COMPONENTS.URL_COMPONENTS, setter: setMeasure },
+      { key: URL_COMPONENTS.REPORT, setter: openEnlargedDialog },
+      { key: URL_COMPONENTS.MEASURE_PERIOD, setter: updateCurrentMeasureConfigOnceHierarchyLoads },
+    ],
```
kael89, line 242: Although this is fine, you could probably provide an object here, and then iterate over its entries:

```js
{
  [URL_COMPONENTS.PROJECT]: setProject,
  [URL_COMPONENTS.ORG_UNIT] setOrgUnit,
  // etc...
}
```

kael89 (approved):
Great work! A few suggestions but they are all optional, approved!

packages/web-frontend/src/sagas.js
tcaiger: great suggestion

packages/web-frontend/src/sagas.js
tcaiger: Unfortunately you can not call yield from within an inner function. It must be from within the main generator function.

kael89: No worries!

packages/web-frontend/src/sagas.js
tcaiger: Good point. I wanted to make sure `project = {}` didn't return a false positive. So I've update the check to be: `Object.keys(project).length > 0`

tcaiger:
merging into main feature PR.

PR #1529, opened 2020-10-23
Allow exporting empty survey response

edmofro:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/1466

Now outputs the empty survey response (i.e. info row labels) as well as the "no data" comment if there are no survey responses.

Also cleaned up the code in the immediate area a little, and fixed a bug where the survey responses were being exported in an indeterminate order.

### DIFF
--- a/packages/meditrak-server/src/routes/exportSurveyResponses.js
+++ b/packages/meditrak-server/src/routes/exportSurveyResponses.js
@@ -93,7 +93,7 @@ export async function exportSurveyResponses(req, res) {
       entities = [entity];
     } else if (countryId) {
       country = await models.country.findById(countryId);
-      entities = await models.entity.find({ country_code: country.code }, { sort: ['name'] });
+      entities = await models.entity.find({ country_code: country.code }, { sort: ['name ASC'] });
     } else if (entityIds) {
       entities = await Promise.all(
         entityIds.split(',').map(entityId => models.entity.findById(entityId)),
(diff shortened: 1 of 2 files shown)


packages/meditrak-server/src/routes/exportSurveyResponses.js
```diff
@@ -227,66 +227,64 @@ export async function exportSurveyResponses(req, res) {
...
+
+      // Add the answers on the right columns to the exportData
+      questionsForExport.forEach((question, questionIndex) => {
+        const exportRow = preQuestionRowCount + questionIndex;
+        surveyResponseAnswers.forEach((answersByQuestionId, surveyResponseIndex) => {
+          const answer = answersByQuestionId[question.id];
+          const exportColumn = infoColumnKeys.length + surveyResponseIndex;
+          exportData[exportRow][exportColumn] = answer || '';
+        });
+      });
```
kael89, line 286: I believe this loop can be consolidated with the one above

kael89 (approved):
Looks good! This will be a really valuable feature. Just one optional suggestion, approving

packages/meditrak-server/src/routes/exportSurveyResponses.js
edmofro: We separated them intentionally, because they are logically responsible for different things - and it's next to no performance hit to do (x + y) amount of effort n times, vs. x amount of effort n times + y amount of effort n times

PR #1530, opened 2020-10-24
1232: Allow filtering options by a previous question

chris-bes:
### Changes:

- Allow filtering options by previous question

### DIFF
--- a/packages/meditrak-app/app/assessment/specificQuestions/AutocompleteQuestion.js
+++ b/packages/meditrak-app/app/assessment/specificQuestions/AutocompleteQuestion.js
@@ -4,22 +4,27 @@
  **/
 
 import React from 'react';
+import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import { View } from 'react-native';
-
+import { getQuestion, getAnswerForQuestion } from '../selectors';
 import { Autocomplete } from '../../widgets/Autocomplete/Autocomplete';
+import { constructRecord } from '../../database/constructRecord';
 
 const OPTIONS_PER_PAGE = 6;
-export class AutocompleteQuestion extends React.Component {
+export class AutocompleteQuestionComponent extends React.Component {
   constructor(props) {
     super(props);
 
     this.optionSet = props.realmDatabase.getOptionSetById(props.optionSetId);
+    const checkMatchesAttributeFilters = this.getOptionAttributeFilters();
+    const filteredResults = this.optionSet.options
+      .sorted('sortOrder')
+      .filter(checkMatchesAttributeFilters);
+
     this.state = {
-      filteredResults: this.optionSet.options.sorted('sortOrder'),
-      optionList: this.buildOptionList(
-        this.optionSet.options.sorted('sortOrder').slice(0, OPTIONS_PER_PAGE),
-      ),
+      filteredResults,
+      optionList: this.buildOptionList(filteredResults.slice(0, OPTIONS_PER_PAGE)),
     };
   }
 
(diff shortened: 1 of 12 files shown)


packages/meditrak-app/app/database/schema.js
```diff
@@ -7,7 +7,7 @@ import * as DataTypes from './types';
 
 export const schema = {
   schema: Object.values(DataTypes),
-  schemaVersion: 17,
+  schemaVersion: 18,
```
chris-bes, line 10: I also increased the version in https://github.com/beyondessential/tupaia/pull/1451. Should I maybe branch off from it?

edmofro: I think it'll be fine - could you just keep an eye to make sure they end up in the same app release?

packages/meditrak-app/app/assessment/specificQuestions/AutocompleteQuestion.js
```diff
@@ -42,15 +47,29 @@ export class AutocompleteQuestion extends React.Component {
 
   buildOptionList = options => options.map(option => option.label || option.value);
 
+  getOptionAttributeFilters = () => {
+    const { attributeAnswers } = this.props;
+
+    return option =>
+      attributeAnswers
+        ? Object.entries(attributeAnswers).every(([key, answer]) => {
+            // No answer was selected for the question to filter, return all
```
edmofro, line 56: Is this comment in the right place?

packages/meditrak-server/src/routes/importSurveys/ConfigImporter/processAutocompleteConfig.js
```diff
@@ -0,0 +1,56 @@
+/**
```
edmofro, line 1: In here there's some quite familiar code - is there stuff we can reuse in a common util from your earlier PRs?

edmofro (changes requested):
Looking good @chris-bes :-)

How's the performance of this? Any lag as you type a search filter on a big option set? It has to run the attributes filter across every option as well, including running the `toJson` every time. Shouldn't be slow, but with a large number of options...maybe?

chris-bes:
> Looking good @chris-bes :-)
> 
> How's the performance of this? Any lag as you type a search filter on a big option set? It has to run the attributes filter across every option as well, including running the `toJson` every time. Shouldn't be slow, but with a large number of options...maybe?

So far no but I only tested with a small number of options. Will test it with more options and let you know!

packages/meditrak-server/src/routes/utilities/translateQuestionDependentFields.js
```diff
@@ -0,0 +1,76 @@
+import { splitStringOn } from './split';
+
+// JSON subfields will be in the format field.subfield, e.g. 'attributes.type'
+const isJsonSubfield = (jsonFieldList, fieldKey) =>
+  jsonFieldList.find(field => fieldKey.startsWith(field));
+
+export const translateQuestionDependentNonJsonFields = (config, nonJsonFieldTranslation) => {
```
chris-bes, line 7: I moved the code that translate/replace the config into 1 util file like this to reduce code duplication. However, it felt a bit awkward for me to translate both non json (eg: parent, grandparent) and json (attributes.type, attributes.parent_project) in 1 method. So I split it into 2 different methods, 1 for translating non json and 1 for translating json which I feel is a bit cleaner (although it have a little bit of code duplication)

chris-bes:
> Looking good @chris-bes :-)
> 
> How's the performance of this? Any lag as you type a search filter on a big option set? It has to run the attributes filter across every option as well, including running the `toJson` every time. Shouldn't be slow, but with a large number of options...maybe?

I've tested it with 100 options and searching was fast and no lag at all!

packages/meditrak-server/src/routes/utilities/translateQuestionDependentFields.js
edmofro: Good call. I would remove the `NonJson` bit from the function and variable names though. Having the other one include `Json` is enough to distinguish the two.

edmofro: Actually I would use the term `Nested` rather than `Json`...

edmofro (approved):
Approved as is, with some optional name tweak suggestions :-)

PR #1531, opened 2020-10-26
1438 fix opacity for Service Suggestions drilldown

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1438

This basically just changes the wrapper background colour to dark blue if the problematic drilldown identified in that bug report is selected
- The reason I didn't just simply change the background colour to dark blue in all cases is because I assume it is set to that colour for a reason (although I'm not exactly sure what that reason is - as far as I can tell the existing colour only shows up during transitions between drilldowns in other drilldown reports)

### DIFF
--- a/packages/web-frontend/src/components/View/index.js
+++ b/packages/web-frontend/src/components/View/index.js
@@ -6,5 +6,5 @@
 export { ChartWrapper } from './ChartWrapper';
 export { MatrixWrapper } from './MatrixWrapper';
 export { View } from './View';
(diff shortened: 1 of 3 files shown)


edmofro (changes requested):
Sorry @avaek but this fix is too specific. I can see why you're being conservative, but introducing a conditional based on a specific report's name is the durum flour from which spaghetti code could proliferate. Perhaps you can look at who set it to the other colour, and ask them why, and even pair to work out a more generic solution?

edmofro:
Closing PR as diff is empty!

PR #1532, opened 2020-10-26
Fix detect stage script

kael89:
### Issue #:
Reinstates functionality back to a working state

### DIFF
--- a/packages/devops/scripts/utility/detectStage.sh
+++ b/packages/devops/scripts/utility/detectStage.sh
@@ -3,6 +3,6 @@
 # Get the stage tag of this ec2 instance from AWS
 TAG_NAME="Stage"
 INSTANCE_ID="$(wget -qO- http://instance-data/latest/meta-data/instance-id)"
-REGION="$(wget -qO- http://instance-data/latest/meta-data/placement/availability-zone | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:')"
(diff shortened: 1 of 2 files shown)


packages/devops/scripts/utility/detectStage.sh
```diff
@@ -3,6 +3,6 @@
 # Get the stage tag of this ec2 instance from AWS
 TAG_NAME="Stage"
 INSTANCE_ID="$(wget -qO- http://instance-data/latest/meta-data/instance-id)"
-REGION="$(wget -qO- http://instance-data/latest/meta-data/placement/availability-zone | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:')"
+REGION="`wget -qO- http://instance-data/latest/meta-data/placement/availability-zone | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:'`"
```
kael89, line 6: I don't know why this doesn't work... It was automatically applied by [shell-format](https://marketplace.visualstudio.com/items?itemName=foxundermoon.shell-format&ssr=false#overview) 😢

IgorNadj (approved):

packages/devops/scripts/utility/detectStage.sh
kael89: I'm not familiar with the `\` syntax for `sed`, I use `/`. In the meantime, I'll have a look at the PR which introduced this change to verify that every other related change is ok

PR #1534, opened 2020-10-26
Hotfix event survey res not syncing

chris-bes:
### Issue: 
Fix an issue with syncing event survey responses that I've found today

### Changes:

After upgrading DHIS2 to 2.35, syncing **event** survey responses started to have problem (my bad, I only smoke tested syncing **analytic** survey responses :( ). It gives the error: `Event.programStage does not point to a valid programStage: W9ijoiwH9mD`.

After investigating on DHIS2 2.35 code, particularly:

1. `ProgramStageCheck.java` (where it gives the error, line 61)
https://github.com/dhis2/dhis2-core/blob/ddb1f87fa26015b5fdb0e374ea863e499ec40280/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/insert/validation/ProgramStageCheck.java

2. And `WorkContext.getProgramStageIdScheme(), line 157` (this is where it find the programStage)
https://github.com/dhis2/dhis2-core/blob/1bd8ee38126fbac22f805db44dad0cdcba7b14a6/dhis-2/dhis-services/dhis-service-dxf2/src/main/java/org/hisp/dhis/dxf2/events/importer/context/WorkContext.java

4. And `getIdentifierBasedOnIdScheme()` line 410 (this is where it find the ID based on the id scheme)
https://github.com/dhis2/dhis2-core/blob/b85261219b82ad71593275bc8ecad65d1f01bfa5/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/IdentifiableObjectUtils.java

3. Also had a look at documentation for IdScheme:
https://docs.dhis2.org/master/en/developer/html/webapi_identifier_schemes.html

It seems like UID is the type of id scheme that we want for most of the objects (except program and dataElement because the event data format we have are using `codes` for those) when posting events. That's why it could not find the program stage because it uses the `code` of the program stage to compare (it should actually use the UID of the program stage to compare). Not sure why it worked for previous version that we had, from looking at the new DHIS2 code there was a major refactoring regarding import validation. So I made this change and have tested a lot with syncing both analytic/event survey responses/answers again (submit/update/delete survey responses) and now it seems fine.

Sorry for not testing thoughout before upgrading the instances... :(

### DIFF
--- a/packages/dhis-api/src/DhisApi.js
+++ b/packages/dhis-api/src/DhisApi.js
@@ -149,8 +149,15 @@ export class DhisApi {
     return this.fetch(endpoint, queryParameters, { method: 'DELETE' });
   }
 
-  async postData(endpoint, data) {
-    return this.post(endpoint, data, { idScheme: 'code', skipAudit: true });
+  async postData(endpoint, data, queryParameters = {}) {
+    const { idScheme = 'code', orgUnitIdScheme, programIdScheme } = queryParameters;
+
+    return this.post(endpoint, data, {
+      idScheme,
+      orgUnitIdScheme,
+      programIdScheme,
+      skipAudit: true,
+    });
   }
 
   async postDataValueSets(data) {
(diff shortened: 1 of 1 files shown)


edmofro (approved):
Great work finding the fix quickly. Please go ahead and merge straight to master and check that it's worked, then merge master in to dev

PR #1537, opened 2020-10-27
1429 update link generation and existed links data for feed items

billli0:
### Issue #:
[Update the link generation on social feed items](https://github.com/beyondessential/tupaia-backlog/issues/1429#issuecomment-716313917)

### Changes:

1. Update value with key `link` in `template_variables` in `feed_item` table, from legacy link format: `https:\\mobile.tupaia.org\facility\{faclitycode}` or `https:\\mobile.tupaia.org\country\{countrycode}`
to `https:\\mobile.tupaia.org\explore\{code}` 

2. Update the link generation in `meditrak-server`

### DIFF
--- a/packages/database/src/migrations/20201026223059-UpdateLinkOnFeedItems-modifies-data.js
+++ b/packages/database/src/migrations/20201026223059-UpdateLinkOnFeedItems-modifies-data.js
@@ -1,0 +1,36 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function(db) {
+  /**
+   * Update value with key `link` in `template_variables`
+   */
+  return db.runSql(`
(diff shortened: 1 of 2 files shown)


packages/database/src/migrations/20201026223059-UpdateLinkOnFeedItems-modifies-data.js
```diff
@@ -0,0 +1,36 @@
...
+
+exports.up = function(db) {
+  /**
+   * Update value with key `link` in `template_variables`
+   */
+  return db.runSql(`
+    update feed_item
+    set template_variables = regexp_replace (
+                      template_variables :: text,
+                      '(?<=https:\\/\\/mobile.tupaia.org\\/)(country|facility)',
```
rohan-bes, line 25: I suppose technically we ought to be escaping the `.` character too since that's a regex special character, so it should be:
`(?<=https:\\/\\/mobile\\.tupaia\\.org\\/)(country|facility)`

But I don't think it will make any difference realistically so feel free to skip it 😄

rohan-bes (approved):
Looks good, great stuff!

packages/database/src/migrations/20201026223059-UpdateLinkOnFeedItems-modifies-data.js
billli0: Thanks Rohan! It is nice to think carefully!

PR #1542, opened 2020-10-27
1305 project permission flow

tcaiger:
### Changes:

- Re-style RequestProjectAccess Modal
- Refactor OverlayDiv & SidePanel to be functional components
- Re-styled no data message on dashboards
---

### Screenshots:
Screenshots can be found in the issue.


packages/web-frontend/src/components/View/NoDataMessage.js
```diff
@@ -29,9 +38,16 @@ export const NoDataMessage = ({ viewContent }) => {
...
 };
 
 NoDataMessage.propTypes = {
   viewContent: PropTypes.shape(VIEW_CONTENT_SHAPE).isRequired,
+  severity: PropTypes.string,
+  AlertComponent: PropTypes.elementType,
+};
+
+NoDataMessage.defaultProps = {
+  AlertComponent: StyledAlert,
```
kael89, line 51: Clever!

packages/web-frontend/src/components/View/NoDataMessage.js
```diff
@@ -29,9 +38,16 @@ export const NoDataMessage = ({ viewContent }) => {
...
   }
 
-  return <StyledAlert severity="info">{message}</StyledAlert>;
+  return <AlertComponent severity={severity}>{message}</AlertComponent>;
 };
 
 NoDataMessage.propTypes = {
   viewContent: PropTypes.shape(VIEW_CONTENT_SHAPE).isRequired,
+  severity: PropTypes.string,
+  AlertComponent: PropTypes.elementType,
```
kael89, line 47: `NoDataMessage`, where `PlainNoDataMessage` is passed in for `AlertComponent`. We could consider

* Making this the default value here and delete `PlainNoDataMessage`
* Do not support passing a component in, but always use `PlainNoDataMessage`

But I'm fine if you think that the existing pattern is a good component abstraction practice

kael89 (approved):
Excellent work!!

packages/web-frontend/src/components/View/NoDataMessage.js
tcaiger: Thanks Kostas! well picked up. I will the second option and make it as simple as possible.

PR #1544, opened 2020-10-27
1451: Create test admin user

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1451

### Changes:
* Create test admin user, which is granted access to all top-level permissions for all countries
* Wait for `view` request instead of `dashboard` to tell when a report has loaded
* Refactor test user detail specification

### DIFF
--- a/packages/web-frontend/cypress.json
+++ b/packages/web-frontend/cypress.json
@@ -3,6 +3,7 @@
   "experimentalFetchPolyfill": true,
   "ignoreTestFiles": "snapshots.js",
   "integrationFolder": "cypress/e2e",
+  "requestTimeout": 20000,
   "useRelativeSnapshots": true,
   "viewportHeight": 900,
   "viewportWidth": 1400
--- a/packages/web-frontend/cypress.json
+++ b/packages/web-frontend/cypress.json
@@ -1,9 +1,9 @@
 {
   "baseUrl": "http://localhost:8088",
-  "integrationFolder": "cypress/e2e",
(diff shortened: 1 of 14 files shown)


packages/web-frontend/cypress.json
```diff
@@ -1,9 +1,10 @@
 {
   "baseUrl": "http://localhost:8088",
-  "integrationFolder": "cypress/e2e",
+  "experimentalFetchPolyfill": true,
   "ignoreTestFiles": "snapshots.js",
-  "viewportHeight": 900,
-  "viewportWidth": 1400,
+  "integrationFolder": "cypress/e2e",
+  "requestTimeout": 20000,
```
kael89, line 6: This is the only added field here (the rest are just sorted).

The original timeout is `5000`, which is certainly not enough for some reports. We may find out that we need to adjust this even more, or leave it there as a canary for reports taking too long to load

Related to this, a positive side effect of having e2e tests is that they can serve as a rough benchmarks:

image

packages/web-frontend/cypress/config/orgUnitMap.json
```diff
@@ -138,11 +138,6 @@
     "district": "TK_Atafu",
     "facility": "TK_Ataf"
   },
-  "TL": {
```
kael89, line 141: TL is disabled for all users currently

tcaiger (approved):
Great addition! Everything looks good to me.

PR #1546, opened 2020-10-28
Throw descriptive error message for absent .env

kael89:
### Issue #:
[No issue]


kael89:
@edmofro I assigned this to you directly because we were discussing about it the other day, and also I have a question for you related to this

packages/database/src/getConnectionConfig.js
```diff
@@ -3,27 +3,44 @@
...
+  database: process.env.CI_TEST_DB_NAME,
+  ssl: null,
+};
+
+const SERVER_CONNECTION_CONFIG = {
+  host: process.env.DB_URL,
+  user: process.env.DB_USER,
+  password: process.env.DB_PASSWORD,
+  database: process.env.DB_NAME,
+  ssl: ['true', undefined].includes(process.env.DB_DISABLE_SSL)
```
kael89, line 19: So here is a change: we currently set `ssl` to null only if `DB_DISABLE_SSL` is explicitly set to `true`. It seems to me that it should be the other way around: SSL should be disabled by default, and people can turn it on. But let me know if my understanding is wrong.

If this is correct, we may want to rename this to `DB_ENABLE_SSL` to make this "opt-in" behaviour clearer.

Also, I couldn't find a `.env` file in Last Pass where this was set to a value other than `true`, so I'm wondering whether it is used at all?

edmofro: It's set to false if you are developing locally but want to connect to a remote database. In future, we aim to move the database across to a service like RDS, so we'd also want to use an ssl connection then. 

I'd agree with your suggestion of the rename and default

packages/auth/.env.example
```diff
@@ -1,4 +1,4 @@
-DB_DISABLE_SSL=
+DB_ENABLE_SSL=
```
kael89, line 1: Not sure if we should include this optional field in the examples or just skip it, I chose to include it

edmofro: Yeah I think it's good to include for completeness. Please add a note that the release process needs to include some manual changes to environment variables (the process is somewhat arduous)

kael89: I will add it (by the way I am the release manager this week). Could you explain the process to me in person (or point me to related documentation)?

packages/database/src/getConnectionConfig.js
```diff
@@ -3,27 +3,45 @@
...
+          rejectUnauthorized: false,
+        }
+      : null,
+};
+
+const validateConfig = config => {
+  if (config.ssl) {
+    return;
+  }
+
```
edmofro, line 32:
```suggestion
```
If ssl is set, it doesn't guarantee we have the other variables set up correctly (and we still need to)

edmofro (approved):
Very nice breaking out two consts, that's much more clear.

Pre-approved with one suggested change

packages/database/src/getConnectionConfig.js
kael89: I wasn't sure about that and actually I should have left a question-comment! Thanks for catching this!

packages/auth/.env.example
edmofro: For sure. This is the best documentation so far, but I will take you through it in person and we can add a doc just for that process

edmofro: https://docs.beyondessential.com.au/books/dev-ops/page/tupaia-main-server-structure-overview#bkmrk-environment-variable

packages/database/src/getConnectionConfig.js
```diff
@@ -3,27 +3,40 @@
...
-        ssl:
-          process.env.DB_DISABLE_SSL === 'true'
-            ? null
-            : {
-                // Test server cannot turn on ssl, so sets the env to disable it
-                rejectUnauthorized: false,
-              },
-      };
+  validateEnv();
+  // Note: Must use functions to guarantee environment variables have loaded
```
kael89, line 40: Indeed this is the case... In previous commits, I removed the comment (and used `const`) without a problem, but I was probably lucky

This functionality now fails if functions are not used to retrieve the `.env` variables

edmofro (approved):

kael89:
@edmofro I'm going to merge this into `dev`, and follow the new [.env variable instructions](https://github.com/beyondessential/tupaia-backlog/issues/827#issuecomment-719098624) to remove `DB_SSL_DISABLE` (not required anymore) for the packages whose `.env.example` files were updated in this PR

edmofro:
Sounds good!

kael89:
@edmofro as discussed, it seems that in our current prod/dev instances (and LastPass), the following applies:

```
# meditrak-server
DB_DISABLE_SSL="false"
# web-config-server
DB_DISABLE_SSL="true"
```

Which implies that the presence of `DB_DISABLE_SSL` is significant in some cases. However, we have now removed support for this variable, but everything is working fine in `dev`. It may be that at some point (early stages of Tupaia development?) devs used SSL to connect to a database, and hence the need for this variable.

Given that we want to be able to connect to a remote DB, we can still retain support for `DB_ENABLE_SSL`, but remove it from all existing `.env` files since it is not currently required. The other option would be to completely remove it, and reintroduce it when required.

PR #1552, opened 2020-10-29
1278: Added 'report' table and model

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1278:
Also updated eslint config for reports package

### DIFF
--- a/packages/database/src/migrations/20201028234713-AddReportsTable-modifies-schema.js
+++ b/packages/database/src/migrations/20201028234713-AddReportsTable-modifies-schema.js
@@ -1,0 +1,36 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`
+    CREATE TABLE report (
+      id TEXT PRIMARY KEY,
+      code TEXT NOT NULL,
+      config JSONB NOT NULL,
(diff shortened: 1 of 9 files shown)


packages/database/src/migrations/20201028234713-AddReportsTable-modifies-schema.js
```diff
@@ -0,0 +1,36 @@
...
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`
+    CREATE TABLE report (
+      id TEXT PRIMARY KEY,
+      code TEXT NOT NULL,
+      config JSONB NOT NULL,
+      permission_group_id TEXT NOT NULL
```
kael89, line 23: 1. Let's add a unique `code` constraint
2. Let's make `permission_group_id` a foreign key

kael89 (changes requested):
Great work, just a few schema tweaks requested! Interesting to know that permissions can now be attached to individual reports

packages/database/src/migrations/20201028234713-AddReportsTable-modifies-schema.js
rohan-bes: Ah great point! Btw, do you think NOT NULL is appropriate? A a few of the other tables with this foreign key support NULL but I'm not sure what their behaviour is when it's missing. In my mind you either want the report to be guarded by permissions, or the permission group should just be public, so that's why I went with NOT NULL

kael89: Hm not sure, perhaps our code treats `NULL` as `Public` in which case it will be fine! On the other hand, using an explicit group will make things more straightforward in code (and TypeScript).

I guess that wasn't much help but I don't know the answer! 😛 If it's tricky to figure out, you can always start with a strict schema and loosen up later

kael89 (approved):
Nice on!

PR #1556, opened 2020-10-29
827: Tweak deployment process to allow for e2e test deployments

edmofro:


edmofro:
@kael89 note that tests are passing, it's just a migration that is failing as it depends on an admin panel data change that was done on prod more recently than e2e baseline instance was frozen

edmofro:
(which may become a general problem to solve)

packages/devops/scripts/utility/getEC2TagValue.sh
```diff
@@ -1,8 +1,8 @@
...
 
 # Get the stage tag of this ec2 instance from AWS
-TAG_NAME="Stage"
+TAG_NAME=$1
 INSTANCE_ID="$(wget -qO- http://instance-data/latest/meta-data/instance-id)"
 REGION="`wget -qO- http://instance-data/latest/meta-data/placement/availability-zone | sed -e 's:\([0-9][0-9]*\)[a-z]*\$:\\1:'`"
-STAGE="$(aws ec2 describe-tags --filters "Name=resource-id,Values=$INSTANCE_ID" "Name=key,Values=$TAG_NAME" --region $REGION --output=text | cut -f5)"
-echo $STAGE
+TAG_VALUE="$(aws ec2 describe-tags --filters "Name=resource-id,Values=$INSTANCE_ID" "Name=key,Values=$TAG_NAME" --region $REGION --output=text | cut -f5)"
+echo $TAG_VALUE
```
kael89, line 8: Nice generalisation!

packages/devops/scripts/deployment/deployLatestRepositories.sh
```diff
@@ -36,16 +45,9 @@ for PACKAGE in "meditrak-server" "web-config-server" "web-frontend" "admin-panel
...
-        SSM_PATH="/${PACKAGE}/dev"
-        $(aws ssm get-parameters-by-path --with-decryption --path $SSM_PATH |
-            jq -r '.Parameters| .[] | .Name + "=\"" + .Value + "\""  ' |
-            sed -e "s~${SSM_PATH}/~~" >.env)
-        # Replace any instances of "dev" in the .env file (e.g. to point urls in the right place)
-        sed -i -e "s/dev/${STAGE}/g" .env
-    fi
+    # Replace any instances of the placeholder [branch-name] in the .env file with the actual branch
+    # name (e.g. [branch-name]-api.tupaia.org -> specific-branch-api.tupaia.org)
+    sed -i -e "s/\[branch-name\]/${BRANCH}/g" .env
```
kael89, line 50: Clever idea using a placeholder!! I guess this has been updated in AWS SSM, and we don't need to update our Last Pass `.env` files? Since if `[branch-name]` is not found, then no replacement will occur, which is fine in case of `[DEV]` and `[PROD]` .env files

scripts/bash/syncDotEnvWithParameterStore.sh
```diff
@@ -0,0 +1,65 @@
...
+if [[ -z $PATH_TO_PACKAGE ]]; then
+    # Use default path to package, based on ubuntu server setup
+    PATH_TO_PACKAGE="/home/ubuntu/tupaia/packages/${PACKAGE_NAME}"
+fi
+
+if [[ $DIRECTION == "upload" ]]; then
+    ${DIR}/storeDotEnvOnParameterStore.sh ${PATH_TO_PACKAGE} ${PARAMETER_STORE_PATH_PREFIX}
+else
+    ${DIR}/fetchDotEnvFromParameterStore.sh ${PATH_TO_PACKAGE} ${PARAMETER_STORE_PATH_PREFIX}
+fi
```
kael89, line 65: 💯

kael89 (approved):
Amazing work @edmofro !!

Do we need to update any documentation, either in the repo or in BookStack (eg https://docs.beyondessential.com.au/books/dev-ops/page/updating-environment-variables)?

Also, I see that some `.py` scripts in `meditrak-server` are deleted - I guess we don't use those anymore

kael89:
> (which may become a general problem to solve)

Good point! I guess we have to skip `-modifies-data` migrations in e2e deployments... Hopefully that won't be too trickuy

edmofro:
The modifies data can be important in the context of having a frozen database with data in. They can be safely skipped on the codeship machine because it starts with zero data. But if we make a code change that involves, e.g. updating all the dashboard report configs to include an additional key, we need those data changing ones to run.

edmofro:
Regarding the updated documentation, I've copy pasted from my vscode to the issue - see https://github.com/beyondessential/tupaia-backlog/issues/827#issuecomment-719098624

packages/devops/scripts/deployment/deployLatestRepositories.sh
edmofro: Updating them for dev will need to be part of the release process. See https://github.com/beyondessential/tupaia-backlog/issues/827#issuecomment-719098624

scripts/bash/syncDotEnvWithParameterStore.sh
edmofro: I should have left a comment about motivation here. When we need to update an env var, we will need to change that variable, but also make sure that all the variables in the .env we're uploading contain `[branch-name]` rather than `dev`, so that the replacement process still works. I figured this would be a bit of a burden to do manually, and could result in mistakes, so I added this script that will pull down the env vars in their pure form as stored on SSM, without running the replacement regex. This means the steps to update become: pull from ssm, update _that_ file, push to ssm.

kael89: Very good call! Perhaps we can add a script that does all 3 steps

edmofro: Oh yeah that could be cool...and then also makes the replacements again at the end (which is a weakness of this as it stands)

kael89:
> The modifies data can be important in the context of having a frozen database with data in. They can be safely skipped on the codeship machine because it starts with zero data. But if we make a code change that involves, e.g. updating all the dashboard report configs to include an additional key, we need those data changing ones to run.

That's definitely true! In that case, we need to distinguish between `backbone` data (visualisations..) and `volatile` data (survey responses...). (Happy to introduce better terms than those here!) A couple of solutions to that could be:

* Educate devs to catch possible migration errors related to not found `volatile` data, and only throw them if the environment is not "e2e"
* Specify a list of tables for which "not found record" data should be ignored if env = "e2e" (this one may be trickier to do correctly in a generic way)

edmofro:
I think the former solution sounds good. Will raise it in dev chat

kael89:
A related problem is when people change `backbone` data via an admin panel: the change will not be reflected in the "e2e" environment

PR #1563, opened 2020-11-03
Fix e2e migrations breaking: delete any survey responses attached to the entity being deleted

edmofro:
While this is updating a migration that has already been released, a strict no-no, in this case it's ok because we don't care about the change running on production. In production, the migration will be marked as already run, so the change will have no effect. In e2e test environments, it will run with the change, so those environments will be fixed.


billli0 (approved):
Looks good!

PR #1568, opened 2020-11-04
1247 Add PSSS project and attach countries list with alternative hierarchy

EMcQ-BES:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/1247

### Changes:

- Add a migration file for creating PSSS project

### DIFF
--- a/packages/database/src/migrations/20201103092344-AddPSSSProject-modifies-data.js
+++ b/packages/database/src/migrations/20201103092344-AddPSSSProject-modifies-data.js
@@ -17,9 +17,9 @@ exports.setup = function (options, seedLink) {
 };
 
 const PROJECT_CODE = 'psss';
+const ENTITY_NAME = 'PSSS';
 const DASHBOARDGROUP_NAME = 'PSSS';
 const USER_GROUP = 'PSSS';
-const NZ_COUNTRY_CODE = 'NZ';
 
 const COUNTRY_LIST = [
   'AS',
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20201103092344-AddPSSSProject-modifies-data.js
```diff
@@ -0,0 +1,120 @@
...
+    code: NZ_COUNTRY_CODE,
+    name: 'New Zealand',
+  });
+
+  // Add project entity and entity hierarchy for PSSS
+  await insertObject(db, 'entity', {
+    id: generateId(),
+    code: PROJECT_CODE,
+    parent_id: await codeToId(db, 'entity', 'World'),
+    name: DASHBOARDGROUP_NAME,
```
edmofro, line 76: Kinda weird variable name to use here. It is the same though, so maybe just a variable rename?

edmofro (changes requested):
Code all lgtm, with one suggested rename. I'm wondering about adding a country through a migration rather than through the admin panel process of creating a country then importing entities (of just that country) through the admin panel though. That admin panel process has some side effects, such as grabbing the geojson for the country from and adding that to the database.

So, please change this to assume NZ exists, and then add a data steward task to add it through the admin panel.

An additional complexity: this would then be an example of a migration relying on something existing in the database that doesn't exist on the end to end test instance (frozen a while ago, so won't have NZ). So please also add the first case of graceful error handling if we're on the e2e instance. How do you know that? Not sure.

EMcQ-BES:
> So, please change this to assume NZ exists, and then add a data steward task to add it through the admin panel.

Edwin touched base with my via slack to confirm we can instead leave NZ off the list of countries for now. So I've removed it completely instead.

edmofro (approved):

PR #1570, opened 2020-11-04
Lazy load db

kael89:
### Issue #:
[No issue] Fixes the following error currently appearing in my local (haven't seen this in CI/CD yet)

This error is probably harmless (given the way it is triggered), but I thought that it's still a good idea to fix it. See my comment in `getModels.js` for more details

Weirdly enough, this doesn't come up in Codeship, but I still think it's a worthwhile change.


packages/data-broker/src/getModels.js
```diff
@@ -1,14 +0,0 @@
-/**
- * Tupaia
- * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
- */
-
-import { ModelRegistry, TupaiaDatabase } from '@tupaia/database';
-
-import { modelClasses } from './modelClasses';
-
-const database = new TupaiaDatabase();
```
kael89, line 10: So this is what was causing the issue...  The database will be unconditionally instantiated every times this file is loaded

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
```diff
@@ -4,26 +4,30 @@
  */
 
 import { DataBroker } from '../../DataBroker';
-import * as GetModels from '../../getModels';
 import { DATA_BY_SERVICE, DATA_ELEMENTS, DATA_GROUPS } from './DataBroker.fixtures';
 import { stubCreateService, createModelsStub, createServiceStub } from './DataBroker.stubs';
 
+const dataSources = Object.values(DATA_ELEMENTS).concat(Object.values(DATA_GROUPS));
+const mockModels = createModelsStub(dataSources);
```
kael89, line 11: This variable must start with `mock`, since it's used inside `jest.fn().mockImplementation` below. From the [jest docs](https://jestjs.io/docs/en/es6-class-mocks#calling-jestmockdocsenjest-objectjestmockmodulename-factory-options-with-the-module-factory-parameter):

> A limitation with the factory parameter is that, since calls to jest.mock() are hoisted to the top of the file, it's not possible to first define a variable and then use it in the factory. An exception is made for variables that start with the word 'mock'.

edmofro (approved):
Nice fix

kael89:
@edmofro I'm thinking this would require just a smoke test in a feature instance (actually making sure that data-broker is not fundamentally broken and reports load). Are you happy for me to do this in `dev`?

edmofro:
Yep

PR #1572, opened 2020-11-04
Set up Raw Data download for MS1 Administration

billli0:
### Issue #:
[1426 - Set up Raw Data download for MS1](https://github.com/beyondessential/tupaia-backlog/issues/1426)

### Changes:

Manual steps in Admin Panel:
1. Create new permission group  `MS1 Administration` under permission `Admin`
2. Add Kantabu to this permission group with Entity `Kiribati`

Then do migration:
1. Create new dashboard group for `MS1 Administration`
2. Create new dashboard report with surveys from survey group `Kiribati HIS`
3. Change map overlay `id: MS1_COMPLETENESS` permission to this new permission group


packages/database/src/migrations/20201104032225-AddDashboardGroupForMS1Administration-modifies-data.js
```diff
@@ -0,0 +1,104 @@
...
+        "organisationLevel" ,
+        "userGroup" ,
+        "organisationUnitCode" ,
+        "dashboardReports" ,
+        "name",
+        "code",
+        "projectCodes")
+        
+      VALUES (
+         ${dashboardGroupId},
```
chris-bes, line 34: I don't think you need to hardcode dashboardGroupId when inserting a new dashboard group. As I remember it is automatically generated. I think it's actually safer to let it self generated because other devs might insert a new dashboard group with the same id at the same time

packages/database/src/migrations/20201104032225-AddDashboardGroupForMS1Administration-modifies-data.js
```diff
@@ -0,0 +1,104 @@
...
+    UPDATE "mapOverlay" 
+    SET "userGroup" = 'MS1 Administration'
+    WHERE id = '${mapOverlayId}';
+  `);
+};
+
+exports.down = async function (db) {
+  await db.runSql(`
+    DELETE FROM "dashboardGroup" 
+    WHERE id = ${dashboardGroupId};
```
chris-bes, line 87: If you decide to remove the id when inserting dashboard group, we can change this to `WHERE code = ${dashboardGroupCode}`

chris-bes (approved):
Looks awesome! I just left a minor suggestion so pre-approve!

packages/database/src/migrations/20201104032225-AddDashboardGroupForMS1Administration-modifies-data.js
billli0: That's something that I didn't know. Thanks @chris-bes for sharing!

billli0:
Test note: Not sure whether it will affect anything after changing the map overlay permssion from 'Admin' to 'MS1 Administration'.

PR #1574, opened 2020-11-05
1492 improve e2e test config generation

kael89:
### Issue #:
http://github.com/beyondessential/tupaia-backlog/issues/1492

### Changes:
As per the card:
* Skip drill down reports
* _(Updated 6/11)_ Skip reports with no data builder
* Take `attributes` into account when selecting an org unit for a report

Additional:
* Moved `filterEntities` in `@tupaia/utils` and generalised a bit
* Changed a bit the warning logs
* Added and used some new utils

### Screenshots:
_(Updated 6/11)_

Screen Shot 2020-11-06 at 4 27 45 pm


packages/utils/src/array.js
```diff
@@ -33,3 +33,5 @@ export const max = array =>
   Array.isArray(array) && array.length > 0
     ? array.reduce((result, value) => Math.max(value, result), Number.NEGATIVE_INFINITY)
     : undefined;
+
+export const toArray = input => (Array.isArray(input) ? input : [input]);
```
kael89, line 37: I would be also keen to name this just `array`, but I went with the "safest" approach

packages/utils/src/filterEntities.js
```diff
@@ -0,0 +1,71 @@
...
+ * Example entityFilter format:
+ * ```
+ * {
+ *    code: 'ABC'
+ *    attributes: { type: 'Primary' },
+ * }
+ * ```
+ */
+export const filterEntities = (entities, entityFilter) => {
+  let filteredEntities = entities;
```
kael89, line 57: I generalised this function a bit since it is now part of `@tupaia/utils`. We could probably make this even more generic, something like:

`export const filterObjects = (objects, filter)`

The `switch` block below explicitly uses different filter logic (plain/object) for different field keys. In practice, we don't need to be aware of the object's schema: we can figure out if a field is an object by checking its value.

Also, I have the suspicion that `FILTER_TYPE_TO_METHOD` has some commonalities with `checkValueSatisfiesCondition`.

Finally, we probably can avoid using `let`, and rather use chained array methods (`every`/`filter` etc).

I decided to not do all this refactoring since it felt out of scope for this card, but now that I think of it we're gonna do a regression testing of this functionality anyway, so it may be worth it! But if I do, perhaps it should be done as a separate PR for ease of reviewing. I would love a second opinion so please let me know!

packages/web-frontend/cypress/scripts/generateDashboardReportConfig.js
```diff
@@ -117,9 +124,7 @@ const getDashboardReportIdToGroups = async database => {
  * that matches the group's org unit code and level
  */
 export const generateDashboardReportConfig = async ({ database, logger }) => {
-  const reports = await database.executeSql(
-    `SELECT id, "viewJson"->>'name' as name from "dashboardReport"`,
-  );
+  const reports = await database.executeSql(`SELECT id, "drillDownLevel" from "dashboardReport"`);
```
edmofro: I guess `name` wasn't being used

packages/web-frontend/cypress/support/helpers.js
```diff
@@ -21,6 +23,13 @@ export class EmptyConfigError extends Error {
   }
 }
 
+export class TestUserPasswordUndefinedError extends Error {
+  constructor() {
+    super('Please specify a value for CYPRESS_TEST_USER_PASSWORD in packages/web-frontend/.env');
+    this.name = 'TestUserPasswordUndefinedError';
```
edmofro, line 29: For my own interest - does setting a name on an error give nicer console logs?

packages/web-frontend/cypress/README.md
```diff
@@ -8,7 +8,12 @@
 
    ⚠️: In **Step 2. Install node dependencies** you need to run the commands under a Windows terminal, see the note in that section.
 
-2. Use a **[LOCAL]** `.env` file for `web-frontend`, so that it points to a local `web-config-server`
+2. Point `web-frontend` to a local instance of `web-config-server`. In `packages/web-frontend/.env`, set
```
edmofro, line 11: Presuming these instructions are targeted at running locally

packages/utils/src/array.js
edmofro: I prefer `toArray` :+1:

packages/web-frontend/cypress/scripts/generateDashboardReportConfig.js
kael89: Indeed! In the latest version of this file I just `SELECT *` (I have pushed some commits after the initial PR)

edmofro: Ignore this - it gets changed to `*` in a later commit.

packages/web-frontend/cypress/support/helpers.js
kael89: Yeap, the stack trace of an uncaught error will be like this:

image

instead of just `Error: ...`

edmofro: Very cool, and good to know

packages/utils/src/__tests__/object.test.js
```diff
@@ -359,4 +360,19 @@ describe('object', () => {
       });
     });
   });
+
+  describe('filterValues', () => {
+    const object = { a: 1, b: 2, alpha: 1 };
+
+    const testData = [
+      ['no entry passes', value => value === 3, {}],
```
edmofro, line 368: Could split this into two tests: `no entries in object`, and `no entry passes`

packages/utils/src/__tests__/object.test.js
```diff
@@ -359,4 +360,19 @@ describe('object', () => {
...
+      ['one entry passes', value => value === 2, { b: 2 }],
+      ['multiple entries pass', value => value === 1, { a: 1, alpha: 1 }],
```
edmofro, line 370: These two should have some negative entries in the object as well

packages/web-frontend/cypress/README.md
kael89: They do! But I guess it's a good idea to clarify that - will do

packages/utils/src/__tests__/object.test.js
kael89: Good idea!

packages/utils/src/__tests__/object.test.js
```diff
@@ -362,17 +362,21 @@ describe('object', () => {
   });
 
   describe('filterValues', () => {
-    const object = { a: 1, b: 2, alpha: 1 };
+    const object = { a: 1, b: 2, alpha: 1, negativeA: -1, negativeB: -2 };
 
     const testData = [
       ['no entry passes', value => value === 3, {}],
       ['one entry passes', value => value === 2, { b: 2 }],
       ['multiple entries pass', value => value === 1, { a: 1, alpha: 1 }],
```
edmofro: In these ones too

edmofro (approved):
🙌

packages/web-config-server/src/apiV1/dataBuilders/generic/latestData/multiDataValuesLatestSurvey.js
```diff
@@ -13,7 +13,7 @@ export const multiDataValuesLatestSurvey = async (
   );
   const latestPeriod = getMostRecentPeriod(surveyDatesResponseDataValues.map(srv => srv.period));
   if (!latestPeriod) {
-    return null;
+    return { data: [] };
```
kael89, line 16: @edmofro can I also sneak this fix in here? This is a current issue in `dev` and `prod` that the e2e tests found. It fixes the last report that consistently fails the tests because of a code error (at least with the current orgUnitMap) 🎉

This also made me think that we should probably also test reports when no data are present. One way to do so is provide an alternative orgUnitMap and also run the tests against that.

edmofro: Yes and good idea!

kael89:
@edmofro this is `testing passed` now on the feature branch, but I had to resolve some conflicts with `dev` in https://github.com/beyondessential/tupaia/pull/1574/commits/bedcb3eca8adbe95dcabc386cd4286b21b86befa. Feel free to have another look if you want!

edmofro:
lgtm :+1:

PR #1575, opened 2020-11-05
fix image uploads

tcaiger:
### Issue #: [960](https://github.com/beyondessential/tupaia-backlog/issues/960)

The issue was actually a regression caused by a previous feature for uploading profile images. It looks like that change was never actually needed and always saving the image as a png works fine. I regression tested the profile image uploads and they still work after this change.

### Changes:

- Fix s3 image uploads file types
- Fix create thumbnail logging
---


edmofro (approved):
lgtm

PR #1577, opened 2020-11-06
Include cases in hierarchies for STRIVE and Fanafana Ola

edmofro:
The bug @avaek identified in [this comment](https://github.com/beyondessential/tupaia-backlog/issues/478#issuecomment-719881621) exists because when a "case" is created, it isn't attached to any entity hierarchy (because cases aren't a canonical entity type; and there are no entity relation links joining it manually to a hierarchy).

This means that when it syncs, and we try to find an org unit ancestor to put it under on DHIS2, we can't find one because there are no ancestor_descendant_relation links at all.

Two changes in this PR fix that:
- A hierarchy can have a custom set of canonical types (have included cases for STRIVE and Fanafana Ola)
  - This can be reviewed a commit at a time, from the third commit onwards
- When we sync, if a case doesn't exist in the default "explore" hierarchy, find another hierarchy that does contain the case
  - All contained within the first commit


packages/database/src/migrations/20201105014950-AddEntityTypesToHierarchy-modifies-schema.js
```diff
@@ -0,0 +1,27 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.addColumn('entity_hierarchy', 'canonical_types', { type: 'text[]' });
```
kael89, line 18: We could default this to an empty array

packages/database/src/cachers/EntityHierarchyCacher.js
```diff
@@ -20,6 +20,9 @@ export class EntityHierarchyCacher {
     this.changeHandlerCancellers[1] = this.models.entityRelation.addChangeHandler(
       this.handleEntityRelationChange,
     );
+    this.changeHandlerCancellers[2] = this.models.entityHierarchy.addChangeHandler(
+      this.handleEntityHierarchyChange,
+    );
```
kael89, line 25: Although this is very readable, you can optionally turn it into a standard array assignment

```js
this.changeHandlerCancellers = [ ... ];
```

packages/meditrak-server/src/externalApiSync/ExternalApiSyncQueue.js
```diff
@@ -78,8 +78,16 @@ export class ExternalApiSyncQueue {
...
-    const changeDetails = await this.detailGenerator.generateDetails(validUpdates);
-    return this.persistToSyncQueue(validUpdates, changeDetails);
+    try {
+      const changeDetails = await this.detailGenerator.generateDetails(validUpdates);
+      return this.persistToSyncQueue(validUpdates, changeDetails);
+    } catch (e) {
+      // Something went wrong with generating change details, possibly because the entity hierarchy
+      // cache is still being built as the result of an entity change. Put this batch of changes
+      // back on the queue to attempt processing again
+      this.unprocessedChanges.push(...validUpdates);
```
kael89, line 88: The error handling here makes sense in the general case! Let's also keep in mind the cases where this could continuously fail for a legit reason. We may want to handle them in a different way (not retry).

One approach could be to retry a max number of times per batch of updates. This would also require assigning batches an id of sort.

We can probably ignore this for now and deal with it if such a situation arises

kael89 (approved):
Awesome work @edmofro !! The addition of custom canonical types per hierarchy feels very natural and useful!

Approving since most of my suggestions are optional. The only one I feel should probably be addressed is defaulting `canonical_types` to an empty array in the DB.

Let's also make sure that non org units types do not get exposed in the front-end (eg Strive cases do not show up in the Search Bar). I don't remember if the filtering is done in the back-end or in the front-end

packages/database/src/cachers/EntityHierarchyCacher.js
edmofro: Much more elegant!

packages/meditrak-server/src/externalApiSync/ExternalApiSyncQueue.js
edmofro: Yep it's a good point. We should notice if this happens for real - it would be the same as the current behaviour, which is that the sync queue would get blocked completely. This doesn't make things any worse ;-)

packages/database/src/modelClasses/Entity.js
```diff
@@ -169,11 +168,30 @@ export class EntityType extends DatabaseType {
...
+    const hierarchiesIncludingEntity = await this.otherModels.entityHierarchy.find(
+      {
+        ancestor_id: this.id,
+        [QUERY_CONJUNCTIONS.OR]: {
+          descendant_id: this.id,
+        },
+      },
+      {
+        joinWith: TYPES.ANCESTOR_DESCENDANT_RELATION,
+        joinCondition: ['entity_hierarchy.id', 'entity_hierarchy_id'],
```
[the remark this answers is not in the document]
edmofro: Oh wow, I forgot about that default behaviour! Will update.

packages/meditrak-server/src/externalApiSync/ExternalApiSyncQueue.js
kael89: Good point! An additional justification for not doing anything extra at the moment 🙌

packages/database/src/migrations/20201105014950-AddEntityTypesToHierarchy-modifies-schema.js
edmofro: Done

edmofro:
@kael89 merging to free up Andrew's testing, but please re-review after merge

edmofro:
Regarding "Let's also make sure that non org units types do not get exposed in the front-end (eg Strive cases do not show up in the Search Bar)."

They will get exposed. I think that's ok, so long as it's just within the STRIVE project?

kael89:
> Regarding "Let's also make sure that non org units types do not get exposed in the front-end (eg Strive cases do not show up in the Search Bar)."
> 
> They will get exposed. I think that's ok, so long as it's just within the STRIVE project?

I remember I had fixed this as a bug in the past (pre-project era, I believe) where the user would type something in the SearchBar and Strive cases would show up. I shouldn't necessarily be a blocker for testing this card at this stage, but before releasing it we should either address it or communicate it as a conscious change in functionality. If not, people may identify this as a bug.

It may (or may not...) be as simple as filtering the results that get returned in the front-end to include only org unit types. We also need to think whether similar issues may occur in other places where the front-end requests a hierarchy (not sure if others exist).

edmofro:
Yeah it's a good point about making sure the project is ok with this. I'll check in.

We couldn't make it as simple as only including org units, as that would exclude e.g. schools from Laos Schools.

kael89:
> Yeah it's a good point about making sure the project is ok with this. I'll check in.
> 
> We couldn't make it as simple as only including org units, as that would exclude e.g. schools from Laos Schools.

That's a very good point! Probably we need to define a new subset of entity types, which is the types that are displayed in the front-end. This can be defined repo-wide for now, and in a per-project basis if the need arises in the future

PR #1578, opened 2020-11-06
835 admin panel export modal

tcaiger:
### Changes:
- Refactor ExportModal to take jsx form fields as children
- Add SurveyResponsesExportModal
- Refactor Export modal to use local state instead of redux state
- Refactor Import modal to use local state instead of redux state
- Remove Countries Page (as it is no longer need since it only existed to export SurveyResponses)

---

### Screenshots:

Screen Shot 2020-11-06 at 4 57 17 PM


rohan-bes (approved):
Looks really good @tcaiger, nice work 😄

edmofro:
Reinstating `CountriesPage.js`, as we use it to create new countries in the latest dev, but have removed the export survey responses button from it

PR #1579, opened 2020-11-08
1379: Add checkConditions/boolean helper

chris-bes:
### Changes:

- Added expression-parser package which wraps around mathjs


packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -0,0 +1,110 @@
...
+    this.parser.set(name, value);
+  }
+
+  /**
+   * Sanitize any invalid variables in the scope and expression.
+   * Eg: variables that start with numbers
+   * @param {*} expression
+   * @param {*} originalScope
+   */
+  sanitizeExpression(expression, originalScope) {
```
chris-bes, line 62: One limitation of mathjs package is it does not allow you to have a variable name starting with `numbers` in an expression (https://github.com/josdejong/mathjs/issues/688#issuecomment-236675020). This is because mathjs supports implicit multiplication (eg: if we have expression `5abc > 3`, it will be parsed as `5 * abc > 3`). This is the default behaviour of mathjs parser and there's no way to turn this off. I tried forking the package, looking at the parser logic and it doesn't seem easy to do this too

In Tupaia I think we do have some cases when we want variables to start with numbers, examples:
- In survey question config, we always want reference questions using uuid so that we can get the answers from the state.
- Some question codes do start with numbers. And we probably always want to use question codes as variables in indicators/report package.

So I was torn with 2 solutions:
1. In any places that use this package, we just have to make sure when constructing expressions, we don't use any variables starting with numbers (eg: in the config of survey question, always add a prefix `$`, formula will look like '$5fa0cdcd5da1e614f30001fd+ $5fa0cdcd5da1e623c30001eg...')

2. Have this package automatically sanitize the variables and expression before actual evaluation.

2nd option seems like the best solution. The downside of 2nd option is we will not be able to implement a util that can get the variables from an expression (similar to `getVariables` in @arithmetic package, which is very handy). Mathjs does support a way of extracting variables from just an expression. But this is only accurate if we don't have any variables starting with numbers in the expression.

So the trade off for the 2nd option is when we construct an expression, we will always have to define a set of variable names expected in the expression. (Eg: in meditrak app, if we use @arithmetic package, we can just grab the variables from the formula and get the values of those variables. Whereas if we use this package, beside the formula, we will also need to store in the config something like:
```
config: {
	formula: '$5fa0cdcd5da1e614f30001fd + $5fa0cdcd5da1e623c30001eg',
	expectedVariables: ['$5fa0cdcd5da1e614f30001fd', '$5fa0cdcd5da1e623c30001eg'],
}
```

What do you reckon?

packages/expression-parser/src/expression-parser/BooleanExpressionParser.js
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * mathjs does not have any built in regex function.
+   * Import this to support matching regex
+   * Usage:
+   * parser.set('x', '5');
+   * parser.evaluate('matchRegex("x", "5")'); -> returns true
+   * parser.evaluate('matchRegex("x", "1")'); -> returns false
+   */
+  importMatchRegexFunction() {
```
chris-bes, line 29: Supported comparisons in `checkValueSatisfiesCondition,js`
```
const OPERATOR_TO_VALUE_CHECK = {
  '=': (value, target) => value === target,
  '>=': (value, target) => value >= target,
  '<=': (value, target) => value <= target,
  '>': (value, target) => value > target,
  '<': (value, target) => value < target,
  range: (value, target) => target[0] <= value && value <= target[1],
  rangeExclusive: (value, target) => target[0] < value && value < target[1],
  regex: (value, target) => !!value.match(target),
  in: (value, target) => target.includes(value),
};
```

mathjs supports all the above comparisons excepts regex. Luckily it allows us to import new util

full-testable-package.Dockerfile
```diff
@@ -64,6 +64,7 @@ COPY packages/database/. ./packages/database
 COPY packages/dhis-api/. ./packages/dhis-api
 COPY packages/indicators/. ./packages/indicators
 COPY packages/utils/. ./packages/utils
+COPY packages/expression-parser/. ./packages/expression-parser
```
edmofro, line 67: About 20 lines above you should also copy across the package.json

scripts/bash/getInternalDependencies.sh
```diff
@@ -1,2 +1,2 @@
 #!/bin/bash
-echo "access-policy" "aggregator" "auth" "database" "data-api" "data-broker" "dhis-api" "indicators" "utils" "ui-components" "weather-api"
+echo "access-policy" "aggregator" "auth" "database" "data-api" "data-broker" "dhis-api" "indicators" "utils" "ui-components" "weather-api" "expression-parser"
```
edmofro, line 2: alphabetical plz

tupaia-packages.code-workspace
```diff
@@ -43,6 +43,10 @@
     {
       "name": "export-chart-lambda",
       "path": "packages/export-chart-lambda"
+    },
+	{
```
edmofro, line 47: spacing here?

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Set the parser's scope.
+   * @param {*} scope
+   * @param {*} defaultScope
+   */
+  setScope(scope = {}, defaultScope = {}) {
+    Object.entries(scope).forEach(([name, value]) => {
+      const expressionValue = value || defaultScope[name] || 0;
+      this.set(name, expressionValue);
```
edmofro, line 43:
```suggestion
      this.parser.set(name, expressionValue);
```
And delete the one below - doesn't seem worth its own function to me?

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -0,0 +1,110 @@
...
+
+    return result;
+  }
+
+  /**
+   * Set the parser's scope.
+   * @param {*} scope
+   * @param {*} defaultScope
+   */
+  setScope(scope = {}, defaultScope = {}) {
```
edmofro, line 40:
```suggestion
  setScope(scope = {}) {
```
Unused?

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -0,0 +1,110 @@
...
+    // In Tupaia, we sometimes want to support variables starting with numbers like data code or question id.
+    // So to support it all the time, add a prefix $ to prevent implicit multiplication parsing
+    if (this.startsWithNumber(name)) {
+      sanitizedName = `$${sanitizedName}`;
+    }
+
+    return sanitizedName;
+  }
+
+  startsWithNumber(s) {
```
edmofro, line 99: I feel this should be a util above rather than within the class. `this.startsWithNumber` sounds like you're asking something of the class instance to me

packages/expression-parser/src/expression-parser/BooleanExpressionParser.js
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { ExpressionParser } from './ExpressionParser';
+
+export class BooleanExpressionParser extends ExpressionParser {
+  constructor() {
+    super();
+    this.importFunctions();
+  }
```
edmofro, line 12: If this gets extended to any other `ExpressionParser`s, I'd suggest we have a stub `importFunctions` in the base class, so the children don't have to override the constructor

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -0,0 +1,110 @@
...
+    return s.match(/^\d/);
```
edmofro, line 100: Should we allow actual number constants to go through as numbers? I'm not entirely sure if our use cases need this... but perhaps we should check for `/^\d.*[^\d]+` (a number followed by at least one non number - better check my working though)

edmofro: Your tests actually test this - so I guess that is working already and I just don't understand regex!

packages/expression-parser/src/expression-parser/ExpressionParser.js
edmofro: A couple of questions:
- @rohan-bes because of the way these are sanitized, in the data processing pipeline within a report builder, would you have to remember to use `$.row.$5fa0` rather than `$.row.5fa0` in any steps afterwards, or does the context of the expression only last for the single transform step, then we get the result and never see the sanitized variable names again?
- could we implement a `getVariables` which gets the variables from mathjs and then does a desanitize step on the variable names?
- I wonder if we should put a `$` in front of all of the variables in the `scope`, not just those with a leading number? Could cause less confusion if it's a hard rule that we can always expect

packages/expression-parser/src/expression-parser/ExpressionParser.js
edmofro: Ah right, it only checks those that are passed in through `scope`! The others are assumed to be constants.

edmofro (changes requested):
Very clean object oriented design @chris-bes - nice work :-)

packages/expression-parser/src/expression-parser/ExpressionParser.js
chris-bes: @edmofro I can answer the last 2 points:

> could we implement a getVariables which gets the variables from mathjs and then does a desanitize step on the variable names?

If what you meant is similar to `getVariables(formula)` in @beyondessential/arithmetic package (which just extracts the variables from the formula without any other info) then yes it's possible. But we have to make sure that the variables in the formula do not start with numbers (Eg: `getVariables('$5fa0 + $4fa1')` -> returns [`$5fa0`, `$4fa1`]). If `getVariables('5fa0 + 4fa1')`, it will return [`fa0`, `fa1`].

> I wonder if we should put a $ in front of all of the variables in the scope, not just those with a leading number? Could cause less confusion if it's a hard rule that we can always expect

This sounds better to me too!

edmofro: Yeah I'm suggesting that we:
- start with `tupaiaExpressionParserInstance.getVariables('5fa0 + 4fa1')`
- sanitize the expression, using your existing logic, creating `'$5fa0 + $4fa1'`
- run mathjs' get variables function, which returns `['$5fa0', '$4fa1']`
- desanitize, stripping the leading `$` -> `['5fa0', '4fa1']`
- return that

Would that work?

chris-bes: > sanitize the expression, using your existing logic, creating '$5fa0 + $4fa1'

Yep that would work, but only if the scope already has `5fa0` and `4fa1` variables set in it. This is because the existing sanitize expression logic relies on the variables set in the scope to sanitize the expression (otherwise I don't think there's a way to know what are the variables expected in the formula, unless we use regex which can be complicated since mathjs has a lot of built in syntax)

edmofro: I see what you mean 😞

rohan-bes: > @rohan-bes because of the way these are sanitized, in the data processing pipeline within a report builder, would you have to remember to use $.row.$5fa0 rather than $.row.5fa0 in any steps afterwards, or does the context of the expression only last for the single transform step, then we get the result and never see the sanitized variable names again?

I believe that `$.row.5fa0` should be fine. 
1. As you say, the expression context is refreshed between each transform step.
2. Even if it weren't, when an object is added to the expression scope (as we do with the `$` object currently in the reporting package) it seems that you are able to reference the children fine even if their names start with a number. ie. `$.row.5fa0` works fine without any extra sanitization

>In any places that use this package, we just have to make sure that when constructing expressions, we don't use any variables starting with numbers (eg: in the config of survey question, always add a prefix $, formula will look like '$5fa0cdcd5da1e614f30001fd+ $5fa0cdcd5da1e623c30001eg...'). This seems very inconvenient and ugly...

@chris-bes I don't necessarily think this is too ugly... Though I don't like the idea of us being forced into using such a syntax simply because of this mathjs implicit multiplication logic...

packages/expression-parser/src/expression-parser/ExpressionParser.js
chris-bes: I was thinking so that when setting a scope value outside of this package, we don't have to do something like `expressionParser.parser.set('a', 5)`. But it's very trivial...

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -89,22 +101,20 @@ export class ExpressionParser {
...
-
   /**
    * Clear the scope of the parser.
    * This will not remove any imported functions.
    */
   clearScope() {
     this.parser.clear();
   }
+
+  importFunctions() {}
```
edmofro, line 119: Perhaps a comment that this can be overridden if needed

packages/expression-parser/src/expression-parser/ExpressionParser.js
edmofro: Oh true. Do you think that will actually be used? I would have imagined the use cases to either be one of the following, rather than setting scope a field at a time:
````
const scope = { a: 1, b: 2 };
const expression1 = 'a + b';
const expression2 = 'a x b';

// option 1
const parser = new ExpressionParser(scope);
parser.evaluate(expression1);
parser.evaluate(expression2);

// option 2
const parser = new ExpressionParser();
parser.setScope(scope);
parser.evaluate(expression1);
parser.evaluate(expression2);

// option 3
const parser = new ExpressionParser();
parser.evaluate(expression1, scope); // order may be reversed
parser.evaluate(expression2, scope);
```

packages/expression-parser/src/expression-parser/ExpressionParser.js
edmofro: I agree on both points. @rohan-bes and @chris-bes I think this is the major point to resolve before this PR can be approved - could you two work on it together?

rohan-bes: @edmofro actually I believe I just lied accidentally, I double checked this and `$.row.5fa0` does actually fail unfortunately 😞 You can use `$.row['5fa0']` fine, however that's not intuitive at all...

@chris-bes let's organise some time today to work out a solution for this?

chris-bes: Had a discussion with Rohan about this. We agree that:
- The most ideal way to solve this is to modify the parser logic in mathjs to disable the implicit multiplication feature so that we can freely use variables starting with numbers. We would never want to use implicit multiplication in our expressions as well.
- The sanitizing expression approach is probably not the best way and can introduce complications in the future.

Given there's time constraint because both the reporting package and CalculatedQuestion need this soon. I think for now we can:
- Go with the 1st proposed approach above which is just prefix variables with `$` when constructing expressions and remove the sanitizing logic. Also, in this `expression-parser` package, we can add a validate method to detect if there's any implicit multiplication found in the expression and throw an error so that the users know.
- Create a technical debt ticket to fork the mathjs package and disable the implicit multiplication feature? (I'm happy to volunteer to work on it if we agree with this approach).

What do you reckon @edmofro

edmofro: Yeah great conclusions. Thanks guys.

edmofro (approved):
It might be useful to make sure the dev finds it easy to understand why the implicit multiplication occurred:
- add a note to the readme or comment somewhere saying that variables shouldn't start with a number
- tweak the error message to include "Perhaps you used a variable starting with a number?" or similar as a quick hint

Otherwise, this looks great :-)

chris-bes:
@edmofro Should we merge this one straight to dev and it can be a `no testing required` as this is just an independent package? If so then @rohan-bes can merge dev back to the reporting package and use it

edmofro:
Yep, sounds good!

PR #1584, opened 2020-11-09
Fix getDataValuesInSets

Alastair-L:
### Issue #:
No issue

Fixes a bug introduced here: https://github.com/beyondessential/tupaia/pull/1442/files#diff-f74381228000d462c6a6d7428aacab2b375439b83e439ecac230a2cd6abc3eecR44-R48

PR #1442, opened 2020-10-06
1050 district details table laos

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1050
### Changes:
2 new pieces of functionality:
- `$countDescendantsMatchingCondition`
- `includeSiblingData`: A key for entity aggregation

Also a very minor refactor of `fetchDataSourceEntities` in order to make it easier to follow.

PR #1585, opened 2020-11-09
Fix getDataValuesInSets

Alastair-L:
### Issue #:
No issue

Fixes a bug introduced here: https://github.com/beyondessential/tupaia/pull/1442/files#diff-f74381228000d462c6a6d7428aacab2b375439b83e439ecac230a2cd6abc3eecR44-R48

Just a rename of: https://github.com/beyondessential/tupaia/pull/1584

### DIFF
--- a/packages/web-config-server/src/dhis/DhisApi.js
+++ b/packages/web-config-server/src/dhis/DhisApi.js
@@ -27,7 +27,9 @@ export class DhisApi extends BaseDhisApi {
       throw new Error('Cannot getDataValuesInSets until fetchDataSourceEntities is injected');
     }
 
-    const dataSourceEntities = await this.fetchDataSourceEntities(entity, Entity.FACILITY);
+    const dataSourceEntities = await this.fetchDataSourceEntities(entity, {
+      dataSourceEntityType: Entity.FACILITY,
+    });
     const organisationUnitCodes = dataSourceEntities.map(e => e.code);
 
     return super.getDataValuesInSets({ ...query, organisationUnitCodes });


edmofro:
This really needs a hot-fix to master right? Are there any cases where it really does break things?

edmofro (approved):

Alastair-L:
I'm not too sure. I checked about 5 reports (and the MS1_COMPLETENESS overlay) and they all seemed to work in both versions. 

In fact I don't know how this particular report doesn't break here: the entity is the country so the nearest org unit descendants should be the provinces, rather that the desired facilities. 
However as the screenshot shows it works on prod at the moment...
image
image


I haven't worked with this functionality before so I'm not sure if there are any tricks... do you know why it is working at the moment? I don't know if I want to merge it without understanding it, but hopefully you can shed some light on what is going on (or just let me know the next steps).

edmofro:
Oh yeah, cool. So that will be DHIS2 being our friend - it returns all data at or below the specified org unit by default. We used to utilise this a lot, but with alternative hierarchies we no longer can. It explains why it's still working though - in these reports we're probably not in an alternative hierarchy right? So the DHIS2 hierarchy is picking up all of the correct data from descendants.

Alastair-L:
Ahh that makes sense then. In that case, for it to break, we would need to be 
a) At an inconvenient level, and
b) Within an alternative hierarchy

I can't find an example of that, which suggests to me that we would be fine merging this to dev first as a matter of principal?

Shall I create a card for testing it, or do you not think it needs it?

edmofro:
_or_ use a data service other than DHIS2 analytics (e.g. internal data)

If still no examples, merging to dev first sounds fine.

Don't worry about testing.

Alastair-L:
Good point! Almost forgot that, thanks for picking it up

I did another check with that in mind and still couldn't find anything (at least according to the `data_source` table).

Will squash and merge.

PR #1586, opened 2020-11-09
1358 fix login form renders twice

tcaiger:
### Issue #: [1358 fix login form renders twice](https://github.com/beyondessential/tupaia-backlog/issues/1358)

I tested the login form and it no longer loads twice. I believe the changes I made to the App.js and login flow in the sagas fixed this ([1528](https://github.com/beyondessential/tupaia/pull/1528)). However it could have been something else that fixed it - I'm not 100% sure to be honest but the main thing is that it works. I removed the hack that was in place in the end-to-end tests and ran them and the login flow works as expected.

**Notes on End To End Testing**
Setting up and running the tests went really smoothly for me. It is awesome to see these tests in working - great work @kael89! On the dev branch the tests were failing after login but on the most up to date branch [1574](https://github.com/beyondessential/tupaia/pull/1574) the tests ran perfectly (I stopped the tests after about 5 minutes).

I did note an issue with the `CYPRESS_TEST_USER_PASSWORD` env variable. It was wrongly named in the `actions.js` (which I've fixed in this PR) and also cypress was not picking up the password variable from my `.env` file after I added it. When I added the env variable to the `cypress.json` file it worked ([as described here](https://docs.cypress.io/guides/guides/environment-variables.html#Option-1-configuration-file)). 

PS. When I ran `cypress:generate-config` the password variable worked fine from the `.env` file.

### DIFF
--- a/packages/web-frontend/cypress/support/actions.js
+++ b/packages/web-frontend/cypress/support/actions.js
@@ -6,7 +6,7 @@
 import { TEST_USER } from '../constants';
 
 const getTestUserPassword = () => {
-  const password = Cypress.env('CYPRESS_TEST_USER_PASSWORD');
+  const password = Cypress.env('TEST_USER_PASSWORD');
   if (!password) {
     throw new Error(
       'Please specify a value for CYPRESS_TEST_USER_PASSWORD in packages/web-frontend/.env',
--- a/packages/web-frontend/cypress/support/actions.js
+++ b/packages/web-frontend/cypress/support/actions.js
@@ -6,7 +6,7 @@
 import { TEST_USER } from '../constants';
 
 const getTestUserPassword = () => {
-  const password = Cypress.env('TEST_USER_PASSWORD');
+  const password = Cypress.env('CYPRESS_TEST_USER_PASSWORD');
   if (!password) {
     throw new Error(
       'Please specify a value for CYPRESS_TEST_USER_PASSWORD in packages/web-frontend/.env',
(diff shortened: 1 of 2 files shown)


kael89 (changes requested):
Great to know that this is not required anymore! Just one point

packages/web-frontend/cypress/support/actions.js
```diff
@@ -6,7 +6,7 @@
 import { TEST_USER } from '../constants';
 
 const getTestUserPassword = () => {
-  const password = Cypress.env('TEST_USER_PASSWORD');
+  const password = Cypress.env('CYPRESS_TEST_USER_PASSWORD');
```
kael89, line 9: Did this work for you? It doesn't for me...

[Cypress.env](https://docs.cypress.io/api/cypress-api/env.html#Syntax) loads environment variables prefixed with `CYPRESS_`, but the prefix is omitted when retrieving the variable. We use [cypress-dotenv](https://www.npmjs.com/package/cypress-dotenv) to load those from a `.env` file, but the same convention still applies

kael89 (approved):
Feel free to merge straight into `dev` as this is dev related

packages/web-frontend/cypress/support/actions.js
tcaiger: Thanks for the explanation. My bad, I thought that it didn't work initially but I have re-tested it this morning and it works perfectly with the environment variables set up as you have them. I will put that down to my user error 👍

PR #1588, opened 2020-11-09
1132: Add create handler for /userEntityPermissions route

EMcQ-BES:
### Changes:

- Add CreateUserEntityPermissions.js
- Add unit tests for user entity permissions CRUD handling

### DIFF
--- a/packages/meditrak-server/src/routes/userEntityPermissions/CreateUserEntityPermissions.js
+++ b/packages/meditrak-server/src/routes/userEntityPermissions/CreateUserEntityPermissions.js
@@ -1,0 +1,39 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { CreateHandler } from '../CreateHandler';
+import {
+  assertAnyPermissions,
+  assertBESAdminAccess,
+  assertTupaiaAdminPanelAccess,
+} from '../../permissions';
+import { assertUserEntityPermissionUpsertPermissions } from './assertUserEntityPermissionPermissions';
+
+/**
+ * Handles POST endpoints:
+ * - /userEntityPermissions
+ */
+
+export class CreateUserEntityPermissions extends CreateHandler {
+  async assertUserHasAccess() {
+    await this.assertPermissions(
+      assertAnyPermissions(
+        [assertBESAdminAccess, assertTupaiaAdminPanelAccess],
+        'You need either BES Admin or Tupaia Admin Panel access to create user entity permissions',
+      ),
+    );
+  }
+
+  async createRecord() {
+    // Check Permissions
+    const userEntityPermissionsChecker = accessPolicy =>
+      assertUserEntityPermissionUpsertPermissions(accessPolicy, this.models, this.newRecordData);
+    await this.assertPermissions(
+      assertAnyPermissions([assertBESAdminAccess, userEntityPermissionsChecker]),
+    );
+
+    return this.insertRecord();
+  }
+}
--- a/packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
+++ b/packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
@@ -39,11 +39,9 @@ export const assertUserEntityPermissionEditPermissions = async (
   await assertUserEntityPermissionUpsertPermissions(accessPolicy, models, updatedFields);
 
   // Final check to make sure we're not editing a BES admin access permission
-  // There's nothing we could edit that couldn't be abused
+  // Changing any of the pieces of data in a BES admin UEP is abusable, so completely forbid it
   const userEntityPermission = await models.userEntityPermission.findById(userEntityPermissionId);
-  const permissionGroup = await models.permissionGroup.findById(
-    userEntityPermission.permission_group_id,
-  );
+  const permissionGroup = await userEntityPermission.permissionGroup();
   if (permissionGroup.name === BES_ADMIN_PERMISSION_GROUP) {
     throw new Error('Need BES Admin access to make this change');
   }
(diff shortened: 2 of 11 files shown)


packages/meditrak-server/src/routes/userEntityPermissions/CreateUserEntityPermissions.js
```diff
@@ -0,0 +1,39 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { CreateHandler } from '../CreateHandler';
+import {
+  assertAnyPermissions,
+  assertBESAdminAccess,
+  assertTupaiaAdminPanelAccess,
+} from '../../permissions';
+import { assertUserEntityPermissionUpsertPermissions } from './assertUserEntityPermissionPermissions';
```
edmofro, line 12: Great file name

packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
```diff
@@ -23,6 +27,52 @@ export const assertUserEntityPermissionPermissions = async (
...
+  userEntityPermissionId,
+  updatedFields,
+) => {
+  // Check we have permission to access the record we're trying to edit
+  await assertUserEntityPermissionPermissions(accessPolicy, models, userEntityPermissionId);
+  // Check we have permission for the changes
+  await assertUserEntityPermissionUpsertPermissions(accessPolicy, models, updatedFields);
+
+  // Final check to make sure we're not editing a BES admin access permission
+  // There's nothing we could edit that couldn't be abused
```
edmofro, line 42:
```suggestion
  // which is the only thing we could edit that could be abused
```
I didn't quite get the comment - maybe it means this?

packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
```diff
@@ -23,6 +27,52 @@ export const assertUserEntityPermissionPermissions = async (
...
+  const userEntityPermission = await models.userEntityPermission.findById(userEntityPermissionId);
+  const permissionGroup = await models.permissionGroup.findById(
```
edmofro, line 44:
```suggestion
  const permissionGroup = await userEntityPermission.permissionGroup();
```
We often have this pattern when there's a foreign key relationship (have to manually set it up, so you need to check the model class each time, but in this case it definitely exists)

packages/meditrak-server/src/routes/userEntityPermissions/assertUserEntityPermissionPermissions.js
```diff
@@ -23,6 +27,52 @@ export const assertUserEntityPermissionPermissions = async (
...
+    if (permissionGroup.name === BES_ADMIN_PERMISSION_GROUP) {
+      throw new Error('Need BES Admin access to make this change');
+    }
+  }
+  if (entityId) {
+    const entity = await models.entity.findById(entityId);
+    if (!accessPolicy.allows(entity.country_code, TUPAIA_ADMIN_PANEL_PERMISSION_GROUP)) {
+      throw new Error('Need Admin Panel access to the updated entity');
+    }
+  }
```
edmofro, line 73: All of the potential abuse cases are very nicely handled. I am a bit wary about the number of individual db calls we're adding in every time someone hits an endpoint, but we can optimise when we experience performance issues rather than making this code uglier now

packages/meditrak-server/src/tests/routes/userEntityPermissions/DeleteUserEntityPermissions.test.js
```diff
@@ -0,0 +1,111 @@
...
+      it('Throw a permissions gate error if we do not have BES admin or Tupaia Admin panel access anywhere', async () => {
+        const policy = {
+          DL: ['Public'],
+        };
+        await prepareStubAndAuthenticate(app, policy);
+        const { body: result } = await app.delete(
+          `userEntityPermissions/${vanuatuPublicPermission.id}`,
+        );
+
+        expect(result).have.to.keys('error');
```
edmofro, line 81: Perhaps also check the record still exists in the db

packages/meditrak-server/src/tests/routes/userEntityPermissions/EditUserEntityPermissions.test.js
```diff
@@ -0,0 +1,181 @@
...
+        expect(result.user_id).to.equal(userAccountId2);
+      });
+      it('Edit any user entity permission as BES Admin', async () => {
+        await prepareStubAndAuthenticate(app, BES_ADMIN_POLICY);
+        await app.put(`userEntityPermissions/${kiribatiBESPermission.id}`, {
+          body: {
+            user_id: userAccountId2,
+          },
+        });
+        const result = await models.userEntityPermission.findById(kiribatiBESPermission.id);
```
edmofro, line 175: Yeah good case to test, that BES can even edit BES

packages/meditrak-server/src/tests/routes/userEntityPermissions/CreateUserEntityPermissions.test.js
```diff
@@ -0,0 +1,149 @@
...
+          body: {
+            user_id: userAccountId,
+            permission_group_id: publicPermissionGroupId,
+            entity_id: laosEntityId,
+          },
+        });
+
+        expect(result).to.have.keys('error');
+      });
+      it('Throw an exception when trying to create a user entity permission with BES admin access when current user lacks BES admin access', async () => {
```
edmofro, line 98: prettier should tell you to put spaces between cases

edmofro (changes requested):
Along with a few comments inline:
- same comments as the previous PR
- it would be nice to do more checking of the db state after the endpoint has been handled, whether positively or negatively (nervous that one day we'll include the key `error` in the result as an empty string or something when there was no error, and then these tests could give false positives)

edmofro (approved):

PR #1593, opened 2020-11-10
Exclude cases from showing on tupaia frontend

edmofro:
Relates to https://github.com/beyondessential/tupaia/pull/1577#issuecomment-724264862

### DIFF
--- a/packages/web-config-server/src/apiV1/organisationUnit.js
+++ b/packages/web-config-server/src/apiV1/organisationUnit.js
@@ -27,7 +27,10 @@ export default class extends RouteHandler {
     const projectEntity = await project.entity();
     const country = await this.entity.countryEntity();
     const countryDescendants = country
-      ? await country.getDescendants(project.entity_hierarchy_id)
+      ? await country.getDescendants(project.entity_hierarchy_id, {
+          comparator: 'not in',
+          comparisonValue: this.models.entity.typesExcludedFromWebFrontend,
+        })
       : [];
     const orgUnitHierarchy = await this.filterForAccess([
       projectEntity,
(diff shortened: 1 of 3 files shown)


packages/web-config-server/src/apiV1/organisationUnit.js
```diff
@@ -63,11 +68,13 @@ export default class extends RouteHandler {
   }
 
   async filterForAccess(entities) {
-    return (await Promise.all(
-      entities
-        .filter(entity => entity)
-        .map(async entity => (await this.checkUserHasEntityAccess(entity)) && entity),
-    )).filter(entity => entity);
+    return (
```
edmofro, line 71: Prettier made this change

packages/web-config-server/src/apiV1/organisationUnitSearch.js
```diff
@@ -42,7 +42,12 @@ export default class extends RouteHandler {
     const project = await this.fetchProject();
     const projectEntity = await project.entity();
 
-    const allEntities = await projectEntity.getDescendants(project.entity_hierarchy_id);
+    const allEntities = await projectEntity.getDescendants(project.entity_hierarchy_id, {
+      type: {
+        comparator: 'not in',
+        comparisonValue: this.models.entity.typesExcludedFromWebFrontend,
+      },
+    });
```
kael89, line 50: This query is the same with the other one in `organisationUnit.js`, does it make sense to extract them in a common method (eg `getDescendantsForFrontEnd`)?

From one hand, we shouldn't go very far re-defining methods for every param combination, on the other this encapsulates specific business logic so it's probably worth it. I will leave the decision to you! 😎

kael89 (approved):
Thanks for taking care of this @edmofro !

packages/web-config-server/src/apiV1/organisationUnitSearch.js
edmofro: Yeah it's a question I thought about, but only for 30 seconds, and then decided it's simple enough to repeat both places (and actually reads fairly fluently to my eyes)

PR #1594, opened 2020-11-10
Not to rull out the first row

billli0:
### Issue #:
[Villages out of order #1422](https://github.com/beyondessential/tupaia-backlog/issues/1422)
### Changes:

- Not to rull out the first row


nick-beyondessential (approved):
Nice work @billli0

PR #1597, opened 2020-11-10
Add fiji districts to WISH entity relations

nick-beyondessential:
Add Fiji districts to WISH entity_relations to fix WISH custom downloads

### DIFF
--- a/packages/database/src/migrations/20201110210258-AddFijiDistrictsToEntityRelation-modifies-data.js
+++ b/packages/database/src/migrations/20201110210258-AddFijiDistrictsToEntityRelation-modifies-data.js
@@ -1,0 +1,48 @@
+'use strict';
+
+import { arrayToDbString, generateId, insertObject } from '../utilities';
+
+var dbm;
+var type;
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20201110210258-AddFijiDistrictsToEntityRelation-modifies-data.js
```diff
@@ -0,0 +1,48 @@
...
+      child_id: district,
+      entity_hierarchy_id: hierarchyId,
+    }),
+  );
+};
+
+exports.down = function (db) {
+  return db.runSql(
+    `DELETE FROM entity_relation 
+    WHERE child_id IN (${arrayToDbString(districts)});`,
```
edmofro, line 42: I'm approving since we don't use downs, but it'd be safer here to check the hierarchy id as well

edmofro (approved):
Nice Nick! Hopefully this fixes the issue?

edmofro (approved):

PR #1601, opened 2020-11-11
1281: Auth handling for Reporting package

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1281:

Worked out pretty cleanly in the end, thanks to our neatly abstracted auth and access-policy packages 😄 

For now, just supporting auth via username/pass, however in the future when we wire this up with the PSSS backend, we need to add support for auth via user token. This should be fairly easy to do, and can be copied from meditrak-server 👍

### DIFF
--- a/packages/reports/examples.http
+++ b/packages/reports/examples.http
@@ -1,0 +1,9 @@
+# Fetch Reports
+
+POST http://localhost:8030/v1/fetchReport/8?organisationUnitCode=TO HTTP/1.1
+content-type: application/json
+
+{
+    "emailAddress": "username",
+    "password": "password"
+}
--- a/packages/reports/src/app/addRoutesToApp.ts
+++ b/packages/reports/src/app/addRoutesToApp.ts
@@ -13,7 +13,7 @@ export function addRoutesToApp(app) {
   app.use(authenticationMiddleware);
 
   /**
-   * GET routes
-   **/
+   * POST routes
+   */
   app.post('(/v[0-9]+)?/fetchReport/:reportCode', fetchReport);
 }
--- a/packages/reports/src/app/addRoutesToApp.ts
+++ b/packages/reports/src/app/addRoutesToApp.ts
@@ -4,10 +4,16 @@
  */
 
 import { fetchReport } from '../routes';
+import { authenticationMiddleware } from '../auth';
 
 export function addRoutesToApp(app) {
   /**
+   * Attach authentication to each endpoint
+   */
+  app.use(authenticationMiddleware);
+
+  /**
    * GET routes
    **/
-  app.get('(/v[0-9]+)?/fetchReport/:reportCode', fetchReport);
+  app.post('(/v[0-9]+)?/fetchReport/:reportCode', fetchReport);
 }
--- a/packages/reports/src/routes/fetchReport.ts
+++ b/packages/reports/src/routes/fetchReport.ts
@@ -5,10 +5,11 @@
 
 import { respond } from '@tupaia/utils';
 import { createAggregator } from '@tupaia/aggregator';
-import { Request, Response } from 'express';
+import { Response } from 'express';
 import { ParamsDictionary, Query } from 'express-serve-static-core';
 import { Aggregator } from '../aggregator';
 import { BuildReport, ReportBuilder } from '../reportBuilder';
+import { ReportsRequest } from '../types';
 
 export interface FetchReportQuery extends Query {
   organisationUnitCode: string;
@@ -26,7 +27,7 @@ class FetchReportRouteHandler {
     this.aggregator = createAggregator(Aggregator);
   }
 
-  fetchReport = async (req: Request, res: Response): Promise<void> => {
+  fetchReport = async (req: ReportsRequest, res: Response): Promise<void> => {
     const params = req.params as FetchReportParams;
     const query = req.query as FetchReportQuery;
     const report = await req.models.report.findOne({ code: params.reportCode });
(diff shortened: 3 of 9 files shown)


packages/reports/examples.http
```diff
@@ -0,0 +1,9 @@
+# Fetch Reports
+
+POST http://localhost:8030/v1/fetchReport/8?organisationUnitCode=TO HTTP/1.1
+content-type: application/json
+
+{
+    "emailAddress": "username",
+    "password": "password"
+}
```
rohan-bes, line 9: For use with vscode RestClient plugin

packages/reports/src/app/addRoutesToApp.ts
```diff
@@ -4,10 +4,16 @@
...
-   **/
-  app.get('(/v[0-9]+)?/fetchReport/:reportCode', fetchReport);
+   * Attach authentication to each endpoint
+   */
+  app.use(authenticationMiddleware);
+
+  /**
+   * POST routes
+   */
+  app.post('(/v[0-9]+)?/fetchReport/:reportCode', fetchReport);
```
rohan-bes, line 18: Switched to `post` for this as we want to pass username/pass in the body. Not a bad idea anyway, as we may end up with a more complex request format anyway

yarn.lock
```diff
@@ -14775,11 +14775,6 @@ history@^4.7.2, history@^4.9.0:
     tiny-warning "^1.0.0"
     value-equal "^1.0.1"
 
-hjson@^3.2.1:
-  version "3.2.1"
-  resolved "https://registry.yarnpkg.com/hjson/-/hjson-3.2.1.tgz#20de41dc87fc9a10d1557d0230b0e02afb1b09ac"
-  integrity sha512-OhhrFMeC7dVuA1xvxuXGTv/yTdhTvbe8hz+3LgVNsfi9+vgz0sF/RrkuX8eegpKaMc9cwYwydImBH6iePoJtdQ==
-
```
rohan-bes, line 14782: This package was actually removed in a separate commit, but I hadn't run `yarn` before committing 🤦

packages/reports/src/routes/fetchReport.ts
```diff
@@ -26,13 +27,24 @@ class FetchReportRouteHandler {
...
     }
+
+    const orgUnit = await req.models.entity.findOne({ code: query.organisationUnitCode });
+    if (!orgUnit) {
+      throw new Error(`No entity found with code ${query.organisationUnitCode}`);
+    }
+    const country = await orgUnit.country();
+    const permissionGroup = await req.models.permissionGroup.findById(report.permission_group_id);
+
+    if (!req.accessPolicy.allows(country.code, permissionGroup.name)) {
```
edmofro, line 45:
```suggestion
    if (!req.accessPolicy.allows(entity.country_code, permissionGroup.name)) {
```
Save a fetch? I'm not reviewing this properly - just got curious before heading home!

rohan-bes: Ah good point!

kael89: Needs to be addressed

packages/reports/src/routes/fetchReport.ts
```diff
@@ -26,13 +27,24 @@ class FetchReportRouteHandler {
...
+      throw new Error(`No access`);
```
kael89, line 46: Perhaps we should make this error more descriptive?

kael89 (changes requested):
Great work overall @rohan-bes !! Just a few details

packages/reports/src/types.ts
```diff
@@ -0,0 +1,27 @@
...
+import { ParamsDictionary, Query } from 'express-serve-static-core';
+import { AccessPolicy } from '@tupaia/access-policy';
+import { Authenticator } from '@tupaia/auth';
+import { TupaiaDatabase, ModelRegistry } from '@tupaia/database';
+
+export interface ReportsRequest<
+  P = ParamsDictionary,
+  ResBody = any,
+  ReqBody = ReportsRequestBody,
+  ReqQuery = Query
```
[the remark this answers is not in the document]
kael89: Obviously we will have to move the definitions for `FetchReportParams`, `ReportsRequestBody` from `fetchReports.ts` to here. It may also be worth re-ordering the definitions according to usage (eg `FetchReportParams`, `ReportsRequestBody` etc goes first, then the dependent `ReportsRequest` goes last)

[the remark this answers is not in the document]
rohan-bes: Is it okay if I address this in my next PR? I've already done half of that refactor there already

packages/reports/src/routes/fetchReport.ts
```diff
@@ -26,13 +27,24 @@ class FetchReportRouteHandler {
     this.aggregator = createAggregator(Aggregator);
   }
 
-  fetchReport = async (req: Request, res: Response): Promise<void> => {
+  fetchReport = async (req: ReportsRequest, res: Response): Promise<void> => {
     const params = req.params as FetchReportParams;
     const query = req.query as FetchReportQuery;
```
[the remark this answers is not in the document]
rohan-bes: I like this, will do it with the `types` changes in the next PR 👍

packages/reports/src/types.ts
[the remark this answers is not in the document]
kael89: No problem at all! 🙂

kael89 (approved):

PR #1602, opened 2020-11-11
Delete old laos schools textbook data

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1547


packages/database/src/migrations/20201110234830-DeleteOldLaosSchoolsTextbookData-modifies-data.js
```diff
@@ -0,0 +1,70 @@
...
+exports.up = async function (db) {
+  await db.runSql(`
+    -- Drop triggers so that all the deletes don't flood the dhis sync queue
+    DROP TRIGGER IF EXISTS survey_response_trigger
+      ON survey_response;
+    DROP TRIGGER IF EXISTS answer_trigger
+      ON answer;
+
+    -- Delete the answers
+    DELETE FROM answer a
```
kael89, line 26: I believe you could delete just the survey responses, and the answers would be deleted as well, because `answer.survey_response_id` has an "on delete cascade" constraint with a reference to `survey_response.id`. But it's definitely fine to keep it as is here 🙂

packages/database/src/migrations/20201110234830-DeleteOldLaosSchoolsTextbookData-modifies-data.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`
+    -- Drop triggers so that all the deletes don't flood the dhis sync queue
```
edmofro, line 19:
```suggestion
    -- Drop triggers so that change notifications don't slow the migration down to the point it fails
    -- This is safe because all data is internal, so no change handlers need to handle this for external sync queues
```

packages/database/src/migrations/20201110234830-DeleteOldLaosSchoolsTextbookData-modifies-data.js
```diff
@@ -0,0 +1,70 @@
...
+    DROP TRIGGER IF EXISTS survey_response_trigger
```
kael89, line 20: I would assume that this is not required since Laos surveys do not get synced to DHIS. Again, it's fine to leave this here (although if that's the case indeed, I would be tempted to remove it)

kael89 (approved):
Nice work, approved with a couple of comments

packages/database/src/migrations/20201110234830-DeleteOldLaosSchoolsTextbookData-modifies-data.js
Alastair-L: I didn't know that! Thanks for pointing it out. Saying that, I'm happy to leave it as it is.

packages/database/src/migrations/20201110234830-DeleteOldLaosSchoolsTextbookData-modifies-data.js
kael89: It seems I reviewed this at the same time as @edmofro , whose comment explains why triggers are dropped

PR #1606, opened 2020-11-12
1538 no bar for no data

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1538
### Changes:

Ended up filtering all non-numeric data out in `ChartWrapper`. Logically, I can't think of a reason why there would be a use for non-numeric data, so this seems logical to me, and it can always be changed if needed.

### DIFF
--- a/packages/web-frontend/src/components/View/ChartWrapper/ChartWrapper.js
+++ b/packages/web-frontend/src/components/View/ChartWrapper/ChartWrapper.js
@@ -38,12 +38,15 @@ export class ChartWrapper extends PureComponent {
   }
 
   removeNonNumericData = data =>
-    data.map(dataSeries =>
-      Object.entries(dataSeries).reduce((newDataSeries, [key, value]) => {
-        const isNonNumericData = isDataKey(key) && Number.isNaN(Number(value));
-        return isNonNumericData ? newDataSeries : { ...newDataSeries, [key]: value };
-      }, {}),
-    );
+    data.map(dataSeries => {
+      const filteredDataSeries = {};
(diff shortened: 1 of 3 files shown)


packages/web-frontend/src/components/View/ChartWrapper/ChartWrapper.js
```diff
@@ -27,15 +27,24 @@ export class ChartWrapper extends PureComponent {
...
-      : { ...viewContent, data: this.sortData(data) };
+      : { ...viewContent, data: massagedData };
   }
 
+  removeNonNumericData = data =>
+    data.map(dataSeries =>
+      Object.entries(dataSeries).reduce((newDataSeries, [key, value]) => {
+        const isNonNumericData = isDataKey(key) && Number.isNaN(Number(value));
+        return isNonNumericData ? newDataSeries : { ...newDataSeries, [key]: value };
+      }, {}),
```
edmofro, line 45: I'm not sure how many entries are involved here, but we might want to use a `forEach` if there's the potential for there to be a lot (see https://beyondessential.slack.com/archives/CDGS6P56K/p1585867391005900)

edmofro (approved):
lgtm

packages/web-frontend/src/components/View/ChartWrapper/ChartWrapper.js
Alastair-L: >  2000-3000 times faster!!

Wow! 

There are only as many entries here as legend entries (I have seen maybe 20 max), but will make the improvement anyway.

edmofro (approved):

PR #1608, opened 2020-11-12
Pass event to submitForm

edmofro:
To fix the attached screenshot, which pops up if you're running in dev mode

image

### DIFF
--- a/packages/meditrak-app/app/country/RequestCountryAccessPage.js
+++ b/packages/meditrak-app/app/country/RequestCountryAccessPage.js
@@ -194,7 +194,7 @@ export class RequestCountryAccessPage extends React.Component {
                 <Button
                   title="Request access"
                   isDisabled={isComplete}
-                  onPress={() => this.submitForm()}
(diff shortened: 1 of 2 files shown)


nick-beyondessential (approved):
LGTM

packages/meditrak-app/app/country/RequestCountryAccessPage.js
```diff
@@ -194,7 +194,7 @@ export class RequestCountryAccessPage extends React.Component {
                 <Button
                   title="Request access"
                   isDisabled={isComplete}
-                  onPress={() => this.submitForm()}
+                  onPress={e => this.submitForm(e)}
```
kael89, line 197: I don't think this is required here (the `submitForm` method does not accept any params. No functionality harm passing it in either

edmofro: The double review digs deep...thanks Kostas, have updated directly in dev

PR #1610, opened 2020-11-12
Use http/2 protocol on all requests

edmofro:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/1542

We have been using http/1.1 by default. According to https://developers.google.com/web/tools/chrome-devtools/network/reference?utm_source=devtools#timing-explanation 

> There are already six TCP connections open for this origin, which is the limit. Applies to HTTP/1.0 and HTTP/1.1 only.

This moves us to http/2

Will need to be manually applied - there is no CI/CD set up for nginx settings.

### DIFF
--- a/packages/devops/configs/servers.conf
+++ b/packages/devops/configs/servers.conf
@@ -5,8 +5,8 @@ server {
 }
 server {
 
-  listen       443 ssl;
-  listen       [::]:443 ssl;
+  listen       443 ssl http2;
+  listen       [::]:443 ssl http2;
   server_name  ~^.*admin\.tupaia\.org$;
   root         /home/ubuntu/tupaia/packages/admin-panel/served_build/;
 
@@ -41,8 +41,8 @@ server {
 }
 server {
 
-  listen       443 ssl;
-  listen       [::]:443 ssl;
+  listen       443 ssl http2;
+  listen       [::]:443 ssl http2;
   server_name  ~^.*psss\.tupaia\.org$;
   root         /home/ubuntu/tupaia/packages/psss/served_build/;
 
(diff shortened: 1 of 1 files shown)


kael89 (approved):
Great find!

edmofro:
Note that with this applied, some requests do still time out, but it stops the throttling/red area in [Nick's screenshot](https://user-images.githubusercontent.com/62525113/98604440-c7f7a080-2337-11eb-884f-beac2c04aaef.png) 

Screen Shot 2020-11-12 at 5 21 30 PM

edmofro:
Merged as I've already tested manually on a feature instance and dev, and will now apply manually on production.

kael89:
I thought this would be unrelated to the problems I had in my local environment (https://beyondessential.slack.com/archives/CDGS6P56K/p1605150189099400), but there are some clues that it may indeed be related:

The requests that fail are not slow (I have checked this using time measurement), they are just stalled. We can verify this in the Chrome Network tab:

image

Now some interesting questions arise
1. How come this problem occurs to me now? I tried searching if this was introduced in Chrome recently, but didn't find anything relevant.
2. This seems to happen only for Tongan dashboards - Laos Schools at the country level, a resource intensive page, loads fine

edmofro:
Didn't you say this was local though? As in, web-config-server running locally? In that case it wouldn't be running through nginx at all, so there is no change from this PR

Secondly, you reported this issue around 45 minutes before I made the change manually on a feature branch/dev.

kael89:
Yes, the solution here shouldn't be related to my problem, I just wanted to point out some similarities in case the root cause is similar

PR #1612, opened 2020-11-13
1282: Add reporting package to CI/CD

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1282

Also sorted some items alphabetically and removed an unused dependency


full-testable-package.Dockerfile
```diff
@@ -13,6 +13,7 @@ WORKDIR /tupaia
 COPY package.json ./
 COPY yarn.lock ./
 COPY babel.config.json ./
+COPY .babelrc-ts.js ./
```
kael89, line 16: This is what was causing previous builds for this epic to fail

edmofro (approved):

PR #1614, opened 2020-11-13
1421 refactor enlarged dialog

Alastair-L:
### Issue #:
Refactoring in preparation for https://github.com/beyondessential/tupaia-backlog/issues/1421

Removes the idea of a separate drillDown component and handles all drillDown logic locally in `EnlargedDialog`.


packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
```diff
@@ -52,39 +46,89 @@ const STATUS = {
...
+      },
+      projectCode,
+      startDate,
+      endDate,
+      isExpanded: true,
+      ...drillDownState,
+    };
+    fetchViewData(options);
+    // TODO: Only refetching on parameterValue change is a bit hacky... We rely on not refetching here...
+  }, [startDate, endDate, drillDownState.parameterValue, infoViewKey]);
```
Alastair-L, line 100: This is the section I seem to be having issues with.

I want it to reload on open if and only if:
 - The expanded date range is different to the date range for the small one.
 - The report is a matrix (in small form, it just showed a placeholder)

(https://github.com/beyondessential/tupaia/pull/1623 may help with not reloading after opening)

After opening, it should reload only on
 - Date range change
 - Drill down

Note:
 - It should NOT reload on drill up (this relates to the TODO)

Also, should we consider the cases that we don't support, such as the infoViewKey changing while the report is open?

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
```diff
@@ -133,6 +173,11 @@ const EnlargedDialogComponent = ({
         scroll={isMobile() ? 'body' : 'paper'}
         PaperProps={{ style: getDialogStyle() }}
       >
+        {isLoading && (
```
Alastair-L, line 176: I believe that this sort of thing would look better than what we used to have, which was just a grey filter over the whole screen.

Alastair-L: Even better would be just a loading symbol in place of the report itself, and still show the date range picker, close/back icons and title (as we should have that information).

packages/web-frontend/src/sagas.js
```diff
@@ -1006,29 +1001,62 @@ function getTimeZone() {
...
-      drillDownLevel,
-      ...restOfAction,
-    },
-    fetchDrillDownError,
-  );
-  if (drillDownData) {
-    yield put(
-      fetchDrillDownSuccess(drillDownLevel, { ...drillDownData, parameterLink, parameterValue }),
-    );
+function* fetchEnlargedDialogData(action) {
```
Alastair-L, line 1006: Do you agree with consolidating the three sagas?

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
```diff
@@ -52,39 +46,89 @@ const STATUS = {
...
+
+  const viewContent = contentByLevel?.[drillDownState.drillDownLevel]?.viewContent;
+
+  const { organisationUnitCode, dashboardGroupId, viewId } = getInfoFromInfoViewKey(infoViewKey);
 
-  if (!isVisible) {
-    return null;
-  }
+  React.useEffect(() => {
+    if (!viewId) return;
```
Alastair-L, line 80: should always be there

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
Alastair-L: Discussed with Tom Caiger.

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
Alastair-L: ?

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
```diff
@@ -163,6 +163,7 @@ export class MatrixWrapper extends Component {
     return (
       !shallowEqual(this.props, nextProps) ||
       nextState.isLoading !== this.state.isLoading ||
+      nextState.expandedMatrixData !== this.state.expandedMatrixData ||
```
Alastair-L, line 166: Not quite sure about why this is needed now?

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialogContent.js
```diff
@@ -161,31 +161,56 @@ export class EnlargedDialogContent extends PureComponent {
...
+    // isVisible
 
     if (isExporting) {
       return null;
     }
 
     return (
-      <div style={{ ...styles.toolbar, ...toolbarStyle }}>
-        {this.isExportable() ? (
+      <>
```
Alastair-L, line 183: Functional change: show an x for drillDown reports.

packages/web-frontend/src/sagas.js
```diff
@@ -704,22 +699,22 @@ function* fetchViewData(parameters, errorHandler) {
...
       ? parameters
-      : getDefaultDates(state.global.viewConfigs[infoViewKey]);
+      : getDefaultDates(state.global.viewConfigs[infoViewKey] || {});
+
   // Build the request url
   const {
     organisationUnitCode,
     dashboardGroupId,
     viewId,
-    drillDownLevel,
```
Alastair-L, line 713: Not a functional change, drillDownLevel is in extraUrlParameters now (but no strong preference on this)

packages/web-frontend/src/selectors/dashboardSelectors.js
```diff
@@ -58,12 +59,43 @@ export const selectCurrentInfoViewKey = createSelector(
...
+export const selectShouldUseDashboardData = createSelector(
+  [selectCurrentInfoViewKey, selectCurrentExpandedViewContent, (_, options) => options],
+  (candidateInfoViewKey, candidateViewContent, options) => {
+    const { startDate, endDate, infoViewKey, drillDownLevel } = options;
+
+    if (drillDownLevel > 0) return false;
+    if (candidateInfoViewKey !== infoViewKey) return false;
+    if (!candidateViewContent || candidateViewContent.type === 'matrix') return false;
+
+    const { startDate: defaultStartDate, endDate: defaultEndDate } = getDefaultDates(
```
Alastair-L, line 87: If they open the dashboard group on one day, leave it overnight and open a report the next day, it will refetch the data (perhaps) unnecessarily. But I don't think that's a huge issue

packages/web-frontend/src/actions.js
```diff
@@ -1036,85 +1031,34 @@ export function openEnlargedDialog(viewId) {
   };
 }
```
tcaiger, line 1033: Nice work - it's good to see some of this redux boiler plate go

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
tcaiger: hmmm me neither. I guess react is not picking up that the props have changed in EnlargedDialog when the useEffect kicks in. Perhaps because viewContent is the object that gets passed in to EnlargedDialogContent and it is a nested object (some shallow comparison issue). Anyway if this works then I think it is ok.

packages/web-frontend/src/selectors/dashboardSelectors.js
```diff
@@ -58,12 +59,43 @@ export const selectCurrentInfoViewKey = createSelector(
...
-    selectCurrentInfoViewKey,
-    state => state.dashboard.viewResponses,
+    selectCurrentDashboardGroupIdForExpandedReport,
+    selectCurrentOrgUnitCode,
+    selectCurrentExpandedViewId,
   ],
-  (viewContent, infoViewKey, viewResponses) => {
-    return viewContent || viewResponses[infoViewKey];
+  (dashboardGroupId, organisationUnitCode, viewId) =>
+    !!(dashboardGroupId && organisationUnitCode && viewId),
```
tcaiger, line 75: Just a general note that I am not really a big fan of using derived data like this to show and hide a modal. I think it is better to explicitly have a piece of state for opening and closing the modal. It is a better separation of concerns (api data vs. app state) and also gives you more control (adding a loader, un-mounting etc). However I think that refactoring this is out of scope of this PR

packages/web-frontend/src/selectors/dashboardSelectors.js
tcaiger: I think that is actually a good thing.

packages/web-frontend/src/sagas.js
tcaiger: Yes I think it is definitely an improvement

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
tcaiger: We will need to change the loader somehow so that it only shows when the dropdown level changes and not when the date changes. There is already a loader for the date changing next to the date picker and then the chart themselves don't have a loader so that the chart animation can happen. We are keen to keep the chart animation.

tcaiger: Also, in the existing drill down the container overflow is set to scroll and now it changes sizes. We need to fix this so that the dialog doesn't keep changing size. It might be a matter of just adding a max-height somewhere?

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialogWrapper.js
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import { connect } from 'react-redux';
+import { closeEnlargedDialog } from '../../actions';
+import { selectIsEnlargedDialogVisible } from '../../selectors';
+import { EnlargedDialog } from './EnlargedDialog';
+
+const EnlargedDialogWrapperComponent = ({ isVisible, onCloseOverlay }) => {
```
tcaiger, line 13: hmmm I'm not sure about this component - it seems a bit unnecessary. What do you think about moving the onCloseOverlay down to EnlargedDialog and moving the isVisible up to the Screen components?

tcaiger: The other option would be to reset all the state to default values in the EnlargedDialog component but that could get messy. One potential solution for that is to use the `useReducer` hook. I'm not sure if you have seen that but it is good when there are a lot of interacting uses of `useState` in a component. I think it might be going too far to use that here but it is good to know about.

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
```diff
@@ -3,32 +3,25 @@
  * Copyright (c) 2019 Beyond Essential Systems Pty Ltd
  */
 
-import React from 'react';
+import CircularProgress from '@material-ui/core/CircularProgress';
+import Dialog from '@material-ui/core/Dialog';
 import PropTypes from 'prop-types';
+import React from 'react';
```
tcaiger, line 9:
```suggestion
import React, { useState, useEffect, useRef } from 'react';
```

tcaiger: Then you can refactor all the `React.useState` etc below to be `useState`etc

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
```diff
@@ -52,39 +45,116 @@ const STATUS = {
...
+    organisationUnitName,
+    onSetDateRange,
+    startDate: startDateForTopLevel,
+    endDate: endDateForTopLevel,
+    isLoading,
+    projectCode,
+    infoViewKey,
+    fetchViewData,
+    drillDownDatesByLevel,
+  } = props;
```
tcaiger, line 77: Any reason for not simply de-structuring at the top?

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialogWrapper.js
```diff
@@ -0,0 +1,42 @@
+/**
```
tcaiger, line 1: it's minor but this is the official header now.

`/*
 * Tupaia
 */`

tcaiger (changes requested):
Nice work - a big refactor and it's looking way better. I left a few comments and questions to address. Let me know if you need any clarification.

ps. another minor one that didn't fit anywhere
Don't need to export `EnlargedDialogContent`from EnlargedDialog/index.js

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
Alastair-L: Oops. It was for console.logging, I'll put it back

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
Alastair-L: Awesome, thanks.

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
Alastair-L: > We will need to change the loader somehow so that it only shows when the dropdown level changes and not when the date changes.
👍 

> We need to fix this so that the dialog doesn't keep changing size.

In investigating this, it looks like the only reason it used to be the same size is that the underlying chart was still being rendered and setting the size, and the `drillDownOverlay` just conformed to that. 

My understanding has been completely changed. Instead of navigating to a new, related report, we were showing some deeper information over the top of the chart. 

Perhaps the best way to go about this is to revert specifically the drillDownOverlay changes? Now that I understand the old system better, it may not have needed a refactor at all? We can still delete all the redux stuff? What do you think?

packages/web-frontend/src/selectors/dashboardSelectors.js
Alastair-L: Interesting, thank you for that. Will definitely consider this in future.

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialogWrapper.js
Alastair-L: > What do you think about moving the onCloseOverlay down to EnlargedDialog and moving the isVisible up to the Screen components?

I've gone with this approach.

> One potential solution for that is to use the useReducer hook

I hadn't seen it before but it looks interesting! Will definitely keep it in mind for the future.

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
tcaiger: hmmm if you are happy to do that then that sounds like a good idea to me. and yep if we can still remove the redux stuff that would be good.

Alastair-L: This little bit of gap at the sides and transparency matches my new understanding... 
It definitely needs a back button, but after that do you think it's more clear?
image

Alastair-L: I'm going to commit here but make it match the existing functionality, just thought this might be worth a look.

packages/web-frontend/src/screens/desktop/RootScreen/index.js
```diff
@@ -36,13 +39,21 @@ export const RootScreen = () => {
         </div>
         <OverlayDiv />
         <SessionExpiredDialog />
-        <EnlargedDialog />
+        {enlargedDialogIsVisible ? <EnlargedDialog /> : null}
```
tcaiger, line 42: Nice this looks better to me. More clearly shows what is happening IMO

tcaiger (approved):
Beautiful!

PR #1617, opened 2020-11-15
Don't allow overwrite of s3 images

edmofro:
### Changes:

- Check whether file exists on S3 before uploading, and throw an error if it does
- Get rid of unused code within `deleteImage.js`


packages/meditrak-server/src/s3/S3Client.js
```diff
@@ -0,0 +1,52 @@
...
+      this.s3
+        .headObject({
+          Bucket: BUCKET_NAME,
+          Key: fileName,
+        })
+        .on('success', () => {
+          resolve(true);
+        })
+        .on('error', () => {
+          resolve(false);
```
kael89, line 25: We could check that the error thrown is specifically because of the file not found, and not for another reason

kael89 (approved):
A PR review starts great when it starts with a class! Looking good, just one comment but a detail probably. Approving!

packages/meditrak-server/src/s3/S3Client.js
edmofro: True, that would be a more cautious approach!

edmofro: I was just about to go and do that (because I think it's a good idea) and then I remembered how we have to cut a few corners if we hope to deliver psss by 11th Dec, so I'm going to ignore your suggestion - sorry!

kael89: No worries - I think it's looking great as is!

PR #1618, opened 2020-11-16
Change unfpa default measure

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1094
### Changes:
Just needed to change the default map overlay. It still works at regional level as the new default is not available there.


chris-bes (approved):
LGTM!

PR #1622, opened 2020-11-16
1477: Finalise Fetch for Reporting Package

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1477:

New config:
```
"fetch": {
  "dataElements": [ "BCD46", "BCD47", "BCD48"]
}

or

"fetch": {
  "dataGroups": [ "HP02"]
}
```

Re-wrote the fetch step to function like the transform step internally. A few refactors in here as well.


rohan-bes:
Don't worry about the build failing, it's only because I accidentally gave the branch an invalid name 🤦

packages/reports/src/reportBuilder/fetch/functions/dataGroups.ts
```diff
@@ -0,0 +1,57 @@
...
+      startDate,
+      endDate,
+    },
+    {},
+  )) as Row[];
+  response.forEach(row => {
+    Object.entries(row.dataValues).forEach(([key, value]) => {
+      row[key] = value;
+    });
+    delete row.dataValues;
```
kael89, line 32: We should somehow make sure that `key`s do not conflict with other event metadata keys (`eventDate` etc) This will most probably not be a problem in practice, since we use different naming conventions for those (upper SNAKE_CASE for data element keys, camelCase for metadata)

packages/reports/src/reportBuilder/fetch/functions/dataGroups.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+  return {
+    results: response,
+  };
+};
+
+const buildParams = (params: unknown): DataGroupsFetchParams => {
+  if (!Array.isArray(params)) {
+    throw new Error(`Expected params object but got ${params}`);
```
kael89, line 41:
```suggestion
    throw new Error(`Expected an array but got ${params}`);
```

packages/reports/src/reportBuilder/fetch/types.ts
```diff
@@ -0,0 +1,5 @@
+import { Row } from '../types';
+
+export interface FetchResponse {
+  results: Row[];
+}
```
kael89, line 5: Separating types in different files is a new pattern which makes a lot of sense! 👌

kael89 (changes requested):
This is some really nice piece of work @rohan-bes !! I basically liked everything I saw at an architectural level, and the implementation is also great 🚀 

A few details but looks great overall!

packages/reports/src/reportBuilder/fetch/fetch.ts
```diff
@@ -1,26 +1,43 @@
...
-  const fetchFunctionKey: string = Object.keys(firstFetch)[0];
-  if (isValidFetchFunction(fetchFunctionKey)) {
-    const fetchFunction = fetchFunctions[fetchFunctionKey];
-    return fetchFunction(aggregator, query, firstFetch[fetchFunctionKey]);
+  return fetcher.call(aggregator, query);
+};
+
+const buildParams = (params: unknown): FetchParams => {
+  if (typeof params !== 'object' || params === null) {
+    throw new Error(`Expected object but got ${params}`);
```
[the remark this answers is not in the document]
rohan-bes: @kael89 yeah I totally agree. Especially since eventually we will need to do validation as part of a separate process to calculating the report (ie. when the user submits a report config).

I still like this model of doing recursive validation as it think it makes each single validation level easier to read, but yeah we should probably compartmentalise it all within a `validate` folder or something

packages/reports/src/reportBuilder/fetch/functions/dataElements.ts
```diff
@@ -0,0 +1,43 @@
...
+    {
+      aggregationType: 'RAW',
+    },
+  )) as FetchResponse;
+  return response;
+};
+
+const buildParams = (params: unknown): DataElementsFetchParams => {
+  if (!Array.isArray(params)) {
+    throw new Error(`Expected params object but got ${params}`);
```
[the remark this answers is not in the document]
rohan-bes: Good catch!

packages/reports/src/reportBuilder/fetch/functions/dataGroups.ts
```diff
@@ -0,0 +1,57 @@
...
+    programCode,
+    {
+      useDeprecatedApi: false,
+      dataServices: { isDataRegional: true },
+      organisationUnitCodes: [organisationUnitCode],
+      startDate,
+      endDate,
+    },
+    {},
+  )) as Row[];
```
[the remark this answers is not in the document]
rohan-bes: That's a good idea. I think after we deliver PSSS it will be a good time for us to find the right place to define common types between Tupaia packages (Analytic, Event, AnalyticCluster, etc.)

packages/reports/src/reportBuilder/fetch/functions/dataGroups.ts
rohan-bes: Oh good point... Tbh I'm not really happy with how I'm handling events here... Perhaps we just leave them as they come from the Aggregator, and then have a named transform format the data correctly, eg.
```
{
  "fetch": { "dataGroups": ["HP02"] },
  "transform": [
    "expandDataGroups",
    ...
  ]
}
```

Since we don't need to be using events in the short term, I'm happy to make that decision at a later date. But yeah for the record I still don't know the best way to handle events...

packages/reports/src/reportBuilder/fetch/types.ts
rohan-bes: Yeah, I agree. I did it to avoid dependency cycles, but honestly it just makes things a lot easier to read.

I am a fan tho of having multiple different `types.ts` files within various packages. I'm not entirely sure why, but my code feels more appropriately encapsulated that way.

packages/reports/src/reportBuilder/fetch/functions/dataGroup.ts
```diff
@@ -1,56 +1,59 @@
...
-  const { programCode } = params;
+  const { dataGroupCode } = params;
   const { organisationUnitCode, startDate, endDate } = query;
   const response = (await aggregator.fetchEvents(
-    programCode,
+    dataGroupCode,
     {
       useDeprecatedApi: false,
-      dataServices: { isDataRegional: true },
+      dataServices: [{ isDataRegional: true }],
```
kael89, line 20: Oops, good catch - I missed that!

packages/reports/src/reportBuilder/fetch/functions/dataGroup.ts
```diff
@@ -1,56 +1,59 @@
 import { Aggregator } from '../../../aggregator';
-import { FetchReportQuery } from '../../../types';
+import { FetchReportQuery, Event } from '../../../types';
 import { FetchResponse } from '../types';
-import { Row } from '../../types';
 
 type DataGroupsFetchParams = {
-  programCode: string;
+  dataGroupCode: string;
```
kael89, line 6: I love this renaming!

packages/reports/src/reportBuilder/fetch/functions/dataGroup.ts
```diff
@@ -1,56 +1,59 @@
...
-  )) as Row[];
-  response.forEach(row => {
-    Object.entries(row.dataValues).forEach(([key, value]) => {
-      row[key] = value;
-    });
-    delete row.dataValues;
+  )) as Event[];
+  const rows = response.map(event => {
+    const { dataValues, ...restOfEvent } = event;
+    return { ...dataValues, ...restOfEvent };
```
kael89, line 29: This is better!

packages/reports/src/reportBuilder/fetch/functions/dataElements.ts
```diff
@@ -0,0 +1,43 @@
+import { Aggregator } from '../../../aggregator';
+import { FetchReportQuery } from '../../../types';
+import { FetchResponse } from '../types';
+
+type DataElementsFetchParams = {
```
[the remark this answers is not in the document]
kael89: Not resolved

packages/reports/src/types.ts
```diff
@@ -9,19 +9,38 @@ import { AccessPolicy } from '@tupaia/access-policy';
...
-interface ReportsRequestBody {
-  emailAddress: string;
-  password: string;
+export interface Event {
+  event: string;
+  eventDate: string;
+  orgUnitName: string;
+  orgUnit: string;
+  dataValues?: {
+    [key: string]: string | number | boolean;
```
kael89, line 44: This is my only suggestion... I believe we don't cast event data values to `boolean`, but I'm not 100% sure

```suggestion
    [key: string]: string | number;
```

kael89 (approved):
Nicely done! Just one unresolved suggestion (in `dataElements.ts`) and a new suggestion left, but both are details so I'm going to approve 🙌

packages/reports/src/types.ts
rohan-bes: Ahh true... boolean values are regarded as numbers huh? I'm happy to change this 👍

kael89 (approved):

PR #1623, opened 2020-11-16
1578: Do not reload date range report

kael89:
### Issue #:
Fix https://github.com/beyondessential/tupaia-backlog/issues/1578

### DIFF
--- a/packages/web-config-server/src/apiV1/view.js
+++ b/packages/web-config-server/src/apiV1/view.js
@@ -1,9 +1,12 @@
-import { convertDateRangeToPeriods, replaceValues, CustomError } from '@tupaia/utils';
+import { convertDateRangeToPeriods, CustomError, replaceValues, utcMoment } from '@tupaia/utils';
 import { getDhisApiInstance } from '/dhis';
 import { isSingleValue } from './utils';
 import { DataAggregatingRouteHandler } from './DataAggregatingRouteHandler';
 import { DashboardPermissionsChecker } from './permissions';
 import { getDataBuilder } from '/apiV1/dataBuilders/getDataBuilder';
+
+// Same as in @tupaia/web-frontend
+const DEFAULT_MIN_DATE = '20150101';
 
 const viewFail = {
   type: 'View Error',
@@ -14,13 +17,6 @@ const noViewWithId = {
   responseText: {
     status: 'viewError',
     details: 'No view with corresponding id',
-  },
-};
-
-const viewNotInGroup = {
-  responseText: {
-    status: 'viewError',
-    details: 'Dashboard group does not contain view',
   },
 };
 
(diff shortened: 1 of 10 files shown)


packages/web-config-server/src/apiV1/view.js
```diff
@@ -1,10 +1,13 @@
-import { convertDateRangeToPeriods, replaceValues, CustomError } from '@tupaia/utils';
+import { convertDateRangeToPeriods, CustomError, replaceValues, utcMoment } from '@tupaia/utils';
 import { getDhisApiInstance } from '/dhis';
 import { isSingleValue } from './utils';
 import { DataAggregatingRouteHandler } from './DataAggregatingRouteHandler';
 import { DashboardPermissionsChecker } from './permissions';
 import { getDataBuilder } from '/apiV1/dataBuilders/getDataBuilder';
 
+// Same as in @tupaia/web-frontend
+const DEFAULT_MIN_DATE = '20150101';
```
kael89, line 9: This is also defined in `web-fronted/src/components/DateRangePicker/constants.js`.

Ideally we would define this constant in a single place (eg the `utils` package), but we've had a bit of trouble loading functionality from other tupaia packages in `web-frontend`. Hopefully the above comment helps!

packages/web-config-server/src/apiV1/view.js
```diff
@@ -113,7 +109,12 @@ export default class extends DataAggregatingRouteHandler {
 
   // common view translation (for all possible views)
   addViewMetaData = inJson => {
-    const { viewJson, query, startDate, endDate } = this;
+    const {
+      viewJson,
+      query,
+      startDate = DEFAULT_MIN_DATE,
+      endDate = utcMoment().format('YYYYMMDD'),
```
kael89, line 116: The front-end uses these values as defaults for start and end dates, so we may as well as specify them earlier on in the data flow. Ideally we would infer them from the results, but it's hard to do so for all data builders given the existing architecture. The new `reports` package will probably make this easier!

chris-bes (approved):
LGTM!

kael89:
@chris-bes after consulting @edmofro I realised that there is no need to have the default date range logic in `web-config-server`, but we should move it in the front end saga level. Could you please re-review ?

packages/web-frontend/src/actions.js
```diff
@@ -697,17 +697,13 @@ export function fetchDashboardItemData(
   dashboardGroupId,
   viewId,
   infoViewKey,
-  startDate,
-  endDate,
```
kael89, line 701: Those were never used in practice

chris-bes (approved):

kael89:
@Alastair-L this is "ready for test" now, just bringing it in your attention in case it conflicts with https://github.com/beyondessential/tupaia-backlog/issues/1421

Alastair-L:
Thanks @kael89 for bringing it to my attention!

PR #1624, opened 2020-11-16
1376 add min/max lines to MOS report

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1376

Basically this migration is only adding two new lines to the `chartConfig` of the `viewJson` for this report but I couldn't figure out a way to add these lines without including the entire `chartConfig` in this migration
- The two new lines are: `"referenceValue": 12` & `"referenceValue": 6`
- The only way I could figure out to include this in the `chartConfig` was to include it with a specific data element - this is probably not ideal, since these `referenceValue`'s apply to all the data elements & not just the ones I associated them with in the `viewJson` but I couldn't figure out how to do this a different way & it works like it should in the front-end (but it's probably confusing if someone was reading the `viewJson` & trying to figure out how it worked)


packages/database/src/migrations/20201116013159-AddMinMaxValuestoMOSReport-modifies-data.js
```diff
@@ -0,0 +1,151 @@
...
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function(db) {
+  return db.runSql(`
+      UPDATE "dashboardReport"
+      SET "viewJson" = "viewJson" || '{"chartConfig": {
```
Alastair-L, line 20: My personal preference would be to have the chartConfig as a javascript object, and use JSON.stringify() to insert it into the database. This also means that you can make the changes more clear.
```javascript
const OLD_CHART_CONFIG = { ... };

const toInsert = {
   ...OLD_CHART_CONFIG,
   UNFPA_RH_MOS_Condoms_male: {
       ...OLD_CHART_CONFIG.UNFPA_RH_MOS_Condoms_male,
       referenceValue: 12,
    },
   UNFPA_RH_MOS_Condoms_female: {
       ...OLD_CHART_CONFIG.UNFPA_RH_MOS_Condoms_female,
       referenceValue: 6,
    },
}

```

Alastair-L (changes requested):
An alternate suggestion, is to add the following to the chartConfig:
```javascript
    "reference_1": {
      "referenceValue": 12
    },
    "reference_2": {
      "referenceValue": 6
    },
```


Then somehow hide them on the legend.
What do you think?

Saying that, I'm happy with your solution, just wanted to point this one out in case you'd prefer it as you mentioned in the description.

avaek:
Thanks for the great suggestions @Alastair-L - I think the second one is the coolest but as far as I can tell theres currently no way to hide these items from the legend of the report so to do it this way would seem to require new functionality. Since I didn't want to spend too long on this card I went with your first suggestion - I tested it locally and it worked great but let me know if theres any code formatting changes that are required before I move this to `QA on Feature`

packages/database/src/migrations/20201116013159-AddMinMaxValuestoMOSReport-modifies-data.js
```diff
@@ -0,0 +1,107 @@
...
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const OLD_CHART_CONFIG = {
+  "UNFPA_RH_MOS_Condoms_male": {
```
Alastair-L, line 18: It doesn't really matter, but one of the advantages of putting this into an object is that you can remove the quotes around keys, which your linter should be able to do automatically.

Alastair-L (approved):
One small comment, but other than that looks good! Approved :)

avaek:
Thanks for your help with this @Alastair-L !
- I've changed the code as per Erin's request here ( https://github.com/beyondessential/tupaia-backlog/issues/1376#issuecomment-731858682 ) so now the reference lines display a label as well as implementing some of your earlier comments on this PR

packages/web-frontend/src/components/View/ChartWrapper/ReferenceLabel.js
```diff
@@ -12,11 +12,13 @@ function ReferenceLabel(props) {
   const x = viewBox.width / 2 + 30;
   const y = viewBox.y + 15;
 
+  if(typeof value != 'undefined' && value !=null) {
```
Alastair-L, line 15: Perhaps it would be worth spending some time getting your linter working? Couple spacing issues here, but the big one is that `!= null` will also catch `undefined`, so there's no need for the first check.
image

Alastair-L:
```suggestion
  if(value == null) return null;
```
Also, I believe that this is more clear, we
a) Avoid a level of nesting 
b) Make sure to return something no matter what path the code takes
c) Match a pattern which is used all over the frontend. `if(x) return null;`

packages/database/src/migrations/20201116013159-AddMinMaxValuestoMOSReport-modifies-data.js
```diff
@@ -0,0 +1,109 @@
...
+    label: 'Implant',
+    legendOrder: 4,
+  },
+};
+
+const NEW_CHART_CONFIG = {
+  ...OLD_CHART_CONFIG,
+  max_reference_line: {
+      referenceValue: 12,
+      color: '#ffffff',
```
Alastair-L, line 79: I don't think we use `color`?

Alastair-L (approved):
Looks great! Just a few suggestions on style, but I'm approving nonetheless.

Alastair-L:
Note, I just checked out her comment and it looks like the max and min were coloured? Not sure if that's a requirement, perhaps you should just double check that it isn't. (Also, her label reads `6 (Min)`)

avaek:
Thanks for the review @Alastair-L - I've implemented your suggestions if you want to give it one final quick review to make sure they're implemented correctly (I also made a couple of other changes my linter suggested, sorry for not running it earlier - it stopped working during development & I was too hasty to fix it before submitting this PR)

> Note, I just checked out her comment and it looks like the max and min were coloured? Not sure if that's a requirement, perhaps you should just double check that it isn't. (Also, her label reads 6 (Min))
- Thanks, I set it up the way I did cos I thought it looked prettier but I'll talk to Erin to confirm that it's okay

Alastair-L:
Approved! Thanks for taking the time to satisfy my nitpicking!

> Thanks, I set it up the way I did cos I thought it looked prettier but I'll talk to Erin to confirm that it's okay

Thanks,  I'm not against you improving it, just wanted to catch it in case it was accidental.

avaek:
Hey @Alastair-L - based on some feedback during testing I've made some more changes exclusively to the `ReferenceLabel.js` file to implement the discussion from this comment ( https://github.com/beyondessential/tupaia-backlog/issues/1376#issuecomment-737665028 ) - do you think you could re-review this PR for the (hopefully) final time?

packages/web-frontend/src/components/View/ChartWrapper/ReferenceLabel.js
```diff
@@ -7,14 +7,21 @@
 
 import React from 'react';
 
+function returnTrueIfMax(value) {
+  if (value.includes('MAX') || value.includes('Max') || value.includes('max')) return true;
```
Alastair-L, line 11: Possibly you could use `value.toLowerCase().includes('max')`?
image

Alastair-L: Also if you want to be super fancy this whole function could be one line:
```
const isMaxLabel = value => value.toLowerCase().includes('max');
```

packages/web-frontend/src/components/View/ChartWrapper/ReferenceLabel.js
```diff
@@ -7,14 +7,21 @@
...
+  return false;
+}
+
 function ReferenceLabel(props) {
   const { fill, fontSize, value, viewBox } = props;
   const x = viewBox.width / 2 + 30;
-  const y = viewBox.y + 15;
+  const defaultY = viewBox.y + 15;
+  const maxY = viewBox.y - 5;
```
Alastair-L, line 19: maxY tells me this is the highest that the Y could be, not that it is the y value for a y label. Maybe we could just do away with the extra variable? :

`const y = isMaxLabel(value) ? viewBox.y - 5 : viewBox.y + 15;`

packages/web-frontend/src/components/View/ChartWrapper/ReferenceLabel.js
```diff
@@ -7,14 +7,21 @@
...
+  const getYPosition = returnTrueIfMax(value) ? maxY : defaultY;
 
   if (value == null) return null;
   return (
-    <text x={x} y={y} fill={fill} fontSize={fontSize || 14} fontWeight="bolder">
+    <text x={x} y={getYPosition} fill={fill} fontSize={fontSize || 14} fontWeight="bolder">
```
Alastair-L, line 24: Side note, `getX`, to me, implies that this is a function, whereas it is just a value in this case.

Alastair-L (changes requested):
Few little cleanups which might be nice before I approve it :). But looks good 👍

avaek:
Thanks for the great review @Alastair-L - those suggestions make the code much cleaner. I've applied those changes to the PR, let me know how it looks

Alastair-L (approved):
Great work! Sorry to not review this before this week's release though :(

PR #1626, opened 2020-11-16
1391 hide services provided overlay within unfpa

PShoe:
### Changes: Hiding a map overlay group by deleting it from map_overlay_group_relation 

- Please be harsh on my first PRs, it's how I'll learn : ) 

---


packages/database/src/migrations/20201116024506-RemoveMapOverlayUNFPA-modifies-data.js
```diff
@@ -0,0 +1,40 @@
...
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function(db) {
+  return db.runSql(`
+    DELETE FROM map_overlay_group_relation where child_id IN (SELECT id FROM map_overlay_group where name = 'Services provided');
```
kael89, line 22: Hey @PShoe ! This will remove `Services Provided` from all projects. What we want to do instead is remove it from the `unfpa` project specifically.

Map overlays are attached to projects via the `mapOverlay.projectCodes` column. Thus, we should remove all overlays belonging to the `Services Provided` group from the `unfpa` project.  After this is done, the group will be empty for UNFPA and it shouldn't show up in the front end.

Please let me know if you need more help! 🙌

kael89 (changes requested):

packages/database/src/migrations/20201116024506-RemoveMapOverlayUNFPA-modifies-data.js
PShoe: @kael89 was this more of what you had in mind? 
Thank you for the extra-effort review and sql help 🙏

packages/database/src/migrations/20201116024506-RemoveMapOverlayUNFPA-modifies-data.js
```diff
@@ -0,0 +1,49 @@
...
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  const servicesProvidedId = await db.runSql(`
+    SELECT id FROM map_overlay_group WHERE code = 'Services_provided'
+  `);
```
kael89, line 22: Did the above code work for you locally?

```suggestion
  const { rows } = await db.runSql(`
    SELECT id FROM map_overlay_group WHERE code = 'Services_provided'
  `);
  const [{ id: servicesProvidedId }] = rows;
```

Similarly for the `down` migration

kael89 (changes requested):

packages/database/src/migrations/20201116024506-RemoveMapOverlayUNFPA-modifies-data.js
PShoe: @kael89 
It worked when I had the id hardcoded in postico (hah), and didn't throw an error when I ran the migration up/down locally in my terminal. Which probably misled me. I may need help with how to properly run multiple lines of code in positco's SQL query editor. Or perhaps async functionality is not possible? Might also need a 5 min demo of how to better test my own work. (One day if someone has time?)
Thank you! 

Have pulled the id out from the row by using your suggestion. Have another look.

kael89: No worries @PShoe ! Happy to go through with you how you could test this feature locally and make sure that the functionality is correct

kael89 (approved):
That should do! 🚀

PR #1628, opened 2020-11-16
1229: Allow creating new options in an option set

chris-bes:
### Changes:

- Added functionality to allow creating a new option from an option set


packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
```diff
@@ -45,49 +45,59 @@ async function saveAnswer(models, answer, surveyResponseId) {
...
+      },
+    );
+    if (!surveyResponse.submission_time) {
+      // Survey responses from older versions of Tupaia MediTrak may not have a submission time
+      surveyResponse.submission_time = surveyResponse.end_time;
+      await surveyResponse.save();
     }
-    await Promise.all(answers.map(answer => saveAnswer(models, answer, surveyResponse.id)));
   } catch (error) {
-    throw error; // Throw error up for caller to handle
```
edmofro, line 87: Thank you for removing this redundant code! 🙏

packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
```diff
@@ -45,49 +45,59 @@ async function saveAnswer(models, answer, surveyResponseId) {
...
+    throw new DatabaseError(`creating/updating survey response with id ${surveyResponseId}`, error);
   }
+  await Promise.all(answers.map(answer => saveAnswer(models, answer, surveyResponse.id)));
 }
 
+const createOptions = async (models, optionsCreated) => {
+  return Promise.all(
+    optionsCreated.map(async option => {
```
edmofro, line 86: It might be good to process these sequentially. If the app has created two in the same option set, we may end up putting them as the same sort order if this happens in parallel: option a might be being added when option b looks for the largest sort order, and the db doesn't yet know about option a being added.

edmofro (changes requested):
Thanks for this @chris-bes, it's a solid set up. I have suggestions, but they are mainly minor (with one exception relating to a realm migration)

packages/meditrak-app/app/database/types/Option.js
```diff
@@ -11,8 +11,9 @@ export class Option extends RealmObject {
       id: this.id,
       value: this.value,
       label: this.label,
-      sortOrder: this.sortOrder,
+      sort_order: this.sortOrder,
       attributes: this.attributes ? JSON.parse(this.attributes) : {},
+      option_set_id: this.optionSetId,
```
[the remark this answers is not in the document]
chris-bes: I had a try with inverse relationship in Realm and seems like the linkingObjects property is _always going to be a collection_ containing all objects linked to that one (so maybe it's only meant for many to many relationships?). However, the relationship between Option and OptionSet is one to many (1 option only belongs to 1 option set). So to make it work, we have to add this to properties:
```
optionSets: {
      type: 'linkingObjects',
      objectType: 'OptionSet',
      property: 'options',
    },
```

And then maybe have a separate getter method to get our option set:
```
getOptionSet() {
    return this.optionSets[0];
  }
```

It seems that's the only work around. What do you reckon about this? I'm a bit afraid that it might cause some confusion when someone tries to access the Option's optionSet property. I guess we can also add some documentation?

chris-bes:
Really sorry about the delay... I accidentally messed up my meditrak-app environment, worse is I'm using windows so it took me a while to make it running again...

packages/meditrak-server/src/routes/postChanges.js
```diff
@@ -110,15 +117,15 @@ const constructEntitiesCreatedValidators = models => ({
 const constructIsValidEntity = models => async value =>
   new ObjectValidator(constructEntitiesCreatedValidators(models)).validate(value);
 
-const constructOptionsCreatedValidators = () => ({
+const constructOptionsCreatedValidators = models => ({
   id: [hasContent, takesIdForm],
   value: [hasContent],
-  option_set_id: [hasContent, takesIdForm],
+  option_set_id: [hasContent, takesIdForm, optionSetExist(models)],
```
edmofro, line 123:
```suggestion
  option_set_id: [hasContent, takesIdForm, constructRecordExistsWithId(models.optionSet)],
```
And can delete your custom validator

packages/meditrak-app/app/database/types/Option.js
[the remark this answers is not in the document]
edmofro: True! So we'd need to set up the primary link to be the other way around, so that `option.optionSet` was directly linked, and then `optionSets.options` used a `linkingObjects` list.

Can you spend 30 minutes looking at how hard it would be to migrate the primary link to be the other way around? If it's complex, just leave it as you have it here - it's pretty straightforward.

edmofro (approved):
One little tweak suggested

packages/meditrak-app/app/database/types/Option.js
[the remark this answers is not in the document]
chris-bes: I did also have a try with the other way around too, but it didn't work. The schema I had was:
```
Option: {
	optionSet: 'OptionSet'
}
```
and

```
OptionSet: {
	options: {
		type: 'linkingObjects',
		objectType: 'Option',
		property: 'optionSet',
    },
}
```

The reason it didn't work was, in our case, we want the Option.optionSet to be automatically updated/linked whenever we add a new option into OptionSet.options. But with the schema above, it's the other way around, the list of options will be automatically updated when we set optionSet in an option (because options are the `linkingObjects`). It also requires to change the logic of adding options into option set. So that's why I kinda gave up... Hope you're fine with that.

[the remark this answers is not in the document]
chris-bes: Have set it up! It's working fine now!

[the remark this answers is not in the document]
edmofro: Amazing! Thank you @chris-bes, I appreciate you putting in the extra effort

packages/meditrak-app/app/database/types/OptionSet.js
```diff
@@ -24,7 +24,11 @@ OptionSet.schema = {
   properties: {
```
edmofro, line 24: Can probably delete `addOptionIfUnique` from the model now.

packages/meditrak-app/app/database/types/OptionSet.js
```diff
@@ -24,7 +24,11 @@ OptionSet.schema = {
...
     id: 'string',
     name: { type: 'string', default: 'OptionSet not properly synced' },
-    options: 'Option[]',
+    options: {
```
edmofro, line 27: I presume the order the options are added doesn't matter, because they are ordered by `sortOrder` when pulled from the db?

edmofro (approved):
Please also add a comment to the testing notes that the upgrade from a fully synced previous version of the app should work, with regard to options

packages/meditrak-app/app/database/types/OptionSet.js
chris-bes: Sorry, I lost track of your comments in my email. Yep I think so, and when Autocomplete question is rendered, it always sorts the options by sortOrder

packages/meditrak-app/app/sync/syncMigrations.js
```diff
@@ -38,6 +38,10 @@ const migrations = {
   7: async (synchroniser, setProgressMessage) => {
     await resyncRecordTypes(synchroniser, setProgressMessage, ['survey_screen_component']);
   },
+  // Resync all entities so that the new attributes column come through
+  8: async (synchroniser, setProgressMessage) => {
+    await resyncRecordTypes(synchroniser, setProgressMessage, ['entity']);
+  },
```
chris-bes, line 44: @edmofro  yesterday when I was testing the upgrade with Sepi I found out that none of the attributes column in Entity were synced :(. `Attributes` data in Entity is new so I think resyncing all of the entities is the only way. What do you reckon?

edmofro: Damn damn damn, this is like 14000 records. I guess you're right though! What a shame, sucks for those people on low speed connections in the Pacific. Still, great catch guys.

packages/meditrak-app/app/database/schema.js
```diff
@@ -7,7 +7,7 @@ import * as DataTypes from './types';
 
 export const schema = {
   schema: Object.values(DataTypes),
-  schemaVersion: 18,
+  schemaVersion: 19,
```
chris-bes, line 10: I think schema version 18 for meditrak-app hasn't been released (I've increased from 17 to 18 in one of my recent  meditrak-app task that hasn't been released). So maybe what you just pushed is actually the fix for it?

edmofro: That's a good point, but I'm still ok with this going to 19 (and don't think bumping the app version was the fix)

I think this was the fix for Sepi's test, which was that she had a dev apk based on 18, and was upgrading to the feature branch here which was also 18. It's fine for us to jump 2 schema versions in one release, to make sure testing can happen properly :-)

edmofro (approved):

PR #1631, opened 2020-11-17
1591: Test data on production (UNFPA project, under Vanuatu)

chris-bes:
### Changes:

- Deleted Test District

- Will also have to delete the entity from DHIS2. Will do that when this is merged into dev/prod


edmofro (approved):

chris-bes:
@edmofro I'll merge this so it can be tested on dev

PR #1632, opened 2020-11-17
Start reports backend in CI/CD

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1282

I forgot that the `reports` package is a server, not an internal dependency 🤦‍♂️


edmofro (approved):
lgtm!

As discussed, the nginx conf will need to be manually deployed on production

edmofro:
Actually I just double-checked, and in fact it does get picked up on each deployment separately, so no need to manually deploy on production _until we release_ (but please do add it to the manual release steps for whichever release notes this goes out with)

kael89:
Merging in the feature branch as the api was successfully deployed in a feature instance.

### Deployment steps

- [ ]  Create a `pm2` process for `report-server`:
SSH in a (dev/prod) instance, the
```bash
  cd tupaia
   yarn workspace @tupaia/reports build
   pm2 start --name report-server dist --wait-ready --listen-timeout 15000 --time
```
- [ ] Add Route53 entries for `report-api.tupaia.org` in prod and dev
- [ ] Upload the `.env` variables  in AWS SSM - see https://docs.beyondessential.com.au/books/dev-ops/page/updating-environment-variables

PR #1633, opened 2020-11-17
1633 Permission checking: Get disasters

edmofro:

### DIFF
--- a/packages/meditrak-server/src/tests/routes/GETDisasters.test.js
+++ b/packages/meditrak-server/src/tests/routes/GETDisasters.test.js
@@ -1,0 +1,105 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { expect } from 'chai';
+import { findOrCreateDummyRecord } from '@tupaia/database';
+import { Authenticator } from '@tupaia/auth';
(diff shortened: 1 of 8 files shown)


packages/meditrak-server/src/tests/routes/GETDisasters.test.js
```diff
@@ -0,0 +1,105 @@
...
+    });
+    filterString = `filter={"id":{"comparator":"in","comparisonValue":["${disaster1.id}", "${disaster2.id}", "${disaster3.id}"]}}`;
+  });
+
+  afterEach(() => {
+    Authenticator.prototype.getAccessPolicyForUser.restore();
+  });
+
+  describe('GET /disasters/:id', async () => {
+    it('Sufficient permissions: Should return a requested disaster that user has access to', async () => {
```
EMcQ-BES, line 55: Something I've been trying to do more lately is avoid 'Should' in unit test descriptions, it's kind of redundant because every test should do something

edmofro: True, thanks for the reminder! We did discuss this in dev chat a while back too.

EMcQ-BES (approved):
Looks good! Longer term if we end up adding other CRUD endpoints for disasters we might wanna split out the permission checks to be reused, but for right now I think that's no big deal.

PR #1634, opened 2020-11-17
Add documentation for Jest test in Auth package

billli0:
### Issue #:
Temporarily there could be a timeout problem when testing. It will be investigated after all packages' tests have been converted to Jest. A documentation can reduce confusion for testing in this package.

### Changes:
Add documentation for termporary use.

### DIFF
--- a/packages/auth/README.md
+++ b/packages/auth/README.md
@@ -22,3 +22,8 @@ policy
 Note that parsing of that access policy is handled by the separate isomorphic package
 @tupaia/access-policy, as it is required on both the back ends and front end clients (admin-panel,
 meditrak-app, etc.)
+
+### Test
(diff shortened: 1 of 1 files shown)


kael89 (changes requested):
Hi @billli0 , I do not see much value in adding a documentation note for this. The auth package does not change often so most likely developers will not have to run its tests locally.

The biggest problem is solving this issue in our CI/CD pipeline, but documentation will not solve it, of course. I am sorry but I will have to suggest closing this PR without merging it. I created [1595: Fix auth tests failing locally](https://github.com/beyondessential/tupaia-backlog/issues/1595) so that we take the time to address this properly 🙌

billli0:
No worries! Today Andrew was having trouble testing locally so I think that could be helpful. Since there is a feature branch for it, yes we can close this one.

PR #1636, opened 2020-11-17
1130 get survey permissions

EMcQ-BES:
### Changes:

- Add GETSurvey and GETSurveyScreenComponent permissions handlers
- Add unit tests for those endpoints


edmofro (changes requested):
Same suggestions as previous PR too :-)

packages/meditrak-server/src/routes/CRUDHandler/helpers.js
```diff
@@ -3,8 +3,8 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-// if the endpoint is /survey/5a5d1c66ae07fb3fb025c3a3/answer, the resource is 'survey'
+// if the endpoint is /surveys/5a5d1c66ae07fb3fb025c3a3/answers, the resource is 'surveys'
 export const extractResourceFromEndpoint = endpoint => endpoint.split('/')[1];
 
-// if the endpoint is /survey/5a5d1c66ae07fb3fb025c3a3/answer, the child resource is 'answer'
+// if the endpoint is /surveys/5a5d1c66ae07fb3fb025c3a3/answers, the child resource is 'answers'
```
EMcQ-BES, line 9: Just a thought but have we accounted for every place we may need to singularise a resource? I know there's a line in the CRUDHandler where we singularise while looking up a model.

From what I've looked at I'm fairly sure we're fine but just wanted to note the thought.

edmofro: Yeah good thought. I'm certainly not 100% confident, but hopefully we find anywhere that's not in testing

edmofro (approved):

PR #1640, opened 2020-11-17
Sort reports package.json

kael89:
Sorted using `npx sort-package-json`


rohan-bes (approved):
Nice, this is awesome!

PR #1644, opened 2020-11-18
1384 Fix yarn `start-dev` when src is typescript

chris-bes:
### Changes:

- Modified backendStartDev to allow starting ts server

### DIFF
--- a/packages/reports/package.json
+++ b/packages/reports/package.json
@@ -20,7 +20,7 @@
     "lint:fix": "yarn lint --fix",
     "prestart": "npm run -s build",
     "start": "node dist",
-    "start-dev": "../../scripts/bash/backendStartDev.sh 9997",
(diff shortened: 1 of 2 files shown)


packages/reports/package.json
```diff
@@ -20,7 +20,7 @@
     "lint:fix": "yarn lint --fix",
     "prestart": "npm run -s build",
     "start": "node dist",
-    "start-dev": "../../scripts/bash/backendStartDev.sh 9997",
+    "start-dev": "../../scripts/bash/backendStartDev.sh 9997 -ts",
```
rohan-bes, line 23: is `9997` here the debugger port? Perhaps we can change it to something other than 9997 since that's the same as `web-config-server`

`9995`?

rohan-bes (approved):
Awesome work man, so glad this is finally fixed 😄 

Just one comment to change the debugger port, but approving anyway 👍

PR #1645, opened 2020-11-18
1283: Cleanup Reporting Package

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1283:

- Removed QueryBuilder and buildAggregation
-- These weren't actually necessary and were just a relic of when I copied everything over from web-config-server to get things working in the first place
- Added @types declarations for the 3rd party dependencies
- Added proper error handling for `fetchReport` endpoint
- Reworked Aggregator to be a lot more simple
- Changed to request param to `organisationUnitCodes` in order to support multiple orgUnitCodes


packages/report-server/src/aggregator/Aggregator.ts
```diff
@@ -1,56 +1,38 @@
...
-  }
-
   async fetchAnalytics(
-    dataElementCodes,
-    originalQuery,
-    replacementValues,
-    initialAggregationOptions = {},
+    dataElementCodes: string[],
+    organisationUnitCodes: string,
+    period = getDefaultPeriod(),
```
kael89, line 9: `period` is a DHIS specific concept, I believe we support just `startDate` and `endDate` in our internal data fetching (see https://github.com/beyondessential/tupaia/blob/dev/packages/data-api/src/fetchData.js#L39). I see us using `period` less and less in the future

But you can leave it as such for now and come back at a later stage to update it

packages/report-server/src/aggregator/Aggregator.ts
```diff
@@ -1,56 +1,38 @@
...
-      entityAggregationOptions,
-      hierarchyId,
+    return super.fetchAnalytics(
+      dataElementCodes,
+      {
+        organisationUnitCodes: organisationUnitCodes.split(','),
+        period,
+        dataServices: [{ isDataRegional: true }],
+      },
+      { aggregationType: 'RAW' },
```
kael89, line 18: It's a pity that `aggregateAnalytics` uses `MOST_RECENT` by default (one would expect it to be `RAW`)...

packages/report-server/src/app/addRoutesToApp.ts
```diff
@@ -15,5 +33,15 @@ export function addRoutesToApp(app) {
...
+  app.post('(/v[0-9]+)?/fetchReport/:reportCode', wrappedFetchReportWithAsyncCatch);
+
+  app.use(handleError);
 }
+
+const handleError = (err, req: ReportsRequest, res: Response) => {
+  let error = err;
+  if (!error.respond) {
+    error = new InternalServerError(err);
+  }
```
kael89, line 45: This is a way to avoid "let", although the existing code will also do!

```suggestion
  const error = err.respond ? err : new InternalServerError(err);
```

packages/report-server/src/routes/fetchReport.ts
```diff
@@ -24,15 +24,30 @@ class FetchReportRouteHandler {
...
-    const orgUnit = await models.entity.findOne({ code: query.organisationUnitCode });
-    if (!orgUnit) {
-      throw new Error(`No entity found with code ${query.organisationUnitCode}`);
-    }
     const permissionGroup = await models.permissionGroup.findById(report.permission_group_id);
+    const orgUnitCodes = query.organisationUnitCodes.split(',');
+    const orgUnitsAndCodes = await Promise.all(
+      orgUnitCodes.map(async orgUnitCode => ({
+        orgUnitCode,
+        orgUnit: await models.entity.findOne({ code: orgUnitCode }),
```
kael89, line 32: Interesting structure! Another option would be creating an `orgUnitByCode` dictionary, and check that all org units in the input are a key in it

kael89 (approved):
Awesome cleanup @rohan-bes !! Just one optional  suggestion, which is a matter of style actually so up to you

_Cleanup tip (which you may already know): `yarn workspace @tupaia/report-server lint` will show all linting errors (including TS ones)_

💯 👌

packages/report-server/src/aggregator/Aggregator.ts
rohan-bes: I think I may leave it for now if that's okay, and we can redesign this when we are actually implementing some reports. `startDate` and `endDate` are more effecient and sensible, so I'm happy with going with those in the future

packages/report-server/src/routes/fetchReport.ts
rohan-bes: Yeah it is a bit funny, I mainly did it so that I could play nicely with `Promise.all()`

Tbh I still haven't found a way of handling loops with async calls in a way that I'm happy with 😝 

I might leave things how they are for now? If you have a clean example of how to do this with an `orgUnitByCode` dictionary tho I would happily take it 😄

packages/report-server/src/aggregator/Aggregator.ts
kael89: Good call!

packages/report-server/src/routes/fetchReport.ts
kael89:
```js
import { reduceToSet } from '@tupaia/utils';

const requestedOrgUnitCodes = query.organisationUnitCodes.split(',');
const orgUnits = await models.entity.find({ code: orgUnitCodes });

const resultOrgUnitCodes = reduceToSet(orgUnits, 'code');
requestedOrgUnitCodes.forEach(code => {
  if (!resultOrgUnitCodes.has(code)) {
    throw new Error(`No entity found with code ${invalidOrgUnit.orgUnitCode}`);
  }
});
```

Actually, could you change that in a new PR cause this reduces multiple DB queries to one @rohan-bes ?

rohan-bes: Ah ofcourse! Why do i always forget you can return multiple db entries with one call 🤦 

Will do this 👍

PR #1646, opened 2020-11-18
create new datatrak app

tcaiger:
### Changes:

- Create new datatrak package

This is just the initial setup which is needed for all the other development. There is another [task to clean up the package ](https://github.com/beyondessential/tupaia-backlog/issues/1605)and remove all the meditrak legacy code etc.

---


edmofro (approved):
Will need to manually apply these two changes once they're in dev as well
- https://github.com/beyondessential/tupaia/pull/1628
- https://github.com/beyondessential/tupaia/pull/1510

PR #1650, opened 2020-11-19
1597 strive data grouped incorrectly

IgorNadj:
### Changes:

- Fix duplicate name causes incorrect grouping
- Add tests to cover bad grouping

---


packages/web-config-server/src/tests/apiV1/dataBuilders/helpers/groupEvents.test.js
```diff
@@ -48,139 +30,353 @@ const PARENT_ORG_UNIT = {
...
+        }),
+      ).to.eventually.deep.equal({
+        Wellington: [
+          {
+            orgUnit: 'NZ_WELLINGTON_CITY',
+            orgUnitName: 'Wellington',
+            dataValues: [{ dataElement: 'A', value: '1' }],
+          },
+        ],
+        "Other (NZ_WELLINGTON_OTHER)": [
```
IgorNadj, line 218: I chose to put the code in the output as a disambiguator. It will of-course look bad, and isn't meant to be used like this, but instead as a way to show the data steward/developer that they should clean up their data.

The alternative of throwing an error when we come across a duplicate, I thought was too dramatic.

packages/web-config-server/src/apiV1/dataBuilders/helpers/groupEvents.js
```diff
@@ -10,29 +10,71 @@ const getOrgUnits = async (models, { parentCode, type, hierarchyId }) => {
...
+/**
+ * Given an object of eventsByOrgUnitCode, returns the same object with the code changed to name, but
+ * disambiguated, e.g. 'Melbourne', 'Other', 'Other (1)'
+ *
+ * @param allOrgUnits
+ * @param eventsByOrgUnitCode
+ * @returns {{}}
+ * @private
+ */
+const mapOrgUnitCodeToUniqueOrgUnitName = (allOrgUnits, eventsByOrgUnitCode) => {
```
IgorNadj, line 22: There are no cases in production where this is an issue. 

The bug that is being fixed is fixed entirely in the methods `groupByAllOrgUnitNames` and `groupByAllOrgUnitParentNames` by changing to using codes during processing instead of names.

This method simply converts an object keyed by org unit code to an object keyed by org unit name.

The extra work it does to disambiguate non-unique names is done for completeness, as we may run across this problem in the future. And I have added tests to cover this.

packages/web-config-server/src/apiV1/dataBuilders/helpers/groupEvents.js
```diff
@@ -10,29 +10,71 @@ const getOrgUnits = async (models, { parentCode, type, hierarchyId }) => {
...
+  for (const orgUnitCode of Object.keys(eventsByOrgUnitCode)) {
+    orgUnitsByCode[orgUnitCode] = allOrgUnits.find(orgUnit => orgUnit.code === orgUnitCode);
+  }
+
+  const eventsByUniqueOrgUnitName = {};
+
+  for (const orgUnitCode of Object.keys(eventsByOrgUnitCode)) {
+    const orgUnit = orgUnitsByCode[orgUnitCode];
+
+    const isNameUnique =
```
nick-beyondessential, line 34: This is a nice method to filter uniqueness.

packages/web-config-server/src/apiV1/dataBuilders/helpers/groupEvents.js
```diff
@@ -10,29 +10,71 @@ const getOrgUnits = async (models, { parentCode, type, hierarchyId }) => {
...
+      Object.values(orgUnitsByCode)
+        .filter(otherOrgUnit => otherOrgUnit.code !== orgUnitCode)
+        .filter(otherOrgUnit => otherOrgUnit.name === orgUnit.name).length === 0;
+
+    if (isNameUnique) {
+      eventsByUniqueOrgUnitName[orgUnit.name] = eventsByOrgUnitCode[orgUnitCode];
+    } else {
+      const uniqueName = `${orgUnit.name} (${orgUnitCode})`;
+      eventsByUniqueOrgUnitName[uniqueName] = eventsByOrgUnitCode[orgUnitCode];
+    }
```
nick-beyondessential, line 44: This could be shortened to two lines if you wanted:
```suggestion
      const uniqueName = isNameUnique ? orgUnit.name : `${orgUnit.name} (${orgUnitCode})`;
      eventsByUniqueOrgUnitName[uniqueName] = eventsByOrgUnitCode[orgUnitCode];
```

nick-beyondessential (approved):
Looks good Igor, approving with one minor suggestion.

PR #1652, opened 2020-11-19
1478: Switch report-server to use @tupaia/expression-parser

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1478:

Centralised transform parsing logic into `TranformParser` which extends the `@tupaia\expression-parser` `ExpressionParser` class.


packages/report-server/reports.hjson
```diff
@@ -1,262 +0,0 @@
-{
```
rohan-bes, line 1: I've got a copy of this file if anyone wants it to copy/learn different report configs from

packages/report-server/src/reportBuilder/functions/basic.ts
```diff
@@ -1,11 +1,19 @@
...
-  return values.reverse()[0];
+  if (!Array.isArray(values)) {
+    throw new Error(`Function 'last' expected an array, but got: ${values}`);
+  }
+
+  if (values.length < 1) {
+    return undefined;
+  }
+
+  return values[values.length - 1];
```
rohan-bes, line 16: Pretty funny bug here, we were previously reversing the same array each time this was called, which lead to it returning last, then first, then last, etc... 😝

packages/report-server/src/reportBuilder/transform/parser/tranformParser.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Lookups object for rows within the transform data
+ *
+ * eg. if rows = [{BCD1: 5}, {BCD1: 8}, {BCD1: 4}], and currentRow = 1
+ * '$row.BCD1' => 8
+ * '$all.BCD1' => [5, 8, 4]
+ * '$allPrevious.BCD1' => [5, 8]
+ * '$where(f($otherRow) = $otherRow.BCD1 < $row.BCD1).BCD1' => [5, 4]
+ * '$row.BCD1 + sum($allPrevious.BCD1)' => 21
+ */
+type Lookups = {
```
rohan-bes, line 23: If anyone has a better name than `Lookup` I'm all ears. I'm not too happy with it but nothing better comes to mind...

packages/report-server/src/reportBuilder/transform/parser/tranformParser.ts
```diff
@@ -0,0 +1,89 @@
+/**
```
chris-bes, line 1: Should we rename this file to be `TransformParser` instead? Coz that's how we often name class files in tupaia?

chris-bes (approved):
Looks really awesome to me and I love the TransformParser class! Some nice tests added too. Glad the ExpressionParser helps a bit. I cannot find anything major so approving

PR #1653, opened 2020-11-19
Permissions checking: Update and Delete surveyResponses/answers

EMcQ-BES:
### Changes:

- Add delete and edit handlers for surveyResponses and answers endpoints
- Standardize endpoints as pluralized

This is branched from #1184, may cause some conflicts when we get to merging them both onto the epic branch.


edmofro (approved):
Nice work - I particularly like the survey entity pair stuff.

PR #1657, opened 2020-11-19
1601: Indicators - support intermediary values ("parameters")

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1601

A lot of changes here so I suggest reviewing one commit at a time


packages/indicators/src/IndicatorApi.ts
```diff
@@ -30,16 +30,23 @@ export class IndicatorApi implements IndicatorApiInterface {
...
 
-  async buildAnalytics(codes: string[], fetchOptions: FetchOptions): Promise<Analytic[]> {
-    const indicators = await this.models.indicator.find({ code: codes });
+  async buildAnalytics(indicatorCodes: string[], fetchOptions: FetchOptions): Promise<Analytic[]> {
+    const indicators = await this.models.indicator.find({ code: indicatorCodes });
+    return this.buildAnalyticsForIndicators(indicators, fetchOptions);
+  }
+
+  async buildAnalyticsForIndicators(
+    indicators: Omit<IndicatorFields, 'id'>[],
```
edmofro: You're using the Omit often enough to almost justify a new type

edmofro: Haha ignore - you do this next commit

packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.fixtures.ts
```diff
@@ -40,12 +39,15 @@ const ARRAY_ANALYTICS: ArrayAnalytic[] = [
...
   ['E_Pg_20192020', 'PG', '2020', 55],
+
+  ['Population', 'TO', '2019', 100],
+  ['Population', 'TO', '2020', 75],
 ];
 
-export const ANALYTIC_RESPONSE_FIXTURES: AnalyticResponseFixture[] = ARRAY_ANALYTICS.map(
-  ([dataElement, organisationUnit, period, value]) => ({
-    code: dataElement,
+export const ANALYTIC_RESPONSE_FIXTURES: AnalyticResponseFixture[] = AGGREGATOR_ARRAY_ANALYTICS.map(
```
edmofro, line 47: Another option
```
AGGREGATOR_ARRAY_ANALYTICS.map(arrayToAnalytic).map(analytic => ({ code: analytic.code, aggregations: [{ type: 'FINAL_EACH_YEAR' }], analytic }));
```

kael89: I like it!

packages/utils/src/__tests__/validation/validatorFunctions.test.js
```diff
@@ -78,4 +96,54 @@ describe('validatorFunctions', () => {
...
+    describe('fails if all values are of different types', () => {
+      const testData = [
+        ['array', [[1, '2'], { a: 1 }], `${JSON.stringify({ a: 1 })} is not an array`],
+        ['object', [{ a: 1 }, [{ b: 2 }]], `${JSON.stringify([{ b: 2 }])} is not an object`],
+        ['number', [1, '2'], "'2' is not a number"],
+        ['string', ['1', 2], '2 is not a string'],
+        ['boolean', [true, 'false'], "'false' is not a boolean"],
+      ];
+
+      it.each(testData)('%s', (type, value, expectedError) => {
```
edmofro, line 151: Clever use of the `type` as the test case description

packages/utils/src/validation/validatorFunctions.js
```diff
@@ -191,11 +193,16 @@ export const constructRecordExistsWithId = (modelOrDatabase, recordType) => asyn
...
-export const constructIsEmptyOr = validatorFunction => (value, object, key) => {
+/**
+ * @param {Function|Function[]} validators
+ */
+export const constructIsEmptyOr = validators => async (value, object, key) => {
   if (value !== undefined && value !== null && value !== '') {
-    return validatorFunction(value, object, key);
+    const runValidator = async validatorFunction => {
+      await validatorFunction(value, object, key);
+    };
```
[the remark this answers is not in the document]
edmofro: One difference is that we're no longer returning `true`. Do other validators return `undefined` as well? I know originally they were set up to "return `true` or throw an error"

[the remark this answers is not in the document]
kael89: Good question! A lot of them do not return `true`. I personally use one of the following two patterns for validation functions:
1. return either `true`/`false
2. throw an error if invalid (return void)

But I'm happy to add `return true` if you think it's preferable, although I would also suggest removing those statements from existing validators at some point

packages/indicators/src/builders/buildArithmetic.ts
```diff
@@ -102,18 +125,48 @@ const buildAnalyticValues = (analyticClusters: AnalyticCluster[], formula: strin
...
+  const formulaAnalytics = await fetchAnalytics(aggregator, aggregationsByCode, fetchOptions);
+  const formulaElements = Object.keys(aggregationsByCode);
 
-  return { analytics, dataElements };
+  const parameterAnalytics = await api.buildAnalyticsForIndicators(parameters, fetchOptions);
+  const parameterElements = parameters.map(p => p.code);
+
+  return {
+    analytics: formulaAnalytics.concat(parameterAnalytics),
+    dataElements: getUniqueEntries(formulaElements.concat(parameterElements)),
```
edmofro, line 166: Why would there be the same code in formulaElements and parameterElements? (i.e. why the `getUniqueEntries`)

packages/utils/src/validation/validationTypes.js
```diff
@@ -47,23 +48,36 @@ const getValidationType = value => {
       return VALIDATION_TYPES.BOOLEAN;
     case 'undefined':
       return VALIDATION_TYPES.UNDEFINED;
+    case 'object': {
+      if (Array.isArray(value)) {
+        return VALIDATION_TYPES.ARRAY;
+      }
+      if (isPojo(value)) {
+        return VALIDATION_TYPES.OBJECT;
```
edmofro, line 56: Nice, I like that this is explicitly checked now

edmofro (approved):
Wow, I didn't have any comments of significance across this large PR - amazing work!

packages/utils/src/validation/validatorFunctions.js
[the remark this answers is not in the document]
edmofro: No I'm fine to remove it, especially given we will need to regression test the `async` part anyway

packages/indicators/src/builders/buildArithmetic.ts
kael89: By accident 😛 Parameters would then take precedence over formula elements. Unique entries are used here mainly because  an element may appear more than once in a formula, so it's just about keeping our data structures clean (it's optional in practice)

kael89: By the way I haven't added a test case to assert the above behaviour, since I thought we could be classified as "unpredictable" (not part of the indicators' contract)

edmofro: Yep, happy with that!

PR #1664, opened 2020-11-22
Make endpoints consistently plural

edmofro:


packages/meditrak-server/src/app/addRoutesToApp.js
```diff
@@ -158,12 +158,13 @@ export function addRoutesToApp(app) {
...
-  app.post('(/v[0-9]+)?/user', registerUserAccount);
-  app.post('(/v[0-9]+)?/userAccount', createUserAccount);
+  app.post('(/v[0-9]+)?/user', registerUserAccount); // used for user registration on tupaia.org etc.
+  app.post('(/v[0-9]+)?/users', createUserAccount); // used by admin panel to directly create users
   app.post('(/v[0-9]+)?/userEntityPermissions', createUserEntityPermissions);
   app.post('(/v[0-9]+)/me/requestCountryAccess', requestCountryAccess);
   app.post('(/v[0-9]+)/me/changePassword', changePassword);
-  app.post('(/v[0-9]+)/surveyResponse', surveyResponse);
+  app.post('(/v[0-9]+)/surveyResponse', surveyResponse); // used by mSupply to directly submit data
+  app.post('(/v[0-9]+)/surveyResponses', surveyResponse);
```
EMcQ-BES, line 167: If we need to support both we could use the regex matching from express

```suggestion
  app.post('(/v[0-9]+)/surveyResponses?', surveyResponse);
```

Although writing it like this may make it more obvious what the line is doing

```suggestion
  app.post('(/v[0-9]+)/surveyResponse(s)?', surveyResponse);
```

EMcQ-BES (approved):

packages/meditrak-server/src/app/addRoutesToApp.js
edmofro: That's a solid suggestion, and I'm happy to make it, but I did think the comment on the line earlier makes things quite clear. What do you reckon?

PR #1669, opened 2020-11-22
1407 psss fix persisted state

tcaiger:
### Changes:

- Fix persisted state bug where no state was being remembered on non-development environments
- clean up store initialisation code
- Add remember me functionality (we forgot to create a card for this and there was an unused checkbox on the login form so I took this chance to add this)


packages/psss/src/utils/PersistGateProvider.js
```diff
@@ -8,23 +8,21 @@ import { PersistGate } from 'redux-persist/integration/react';
 import PropTypes from 'prop-types';
 
 function initPersistor(store) {
-  const persistor = persistStore(store, null);
+  const persistor = persistStore(store);
 
-  // if you run into problems with redux state, call "purge()" in the dev console
```
tcaiger, line 13: This was never being used and was a bit distracting in the code so I removed it.

packages/psss/src/containers/__tests__/weeklyReportsPanel.test.js
```diff
@@ -54,7 +56,13 @@ jest.mock('@tupaia/ui-components', () => ({
 }));
 
 function renderWeeklyReportsPanel() {
-  render(<WeeklyReportsPanel />, defaultState);
+  const history = createMemoryHistory();
+  render(
+    <Router history={history}>
```
edmofro, line 61: Is this replacing any existing history behaviour, or is it new?

packages/psss/src/utils/PersistGateProvider.js
```diff
@@ -8,23 +8,25 @@ import { PersistGate } from 'redux-persist/integration/react';
...
+  if (!window.sessionStorage.getItem('PSSS:activeSession')) {
+    const rememberMe = window.localStorage.getItem('PSSS:rememberMe');
 
-  window.purge = () => {
-    window.localStorage.setItem('queuePurge', 'true');
-    window.location.reload();
-  };
+    // If it is a new session and the user has not ticked rememberMe, then remove the user data from localStorage
+    if (rememberMe === 'false') {
+      persistor.purge();
```
edmofro, line 20: Ideally we'd like to keep their email address, but throw away everything else (so a partial rather than full purge). That could be easier through a redux persist rehydration transform, rather than through session and local storage (and would also mean the action handling all went through redux rather than setting two different things). This is a nice to have, so up to you whether you give it a go

edmofro (approved):
Approved but with a couple of questions

packages/psss/src/containers/__tests__/weeklyReportsPanel.test.js
tcaiger: This is new. The tests used to not need a router but recent changes mean that it does and so the tests started failing. I only just realised because psss tests are not on CI/CD yet. There is another card for that.

packages/psss/src/utils/PersistGateProvider.js
tcaiger: I initially set it up using a transform (in the same way as AdminPanel) but there is a big advantage to doing it this way which is:
- the redux data gets stored for the duration of the session even if rememberMe is not ticked. This is great because the user can still refresh the page and not get logged out or loose their page

On the admin panel if you did not tick the rememberMe checkbox and refresh you get booted out which could be frustrating

edmofro: Ah yep cool. For my own learning: what are the boundaries of a session?

tcaiger:
merging into dev and will be tested as part of psss phase1 testing

PR #1674, opened 2020-11-23
Assign correct hotel names to make them consistent

billli0:
### Issue #:
[1444: Dashboard: Quarantine sites by flight (COVID-19)](https://github.com/beyondessential/tupaia-backlog/issues/1444)

### Changes:
Reassign correct hotel names to the quarantine site names of inconsistent answers

### DIFF
--- a/packages/database/src/migrations/20201123014943-IntegrateInconsistentAnswersForHotelNames-modifies-data.js
+++ b/packages/database/src/migrations/20201123014943-IntegrateInconsistentAnswersForHotelNames-modifies-data.js
@@ -46,25 +46,22 @@ const convertData = [
 ];
 
 function checkHotelName(answer) {
-  const newAnswer = answer;
-  convertData.forEach(({ incorrectNames, correctName }) => {
-    if (incorrectNames.includes(answer.text)) {
-      newAnswer.text = correctName;
-    }
-  });
-  return newAnswer;
+  for (const { incorrectNames, correctName } of convertData) {
+    if (incorrectNames.includes(answer.text)) return { id: answer.id, text: correctName };
+  }
+  return null;
 }
 
 exports.up = async function (db) {
-  let { rows: answers } = await db.runSql(`
+  const { rows: answers } = await db.runSql(`
     select a.id, a."text" from answer a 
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20201123014943-IntegrateInconsistentAnswersForHotelNames-modifies-data.js
```diff
@@ -0,0 +1,77 @@
...
+  return newAnswer;
+}
+
+exports.up = async function (db) {
+  let { rows: answers } = await db.runSql(`
+    select a.id, a."text" from answer a 
+    join question q on a.question_id = q.id 
+    where q.text = 'Quarantine Site'
+  `);
+  answers = answers.map(a => ({ ...a, text: a.text.trim() }));
```
nick-beyondessential, line 64: Personally I'd const a new variable rather than let it but don't think it makes much difference, I just try avoiding lets as much as possible.

packages/database/src/migrations/20201123014943-IntegrateInconsistentAnswersForHotelNames-modifies-data.js
```diff
@@ -0,0 +1,77 @@
...
+  for (const answer of answers) {
+    const newAnswer = checkHotelName(answer);
+    await updateValues(db, 'answer', { text: newAnswer.text }, { id: newAnswer.id });
```
nick-beyondessential, line 67: @billli0 So does this update every record from the answers fetched whether changed or not? Would it be better to build an array of answers of ones changed by the check an only update those records?

billli0: Hi Nick. You are right, that makes more sense. I will change it now.

nick-beyondessential (approved):
LGTM

edmofro:
@billli0 should this be closed?

billli0:
@edmofro Yeah this can be closed. I will leave a reference to the ticket, which might be helpful in the future.

PR #1677, opened 2020-11-23
Support array and object aggregation in indicators

kael89:
### Issue #:
No issue - required for PSSS indicators (eg https://github.com/beyondessential/tupaia-backlog/issues/1338)

With this feature we'll be able to specify aggregations such as the following:

```js
aggregation: {
  type: 'FINAL_EACH_WEEK',
  config: { periodType: 'week', offset: 1 },
}
```


packages/indicators/src/builders/buildArithmetic/aggregationConfig.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { getVariables } from '@beyondessential/arithmetic';
+
+import { toArray } from '@tupaia/utils';
+import { ArithmeticConfig, AggregationDescriptor, AggregationSpecs } from './types';
+
+enum AggregationType {
```
kael89, line 11: Here I used the Typescript convention for [enums](https://www.typescriptlang.org/docs/handbook/enums.html):

* Singular, not plural word
* PascalCase for the enum and its members

I also considered using our existing convention for JavaScript "enum" constants:

* Plural word
* upper SNAKE_CASE

I'm still not sure about what's the best approach. I like the TypeScript convention, but consistency is key.

packages/indicators/src/builders/buildArithmetic/aggregationConfig.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+      if (Array.isArray(aggregation)) {
+        return AggregationType.Array;
+      }
+      if ('type' in aggregation) {
+        return AggregationType.Object;
+      }
+      return AggregationType.Dictionary;
+    default:
+      return AggregationType.Invalid;
```
rohan-bes, line 40: If we already know the aggregation type is invalid at this point, perhaps it's best just to throw an error here instead of having this type on the enum?

rohan-bes (changes requested):
Nice work! Just a comment about how the `Invalid` aggregation config type, as currently it seems we aren't handling it?

Happy with everything else tho 😄

packages/indicators/src/builders/buildArithmetic/configValidators.ts
```diff
@@ -0,0 +1,52 @@
...
+  defaultValues: Record<string, unknown>,
+  config: ArithmeticConfig,
+) => {
+  const variables = getVariables(config.formula);
+  Object.keys(defaultValues).forEach(code => {
+    if (!variables.includes(code)) {
+      throw new Error(`'${code}' is in defaultValues but not referenced in the formula`);
+    }
+  });
+};
```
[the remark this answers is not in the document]
kael89: Yeap that was @Alastair-L 's work 👏

packages/indicators/src/builders/buildArithmetic/aggregationConfig.ts
kael89: Good suggestion! Will do that

rohan-bes (approved):

PR #1543, opened 2020-10-27
Allow importing new responses without an initial response

edmofro:

PR #1681, opened 2020-11-23
1026: Fix import survey response permission check

chris-bes:
### Changes:

- Add validation if the submitted survey response entities belong to the
- Fixed some code conflicts that was not resolved from the last merge


packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
```diff
@@ -3,30 +3,55 @@
...
+      // Check if the country of the submitted survey response(s) matches with the survey countries.
+      if (!survey.country_ids.includes(surveyResponseCountry.id)) {
+        const surveyCountries = await models.country.findManyById(survey.country_ids);
+        const entities = entitiesByCountryCode[surveyResponseCountry.code];
+        const entityCodesString = entities.map(e => e.code).join(', ');
+        const surveyCountryNamesString = surveyCountries.map(s => s.name).join(', ');
+        throw new Error(
+          `Some survey response(s) are submitted against entity code(s) (${entityCodesString}) that do not belong to the countries (${surveyCountryNamesString}) of the survey '${survey.name}'`,
+        );
+      }
```
chris-bes, line 46: This if block is to fix the bug that Klaus found:
Comment: (https://github.com/beyondessential/tupaia-backlog/issues/1026#issuecomment-711557748)

> So survey response import doesn't validate against the countries of the survey? I just successfully imported a survey response with Fiji entity codes but the survey countries are Vanuatu, Demo Land and Solomon Islands. I'm confused.

Basically we also need to check if the entities of the submitted survey responses belong to the survey countries

packages/meditrak-server/src/routes/importSurveyResponses/importSurveyResponses.js
```diff
@@ -312,3 +311,15 @@ function getCellContents(sheet, columnIndex, rowIndex) {
...
+  const firstSurveyResponseId = getColumnHeader(sheet, INFO_COLUMN_HEADERS.length);
+  if (!firstSurveyResponseId) {
+    throw new Error(
+      'Each tab of the import file must have at least one previously submitted survey as the first entry',
+    );
+  }
+
+  const firstSurveyResponse = await models.surveyResponse.findById(firstSurveyResponseId);
+  return models.survey.findById(firstSurveyResponse.survey_id);
+}
```
chris-bes, line 331: There were some code conflicts that were not resolved when merging dev back into the epic branch so the changes in this file are mostly for cleaning it up

packages/meditrak-server/src/routes/importSurveyResponses/importSurveyResponses.js
```diff
@@ -312,3 +317,15 @@ function getCellContents(sheet, columnIndex, rowIndex) {
...
```
chris-bes, line 331: This was removed when the new Admin Panel feature was merged into the epic branch, not sure why but it is still needed when we're just updating existing survey responses

edmofro: Sorry @chris-bes this was probably my poor conflict resolution!

EMcQ-BES:
Can you update the unit tests to match these changes? Looks like some of them are using the error message for comparison, which has now changed

chris-bes:
I've noticed some of the tests I wrote before were failing. So I've pushed some more changes that fixed them. It's now down to around 20 tests failing

chris-bes:
> These lines shouldn't be necessary, as of #1543

I put the method back because I need a way of extracting the survey names to perform permission check when the users choose to import only existing survey responses (so no `surveyNames` provided in the query). The method is not used in the actual import logic

edmofro:
Oh that makes sense! Thanks.

EMcQ-BES (approved):
This all looks good to me, I've actually branched off to leverage some of this work so I'm pretty happy with having seen how it all works

PR #1682, opened 2020-11-24
1645 unit tests passing

edmofro:
Down to ~40 unit tests failing again, which contributes towards but does not achieve #1645


packages/meditrak-server/src/tests/routes/GETAnswers.test.js
```diff
@@ -37,6 +37,8 @@ describe('Permissions checker for GETAnswers', async () => {
   let filterString;
 
   before(async () => {
+    await resetTestData();
```
EMcQ-BES, line 40: Would it make more sense to reset test data after a batch of tests? And then a single batch can be assumed to have no side effects?

For now I suppose it makes more sense to keep it rather than find all the places that assumption isn't true

EMcQ-BES (approved):
Mostly just cleaning up data, not much new functionality, unit test failures on codeship reduced drastically. I'm happy with that.

packages/meditrak-server/src/tests/routes/GETAnswers.test.js
edmofro: I tried adding a reset to every batch by using a `beforeEach` in the scaffolding file, but it broke a lot more than it fixed, so yeah...good idea for the future.

PR #1683, opened 2020-11-24
1421 report dates in url

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1421

I have decided to leave all dates in `day` form, like so: `reportPeriod=1st_Sep_2020-30th_Sep_2020`
This has 2 advantages:
1. It better represents the actual state.
2. It can be interpreted without knowing the periodGranularity.

I think possibly the plan should be to pass start and end dates around in the string format? i.e. `W12020`. Either way, I don't think it's in the scope of this PR?


packages/web-frontend/src/historyNavigation/constants.js
```diff
@@ -14,14 +14,9 @@ const DASHBOARD = 'DASHBOARD';
 const MEASURE = 'MEASURE';
 const REPORT = 'REPORT';
 const PASSWORD_RESET_TOKEN = 'PASSWORD_RESET_TOKEN';
-const TIMEZONE = 'TIMEZONE';
-const START_DATE = 'START_DATE';
-const END_DATE = 'END_DATE';
-const DISASTER_START_DATE = 'DISASTER_START_DATE';
-const DISASTER_END_DATE = 'DISASTER_END_DATE';
```
kael89, line 21: Yay for all the simplification!

packages/web-frontend/src/selectors/dashboardSelectors.js
```diff
@@ -15,6 +20,11 @@ export const selectCurrentExpandedViewId = createSelector([selectLocation], loca
   getLocationComponentValue(location, URL_COMPONENTS.REPORT),
 );
 
+export const selectCurrentExpandedDates = createSelector([selectLocation], location => {
+  const startDateString = getLocationComponentValue(location, URL_COMPONENTS.REPORT_PERIOD) ?? '';
+  return convertUrlPeriodStringToDateRange(startDateString);
```
kael89, line 25:
```suggestion
  const periodString = getLocationComponentValue(location, URL_COMPONENTS.REPORT_PERIOD) ?? '';
  return convertUrlPeriodStringToDateRange(periodString);
```

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialogContent.js
```diff
@@ -219,6 +219,10 @@ export class EnlargedDialogContent extends PureComponent {
     const { onSetDateRange, isLoading, viewContent, isExporting } = this.props;
     const { periodGranularity, startDate, endDate } = viewContent;
 
+    if (!periodGranularity) {
+      return null;
+    }
```
kael89, line 224: Question: this makes a lot of sense, how come it wasn't an issue in the past? Or was it (does it fix an existing bug?)

kael89 (changes requested):
Awesome work @Alastair-L !! It's really exiting seeing this is almost ready to go, it will be a very useful functionality 🚀

Only a small renaming requested + a question

packages/web-frontend/src/historyNavigation/constants.js
Alastair-L: Tom's new export system is amazing

packages/web-frontend/src/selectors/dashboardSelectors.js
Alastair-L: Oops! Good catch

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialogContent.js
Alastair-L: I believe that this is completely unneeded! The only place this is rendered is in this file (line 122) and states:
`{periodGranularity && this.renderPeriodSelector()}`

kael89 (approved):
Looks great!

PR #1684, opened 2020-11-24
1657: Fix date range for aggregations

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1657

### Changes:
* Adds support for date range calculation when an aggregation array is used. This is a fix for https://github.com/beyondessential/tupaia-backlog/issues/1657
* Adjust date range for `OFFSET_PERIOD` aggregation to compensate for the provided offset


packages/aggregator/src/analytics/aggregateAnalytics/adjustTemporalDimensionsToAggregation.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+export const adjustTemporalDimensionsToAggregation = (temporalDimensions, aggregationOptions) => {
+  const adjustedDateRanges = getAggregations(aggregationOptions)
+    .map(aggregation => adjustDateRangeToAggregation(temporalDimensions, aggregation))
+    .filter(dr => !!dr);
+
+  if (adjustedDateRanges.length === 0) {
+    // No need to adjust the provided date ranges, returning the input
+    return temporalDimensions;
+  }
```
kael89, line 58: We could avoid this special logic by including the `temporalDimensions` from the input in the `adjustedDateRanges`, and then do the same steps as below (get date range extremes, calculate period from start and end dates).

However, there could be a regression risk in cases where `period` does not correspond to `startDate` and `endDate`, for whatever reason. I decided to avoid this by using the special logic above, as was the case in the existing code.

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/sumPreviousPerPeriod.js
```diff
@@ -87,3 +94,13 @@ const calculatePeriodsFromAnalytics = (analytics, aggregationPeriod, requestedPe
 
   return getPeriodsInRange(startPeriod, endPeriod);
 };
+
+export const getDateRangeForSumPreviousPerPeriod = (dateRange, _, periodType) => {
```
kael89, line 98: Using classes for aggregations could allow for a nice pattern, where two public methods would be part of their interface: `aggregate`, and `getDateRange`

rohan-bes: Yeah, it really seems like a good case from some class based logic here huh...

Still best to leave things as is for now.

rohan-bes (approved):
Looks good, nice work!

packages/aggregator/src/analytics/aggregateAnalytics/adjustTemporalDimensionsToAggregation.js
```diff
@@ -0,0 +1,61 @@
...
+ * @property {string} startDate
+ * @property {string} endDate
+ */
+
+/**
+ * Returns the minimum startDate and the maximum endDate in the provided ranges
+ */
+const getDateRangeExtremes = dateRanges => {
+  const [startDate] = dateRanges.map(dr => dr.startDate).sort(comparePeriods);
+  const [endDate] = dateRanges.map(dr => dr.endDate).sort((a, b) => comparePeriods(a, b) * -1);
```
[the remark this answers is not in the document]
kael89: Oh I like that!!

PR #1685, opened 2020-11-24
1338: Add PSSS WoW increase indicators

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1338

#### Dependencies (pending merge)
- [ ] https://github.com/beyondessential/tupaia/pull/1677
- [ ] https://github.com/beyondessential/tupaia/pull/1684

### DIFF
--- a/packages/database/src/migrations/20201117053359-AddPsssWoWIncreaseIndicators-modifies-data.js
+++ b/packages/database/src/migrations/20201117053359-AddPsssWoWIncreaseIndicators-modifies-data.js
@@ -1,0 +1,139 @@
+'use strict';
+
+import { arrayToDbString, generateId, insertObject } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const CONDITION_CODES = ['AFR', 'DIA', 'ILI', 'PF', 'DLI'];
+
+// eg 'PSSS_Confirmed_AFR_Site_Average'
+const psssCode = (descriptor, condition = '', isConfirmed = false) => {
+  const parts = ['PSSS', descriptor];
+  if (condition) {
+    parts.splice(1, 0, condition);
+  }
+  if (isConfirmed) {
+    parts.splice(1, 0, 'Confirmed');
+  }
+
+  return parts.join('_');
+};
+
+const insertIndicator = async (db, indicator) => {
+  await insertObject(db, 'indicator', { ...indicator, id: generateId() });
+  await insertObject(db, 'data_source', {
+    id: generateId(),
+    code: indicator.code,
+    type: 'dataElement',
+    service_type: 'indicator',
+  });
+};
+
+const createTotalSitesReportedIndicator = db =>
+  insertIndicator(db, {
+    code: psssCode('Total_Sites_Reported'),
+    builder: 'arithmetic',
+    config: {
+      // Just use a single data element for now, this logic will be extended in the future
+      formula: psssCode('Sites_Reported'),
+      aggregation: 'FINAL_EACH_WEEK',
+    },
+  });
+
+const createTotalCasesIndicator = async (db, condition) =>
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20201117053359-AddPsssWoWIncreaseIndicators-modifies-data.js
```diff
@@ -0,0 +1,139 @@
...
+    config: {
+      formula: `${cases} / ${sitesReported}`,
+      aggregation: 'FINAL_EACH_WEEK',
+    },
+  });
+};
+
+const createWowIncreaseIndicator = async (db, condition, isConfirmed) => {
+  const currentAverage = psssCode('Site_Average', condition, isConfirmed);
+  const prevAverage = 'siteAveragePrevWeek';
```
kael89, line 82: It may be useful to separate "data source" and "parameter" variables in a formula by using a different conventions. Since we use SNAKE_CASE for data sources, we could use cameCase for parameters. The formula will look like:

`PSSS_AFR_Site_Average / siteAveragePrevWeek`

packages/database/src/migrations/20201117053359-AddPsssWoWIncreaseIndicators-modifies-data.js
```diff
@@ -0,0 +1,139 @@
...
+
+const insertIndicator = async (db, indicator) => {
+  await insertObject(db, 'indicator', { ...indicator, id: generateId() });
+  await insertObject(db, 'data_source', {
+    id: generateId(),
+    code: indicator.code,
+    type: 'dataElement',
+    service_type: 'indicator',
+  });
+};
```
rohan-bes, line 42: This seems really handy, as I'm sure many of us may forget to create the `data_source` for the indicator when creating one. Think it's worth moving this to migration utils?

packages/database/src/migrations/20201117053359-AddPsssWoWIncreaseIndicators-modifies-data.js
```diff
@@ -0,0 +1,139 @@
...
+      aggregation: 'FINAL_EACH_WEEK',
+      parameters: [
+        {
+          code: prevAverage,
+          builder: 'arithmetic',
+          config: {
+            formula: currentAverage,
+            aggregation: [
+              'FINAL_EACH_WEEK',
+              { type: 'OFFSET_PERIOD', config: { periodType: 'week', offset: 1 } },
```
rohan-bes, line 98: So I'm guessing that once https://github.com/beyondessential/tupaia/pull/1684 is merged, then this will work even if the `offsetPeriod` doesn't fall within the requested period range? Ie. If I request this indicator with a period range of just a single week, it will still work correctly?

I'll review https://github.com/beyondessential/tupaia/pull/1684 as soon as I finish this one 😄

packages/database/src/migrations/20201117053359-AddPsssWoWIncreaseIndicators-modifies-data.js
kael89: That's a great point @rohan-bes ! And as a matter of fact I initially forgot to add the data sources, even though I've done this in the past. Then I lost time trying to understand why my data is not fetched properly 😆


The problem with migrations is that they are historically bound to the state of the schema when they run. This means that the same util may not work in the future (eg if we break `data_source` in two tables: `data_element` and `data_group`)

I'm still happy to add this to the `Utility Templates` section of the migration utils, as a source of documentation.

Idea: we could establish a new pattern: "versioned" functions, similar to versioned apis. Those could be either
`insertIndicatorV1`, or `insetIndicators(args, version)` where different logic is executed according to the specified version. I don't know if it's worth the maintenance overhead, just bringing this up for brainstorming.

packages/database/src/migrations/20201117053359-AddPsssWoWIncreaseIndicators-modifies-data.js
kael89: Yes it will still work correctly. This taken care of in [getDateRangeForOffsetPeriod](https://github.com/beyondessential/tupaia/blob/723e957eb283f525d0713b44a8c0bc688d671856/packages/aggregator/src/analytics/aggregateAnalytics/aggregations/offsetPeriod.js#L48), which will extend the requested range accordingly to compensate for the requested offset

PR #1686, opened 2020-11-24
Set up PSSS server

edmofro:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/1649

Goes a little further than the issue linked

Heavily inspired by report-server

I had started with a clean set of commits, but made some big design changes for the last few, so it is probably best to read the first few commits individually, and then the last four or five as a block


packages/psss-server/src/app/addRoutesToApp.ts
```diff
@@ -0,0 +1,44 @@
...
+const handleWith = (Handler: any) => (req: PsssRequest, res: Response, next: NextFunction) => {
+  const handler = new Handler(req, res, next);
+  return handler.handle();
+};
+
+const handleError = (
+  err: RespondingError | Error,
+  req: PsssRequest,
+  res: Response,
+  next: NextFunction,
```
edmofro, line 20: @rohan-bes this needed four arguments to know it should handle errors (plus see the first conditional taken from web-config-server)

packages/psss-server/examples.http
```diff
@@ -0,0 +1,19 @@
+### /login
+POST http://localhost:8040/v1/login HTTP/2.0
+content-type: application/json
+
+{
+    "emailAddress": "username",
+    "password": "password"
+}
```
rohan-bes, line 8: We are going to add a `http.env` file soon to the reporting package so that devs can save credentials there without needing to worry about accidentally committing them to the repo (https://github.com/beyondessential/tupaia-backlog/issues/1662)
We will do it for psss-server there too!

packages/psss-server/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@tupaia/psss-server",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Server for building report data",
```
rohan-bes, line 5:
```suggestion
  "description": "Backend server for PSSS application",
```
?

packages/psss-server/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "babel src --out-dir dist --source-maps --ignore 'src/__tests__/**' --extensions '.ts' --config-file '../../.babelrc-ts.js'",
+    "lint": "tsc; eslint . --ext .ts",
+    "lint:all": "yarn run lint \"src/**/*.{ts,jsx}\"",
+    "lint:fix": "yarn lint --fix",
+    "prestart": "npm run -s build",
+    "start": "node dist",
+    "start-dev": "../../scripts/bash/backendStartDev.sh 9995 -ts",
```
rohan-bes, line 23:
```suggestion
    "start-dev": "../../scripts/bash/backendStartDev.sh 9991 -ts",
```
May want to change the debugger port, incase you want to run `report-server` and `psss-server` at the same time?

packages/psss-server/src/app/addRoutesToApp.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { Express, Request, Response, NextFunction } from 'express';
+import { InternalServerError, RespondingError } from '@tupaia/utils';
+import { PsssRequest } from '../types';
+import { LoginHandler, LogoutHandler, TestHandler } from '../routes';
+import { RouteHandler } from '../routes/RouteHandler';
+
+const handleWith = (Handler: typeof RouteHandler) => (
```
rohan-bes, line 12: Oh this is really cool, I didn't realise you could use `typeof` on a class like this 👍

packages/psss-server/src/models/PsssSession.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { DatabaseModel, DatabaseType } from '@tupaia/database';
+import { AccessPolicy } from '@tupaia/access-policy';
+
+export class PsssSessionType extends DatabaseType {
+  static databaseType = 'psss_session';
```
rohan-bes, line 11: Spell `PsssSession` fast 7 times in a row 😝

packages/psss-server/src/routes/LogoutHandler.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { UnauthenticatedRouteHandler } from './UnauthenticatedRouteHandler';
+
+export class LogoutHandler extends UnauthenticatedRouteHandler {
+  async buildResponse() {
+    this.req.sessionCookie?.reset();
```
rohan-bes, line 10: Oooh love the use of this `sessionCookie?.reset()`. I keep forgetting this functionality exists now 😂

packages/psss-server/src/routes/RouteHandler.ts
```diff
@@ -0,0 +1,72 @@
...
+      throw new UnauthenticatedError('User not authenticated');
+    }
+
+    const session: PsssSessionType = await this.sessionModel.findById(sessionId);
+    if (!session) {
+      throw new UnauthenticatedError('Session not found in database');
+    }
+
+    const { accessPolicy } = session;
+    const authorized = accessPolicy.allowsAnywhere(PSSS_PERMISSION_GROUP);
```
rohan-bes, line 61: Will the `PSSS Sentinel Site Officer` user group have any access to the PSSS app? Or is this just placeholder code for now before something more complex is introduced?

rohan-bes (commented):
Mostly finished review, but may do one more check through later (have a meeting now)

Looks really good, and I'll defs copy some of the error handling code through 👍

packages/psss-server/src/routes/RouteHandler.ts
edmofro: No I'm pretty sure they have no access to the psss app

packages/psss-server/src/models/PsssSession.ts
edmofro: Haha I thought the same! Originally I had it with capital S's, but that was just impossible to parse

packages/psss-server/package.json
edmofro: Good call

packages/psss-server/package.json
edmofro: Oops, thanks

packages/psss-server/examples.http
edmofro: Ah awesome! Let me know when you've done that.

packages/database/src/migrations/20201124054820-AddPsssSessionTable-modifies-schema.js
```diff
@@ -0,0 +1,36 @@
...
+
+exports.up = function (db) {
+  return db.createTable('psss_session', {
+    columns: {
+      id: { type: 'text', primaryKey: true },
+      email: { type: 'text', notNull: true },
+      access_policy: { type: 'text', notNull: true },
+      access_token: { type: 'text', notNull: true },
+      refresh_token: { type: 'text', notNull: true },
+    },
```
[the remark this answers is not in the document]
edmofro: Yeah I actually thought about storing access_policy as jsonb, but because the constructor of `AccessPolicy` handles a string and parses it, there's really no need, so I leaned towards keeping this database table as dumb as possible. I can't think of instances where we would be searching by properties within the access policy, and we certainly shouldn't be manipulating it. 

There's not a huge argument either way though, so I'm happy to be talked around

packages/psss-server/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "test": "jest",
+    "test:coverage": "jest --coverage",
+    "test:watch": "jest --watch"
+  },
+  "dependencies": {
+    "@babel/polyfill": "^7.0.0",
+    "@tupaia/access-policy": "3.0.0",
+    "@tupaia/database": "1.0.0",
+    "@tupaia/utils": "1.0.0",
+    "@types/client-sessions": "^0.8.0",
```
rohan-bes, line 34: I think `@types` dependencies are just required as `devDependencies`. There's also an argument for adding them to the `root` devDependencies, but I'm not too fussed 😄

packages/psss-server/.env.example
```diff
@@ -0,0 +1,8 @@
+CLIENT_BASIC_AUTH_HEADER=
+DB_NAME=
+DB_PASSWORD=
+DB_URL=
+DB_USER=
```
rohan-bes, line 5: So in the end I guess we did need a db connection for the sake of session management huh? Makes me wonder if perhaps switching to a different system for session management might not be a bad idea? Perhaps a small file based db per app which just handles it's own sessions?

Obviously not for now, just a general thought 👍

packages/devops/configs/servers.conf
```diff
@@ -74,6 +74,50 @@ server {
         internal;
   }
 
+}
+server {
+
+  listen       443 ssl http2;
+  listen       [::]:443 ssl http2;
+  server_name  ~^.*psss-api\.tupaia\.org$;
```
rohan-bes, line 82: not actually related to your changes at all, but this reminds me that we need to add `psss` to the reserved endings in `validateBranchName.sh`

packages/database/src/migrations/20201124054820-AddPsssSessionTable-modifies-schema.js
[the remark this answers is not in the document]
rohan-bes: Yeah I don't think there's a particularly strong argument for it, outside of:
1. It's consistent with the existing `userSession` table (which we have already deviated from a bit here, though for good reasons imo)
2. Having `jsonb` field type will enforce that we enter data in the correct format. But given that it's created programmatically once again I don't see this as a big issue.
3. Most importantly, on the incredibly rare occasion that I want to manually inspect the policy in my db viewer, it displays in pretty json by default 😛 

So yeah, I'm not too fussed either 👍

rohan-bes (approved):
Just a few more general comments, and a request to move the `@types` dependency to `devDependencies`

Approving anyway as it's just a minor thing. Awesome work, great that you got a hang of typescript so quickly, and the package is looking very clean 👍

packages/database/src/migrations/20201124054820-AddPsssSessionTable-modifies-schema.js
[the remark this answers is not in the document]
edmofro: Well they're not super strong, but I've been swayed. Have updated in the subsequent PR :-)

packages/psss-server/package.json
edmofro: Oh right cool! Will update in the PR off this one, because I made the same mistake there

packages/psss-server/.env.example
edmofro: Yeah I actually looked into other options, including a file based db. They come with some risks that would need to be thought through though, and in the end I figured...I've got ready access to a nice database, I might as well use it for now. Notice though that the only thing passed around is the model for the psss_session table, so it kinda restricts it to have single-table access.

packages/devops/configs/servers.conf
edmofro: Oh true - I might as well slip this into the next PR!

PR #1687, opened 2020-11-24
1023 Permission Checking: Submit survey responses

EMcQ-BES:
### Changes:

- Added permissions checking to POST surveyResponse and postChanges endpoints

### DIFF
--- a/packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
+++ b/packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
@@ -4,7 +4,7 @@
  */
 
 import { flattenDeep, groupBy, keyBy } from 'lodash';
-import { reduceToDictionary } from '@tupaia/utils';
+import { getUniqueEntries, reduceToDictionary } from '@tupaia/utils';
 
 export const assertCanImportSurveyResponses = async (
   accessPolicy,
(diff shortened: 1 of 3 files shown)


packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
```diff
@@ -57,3 +57,29 @@ export const assertCanImportSurveyResponses = async (
...
+  // Assumes the data has already been validated
+  const entitiesGroupedBySurveyName = {};
+
+  for (const response of surveyResponses) {
+    // There are three valid ways to refer to the entity in a batch change
+    const entityCode =
+      response.entity_code ||
+      (response.entity_id
+        ? (await models.entity.findById(response.entity_id)).code
+        : (await models.facility.findById(response.clinic_id)).code);
```
kael89, line 75: Let's break this into a separate method, where we return early if a value is found, to avoid nesting and parentheses

packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
```diff
@@ -57,3 +57,29 @@ export const assertCanImportSurveyResponses = async (
...
   return true;
 };
+
+export const assertSurveyResponseBatchPermissions = async (
+  accessPolicy,
+  models,
+  surveyResponses,
+) => {
+  // Assumes the data has already been validated
+  const entitiesGroupedBySurveyName = {};
```
kael89, line 67: The name you used is still fine! I believe the `grouped` part is implied by the plural in `entities`. Those are implicit conventions used in our codebase:
* `entityBySurveyName` (or more commonly `surveyNameToEntity`) : object of `surveyName` key => `entity` value
* `entitiesBySurveyName`: object of `surveyName` key => `entity[]` value

```suggestion
  const entitiesBySurveyName = {};
```

This is also consistent to the name used in `assertCanImportSurveyResponses()`

packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
```diff
@@ -57,3 +57,29 @@ export const assertCanImportSurveyResponses = async (
...
+
+  for (const response of surveyResponses) {
+    // There are three valid ways to refer to the entity in a batch change
+    const entityCode =
+      response.entity_code ||
+      (response.entity_id
+        ? (await models.entity.findById(response.entity_id)).code
+        : (await models.facility.findById(response.clinic_id)).code);
+    const surveyName = (await models.survey.findById(response.survey_id)).name;
```
kael89, line 76: A consideration here is that we will always perform 1 or 2 db queries per survey response

1. Potentially one for the entity (if no `entity_code` is provided)
2. Definitely one for the survey name

But it may be that eg 20 survey responses are submitted for a single survey and entity, which would make most of those queries redundant.

Overcoming this would require a more elaborate algorithm where we find all distinct entity/clinic/survey ids in the responses first, fetch the records next (3 queries max, for the whole batch) and then transform them in appropriate dictionaries to be used to eventually create `entitiesBySurveyName`.

I understand that this makes code more complicated and they say that premature optimisation is the root of all evil, so let's leave this as is. You may still want to attempt an compromise by pre-fetching distinct surveys, since this wouldn't be too complicated. `reduceToDictionary` is a useful util to then map survey ids to names.

packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
```diff
@@ -57,3 +57,29 @@ export const assertCanImportSurveyResponses = async (
 
   return true;
 };
+
+export const assertSurveyResponseBatchPermissions = async (
```
kael89, line 61: To use the existing `assertCan{{verb}}{{noun}}` assertion naming pattern

```suggestion
export const assertCanSubmitSurveyResponses = async (
```

kael89 (changes requested):
Looks good overall, nice work!

packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
EMcQ-BES: Fair enough, I've added pre-fetching of unique surveys for now

packages/meditrak-server/src/routes/importSurveyResponses/assertCanImportSurveyResponses.js
```diff
@@ -58,28 +58,42 @@ export const assertCanImportSurveyResponses = async (
...
+  if (surveyResponse.entity_id) {
+    const entity = await models.entity.findById(surveyResponse.entity_id);
+    return entity.code;
+  }
+  if (surveyResponse.clinic_id) {
+    const clinic = await models.facility.findById(surveyResponse.clinic_id);
+    return clinic.code;
+  }
+
+  throw new Error('Survey response change does not contain valid entity reference');
```
kael89, line 76: Nice! 👌

kael89 (approved):
Looks great, thanks for implementing the suggestions!!

PR #1690, opened 2020-11-25
Refactor psss to react query

tcaiger:
There is no issue - this is just a refactor to improve the maintainability of the code in preparation for integration with the server

### Changes:

- Add [react-query](https://github.com/tannerlinsley/react-query) to project
- Add useTableQuery hook which uses react-query
- Refactor fetching tables to useTableQuery

There is another PR for setting up the post handling in react-query
---


packages/psss/src/index.js
```diff
@@ -5,12 +5,14 @@
...
 import { render as renderReactApp } from 'react-dom';
 import App from './App';
+import { ReactQueryDevtools } from 'react-query-devtools';
 import { AppProviders } from './AppProviders';
 import { initStore } from './store/store';
 
 const render = () => {
   return renderReactApp(
     <AppProviders store={initStore()}>
+      <ReactQueryDevtools />
```
edmofro, line 15: Should we be turning this off in production? (honestly don't know)

edmofro (changes requested):
Absolutely love the query hook - it's awesome!

Seems a bit of a loss to take away `ConnectedTable`, given that almost all of the tables use the same state in the same exact way? Could reduce on what seems like unnecessary boilerplate to add a wrapper component back in imo. Should also make it more maintainable - if we want to e.g. change a prop name, there's just one file to change it in.

packages/psss/src/index.js
tcaiger: That's a good question. I hadn't thought of that and had to look it up but no it's fine to leave it like that - it turns off automatically in production.

https://github.com/tannerlinsley/react-query-devtools/blob/master/index.js

tcaiger:
> Absolutely love the query hook - it's awesome!
> 
> Seems a bit of a loss to take away `ConnectedTable`, given that almost all of the tables use the same state in the same exact way? Could reduce on what seems like unnecessary boilerplate to add a wrapper component back in imo. Should also make it more maintainable - if we want to e.g. change a prop name, there's just one file to change it in.

Yeah I was on the fence when I removed that as well. I don't really mind but the reason I decided to remove it was that it seemed like one abstraction too many that wouldn't do much and only save 4 or 5 lines of code in each of the tables. I kind of feel like the extra abstraction would make the use of the querying and the table less transparent and there are enough differences in the tables (TableHeader, TableBody, onRowClick) that it is nice to see how they are configured without looking in another place.

edmofro:
Sweet, well that's a decent justification :-)

edmofro (approved):

PR #1692, opened 2020-11-25
1658: Use expression parser in indicators

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1658


kael89:
@chris-bes Unfortunately the commit history is a bit messed up, cause I "squashed and merged" the base branch for this 🤦‍♂️ You can still follow commit history from **Add TS type generation for expression-parser** onwards, or you can just have a look at the final diff

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -56,6 +56,14 @@ export class ExpressionParser {
...
   }
 
+  evaluateToNumber(expression) {
+    const result = this.evaluate(expression);
+    if (typeof result === 'boolean') {
+      return result ? 1 : 0;
+    }
+    return result;
+  }
+
```
chris-bes, line 66: Nice!

packages/indicators/src/builders/buildArithmetic/buildArithmetic.ts
```diff
@@ -40,14 +40,16 @@ const fetchAnalyticClusters = async (
...
-      organisationUnit,
-      period,
-      value: runArithmetic(formula, dataValues),
-    }))
+const buildAnalyticValues = (analyticClusters: AnalyticCluster[], formula: string) => {
+  const parser = new ExpressionParser();
+
+  return analyticClusters
+    .map(({ organisationUnit, period, dataValues }) => {
+      parser.setScope(dataValues);
```
chris-bes, line 48: `parser.setScope()` will just keep setting new values into the scope without clearing the previous values. Do you think there might be any side effects if we use the same parser (which will share same scope) to evaluate the formula for multiple analytic clusters? Or is it safer to maybe instantiate new parser in the for loop?

packages/indicators/src/__tests__/builders/buildArithmetic/buildArithmetic.test.ts
```diff
@@ -358,60 +358,77 @@ describe('buildArithmetic', () => {
...
-          formula: 'A_To_2020 + Undefined_To_2019',
-          aggregation: 'FINAL_EACH_YEAR',
-        },
-        [],
+    const testData: [string, string, number[]][] = [
+      ['simple expression - integer result', 'One + Two', [3]],
+      ['simple expression - float result', 'One / Two', [1 / 2]],
+      ['complex expression', '((One + Two) * Three) / (Four - Five)', [((1 + 2) * 3) / (4 - 5)]],
+      ['division with zero', 'One / (One + Two - Three)', []],
+      ['some data elements are undefined in the orgUnit/period combo', 'One + Undefined', []],
```
chris-bes, line 372: Much more readable!

chris-bes (commented):
Looks cool to me! I just have 1 question about `parser.setScope()` in `buildAnalyticValues`. If you think there are no side effects doing that then I'm happy to approve this PR!

packages/indicators/src/builders/buildArithmetic/buildArithmetic.ts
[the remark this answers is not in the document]
kael89: That's a really good observation! I believe it won't cause issues in the current implementations, since we make sure that every cluster includes all data elements (see `checkClusterIncludesAllElements`), so the scope should be complete every time.

But this may change in the future and has a bit of a risk. I will confirm that there is no significant performance toll, and if there is not I'll change it as per your suggestion!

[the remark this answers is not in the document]
kael89: Or I could simply use `clearScope`, and problem solved!

chris-bes (approved):
LGTM!

PR #1693, opened 2020-11-25
1521 Fix org unit children not showing in mobile

IgorNadj:
### Changes:

- Fix org unit children not showing in mobile


kael89 (approved):
Cute little fix!

PR #1695, opened 2020-11-25
1409: PSSS Confirmed Weekly Report

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1409:

Added `Confirmed Weekly Report` Report

### DIFF
--- a/packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
+++ b/packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
@@ -1,0 +1,91 @@
+'use strict';
+
+import { codeToId, insertObject, generateId } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+const permissionGroupNameToId = async (db, name) => {
+  const record = await db.runSql(`SELECT id FROM permission_group WHERE name = '${name}'`);
+  return record.rows[0] && record.rows[0].id;
+};
+
+const REPORT_CODE = 'PSSS_CONFIRMED_WEEKLY_REPORT';
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  return insertObject(db, 'report', {
+    id: generateId(),
+    code: REPORT_CODE,
+    config: {
+      fetch: {
+        dataElements: [
+          'PSSS_Confirmed_Sites_Reported',
+          'PSSS_Confirmed_Sites',
+          'PSSS_Confirmed_AFR_Cases',
+          'PSSS_Confirmed_AFR_Alert_Threshold_Crossed',
+          'PSSS_Confirmed_DIA_Cases',
+          'PSSS_Confirmed_DIA_Alert_Threshold_Crossed',
+          'PSSS_Confirmed_ILI_Cases',
+          'PSSS_Confirmed_ILI_Alert_Threshold_Crossed',
+          'PSSS_Confirmed_PF_Cases',
+          'PSSS_Confirmed_PF_Alert_Threshold_Crossed',
+          'PSSS_Confirmed_DLI_Cases',
+          'PSSS_Confirmed_DLI_Alert_Threshold_Crossed',
+        ],
+      },
+      transform: [
+        'keyValueByDataElementName',
+        {
+          transform: 'select',
+          "'period'": "convertToPeriod($row.period, 'WEEK')",
+          '...': '*',
+        },
+        {
+          transform: 'aggregate',
+          period: 'group',
+          organisationUnit: 'group',
+          '...': 'last',
+        },
+        {
+          transform: 'select',
+          "'AFR'": '$row.PSSS_Confirmed_AFR_Cases',
+          "'AFR Threshold Crossed'": '$row.PSSS_Confirmed_AFR_Alert_Threshold_Crossed == 1',
+          "'DIA'": '$row.PSSS_Confirmed_DIA_Cases',
+          "'DIA Threshold Crossed'": '$row.PSSS_Confirmed_DIA_Alert_Threshold_Crossed == 1',
+          "'ILI'": '$row.PSSS_Confirmed_ILI_Cases',
+          "'ILI Threshold Crossed'": '$row.PSSS_Confirmed_ILI_Alert_Threshold_Crossed == 1',
+          "'PF'": '$row.PSSS_Confirmed_PF_Cases',
+          "'PF Threshold Crossed'": '$row.PSSS_Confirmed_PF_Alert_Threshold_Crossed == 1',
+          "'DLI'": '$row.PSSS_Confirmed_DLI_Cases',
+          "'DLI Threshold Crossed'": '$row.PSSS_Confirmed_DLI_Alert_Threshold_Crossed == 1',
+          "'Sites'": '$row.PSSS_Confirmed_Sites',
+          "'Sites Reported'": '$row.PSSS_Confirmed_Sites_Reported',
+          '...': ['period', 'organisationUnit'],
+        },
+      ],
+    },
+    permission_group_id: await permissionGroupNameToId(db, 'PSSS'),
+  });
+};
+
+exports.down = function (db) {
+  return db.runSql(`
+    DELETE FROM report
+    WHERE code = '${REPORT_CODE}';
+  `);
+};
+
+exports._meta = {
+  version: 1,
+};
--- a/packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
+++ b/packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
@@ -11,7 +11,7 @@ const permissionGroupNameToId = async (db, name) => {
   return record.rows[0] && record.rows[0].id;
 };
 
-const REPORT_CODE = 'PSSS_CONFIRMED_WEEKLY_REPORT';
+const REPORT_CODE = 'PSSS_Confirmed_Weekly_Report';
 
 /**
  * We receive the dbmigrate dependency from dbmigrate initially.
@@ -46,17 +46,8 @@ exports.up = async function (db) {
       },
       transform: [
         'keyValueByDataElementName',
-        {
-          transform: 'select',
-          "'period'": "convertToPeriod($row.period, 'WEEK')",
-          '...': '*',
-        },
-        {
-          transform: 'aggregate',
-          period: 'group',
-          organisationUnit: 'group',
-          '...': 'last',
-        },
+        'convertPeriodToWeek',
+        'aggregateFirstValuePerPeriodPerOrgUnit',
         {
           transform: 'select',
           "'AFR'": '$row.PSSS_Confirmed_AFR_Cases',
--- a/packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
+++ b/packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
@@ -47,7 +47,7 @@ exports.up = async function (db) {
       transform: [
         'keyValueByDataElementName',
         'convertPeriodToWeek',
-        'aggregateFirstValuePerPeriodPerOrgUnit',
+        'firstValuePerPeriodPerOrgUnit',
         {
           transform: 'select',
           "'AFR'": '$row.PSSS_Confirmed_AFR_Cases',
--- a/packages/report-server/src/reportBuilder/reportBuilder.ts
+++ b/packages/report-server/src/reportBuilder/reportBuilder.ts
@@ -9,33 +9,25 @@ interface BuildReport {
 }
 
 export class ReportBuilder {
-  readonly config: ReportConfig;
+  config?: ReportConfig;
 
-  readonly aggregator: Aggregator;
+  testData?: Row[];
 
-  readonly query: FetchReportQuery;
-
-  readonly testData: Row[];
-
-  constructor(
-    config: ReportConfig,
-    aggregator: Aggregator,
-    query: FetchReportQuery,
-    testData: Row[] = [],
-  ) {
+  setConfig = (config: ReportConfig) => {
     this.config = config;
-    this.aggregator = aggregator;
-    this.query = query;
-    this.testData = testData;
-  }
+  };
 
-  build = async (): Promise<BuildReport> => {
+  setTestData = (testData: Row[]) => {
+    this.testData = testData;
+  };
+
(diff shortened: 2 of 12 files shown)


packages/report-server/src/types.ts
```diff
@@ -18,9 +18,19 @@ export interface FetchReportParams extends ParamsDictionary {
...
+    dataGroups?: string[];
+  };
+  transform: (string | Record<string, unknown>)[];
+}
+
 interface ReportsRequestBody {
   emailAddress: string;
   password: string;
+  testConfig?: ReportConfig;
+  testData?: Record<string, string | number>[];
```
rohan-bes, line 33: Added these two optional parameters, which allow to provide either a `testConfig` object (representing a report.config object) or `testData` representing an array or analytics or events. These can be really useful for testing when data isn't available, or for iteratively developing a transform config without needing to edit the database.

I'm thinking that ultimately it might be nice to break this out into a separate endpoint (perhaps `/report/:id/fetch` and `/report/:id/test`) but I still haven't made my mind up about it yet

packages/report-server/src/routes/fetchReport.ts
```diff
@@ -18,7 +18,19 @@ class FetchReportRouteHandler {
...
+      const reportBuilder = new ReportBuilder(
+        body.testConfig,
+        this.aggregator,
+        query,
+        body.testData,
+      );
+      const data = await reportBuilder.build();
+      respond(res, data, 200);
+      return;
+    }
```
kael89, line 32: Let's encapsulate this in a separate function

packages/report-server/src/types.ts
kael89: A separate endpoint would be great for keeping things clean and decoupled! As long as the actual behaviour can easily be invoked with test data passed in (to avoid duplication)

I guess your current implementation is still pretty clean, so we may as well leave as is for now

packages/report-server/src/types.ts
```diff
@@ -40,7 +50,5 @@ export interface Event {
   eventDate: string;
   orgUnitName: string;
   orgUnit: string;
-  dataValues?: {
-    [key: string]: string | number;
-  };
+  dataValues?: Record<string, string | number>;
```
kael89, line 53: 👍

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
```diff
@@ -0,0 +1,91 @@
...
+var dbm;
+var type;
+var seed;
+
+const permissionGroupNameToId = async (db, name) => {
+  const record = await db.runSql(`SELECT id FROM permission_group WHERE name = '${name}'`);
+  return record.rows[0] && record.rows[0].id;
+};
+
+const REPORT_CODE = 'PSSS_CONFIRMED_WEEKLY_REPORT';
```
kael89, line 14: Perhaps we want to use "first letter upper" snake case for report codes? `PSSS_Confirmed_Weekly_Report`, similar to existing `dashboardReport` ids

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
```diff
@@ -0,0 +1,91 @@
...
+          'PSSS_Confirmed_PF_Alert_Threshold_Crossed',
+          'PSSS_Confirmed_DLI_Cases',
+          'PSSS_Confirmed_DLI_Alert_Threshold_Crossed',
+        ],
+      },
+      transform: [
+        'keyValueByDataElementName',
+        {
+          transform: 'select',
+          "'period'": "convertToPeriod($row.period, 'WEEK')",
```
kael89, line 51: Does this need to be wrapped in single quotes? Or would the following also work?

```suggestion
          "period": "convertToPeriod($row.period, 'WEEK')",
```

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
```diff
@@ -0,0 +1,91 @@
...
+          '...': '*',
```
kael89, line 52: I can see this happening a lot, so we could consider making it the default behaviour (no harm keeping additional fields, even if they are eventually not needed)

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
```diff
@@ -0,0 +1,91 @@
...
+        {
+          transform: 'aggregate',
+          period: 'group',
+          organisationUnit: 'group',
+          '...': 'last',
+        },
+        {
+          transform: 'select',
+          "'AFR'": '$row.PSSS_Confirmed_AFR_Cases',
+          "'AFR Threshold Crossed'": '$row.PSSS_Confirmed_AFR_Alert_Threshold_Crossed == 1',
```
kael89, line 63: I'm not sure if our front-end expects a boolean value here, or `1`/`0`, or `Yes`/`No`

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
```diff
@@ -0,0 +1,91 @@
...
+          'PSSS_Confirmed_ILI_Cases',
+          'PSSS_Confirmed_ILI_Alert_Threshold_Crossed',
+          'PSSS_Confirmed_PF_Cases',
+          'PSSS_Confirmed_PF_Alert_Threshold_Crossed',
+          'PSSS_Confirmed_DLI_Cases',
+          'PSSS_Confirmed_DLI_Alert_Threshold_Crossed',
+        ],
+      },
+      transform: [
+        'keyValueByDataElementName',
```
kael89, line 48: I love this succinct transformation!

kael89 (changes requested):
Excellent work!! I love the new report config, it's very declarative!

So good to see all this work coming together!! 👏🚀

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
rohan-bes: So, the wrapping of keys in the `select` transform in quotes is because these keys are actually run through the `expression-parser` in order to evaluate their value. Doing this allows for some useful tricks like being able to key one value in a row by another value, eg. `"$row.dataElement" : "$row.value"`

However, this is obviously not the common case, so really we should have some logic in place to know whether or not we want to run the key through the `expression-parser`. I just haven't implemented that as of yet 😝 

I would like it do it soon tho, and if you don't wanna create any reports until we've got this done I can understand.

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
rohan-bes: As discussed since this is for `PSSS` we can pass through a boolean value 👍

packages/report-server/src/types.ts
rohan-bes: Yeah I think both endpoints would use the same `ReportBuilder` class, just only one has the ability to add the test config / test data. 

The idea around splitting into separate endpoints is partially cos the decoupling would be nice, but also a security thing. Regular users should be able to fetch data, but only Admin users (or some other group) should be able to test different configs out

packages/report-server/src/reportBuilder/reportBuilder.ts
```diff
@@ -9,22 +9,33 @@ interface BuildReport {
...
 
-  constructor(report, aggregator: Aggregator, query: FetchReportQuery) {
-    this.report = report;
+  readonly testData: Row[];
+
+  constructor(
+    config: ReportConfig,
+    aggregator: Aggregator,
+    query: FetchReportQuery,
+    testData: Row[] = [],
```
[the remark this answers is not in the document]
rohan-bes: Yeah I liked this idea, so I gave it a shot in the latest commit.

packages/report-server/src/types.ts
kael89: Most definitely!

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
```diff
@@ -0,0 +1,82 @@
...
+        'convertPeriodToWeek',
+        'aggregateFirstValuePerPeriodPerOrgUnit',
```
rohan-bes, line 50: I went with this `aggregateFirstValuePerPeriodPerOrgUnit` as there really only should be 1 value per dataElement per week. This is part of the business logic of PSSS. It's possible to use a `mostRecent` type aggregation here but those really ought to be proceeded with a `sort` which is a bit tricky to do here as we want to transform the period type, and also is a little inefficient.

Given that it's a case that's not meant to happen, I'm happy with going with `first` here (a bit like when you know there should just be a single matching value in an array you use `array.find()`) 👍

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
kael89: No worries, it's not big deal! It's great that we have a plan for this

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
kael89: Sounds good to me! How about naming it `keepFirstValuePerPeriodPerOrgUnit` or `pickFirstValuePerPeriodPerOrgUnit`

packages/report-server/src/reportBuilder/reportBuilder.ts
```diff
@@ -9,33 +9,25 @@ interface BuildReport {
...
+
+  setTestData = (testData: Row[]) => {
     this.testData = testData;
-  }
+  };
 
-  build = async (): Promise<BuildReport> => {
+  build = async (aggregator: Aggregator, query: FetchReportQuery): Promise<BuildReport> => {
+    if (!this.config) {
+      throw new Error('Report requires a config be set');
```
kael89, line 26: Nice!

packages/report-server/src/reportBuilder/transform/aliases/aggregateAliases.ts
```diff
@@ -12,3 +12,13 @@ export const aggregateMostRecentValuePerOrgUnit = () =>
...
+
+export const aggregateFirstValuePerPeriodPerOrgUnit = () =>
+  buildTransform([
+    {
+      transform: 'aggregate',
+      organisationUnit: 'group',
+      period: 'group',
+      '...': 'first',
+    },
+  ]);
```
kael89, line 24: So easy to add! That's a sign of a great architecture under the hood 😉

packages/report-server/src/routes/fetchReport.ts
```diff
@@ -19,51 +19,66 @@ class FetchReportRouteHandler {
...
+  models,
+  accessPolicy,
+  report,
+  organisationUnitCodes: string[],
+) => {
+  const permissionGroup = await models.permissionGroup.findById(report.permission_group_id);
+  const orgUnitsAndCodes = await Promise.all(
+    organisationUnitCodes.map(async orgUnitCode => ({
+      orgUnitCode,
+      orgUnit: await models.entity.findOne({ code: orgUnitCode }),
```
kael89, line 64: Could we change this to be a single query, as discussed in the past? (Can do in a separate PR)

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -54,6 +54,12 @@ const drop = (existingRow: Row, field: string, value: FieldValue) => {
   // Do nothing, don't add the field to the existing row
 };
 
+const first = (existingRow: Row, field: string, value: FieldValue) => {
+  if (!existingRow[field]) {
+    existingRow[field] = value;
+  }
```
kael89, line 60: (value may be `0`)

```suggestion
  if (!(field in existingRow)) {
    existingRow[field] = value;
  }
```

kael89 (changes requested):
I love the new report config, so simple! Almost there

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
rohan-bes: I was prefixing the names of all these aliases for `aggregate` transformations with `aggregate` as a bit of a hint to the user that 'hey, this is done using aggregation just so ya know` 😛 

But I do agree it's a bit awkward 👍 

How would you feel about just simply: `firstValuePerPeriodPerOrgUnit`? Then we could also have `sumValuesPerPeriod`, or `minValuePerOrgUnit`, etc. ? Or should it really have a verb in front for it to read like a transformation..?

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
rohan-bes: 🤦 why do I still always do this!! Thanks, great catch!!

kael89: I've been burned by that countless times...

packages/database/src/migrations/20201123061122-AddConfirmedWeeklyDataReportForPSSS-modifies-data.js
kael89: I really like `firstValuePerPeriodPerOrgUnit`, reads very natural!

packages/report-server/src/routes/fetchReport.ts
rohan-bes: Yup, good point, did this 👍

packages/report-server/src/routes/fetchReport.ts
```diff
@@ -55,25 +55,21 @@ const checkUserHasAccessToReport = async (
...
+  const foundOrgUnits = await models.entity.find({ code: requestedOrgUnitCodes });
+  const foundOrgUnitCodes = foundOrgUnits.map(orgUnit => orgUnit.code);
 
-  const countryCodes = new Set(
-    orgUnitsAndCodes.map(orgUnitAndCode => orgUnitAndCode.orgUnit.country_code),
+  const missingOrgUnitCodes = requestedOrgUnitCodes.filter(
+    orgUnitCode => !foundOrgUnitCodes.includes(orgUnitCode),
   );
+  if (missingOrgUnitCodes.length > 0) {
+    throw new Error(`No entities found with codes ${missingOrgUnitCodes}`);
```
kael89, line 69: Great implementation!

kael89 (approved):
Looks great! A historical moment, the first revamped `report` is ready to be merged! 🚀

PR #1698, opened 2020-11-25
1651 connect to meditrak server, 1654 maintain auth tokens

edmofro:
### Issue #:
- https://github.com/beyondessential/tupaia-backlog/issues/1651
- https://github.com/beyondessential/tupaia-backlog/issues/1654


packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { ApiConnection } from './ApiConnection';
+
+const { MEDITRAK_API_URL = 'http://localhost:8090/v2' } = process.env;
+
+export class MeditrakConnection extends ApiConnection {
+  baseUrl = MEDITRAK_API_URL;
+}
```
kael89, line 12: Although this pattern looks cool, I'm not sure if we need this and `ReportConnection` as separate classes: we could just pass `baseUrl` in the constructor.

Unless if we want to add custom behaviour in these classes in the very near future

packages/psss-server/src/connections/ApiConnection.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+import { fetchWithTimeout, checkStatusAndParseResponse, stringifyQuery } from '@tupaia/utils';
+import { FetchConfig, QueryParameters, RequestBody } from '../types';
+
+interface AuthHandler {
+  getAuthHeader: () => Promise<string>;
+}
+
+export class ApiConnection {
```
kael89, line 13: This class is awesome!! It's so good that we may find ourselves moving it in `utils` (`utils-ts`?) at some stage (not neccessary now)

packages/psss-server/src/connections/AuthConnection.ts
```diff
@@ -0,0 +1,59 @@
...
+const {
+  MEDITRAK_API_CLIENT_NAME,
+  MEDITRAK_API_CLIENT_PASSWORD,
+  MEDITRAK_API_URL = 'http://localhost:8090/v2',
+} = process.env;
+
+const basicAuthHandler = {
+  getAuthHeader: async () => {
+    const credentials = `${MEDITRAK_API_CLIENT_NAME}:${MEDITRAK_API_CLIENT_PASSWORD}`;
+    return `Basic ${Buffer.from(credentials).toString('base64')}`;
```
kael89, line 18: Detail, but this can also be a constant in the outer scope

packages/utils/src/errors.js
```diff
@@ -12,8 +12,9 @@ import { respond } from './respond';
  * Logged errors print out to the server's logs so that we have a record of all errors. In future
  * this may change to saving the error info to the database, notifying the admin, or similar
  */
-class LoggedError {
+class LoggedError extends Error {
```
kael89, line 15: 👌

packages/psss-server/src/routes/LogoutRoute.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { Route } from './Route';
+
+export class LogoutRoute extends Route {
+  async buildResponse() {
+    await this.sessionModel.deleteById(this.sessionCookie.id);
+    this.sessionCookie?.reset();
```
kael89, line 11: I'm not sure if this will work if `reset` is not defined, since it will try to invoke `undefined` as a function?

(I realise the existing code was like that, but it may be still worth checking)

packages/psss-server/src/app/addRoutesToApp.ts
```diff
@@ -6,15 +6,15 @@
...
+import { Route } from '../routes/Route';
 
-const handleWith = (Handler: typeof RouteHandler) => (
+const handleWith = (RouteClass: typeof Route) => (
   req: PsssRequest,
   res: Response,
   next: NextFunction,
 ) => {
-  const handler = new Handler(req, res, next);
+  const handler = new RouteClass(req, res, next);
```
kael89, line 17:
```suggestion
  const route = new RouteClass(req, res, next);
```

packages/admin-panel/src/api/TupaiaApi.js
```diff
@@ -139,25 +139,7 @@ export class TupaiaApi {
...
-        !responseJson.error
-      ) {
-        throw new Error(responseJson.message);
-      }
-      if (responseJson.error) {
-        throw new Error(responseJson.error);
-      }
-    }
-    return response;
+    return checkStatusAndParseResponse(response);
```
kael89, line 142: Nice!

packages/psss-server/src/types.ts
```diff
@@ -7,13 +7,15 @@ import { Request } from 'express';
 import { ParamsDictionary, Query } from 'express-serve-static-core';
 import { PsssSessionModel, PsssSessionType } from './models';
 
+export type AccessPolicyObject = Record<string, Array<string>>;
```
kael89, line 10: (Both are correct)

```suggestion
export type AccessPolicyObject = Record<string, string[]>;
```

packages/psss-server/src/types.ts
```diff
@@ -22,6 +24,53 @@ export interface PsssRequest<>extends Request<ParamsDictionary, unknown, PsssReq
...
+}
+
+export interface SessionDetails {
+  email: string;
+  accessPolicy: AccessPolicyObject;
+  accessToken: string;
+  refreshToken: string;
+}
+
+export interface SessionColumns {
```
kael89, line 69: We have named a similar type in indicators `IndicatorFields`, may be a good idea to rename to `SessionFields`

packages/psss-server/src/models/PsssSession.ts
```diff
@@ -4,19 +4,107 @@
...
 import { AccessPolicy } from '@tupaia/access-policy';
 
+import { AccessPolicyObject, SessionColumns, SessionDetails } from '../types';
+import { AuthConnection } from '../connections/AuthConnection';
+
+const getTokenExpiry = (accessToken: string) => {
+  const { exp: expiryAuthServerClock, iat: issuedAtAuthServerClock } = jwt.decode(accessToken);
+  // subtract 3 seconds to account for latency since generation on the auth server
+  const validForSeconds = expiryAuthServerClock - issuedAtAuthServerClock - 3;
+  const expiryPsssServerClock = Date.now() + validForSeconds * 1000;
```
kael89, line 18: I have a question about that but I'll ask offline

packages/psss-server/src/models/PsssSession.ts
```diff
@@ -4,19 +4,107 @@
...
+  authConnection: AuthConnection;
+
+  refreshAccessTokenPromise: Promise<void> | null;
+
   static databaseType = 'psss_session';
 
+  constructor(model: PsssSessionModel, fieldValues: SessionColumns) {
+    super(model, fieldValues);
+
+    // explicitly reassign all field values to satisfy typescript
```
kael89, line 44: It would be good to go around this at some point; but it's fine for now

packages/psss-server/src/models/PsssSession.ts
```diff
@@ -4,19 +4,107 @@
...
+      this.refreshAccessTokenPromise = refreshAndUpdate();
+    }
+    return this.refreshAccessTokenPromise;
+  }
+
+  async updateSessionDetails({ accessToken, accessPolicy }: SessionDetails) {
+    this.access_token = accessToken;
+    this.access_token_expiry = getTokenExpiry(accessToken);
+    this.access_policy = accessPolicy;
+    return this.save(); // update the underlying db record
```
kael89, line 89: Optional suggestion

```suggestion
    return this.save();
```

kael89 (changes requested):
Awesome work @edmofro !! It was a pleasure reading this PR, I have the suspicion that TS may actually make reviewing easier (🤞)

 All my suggestions are about details

packages/psss-server/src/connections/MeditrakConnection.ts
edmofro: Yep good point, but indeed the plan is that each will be filled out with some more custom behaviour as we implement various psss endpoints that need to connect. If they're still empty in two weeks, I agree we should change the pattern!

packages/psss-server/src/connections/AuthConnection.ts
edmofro: True. I'm not sure I have a preference either way, so I've swapped it to your suggestion

packages/utils/src/errors.js
edmofro: Thanks TypeScript!

packages/psss-server/src/routes/LogoutRoute.ts
edmofro: Yeah this is one of the ts errors I gave up on. I know that at the point `buildResponse` is called that `sessionCookie` will be defined on the class, but that's just because I know the logic of the parent class. It doesn't get defined in the constructor, so I can't make it mandatory...thought I would leave it for later!

packages/psss-server/src/app/addRoutesToApp.ts
edmofro: Good spot!

packages/psss-server/src/types.ts
edmofro: Aaah thank you, this is prettier

packages/psss-server/src/types.ts
edmofro: Very happy to!

packages/psss-server/src/models/PsssSession.ts
edmofro: Darn I never got back to you on this...

packages/psss-server/src/models/PsssSession.ts
edmofro: Totally. I wish we didn't need to, but I also understand that TypeScript can't just come up with types out of thin air.

packages/psss-server/src/models/PsssSession.ts
edmofro: Fair. It's pretty obvious.

edmofro:
@kael89 have responded to all your comments except the one you wanted to talk about in person

edmofro:
@kael89 I'm going to merge this into the base branch to check that tests pass when merged together. If you have any additional comments we can address them there

packages/psss-server/src/models/PsssSession.ts
kael89: No worries, it seems like you know what you're doing here - it would be mainly for my understanding. Can do post merge

packages/psss-server/src/routes/LogoutRoute.ts
kael89: Good call!

kael89 (commented):
Looks good post merge!

PR #1700, opened 2020-11-26
1338, 1339: Add PSSS Alert Threshold indicators

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1339
https://github.com/beyondessential/tupaia-backlog/issues/1340

#### Dependencies (pending merge)
- [x] https://github.com/beyondessential/tupaia/pull/1677
- [x] https://github.com/beyondessential/tupaia/pull/1684

### DIFF
--- a/packages/database/src/migrations/20201125021221-AddPsssAlertThresholdIndicators-modifies-data.js
+++ b/packages/database/src/migrations/20201125021221-AddPsssAlertThresholdIndicators-modifies-data.js
@@ -1,0 +1,125 @@
+'use strict';
+
+import { arrayToDbString, generateId, insertObject } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const CONDITION_CODES = ['AFR', 'DIA', 'ILI', 'PF', 'DLI'];
+
+// eg 'PSSS_Confirmed_AFR_Site_Average'
+const psssCode = (descriptor, condition = '', isConfirmed = false) => {
+  const parts = ['PSSS', descriptor];
+  if (condition) {
+    parts.splice(1, 0, condition);
+  }
+  if (isConfirmed) {
+    parts.splice(1, 0, 'Confirmed');
+  }
+
+  return parts.join('_');
+};
+
+const insertIndicator = async (db, indicator) => {
+  await insertObject(db, 'indicator', { ...indicator, id: generateId() });
+  await insertObject(db, 'data_source', {
+    id: generateId(),
+    code: indicator.code,
+    type: 'dataElement',
+    service_type: 'indicator',
+  });
+};
+
+const createAlertThresholdLevelIndicator = async (db, condition, isConfirmed) => {
+  if (condition === 'AFR') {
+    // The threshold level for AFR is always 1
+    return true;
+  }
+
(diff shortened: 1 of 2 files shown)


tupaia-packages.code-workspace
```diff
@@ -95,6 +95,7 @@
       "Codeship",
       "DHIS",
       "MediTrak",
+      "PSSS",
```
rohan-bes, line 98: Just a warning, I think @edmofro has also done this in: https://github.com/beyondessential/tupaia/pull/1686/files#diff-65619f8f662c566b8a93965b22b0b01a7aa4668821d23dd9e727b975ed5f6291R102

packages/database/src/migrations/20201125021221-AddPsssAlertThresholdIndicators-modifies-data.js
```diff
@@ -0,0 +1,125 @@
...
+  if (condition === 'DLI') {
+    averageFormula = '(siteAverage1WeekAgo + siteAverage2WeeksAgo + siteAverage3WeeksAgo) / 3';
+    parameters.push(createParameter('siteAverage3WeeksAgo', 3));
+  }
+
+  return insertIndicator(db, {
+    code: psssCode('Alert_Threshold_Level', condition, isConfirmed),
+    builder: 'arithmetic',
+    config: {
+      formula: `2 * (${averageFormula})`,
```
rohan-bes, line 77: While I'm largely not fussed, I am a little curious to know if `mathjs` knows how to optimise this? Given that for `PF`, `DIA`, and `ILI` this formula is `2 * (siteAverage1WeekAgo + siteAverage2WeeksAgo) / 2` => `siteAverage1WeekAgo + siteAverage2WeeksAgo`?

I wonder if there would be any performance hit for an sub-optimised (mathematically) expression or not...

But yeah, not a real issue 👍

rohan-bes (approved):
Looks awesome! Great work 😄 

Just the one comment about the clash with Edwin's PR, but approving anyway

tupaia-packages.code-workspace
kael89: Thanks for flagging this! Hopefully the conflict would be resolved automatically, but if not it should be easy to resolve it

packages/database/src/migrations/20201125021221-AddPsssAlertThresholdIndicators-modifies-data.js
kael89: That's a good observation! While scanning the math.js docs, I bumped into the [simplify](https://mathjs.org/examples/algebra.js.html) API:

```js
console.log(simplify('3 + 2 / 4').toString()) // '7 / 2'
console.log(simplify('2x + 3x').toString()) // '5 * x'
```

We could simplify operations in the `arithmetic` indicator level, but I don't know if simplify + evaluate would be actually more performant than just evaluate... 

For this specific case, I really hope that 2 * x / 2 is performant enough, cause if not we are into serious trouble. But I'd bet it will be very, very fast

rohan-bes: Yeah I imagine `simply` then `evaluate` is probably more expensive than just running `evaluate`. Perhaps when we allow for creating indicators via the Admin panel, we can run `simplify` over the expression before saving it to the db? 

But yeah, can't imagine a lot of scenarios where it would create noticeable performance issues, and sometimes an _unsimplified_ expression reads a little more clearly to a human so we wouldn't want to lose that I guess 😄

kael89: Exactly, I thought of "simplifying" the query but then I decided not to, so that the calculation logic is more obvious to humans

PR #1701, opened 2020-11-26
Fix getDateRangeForOffsetPeriod

kael89:
@rohan-bes adding you as a reviewer as you reviewed the original PR

### DIFF
--- a/packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/offsetPeriod.test.js
+++ b/packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/offsetPeriod.test.js
@@ -4,160 +4,208 @@
  */
 
 import { arrayToAnalytics } from '@tupaia/data-broker';
-import { offsetPeriod } from '../../../../analytics/aggregateAnalytics/aggregations/offsetPeriod';
+import {
+  getDateRangeForOffsetPeriod,
+  offsetPeriod,
+} from '../../../../analytics/aggregateAnalytics/aggregations/offsetPeriod';
 
-describe('offsetPeriod()', () => {
-  it('throws an error if `periodType` is not provided', () => {
-    expect(() => offsetPeriod([], { offset: 1 })).toThrow("'periodType' is required");
-  });
+describe('offsetPeriod', () => {
+  const testConfigValidation = () => {
+    it('throws an error if `periodType` is not provided', () => {
+      expect(() => offsetPeriod([], { offset: 1 })).toThrow("'periodType' is required");
+    });
 
-  it('throws an error if `offset` is not provided', () => {
-    expect(() => offsetPeriod([], { periodType: 'day' })).toThrow("'offset' is required");
-  });
+    it('throws an error if `offset` is not provided', () => {
+      expect(() => offsetPeriod([], { periodType: 'day' })).toThrow("'offset' is required");
+    });
+  };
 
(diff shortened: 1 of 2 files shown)


packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/offsetPeriod.test.js
```diff
@@ -4,160 +4,208 @@
...
+    });
+  };
 
-  it('`periodType` is case insensitive', () => {
-    const analytics = [
-      { dataElement: 'BCD1', organisationUnit: 'TO', period: '20190101', value: 1 },
-    ];
-    const offset = 1;
+  describe('offsetPeriod()', () => {
+    testConfigValidation();
```
kael89, line 24: The only changes here are

1. I created and used the `testConfigValidation` function
2. I wrapped this block in an outer `offsetPeriod` block

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/offsetPeriod.test.js
```diff
@@ -4,160 +4,208 @@
...
+        const periodType = 'month';
+        // This is an edge case where applying a '-' offset is not a reversible with a '+' offset
+        assertPeriodOffsetIsApplied('20190531', { periodType, offset: -1 }, '20190430');
+        assertPeriodOffsetIsApplied('20190430', { periodType, offset: +1 }, '20190530');
+      });
     });
+  });
+
+  describe('getDateRangeForOffsetPeriod()', () => {
+    testConfigValidation();
```
kael89, line 174: Everything is new from here on

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/offsetPeriod.js
```diff
@@ -47,11 +47,8 @@ const compensateForPeriodOffset = (date, { periodType, offset }) =>
...
 export const getDateRangeForOffsetPeriod = (dateRange, config) => {
   validateOffsetPeriodConfig(config);
-  const { offset } = config;
-  const { startDate, endDate } = dateRange;
-
   return {
-    startDate: offset > 0 ? compensateForPeriodOffset(startDate, config) : startDate,
-    endDate: offset < 0 ? compensateForPeriodOffset(endDate, config) : endDate,
+    startDate: compensateForPeriodOffset(dateRange.startDate, config),
+    endDate: compensateForPeriodOffset(dateRange.endDate, config),
```
kael89, line 52: I realised we should add offset compensation to both dates: if not, we may get more analytics than we want, which doesn't play nice with "most_recent" logic

Eg, for
```js
{
  startDate: '20201123',
  endDate: '20201130',
  aggregation: { type: 'OFFSET_PERIOD', config: { periodType: 'week', offset: 1 }}
}
```

which is equivalent to "fetch my the results for the previous period", we want to get results in `['20201116', '20201123']`, then add 1 week to their period

rohan-bes: Ahhh this makes total sense, good pickup 👍

rohan-bes (approved):
Looks good, nice fix

kael89:
This aggregation type is not currently used in `dev`, so no testing would be required - merging

PR #1702, opened 2020-11-26
Add country permissions back into BCD survey

nick-beyondessential:
https://github.com/beyondessential/tupaia-backlog/issues/886 Papua New Guinea broken raw data download

Raw Data Downloads for 'Basic Clinic Data' and 'Childhood Diarrhoea' broken for Papua New Guinea.
This is because country permissions were overwritten on import (address with https://github.com/beyondessential/tupaia-backlog/issues/1042 )
Fiji and Samoa also added back in.

Change Raw Data Downloads survey 'CD' to 'DIA'
Add PG to DIA survey


packages/database/src/migrations/20201126030925-AddPngRawDataDownloadSurveyPermissions-modifies-data.js
```diff
@@ -0,0 +1,44 @@
...
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const SURVEY_CODE = 'BCD';
+const COUNTRY_CODES = ['FJ', 'PG', 'PH', 'LA', 'WS'];
```
edmofro, line 18: Where did you get the list of countries that should have access? I can see a BCD survey for each of Fiji and Samoa in the admin panel, so do they need access to the generic one?

edmofro (changes requested):
The migration code all looks good, I just have a question on the specific countries used.

This should all also be admin panel tasks rather than migrations - could you add a card to allow editing the list of countries attached to each survey through the admin panel (currently only possible through a reimport)

packages/database/src/migrations/20201126030925-AddPngRawDataDownloadSurveyPermissions-modifies-data.js
nick-beyondessential: I got the list of countries from 
  select * from "dashboardGroup" where 'Raw_Data_Core_Surveys' = any("dashboardReports"); 
Which yielded FJ', 'PG', 'PH', 'WS'  and  'LA' was the country_id that was already listed in that survey (along with PH).

nick-beyondessential: https://github.com/beyondessential/tupaia-backlog/issues/1678

edmofro: That's a fair way to try to work it out, but I'm afraid doesn't get us the true list we need. Survey permissions don't generally relate directly to dashboard report permissions. For example, it seems like `FM` should have access to BCD (well, at least, they've submitted some before) but aren't in that list.

I also don't really understand why Fiji has access to download the generic BCD when they have their own custom BCDF (same goes for Samoa). 

It seems like a broader data stewardship issue is at play here - which BCD should each country have access to. I believe that each one should either have access to a country specific one, or the generic one.

A couple of semi-informative queries:
```
-- all BCD surveys (a shame that not all have been coded with the ISO country code suffix, would be good to standardise that)
select code, name, country_ids from survey where code like 'BCD%'; 

-- a map of which countries have filled out which BCD surveys (one from PG for the BCD_VU survey!?)
select country_code, survey.code from survey_response join survey on survey_id = survey.id join entity on entity_id = entity.id where survey.code like 'BCD%' group by country_code, survey.code;`
```

nick-beyondessential: That is fair enough though Fiji does currently have BCD in their download report:
image
which is currently errorring from lack of permissions (I'm assuming it previously worked.) So should this be updated to BCDF (as a new card?)

nick-beyondessential: New list of countries based on use would be 'CI, FM, KI, LA, PG, PH, TO, VE, VU', With FJ and WS to have their data download reports updated to their specific BCD survey?

edmofro: Can you check that against the list of all countries, minus those with specific BCD surveys?

Yes, I think that each country with a specific BCD survey should have that in their raw data downloads, rather than the general one (unless they historically also have data under that, in which case maybe both)

These are really data steward calls though. @gerrybes when you're back next week could you take a look at what data is where for BCD surveys, standardise the survey codes, etc.?

nick-beyondessential: Ah I see so countries minus specific BCD surveys becomes (KH, MH, NR, NU, PW, TV, AS, GU, PF, MP, PI, WF, NO, NC, AU, NZ) + (CI, FM, KI, LA, PG, PH, TO, VE, VU); 

Maybe NO (No Country shouldn't be included?)

So Should we add FJ & WS to fix their currently broken reports and create a new card for @gerrybes standardisation?

edmofro: I'm confused about your list. What's the difference between the two sections? And why are TO, VU, VE in the list?

edmofro: Agree that NO shouldn't be included!

Please add all of the ones that should be able to collect data against BCD, which is basically all countries except those with their own specific BCD survey (could check that with Erin and/or Mike)

nick-beyondessential: First section is countries not in:
 `select country_code, survey.code 
from survey_response 
join survey on survey_id = survey.id join entity on entity_id = entity.id 
where survey.code like 'BCD%' group by country_code, survey.code;`

Second list are countries in the above survey query with BCD survey

edmofro: As for "Should we add FJ & WS to fix their currently broken reports" - there's not much point if they don't have any data. Much better would be change their raw data download to have the actual survey they fill out, rather than the general one

nick-beyondessential: As for the ones like Tonga that have both a specific and general download?

edmofro: Good question! I'm going to tap out and let @avaek make some of those calls - he's been around when a lot of these were transitioning from general to specific, and might have a better understanding of what to do here.

avaek: Happy to chime in on this - I think you guys have mostly ironed this out but one more thing to consider is that a lot of the Basic Clinic Data survey responses that exist are not "real" data & are basically just survey responses submitted by me or another data steward to give an answer to the `BCD1` question (in order to mark a facility as operational / closed / etc). In cases like this the data doesn't really need to be exported since it's more just for internal use

Below is my take on what should be exported by which countries when Basic Clinic Data is selected in the Raw Data Downloads tool:

_Countries that should just export regular BCD survey responses with this tool:_
- `PH` (note - all these survey responses were submitted by Michael but they seem to contain "real" data)
---

_Countries that should just export country specific BCD survey responses with this tool:_ 
- `CK`
- `KI`
- `SB`
- `TK`
- `TO`
- `VE` (note - these were all submitted by me but they seem to contain "real" data)
- `VU`
- `WS`
- You'll notice that some of these are countries you identified as having both regular BCD & country specific BCD survey responses. The reason I think only the country specific BCD survey responses should be exported is because these are the only survey responses with "real" data - the regular BCD survey responses for countries in this section were just submitted by me / data stewards to give an answer to `BCD1` for certain facilities in that country

---

_Countries that shouldn't export any Basic Clinic Data survey responses (even though they have survey responses for either):_ 
- `CI`
- `FM`
- `LA`
- `PG`
- `TL` (note - this country actually has real data but I believe we don't display this country on our site anymore so it can be ignored)
- These should be ignored because there is no "real" data for the entire country & is all just survey responses containing an answer just for `BCD1` that was submitted by me or another data steward

---

If my above suggestion would require significant code changes to implement then perhaps theres a way to transform this into some kind of data task to get this working the way that we want it to (e.g. reimporting surveys with certain codes etc)

nick-beyondessential: Amazingly clear and thorough description as usual, Thanks @avaek.

edmofro: Yes thanks for the illuminating comments and taking the time to explain thoroughly. 

That makes it clear which should be available in which raw data download. Could you also list which countries should have permission to fill in a BCD? Hint: a way to figure this out is to filter the `api_request_log` table by endpoints like `import/survey` with query params containing `BCD`, and see if there's historically been a consistent set of country ids (along with a bit of reasoning about which countries should really have access - potentially all without a specific BCD). The permission @nick-beyondessential is setting up is more related to filling in surveys than it is to viewing data (the latter is a side effect of raw data downloads routing through meditrak-server).

@nick-beyondessential, I also wonder if it would be possible to convert these raw data downloads to use your new and improved download system so that there is no routing via meditrak-server?

I guess then there are three independent tasks:
- reset `country_ids` on the BCD survey to whatever it was before
- split out raw data downloads to be more country specific, using the appropriate BCD survey in each case
- transfer raw data downloads from direct-from-meditrak-server to use the regular data fetching and table building procedure

avaek: Regarding searching the `api_request_log` table - I just took a quick look at the `import/surveys` endpoint & I couldn't find anything that referred specifically to the _Basic Clinic Data_ survey in the `query` column (the only results were related to country specific BCD surveys)

nick-beyondessential: Hi @enunan, we have had another case of survey country_ids being overwritten, and wondering if you are able offer any insight into which countries should have access to the BCD survey? @avaek has identified permissions for raw data downloads but this doesn't necessarily match up with which countries should have access to survey.

enunan: Hey, I think all countries should have access to BCD as this (and Facility Fundamentals) is what makes up our basic data available to the public about each facility. 
Some countries (e.g. Samoa) have made country specific edits to their BCD, but others should be quite uniformly available in all countries.

nick-beyondessential: Thanks @enunan, that makes the first task easier.

nick-beyondessential:
- I have updated to the country_id list to all countries as per https://github.com/beyondessential/tupaia/pull/1702#discussion_r533047070.
- Removed the CD to DIA migration as this has been completed as a data task.
- transfer raw data downloads from direct-from-meditrak-server to use the regular data fetching and table building procedure will be handled in https://github.com/beyondessential/tupaia-backlog/issues/1760

edmofro:
Hey @nick-beyondessential, thanks for the clear update!

I think Erin's message https://github.com/beyondessential/tupaia/pull/1702#discussion_r533047070 meant all countries should have access to _some_ BCD, so those with a specific BCD survey shouldn't also have access to the general BCD. It seems the list of country_ids now includes SB etc., but it shouldn't. Could you please update again?

nick-beyondessential:
Thanks @edmofro, that makes sense. The above country codes have been removed.

edmofro (approved):
Wahoo!

PR #1703, opened 2020-11-26
1659: Accept access token for authentication in report server

chris-bes:
### Changes:

- In authentication middleware of report server, take in Bearer Auth Header with an Access Token


packages/report-server/src/auth/middleware.ts
```diff
@@ -1,20 +1,30 @@
...
-    const { user, accessPolicy } = await authenticator.authenticatePassword({
-      emailAddress,
-      password,
-      deviceName: 'reports',
-    });
-    if (user) {
+    const userId = await authenticateUser(req);
+    if (userId) {
+      const { authenticator } = req;
+      const accessPolicy = await authenticator.getAccessPolicyForUser(userId);
```
chris-bes, line 26: Do we still want to support authenticate by username and password alongside with access token? I thought that since we're only connecting report-server up to psss-server for now it's fine to leave it accepting only access token (similar to meditrak-server). Otherwise we will have to add an extra query parameter `authType` and provide it for every request

scripts/bash/backendStartDev.sh
```diff
@@ -39,7 +39,7 @@ done
 
 # Start server command for TS
 if [[ ${type_script} == true ]]; then
-    start_server="nodemon -w src --exec \"babel-node --extensions \".ts\" src/index.ts --inspect=${inspect_port} --config-file '../../.babelrc-ts.js'\""
+    start_server="nodemon -w src -e ts,json --exec \"babel-node --extensions \".ts\" src/index.ts --inspect=${inspect_port} --config-file '../../.babelrc-ts.js'\""
```
chris-bes, line 42: For fixing an issue that start-dev psss-server does not hot reload when skipping internal dependencies

packages/auth/src/security.js
```diff
@@ -1,6 +1,6 @@
 /**
```
chris-bes, line 1: Moving this file to central auth package so that it can be used by both meditrak-server and report-server

packages/report-server/src/auth/middleware.ts
rohan-bes: You make a good argument for removing the user/pass auth

I do however find basic user/pass auth quite useful when testing the report-server. Otherwise we'd need to send an initial request to the /auth endpoints of the meditrak/psss server to get a valid auth token right?

This also shouldn't be too hard to set up using VSCode Rest Client, but if you're testing locally does still require you to have one of those servers up and running...

What do you rekon? I think I lean slightly towards supporting both...

scripts/bash/backendStartDev.sh
rohan-bes: Nice one!

rohan-bes (commented):
What are your thoughts on supporting both user/pass and token auth?

Happy with everything else tho 👍 

I think there's a messed up import or something in the meditrak-server tests tho? The build is failing?

chris-bes:
Not sure why I couldn't reply to your comment

> You make a good argument for removing the user/pass auth
> 
> I do however find basic user/pass auth quite useful when testing the report-server. Otherwise we'd need to send an initial request to the /auth endpoints of the meditrak/psss server to get a valid auth token right?
> 
> This also shouldn't be too hard to set up using VSCode Rest Client, but if you're testing locally does still require you to have one of those servers up and running...
> 
> What do you rekon? I think I lean slightly towards supporting both...

I think you raised a good point that if we only support accept token in report-server for authentication, then we have to have either psss-server or meditrak-server up running to test the report-server locally.

How about making access-token authentication as the default way. And if you want to authenticate by username/pass, provide a query parameter `authType=password`?

rohan-bes:
> How about making access-token authentication as the default way. And if you want to authenticate by username/pass, provide a query parameter `authType=password`?

This is the right solution! Perhaps instead of using an `authType` query param, we can just use the authorization header `Authorization: Basic`? And then include email/password in the request body?
I know that providing a base64 encoded user/pass in the auth header itself is standard for Basic auth, so if you feel like going that far I think that could be good too, but keeping them as plaintext in the body also seems fine to me...

chris-bes:
> I know that providing a base64 encoded user/pass in the auth header itself is standard for Basic auth, so if you feel like going that far I think that could be good too, but keeping them as plaintext in the body also seems fine to me...

That's a great suggestion, I went with sending username/password using Basic Auth Header instead of putting them in the body

packages/report-server/src/auth/middleware.ts
```diff
@@ -1,20 +1,56 @@
...
+  if (tokenUserID) {
+    return tokenUserID;
+  }
+
+  throw new UnauthenticatedError('Could not authenticate with the provided access token');
+}
+
+const authenticateBasicAuthHeader = async (req: ReportsRequest, authHeader: string) => {
+  const usernamePassword = Buffer.from(authHeader.split(' ')[1], 'base64').toString();
+  const [username, password] = usernamePassword.split(':');
```
rohan-bes, line 33: I'm a little worried that this may fail if the user has a `:` in their password? I think it's safe for us to assume that `:` can't be in the username since it's invalid in an email.

Perhaps something like: 
```
if (!usernamePassword.includes(':')) {
  throw new UnauthenticatedError('Invalid Basic auth credentials');
}
const username = usernamePassword.split(':')[0];
const password = usernamePassword.substring(username.length, usernamePassword.length);
```

Just double check me on that `substring` logic cos my small brain always screws up string indexes 😝

rohan-bes (changes requested):

packages/report-server/src/auth/middleware.ts
chris-bes: Oh man... nice catch 🤦

chris-bes: Fixed!

rohan-bes (approved):
Awesome work @chris-bes , this looks great!

PR #1705, opened 2020-11-26
Migration to update Raw Data Download reports survey codes CD to  DIA

nick-beyondessential:
### Changes: 
- Migration to update Raw Data Download reports survey code for "Childhood Diarrhoea" from 'CD' to "DIA"
- updated 'Raw_Data_CK_Surveys',
                  'Raw_Data_DL_Surveys',
                  'Raw_Data_VU_Surveys',
                  'Raw_Data_TL_Surveys',
                  'Raw_Data_KI_Surveys',
                  'Raw_Data_TK_Surveys',
                  'Raw_Data_SB_Surveys',
                  'Raw_Data_VE_Surveys',
                  'Raw_Data_TO_Surveys'

-Raw_Data_Core_Surveys is handled by https://github.com/beyondessential/tupaia/pull/1702

### DIFF
--- a/packages/database/src/migrations/20201126075832-ChangeRawDataDownloadSurveyCodeFromCDtoDIA-modifies-data.js
+++ b/packages/database/src/migrations/20201126075832-ChangeRawDataDownloadSurveyCodeFromCDtoDIA-modifies-data.js
@@ -1,0 +1,45 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const reportIds = `'Raw_Data_CK_Surveys',
+                  'Raw_Data_DL_Surveys',
+                  'Raw_Data_VU_Surveys',
+                  'Raw_Data_TL_Surveys',
+                  'Raw_Data_KI_Surveys',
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20201126075832-ChangeRawDataDownloadSurveyCodeFromCDtoDIA-modifies-data.js
```diff
@@ -0,0 +1,45 @@
...
+                  'Raw_Data_KI_Surveys',
+                  'Raw_Data_TK_Surveys',
+                  'Raw_Data_SB_Surveys',
+                  'Raw_Data_VE_Surveys',
+                  'Raw_Data_TO_Surveys'`;
+
+exports.up = function(db) {
+  return db.runSql(`
+    update "dashboardReport" 
+    set "dataBuilderConfig" = jsonb_set("dataBuilderConfig", '{surveys,2,code}', '"DIA"')
```
nick-beyondessential, line 30: Might have been preferable to have a fancier way of discovering index for each report but as this survey is listed in the same position for each report we can get away with supplying an index.

rohan-bes (commented):
While this looks totally fine to me, are we not able to now fix issues like this via the Admin Panel? That seems preferable to writing a migration, but perhaps there's something I'm missing?

nick-beyondessential:
@rohan-bes you are probably correct, this could be done via admin, migration just felt more native to me.
What would be the procedure regarding testing etc. Change on dev, get a tester to test there, then change on prod. Or Just chane on Prod and test there?

nick-beyondessential:
closing to edit via admin-panel

PR #1713, opened 2020-11-29
1652 psss server integration

tcaiger:
### Issue #: [1652 psss server integration](https://github.com/beyondessential/tupaia-backlog/issues/1652)

Integration with the new psss-server.

### Changes:

- Added axios for making requests
- Added api request methods as an abstraction layer
- Cleaned up store init
- Update psss-server to return full user object
- Update psss-server cors policy
- Moved useTableQuery into api directory
- move login flow from redux into component
- Add global handling of 401 errors and logout user

---


packages/psss-server/src/connections/AuthConnection.ts
```diff
@@ -52,6 +52,6 @@ export class AuthConnection extends ApiConnection {
       throw new Error('Invalid response from auth server');
     }
     const { accessPolicy, email } = user;
-    return { accessToken, refreshToken, accessPolicy, email };
+    return { accessToken, refreshToken, accessPolicy, email, user };
```
tcaiger, line 55: The front end needs the user object to display the users name (and less importantly the users other details for the profile section). It may be better to restructure the response rather than just adding in the extra object. If we do that then the 'types' need to be refactored too.

packages/psss-server/src/routes/LoginRoute.ts
```diff
@@ -16,9 +19,14 @@ export class LoginRoute extends UnauthenticatedRoute {
...
+    const authorized = accessPolicy.allowsAnywhere(PSSS_PERMISSION_GROUP);
+    if (!authorized) {
+      throw new PermissionsError('User not authorized for PSSS');
+    }
+
     // set sessionId cookie
     this.req.sessionCookie = { id, email };
 
-    return { accessPolicy };
+    return { user: response.user, accessPolicy };
```
tcaiger, line 30: The front end needs the user object. See comment above.

packages/psss/src/api/api.js
```diff
@@ -0,0 +1,53 @@
...
+
+  try {
+    const { data } = await axios(`${PSSS_API_URL}/${endpoint}`, {
+      headers,
+      timeout,
+      ...options,
+    });
+    return data;
+  } catch (error) {
+    if (error.response) {
```
tcaiger, line 33: I would expect data.error to always have the error message but I am also doing my best to always show a good error message. One thing I noticed is that all the errors from meditrak server come prepended with 'Internal server error'. For example 'Internal server error: Incorrect email or password'. I could parse it out but another option would be to not add it on the server?

packages/psss/src/api/useTableQuery.js
```diff
@@ -0,0 +1,26 @@
+/*
```
tcaiger, line 1: This file is un-changed - just moved from the hooks directory

packages/psss/src/index.js
```diff
@@ -4,14 +4,14 @@
...
-import App from './App';
 import { ReactQueryDevtools } from 'react-query-devtools';
+import App from './App';
 import { AppProviders } from './AppProviders';
-import { initStore } from './store/store';
+import { store } from './store/store';
 
 const render = () => {
   return renderReactApp(
-    <AppProviders store={initStore()}>
```
tcaiger, line 14: an api class is no longer injected into the store so no longer need to have this be a function

packages/psss/src/utils/PersistGateProvider.js
```diff
@@ -17,6 +18,7 @@ function initPersistor(store) {
 
     // If it is a new session and the user has not ticked rememberMe, then remove the user data from localStorage
     if (rememberMe === 'false') {
+      logoutUser();
```
tcaiger, line 21: hit the logout endpoint and clear session cookie

packages/psss-server/src/connections/AuthConnection.ts
```diff
@@ -24,11 +24,11 @@ export class AuthConnection extends ApiConnection {
     super(basicAuthHandler);
   }
 
-  async login({ emailAddress, password }: Credentials) {
+  async login({ emailAddress, password, deviceName }: Credentials) {
     const response = await this.post(
       'auth',
       { grantType: 'password' },
-      { emailAddress, password, deviceName: 'PSSS Server' },
+      { emailAddress, password, deviceName: `PSSS-SERVER: ${deviceName}` },
```
tcaiger, line 31: I assumed that we still want the deviceName captured but let me know if not and i can remove it.

packages/psss-server/src/routes/LogoutRoute.ts
```diff
@@ -3,12 +3,20 @@
...
 
-import { Route } from './Route';
+import { UnauthenticatedRoute } from './UnauthenticatedRoute';
 
-export class LogoutRoute extends Route {
+export class LogoutRoute extends UnauthenticatedRoute {
   async buildResponse() {
-    await this.sessionModel.deleteById(this.sessionCookie.id);
-    this.sessionCookie?.reset();
+    const sessionId = this.sessionCookie?.id;
```
tcaiger, line 10: I thought that it would be better if the logout route didn't return an error if it was called without a sessionCookie in the request (which means it had to be unauthenticated) as it could lead to a few issues including potentially getting stuck in a 401 loop.

packages/psss/src/store/auth.js
```diff
@@ -5,69 +5,48 @@
...
-    const userDetails = await api.reauthenticate({
-      emailAddress: email,
-      password,
-      deviceName,
-    });
-    dispatch(loginSuccess(userDetails));
-  } catch (error) {
-    dispatch(loginError(error.message));
-  }
+  const user = await loginUser({
```
edmofro, line 19: How does the error handling work here?

edmofro: (or in the others below)

edmofro (changes requested):
Lots of simplification here, big wins!

packages/psss/src/api/requests/auth.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const getUser = () => get('me');
+
+export const updateUser = data =>
+  put('me', {
+    data,
+  });
+
+export const updatePassword = data => post('me/changePassword', { data });
```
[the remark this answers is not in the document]
tcaiger: yes good idea!

packages/psss/src/store/auth.js
tcaiger: The error will bubble up and get handled in the try catch of the component (eg. LoginFormComponent). This does create a risk that if someone in the future dispatched login without a try catch in a component then there would be a risk of an uncaught exception. If the errors were handled in the store code then, the same risk would just exist in a different place.

tcaiger: I did a little bit of extra research around handling unchaught exceptions. The good news is that for all the queries that use react-query (ie. all but the 3 in this file). Then there is no risk because, react-query handles the thrown error and simply returns the error object with the query response.

edmofro (approved):

PR #1715, opened 2020-11-30
Check out latest code before deploying packages

edmofro:
Issue: https://github.com/beyondessential/tupaia-backlog/issues/1694

### DIFF
--- a/packages/devops/scripts/deployment/startup.sh
+++ b/packages/devops/scripts/deployment/startup.sh
@@ -13,9 +13,9 @@ echo "Starting up instance for ${STAGE}"
 
 # Set the branch based on STAGE
 if [[ $STAGE == "production" ]]; then
-    BRANCH="master"
+    export BRANCH="master"
 else
-    BRANCH="$STAGE"
+    export BRANCH="$STAGE"
 fi
 
 # Fetch the latest code
--- a/packages/devops/scripts/deployment/startup.sh
+++ b/packages/devops/scripts/deployment/startup.sh
@@ -11,9 +11,19 @@ DIR=$(dirname "$0")
 export STAGE=$(${DIR}/../utility/getEC2TagValue.sh Stage)
 echo "Starting up instance for ${STAGE}"
 
-# Fetch and deploy the latest for each package based on the stage, including injecting environment
-# variables from parameter store into the .env file
-${HOME_DIRECTORY}/packages/devops/scripts/deployment/deployLatestRepositories.sh
+# Set the branch based on STAGE
+if [[ $STAGE == "production" ]]; then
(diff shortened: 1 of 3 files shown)


packages/devops/scripts/deployment/startup.sh
```diff
@@ -11,9 +11,19 @@ DIR=$(dirname "$0")
...
+else
+    BRANCH="$STAGE"
+fi
+
+# Fetch the latest code
+${HOME_DIRECTORY}/packages/devops/scripts/deployment/checkoutLatest.sh
+
+# Deploy each package based on the stage, including injecting environment variables from parameter
+# store into the .env file
+${HOME_DIRECTORY}/packages/devops/scripts/deployment/deployPackages.sh
```
edmofro, line 26: This runs the version of `deployPackages.sh` on the newly checked out branch, rather than whatever was on master during the overnight clone.

packages/devops/scripts/deployment/startup.sh
```diff
@@ -11,9 +11,19 @@ DIR=$(dirname "$0")
...
-# Fetch and deploy the latest for each package based on the stage, including injecting environment
-# variables from parameter store into the .env file
-${HOME_DIRECTORY}/packages/devops/scripts/deployment/deployLatestRepositories.sh
+# Set the branch based on STAGE
+if [[ $STAGE == "production" ]]; then
+    BRANCH="master"
+else
+    BRANCH="$STAGE"
+fi
+
```
kael89, line 20: I believe this is not required in this script

```suggestion
```

kael89 (changes requested):
Very valuable change! Just a suggestion

packages/devops/scripts/deployment/startup.sh
edmofro: Oh oops - actually it is required here, but not required in `checkoutLatest.sh`. Will update - good catch!

packages/devops/scripts/deployment/startup.sh
```diff
@@ -11,9 +11,19 @@ DIR=$(dirname "$0")
...
 
-# Fetch and deploy the latest for each package based on the stage, including injecting environment
-# variables from parameter store into the .env file
-${HOME_DIRECTORY}/packages/devops/scripts/deployment/deployLatestRepositories.sh
+# Set the branch based on STAGE
+if [[ $STAGE == "production" ]]; then
+    export BRANCH="master"
+else
+    export BRANCH="$STAGE"
+fi
```
kael89, line 19: Ideally we'd pass those in as an argument in the scripts invoked below

kael89 (approved):
Approved with a note

PR #1716, opened 2020-11-30
1367 1370 PSSS Confirmed weekly reports

chris-bes:
### Issue:
https://github.com/beyondessential/tupaia-backlog/issues/1367
https://github.com/beyondessential/tupaia-backlog/issues/1370

### Changes:
- Added PSSS endpoints for Confirmed weekly reports
- /confirmedWeeklyReport?startWeek=2020W1&endWeek=2020W1
- /confirmedWeeklyReport/:orgUnitCode?startWeek=2020W1&endWeek=2020W8


packages/database/src/migrations/20201129051349-DropPSSSSessionNotifyChangesTrigger-modifies-schema.js
```diff
@@ -0,0 +1,27 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql('DROP TRIGGER IF EXISTS psss_session_trigger ON psss_session;');
```
chris-bes, line 18: When `psss_session` table was created, a change notify trigger was also created which was unncessary. The trigger also caused a message `payload string too long` whenever the session record is updated (mainly because we store the access_policy data which can be very long). So this is for dropping it

packages/psss-server/src/app/attachModels.ts
```diff
@@ -3,17 +3,19 @@
...
 /**
  * Attach psss session model to the request
  */
-export const attachSessionModel = (sessionModel: PsssSessionModel) => (
+export const attachModels = (models: ModelRegistry, sessionModel: PsssSessionModel) => (
   req: PsssRequest,
   res: Response,
   next: NextFunction,
 ) => {
+  req.models = models;
```
chris-bes, line 18: Need this to get the entity records from the database in the endpoints. I also think we will need to use it more in the future

packages/psss-server/src/connections/ApiConnection.ts
```diff
@@ -23,7 +23,7 @@ export class ApiConnection {
     return this.request('GET', endpoint, queryParameters);
   }
 
-  post(endpoint: string, queryParameters: QueryParameters, body: RequestBody) {
+  post(endpoint: string, queryParameters: QueryParameters, body?: RequestBody) {
```
chris-bes, line 26: Making this optional because fetchingReport is a post but only requires body when sending test data

packages/database/src/migrations/20201129051349-DropPSSSSessionNotifyChangesTrigger-modifies-schema.js
edmofro: Thanks Chris, great job working that one out and fixing it up!

packages/psss-server/src/app/attachModels.ts
edmofro: The idea was that psss-server shouldn't have access to the database at all (session table was stored in the same database purely for convenience)

packages/psss-server/src/connections/ApiConnection.ts
edmofro: I'm not sure it should be a post then?

packages/psss-server/src/connections/ReportConnection.ts
```diff
@@ -9,4 +9,15 @@ const { REPORT_API_URL = 'http://localhost:8030/v2' } = process.env;
...
   baseUrl = REPORT_API_URL;
+
+  async fetchReport(reportCode: string, orgUnitCodes: string[], periods: string[] = []) {
+    if (!orgUnitCodes || !orgUnitCodes.length) {
+      throw new Error(`No organisationUnitCodes provided`);
+    }
+
+    return this.post(`fetchReport/${reportCode}`, {
+      organisationUnitCodes: orgUnitCodes.join(','),
+      period: periods.join(';')
```
edmofro, line 20: So the report server expects period to be specified in the same way as `DHIS2`, rather than as `startDate` and `endDate`?

packages/psss-server/src/connections/ReportConnection.ts
```diff
@@ -9,4 +9,15 @@ const { REPORT_API_URL = 'http://localhost:8030/v2' } = process.env;
 
 export class ReportConnection extends ApiConnection {
   baseUrl = REPORT_API_URL;
+
+  async fetchReport(reportCode: string, orgUnitCodes: string[], periods: string[] = []) {
+    if (!orgUnitCodes || !orgUnitCodes.length) {
+      throw new Error(`No organisationUnitCodes provided`);
+    }
+
+    return this.post(`fetchReport/${reportCode}`, {
```
edmofro, line 18: I definitely find it weird that the fetching of data uses a `post`!

packages/psss-server/src/connections/ReportConnection.ts
chris-bes: Yep from what I understand

packages/psss-server/src/connections/ReportConnection.ts
chris-bes: I think it was because initially we authenticate the requests using username/password (stored in the request body). So it was changed from a get to a post. Now we no longer do that but if we send test data (which added by @rohan-bes ) then we also need to store it in the request body

packages/psss-server/src/connections/ReportConnection.ts
edmofro: Oh right! Cool...I guess that can complicate things, but might be needed to support specific use cases.

packages/psss-server/src/connections/ReportConnection.ts
edmofro: Ok maybe that test data thing should be a special case - I think in general this should just be a `GET` with an Authorization header

packages/psss-server/src/routes/reports/ConfirmedWeeklyReportRoute.ts
```diff
@@ -0,0 +1,32 @@
...
+    return {
+      startWeek,
+      endWeek,
+      data: reportData,
+    };
+  }
+
+  async getAccessibleCountryCodes() {
+    const { accessPolicy } = await this.getSession();
+    const accessibleEntityCodes = accessPolicy.getEntitiesAllowed(PSSS_PERMISSION_GROUP);
```
edmofro, line 28: Assumption: these will be countries already. No need to drill into entity table.

packages/psss-server/src/routes/reports/ConfirmedCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,38 @@
...
+    return {
+      startWeek,
+      endWeek,
+      country: countryCode,
+      data: reportData,
+    };
+  }
+
+  async getCountryCode() {
+    const { organisationUnitCode } = this.req.params;
```
edmofro, line 29: Assumption: will be a country, so no need to drill into entity table

packages/psss-server/src/connections/ReportConnection.ts
rohan-bes: @edmofro @chris-bes sums it up correctly! Now I believe we can switch back to `fetchReport` being a `GET` endpoint, and we can switch the `testData`/`testConfig` behaviour on to a separate endpoint entirely (which would still be `POST`) 👍

packages/psss-server/src/routes/reports/ConfirmedCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,38 @@
...
+    const { accessPolicy } = await this.getSession();
+    const entity = await this.models.entity.findOne({ code: organisationUnitCode });
+    if (!entity || !accessPolicy.allows(entity.country_code, PSSS_PERMISSION_GROUP)) {
```
edmofro, line 32: I believe we talked about the report-server verifying access, so no need to do that here - is that right @rohan-bes?

edmofro (changes requested):
All very tidy but some queries and suggestions

packages/psss-server/src/connections/ReportConnection.ts
kael89: Related: https://github.com/beyondessential/tupaia/pull/1645#discussion_r527298489

packages/psss-server/src/routes/reports/ConfirmedCountryWeeklyReportRoute.ts
rohan-bes: Correct @edmofro!

packages/psss-server/src/routes/reports/ConfirmedWeeklyReportRoute.ts
chris-bes: Ah cool, if so I think I can remove the ModelRegistry from the request

packages/psss-server/src/routes/reports/ConfirmedCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { Route } from '../Route';
+
+export class ConfirmedCountryWeeklyReportRoute extends Route {
+  async buildResponse() {
+    const { startWeek, endWeek } = this.req.query;
+    const { organisationUnitCode: countryCode } = this.req.params;
+    const reportData = await this.reportConnection?.fetchReport(
```
edmofro, line 12: Sorry for more back and forth but I just thought of error handling here. Do we want to handle anything specifically, or is the generic error handling wrapped around all endpoints enough?

edmofro (approved):
Nice! One pending question, may not be necessary

packages/psss-server/src/routes/reports/ConfirmedCountryWeeklyReportRoute.ts
edmofro: For example, if the report-server sends back a specific error code, perhaps we want to pass that same error code on to the front end? Not sure though...let me know your thoughts and I can return with a fresh brain tomorrow!

chris-bes: That sounds very useful! I guess we can do this at the ApiConnection level so that it's available to all the Api connections. I had a look at the `verifyResponseStatus` util that you used in ApiConnection and maybe we can change it to something like the below?
```
const throwCustomError = (status, errorMessage) => {
  const statusCode = status || 500;
  throw new CustomError({
    responseStatus: statusCode,
    responseText: errorMessage,
  });
};

export const verifyResponseStatus = async response => {
  if (!response.ok) {
    let responseJson;
    try {
      responseJson = await response.json();
    } catch (error) {
      throw new Error(`Network error ${response.status}`);
    }
    if (
      response.status &&
      (response.status < 200 || response.status >= 300) &&
      !responseJson.error
    ) {
      throwCustomError(response.status, responseJson.message);
    }
    if (responseJson.error) {
      throwCustomError(response.status, responseJson.error);
    }
  }
};
```

edmofro: Sounds great!

chris-bes: As discussed, will resolve this in another PR later!

PR #1719, opened 2020-11-30
Hotfix: set setOverlayComponent(LANDING) on sucess

nick-beyondessential:
### Changes:
After sucessful login to tupaia, user should be directed to project landing page.
image

### DIFF
--- a/packages/web-frontend/src/sagas.js
+++ b/packages/web-frontend/src/sagas.js
@@ -1049,7 +1049,9 @@ function* fetchLoginData(action) {
   if (action.loginType === LOGIN_TYPES.MANUAL) {
     const { routing: location } = yield select();
     yield call(fetchProjectData);
-    yield put(setOverlayComponent(LANDING));
+    const { PROJECT } = decodeLocation(location);
+    const overlay = PROJECT === 'explore' ? LANDING : null;
+    yield put(setOverlayComponent(overlay));
     yield call(handleLocationChange, {
       location,
       // Assume an empty location string so that the url will trigger fetching fresh data
--- a/packages/web-frontend/src/sagas.js
+++ b/packages/web-frontend/src/sagas.js
@@ -1048,8 +1048,8 @@ function* watchLogoutSuccess() {
 function* fetchLoginData(action) {
   if (action.loginType === LOGIN_TYPES.MANUAL) {
     const { routing: location } = yield select();
-    yield put(setOverlayComponent(null));
     yield call(fetchProjectData);
+    yield put(setOverlayComponent(LANDING));
     yield call(handleLocationChange, {
       location,
       // Assume an empty location string so that the url will trigger fetching fresh data


tcaiger (approved):
Nice - looks good to me!

edmofro:
This should be done using the hot fix process please! Follow the hot fix from @kael89 on Friday for how that works

tcaiger (changes requested):
@nick-beyondessential - I just got my head back into how this works and this will cause a regression on the scenario where user goes to a private project url while not logged in. In that case, it should take them to the login and then after they login it should send them to the correct url. I will have a quick look and see if I can propose a solution.

packages/web-frontend/src/sagas.js
```diff
@@ -1048,8 +1048,8 @@ function* watchLogoutSuccess() {
 function* fetchLoginData(action) {
   if (action.loginType === LOGIN_TYPES.MANUAL) {
     const { routing: location } = yield select();
-    yield put(setOverlayComponent(null));
     yield call(fetchProjectData);
+    yield put(setOverlayComponent(LANDING));
```
tcaiger, line 1052:
```suggestion
    yield call(fetchProjectData);
    const { PROJECT } = decodeLocation(location);
    const overlay = PROJECT === 'explore' ? LANDING : null;
    yield put(setOverlayComponent(overlay));```

I think that should do it.
ps. I assume that it is just on the 'explore' project that we want to show the landing page after login.

tcaiger: I think that should do it.
ps. I assume that it is just on the 'explore' project that we want to show the landing page after login.

nick-beyondessential: Hrm...this seems to work, if I go into private project url unlogged in, then it brings up log in form, after log in sucess it brings up the project page briefly then closes it. This may be acceptable?

nick-beyondessential: I guess the other senerio that it should bring up the project page is where a user links into a project they don't have access to.

nick-beyondessential: Actually no, if a user doesn't have access to the project after logging in, they get the request access modal, which is ideal!
Only the half a second flash of the project page before closing seems to be the only issue.

tcaiger: hmmm given that this is a hotfix, I'd say it is acceptable to have the flash. Obviously it's not ideal but it is also not the main flow that users take. I had a quick look and am not sure why there is the flash as the default state for the overlay is null which was my first thought. Maybe have a quick look at fixing it and if you can't nail it down in then create a bug card for it to be picked up later.

tcaiger (commented):
Approving as this fix addresses the main issue. Please make a card to pick up the overlay flash if required.

nick-beyondessential:
Thanks @tcaiger will fix on new card: https://github.com/beyondessential/tupaia-backlog/issues/1698

PR #1720, opened 2020-11-30
1090 Sort order of org unit columns in matricies

IgorNadj:
### Changes:

- Sort matrix columns by org unit name
- Sort country names in project description
- Sort country names in project cards

---


edmofro (approved):

IgorNadj:
@edmofro added a commit

packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/ProjectCard.js
```diff
@@ -126,7 +126,7 @@ export const ProjectCard = ({
       <Title>{name}</Title>
       <FullWidthRow>{description}</FullWidthRow>
       <Footer>
-        <Countries>{name === 'Disaster Response' ? 'Global' : names.join(', ')}</Countries>
+        <Countries>{name === 'Disaster Response' ? 'Global' : names.sort().join(', ')}</Countries>
```
edmofro, line 129: The fix should be on the server instead, if we always want them sorted alphabetically. Then again, it's kind of a presentational concern, so this is still approved :+1:

PR #1729, opened 2020-12-01
Use Windows friendly paths in scripts

kael89:
### Issue #:
[No issue] - bumped into this when working on something else. Addresses two issues:

1. Make yarn scripts compatible with Windows

It seems that Windows are not happy with `'`, that's why we have to use `\"` all over the place. In practice, this enables
1. Building packages under Windows (probably most devs do this in WSL currently)
2. Running `yarn refresh-database -m` successfully (the `-m` part would fail, because it invokes `yarn migrate` which includes a non-Windows friendly path)

So no significant wins at this stage, but still seems as a better practice to me

2. Fix babel caching

By default, `babel` does not delete existing files in the `dist` folder, just adds/updates them. This can result to broken code when someone eg moves `function.js` to `function/index.js`. Both will still exist under the `dist` folder, which may create issues when `function.js` is now broken (eg references a dependency which is now deleted).

This actually happened when we removed `@beyondessential/arithmetic` from `indicators`, and moved `buildArithmetic.ts` under a `buildArithmetic` folder


packages/access-policy/package.json
```diff
@@ -11,7 +11,7 @@
   "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
   "main": "dist/index.js",
   "scripts": {
-    "build": "babel src --out-dir dist --source-maps --config-file '../../babel.config.json'",
+    "build": "babel src --delete-dir-on-start --out-dir dist --source-maps --config-file \"../../babel.config.json\"",
```
kael89, line 14: The `delete-dir-on-start` option is annoyingly not documented! I found it while searching around for similar issues. It was added in this PR: https://github.com/babel/babel/pull/6187

I guess we could also simply delete the `dist` file before we run `babel`

rohan-bes (approved):
Looks good, great find on the `delete-dir-on-start` arg!!

kael89:
@rohan-bes thanks! I'll test all is good with a feature branch before merging this

scripts/bash/buildInternalDependencies.sh
```diff
@@ -52,6 +56,10 @@ if [[ $watch == "true" ]]; then
     echo "yarn concurrently ${build_commands[@]}"
     eval "yarn concurrently ${build_commands[@]}"
 else
+    echo "Deleting existing internal dependency builds"
+    echo "${delete_command}"
+    eval "${delete_command}"
```
kael89, line 61: @rohan-bes I reverted my changes in the [Fix babel caching commit](https://github.com/beyondessential/tupaia/pull/1729/commits/2e06531ba05255f4ca153e855a361aec4d776aef). In that commit, we delete the `dist` folder before building a package. While this works in actual deployments, `yarn build-internal-dependencies` doesn't work in watch mode: I suspect this is because watch mode expects a `dist` folder to pre-exist

@edmofro Tagging you here cause I think you should also have a look at this

kael89: By the way, I created a test feature deployment using this branch it seems to be working fine

kael89: Also, deleted files will still not be picked up in `watch` mode. This is the existing behaviour, which can result in weird scenarios during local development: eg

1. A dev deletes a file in a dependency. There forget to delete an `import` of this file
2. The change is not being picked up. However, the dev assumes that hot reload would have picked all changes
3. Dev tests their code locally, everything looks good, so they commit & push their change. They haven't run `build-internal-dependencies` since they pulled their base branch - theoretically, they shouldn't have to
4. They deploy their code and it doesn't work

But at least now deployment builds map accurately to the repository code.

edmofro: This lgtm @kael89, thanks for trying out that undocumented babel command, it seemed good, but this also seems a solid strategy. I'm not worries about that scenario you mentioned above (and nor is it a regression)

PR #1731, opened 2020-12-01
fetchProjectData after setOverlayComponent login

nick-beyondessential:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/1698

reordering fetchLoginData saga to set setOverlayComponent prior to fetchProjectData to avoid landing page flashing up after loging into a private project


tcaiger (approved):
Simply genius! Thanks for following up on this.

PR #1735, opened 2020-12-01
1368: PSSS app endpoint: GET Unconfirmed Weekly Report per Week for a Country

chris-bes:
### Changes:

- Added PSSS Indicator for Total Sites
- Added get unconfirmed weekly report endpoint.
- Changed verifyResponse to also pass the error code back


packages/utils/src/request.js
```diff
@@ -80,14 +89,14 @@ export const verifyResponseStatus = async response => {
...
+      response.status &&
+      (response.status < 200 || response.status >= 300) &&
       !responseJson.error
     ) {
-      throw new Error(responseJson.message);
+      throwCustomError(response.status, responseJson.message);
     }
     if (responseJson.error) {
-      throw new Error(responseJson.error);
+      throwCustomError(response.status, responseJson.error);
```
chris-bes, line 99: This is for resolving something that I promised Edwin I would resolve later. Context is all here: https://github.com/beyondessential/tupaia/pull/1716#discussion_r532389822

packages/database/src/migrations/20201201022321-AddPSSSTotalSitesIndicator-modifies-data.js
```diff
@@ -0,0 +1,46 @@
...
+  });
+};
+
+exports.down = function (db) {
+  return null;
+};
+
+exports._meta = {
+  version: 1,
+};
```
chris-bes, line 46: There was a missing indicator `PSSS_Total_Sites` which is used in the report. So this migration is for creating it

rohan-bes (changes requested):
Just wanna hear your thoughts on the naming convention for confirmed/unconfirmed but totally happy with everything else. These endpoints are looking great 🎉

packages/psss-server/src/app/addRoutesToApp.ts
```diff
@@ -50,6 +51,10 @@ export function addRoutesToApp(app: Express) {
     '/v1/confirmedWeeklyReport/:organisationUnitCode',
     handleWith(ConfirmedCountryWeeklyReportRoute),
   );
+  app.get(
+    '/v1/unconfirmedWeeklyReport/:organisationUnitCode',
+    handleWith(UnconfirmedCountryWeeklyReportRoute),
+  );
```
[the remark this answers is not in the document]
chris-bes: I actually was not aware of the terms we're exposing externally... so that's a good point you brought up. I second your opinion that we should make it consistent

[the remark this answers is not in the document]
kael89: I agree with you @rohan-bes , this is consistent with the naming conventions used in indicators and data elements. We can probably skip `unconfirmed` even in the code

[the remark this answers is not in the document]
edmofro: Agreed :+1:

packages/psss-server/src/routes/reports/ConfirmedCountryWeeklyReportRoute.ts
```diff
@@ -12,7 +12,7 @@ export class ConfirmedCountryWeeklyReportRoute extends Route {
     const reportData = await this.reportConnection?.fetchReport(
       'PSSS_Confirmed_Weekly_Report',
       [countryCode],
-      [startWeek], //Only need to send startWeek period because this is a weekly report
+      [startWeek, endWeek],
```
chris-bes, line 15: Fixed a mistake from last PR

packages/report-server/.env.example
```diff
@@ -7,4 +7,5 @@ DHIS_CLIENT_ID=
 DHIS_CLIENT_SECRET=
 DHIS_PASSWORD=
 DHIS_USERNAME=
+JWT_SECRET=
```
rohan-bes, line 10: Ahh thanks for adding this 👍

rohan-bes (approved):
Looks great! 👍

PR #1741, opened 2020-12-02
1666 add project palau olangch

nick-beyondessential:
https://github.com/beyondessential/tupaia-backlog/issues/1666 Create project card - Palau

In this PR:
- migration to add Palau MoH Project Olangch 
- hide null logo images in project request modal


packages/database/src/migrations/20201201082830-AddProjectPalauOlangch-modifies-data.js
```diff
@@ -0,0 +1,98 @@
...
+    parent_id: await codeToId(db, 'entity', PROJECT_CODE),
+    child_id: await codeToId(db, 'entity', COUNTRY_CODE),
+    entity_hierarchy_id: await hierarchyNameToId(db, PROJECT_CODE),
+  });
+  await insertObject(db, 'project', {
+    id: generateId(),
+    code: PROJECT_CODE,
+    description: 'HIS for Palau’s public health data, integrating DHIS2 and mSupply.',
+    sort_order: 10,
+    image_url: '', // 'https://tupaia.s3-ap-southeast-2.amazonaws.com/uploads/palau_olangch_background.png',
```
nick-beyondessential, line 74: Logo image is a temp image.
Background image hasn't been supplied yet. Another option would be to add a blank image with this name to aws - so that when real one comes through, doesn't need code change.

nick-beyondessential: Went with uploading a blank image.

packages/database/src/migrations/20201201082830-AddProjectPalauOlangch-modifies-data.js
```diff
@@ -0,0 +1,99 @@
...
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const COUNTRY_CODE = 'PW';
+const PROJECT_CODE = 'olangch_palau';
+const PROJECT_NAME = 'MoH Project Olangch';
+const DASHBOARDGROUP_NAME = 'General';
+const DASHBOARDGROUP_CODE = 'PW_Olangch';
```
kael89, line 23: According to an implicit existing convention:

```suggestion
const DASHBOARDGROUP_CODE = 'OLANGCH_PALAU_Project';
```

Although I don't know why we use upper case for the project name, `Olangch_Palau_Project` would work just as fine for me, let's stick to the existing convention

Alastair-L, line 23: Perhaps add a 'project'
image

kael89 (changes requested):
I saw that @Alastair-L just picked this up so I'll leave the rest of the review to him - just a suggestion!

packages/database/src/migrations/20201201082830-AddProjectPalauOlangch-modifies-data.js
Alastair-L: Also I think most project level dashboards are called the name of the project (e.g. UNFPA), but that's not a big deal because Georgia told you otherwise

Alastair-L (approved):
Nice one! There's a lot to do when adding a project though. One note on code conventions.

packages/database/src/migrations/20201201082830-AddProjectPalauOlangch-modifies-data.js
nick-beyondessential: Ah yes good pick up thanks!

nick-beyondessential: I think standardisation is a good idea here, and i was my mistake. Will update.

packages/database/src/migrations/20201201082830-AddProjectPalauOlangch-modifies-data.js
```diff
@@ -0,0 +1,99 @@
...
+    dashboardReports: '{project_details}',
+    name: DASHBOARDGROUP_NAME,
+    code: DASHBOARDGROUP_CODE,
+    projectCodes: `{${PROJECT_CODE}}`,
+  });
+  await insertObject(db, 'dashboardGroup', {
+    organisationLevel: 'Country',
+    userGroup: USER_GROUP,
+    organisationUnitCode: COUNTRY_CODE,
+    dashboardReports: '{23,19,8,26}',
```
kael89, line 46: @nick-beyondessential also another thing, I don't know if we'll want to show the above reports in this project? I know that we show them in `General` dashboard groups, but those are attached to the older projects of our platform (`explore`, `disaster`, `imms`, `fanafana`). To be honest, I'm not sure if we should show them in all those projects, perhaps we keep them because people are used into seeing them there.

I'd double check with Georgia whether `General` is the most appropriate name for this group, (perhaps it's not and we can suggest `Olangch` for now, as @Alastair-L pointed out). I would also confirm whether those reports here should be displayed in the project's dashboard

nick-beyondessential: Georgia is happy to follow convention. I have also removed those reports as they were just the ones in General and have used project_details until given specific reports.

kael89: Sounds great, thanks!

kael89 (approved):

PR #1745, opened 2020-12-02
1696 upgrade react navigation

chris-bes:
### Changes:

- Example

---


edmofro:
@chris-bes now I've got this checked out I've noticed those transitions still aren't the same as they used to be. Can you compare an old version what they do now? The login/sync/main screen ones previously were just a quick-replace, whereas now they're a slide-from-right.

chris-bes (commented):
I did look through in detail as much as I can but I couldn't really see any major problems...

PR #1746, opened 2020-12-02
1499: PSSS Submit Confirmed Weekly Report

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1499:

Submitting survey responses from `psss-server` for Confirmed Weekly Reports.

Required changing `meditrak-server` a bit to allow for submitting by user.email and question.code.

Not 100% finished as still need a bit of confirmation on the period and answers formats from the front-end


packages/psss-server/src/types.ts
```diff
@@ -36,7 +36,7 @@ export interface Credentials {
 
 export type QueryParameters = Record<string, string>;
 
-export type RequestBody = Record<string, unknown>;
+export type RequestBody = Record<string, unknown> | Record<string, unknown>[];
```
rohan-bes, line 39: The body of the `/changes` endpoint in `meditrak-server` is an array of JSON objects

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -3,10 +3,112 @@
...
+    return results.length > 0 ? results[0] : undefined;
+  }
+
+  async updateSurveyResponse(
+    surveyResponse: SurveyResponseObject,
+    answers: Record<string, string>,
+  ) {
+    const existingAnswers = (await this.get(`surveyResponses/${surveyResponse.id}/answers`, {
+      columns: `["${ANSWER_FIELDS.join('","')}"]`,
+    })) as AnswerObject[];
```
rohan-bes, line 54: Unfortunately, for now we still need to query `meditrak-server` to get the ids of the answers we will be overwriting. I left it this way as otherwise it would be a bit more complicated and risky to implement logic on the `meditrak-server` side that determined if answers ought to be overwritten or not...

In the future though, we should definitely go down that route 👍

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -3,10 +3,112 @@
...
 
 const { MEDITRAK_API_URL = 'http://localhost:8090/v2' } = process.env;
 
+const SURVEY_RESPONSE_FIELDS = <const>['entity.code', 'survey.code', 'submission_time', 'id'];
+const ANSWER_FIELDS = <const>['question.code', 'id'];
+const PSSS_SURVEY_RESPONSE_ANSWER_TYPE = 'Number'; // All survey response answers are 'Number' in PSSS
+
+type SurveyResponseObject = {
+  [key in typeof SURVEY_RESPONSE_FIELDS[number]]: string;
+};
```
rohan-bes, line 18: Is this being too fancy? Basically I'm constructing a type here who's fields are defined in an array. It's pretty nifty but I dunno if it's necessary 🤷

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -3,10 +3,112 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
+import { generateId } from '@tupaia/database';
```
rohan-bes, line 6: I believe this is a safe way of generating id's for creating new survey responses / answers?

packages/meditrak-server/src/routes/postChanges.js
```diff
@@ -37,6 +44,7 @@ export async function postChanges(req, res) {
     if (!ACTION_HANDLERS[action]) {
       throw new ValidationError(`${action} is not a supported change action`);
     }
+    await PAYLOAD_TRANSLATORS[action](models, payload);
```
rohan-bes, line 47: This was my solution of how to submit survey responses by code/email/etc. Basically we take in the payload, and we `translate` the unsupported fields (`question_code`, `user_email`, etc.) into the expected fields (`question_id`, `user_id`, `assessor_name`, etc.).

I don't particularly like this solution in the long term, and I think are more thought out refactor of how these endpoints work is called for... Perhaps we associate requests with a user, and allow for a more generic id/code referencing system?

However, for now I think this is lower risk, and quicker and easier to implement...

Happy to get push back on this whole implementation if people find it too hacky 👍

packages/psss-server/src/routes/reports/SubmitConfirmedCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,90 @@
...
+
+type ConfirmedWeeklyReportAnswers = {
+  PSSS_Confirmed_Sites: number;
+  PSSS_Confirmed_Sites_Reported: number;
+  PSSS_Confirmed_AFR_Cases: number;
+  PSSS_Confirmed_DIA_Cases: number;
+  PSSS_Confirmed_ILI_Cases: number;
+  PSSS_Confirmed_PF_Cases: number;
+  PSSS_Confirmed_DLI_Cases: number;
+};
```
rohan-bes, line 20: I'm quite a fan of defining in-file types like this. It helps me do type checking easier and generally makes things clearer to me. However I have a feeling you're meant to define all types in the root `/types` file? Any thoughts on this?

packages/psss-server/src/routes/reports/SubmitConfirmedCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,90 @@
...
+export class SubmitConfirmedCountryWeeklyReportRoute extends Route {
+  async buildResponse() {
+    const { week } = this.req.query;
+    const { organisationUnitCode } = this.req.params;
+
+    const report = await this.reportConnection?.fetchReport(
+      PSSS_WEEKLY_REPORT_CODE,
+      [organisationUnitCode],
+      [week],
+    );
```
rohan-bes, line 31: We fetch the existing unconfirmed data from the report-server here. We could actually construct a separate report for this that only returns the required data? It should have improved performance that way...

packages/psss-server/src/connections/MeditrakConnection.ts
edmofro: Yeah it is...a bit rough making psss-server generate ids though! Can you make a psss-server tech debt card and add this to it, for addressing later?

packages/psss-server/src/routes/reports/SubmitConfirmedCountryWeeklyReportRoute.ts
edmofro: I've had the same question in my mind but don't know the answer!

packages/psss-server/src/types.ts
edmofro: We shouldn't be using the `/changes` endpoint in `meditrak-server` - that's intended for sync. Could you add that to the tech debt card too?

packages/psss-server/src/routes/reports/SubmitConfirmedCountryWeeklyReportRoute.ts
edmofro: Oh or the front end could just send through the data in the request body?

packages/psss-server/src/connections/MeditrakConnection.ts
edmofro: Yeah it is nifty but I'd prefer just listing them out in the object...just an extra thing to think about when you're working out the type. 

To confirm my understanding: `number` is a special keyword that TypeScript uses when mapping an array out into an object?

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -3,10 +3,112 @@
...
+
 export class MeditrakConnection extends ApiConnection {
   baseUrl = MEDITRAK_API_URL;
+
+  async findSurveyResponse(surveyCode: string, orgUnitCode: string, period: string) {
+    const [startDate, endDate] = convertPeriodStringToDateRange(period);
+    const results = (await this.get(`surveyResponses/`, {
+      page: '0',
+      pageSize: '1',
+      columns: `["${SURVEY_RESPONSE_FIELDS.join('","')}"]`,
```
edmofro, line 32: Oh I see why it's doubly helpful. Yeah still...probably better just to have some duplicate code than the extra layer of abstraction to think about when looking at the type. Makes sense to use the array here though of course! Is there a way to do it the other way around? Object.keys on a type?

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -3,10 +3,112 @@
...
+      filter: JSON.stringify({
+        'survey.code': { comparisonValue: surveyCode, castAs: 'text' },
+        'entity.code': { comparisonValue: orgUnitCode, castAs: 'text' },
```
edmofro, line 35: Are these casts necessary?

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -3,10 +3,112 @@
...
+        submission_time: {
+          comparator: 'BETWEEN',
+          comparisonValue: [startDate, endDate],
+          castAs: 'date',
+        },
+      }),
+      sort: '["submission_time DESC"]',
```
edmofro, line 42: Nice, so if there happens to be more than one it will update the most recent.

packages/psss-server/src/connections/MeditrakConnection.ts
edmofro: That's ok, when a survey response exists we're updating using the id so I feel alright about using ids to update answers too.

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -3,10 +3,112 @@
...
+
+  async updateSurveyResponse(
+    surveyResponse: SurveyResponseObject,
+    answers: Record<string, number>,
+  ) {
+    const existingAnswers = (await this.get(`surveyResponses/${surveyResponse.id}/answers`, {
+      columns: `["${ANSWER_FIELDS.join('","')}"]`,
+    })) as AnswerObject[];
+
+    const newAnswers = existingAnswers.map(existingAnswer => ({
```
edmofro, line 56: We might drop an answer to a question that wasn't answered in the first survey response? Perhaps that's an edge case that never comes up in the flow we're using? I.e., if there's an existing survey response, it will have all questions answered?

packages/meditrak-server/src/routes/postChanges.js
edmofro: I'm actually ok with this implementation - I don't think we need to have a different mapping for incoming data per client, given that they're pushing to us. But I don't think it's quite in the right place - I'd like `/changes` to be restricted to supporting syncing clients (so far just the data collection app).

packages/meditrak-server/src/routes/postChanges.js
```diff
@@ -72,6 +80,18 @@ const PAYLOAD_VALIDATORS = {
   },
 };
 
+/**
+ * Contains functions that accept the payload, and translate the data so that the required fields are present
+ * for the relevant change action, returning true if successful and throwing an error if not
+ */
+const PAYLOAD_TRANSLATORS = {
+  [SUBMIT_SURVEY_RESPONSE]: async (models, payload) =>
```
edmofro, line 88: Pretty clean and matches the validator pattern, but I would prefer it wasn't mutating tbh. Do you think it would be a big performance impact to deep copy the payload and return the translated version?

packages/meditrak-server/src/routes/utilities/translateObjectFields.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+export const translateObjectFields = async (object, objectTranslators) => {
```
edmofro, line 6: Nice little util!

packages/meditrak-server/src/routes/postChanges.js
```diff
@@ -92,12 +112,41 @@ async function validateSurveyResponseObject(models, surveyResponseObject) {
...
+  }
+
+  const surveyResponseTranslators = constructSurveyResponseTranslators(models);
+  await translateObjectFields(surveyResponseObject, surveyResponseTranslators);
+
+  const { answers } = surveyResponseObject;
+  if (Array.isArray(answers)) {
+    const answerTranslators = constructAnswerTranslators(models);
+    for (let i = 0; i < answers.length; i++) {
+      await translateObjectFields(answers[i], answerTranslators);
```
edmofro, line 127: Could be quite a bit more efficient to fetch the question code to id mapping in one db query, and then use that mapping object during each translation

edmofro (approved):
I have some suggestions but in the interest of time I'm happy to delay them into the tech debt card, so approved with decisions about what to change now vs. leave for later over to you @rohan-bes

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -3,10 +3,112 @@
...
+    const newAnswers = Object.entries(answers).map(([questionCode, value]) => ({
+      id: generateId(),
+      type: PSSS_SURVEY_RESPONSE_ANSWER_TYPE,
+      question_code: questionCode,
+      body: value,
+    }));
+
+    const date = new Date().toISOString();
+
+    return this.post(`changes`, {}, [
```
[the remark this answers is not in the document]
rohan-bes: I looked into it a bit, but it seemed as though we'd need to make a call per answer to `/surveyResponse/:id/answers/:id` in order to update the existing answers? That's why I went with `/changes` as it seemed to be more dedicated to updating/creating survey responses? Let's have a quick chat about this today?

packages/psss-server/src/connections/MeditrakConnection.ts
rohan-bes: Yeah I think that's an invalid use case for PSSS. All survey responses must contain case data for each syndrome

edmofro: Great - makes things easy :-)

packages/meditrak-server/src/routes/postChanges.js
rohan-bes: Probably wouldn't be the biggest performance hit, though it would affect all survey_responses submitted via this endpoint. Given that almost all of these come from meditrak-app and don't require any translation, doing a deep copy for each seems a bit risky...
What I might do is some kind of upfront check whether or not a response requires any translation before copying, which I think might even be more performant that what we have currently 👍

packages/meditrak-server/src/routes/utilities/translateObjectFields.js
rohan-bes: I actually thought about putting this into `@tupaia/utils` where `ObjectValidator` sits. But for now it still feels a little too custom to me?

packages/psss-server/src/connections/MeditrakConnection.ts
rohan-bes: https://github.com/beyondessential/tupaia-backlog/issues/1755

packages/psss-server/src/connections/MeditrakConnection.ts
rohan-bes: >Yeah it is nifty but I'd prefer just listing them out in the object...just an extra thing to think about when you're working out the type.

Yeah, I think you're right.

>To confirm my understanding: number is a special keyword that TypeScript uses when mapping an array out into an object?

Yup, tbh this still feels a little like magic to me 🤷

packages/psss-server/src/connections/MeditrakConnection.ts
rohan-bes: >Is there a way to do it the other way around? Object.keys on a type?

I wish!! I've actually been looking around for this for a while. There are custom libraries that do this (https://github.com/kimamula/ts-transformer-keys) but natively I don't think this is possible 😞

edmofro: Oh well. We repeat the keys in the filter again, which makes them somehow feel cheap to repeat in other places haha

packages/psss-server/src/routes/reports/SubmitConfirmedCountryWeeklyReportRoute.ts
edmofro: We talked about this and confirmed that requesting the data here again is safest, though might be made faster by the addition of another report. Happy to resolve this comment

packages/psss-server/src/connections/MeditrakConnection.ts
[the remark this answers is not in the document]
edmofro: Talked about this and decided to push it out into the tech debt card, but that we would aim to one day have it avoid the `/changes` endpoint

packages/meditrak-server/src/routes/postChanges.js
edmofro: Oh yeah true, it would affect all those meditrak changes. Cool a check sounds good, in the absence of moving to a completely different endpoint handler.

packages/meditrak-server/src/routes/utilities/translateObjectFields.js
edmofro: Yeah...I think I agree with that. But keep it in mind when working across other packages of course!

packages/psss-server/src/connections/MeditrakConnection.ts
rohan-bes: Unfortunately, I think this may be an area where we will struggle with the whole `endTime` vs `submissionTime` thing. So in theory we want to be working with the most recent `endTime` (ie. the most recently edited) survey response. However, since the `report-server` has no access to that, it can only return the most recent `submissionTime`. I figured it's more important for this endpoint to be editing the survey response that the user is seeing data for in PSSS, but it does sorta mean that a survey response that comes through via meditrak which doesn't have a more recent `submission_time` will kinda get lost in limbo 🤷

edmofro: Oh man, I was caught out by my own terrible name here 🤦 

Yep all good, there's another card that end_time stuff is deferred into right?

rohan-bes: Yup: https://github.com/beyondessential/tupaia-backlog/issues/1691

packages/psss-server/src/connections/MeditrakConnection.ts
rohan-bes: I am not sure, I copied them from the admin panel, I'll check 👍

packages/meditrak-server/package.json
```diff
@@ -47,7 +47,7 @@
     "form-data": "^2.3.3",
     "format-link-header": "^2.1.0",
     "http": "^0.0.0",
-    "lodash": "^4.17.11",
```
rohan-bes, line 50: So it looks like `lodash` wasn't being used anywhere in `meditrak-server`?

packages/meditrak-server/src/routes/postChanges.js
rohan-bes: While you're totally right about this, I'm thinking of maybe leaving it as is for now? Doing the fetch upfront would be a bit of a tricky code refactor, and given that this code will probably be totally refactored soon enough I'm okay with leaving it a little inefficient?

packages/psss-server/src/routes/reports/SubmitConfirmedCountryWeeklyReportRoute.ts
rohan-bes: I think generally we should try and sketch up some guidelines for good typescript practices. I'll need to do some more research and develop more experience. If you're happy with this I'd like to keep it?

packages/psss-server/src/connections/MeditrakConnection.ts
rohan-bes: They were not necessary, so I removed them

packages/psss-server/src/routes/reports/SubmitConfirmedCountryWeeklyReportRoute.ts
edmofro: Happy as larry

packages/meditrak-server/src/routes/postChanges.js
edmofro: Yep :+1:

packages/meditrak-server/package.json
edmofro: Nice!

PR #1748, opened 2020-12-02
1696 upgrade react native etc

edmofro:


packages/meditrak-app/app/authentication/actions.js
```diff
@@ -49,7 +49,7 @@ export const login = (emailAddress, password) => async (
   const loginCredentials = {
     emailAddress,
     password,
-    deviceName: DeviceInfo.getDeviceName(),
+    deviceName: await DeviceInfo.getDeviceName(),
```
chris-bes, line 52: I guess this is why it comes back as unknown for you before?

chris-bes (commented):
This looks fine to me but to be honest I'm not very confident since it has a lot of stuff that I'm still very green :(. Wouldn't mind if this gets another pair of eyes on it...

packages/meditrak-app/app/authentication/actions.js
edmofro: No the `await` was in place when I reported that. Still not sure, I think it's just the Genymotion device I have set up doesn't give a name.

edmofro:
> This looks fine to me but to be honest I'm not very confident since it has a lot of stuff that I'm still very green :(. Wouldn't mind if this gets another pair of eyes on it...

I think that's fine tbh @chris-bes - anyone at BES will be great to this stuff, so your approval is enough for me given how busy the rest of the team is! :-)

chris-bes (approved):
Did the last smoke test here (https://github.com/beyondessential/tupaia-backlog/issues/1696#issuecomment-742857175). Approve!

PR #1758, opened 2020-12-03
1628 handle report weeks

tcaiger:
### Changes:
- Determine if the upcoming week is submitted and correct upcoming report data in `useUpcomingReport.js` 
- Create `UpcomingReportCard` to display fetched data

Side not: Stubbed out edit profile section until [this](https://github.com/beyondessential/tupaia-backlog/issues/1117) is done


packages/psss/src/containers/UpcomingReportCard.js
```diff
@@ -0,0 +1,119 @@
...
+        <DateSubtitle variant="subtitle2" gutterBottom>
+          {getFormattedStartByPeriod(period, 'LLL d, yyyy')} -{' '}
+          {getFormattedEndByPeriod(period, 'LLL d, yyyy')}
+        </DateSubtitle>
+        <StyledButton fullWidth onClick={() => handleOpen()}>
+          {buttonText}
+        </StyledButton>
+      </CardContent>
+      <CardFooter>
+        <BarMeter value={22} total={30} legend="Sites reported" />
```
tcaiger, line 102: This is still dummy data while waiting on the indicator

packages/psss/src/components/DateToolbar/DateToolbar.js
```diff
@@ -4,43 +4,50 @@
  */
```
tcaiger, line 4: @kael89 Sorry I made some last minute changes to the `DateToolbar` this morning because when I was testing I noticed some inconsistencies (UTC vs NZT) with the dates I was getting from date-fns and the tupaia period utils. My fix is to update everywhere to use the utls and now it works well. I should have listened to you the first time when you said to encapsulate the week formatting in tupaia utls 😄

packages/psss/src/containers/UpcomingReportCard.js
```diff
@@ -0,0 +1,124 @@
...
+const DateSubtitle = styled(Typography)`
+  margin-top: 1rem;
+  margin-bottom: 1rem;
+  font-weight: 400;
+  color: ${props => props.theme.palette.text.secondary};
+`;
+
+const getDisplayDays = days => {
+  const d = Math.abs(days);
+  return `${Math.abs(d)} day${d > 1 ? 's' : ''}`;
```
kael89, line 37: No need to use `Math.abs` twice:

```suggestion
  const d = Math.abs(days);
  return `${d} day${d > 1 ? 's' : ''}`;
```

packages/psss/src/containers/UpcomingReportCard.js
```diff
@@ -0,0 +1,124 @@
...
+        label={<Warning color="error" />}
+      />
+    );
+  }
+
+  if (status === REPORT_STATUSES.UPCOMING && days === 0) {
+    return (
+      <CardHeader color="error" title="Confirmation due today" label={<Warning color="error" />} />
+    );
+  }
```
kael89, line 57: Since 0 is `<= 3`, the 2nd check would never get evaluated. Grouping by `status` may also enhance readability:

```suggestion
  if (status === REPORT_STATUSES.UPCOMING && days <= 3) {
    const title = days === 0 ? 'Confirmation due today' : `Confirmation due in ${displayDays}`;
    return <CardHeader color="error" title={title} label={<Warning color="error" />} />;
  }
```

packages/psss/src/api/queries/useUpcomingReport.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  const status = days > 0 ? REPORT_STATUSES.UPCOMING : REPORT_STATUSES.OVERDUE;
+
+  return {
+    ...query,
+    data,
+    reportStatus: status,
+    period: lastPeriod,
+    days,
```
kael89, line 53: Perhaps using `daysTillDueDay` instead of `days` would be even clearer (while the latter is pretty clear here, I was a bit confused when confirming its usage in `UpcomingReportCard`

packages/psss/src/containers/UpcomingReportCard.js
```diff
@@ -0,0 +1,124 @@
...
+    <Card variant="outlined">
+      <HeaderComponent days={days} status={reportStatus} />
+      <CardContent>
+        <Typography variant="h4">Week {getWeekNumberByPeriod(period)}</Typography>
+        <Typography variant="h4" gutterBottom>
+          Upcoming Report
+        </Typography>
+        <DateSubtitle variant="subtitle2" gutterBottom>
+          {getFormattedStartByPeriod(period, 'LLL d, yyyy')} -{' '}
+          {getFormattedEndByPeriod(period, 'LLL d, yyyy')}
```
kael89, line 100: I guess we could use [getDisplayDatesByPeriod](https://github.com/beyondessential/tupaia/blob/1628-handle-report-weeks/packages/psss/src/utils/period.js#L21) here? There's also another potential use case in [DateToolbar](https://github.com/beyondessential/tupaia/blob/1628-handle-report-weeks/packages/psss/src/components/DateToolbar/DateToolbar.js#L119).

Start date format is different in all those cases, but if consistency in the UI is beneficial in all those cases we should use the util

kael89 (changes requested):
As always, great work @tcaiger ! Just a few details here and there

packages/psss/src/containers/UpcomingReportCard.js
tcaiger: hmmm I thought about this when putting it together and decided that they are different enough to be implemented separately. I think they are also the kind of thing a designer might tweak in one place or another but not everywhere etc

kael89: In that case let's leave those as they are!

packages/psss/src/containers/UpcomingReportCard.js
tcaiger: ah yes good catch and even better fix. Much more readable. Thanks

tcaiger:
Thanks again for the great suggestions. I have incorporated them and it's ready for a re-review.

kael89 (approved):
Nice

PR #1760, opened 2020-12-03
1408 psss table expander

tcaiger:
### Issue #: [1408](https://github.com/beyondessential/tupaia-backlog/issues/1408)

This is for a requested design change to how the tables expand.

### Changes:

- add table expansion button to tables in ui components
- use expanding button in psss
- disable in psss where it is not needed by setting to null

### Screenshots
See screenshot in [the issue](https://github.com/beyondessential/tupaia-backlog/issues/1408)


packages/psss/src/store/store.js
```diff
@@ -22,6 +21,4 @@ const enhancers = composeEnhancers(applyMiddleware(thunk));
 
 const store = createStore(persistedReducer, {}, enhancers);
 
-apiErrorHandler(store);
```
tcaiger, line 25: Removing the global 401 handler for now as it introduced some regressions around catching 401s that should be caught elsewhere in the app. I have created another card for it so that it can be done carefully later and tested properly. @see https://app.zenhub.com/workspaces/sprint-board-5eea9d3de8519e0019186490/issues/beyondessential/tupaia-backlog/1748

packages/psss/src/components/Table/TableRow.js
```diff
@@ -2,11 +2,11 @@
...
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
-import styled from 'styled-components';
-import { ExpandableTableRow } from '@tupaia/ui-components';
+import styled, { css } from 'styled-components';
+import { TableRow } from '@tupaia/ui-components';
 import * as COLORS from '../../constants/colors';
 
-export const SimpleTableRow = styled(ExpandableTableRow)`
+export const SimpleTableRowStyles = css`
```
kael89, line 9: Nice style reuse here!!

I believe this is not used elsewhere?

```suggestion
const SimpleTableRowStyles = css`
```

packages/psss/src/components/Table/TableRow.js
```diff
@@ -50,7 +56,13 @@ export const BorderlessTableRow = styled(SimpleTableRow)`
   }
 `;
 
-export const DottedTableRow = styled(SimpleTableRow)`
+export const BorderlessTableRow = styled(SimpleTableRow)`
+  ${BorderlessTableRowStyles}
+`;
+
+export const DottedTableRowStyles = css`
```
kael89, line 63:
```suggestion
const DottedTableRowStyles = css`
```

packages/ui-components/src/components/Table/ExpandableTable.js
```diff
@@ -4,13 +4,24 @@
...
 import { ExpandableTableBody } from './ExpandableTableBody';
 
-export const ExpandableTable = React.memo(({ Body, SubComponent, ...tableProps }) => {
-  const TableBody = tableBodyProps => <Body {...tableBodyProps} SubComponent={SubComponent} />;
-  return <Table Body={TableBody} {...tableProps} />;
+const StyledTable = styled(Table)`
+  th:last-child,
+  td:last-child {
+    padding-right: 65px;
+  }
```
kael89, line 16: Those styles are repeated in `ExpandableTableBody` and `ExpandableTableHeader`. I do realise that sometimes it's better to repeat a styling value rather than over-abstract its definition. Thus, I have not suggestion to make regarding this, just flagging it

packages/ui-components/src/components/Table/TableRow.js
```diff
@@ -22,32 +24,44 @@ export const TableCell = styled(MuiTableCell)`
...
+          key={key}
+          style={{ width, position: 'relative' }}
+          align={align}
+        >
+          {CellComponent ? (
+            <CellComponent {...rowData} displayValue={value} columnKey={key} />
+          ) : (
+            displayValue
+          )}
+          {ExpandButton && index >= columns.length - 1 && ExpandButton}
```
kael89, line 45:
```suggestion
          {index === columns.length - 1 && ExpandButton}
```

kael89 (approved):
Awesome! Great UI styling (like the one you implement) looks almost like doing magic to me. Approving this since all my suggestions are about details really!

packages/psss/src/components/Table/TableRow.js
tcaiger: yeah I only recently found out that you can use styled-components to create css like this. pretty handy.

PR #1627, opened 2020-11-16
1361 add title and export date to exported chart

billli0:
### Issue #:
[1361 - Change to Exported Visuals](https://github.com/beyondessential/tupaia-backlog/issues/1361)

### Changes:

1. Add title at the top and export date at the bottom of exports excel files, whicn includes: 
    - Raw data download 
    - Custom data download 
    - Matrix data exports

2. Refactor the code in `packages/meditrak-server/src/routes/exportSurveyResponses.js`

PR #1767, opened 2020-12-04
1740: Hotfix unable to export survey responses via the admin panel

chris-bes:
### Changes:

- `timeZone` (from browser) query parameter was added into the export survey response meditrak-server endpoint for this ticket (https://github.com/beyondessential/tupaia/pull/1627). It was only done for web-frontend. We should also do it for admin-panel. So this PR is just for sending through timeZone from Admin-Panel when exporting. A default timeZone to fall back was also added to in the endpoint

### DIFF
--- a/packages/admin-panel/src/pages/resources/SurveyResponsesPage.js
+++ b/packages/admin-panel/src/pages/resources/SurveyResponsesPage.js
@@ -6,6 +6,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import moment from 'moment';
+import { getTimeZone } from '../../utilities';
 import { ResourcePage } from './ResourcePage';
 import { SurveyResponsesExportModal } from '../../importExport';
 
@@ -65,6 +66,9 @@ export const SURVEY_RESPONSE_COLUMNS = [
     actionConfig: {
       exportEndpoint: 'surveyResponse',
       fileName: 'Survey Response',
+      extraQueryParameters: {
+        timeZone: getTimeZone(),
+      },
     },
   },
 ];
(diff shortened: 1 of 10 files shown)


IgorNadj (approved):
Caveat - I don't know anything about the original change but this seems to fix the missing tz issue, approving

packages/meditrak-server/src/routes/exportSurveyResponses.js
```diff
@@ -68,7 +68,7 @@ export async function exportSurveyResponses(req, res) {
     latest = false,
     startDate,
     endDate,
-    timeZone,
+    timeZone = 'Australia/Melbourne',
```
kael89, line 71: I'm not sure if we should provide a default here or let it fail. I'd also like @edmofro 's opinion on that, but for now let's use `UTC` as a default since it matches our server's timezone

packages/admin-panel/src/pages/resources/SurveyResponsesPage.js
```diff
@@ -65,6 +66,9 @@ export const SURVEY_RESPONSE_COLUMNS = [
     actionConfig: {
       exportEndpoint: 'surveyResponse',
       fileName: 'Survey Response',
+      extraQueryParameters: {
+        timeZone: getBrowserTimeZone(),
+      },
```
kael89, line 71: Nice and extensible pattern!

kael89 (changes requested):
You are a legend for fixing this @chris-bes !! Just one suggestion and one question

kael89 (approved):

packages/meditrak-server/src/routes/exportSurveyResponses.js
edmofro: In other places we use Auckland as the default, because timezone wise it aligns more closely with Pacific Island Countries we work in than Melbourne. I think UTC is fine here though, because it's more predictable than anything specific, and won't actually be triggered under our current use cases.

Great fix btw @chris-bes! It seems it's about time to put together some more formal regression matrices.

PR #1768, opened 2020-12-04
1723: Hotfix export issue for surveys with an autocomplete field

kael89:
### Issue #:
Addresses https://github.com/beyondessential/tupaia-backlog/issues/1723

### DIFF
--- a/packages/meditrak-server/src/routes/exportSurveys/cellBuilders/AutocompleteConfigCellBuilder.js
+++ b/packages/meditrak-server/src/routes/exportSurveys/cellBuilders/AutocompleteConfigCellBuilder.js
@@ -12,4 +12,8 @@ export class AutocompleteConfigCellBuilder extends KeyValueCellBuilder {
     }
     return value;
   }
+
+  extractRelevantObject({ autocomplete }) {
+    return autocomplete;
+  }
 }
--- a/packages/meditrak-server/src/routes/exportSurveys/cellBuilders/AutocompleteConfigCellBuilder.js
+++ b/packages/meditrak-server/src/routes/exportSurveys/cellBuilders/AutocompleteConfigCellBuilder.js
@@ -12,8 +12,4 @@ export class AutocompleteConfigCellBuilder extends KeyValueCellBuilder {
     }
     return value;
   }
-
-  extractRelevantObject({ autocomplete }) {
-    return autocomplete;
-  }
 }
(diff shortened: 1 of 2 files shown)


nick-beyondessential (approved):

packages/meditrak-server/src/routes/exportSurveys/cellBuilders/AutocompleteConfigCellBuilder.js
```diff
@@ -12,8 +12,4 @@ export class AutocompleteConfigCellBuilder extends KeyValueCellBuilder {
     }
     return value;
   }
-
-  extractRelevantObject({ autocomplete }) {
-    return autocomplete;
-  }
```
chris-bes, line 18: My apology @kael89 , I think I confused you on Friday... The `autocomplete` config was added recently into Autocomplete questions. However it is optional (so by default the config column is `{}`, but if you enable the new feature in Autocomplete questions, the config column would be `{ autocomplete: { createNew: true } }`. So that's why when you searched `config::text like '%autocomplete%'` in the current db you couldn't see any with (because data stewards haven't started to use it yet).

We can change this to the below maybe?
```
extractRelevantObject(object) {
    return object && object.autocomplete ? object.autocomplete : object;
}
```
Sorry for the hassle

kael89: No worries @chris-bes , it makes sense now!

chris-bes (approved):
Looka great!

packages/meditrak-server/src/routes/exportSurveys/cellBuilders/KeyValueCellBuilder.js
```diff
@@ -34,7 +34,7 @@ export class KeyValueCellBuilder {
     }
     const fullObject =
       typeof jsonStringOrObject === 'string' ? JSON.parse(jsonStringOrObject) : jsonStringOrObject;
-    const object = this.extractRelevantObject(fullObject);
+    const object = this.extractRelevantObject(fullObject) || {};
```
kael89, line 37: @chris-bes I changed the behaviour in the parent class to make config for object fields optional. I believe this is a reasonable default.

If a config is required, child classes of `KeyValueCellBuider` could throw an explicit error in their implementation of the `extractRelevantObject` method. Even better, this should be caught up using validation during import (see eg [EntityConfigValidator](https://github.com/beyondessential/tupaia/blob/dev/packages/meditrak-server/src/routes/importSurveys/Validator/ConfigValidator/EntityConfigValidator.js)

kael89:
Changed the target branch to a hotfix branch, merging

PR #1772, opened 2020-12-07
Handle no baseline data differently

Alastair-L:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1751
### Changes:
Don't return `{ data: [] }` when there is no data for a matrix, it requires the keys `columns` and `rows` instead. In this case, I went with not having a different code path for no data, which could cause a bit of regression, but seems a lot cleaner to me? What do you think?

Note: the affected matrix is the only one with this dataBuilder:
image

Screenshots on card


kael89 (approved):
Great fix, using the common code path seems like the way to go here!

There is another hotfix here: https://github.com/beyondessential/tupaia/pull/1768 so let's release those together - I'll create a hotfix branch.

Alastair-L:
Sounds good 👍

kael89:
Changed the target branch to a hotfix branch, merging

PR #1774, opened 2020-12-07
1442 add legend to matrix

IgorNadj:
### Changes:

- Change dots matrix schema around conditionals
- Add legend to dot matricies
- Add legend labels to UNFPA matricies 

---


packages/web-frontend/src/components/View/MatrixWrapper/styles/getStyles.js
```diff
@@ -298,6 +298,20 @@ const getStyles = (isPrintMode, cellCharacters) => {
       position: 'absolute',
       top: 20,
     },
+    footer: {
```
Alastair-L, line 301: This seems like a strange pattern, to calculate the style high in the hierarchy tree rather than passing through the information required to calculate it (in this case none) close to where they are rendered, but given it is the existing pattern it makes sense to use it...

packages/web-frontend/src/components/View/MatrixWrapper/components/FooterRow.js
```diff
@@ -0,0 +1,47 @@
...
+import PropTypes from 'prop-types';
+import { getMarkerForOption } from '../../../Marker';
+import { LEGEND_COLOR_ICON } from '../../../Marker/markerIcons';
+import { CONDITIONAL_MATRIX_CONDITION_SHAPE } from '../../propTypes';
+
+const Legend = ({ legend, styles }) => {
+  return (
+    <div style={styles.footer.legend}>
+      {legend.map(legendItem => {
+        return <LegendItem {...legendItem} key={legendItem.legendLabel} styles={styles} />;
```
Alastair-L, line 16: Sorry, I'm still not completely familiar with react, is this `key` a special property that's needed here, and does it require unique values?

Alastair-L (approved):
LGTM! Great functionality!

packages/web-frontend/src/components/View/MatrixWrapper/styles/getStyles.js
IgorNadj: Yeah was strange to me too, went with following existing code for consistency

packages/web-frontend/src/components/View/MatrixWrapper/components/FooterRow.js
IgorNadj: Ah yep, React wants one for every "list" of elements so it can re-render faster - https://reactjs.org/docs/lists-and-keys.html#keys

PR #1777, opened 2020-12-07
Generalise syndrome data processing

kael89:


packages/psss/src/api/queries/useSingleWeeklyReport.js
```diff
@@ -6,6 +6,14 @@
 import { queryCache } from 'react-query';
 import { useTableData } from './useTableData';
 
+const SYNDROMES = {
+  AFR: 'Acute Fever and Rash (AFR)',
```
kael89, line 10: I used uppercase ids assuming that they are arbitrary. This allows us to more easily construct keys for data returned from the existing reports

tcaiger:
Thank you Kostas this is much better. I like that the syndromes are simply presented at the top as a const and then parsed below. It is both readable and dry.

PR #1778, opened 2020-12-07
Reworked URL params for fetchReport to accept startDate/endDate

rohan-bes:
These params are required to do aggregation correctly for some aggregation types

### DIFF
--- a/packages/report-server/src/aggregator/Aggregator.ts
+++ b/packages/report-server/src/aggregator/Aggregator.ts
@@ -4,20 +4,29 @@
  */
 
 import { Aggregator as BaseAggregator } from '@tupaia/aggregator';
-import { getDefaultPeriod } from '@tupaia/utils';
+import { getDefaultPeriod, convertPeriodStringToDateRange } from '@tupaia/utils';
 import { Event } from '../types';
+
+type PeriodParams = {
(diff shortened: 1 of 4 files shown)


edmofro (approved):
Approved but I have two questions:
- you're sure we want to bother continuing to support `period` as a param into the report-server?
- why does fetchEvents need period if it has startDate and endDate? Does it use it for selected the return period type?

rohan-bes:
> * you're sure we want to bother continuing to support `period` as a param into the report-server?
Honestly, I suspect we may drop `period` but I figured it may be worthy of some discussion before hand, and I didn't want to increase the scope of this PR as it was intended as a quick fix for the indicators not showing up in reports. 
I think it would be good to know if we plan on phasing out `period` entirely throughout `tupaia`. Supporting both start/end date and period can be a bit of a pain so I'm leaning towards that though I know there will be regression risks.

> * why does fetchEvents need period if it has startDate and endDate? Does it use it for selected the return period type?
I really just left `period` in there as I wasn't entirely sure of the possible flow on effects it could have by removing it. It seems as though the period is incorporated into the dhis request.

edmofro:
Cool - could you add point 2 investigation into the tech debt card for psss-server?

PR #1781, opened 2020-12-07
get cors policy origin from env varialbe if it exists

tcaiger:
### Issue:
The cors policy in psss-server needs to specify the allowed origins or else it throws an error.

### Changes:

- Added psss app url as an env variable
- Set env variable in cors policy

---


edmofro (approved):
Nice!

PR #1783, opened 2020-12-08
1664 add percentage change value

tcaiger:
### Changes:

- Added percentage change value to country report table

---

### Screenshots:
See issue

### DIFF
--- a/packages/psss/src/components/Table/CellComponents/PercentageChangeCell.js
+++ b/packages/psss/src/components/Table/CellComponents/PercentageChangeCell.js
@@ -25,7 +25,7 @@ export const PercentageChangeCell = ({ percentageChange, className }) => {
     return '-';
   }
 
-  const roundedValue = Math.round((percentageChange - 1) * 100);
+  const roundedValue = Math.round(percentageChange * 100);
 
   if (roundedValue > 0) {
     return <WarningStyleText className={className}>+{roundedValue}%</WarningStyleText>;
--- a/packages/psss/src/components/Table/CellComponents/PercentageChangeCell.js
+++ b/packages/psss/src/components/Table/CellComponents/PercentageChangeCell.js
@@ -21,11 +21,17 @@ const SuccessStyleText = styled.span`
  * Displays a value as a percentage
  */
 export const PercentageChangeCell = ({ percentageChange, className }) => {
-  if (percentageChange > 0) {
-    return <WarningStyleText className={className}>+{percentageChange}%</WarningStyleText>;
+  if (!percentageChange) {
+    return '-';
   }
 
-  return <SuccessStyleText className={className}>{percentageChange}%</SuccessStyleText>;
+  const roundedValue = Math.round(percentageChange);
+
+  if (percentageChange > 0) {
+    return <WarningStyleText className={className}>+{roundedValue}%</WarningStyleText>;
+  }
+
+  return <SuccessStyleText className={className}>{roundedValue}%</SuccessStyleText>;
 };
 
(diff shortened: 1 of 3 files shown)


tcaiger:
@rohan-bes - I have a couple of questions about the percentage change week on week data.

I assume the calculated values are: `previous week case value / current week case value`. For example to get the WoW value for 2019W52 the calc would be 2019W52/2019W51. Is that correct?

tcaiger:
Secondly I'm not sure if I am mis-reading the data or if maybe there is an error in the calculation but I'm not getting the values that I'm expecting. It's best illustrated through an example. Here is one below.

Screen Shot 2020-12-09 at 9 42 53 AM
Screen Shot 2020-12-09 at 9 42 34 AM

packages/psss/src/components/Table/CellComponents/PercentageChangeCell.js
```diff
@@ -21,11 +21,17 @@ const SuccessStyleText = styled.span`
...
  */
 export const PercentageChangeCell = ({ percentageChange, className }) => {
-  if (percentageChange > 0) {
-    return <WarningStyleText className={className}>+{percentageChange}%</WarningStyleText>;
+  if (!percentageChange) {
+    return '-';
   }
 
-  return <SuccessStyleText className={className}>{percentageChange}%</SuccessStyleText>;
+  const roundedValue = Math.round((percentageChange - 1) * 100);
```
tcaiger, line 28: It would be good if the reviewer can check for me that this calculation is correct. The values in the response from the server are the previous week divided by this week (eg. 1.5 or 0.5) and what I need to display is the change from last week to this week (eg. +50% or -50%).

rohan-bes:
@tcaiger So I believe the actual calculation is:
`siteAverage / siteAveragePrevWeek` where `siteAverage = cases / no. reporting sites`

If you do this the numbers are correct 👍 

However, I do think we ought to be doing the `percentageChange - 1` calculation on the server side. That way all you need to do is multiply by 100 to get a percentage figure 😄

tcaiger:
aha I knew there must be something I was missing. That makes total sense. Thanks for clarifying 🙏 

Cool - let me know when you've added the change `percentageChange - 1` to the server

packages/psss/src/components/Table/CellComponents/PercentageChangeCell.js
```diff
@@ -21,11 +21,17 @@ const SuccessStyleText = styled.span`
  * Displays a value as a percentage
  */
 export const PercentageChangeCell = ({ percentageChange, className }) => {
-  if (percentageChange > 0) {
-    return <WarningStyleText className={className}>+{percentageChange}%</WarningStyleText>;
+  if (!percentageChange) {
+    return '-';
```
rohan-bes, line 25: Just want to check that this means that if there's no percentage change in the data payload then we will display `-`? If so this is correct, since sometimes there is no percentage change (ie. if the previous week had 0 cases, we won't display a percentage change).

However, we may want to differentiate between a valid `0%` percentage change and a non-existant percentage change?

```suggestion
  if (percentageChange === undefined) {
    return '-';
```

rohan-bes (changes requested):
Requested one change, but I feel it's a bit subjective, what do you think?

packages/psss/src/components/Table/CellComponents/PercentageChangeCell.js
tcaiger: Yes - great catch! I totally agree

rohan-bes (approved):
Nice!

PR #1789, opened 2020-12-08
1497: PSSS-Server: Submit and Delete Weekly Survey Responses

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1497:

Added two new endpoints for update/create and delete for unconfirmed survey responses.

@tcaiger Let's define the format that the data is coming in from the front-end?

### DIFF
--- a/packages/psss-server/src/app/addRoutesToApp.ts
+++ b/packages/psss-server/src/app/addRoutesToApp.ts
@@ -14,7 +14,7 @@ import {
   ConfirmedCountryWeeklyReportRoute,
   CountryWeeklyReportRoute,
   SubmitConfirmedCountryWeeklyReportRoute,
-  SubmitCountryWeeklyReportRoute,
+  SaveCountryWeeklyReportRoute,
   DeleteCountryWeeklyReportRoute,
 } from '../routes';
 import { Route } from '../routes/Route';
@@ -69,7 +69,7 @@ export function addRoutesToApp(app: Express) {
   /**
    * PUT routes
    */
-  app.put('/v1/weeklyReport/:organisationUnitCode', handleWith(SubmitCountryWeeklyReportRoute));
+  app.put('/v1/weeklyReport/:organisationUnitCode', handleWith(SaveCountryWeeklyReportRoute));
 
   /**
    * DELETE routes
--- a/packages/psss-server/src/app/addRoutesToApp.ts
+++ b/packages/psss-server/src/app/addRoutesToApp.ts
@@ -14,6 +14,8 @@ import {
   ConfirmedCountryWeeklyReportRoute,
   CountryWeeklyReportRoute,
   SubmitConfirmedCountryWeeklyReportRoute,
+  SubmitCountryWeeklyReportRoute,
+  DeleteCountryWeeklyReportRoute,
 } from '../routes';
 import { Route } from '../routes/Route';
 
@@ -59,14 +61,20 @@ export function addRoutesToApp(app: Express) {
    */
   app.post('/v1/login', handleWith(LoginRoute));
   app.post('/v1/logout', handleWith(LogoutRoute));
+  app.post(
+    '/v1/confirmedWeeklyReport/:organisationUnitCode',
+    handleWith(SubmitConfirmedCountryWeeklyReportRoute),
+  );
 
   /**
    * PUT routes
    */
-  app.put(
-    '/v1/confirmedWeeklyReport/:organisationUnitCode',
-    handleWith(SubmitConfirmedCountryWeeklyReportRoute),
-  );
+  app.put('/v1/weeklyReport/:organisationUnitCode', handleWith(SubmitCountryWeeklyReportRoute));
+
+  /**
+   * DELETE routes
+   */
+  app.delete('/v1/weeklyReport/:organisationUnitCode', handleWith(DeleteCountryWeeklyReportRoute));
 
   app.use(handleError);
 }
--- a/packages/psss-server/src/routes/reports/DeleteCountryWeeklyReportRoute.ts
+++ b/packages/psss-server/src/routes/reports/DeleteCountryWeeklyReportRoute.ts
@@ -1,0 +1,27 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { Route } from '../Route';
+
+const SURVEY_CODE = 'PSSS_WNR';
+
+export class DeleteCountryWeeklyReportRoute extends Route {
+  async buildResponse() {
+    const { week } = this.req.query;
+    const { organisationUnitCode } = this.req.params;
+
+    const existingSurveyResponse = await this.meditrakConnection?.findSurveyResponse(
+      SURVEY_CODE,
+      organisationUnitCode,
+      week,
+    );
+
+    if (existingSurveyResponse) {
+      return this.meditrakConnection?.deleteSurveyResponse(existingSurveyResponse);
+    }
+
+    return 'No existing survey response';
+  }
+}
--- a/packages/psss-server/src/routes/reports/SubmitCountryWeeklyReportRoute.ts
+++ b/packages/psss-server/src/routes/reports/SubmitCountryWeeklyReportRoute.ts
@@ -1,0 +1,73 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { RespondingError } from '@tupaia/utils';
+import { Route } from '../Route';
+
+const SURVEY_CODE = 'PSSS_WNR';
+
+type WeeklyReportAnswers = {
+  PSSS_Sites: number;
+  PSSS_Sites_Reported: number;
+  PSSS_AFR_Cases: number;
+  PSSS_DIA_Cases: number;
+  PSSS_ILI_Cases: number;
+  PSSS_PF_Cases: number;
+  PSSS_DLI_Cases: number;
+};
+
+export class SubmitCountryWeeklyReportRoute extends Route {
+  async buildResponse() {
+    const { week } = this.req.query;
+    const { organisationUnitCode } = this.req.params;
+    const answers = mapReqBodyToAnswers(this.req.body);
+
+    const existingSurveyResponse = await this.meditrakConnection?.findSurveyResponse(
+      SURVEY_CODE,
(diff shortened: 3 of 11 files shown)


packages/psss-server/src/app/addRoutesToApp.ts
```diff
@@ -59,14 +61,20 @@ export function addRoutesToApp(app: Express) {
    */
   app.post('/v1/login', handleWith(LoginRoute));
   app.post('/v1/logout', handleWith(LogoutRoute));
+  app.post(
+    '/v1/confirmedWeeklyReport/:organisationUnitCode',
+    handleWith(SubmitConfirmedCountryWeeklyReportRoute),
+  );
```
rohan-bes, line 67: I decided to move this endpoint to `POST`. Given that we don't actually send any data for that endpoint, and instead we are telling it to 'take the unconfirmed data and submit it as confirmed data' it feels more _POSTy_ to me?

An argument could be made to rename the class as well but I don't feel too strongly about it...

packages/psss-server/src/routes/reports/DeleteCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,27 @@
...
+      SURVEY_CODE,
+      organisationUnitCode,
+      week,
+    );
+
+    if (existingSurveyResponse) {
+      return this.meditrakConnection?.deleteSurveyResponse(existingSurveyResponse);
+    }
+
+    return 'No existing survey response';
```
rohan-bes, line 25: I just returned a string here since this didn't feel like an error case? Though I suppose you could argue it's a `404`...  I'm not too sure either way

edmofro: Yeah I'm not sure, but no error seems fine

packages/psss-server/src/app/addRoutesToApp.ts
edmofro: Re the `POST` move - yep that makes sense.

I actually _do_ feel strongly about the class name! I don't mind what we call each of them, but I feel there's already confusion about the different behaviours of the two "Submit" endpoints. One responds to the "Save" button, the other to the "Confirm" button. In the former case, the user knows they're editing data. In the latter, they feel more like they're just flagging data as confirmed - the fact that it's also upserting a survey response is an implementation detail within our system. Perhaps we can reflect that with a difference in the naming of the two handlers? 

And in writing that out I've talked myself around to realising that the handler names probably _should_ relate more to our internal implementation details. So now I don't feel too strongly either! But I would still lean towards naming this one here as "`ConfirmCountryWeeklyReportRoute`".

packages/psss-server/src/routes/reports/SubmitCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,73 @@
...
+  };
+};
+
+const validateIsNumber = (value: unknown): number => {
+  if (typeof value !== 'number') {
+    throw new RespondingError(`Cannot confirm weekly data: invalid data`, 500);
+  }
+
+  return value;
+};
```
edmofro, line 73: Really feels like this should be shared across the two endpoints, rather than duplicated

Either way: copy paste error on the error thrown, should be `Cannot save weekly report`

packages/psss-server/src/routes/reports/SubmitCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,73 @@
...
+
+type WeeklyReportAnswers = {
+  PSSS_Sites: number;
+  PSSS_Sites_Reported: number;
+  PSSS_AFR_Cases: number;
+  PSSS_DIA_Cases: number;
+  PSSS_ILI_Cases: number;
+  PSSS_PF_Cases: number;
+  PSSS_DLI_Cases: number;
+};
```
edmofro, line 19: And here's the argument against defining them inline. If this was originally in the `types.ts` file, you wouldn't have recreated it here (and I think that would be a good thing)

packages/psss-server/src/routes/reports/SubmitCountryWeeklyReportRoute.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (existingSurveyResponse) {
+      return this.meditrakConnection?.updateSurveyResponse(existingSurveyResponse, answers);
+    }
+
+    return this.meditrakConnection?.createSurveyResponse(
+      SURVEY_CODE,
+      organisationUnitCode,
+      week,
+      answers,
+    );
```
edmofro, line 42: Even most of this could be shared across the two handlers...a parent/child class structure could work well here?

edmofro (changes requested):
Keen to DRY things out, but otherwise looking good

tcaiger:
Hey @rohan-bes, looking good. I'm happy with the format that you've set up here. Looks very convenient from a front end point of view. Just to check my understanding, it is as follows, right?

### Update or Create Weekly Report Data
METHOD: `PUT`
ENDPOINT: `weeklyReport/TO?week=2021W07`
JSON BODY:
```
{
    "AFR": 7,
    "DIA": 3,
    "ILI": 9,
    "PF": 11,
    "DLI": 2,
    "Sites": 13,
    "Sites Reported": 16
}
```
And it will update a record if it exists  and create a new one if if it does not.

packages/psss-server/src/app/addRoutesToApp.ts
rohan-bes: I did actually think about renaming it to `ConfirmCountryWeeklyReportRoute` as well, however that's dangerously close to our existing `ConfirmedCountryWeeklyReportRoute` 🙀 

I think actually that we need a full rename on the routes to be:
`FetchCountryWeeklyReportRoute = GET /weeklyReport/:orgUnitCode` 
`FetchConfirmedCountryWeeklyReportRoute = GET /confirmedWeeklyReport/:orgUnitCode`
`FetchConfirmedWeeklyReportRoute = GET /confirmedWeeklyReport`

`SubmitCountryWeeklyReportRoute = PUT /weeklyReport/:orgUnitCode`
`DeleteCountryWeeklyReportRoute = DELETE /weeklyReport/:orgUnitCode`

`ConfirmCountryWeeklyReportRoute = POST /confirmedWeeklyReport/:orgUnitCode or POST /weeklyReport/:orgUnitCode` thoughts on this?

packages/psss-server/src/routes/reports/SubmitCountryWeeklyReportRoute.ts
rohan-bes: > Hey @rohan-bes, looking good. I'm happy with the format that you've set up here. Looks very convenient from a front end point of view. Just to check my understanding, it is as follows, right?
> 
> ### Update or Create Weekly Report Data
> METHOD: `PUT`
> ENDPOINT: `weeklyReport/TO?week=2021W07`
> JSON BODY:
> 
> ```
> {
>     "AFR": 7,
>     "DIA": 3,
>     "ILI": 9,
>     "PF": 11,
>     "DLI": 2,
>     "Sites": 13,
>     "Sites Reported": 16
> }
> ```
> 
> And it will update a record if it exists and create a new one if if it does not.

@tcaiger Yup, that's the format! However, it's simplest I think if we change the keys to match whatever works best for you on the front-end, eg.
```
{
  "afr": 7,
  "sitesReported": 12,
  ...
}
```
So if you have a preference for that just lemme know 😄

rohan-bes:
> Hey @rohan-bes, looking good. I'm happy with the format that you've set up here. Looks very convenient from a front end point of view. Just to check my understanding, it is as follows, right?
> 
> ### Update or Create Weekly Report Data
> METHOD: `PUT`
> ENDPOINT: `weeklyReport/TO?week=2021W07`
> JSON BODY:
> 
> ```
> {
>     "AFR": 7,
>     "DIA": 3,
>     "ILI": 9,
>     "PF": 11,
>     "DLI": 2,
>     "Sites": 13,
>     "Sites Reported": 16
> }
> ```
> 
> And it will update a record if it exists and create a new one if if it does not.

@tcaiger Yup, that's the format! However, it's simplest I think if we change the keys to match whatever works best for you on the front-end, eg.
```
{
  "afr": 7,
  "sitesReported": 12,
  ...
}
```
So if you have a preference for that just lemme know 😄

packages/psss-server/src/routes/reports/SubmitCountryWeeklyReportRoute.ts
rohan-bes: So it's not quite the same as the other one, all the keys are different, which means I think it'd need to be defined twice anyway?
```
type ConfirmedWeeklyReportAnswers = {
  PSSS_Confirmed_Sites: number;
  PSSS_Confirmed_Sites_Reported: number;
  PSSS_Confirmed_AFR_Cases: number;
  PSSS_Confirmed_DIA_Cases: number;
  PSSS_Confirmed_ILI_Cases: number;
  PSSS_Confirmed_PF_Cases: number;
  PSSS_Confirmed_DLI_Cases: number;
};
```

I'll just confirm things with @tcaiger about the params we are passing in, and then I'll get back to you on the DRY-ness

edmofro: Oh fair! My mistake.

packages/psss-server/src/routes/reports/SubmitCountryWeeklyReportRoute.ts
rohan-bes: Ah good catch on the error string and you make a good point about shared code!

I think ideally we'd have a common `utils` package in typescript as I really want to share the code that exists in the `indicators` package which validates objects (`src/builders/helpers.ts#validateConfig`)

What I might do for now is create a similar function in a `utils` directory within psss-server, and create a techdebt ticket to either consolidate such functions into a `@tupaia/ts-utils` package, or for us to convert `@tupaia/utils` to typescript?

edmofro (approved):
Nice balance on the dryness :-)

rohan-bes:
> Nice balance on the dryness :-)

Thanks! I know that potentially the `mapReqBodyToAnswers` and `mapUnconfirmedReportToAnswers` functions could be abstracted out somehow, however since they have lots of little different things about them (ids/error string/type assertions) I was finding it a bit of a pain to do it cleanly. Hope this is good enough 👍

PR #1800, opened 2020-12-09
Reference expansion data by row data id rather than index

edmofro:


rohan-bes (approved):
Ahh I'd seen this bug before a few times! So this is what was causing it 😂 

Nice fix 👍

PR #1801, opened 2020-12-09
Renamed psss-server endpoint handlers

rohan-bes:
Renamed psss-server endpoint handlers to be a bit more clear in their function


chris-bes (approved):
Looks good!

PR #1803, opened 2020-12-09
Laos Schools new dashboardGroups

nick-beyondessential:
### Changes:
a new migration to:
- create new Laos School dashboardGroups to split up large number of reports
- delete old dashboardGroups
- update project default dashboardGroup

-Still needed:-
corrently all reports use '' but dashboardGroup 'Laos Schools MoES View' used 'Laos Schools Super User', these reports are now mixed with 'Laos Schools' reports. Need to confirm which of the new dashbaordGroups if any should use 'Laos Schools Super User.' Most likely would be 'ICT/Utility Data'; (this will only require some reconfiguration of set up object.)

So reports in 'Laos School Super User' and not in 'Laos School User':
```
select distinct unnest("dashboardReports") from "dashboardGroup" where "userGroup" = 'Laos Schools Super User'
except 
select distinct unnest("dashboardReports") from "dashboardGroup" where "userGroup" = 'Laos Schools User';
```
Laos_Schools_Language_Of_Students
Laos_Schools_Raw_Data_Downloads

and neither are in new groups.


IgorNadj (approved):
LGTM!

edmofro:
Seems this is close enough to release to apply the regular process rather than hotfix now! @BinduMadhaviTammireddy is it close to passing testing?

BinduMadhaviTammireddy:
> Seems this is close enough to release to apply the regular process rather than hotfix now! @BinduMadhaviTammireddy is it close to passing testing?

Yes there were few questions being discussed with Nick once done can mark it testing passed.

edmofro:
Cool - let's just make sure it gets through into dev tonight for tomorrow's release. Thanks guys!

PR #1805, opened 2020-12-09
1696 Fix transition animation meditrak-app

chris-bes:
### Changes:

- Disable transition animation in Welcome/Home screens


packages/meditrak-app/app/navigation/Navigator.js
```diff
@@ -40,14 +40,17 @@ const routes = {
...
+    navigationOptions: () => ({ headerShown: false, animationEnabled: false }),
+  },
   [CREATE_ACCOUNT_SCREEN]: {
     screen: CreateUserContainer,
     navigationOptions: () => ({ headerRight: () => null }),
   },
   [HOME_SCREEN]: {
     screen: HomeScreenContainer,
-    navigationOptions: () => ({ headerLeft: () => null }),
+    navigationOptions: () => ({ headerLeft: () => null, animationEnabled: false }),
```
chris-bes, line 53: I couldn't figure out how in old version of meditrak-app, it does not have any transition animation for these screens... So for our upgrade, I just disable the animation in Welcome/Home screens so that they're just quick-replace instead of default horizontal slide from right.

I've also checked navigating to other screens and compare the transitions with our production app and they seem consistent (at least for me)

edmofro (approved):
Thanks @chris-bes!

PR #1808, opened 2020-12-10
1664 psss confirm endpoint

tcaiger:
### Changes:

- Integrate with confirm endpoint
- Calculate weekly status for reports


packages/psss/src/api/queries/__tests/useCountryWeeklyReport.test.js
```diff
@@ -0,0 +1,38 @@
+/*
```
tcaiger, line 1: There is another card for writing unit tests and so more detail and coverage will be added later.

packages/psss/src/api/mutations.js
```diff
@@ -5,20 +5,27 @@
...
+export const useConfirmWeeklyReport = (orgUnit, period) =>
+  useMutation(
+    () =>
+      post(`confirmedWeeklyReport/${orgUnit}`, {
+        params: { week: period },
+      }),
+    {
+      onSuccess: () => {
+        queryCache.invalidateQueries(`weeklyReport/${orgUnit}`);
+        queryCache.invalidateQueries(`confirmedWeeklyReport/${orgUnit}`);
```
rohan-bes, line 25: Should this also invalidate `confirmedWeeklyReport`?

rohan-bes (changes requested):
Just one requested change, looks good!

packages/psss/src/api/mutations.js
tcaiger: Yes unfortunately it needs to because I use that to calculate the week status. There is a task in the backlog to create an indicator for the status and have it included in the server response which I think is more appropriate in the long term. However I thought it would be better to have something working than not for MVP.

tcaiger: ps. by week status, I mean `overdue`, `resubmit`, `submitted`

rohan-bes: Oh sorry, I meant should this invalidate the query for all countries in addition to invalidating the query for a single country? 
```suggestion
        queryCache.invalidateQueries(`confirmedWeeklyReport/${orgUnit}`);
        queryCache.invalidateQueries(`confirmedWeeklyReport`);        
```

tcaiger: ah I see what you mean now. Yes you are correct - for when the user goes back to the home page. Nice

rohan-bes (approved):
Yay! This is so exciting 😸

PR #1809, opened 2020-12-10
1696: Meditrak app small fixes

chris-bes:
### Issue:
- https://github.com/beyondessential/tupaia-backlog/issues/1779
- https://github.com/beyondessential/tupaia-backlog/issues/1696#issuecomment-742204885

### Changes:
- Aligned Tupaia Header Logo to the left in Home Screen
- Fixed an issue that users cannot change password using meditrak-app

### DIFF
--- a/packages/meditrak-app/app/api/TupaiaApi.js
+++ b/packages/meditrak-app/app/api/TupaiaApi.js
@@ -148,7 +148,7 @@ export class TupaiaApi {
     const response = await this.post(
       CHANGE_USER_PASSWORD_ENDPOINT,
       null,
-      JSON.stringify({ oldPassword, newPassword, newPasswordConfirm }),
+      JSON.stringify({ oldPassword, password: newPassword, passwordConfirm: newPasswordConfirm }),
     );
 
     return response;
--- a/packages/meditrak-app/app/assessment/specificQuestions/Instruction.js
+++ b/packages/meditrak-app/app/assessment/specificQuestions/Instruction.js
@@ -65,7 +65,7 @@ const localStyles = StyleSheet.create({
   textMain: {
     fontSize: THEME_FONT_SIZE_THREE,
     lineHeight: getLineHeight(THEME_FONT_SIZE_THREE),
-    fontWeight: '500',
+    fontWeight: 'bold',
   },
   textSub: {
     fontSize: THEME_FONT_SIZE_ONE,
--- a/packages/meditrak-app/app/home/HomeScreen.js
+++ b/packages/meditrak-app/app/home/HomeScreen.js
@@ -20,6 +20,7 @@ import { HomeToolbarContainer } from './HomeToolbarContainer';
 class HomeScreenComponent extends PureComponent {
   static navigationOptions = {
     headerTitle: () => <TupaiaHeaderLogo />,
+    headerTitleAlign: 'left',
   };
 
   constructor(props) {
--- a/packages/meditrak-app/app/home/HomeScreen.js
+++ b/packages/meditrak-app/app/home/HomeScreen.js
@@ -6,20 +6,15 @@ import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import { Animated, StyleSheet, Text, FlatList, View } from 'react-native';
 
-import {
-  THEME_COLOR_ONE,
-  THEME_TEXT_COLOR_FOUR,
-  DEFAULT_PADDING,
-  getGreyShade,
-} from '../globalStyles';
+import { THEME_TEXT_COLOR_FOUR, DEFAULT_PADDING, getGreyShade } from '../globalStyles';
 import { FeedItemContainer } from '../social';
 import { HomeHeaderContainer } from './HomeHeaderContainer';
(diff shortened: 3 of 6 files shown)


packages/meditrak-app/app/api/TupaiaApi.js
```diff
@@ -148,7 +148,7 @@ export class TupaiaApi {
     const response = await this.post(
       CHANGE_USER_PASSWORD_ENDPOINT,
       null,
-      JSON.stringify({ oldPassword, newPassword, newPasswordConfirm }),
+      JSON.stringify({ oldPassword, password: newPassword, passwordConfirm: newPasswordConfirm }),
```
chris-bes, line 151: The issue was just because the parameter names don't match with the ones in meditrak-server endpoint. Another way to fix this would be to change the parameter names in meditrak-server to match these parameters in meditrak-app. That way we can release the fix tomorrow, but it would require changing other places that use this endpoint too (tupaia, psss, admin-panel,...). So thats why I went with changing it here

edmofro: Can we go with accepting both alternatives in meditrak-server? It's just that even if we change it in the new version of the app, users with older versions of the app will still be unable to change their password.

packages/meditrak-app/app/home/HomeScreen.js
```diff
@@ -20,6 +20,7 @@ import { HomeToolbarContainer } from './HomeToolbarContainer';
 class HomeScreenComponent extends PureComponent {
   static navigationOptions = {
     headerTitle: () => <TupaiaHeaderLogo />,
+    headerTitleAlign: 'left',
```
edmofro, line 23: Well that was easy!

packages/meditrak-app/app/api/TupaiaApi.js
chris-bes: that's a good point!

packages/meditrak-app/app/home/HomeScreen.js
```diff
@@ -6,20 +6,15 @@ import React, { PureComponent } from 'react';
...
 import { FeedItemContainer } from '../social';
 import { HomeHeaderContainer } from './HomeHeaderContainer';
-import { TupaiaHeaderLogo } from '../widgets';
+import { TupaiaLogo } from '../widgets';
 import { HomeToolbarContainer } from './HomeToolbarContainer';
 
 class HomeScreenComponent extends PureComponent {
   static navigationOptions = {
-    headerTitle: () => <TupaiaHeaderLogo />,
+    headerTitle: () => <TupaiaLogo />,
```
chris-bes, line 17: There's a default padding in `TupaiaHeaderLogo` that caused the logo not aligned with other components (screenshots: https://github.com/beyondessential/tupaia-backlog/issues/1696#issuecomment-742261830). So I removed the whole `TupaiaHeaderLogo` component because it only contained `TupaiaLogo` and that extra padding

packages/meditrak-app/app/assessment/specificQuestions/Instruction.js
```diff
@@ -65,7 +65,7 @@ const localStyles = StyleSheet.create({
   textMain: {
     fontSize: THEME_FONT_SIZE_THREE,
     lineHeight: getLineHeight(THEME_FONT_SIZE_THREE),
-    fontWeight: '500',
+    fontWeight: 'bold',
```
chris-bes, line 68: Seems like there's a bug with numeric `fontWeight` in react-native (`https://github.com/facebook/react-native/issues/26193`). `bold` is the only value that works...

packages/meditrak-app/app/home/HomeScreen.js
chris-bes, line 17: There's a default padding in TupaiaHeaderLogo that caused the logo not aligned with other components (screenshots: beyondessential/tupaia-backlog#1696 (comment)). So I removed the whole TupaiaHeaderLogo component because it only contained TupaiaLogo and that extra padding

packages/meditrak-app/app/assessment/specificQuestions/Instruction.js
chris-bes, line 68: Seems like there's a bug with numeric fontWeight in react-native (https://github.com/facebook/react-native/issues/26193). bold is the only value that works...

packages/meditrak-server/src/routes/changePassword.js
```diff
@@ -7,7 +7,19 @@ import { editRecord } from './editRecord';
...
+  const {
+    oneTimeLoginToken,
+    oldPassword,
+    password,
+    newPassword,
+    passwordConfirm,
+    newPasswordConfirm,
+  } = body;
+
+  // Support both alternatives so that users using old versions
```
edmofro, line 19:
```suggestion
  // Support both alternatives so that users using versions of meditrak-app prior to 1.9.109
```
Have found it's always handy to be specific, so that reading over it in future we will know when to deprecate

edmofro (approved):
lgtm! 

I'll test the visual changes (esp header padding) on iOS after merging

edmofro (approved):

PR #1810, opened 2020-12-10
1664 psss save report endpoint

tcaiger:
### Changes:

- Handle saving weekly report data to the server
- Make status cell it's own component and add a resubmit button

### DIFF
--- a/packages/psss/src/containers/Tables/CountryReportTable.js
+++ b/packages/psss/src/containers/Tables/CountryReportTable.js
@@ -2,7 +2,7 @@
  * Tupaia
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
-import React, { useContext, useCallback, useState } from 'react';
+import React, { useContext, useCallback, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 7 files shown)


packages/psss/src/containers/Tables/CountryReportTable.js
```diff
@@ -82,18 +82,25 @@ export const CountryReportTable = React.memo(
...
       setTableStatus(TABLE_STATUSES.SAVING);
 
       try {
         await saveReport({
-          orgUnit: countryCode,
-          period: weekNumber,
-          ...fields,
-          sitesReported: parseInt(sitesReportedValue, 10),
-          totalSites: parseInt(totalSitesValue, 10),
+          afr: parseInt(fields['AFR-totalCases']),
```
tcaiger, line 97: This is not the most beautiful piece of code but it is kind of necessary with the way the data is set up in the editable table. I welcome any suggestions for a more local fix or I could create a tech debt card to look at the editable table data keying. Or it might be ok since we are hardcoding the 5 sydrome ids anyway?

rohan-bes (approved):
Looks good, great work!

PR #1811, opened 2020-12-10
1696 get app building on ios

edmofro:
Get iOS app building

Mostly trial and error and looking up StackOverflow answers, so pretty hard to review!


kael89 (approved):
In Stack Overflow we trust!

PR #1812, opened 2020-12-10
Fix to display Pitcairn

tcaiger:
### Changes:

Pitcairn is not showing up properly in psss because the country code in the database is not the expected standard country code. This is a temporary fix to display the country name and country flag correctly and there is a [new card](https://github.com/beyondessential/tupaia-backlog/issues/1782) in the backlog for fixing the underlying issue.

---


rohan-bes (approved):
Neato! Yup we should get this properly fixed soon tho 👍

PR #1813, opened 2020-12-10
Use correct sign when comparing expiry time

edmofro:


tcaiger (approved):
Looks good to me :croc:

PR #1820, opened 2020-12-11
1763: Restructure arithmetic config utils (1/3)

kael89:
### Issue #:
Towards https://github.com/beyondessential/tupaia-backlog/issues/1763, Part 1/3

This is a pure refactoring, mainly file restructuring and moving a few things around. I did this because I was planning to add new config-related functionality, but after https://github.com/beyondessential/tupaia-backlog/issues/1763#issuecomment-742864036 this will most probably not be needed (new functionality won't be in the config level). Still, I think the following refactoring is valuable

### DIFF
--- a/packages/indicators/src/builders/buildArithmetic/config/types.ts
+++ b/packages/indicators/src/builders/buildArithmetic/config/types.ts
@@ -5,25 +5,33 @@
 
 import { Aggregation, Indicator } from '../../../types';
 
-// Corresponds to a single aggregation
+/**
+ * A single aggregation
+ */
 export type AggregationDescriptor = string | Aggregation;
 
-// Specifies the aggregation(s) that should be applied to a list of elements
+/**
+ * A list of aggregation(s) that will be applied to a data element
+ */
 export type AggregationSpecs = AggregationDescriptor | AggregationDescriptor[];
 
+/**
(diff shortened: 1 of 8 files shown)


packages/indicators/src/builders/buildArithmetic/config/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  readonly aggregation: AggregationSpecs | Record<string, AggregationSpecs>;
+  readonly parameters?: Indicator[];
+  readonly defaultValues?: Record<string, number>;
+};
+
+/**
+ * A fully expanded, verbose config containing as much information as possible.
+ * Useful for internal code purposes
+ */
+export type ExpandedArithmeticConfig = {
```
kael89, line 33: `Expanded` is probably fine but I'd be happy to consider other alternatives (`Verbose`? `Full`?)

I don't know if there is a standard word for this behaviour. I also considered `Sanitised`, but I think it mainly describes the removal of dangerous/unnecessary data

rohan-bes: I'm fine with `Expanded` I think it works well

rohan-bes (approved):
Looks good

PR #1824, opened 2020-12-13
1784 psss reports submitted card

tcaiger:
### Changes:

- Add `ReportsSubmittedCard`
- Calculate number of reports submitted for this week in `useReportsSubmitted.js`

---

### Screenshots:
See Issue

### DIFF
--- a/packages/psss/src/api/queries/useReportsSubmitted.js
+++ b/packages/psss/src/api/queries/useReportsSubmitted.js
@@ -1,0 +1,24 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { getCurrentPeriod } from '@tupaia/utils';
+import { getWeekNumberByPeriod } from '../../utils';
+import { useData } from './useData';
+
+export const useReportsSubmitted = countryCodes => {
+  const period = getCurrentPeriod('WEEK');
+  const currentWeekNumber = getWeekNumberByPeriod(period);
+
+  const query = useData('confirmedWeeklyReport', { params: { startWeek: period } });
+  const sites = countryCodes.length;
+  const sitesReported = query?.data.length;
+
+  return {
+    ...query,
+    currentWeekNumber,
+    sites,
+    sitesReported,
+  };
+};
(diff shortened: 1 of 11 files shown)


packages/psss/src/api/queries/useReportsSubmitted.js
```diff
@@ -0,0 +1,24 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { getCurrentPeriod } from '@tupaia/utils';
+import { getWeekNumberByPeriod } from '../../utils';
+import { useData } from './useData';
+
+export const useReportsSubmitted = countryCodes => {
```
rohan-bes, line 10: Sorry to be a stickler for naming, but I think it may be a bit more clear if we name this like `useConfirmedCountries` or something?

Since it seems we are actually interested in just the no. countries with confirmed data in this endpoint. There will be an endpoint eventually which returns the no. sites which have submitted reports in a country, which I think the `useReportsSubmitted` name fits to better?

packages/psss/src/containers/ReportsSubmittedCard.js
```diff
@@ -0,0 +1,50 @@
...
+import { getEntitiesAllowed } from '../store';
+import { connect } from 'react-redux';
+
+const StyledCardContent = styled(CardContent)`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+`;
+
+export const ReportsSubmittedCardComponent = React.memo(({ countryCodes }) => {
```
rohan-bes, line 22: As above, perhaps this component should be renamed to be a bit more in line with the fact that it displays the confirmed countries?

rohan-bes (changes requested):
Looks really good, just a request to rename things a little 👍

packages/psss/src/api/queries/useReportsSubmitted.js
tcaiger: That's a good point. What do you think about combining my `useReportsSubmitted` into `useConfirmedWeeklyReport`? That way it would be drier but a bit less easy to follow.

rohan-bes: I like that you've broken it out into a separate function actually. Even if they use the same query, I like the returned object here.

rohan-bes (approved):
Nice one :D

PR #1826, opened 2020-12-14
797 convert covidau/flutracker data to internal

nick-beyondessential:
(https://github.com/beyondessential/tupaia-backlog/issues/797) Convert Australian covid and flutracking data to internal

In this PR:
- migration to convert covidau data_sources to tupaia
- update covid/flutracking dashboardReports and mapOverlays dataSourceEntityTypes
- delete dhis_sync_logs &  dhis_sync_queues


IgorNadj (approved):
LGTM

chris-bes (approved):
LGTM!

rohan-bes:
Requesting a re-review on this one for the following commits:

1f284df - Had to update the aggregation logic of a bunch of Covid AU dashboards to use the `FINAL_EACH_DAY` aggregation prior to their regular aggregation as DHIS2's `/fetchAnalytics` was previously just returning the final result each day for us

9fda8c2 - Re-writing the aggregation logic for this databuilder was proving too tricky, so decided to implement an indicator to do the `10000 * (dailyTestCount / statePopulation)` for us. Worth noting that the data we will get back is now slightly different (unfortunately slightly less accurate) as we are now rounding prior to summing each day rather than rounding at the end. Looking at the numbers it doesn't seem to change much, however might be worth confirming this is acceptable with someone...

a938574 - Just like the dashboard changes, but just for a single map overlay 👍

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -38,7 +38,7 @@ const firstExistingValue = (...argumentList) => {
  * the built in functions are also considered as Symbol Node (which is similar to variables).
  * So we want to exclude them when getting the variables.
  */
-const BUILT_IN_FUNCTIONS = ['equalText'];
+const BUILT_IN_FUNCTIONS = ['equalText', 'round'];
```
rohan-bes, line 41: This may prove to grow rather annoying over time... @chris-bes I wonder if we can come up with a long term plan to handle needing to do this?

packages/database/src/migrations/20210610002626-RewriteTotalTestsPerCapitaToUseIndicator-modifies-data.js
```diff
@@ -0,0 +1,113 @@
...
+    aggregation: {
+      dailysurvey006: 'FINAL_EACH_DAY',
+      AU_POP002: 'FINAL_EACH_DAY_FILL_EMPTY_DAYS',
+    },
+  },
+};
+
+const DATA_BUILDER_AND_CONFIG = {
+  oldConfig: {
+    dataBuilder: 'sumPreviousValuesPer100kPerDayByOrgUnit',
```
kael89, line 59: Using an indicator is so much better than having to maintain a specific `sumPreviousValuesPer100kPerDayByOrgUnit` data builder!

kael89 (approved):
I reviewed the new code since the oldest commits have already been reviewed by 2 people - looks great! 🎉 

Let me know if you need any help with the DHIS2 data/metadata cleanup - we can do it together with a Strive cleanup.

kael89 (approved):

billli0:
Will merge to epic branch `epic-flu-tracking-visuals` for testing

PR #1831, opened 2020-12-15
1637 psss use local time

tcaiger:
### Changes:

- Replace `getCurrentPeriod` from `@tupaia/utils` with a new util which gets the current period in local time instead of UTC
- Tweak working in `ConfirmedCountriesCard.js`


packages/psss/src/utils/period.js
```diff
@@ -14,9 +13,11 @@ import { DUE_ISO_DAY } from '../constants';
 
 /**
  *
- * Convenience utils that build on top of @tupaia/utils
+ * Convenience utils that build on top of getCurrentPeriod
```
kael89, line 16: I guess this was an unwanted replacement

```suggestion
 * Convenience utils that build on top of @tupaia/utils
```

kael89 (approved):
Looks great, approving with a small comment suggestion

PR #1833, opened 2020-12-15
1636: Added migration to add defaultValues to threshold level indicators

chris-bes:
### Changes:

- As I understand, no data was actually skipped. The problem was caused because it was returning no data value for `siteAverage1WeekAgo` in the analytic cluster. Because of that, it then skipped the whole analytic cluster and return false for the threshold level because of the highlighted code below:
arith

So adding defaultValue 0 for siteAverage1WeekAgo, siteAverage2WeeksAgo and siteAverage3WeeksAgo fixes the problem.

Eg:

Before:
```
{
    "formula": "2 * ((siteAverage1WeekAgo + siteAverage2WeeksAgo) / 2)",
    "parameters": [
        {
            "code": "siteAverage1WeekAgo",
            "config": {
                "formula": "PSSS_Confirmed_DIA_Site_Average",
                "aggregation": [
                    "FINAL_EACH_WEEK",
                    {
                        "type": "OFFSET_PERIOD",
                        "config": {
                            "offset": 1,
                            "periodType": "week"
                        }
                    }
                ]
            },
            "builder": "arithmetic"
        },
        {
            "code": "siteAverage2WeeksAgo",
            "config": {
                "formula": "PSSS_Confirmed_DIA_Site_Average",
                "aggregation": [
                    "FINAL_EACH_WEEK",
                    {
                        "type": "OFFSET_PERIOD",
                        "config": {
                            "offset": 2,
                            "periodType": "week"
                        }
                    }
                ]
            },
            "builder": "arithmetic"
        }
    ],
    "aggregation": "FINAL_EACH_WEEK",
}
```

After:
```
{
    "formula": "2 * ((siteAverage1WeekAgo + siteAverage2WeeksAgo) / 2)",
    "parameters": [
        {
            "code": "siteAverage1WeekAgo",
            "config": {
                "formula": "PSSS_Confirmed_DIA_Site_Average",
                "aggregation": [
                    "FINAL_EACH_WEEK",
                    {
                        "type": "OFFSET_PERIOD",
                        "config": {
                            "offset": 1,
                            "periodType": "week"
                        }
                    }
                ]
            },
            "builder": "arithmetic"
        },
        {
            "code": "siteAverage2WeeksAgo",
            "config": {
                "formula": "PSSS_Confirmed_DIA_Site_Average",
                "aggregation": [
                    "FINAL_EACH_WEEK",
                    {
                        "type": "OFFSET_PERIOD",
                        "config": {
                            "offset": 2,
                            "periodType": "week"
                        }
                    }
                ]
            },
            "builder": "arithmetic"
        }
    ],
    "aggregation": "FINAL_EACH_WEEK",
    "defaultValues": {
        "siteAverage1WeekAgo": 0,
        "siteAverage2WeeksAgo": 0
    }
}
```

---


kael89 (changes requested):
As discussed, unfortunately `defaultValues` will not work in an "average" context since the denominator needs to be adjusted for each missing value

Eg 
`siteAverage1WeekAgo`: 10
`siteAverage2WeeksAgo`: undefined
Setting `defaultValues` to 0, `(siteAverage1WeekAgo + siteAverage2WeeksAgo) / 2` = `(10 + 0)  / 2` = 5

What we need is `AVG(siteAverage1WeekAgo + siteAverage2WeeksAgo)` = 10 / 1 = 10

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -29,7 +30,10 @@ export class ExpressionParser {
   getVariables(expression) {
     this.validate(expression);
     const nodeTree = this.math.parse(expression);
-    return nodeTree.filter(node => node.isSymbolNode).map(({ name }) => name);
+    const result = nodeTree
+      .filter(node => node.isSymbolNode && !Object.keys(this.customFunctions).includes(node.name))
```
chris-bes, line 34: For some reasons mathjs just treat the custom function as a symbol node so this is just to exclude that when we extract variables from the formula

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -15,11 +15,18 @@ import { create, all } from 'mathjs';
  * //Evaluate an expression that has a as a variable
  * expressionParser.evaluate('a + 2'); //returns 3
  */
+
+const average = (...argumentList) => {
+  const existingValues = argumentList.filter(a => a !== 'undefined');
+  const sum = existingValues.reduce((a, b) => a + b, 0);
+  return sum / existingValues.length;
+};
```
kael89, line 23: Let's add a blank line below this

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -15,11 +15,18 @@ import { create, all } from 'mathjs';
...
 export class ExpressionParser {
   constructor() {
     this.math = create(all, {});
     this.parser = this.math.parser();
-    this.importFunctions();
+    this.customFunctions = this.getCustomFunctions();
```
kael89, line 29: Great pattern!

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -15,11 +15,18 @@ import { create, all } from 'mathjs';
...
+    this.math.import(this.customFunctions);
```
kael89, line 30: Hopefully this doesn't add too much overhead, but because of past [performance concerns](https://github.com/beyondessential/tupaia-backlog/issues/1763), could we take 2-3 measurements to verify it?

Just measuring how long it takes it create a single ExpressionParser instance should be enough:
```js
const a = Date.now();
new ExpressionParser();
console.log(`${Date.now() - a{)ms`);
```

kael89 (approved):
Great solution @chris-bes !!

packages/expression-parser/src/expression-parser/ExpressionParser.js
kael89: Resolved! 👍

chris-bes: Realised that 1s after I hit push :'(

kael89:
@chris-bes we also need to update a specific test case for `defaultValue` validation (if you find that this takes too much time, we can just delete this test for now)

packages/expression-parser/src/expression-parser/ExpressionParser.js
chris-bes: Just had a try, it varies around 19ms to 40ms

kael89: This is acceptable, thanks!

chris-bes:
@kael89 I've fixed the failed test!

kael89 (approved):
Even better!

PR #1834, opened 2020-12-15
1763: Optimise indicator data fetches

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1763

Includes already reviewed PRs:
* https://github.com/beyondessential/tupaia/pull/1820

PR #1835, opened 2020-12-15
Use ExpressionParser instance in indicators

kael89:
It seems that this simple change accounts for a huge performance gain!! Eg when fetching data locally for `PSSS_Weekly_Report` in Tonga between 01/01/2020 - 08/01/2020:

* Before: 4.3s
* After: 0.9 - 1.1s

Please note that data fetching can be optimised even further, down to 0.3s for the above example, using a wider architecture change change which we have started introducing in https://github.com/beyondessential/tupaia/pull/1834. Unfortunately, only towards the end of development I became aware of the quick win presented in this PR!

The small changes in this PR result to a UX which seems pretty smooth and acceptable, in my opinion. We can still decide to merge the bigger changes in https://github.com/beyondessential/tupaia/pull/1834, but we should
* weigh the advantages (even more performance gain in a core module - indicators - and great scalability as the number of fetched elements increase) with the disadvantages (more complicated/specific code)
* take the proper time to review the wider architectural change


packages/psss-server/.env.example
```diff
@@ -1,3 +1,4 @@
+CLIENT_BASIC_AUTH_HEADER=
```
rohan-bes, line 1: What is this needed by? I can't seem to see any reference to it in the code...

packages/indicators/src/getExpressionParserInstance.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * We can safely use a Singleton pattern since the creation of ExpressionParser is not coupled to
+ * specific app data/logic
+ */
+export const getExpressionParserInstance = () => {
+  if (!parserInstance) {
+    parserInstance = new ExpressionParser();
+  }
+  return parserInstance;
+};
```
rohan-bes, line 23: I think this is good for now, but perhaps in the future we may want to implement and singleton pattern on the expression-parser library itself? For instance the report-server creates instances of it for each transform, and an instances is being created here whenever the report-server queries for indicators... I think it would be optimal to have one shared instances across the whole process?

rohan-bes (approved):
Awesome! This is great

packages/indicators/src/getExpressionParserInstance.ts
kael89: I thought of that too, and initially when with that. It reads even nice (`ExpressionParser.getSingleton()`)

But then I thought it is a bit risky to share state management between packages. Eg ExpressionParser.setScope will change the state of the instance, and the client will have to explicitly call `clearScope` afterwards. Or some clients may choose to dynamically add some functions to ExpressionParser (I believe math.js makes this possible). It would be harder to track things if another client in another package messes up with your instance.

And since the initialisation is 20-30ms, doing it once per involved package still shouldn't add too much overhead overall (especially given this will happen once per package + per the lifetime of the node process). And it seems a bit easier to track programmatically. What do you think?

packages/psss-server/.env.example
kael89: I had some issues running the server locally so I thought this was required for CORS, but probably it was another .env variable...

I see this is absent in the server so I'll remove it. Great pickup!

packages/indicators/src/getExpressionParserInstance.ts
rohan-bes: Yeah I'm kind of on the fence about it as well... I'm definitely happy with doing this per-package for now 👍

rohan-bes (approved):

PR #1841, opened 2020-12-16
641 confirm button status

tcaiger:
### Issue #: Fix related to testing on [641](https://app.zenhub.com/workspaces/sprint-board-5eea9d3de8519e0019186490/issues/beyondessential/tupaia-backlog/641)

There was a bug where the Confirm Button state was being set after the form is submitted instead of being calculated from the report data.

### Changes:

- Move calculate week status to utils
- Get confirmed report data (from the cache if available) and use it to calculate status in `useSingleWeeklyReport.js`
- Use status for confirm button state in `weeklyReportsPanel.js`


packages/psss/src/utils/calculateWeekStatus.js
```diff
@@ -0,0 +1,28 @@
+/*
```
tcaiger, line 1: This is unchanged code. It's just been extracted into it's own file.

packages/psss/src/containers/Panels/WeeklyReportsPanel.js
```diff
@@ -118,14 +117,14 @@ export const WeeklyReportsPanelComponent = React.memo(
       isLoading,
       data: countryWeekData,
       syndromes: countrySyndromesData,
-      alerts,
```
tcaiger, line 121: alerts wan't being used as it isn't needed yet so just tidying that up here.

packages/psss/src/utils/calculateWeekStatus.js
kael89: Thanks for clarifying this!

packages/psss/src/api/queries/useSingleWeeklyReport.js
```diff
@@ -55,15 +57,15 @@ const getAlerts = row =>
     [],
   );
 
-export const useSingleWeeklyReport = (orgUnit, period, verifiedStatuses, pageQueryKey) => {
-  const query = useData(
-    `weeklyReport/${orgUnit}`,
+const useCachedQuery = (endpoint, orgUnit, period, queryKey) => {
```
kael89, line 60: Great work extracting functionality in a new method! We may even find out that this can be reused in other queries - but till then, we can keep it local to this file

packages/psss/src/api/queries/useSingleWeeklyReport.js
```diff
@@ -75,36 +77,52 @@ export const useSingleWeeklyReport = (orgUnit, period, verifiedStatuses, pageQue
...
       unVerifiedAlerts: [],
+      reportStatus: REPORT_STATUSES.OVERDUE,
     };
   }
 
-  const data = query.data[0];
-  const alerts = getAlerts(data);
-
+  const report = reportQuery.data[0];
+  const confirmedReport = confirmedReportQuery?.data[0];
```
kael89, line 117: A syntax I find very readable (although it's 100% fine to keep those as they are!)

```suggestion
  const [report] = reportQuery.data;
  const [confirmedReport] = confirmedReportQuery?.data;
```

kael89 (approved):
Great fix and the code looks even better now!

packages/psss/src/api/queries/useSingleWeeklyReport.js
tcaiger: aah fancy - I like it. I often forget about array de-structuring.

PR #1858, opened 2021-01-03
Show value if it is singled in chart tooltip

billli0:
### Issue #:
[1660 - Values on hover wrong for UNFPA visuals Q2 2020 onwards](https://github.com/beyondessential/tupaia-backlog/issues/1660)
### Changes:
If dashboard is a chart, always performs `.MultiValueTooltip()`. `.SingleValueTooltip()` couldn't show single value in chart's tooltip properly.

### DIFF
--- a/packages/web-frontend/src/components/View/ChartWrapper/Tooltip.js
+++ b/packages/web-frontend/src/components/View/ChartWrapper/Tooltip.js
@@ -97,10 +97,7 @@ const SingleValueTooltip = ({ valueType, payload, periodGranularity, labelType }
 function Tooltip(props) {
   const filteredPayload = props.payload.filter(({ value }) => value !== undefined);
   if (props.active && filteredPayload.length >= 1) {
-    if (props.chartType) {
-      return <MultiValueTooltip {...props} payload={filteredPayload} />;
-    }
-    if (filteredPayload.length === 1 && !props.presentationOptions) {
+    if (props.payload.length === 1 && !props.presentationOptions) {
       return <SingleValueTooltip {...props} payload={filteredPayload} />;
     }
     return <MultiValueTooltip {...props} payload={filteredPayload} />;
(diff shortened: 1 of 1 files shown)


packages/web-frontend/src/components/View/ChartWrapper/Tooltip.js
```diff
@@ -97,6 +97,9 @@ const SingleValueTooltip = ({ valueType, payload, periodGranularity, labelType }
 function Tooltip(props) {
   const filteredPayload = props.payload.filter(({ value }) => value !== undefined);
   if (props.active && filteredPayload.length >= 1) {
+    if (props.chartType) {
+      return <MultiValueTooltip {...props} payload={filteredPayload} />;
+    }
     if (filteredPayload.length === 1 && !props.presentationOptions) {
       return <SingleValueTooltip {...props} payload={filteredPayload} />;
     }
```
chris-bes, line 105: Hmmm I think this will just apply `MultiValueTooltip` for all of the charts, and maybe in some cases in some charts we do want to use `SingleValueTooltip`?

Also, I think the root cause of this is there are multiple payloads returned, but some (not all) of them have undefined values (eg, in this case, Tonga: 100%, Fiji: undefined, Vanuatu, undefined,...). So to decide whether to use MultiValueTooltip or SingleValueTooltip, it should also consider `props.payload` additionally along with `filterdPayLoad`. That means if we have filteredPayLoad.length === 1 (meaning there's at least 1 payload that has value), and there are multiple possible payloads (props.payload.length > 1), we want to use MultiValueTooltip, otherwise use SingleValueTooltip

What do you think about the below?


```suggestion
    if (filteredPayload.length === 1 && props.payload.length === 1 && !props.presentationOptions) {
      return <SingleValueTooltip {...props} payload={filteredPayload} />;
    }
```

chris-bes: Or could be simplify to:
```suggestion
    if (props.payload.length === 1 && !props.presentationOptions) {
      return <SingleValueTooltip {...props} payload={filteredPayload} />;
    }
```

billli0: Thanks Chris! I believe you are right. Looks error free now. I thought `prop.payload` will be the same as the response, which only have Tonga data in Q32020. Thanks for correcting me!

chris-bes (approved):
Looks great!

PR #1859, opened 2021-01-05
1781: Exclude all BES internal emails from Meditrak Leaderboard

chris-bes:
### Changes:

- Initially we only wanted to exclude Rohan since he imported a lot of PSSS survey responses. But then I can't think of any case that we want to have any internal tupaia users in Meditrak Leaderboard... So went with excluding all the internal tupaia users

### DIFF
--- a/packages/meditrak-server/src/database/models/UserReward.js
+++ b/packages/meditrak-server/src/database/models/UserReward.js
@@ -56,6 +56,7 @@ export class UserRewardModel extends DatabaseModel {
       "'5ea3984a61f76a61f1011fa7'", // Geoff F
       "'5cee0ba0f013d666713786a9'", // mSupply API Client
     ];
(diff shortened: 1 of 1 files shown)


edmofro (approved):

packages/meditrak-server/src/database/models/UserReward.js
```diff
@@ -56,6 +56,7 @@ export class UserRewardModel extends DatabaseModel {
       "'5ea3984a61f76a61f1011fa7'", // Geoff F
       "'5cee0ba0f013d666713786a9'", // mSupply API Client
```
edmofro, line 57: Could you check if any of the above can be removed, i.e. if they have an internal email?

chris-bes: I can confirm that all of them are not bes emails 👍

PR #1861, opened 2021-01-05
1575 add date selector hpu

billli0:
### Issue #:
[1575 - HPU: Add date selector](https://github.com/beyondessential/tupaia-backlog/issues/1575)

### Changes:

- Add a year date selector
- Disable exporting functions in migration


edmofro (approved):
Thanks for taking away those exports at the same time 🙏

PR #1874, opened 2021-01-12
1373: Allow loading more results in Search Bar

chris-bes:
### Changes:

- Implemented Load more results in search org unit bar


rohan-bes (approved):
Nice work Chris, looks good!

PR #1878, opened 2021-01-12
fix data elements and data groups caching

tcaiger:
### Issue #: [2093 Fix Admin Panel Caching Issue](https://github.com/beyondessential/tupaia-backlog/issues/2093)

The issue was that data elements and data groups were displaying the same data. The underlying issue was that the endpoint in Resource Page is used as the id for keying the table data in the redux store. Since these 2 pages use the same endpoint they were sharing a record in the store.

### Changes:

- add optional reduxId prop to resourcePage
- set a reduxId for dataElements and dataGroups


kael89 (approved):
Nice fix!

PR #1879, opened 2021-01-13
1621 samoa confirmed and suspected cases

billli0:
### Issue #:
[1621: Map Overlay: Heatmap - Home village of suspected and confirmed cases](https://github.com/beyondessential/tupaia-backlog/issues/1621)

### Changes:

- Add heatmap map overlay by migration


packages/database/src/migrations/20210106000258-HeatmapOfVillageConfirmedAndSuspectedCases-modifies-data.js
```diff
@@ -0,0 +1,85 @@
...
+
+const mapOverlay = {
+  id: 'COVID_WS_Home_Village_Of_Suspected_And_Confirmed_Cases',
+  name: 'Home village of suspected and confirmed cases',
+  userGroup: 'Donor',
+  dataElementCode: 'value',
+  measureBuilderConfig: {
+    dataValues: {
+      WS_COVID19_Clinsurv_11: '*',
+    },
```
kael89, line 45: I have the impression that this is not required, since we don't need to filter based on any value, just count all responses submitted in a village. The card's requirements state

> Count of cases by WS_COVID19_Clinsurv_11 (Current village)

I believe that in practice this means "count all responses in a village (which is captured via `WS_COVID19_Clinsurv_11`)"

No need to change this though, just for your reference!

```suggestion
```

kael89 (approved):
Looks great @billli0 , approved!

packages/database/src/migrations/20210106000258-HeatmapOfVillageConfirmedAndSuspectedCases-modifies-data.js
billli0: Yes Kostas, you are right. Here the hierachy of entities is `case -> village -> facility -> country`, and we are counting all the `case` (which is `WS_COVID19_Clinsurv_11`) in a village, which are 'confirmed' and 'suspected' cases.

PR #1880, opened 2021-01-13
Override default join for GETSurveys to use left outer

EMcQ-BES:
### Changes:
- Add ability to define default joins for GETHandlers, override on GETSurveys to match behaviour on dev

### DIFF
--- a/packages/meditrak-server/src/routes/GETHandler/helpers.js
+++ b/packages/meditrak-server/src/routes/GETHandler/helpers.js
@@ -83,7 +83,7 @@ export const getQueryOptionsForColumns = (
   columnNames,
   baseRecordType,
   customJoinConditions = {},
-  defaultJoinType = null,
+  joinType = null,
 ) => {
   if (columnNames.some(c => c.startsWith('_'))) {
     throw new ValidationError(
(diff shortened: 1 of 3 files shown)


packages/meditrak-server/src/routes/GETHandler/helpers.js
```diff
@@ -83,6 +83,7 @@ export const getQueryOptionsForColumns = (
   columnNames,
   baseRecordType,
   customJoinConditions = {},
+  defaultJoinType = null,
```
edmofro, line 86:
```suggestion
  joinType = null,
```
At this point it's probably just the join type

packages/meditrak-server/src/routes/GETHandler/helpers.js
```diff
@@ -103,7 +104,7 @@ export const getQueryOptionsForColumns = (
           `${recordType}.id`,
           getForeignKeyColumnName(recordType),
         ];
-        multiJoin.push({ joinWith: recordType, joinCondition });
+        multiJoin.push({ joinWith: recordType, joinCondition, joinType: defaultJoinType });
```
edmofro, line 107: So if `joinType` is null, it will use the original default (which I guess is inner)?

edmofro (approved):
Looks good, two q's, but approved anyway

packages/meditrak-server/src/routes/GETHandler/helpers.js
EMcQ-BES: Yep! Made sure to double check that

PR #1881, opened 2021-01-13
Bug fixes:

EMcQ-BES:
### Changes:
- Fix demo land exception in permissions handling for userEntityPermissions
- Include sub-entities in admin panel permissions util
- Check the content of user entity permission when deleting
- Fix up some unit tests to run more consistently


packages/meditrak-server/src/routes/utilities/getAdminPanelAllowedEntityIds.js
```diff
@@ -6,15 +6,14 @@
...
  *
  * @param {AccessPolicy}  accessPolicy
- * @param {ModelRegistry} models
  *
- * @returns string[] The entity ids
+ * @returns string[] The country codes
  */
 
-export const getAdminPanelAllowedEntityIds = async (accessPolicy, models) => {
+export const getAdminPanelAllowedCountryCodes = async accessPolicy => {
```
edmofro, line 16: I don't think this one needs to be async any more

packages/meditrak-server/src/routes/accessRequests/assertAccessRequestPermissions.js
```diff
@@ -17,10 +17,11 @@ export const assertAccessRequestPermissions = async (accessPolicy, models, acces
...
 
   const entity = await models.entity.findById(accessRequest.entity_id);
-  if (!accessPolicy.allows(entity.country_code, TUPAIA_ADMIN_PANEL_PERMISSION_GROUP)) {
-    throw new Error('Need Admin Panel access to the country this access request is for');
+  const accessibleCountryCodes = await getAdminPanelAllowedCountryCodes(accessPolicy);
+  if (accessibleCountryCodes.includes(entity.country_code)) {
+    return true;
   }
-  return true;
+  throw new Error('Need Admin Panel access to the country this access request is for');
```
edmofro, line 24: I reckon I prefer it the old way. No logical difference right?

edmofro: Just for the slight optimisation that would give us

edmofro: Oh right...`DL`. Cool, this makes sense then. 

Could you please make sure the testers are doing performance testing as well? Not that this would make the difference - but the multiple db calls that the endpoint handlers now go through might. @Klaus-S

edmofro (approved):
Just the `async`, but that's a nit so this is approved :-)

PR #1882, opened 2021-01-13
Pluralise disaster import so the endpoint runs correctly

EMcQ-BES:
### Changes:

- Pluralise disaster import endpoint used by admin panel
- Add CUD handlers for dataSources


packages/meditrak-server/src/app/addRoutesToApp.js
```diff
@@ -250,6 +254,7 @@ export function addRoutesToApp(app) {
   app.put('(/v[0-9]+)/surveyScreenComponents/:recordId', editSurveyScreenComponents);
   app.put('(/v[0-9]+)/answers/:recordId', editAnswers);
   app.put('(/v[0-9]+)/surveyResponses/:parentRecordId/answers/:recordId', editAnswers);
+  app.put('(/v[0-9]+)/dataSources/:recordId', editDataSources);
```
edmofro, line 257: Does this work for editing data elements within the data groups dropdown (if that's possible)? The read uses something like dataSources/:dataGroupId/dataSources

packages/admin-panel/src/pages/resources/SocialFeedPage.js
```diff
@@ -102,7 +102,7 @@ const EDIT_CONFIG = {
 const CREATE_CONFIG = {
   title: 'Add Social Feed item',
   actionConfig: {
-    editEndpoint: 'feedItem',
+    editEndpoint: 'feedItems',
```
edmofro, line 105: This isn't mentioned in the description, but seems legit :+1:

edmofro (approved):

packages/meditrak-server/src/app/addRoutesToApp.js
EMcQ-BES: You can edit and delete via the dropdown, but those actions don't use a parent/child endpoint like the get does so this works for them

edmofro: Grand

PR #1883, opened 2021-01-13
Pluralise export endpoints

EMcQ-BES:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/999
https://github.com/beyondessential/tupaia-backlog/issues/998

### Changes:

- We already pluralised these in the admin panel but forgot to pluralise in the addRoutesToApp.js file

### DIFF
--- a/packages/meditrak-server/src/app/addRoutesToApp.js
+++ b/packages/meditrak-server/src/app/addRoutesToApp.js
@@ -152,9 +152,9 @@ export function addRoutesToApp(app) {
    */
   app.get('(/v[0-9]+)?/changes/count', countChanges);
   app.get('(/v[0-9]+)/export/surveyResponses', exportSurveyResponses);
-  app.get('(/v[0-9]+)/export/surveyResponse/:surveyResponseId', exportSurveyResponses);
+  app.get('(/v[0-9]+)/export/surveyResponses/:surveyResponseId', exportSurveyResponses);
   app.get('(/v[0-9]+)/export/surveys', exportSurveys);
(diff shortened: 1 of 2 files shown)


packages/meditrak-server/src/app/addRoutesToApp.js
```diff
@@ -152,9 +152,9 @@ export function addRoutesToApp(app) {
    */
   app.get('(/v[0-9]+)?/changes/count', countChanges);
   app.get('(/v[0-9]+)/export/surveyResponses', exportSurveyResponses);
-  app.get('(/v[0-9]+)/export/surveyResponse/:surveyResponseId', exportSurveyResponses);
+  app.get('(/v[0-9]+)/export/surveyResponses/:surveyResponseId', exportSurveyResponses);
```
edmofro, line 155: We may use this endpoint in web-config-server as part of the tupaia.org raw data downloads system - could you check?

edmofro (changes requested):

packages/meditrak-server/src/app/addRoutesToApp.js
EMcQ-BES: Did a grep through web-config-server and all references use the batch `/export/surveyResponses` endpoint (which has always been plural) not the individual `/export/surveyResponses/:id` endpoint

edmofro: Great! Thanks for checking

edmofro (approved):

PR #1888, opened 2021-01-14
Add ValueAndPercentageByDataValueByFlightDate

nick-beyondessential:
### Changes:
- Generalise databuilder `ValueAndPercentageByAgeByFlightDate` as `ValueAndPercentageByDataValueByFlightDate` and extend the latter class by the former. 
- Create migration to add Covid Samoa Demographics of individuals in quarantine: Health Conditions report
- export migration delete utilities
- add getPassengersPerDataValue test

### DIFF
--- a/packages/web-config-server/src/apiV1/dataBuilders/modules/covid-samoa/valueAndPercentageByDataValueByFlightDate.js
+++ b/packages/web-config-server/src/apiV1/dataBuilders/modules/covid-samoa/valueAndPercentageByDataValueByFlightDate.js
@@ -64,9 +64,9 @@ export class ValueAndPercentageByDataValueByFlightDate extends DataBuilder {
           const passengersForThisDataKey = getPassengersPerDataValue(flight, cells)[dataKey];
           
           row[column.key] = {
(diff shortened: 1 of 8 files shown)


packages/web-config-server/src/apiV1/dataBuilders/modules/covid-samoa/valueAndPercentageByDataValueByFlightDate.js
```diff
@@ -0,0 +1,107 @@
...
+  getTotalNumPassengers,
+  getPassengersPerDataValue,
+  FLIGHT_DATE,
+} from './flight';
+
+export class ValueAndPercentageByDataValueByFlightDate extends DataBuilder {
+  async build() {  
+    const dataElementCodes = this.getDataElementCodes();
+
+    const events = await this.fetchEvents({ useDeprecatedApi: false, dataElementCodes });
```
rohan-bes, line 18: Gosh we should really properly deprecate this api so we don't have to keep specifying otherwise huh? 😝

rohan-bes (approved):
Nice, looks good.

PR #1893, opened 2021-01-14
redirect to home on logout

tcaiger:
### Changes:

- Redirect to home on logout


edmofro (approved):
Approving this as a useful and good change, but also interested in whether we can clear the reset password state earlier as well

PR #1895, opened 2021-01-15
2123: Couple of performance improvements for report-server and expression-parser

rohan-bes:
### Changes:
- Rewrote keyValueBy aliases to perform transform inline rather than calling a select transform
- Begun caching validated expressions on the expression parser to avoid duplicating this call

### Results:
45,000 row report:
- ~5.5 seconds => ~0.2 seconds


packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -56,6 +61,7 @@ export class ExpressionParser {
         );
       }
     });
+    this.validExpressionCache.add(expression);
```
rohan-bes, line 64: I believe always caching at this point should be safe, as we ought to throw and exception already if the expression is invalid

edmofro (approved):
Side benefit - I find the map operations a fair bit easier to understand!

packages/report-server/src/reportBuilder/transform/aliases/keyValueByFieldAliases.ts
```diff
@@ -3,31 +3,37 @@
...
+/**
+ * { period: '20200101', organisationUnit: 'TO', dataElement: 'CH01', value: 7 }
+ *  => { period: '20200101', organisationUnit: 'TO', CH01: 7 }
+ */
+export const keyValueByDataElementName = () => (rows: Row[]) => {
+  return rows.map(row => {
+    const { dataElement, value, ...restOfRow } = row;
+    return { [dataElement as string]: value, ...restOfRow };
+  });
+};
```
rohan-bes, line 17: Switching from the `select` transform to this inline function yields about a 50x speed improvement (2.2 seconds => 0.042 seconds). While this is a highly optimised function, it does somewhat raise concerns around the ongoing performance of the `select` transform on large datasets...

packages/report-server/src/reportBuilder/transform/transform.ts
```diff
@@ -8,6 +8,7 @@ import { transformBuilders } from './functions';
 import { aliases } from './aliases';
 
 type TransformParams = {
+  name: string;
```
rohan-bes, line 11: This isn't actually doing anything, but is quite handy for debugging purposes so I left it in 👍

rohan-bes:
> Side benefit - I find the map operations a fair bit easier to understand!

@edmofro yeah me too! Tbh I'm considering scrapping the whole LHS evaluation of the `select` field names all together, as I find it:
a) Confusing to read and inconsistent with other configs in the report-server
b) Rare that it's actually useful (the `keyValueBy...` transforms were the only ones using it and as of now they no longer use it)
c) Both makes the config and performance of the most common case (where you just want a user defined string as the key) worse

It was a cool idea with some cute applications, but I think it may be wise to remove it, and if we feel the need we can re-add it back in the future. Thoughts?

edmofro:
Oh yeah, I'm all for removing stuff 😁 

As you say, if we think we need it in future, we'll add it back in at that point!

PR #1899, opened 2021-01-15
1798: Use @tupaia/utils in web-frontend

chris-bes:
### Changes:

After more than a day struggling with this, I've finally realized that it's always been working. I just didnt run web-frontend on wsl in my windows machine... If running on wsl or macOS, you can import internal packages in web-frontend without any problems.

So I took this opportunity to might as well remove the duplicated getBrowserTimezone util in web-frontend and use the one in @tupaia/utils


edmofro (approved):
Sorry you had so much time struggling with this! As a mac user I'm oblivious to what parts do and don't work easily for Windows users. Would be good to share your experience in dev chat for other Windows users.

Nice work cleaning up the duplication

PR #1901, opened 2021-01-17
If image already on S3, log error rather than blocking sync

edmofro:

### DIFF
--- a/packages/meditrak-server/src/dataAccessors/addSurveyImage.js
+++ b/packages/meditrak-server/src/dataAccessors/addSurveyImage.js
@@ -1,11 +1,16 @@
 /**
- * Tupaia MediTrak
- * Copyright (c) 2018 Beyond Essential Systems Pty Ltd
- * This uploads a surveyImage to s3
+ * Tupaia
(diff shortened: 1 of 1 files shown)


kael89 (approved):
That will do!

Idea (for another day): the sync process could have a specific error handling logic, which chooses which errors should block it and which not. With this in place we could still throw the error here, keeping the error handling concern out of this util.

edmofro:
That's a good idea...

Note that the function I put it inside is not the common util that handles the s3 upload, but just the handy function used by the sync endpoint (`/changes`) handler, so it kind of operates in the way you said above, but just not as a generic pattern. A side note off that - the handler helper is probably in the wrong directory!

edmofro:
I have tested this locally fairly thoroughly as part of an investigation into the effect of timeouts, so feel comfortable that it can be merged without independent testing (especially given it is a hot fix and a simple change). What do you think @kael89?

kael89:
> I have tested this locally fairly thoroughly as part of an investigation into the effect of timeouts, so feel comfortable that it can be merged without independent testing (especially given it is a hot fix and a simple change). What do you think @kael89?

Sounds great @edmofro , probably this is the most pragmatic approach for testing changes like this! Go for it

PR #1905, opened 2021-01-17
1622 htmap samoa cnfrmed cases

billli0:
### Issue #:
[1622 - Map Overlay: Heatmap - Home village of confirmed cases](https://github.com/beyondessential/tupaia-backlog/issues/1622)

### Changes:

- In `dataBroker`, let `.pullMetadata()` return empty if `service_type` is `indicator`, because `indicator` service is not support `.pullMetadata()`

### DIFF
--- a/packages/data-broker/src/DataBroker.js
+++ b/packages/data-broker/src/DataBroker.js
@@ -114,6 +114,7 @@ export class DataBroker {
       throw new Error('Cannot pull metadata for data sources belonging to different services');
     }
     const serviceType = dataSources[0].service_type;
+    // `indicator` service is not support `.pullMetadata()`
     if (serviceType === 'indicator') return [{ code: dataSources[0].code, name: undefined }];
     const service = this.createService(serviceType);
     // `dataSourceSpec` is defined  for a single `type`
--- a/packages/data-broker/src/DataBroker.js
+++ b/packages/data-broker/src/DataBroker.js
@@ -113,7 +113,9 @@ export class DataBroker {
     if (countDistinct(dataSources, 'service_type') > 1) {
       throw new Error('Cannot pull metadata for data sources belonging to different services');
     }
-    const service = this.createService(dataSources[0].service_type);
+    const serviceType = dataSources[0].service_type;
+    if (serviceType === 'indicator') return [{ code: dataSources[0].code, name: undefined }];
+    const service = this.createService(serviceType);
     // `dataSourceSpec` is defined  for a single `type`
     const { type } = dataSources[0];
     return service.pullMetadata(dataSources, type, options);
--- a/packages/database/src/migrations/20210113032005-SamoaHeatmapHomeVillageOfConfirmedCases-modifies-data.js
+++ b/packages/database/src/migrations/20210113032005-SamoaHeatmapHomeVillageOfConfirmedCases-modifies-data.js
@@ -1,0 +1,98 @@
+'use strict';
+
+import { insertObject, generateId, codeToId } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+async function deleteItemByOneCondition(db, table, condition) {
+  const [key, value] = Object.entries(condition)[0];
+  return db.runSql(`
+    DELETE FROM
+      "${table}"
+    WHERE
+      "${key}" = '${value}';
+  `);
+}
+const indicator = {
+  id: generateId(),
+  code: 'COVID_SAMOA_HEATMAP_CONFIRMED_CASES',
+  builder: 'arithmetic',
+  config: {
+    formula:
+      '(WS_COVID19_Clinsurv_75 == 0 | WS_COVID19_Clinsurv_137 == 1 | WS_COVID19_Clinsurv_79 == 0 | (WS_COVID19_Clinsurv_83 != 3 & WS_COVID19_Clinsurv_83 != 4) | WS_COVID19_Clinsurv_87 == 0) ? 1 : 0',
+    aggregation: 'COUNT_PER_ORG_GROUP',
+    defaultValues: {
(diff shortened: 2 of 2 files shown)


packages/database/src/migrations/20210113032005-SamoaHeatmapHomeVillageOfConfirmedCases-modifies-data.js
```diff
@@ -0,0 +1,98 @@
...
+  config: {
+    formula:
+      '(WS_COVID19_Clinsurv_75 == 0 | WS_COVID19_Clinsurv_137 == 1 | WS_COVID19_Clinsurv_79 == 0 | (WS_COVID19_Clinsurv_83 != 3 & WS_COVID19_Clinsurv_83 != 4) | WS_COVID19_Clinsurv_87 == 0) ? 1 : 0',
+    aggregation: 'COUNT_PER_ORG_GROUP',
+    defaultValues: {
+      WS_COVID19_Clinsurv_75: 2, // 3. Pending
+      WS_COVID19_Clinsurv_79: 2, // 3. Pending
+      WS_COVID19_Clinsurv_83: 4, // 5. Pending
+      WS_COVID19_Clinsurv_137: 0, // No
+      WS_COVID19_Clinsurv_87: 2, // 3. Pending
```
rohan-bes, line 41: Nice comments here, very handy 😄

packages/database/src/migrations/20210113032005-SamoaHeatmapHomeVillageOfConfirmedCases-modifies-data.js
```diff
@@ -0,0 +1,98 @@
...
+      "${key}" = '${value}';
+  `);
+}
+const indicator = {
+  id: generateId(),
+  code: 'COVID_SAMOA_HEATMAP_CONFIRMED_CASES',
+  builder: 'arithmetic',
+  config: {
+    formula:
+      '(WS_COVID19_Clinsurv_75 == 0 | WS_COVID19_Clinsurv_137 == 1 | WS_COVID19_Clinsurv_79 == 0 | (WS_COVID19_Clinsurv_83 != 3 & WS_COVID19_Clinsurv_83 != 4) | WS_COVID19_Clinsurv_87 == 0) ? 1 : 0',
```
rohan-bes, line 34: Are you sure doing an `|` is the right kind of `or` here? According to the mathjs docs this is a bitwise or, which I believe resolves to a number rather than a `true/false` value. I suspect it will work fine with this conditional, but perhaps it may be safer and read better using a logic or, eg:
```
(WS_COVID19_Clinsurv_75 == 0 OR WS_COVID19_Clinsurv_137 == 1 OR WS_COVID19_Clinsurv_79 == 0 OR (WS_COVID19_Clinsurv_83 != 3 AND WS_COVID19_Clinsurv_83 != 4) OR WS_COVID19_Clinsurv_87 == 0) ? 1 : 0
```

I'm okay with the bitwise or if it works though, especially if the performance is better, I just think it might be risky if someone copied this without realising

packages/data-broker/src/DataBroker.js
```diff
@@ -113,7 +113,10 @@ export class DataBroker {
     if (countDistinct(dataSources, 'service_type') > 1) {
       throw new Error('Cannot pull metadata for data sources belonging to different services');
     }
-    const service = this.createService(dataSources[0].service_type);
+    const serviceType = dataSources[0].service_type;
+    // `indicator` service is not support `.pullMetadata()`
+    if (serviceType === 'indicator') return [{ code: dataSources[0].code, name: undefined }];
```
rohan-bes, line 118: I think if we are going to do this, we might as well just implement `pullMetadata()` for indicators, no?

I think what you've got here should suffice for now, so if you just edit `pullMetadata()` in `data-broker/src/services/indicator/IndicatorService.js` to be:
```
  async pullMetadata(dataSources) {
    return dataSources.map(dataSource => ({ code: dataSource.code, name: undefined }));
  }
```

Ultimately we can properly support this interface just by adding a `name` field to the indicators table, and then fetching the `code` and `name` when `pullMetadata()` is called. But I rekon that can be done in separate ticket (I've raised https://github.com/beyondessential/tupaia-backlog/issues/2137 for this)

rohan-bes (changes requested):
Nice one, good job creating your first indicator 🎉 

Just a request to maybe switch to logical or if the performance is fine, and also to relocate that code which handles the `pullMetadata()` for indicators

packages/database/src/migrations/20210113032005-SamoaHeatmapHomeVillageOfConfirmedCases-modifies-data.js
billli0: Thanks Rohan, for performance purpose I agree to switch to 'OR'.

packages/database/src/migrations/20210113032005-SamoaHeatmapHomeVillageOfConfirmedCases-modifies-data.js
```diff
@@ -0,0 +1,98 @@
...
+      "${key}" = '${value}';
+  `);
+}
+const indicator = {
+  id: generateId(),
+  code: 'COVID_SAMOA_HEATMAP_CONFIRMED_CASES',
+  builder: 'arithmetic',
+  config: {
+    formula:
+      '(WS_COVID19_Clinsurv_75 == 0 OR WS_COVID19_Clinsurv_137 == 1 OR WS_COVID19_Clinsurv_79 == 0 OR (WS_COVID19_Clinsurv_83 != 3 & WS_COVID19_Clinsurv_83 != 4) OR WS_COVID19_Clinsurv_87 == 0) ? 1 : 0',
```
rohan-bes, line 34: Nice, reads a bit easier now too :) Just gotta also change that `&` to `AND`?

rohan-bes (changes requested):

rohan-bes (approved):
Great! Nice work 😄

billli0:
Thanks Rohan!

billli0:
switching '|' to 'OR' will fail the indicator function. I guess it is related to this [card](https://github.com/beyondessential/tupaia-backlog/issues/1379)?

> Error: Error in field 'defaultValues': Invalid syntax: Implicit multiplication found. Perhaps you used a variable starting with a number?


@rohan-bes Would you like to have a look?

PR #1908, opened 2021-01-18
1612,1615,1616 DHIS mapping code to id

IgorNadj:
### Changes:

- [1612](https://github.com/beyondessential/tupaia-backlog/issues/1612) Mapping id to code - org unit/entity
- [1615](https://github.com/beyondessential/tupaia-backlog/issues/1615) Mapping id to code - data element
- [1616](https://github.com/beyondessential/tupaia-backlog/issues/1616) Mapping id to code - program code

---

Test config:
```
INSERT INTO data_service_entity (id, entity_code, config) VALUES (generate_object_id(), 'TO', '{ "dhis_id": "jSk4k1gnSqk" }');
UPDATE data_source SET config = '{"dhisId": "OmssK27WEJB", "isDataRegional": false}' where code = 'HP02';
UPDATE data_source SET config = '{"dhisId": "OzDb71SXJgP", "isDataRegional": true}' where code = 'DP9';
UPDATE data_source SET config = '{"dhisId": "K3wJsRJHMCa", "isDataRegional": true}' where code = 'SS128';
UPDATE data_source SET config = '{"dhisId": "pDhCn9lbi5v", "isDataRegional": true}' where code = 'SS162';
UPDATE data_source SET config = '{"dhisId": "g2KqfrYXLdM", "isDataRegional": true}' where code = 'SS182';
UPDATE data_source SET config = '{"dhisId": "iYzYjNzDnAY", "isDataRegional": true}' where code = 'SS190A';
UPDATE data_source SET config = '{"dhisId": "RngB7fLGI3i", "isDataRegional": true}' where code = 'SS192';
UPDATE data_source SET config = '{"dhisId": "bcg0ApKZTav", "isDataRegional": true}' where code = 'SS219';
```

Test 1:
- Analytics query, with data element code that does not exist in data_source.
- request is not modified.
- dashboard: http://localhost:8000/api/v1/view?dashboardGroupId=228&endDate=2021-01-31&isExpanded=true&organisationUnitCode=TO&projectCode=fanafana&startDate=2015-01-01&timeZone=Australia%2FMelbourne&viewId=12&cacheBreaker=h3iudn
- dhis request: https://dev-aggregation.tupaia.org/api/analytics/rawData.json?paging=false&inputIdScheme=code&outputIdScheme=code&includeMetadataDetails=true&dimension=dx:PercentageCriticalMedicinesAvailable&dimension=ou:TO&dimension=co&dimension=pe:201501;201502;201503;201504;201505;201506;201507;201508;201509;201510;201511;201512;201601;201602;201603;201604;201605;201606;201607;201608;201609;201610;201611;201612;201701;201702;201703;201704;201705;201706;201707;201708;201709;201710;201711;201712;201801;201802;201803;201804;201805;201806;201807;201808;201809;201810;201811;201812;201901;201902;201903;201904;201905;201906;201907;201908;201909;201910;201911;201912;202001;202002;202003;202004;202005;202006;202007;202008;202009;202010;202011;202012;202101

Test 2:
- Another analytics query
- request is modified to swap dataElementCodes (SS128, SS192...) to ids, and org unit TO to id.
- dashboard: http://localhost:8000/api/v1/view?dashboardGroupId=227&endDate=2021-01-18&organisationUnitCode=TO&projectCode=fanafana&startDate=2015-01-01&timeZone=Australia%2FMelbourne&viewId=19&cacheBreaker=u2f11
- dhis request: https://dev-aggregation.tupaia.org/api/analytics/rawData.json?paging=false&inputIdScheme=uid&outputIdScheme=code&includeMetadataDetails=true&dimension=dx:pDhCn9lbi5v;K3wJsRJHMCa;g2KqfrYXLdM;iYzYjNzDnAY;RngB7fLGI3i;bcg0ApKZTav;OzDb71SXJgP&dimension=ou:jSk4k1gnSqk&dimension=co&dimension=pe:201501;201502;201503;201504;201505;201506;201507;201508;201509;201510;201511;201512;201601;201602;201603;201604;201605;201606;201607;201608;201609;201610;201611;201612;201701;201702;201703;201704;201705;201706;201707;201708;201709;201710;201711;201712;201801;201802;201803;201804;201805;201806;201807;201808;201809;201810;201811;201812;201901;201902;201903;201904;201905;201906;201907;201908;201909;201910;201911;201912;202001;202002;202003;202004;202005;202006;202007;202008;202009;202010;202011;202012;20210101;20210102;20210103;20210104;20210105;20210106;20210107;20210108;20210109;20210110;20210111;20210112;20210113;20210114;20210115;20210116;20210117;20210118

Test 3:
- Event analytics query 
- request is modified to swap org unit code TO to id, program code HP02 to id
- dashboard: http://localhost:8000/api/v1/view?dashboardGroupId=318&endDate=2020-12-31&isExpanded=true&organisationUnitCode=TO&projectCode=fanafana&startDate=2020-01-01&timeZone=Australia%2FMelbourne&viewId=TO_Total_Screened_For_NCD_Risk_Factors&cacheBreaker=f78w9i
- dhis request: https://dev-tonga-aggregation.tupaia.org/api/analytics/events/query/OmssK27WEJB?paging=false&dimension=ou:jSk4k1gnSqk&startDate=2020-01-01&endDate=2020-12-31

### DIFF
--- a/packages/data-broker/src/__tests__/services/dhis/DhisInputSchemeResolvingApiProxy.test.js
+++ b/packages/data-broker/src/__tests__/services/dhis/DhisInputSchemeResolvingApiProxy.test.js
@@ -48,6 +48,7 @@ describe('DhisInputSchemeResolvingApiProxy', () => {
     });
 
     it('should not swap any codes for ids if some data elements dont exist in data_source', async () => {
+      // TODO: remove after tupaia-backlog#663 is done
       await proxy.getAnalytics({
         dataElementCodes: ['EL1', 'EL444'], // EL444 does not exist
         organisationUnitCodes: ['ORG1'],
--- a/packages/data-broker/src/__tests__/services/dhis/DhisInputSchemeResolvingApiProxy.test.js
+++ b/packages/data-broker/src/__tests__/services/dhis/DhisInputSchemeResolvingApiProxy.test.js
@@ -1,0 +1,105 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { translateElementKeysInEventAnalytics } from '@tupaia/dhis-api';
+import { createApiProxyStub, createApiStub } from './DhisInputSchemeResolvingApiProxy.stubs';
+
+jest.mock('@tupaia/dhis-api');
+
+const api = createApiStub();
+const proxy = createApiProxyStub(api);
+
+describe('DhisInputSchemeResolvingApiProxy', () => {
+  describe('getAnalytics', () => {
+    it('should swap codes for dhisId if all specified', async () => {
+      await proxy.getAnalytics({
+        dataElementCodes: ['EL1', 'EL2'],
+        organisationUnitCodes: ['ORG1'],
+      });
+      return expect(api.getAnalytics).toHaveBeenCalledOnceWith({
+        dataElementIds: ['dhisId_el1', 'dhisId_el2'],
+        organisationUnitIds: ['dhisId_ou1'],
+        inputIdScheme: 'uid',
+      });
+    });
+
+    it('should not swap any codes for ids if some org units dont have dhisId', async () => {
+      await proxy.getAnalytics({
+        dataElementCodes: ['EL1', 'EL2'],
+        organisationUnitCodes: ['ORG1', 'ORG2'], // org2 not present in mapping table, so doesn't have dhis id
+      });
+      return expect(api.getAnalytics).toHaveBeenCalledOnceWith({
+        dataElementCodes: ['EL1', 'EL2'],
+        organisationUnitCodes: ['ORG1', 'ORG2'],
+      });
+    });
+
+    it('should not swap any codes for ids if some data elements dont have dhisId', async () => {
+      await proxy.getAnalytics({
+        dataElementCodes: ['EL1', 'EL3'], // EL3 does not have dhis id
+        organisationUnitCodes: ['ORG1'],
+      });
+      return expect(api.getAnalytics).toHaveBeenCalledOnceWith({
+        dataElementCodes: ['EL1', 'EL3'],
+        organisationUnitCodes: ['ORG1'],
+      });
+    });
+
+    it('should not swap any codes for ids if some data elements dont exist in data_source', async () => {
+      await proxy.getAnalytics({
+        dataElementCodes: ['EL1', 'EL444'], // EL444 does not exist
+        organisationUnitCodes: ['ORG1'],
+      });
+      return expect(api.getAnalytics).toHaveBeenCalledOnceWith({
+        dataElementCodes: ['EL1', 'EL444'],
+        organisationUnitCodes: ['ORG1'],
+      });
+    });
+  });
+
+  describe('getEventAnalytics', () => {
+    it('should swap codes for dhisId if all specified', async () => {
+      await proxy.getEventAnalytics({
+        dataElementCodes: ['EL1', 'EL2'],
+        organisationUnitCodes: ['ORG1'],
+        programCode: 'G1',
+      });
+      return expect(api.getEventAnalytics).toHaveBeenCalledOnceWith({
+        dataElementIdScheme: 'id',
+        dataElementIds: ['dhisId_el1', 'dhisId_el2'],
+        organisationUnitIds: ['dhisId_ou1'],
+        programId: 'dhisId_g1',
+      });
+    });
+
+    it('should swap any codes that it can and leave the rest', async () => {
+      await proxy.getEventAnalytics({
+        dataElementCodes: ['EL1', 'EL2'],
+        organisationUnitCodes: ['ORG1', 'ORG2'], // org2 not present in mapping table, so doesn't have dhis id
+        programCode: 'G1',
+      });
+      return expect(api.getEventAnalytics).toHaveBeenCalledOnceWith({
+        dataElementIdScheme: 'id',
+        dataElementIds: ['dhisId_el1', 'dhisId_el2'],
+        organisationUnitCodes: ['ORG1', 'ORG2'],
+        programId: 'dhisId_g1',
+      });
+    });
+
+    it('should process the response and swap ids back to codes', async () => {
+      await proxy.getEventAnalytics({
+        dataElementCodes: ['EL1', 'EL2'],
+        organisationUnitCodes: ['ORG1'],
+        programCode: 'G1',
+      });
+      return expect(translateElementKeysInEventAnalytics).toHaveBeenCalledOnceWith(
+        {
+          SOME_EVENT_ANALYTICS_RESPONSE: 1,
+        },
+        { dhisId_el1: 'EL1', dhisId_el2: 'EL2' },
+      );
+    });
+  });
+});
--- a/packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
+++ b/packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
@@ -95,6 +95,7 @@ export class DhisInputSchemeResolvingApiProxy {
       if (!dataElement) {
         // most likely using a data element code that is not defined in data_source, which is supported
         // by this service
+        // TODO: Remove after tupaia-backlog#663 is done
         return false;
       }
 
--- a/packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
+++ b/packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
@@ -1,0 +1,244 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { translateElementKeysInEventAnalytics } from '@tupaia/dhis-api';
+
+export class DhisInputSchemeResolvingApiProxy {
+  constructor(models, api) {
+    this.models = models;
+    this.api = api;
+  }
+
+  async getAnalytics(query) {
+    let modifiedQuery = { ...query };
+
+    if (
+      (await this.allDataElementsHaveDhisId(query)) &&
+      (await this.allOrgUnitsHaveDhisId(query))
+    ) {
+      // the endpoint used /api/analytics/rawData.json only allows a single "inputIdScheme", which means
+      // both the dataElements and orgUnits need to be ids
+      modifiedQuery = await this.replaceDataElementCodesWithIds(modifiedQuery);
+      modifiedQuery = await this.replaceOrgUnitCodesWithIds(modifiedQuery);
+      modifiedQuery.inputIdScheme = 'uid';
+    }
+
+    return this.api.getAnalytics(modifiedQuery);
+  }
+
+  async getEventAnalytics(query) {
+    let modifiedQuery = { ...query };
+
+    // The base method here, DhisApi#getEventAnalytics will convert codes to ids before submitting the
+    // actual data request. It does that with separate calls e.g. /api/dataElements. However, with dhis
+    // setups where the Data Elements do not have codes, this api call will give us no new information.
+    // To prevent this, we pre-emptively swap out codes for ids, using our internal mapping, so that
+    // DhisApi#getEventAnalytics does not need to make these code-to-id conversion calls.
+    const { dataElementIdScheme = 'code' } = query;
+
+    const modifyDataElementsToUseDhisId = await this.allDataElementsHaveDhisId(query);
+
+    if (modifyDataElementsToUseDhisId) {
+      modifiedQuery = await this.replaceDataElementCodesWithIds(modifiedQuery);
+      modifiedQuery.dataElementIdScheme = 'id';
+    }
+
+    if (await this.allProgramsHaveDhisId(query)) {
+      modifiedQuery = await this.replaceProgramCodesWithIds(modifiedQuery);
+    }
+
+    if (await this.allOrgUnitsHaveDhisId(query)) {
+      modifiedQuery = await this.replaceOrgUnitCodesWithIds(modifiedQuery);
+    }
+
+    const response = await this.api.getEventAnalytics(modifiedQuery);
+
+    if (dataElementIdScheme !== 'code') {
+      return response;
+    }
+
+    // The api response will contain data elements with ids, and DhisApi will not be able to translate these
+    // back into codes (because the codes are not set in dhis). So, we have to do it ourselves using the internal
+    // mapping.
+    if (!modifyDataElementsToUseDhisId) {
+      return response;
+    }
+
+    const dataElementIdToCode = {};
+
+    const dataElements = await this.models.dataSource.find({ code: query.dataElementCodes });
+
+    for (const dataElement of dataElements) {
+      if (dataElement.config.dhisId) {
+        dataElementIdToCode[dataElement.config.dhisId] = dataElement.code;
+      }
+    }
+
+    return translateElementKeysInEventAnalytics(response, dataElementIdToCode);
+  }
+
+  /**
+   * @param query {*}
+   * @returns bool
+   * @private
+   */
+  allDataElementsHaveDhisId = async query => {
+    const { dataElementCodes } = query;
+
+    const dataElements = await this.models.dataSource.find({ code: dataElementCodes });
+
+    for (const dataElementCode of dataElementCodes) {
+      const dataElement = dataElements.find(d => d.code === dataElementCode);
+
+      if (!dataElement) {
+        // most likely using a data element code that is not defined in data_source, which is supported
+        // by this service
+        return false;
+      }
+
+      if (!dataElement.config.dhisId) {
+        return false;
+      }
+    }
+
+    return true;
+  };
+
+  /**
+   * @param query {*}
+   * @returns bool
+   * @private
+   */
+  allOrgUnitsHaveDhisId = async query => {
+    const { organisationUnitCode, organisationUnitCodes } = query;
+
+    const orgUnitCodes = organisationUnitCode ? [organisationUnitCode] : organisationUnitCodes;
+
+    const mappings = await this.models.dhisOrgUnitMapping.find({ code: orgUnitCodes });
+
+    for (const orgUnitCode of orgUnitCodes) {
+      const mapping = mappings.find(m => m.code === orgUnitCode);
+
+      if (!mapping) {
+        return false;
+      }
+    }
+
+    return true;
+  };
+
+  /**
+   * @param query {*}
+   * @returns bool
+   * @private
+   */
+  allProgramsHaveDhisId = async query => {
+    const { programCode } = query;
+
+    const dataGroups = await this.models.dataSource.find({ code: programCode });
+
+    for (const dataGroup of dataGroups) {
+      if (!dataGroup.config.dhisId) {
+        return false;
+      }
+    }
+
+    return true;
+  };
+
+  /**
+   * @param query {*}
+   * @returns {*}
+   * @private
+   */
+  replaceDataElementCodesWithIds = async query => {
+    const modifiedQuery = { ...query, dataElementIds: [] };
+
+    const { dataElementCodes } = query;
+
+    const dataElements = await this.models.dataSource.find({ code: dataElementCodes });
+
+    for (const dataElementCode of dataElementCodes) {
+      const dataElement = dataElements.find(d => d.code === dataElementCode);
+
+      if (!dataElement) {
+        throw new Error(
+          'DataElement not found in data_source, attempted to replace its code with an id',
+        );
+      }
+
+      if (!dataElement.config.dhisId) {
+        throw new Error(
+          'DataElement does not have a dhisId, attempted to replace its code with the id',
+        );
+      }
+
+      modifiedQuery.dataElementIds.push(dataElement.config.dhisId);
+    }
+
+    delete modifiedQuery.dataElementCodes;
+    return modifiedQuery;
+  };
+
+  /**
+   * @param query {*}
+   * @returns {*}
+   * @private
+   */
+  replaceOrgUnitCodesWithIds = async query => {
+    const modifiedQuery = { ...query, organisationUnitIds: [] };
+
+    const { organisationUnitCode, organisationUnitCodes } = query;
+
+    const orgUnitCodes = organisationUnitCode ? [organisationUnitCode] : organisationUnitCodes;
+
+    const mappings = await this.models.dhisOrgUnitMapping.find({ code: orgUnitCodes });
+
+    for (const orgUnitCode of orgUnitCodes) {
+      const mapping = mappings.find(m => m.code === orgUnitCode);
+
+      if (!mapping) {
+        throw new Error(
+          'Org Unit not found in dhis_org_unit_mapping, attempted to replace its code with the id',
+        );
+      }
+
+      modifiedQuery.organisationUnitIds.push(mapping.dhis_id);
+    }
+
+    delete modifiedQuery.organisationUnitCode;
+    delete modifiedQuery.organisationUnitCodes;
+    return modifiedQuery;
+  };
+
+  /**
+   * @param query {*}
+   * @returns {*}
+   * @private
+   */
+  replaceProgramCodesWithIds = async query => {
+    const modifiedQuery = { ...query };
+
+    const { programCode } = query;
+
+    const dataGroup = await this.models.dataSource.findOne({ code: programCode });
+
+    if (!dataGroup) {
+      throw new Error(
+        'Program/DataGroup not found in data_source, attempted to replace its code with an id',
+      );
+    }
+
+    if (!dataGroup.config.dhisId) {
+      throw new Error(
+        'Program/DataGroup does not have a dhisId, attempted to replace its code with the id',
+      );
+    }
+
+    modifiedQuery.programId = dataGroup.config.dhisId;
+    delete modifiedQuery.programCode;
+    return modifiedQuery;
+  };
+}
--- a/packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
+++ b/packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
@@ -68,7 +68,10 @@ export class DhisInputSchemeResolvingApiProxy {
 
     const dataElementIdToCode = {};
 
-    const dataElements = await this.models.dataSource.find({ code: query.dataElementCodes });
+    const dataElements = await this.models.dataSource.find({
+      code: query.dataElementCodes,
+      type: 'dataElement',
+    });
 
     for (const dataElement of dataElements) {
       if (dataElement.config.dhisId) {
@@ -87,7 +90,10 @@ export class DhisInputSchemeResolvingApiProxy {
   allDataElementsHaveDhisId = async query => {
     const { dataElementCodes } = query;
 
-    const dataElements = await this.models.dataSource.find({ code: dataElementCodes });
+    const dataElements = await this.models.dataSource.find({
+      code: dataElementCodes,
+      type: 'dataElement',
+    });
 
     for (const dataElementCode of dataElementCodes) {
       const dataElement = dataElements.find(d => d.code === dataElementCode);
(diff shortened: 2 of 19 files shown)


packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+
+  async getEventAnalytics(query) {
+    // console.log('DhisInputSchemeResolvingApi', 'getEventAnalytics', 'query', query);
+    let modifiedQuery = { ...query };
+
+    const originalDataElementIdScheme = query.dataElementIdScheme;
+
+    const areDataElementsUsingDhisId = await this.anyDataElementsHaveDhisId(query);
+
+    if (areDataElementsUsingDhisId) {
```
IgorNadj, line 61: As far as I can tell right now, this api request allows for either codes or ids to be passed in for each individual param, dataElement, program, org unit, so we don't have to convert them all like we do above for getAnalytics.

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+
+      // const dataElements = await this.models.dataSource.find({ id: query.dataElementCodes });
+      //
+      // for (const dataElement of dataElements) {
+      //   if (dataElement.config.dhisId) {
+      //     dataElementIdToCode[dataElement.dhisId] = dataElement.code;
+      //   }
+      // }
+
+      for (const dataElementCode of query.dataElementCodes) {
```
IgorNadj, line 93: This conversion back allows data builders to work with codes instead of ids, for known mappings,

packages/data-broker/src/services/dhis/getDhisApiInstance.js
```diff
@@ -5,13 +5,30 @@
...
+export const getDhisApiInstance = (options, models) => {
+  console.log('getDhisApiInstance');
   const { serverName, serverUrl } = getDhisConfig(options);
   if (!instances[serverName]) {
-    instances[serverName] = new DhisApi(serverName, serverUrl);
+    const api = new DhisApi(serverName, serverUrl);
+
+    const dhisInputSchemeResolvingApi = new DhisInputSchemeResolvingApi(models, api);
+
+    const proxy = new Proxy(api, {
```
IgorNadj, line 20: Ran into some sort of circular import problem, not sure how... but the proxy pattern seems to work well, can debug as usual etc. Might make code navigation harder through IDEs but when this code is cleaned up it should be simple enough for a human to follow.

packages/dhis-api/src/DhisApi.js
```diff
@@ -35,6 +35,7 @@ const LATEST_LOOKBACK_PERIOD = '600d';
 
 export class DhisApi {
```
IgorNadj, line 36: The changes in this class are just to allow the api to take ids instead of codes.

packages/dhis-api/src/__tests__/buildAnalyticsQuery.test.js
```diff
@@ -107,4 +107,117 @@ describe('buildAnalyticsQuery', () => {
       assertArrayHasDimensionWithMembers(results.dimension, 'ou', ['pg_dhisId', 'to_dhisId']);
     });
   });
+
```
IgorNadj, line 110: Tests coming next

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+    this.models = models;
+    this.api = api;
+  }
+
+  async getAnalytics(query) {
+    let modifiedQuery = { ...query };
+
+    // console.log('DhisInputSchemeResolvingApi', 'getAnalytics', 'query', query);
+
+    if (
```
IgorNadj, line 36: I checked DHIS api docs, couldn't find a way of specifying the inputIdScheme individually per dimension, and tried a few different ways in postman, I think it has to be all or nothing here...

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+
+      modifiedQuery.inputIdScheme = 'uid';
+    }
+
+    // console.log('DhisInputSchemeResolvingApi', 'getAnalytics', 'modifiedQuery', modifiedQuery);
+    return this.api.getAnalytics(modifiedQuery);
+  }
+
+  async getEventAnalytics(query) {
+    // console.log('DhisInputSchemeResolvingApi', 'getEventAnalytics', 'query', query);
```
IgorNadj, line 54: The base method here, DhisApi#getEventAnalytics will convert codes to ids before submitting the actual data request. It does that with separate calls e.g. /api/dataElements. To prevent this (because laos does not have any codes), we pre-emptively swap out codes for ids, using our internal mapping.

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+// TODO: move to actual db table
+const DATA_SOURCE_CONFIG_BY_CODE = {
+  // SS162: { dhisId: 'pDhCn9lbi5v' },
+  // SS128: { dhisId: 'K3wJsRJHMCa' },
+  // SS182: { dhisId: 'g2KqfrYXLdM' },
+  // SS190A: { dhisId: 'iYzYjNzDnAY' },
+  // SS192: { dhisId: 'RngB7fLGI3i' },
+  // SS219: { dhisId: 'bcg0ApKZTav' },
+  // DP9: { dhisId: 'OzDb71SXJgP' },
+  HP02: { dhisId: 'OmssK27WEJB' },
```
IgorNadj, line 22: This is a program / dataGroup

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+    return modifiedQuery;
+  };
+
+  /**
+   * @param query {*}
+   * @returns {*}
+   * @private
+   */
+  replaceOrgUnitCodesWithIds = async query => {
+    const modifiedQuery = { ...query, organisationUnitIds: [] };
```
IgorNadj, line 238: Note to self: if we set dhisId on these entities and dataElements, and this proxy runs into an api request that hits any of them, it will require them all to have ids. 

TODO: change the logic here to: if they don't _all_ have ids, use codes instead.

In the case of laos eoc, when building dashboards and map overlays, we will know pretty quickly if we have a missing dhisId, because the whole thing will fall over.

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
edmofro: Could you check whether the schemes described in https://docs.dhis2.org/2.28/en/developer/html/dhis2_developer_manual_full.html#webapi_identifier_schemes apply to the endpoints we use?

packages/data-broker/src/services/dhis/getDhisApiInstance.js
edmofro: Wow, I didn't know this was baked in to js - that's super cool 😎

edmofro: What happens if you just pass in `dhisInputSchemeResolvingApi` as the handler directly? From a brief read of the mdn doc for Proxy I would've thought that would work

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * @param query {*}
+   * @returns bool
+   * @private
+   */
+  anyDataElementsHaveDhisId = async query => {
+    const { dataElementCodes } = query;
+
+    // const dataElements = await this.models.dataSource.find({ code: dataElementCodes });
```
edmofro, line 122: Should even be able to check for the presence of `config.dhisId` in the query, eventually :-)

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * @param query {*}
+   * @returns bool
+   * @private
+   */
+  anyProgramsHaveDhisId = async query => {
+    const { programCode } = query;
+
+    // const dataElements = await this.models.dataSource.find({ code: programCode });
```
edmofro, line 168:
```suggestion
    // const dataGroups = await this.models.dataSource.find({ code: programCode });
```

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
IgorNadj: Ah there that is, thanks. No luck unfortunately...

I had a look at the request class for this in dhis2 and it only seems to have inputIdSchema:

https://github.com/dhis2/dhis2-core/blob/d883e68107f971f00f7b9f42d6dd7b001d6cbfd1/dhis-2/dhis-api/src/main/java/org/hisp/dhis/common/AggregateAnalyticsQueryCriteria.java#L182

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
IgorNadj: Added as comment

packages/data-broker/src/services/dhis/getDhisApiInstance.js
IgorNadj: No, it doesn't work, it needs the proxy specification object, with get() etc.

IgorNadj:
@edmofro this PR has graduated from WIP to a grown up PR, ready for review. One commit at a time might be easiest to follow, although it's not too big anyway.

packages/dhis-api/src/__tests__/DhisApi/testGetEventAnalytics.js
```diff
@@ -59,14 +59,35 @@ export const testGetEventAnalytics = () => {
 
   it('translates codes to ids in the provided query', async () => {
     await dhisApi.getEventAnalytics(QUERY.originalInput);
-    expect(BuildAnalyticsQuery.buildEventAnalyticsQuery).toHaveBeenCalledOnceWith(
+    return expect(BuildAnalyticsQuery.buildEventAnalyticsQuery).toHaveBeenCalledOnceWith(
```
edmofro, line 62: Ooh, good catch

packages/data-broker/src/__tests__/services/dhis/DhisInputSchemeResolvingApiProxy.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+    it('should not swap any codes for ids if some data elements dont exist in data_source', async () => {
+      await proxy.getAnalytics({
+        dataElementCodes: ['EL1', 'EL444'], // EL444 does not exist
+        organisationUnitCodes: ['ORG1'],
+      });
+      return expect(api.getAnalytics).toHaveBeenCalledOnceWith({
+        dataElementCodes: ['EL1', 'EL444'],
+        organisationUnitCodes: ['ORG1'],
+      });
```
edmofro, line 58: @kael89 this can be deleted when https://github.com/beyondessential/tupaia-backlog/issues/663 is complete

@IgorNadj could you add a todo comment?

packages/dhis-api/src/__tests__/DhisApi/testGetEvents.js
```diff
@@ -0,0 +1,94 @@
...
+      organisationUnitId: 'to_dhisId',
+      eventId: 'dhis_event_id1',
+    });
+    return expect(dhisApi.fetcher.fetch).toHaveBeenCalledOnceWith(
+      'events/dhis_event_id1',
+      {
+        program: PROGRAM.id,
+        orgUnit: 'to_dhisId',
+        programIdScheme: 'code',
+        orgUnitIdScheme: 'code',
```
edmofro, line 87: 🤔

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -0,0 +1,244 @@
...
+    const dataElements = await this.models.dataSource.find({ code: dataElementCodes });
+
+    for (const dataElementCode of dataElementCodes) {
+      const dataElement = dataElements.find(d => d.code === dataElementCode);
+
+      if (!dataElement) {
+        // most likely using a data element code that is not defined in data_source, which is supported
+        // by this service
+        return false;
+      }
```
edmofro, line 99: @kael89 can also be deleted after your work on https://github.com/beyondessential/tupaia-backlog/issues/663

@IgorNadj a todo comment here too please

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -0,0 +1,244 @@
...
+
+  /**
+   * @param query {*}
+   * @returns bool
+   * @private
+   */
+  allDataElementsHaveDhisId = async query => {
+    const { dataElementCodes } = query;
+
+    const dataElements = await this.models.dataSource.find({ code: dataElementCodes });
```
edmofro, line 90:
```suggestion
    const dataElements = await this.models.dataSource.find({ code: dataElementCodes, type: 'dataElement' });
```
Same for the groups check below; `data_source.code` may not be unique

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -0,0 +1,244 @@
...
+        throw new Error(
+          'DataElement does not have a dhisId, attempted to replace its code with the id',
+        );
+      }
+
+      modifiedQuery.dataElementIds.push(dataElement.config.dhisId);
+    }
+
+    delete modifiedQuery.dataElementCodes;
+    return modifiedQuery;
```
edmofro, line 182: A more aggressive destructure up front avoids a `delete` later. Up to you though, it's a style thing and is fine either way

```suggestion
    const { dataElementCodes, ...restOfQuery } = query;
    
    const modifiedQuery = { ...restOfQuery, dataElementIds: [] };

    const dataElements = await this.models.dataSource.find({ code: dataElementCodes });

    for (const dataElementCode of dataElementCodes) {
      const dataElement = dataElements.find(d => d.code === dataElementCode);

      if (!dataElement) {
        throw new Error(
          'DataElement not found in data_source, attempted to replace its code with an id',
        );
      }

      if (!dataElement.config.dhisId) {
        throw new Error(
          'DataElement does not have a dhisId, attempted to replace its code with the id',
        );
      }

      modifiedQuery.dataElementIds.push(dataElement.config.dhisId);
    }

    return modifiedQuery;
```

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -0,0 +1,244 @@
...
+   * @param query {*}
+   * @returns {*}
+   * @private
+   */
+  replaceDataElementCodesWithIds = async query => {
+    const modifiedQuery = { ...query, dataElementIds: [] };
+
+    const { dataElementCodes } = query;
+
+    const dataElements = await this.models.dataSource.find({ code: dataElementCodes });
```
edmofro, line 161: Same comment about checking `type` as above

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -0,0 +1,244 @@
...
+   * @param query {*}
+   * @returns {*}
+   * @private
+   */
+  replaceProgramCodesWithIds = async query => {
+    const modifiedQuery = { ...query };
+
+    const { programCode } = query;
+
+    const dataGroup = await this.models.dataSource.findOne({ code: programCode });
```
edmofro, line 226: `type` check needed

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -0,0 +1,244 @@
...
+
+  /**
+   * @param query {*}
+   * @returns bool
+   * @private
+   */
+  allProgramsHaveDhisId = async query => {
+    const { programCode } = query;
+
+    const dataGroups = await this.models.dataSource.find({ code: programCode });
```
edmofro, line 140: `type` check

packages/database/src/migrations/20210121045801-AddTableDhisOrgUnitMapping-modifies-schema.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+exports.up = function (db) {
+  return db.createTable('dhis_org_unit_mapping', {
+    columns: {
+      id: { type: 'text', primaryKey: true },
+      code: { type: 'text', notNull: true, unique: true },
+      dhis_id: { type: 'text', notNull: true },
+    },
+  });
```
edmofro, line 24: What do you think about a more generic table that could be used for other data service entity config later?
```suggestion
  return db.createTable('data_service_entity', {
    columns: {
      id: { type: 'text', primaryKey: true },
      entity_code: { type: 'text', notNull: true, unique: true },
      config: { type: 'jsonb', notNull: true, default: '{}' },
    },
  });
```

That would fit better with our `data_source` model, and be more extensible. Will we ever use it? Hard to know, but I don't think the abstraction adds too much complexity to the overall mapping workflow - do you?

FYI, we do want to break `data_source` out into `data_element` and `data_group` (or even `data_service_element` and `data_service_element_group` to match `data_service_entity`). Having them in the one table at the moment is awkward. See https://github.com/beyondessential/tupaia-backlog/issues/662

edmofro (changes requested):
Only very minor comments - looking awesome overall!

packages/dhis-api/src/__tests__/buildAnalyticsQuery.test.js
```diff
@@ -107,4 +107,34 @@ describe('buildAnalyticsQuery', () => {
...
+        dataElementIds: ['pop01_dhisId'],
+        organisationUnitIds: ['tonga_dhisId'],
+      });
+
+      expect(queries.length).toBe(1);
+
+      const [query] = queries;
+
+      expect(query).toHaveProperty('dimension');
+      expect(query.dimension).toIncludeAllMembers(['dx:pop01_dhisId', 'ou:tonga_dhisId']);
```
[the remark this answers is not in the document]
IgorNadj: I thought I would leave dhis-api to be dumb and just mostly pass along what it has been given in the method args. Our dhis-api seems to be responsible for connecting to the server, auth and some minor param adjustments e.g. prefixing pop01_dhisId with "dx:" to give dx:pop01_dhisId. 

I think the better place for the test on idScheme would be in data-broker, and the test exists: https://github.com/beyondessential/tupaia/pull/1908/files#diff-f2bce110ae2acc9a0486da97440f721a6c5bde2fd3a1b1434df8555df6ae4e66R24

Happy with that?

packages/dhis-api/src/__tests__/DhisApi/testGetEventAnalytics.js
IgorNadj: I'm not sure how but I think it works either way... 
I raised https://github.com/beyondessential/tupaia-backlog/issues/1599 to catch this issue, but even without the return it was working... magic. I added it in for safety.

edmofro: Surprising - I wonder if the whole test suite ended before the async function had completed whether the failure would get missed. Definitely safer the way you've got it.

packages/dhis-api/src/__tests__/buildAnalyticsQuery.test.js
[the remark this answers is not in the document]
edmofro: Yes happy :-)

packages/dhis-api/src/__tests__/DhisApi/testGetEvents.js
IgorNadj: Yeah this is a strange one. It's already like that in production, sending uids to the api but saying they are codes (https://github.com/beyondessential/tupaia/blob/dev/packages/dhis-api/src/DhisApi.js#L234). I will fix it anyway, since this will be regression tested anyway.

One question though: do we want to support this code to id swapping for DhisApi#getEvents()? It is only ever called from data-broker via pullEventsForApi_Deprecated. I'm happy to add it for completeness.

edmofro: Oh right...no I think it's better to be restrictive. Thanks for picking that up.

packages/database/src/migrations/20210121045801-AddTableDhisOrgUnitMapping-modifies-schema.js
IgorNadj: I think making that table change isn't really needed, since 
- we don't have services other than dhis we need this for (so don't need a generic config instead of dhis_id)
- we don't need any other attributes against entity we would want to give to data broker

I don't see anything like that coming up on the horizon, as far as I know, so I'm happy to leave the abstraction out.

The breaking of data_source into two tables shouldn't affect this either, not sure how that's related?

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
IgorNadj: Yeah good point, I might leave it in this case but will definitely destructure more aggressively next time

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+      if (!DATA_SOURCE_CONFIG_BY_CODE[dataElementCode].dhisId) {
+        throw new Error(
+          'DataElement does not have a dhisId, attempted to replace its code with the id',
+        );
+      }
+
+      modifiedQuery.dataElementIds.push(DATA_SOURCE_CONFIG_BY_CODE[dataElementCode].dhisId);
+    }
+
+    delete modifiedQuery.dataElementCodes;
```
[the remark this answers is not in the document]
IgorNadj: Replied in big comment below

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApi.js
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * @param query {*}
+   * @returns {*}
+   * @private
+   */
+  replaceDataElementCodesWithIds = async query => {
+    const modifiedQuery = { ...query, dataElementIds: [] };
+
+    const { dataElementCodes } = query;
```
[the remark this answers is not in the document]
IgorNadj: Replied in big comment below

packages/utils/src/proxy.js
```diff
@@ -0,0 +1,10 @@
+export const createClassExtendingProxy = (targetInstance, extendingClassInstance) => {
```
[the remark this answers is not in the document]
IgorNadj: Added

packages/data-broker/src/__tests__/services/dhis/DhisInputSchemeResolvingApiProxy.test.js
IgorNadj: Added

packages/dhis-api/src/__tests__/DhisApi/testGetEvents.js
IgorNadj: Fixed schemes

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
IgorNadj: Added

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
IgorNadj: Updated

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
IgorNadj: Updated

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
IgorNadj: Updated

packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
IgorNadj: Updated

packages/database/src/migrations/20210121045801-AddTableDhisOrgUnitMapping-modifies-schema.js
edmofro: Those are good points.

The connection breaking `data_source` was just that then we'd have three similar tables, all providing information to data broker about how a given dimension maps from tupaia to the external data service:
- `data_service_element`
- `data_service_element_group`
- `data_service_entity`

But, thinking about it now, that may be jumping the gun. Unlike data elements and data groups, the entity doesn't inform the top level of data broker which service we should use - instead it's information used within a specific data service, so it makes sense to have a more specific table.

Happy to stick with what you've got - but just keen to hear what @kael89 thinks first as he put a lot of work into the data broker system too.

kael89: While working with Tupaia I've seen DHIS specific conventions being generalised time and time again. It may not happen immediately, but it eventually happens after perhaps 1-3 years.

With that in mind, I would also lean towards a more generic table. Even if we only use it for dhis eventually, the extra abstraction doesn't seem to add too much complexity to make us concerned.

On the other hand, it's certainly not the end of the world if this stays DHIS specific for now. Bottom line: I'd be keen to see a more generic table, but would not complain if a specific one is used eventually.

edmofro: Thanks @kael89. A bit of a push there @IgorNadj, but you've got reason enough to stick with your existing table if you still prefer. Over to you :-)

kael89: Yeah I'd also consider this as a "here's one suggestion, but up to the author" situation 😃

IgorNadj: No worries, I have changed the table to the generic version

packages/dhis-api/src/DhisApi.js
```diff
@@ -240,8 +240,8 @@ export class DhisApi {
     const queryParameters = {
       program: resolvedProgramId,
       orgUnit: resolvedOrganisationUnitId,
-      programIdScheme: 'code',
-      orgUnitIdScheme,
+      programIdScheme: 'uid',
+      orgUnitIdScheme: 'uid',
```
edmofro, line 244: Actually I don't think this change is safe. The deprecated functions are still used in a number of places (we have https://github.com/beyondessential/tupaia-backlog/issues/451 to get rid of it), and they'll be assuming the response has codes for org units and programs. Sorry @IgorNadj, I didn't look into it properly when we were talking earlier!

edmofro (changes requested):
Everything looking great except one issue that's my fault

packages/dhis-api/src/DhisApi.js
IgorNadj: Ah, no worries, reverted.

edmofro (approved):

PR #1913, opened 2021-01-18
Fix admin panel survey responses export

tcaiger:
### Changes:

- Update survey response export handler to handle the `entityIds` query param as either an array or a comma separated string.

Note I investigated updating the request on the front end to be a comma separated list instead of an array but it uses a project wide util called `stringifyQuery` in @tupaia/utils/src/request.js.

Screenshot of request unit test showing how arrays of params are converted to a query string in tupaia utils

Screen Shot 2021-01-19 at 12 11 23 PM


EMcQ-BES (approved):
Nice and simple. Thanks for this Tom.

PR #1919, opened 2021-01-19
1799: Update indicator tests

kael89:
### Issue #:
Adds integration tests in `indicators` + updates unit tests for https://github.com/beyondessential/tupaia-backlog/issues/1799

#### Integration test conventions

This PR introduces some conventions for integration tests. I'm happy to discuss their implementation or whether they are beneficial at all!

Integration tests are:
* Placed under an `__integration` folder inside `__tests__`. We could also name this folder `__integration__`
* Config driven - it should be fairly easy to add a new test case
* Mocking should be minimal: network requests to 3rd party APIs, and potentially datetime functionality
* The input and output should be very specific
* They generally test realistic "happy paths" and main error throwing - unit tests can be used to test edge cases (there can be exceptions to this in practice)

A lot of the above are already implied in existing integration tests in our codebase. I used a specific approach here:
* A `.json` file serves as the test config. `.js` files may abstract the config generation, but in this case it's probably easier to use plain config. A drawback is that you can't have comments in `.json`, so we could also try `.jsonc`
* A generic `runItntegrationTest` helper that encourages adhering to the above by standardising test aspects (setup, input, expected output)

My intention would be for the same or similar pattern to be used for API testing (like the one we do in `meditrak-server` currently, using supertest). The used endpoints and expected responses can easily be specified in config. One difference here is that we may choose to place those under an `__api` folder.


packages/indicators/src/__tests__/Builder/ArithmeticBuilder/ArithmeticBuilder.test.ts
```diff
@@ -0,0 +1,26 @@
...
+}));
+
+describe('ArithmeticBuilder', () => {
+  describe('validateConfig()', testConfigValidation);
+
+  describe('getAggregations()', testGetAggregations);
+
+  describe('getElementCodes()', testGetElementCodes);
+
+  describe('buildAnalytics()', testBuildAnalytics);
```
kael89, line 25: Apart from `validateConfig`, all other unit tests here explore aspects that are already semi-tested in integration tests. I thought that it's still probably useful (and "better practice"?) to keep the unit tests, since they make it easy to define and test more elaborate scenarios/edge cases.

This decision was made easier by the fact that most of them were already written. I don't know if I would go this way if I had started by writing integration tests in this package. When to use each category seems like is an interesting problem to me, and probably has no specific answers. I'd be happy to hear your thoughts!

kael89: Also, we don't want to "over-test" a package since tests have maintenance overhead. I tried to be conscious of this but please let me know if you think this is the case here!

For future reference: integration tests probably have minimal maintenance cost and I wish I had them written before starting this refactor! Unit tests required some time to be updated but not too much.

packages/indicators/src/__tests__/__integration/arithmeticAnalytics.fixtures.json
```diff
@@ -0,0 +1,447 @@
...
+          "startDate": "2019-01-01",
+          "endDate": "2019-01-31",
+          "organisationUnitCodes": ["GR"]
+        }
+      },
+      "expected": [
+        {
+          "dataElement": "MonthlyBirths",
+          "organisationUnit": "GR",
+          "period": "20190101",
```
kael89, line 240: This doesn't seem correct, right? `MonthlyBirths` uses a `FINAL_EACH_MONTH` aggregation so one would expect this to be `201901`!

I believe this is a "bug" in this aggregation - as a matter of fact, most current aggregations almost never concern themselves with normalising the return period based on what they do. We'd need to change that to make them compatible with the indicator architecture - probably in https://github.com/beyondessential/tupaia-backlog/issues/2150

We were lucky in `PSSS`: we use week types there which force a period update (periods include a `W` in them so the conversion is inevitable for inclusion/comparison purposes etc)

packages/indicators/src/__tests__/__integration/arithmeticAnalytics.fixtures.json
```diff
@@ -0,0 +1,447 @@
...
+        {
+          "dataElement": "AnnualBirths",
+          "organisationUnit": "GR",
+          "period": "2020",
+          "value": 21
+        }
+      ]
+    },
+    {
+      "description": "Applies correct aggregation (object format)",
```
kael89, line 328: This test case will fail - and that's a good thing!! Again, I believe we got lucky in PSSS - there are a few bugs related to `fetchOptions` expansion during fetching, and during aggregation. Good news - this is also related to your original comment in the previous PR: https://github.com/beyondessential/tupaia/pull/1873#issuecomment-761892449.

I'll submit a new PR to fix this, and may add a couple of more complicated test cases here (nested indicators, including `OFFSET_PERIOD` and `SUM` aggregations, make sure these work correctly together)

edmofro:
Is it expected that this should fail on codeship @kael89?

packages/indicators/src/types.ts
```diff
@@ -21,50 +21,58 @@ export interface AnalyticCluster {
   dataValues: Record<Analytic['dataElement'], Analytic['value']>;
 }
 
-type TypeFields = Record<string, string | number | Record<string, unknown>>;
+type DbValue = string | number | boolean | null | DbValue[] | { [key: string]: DbValue };
```
edmofro, line 24: Some intense type nesting here, I'm impressed TypeScript is happy to evaluate that!

packages/indicators/src/__tests__/__integration/integration.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    indicatorCodes: string[];
+    fetchOptions: FetchOptions;
+  };
+  expected: Analytic[] | string;
+  throws?: boolean;
+}
+
+const dataBroker = new DataBroker();
+// TODO Use the base `@tupaia/database` models instead of `dataBroker.models`
+// after https://github.com/beyondessential/tupaia-backlog/issues/663 is implemented
```
edmofro, line 24: If this is because the `findOrDefault` behaviour is needed, I believe you pulled that into the central model recently?

packages/indicators/src/__tests__/Builder/ArithmeticBuilder/testBuildAnalytics.ts
```diff
@@ -71,35 +72,31 @@ export const testCalculations = () => {
...
         {
           formula: 'A_ToPg_20192020 + B_ToPg_20192020',
           aggregation: 'FINAL_EACH_YEAR',
         },
         [
-          { organisationUnit: 'TO', period: '2019', value: 3 },
-          { organisationUnit: 'TO', period: '2020', value: 30 },
-          { organisationUnit: 'PG', period: '2019', value: 3.3000000000000003 },
-          { organisationUnit: 'PG', period: '2020', value: 33 },
+          ['TO', '2019', 3],
```
edmofro, line 96: This is a nice example of when the values are self describing enough, that the field names really are just clutter - I like the array format :-)

kael89:
> Is it expected that this should fail on codeship @kael89?

Yes - see my [comment](https://github.com/beyondessential/tupaia/pull/1919#discussion_r560586349)

packages/database/src/testUtilities/runIntegrationTests.js
```diff
@@ -0,0 +1,65 @@
...
+ * @property {Object<string, *>} config
+ * @property {Function} [beforeAll]
+ * @property {Function} [afterAll]
+ * @property {Function} it
+ */
+
+/**
+ * @param {IntegrationTestInput} input
+ */
+export const runIntegrationTests = ({
```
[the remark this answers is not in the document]
edmofro: Hmmmm...good question! I don't have a strong opinion. It feels wrong here, but I totally see your point about introducing the `database` dependency into `utils`. Your call :-)

packages/indicators/src/__tests__/Builder/ArithmeticBuilder/ArithmeticBuilder.test.ts
```diff
@@ -0,0 +1,26 @@
...
+import { testBuildAnalytics } from './testBuildAnalytics';
+import { testGetAggregations } from './testGetAggregations';
+import { testGetElementCodes } from './testGetElementCodes';
+import { testConfigValidation } from './testValidateConfig';
+
+jest.mock('../../../Builder/createBuilder', () => ({
+  createBuilder: jest.fn().mockImplementation((indicator: Indicator) => ({
+    getElementCodes: () => indicator.config.elementCodes,
+  })),
+}));
```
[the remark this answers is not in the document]
edmofro: Yeah ... that's a shame. Could you put a comment in `testGetElementCodes` to at least point people in the right direction if they're looking?

packages/indicators/src/__tests__/Builder/ArithmeticBuilder/ArithmeticBuilder.test.ts
edmofro: I'm keen to keep both, seeing as we have them. In the end, if we find ourselves spending inordinate amounts of time updating unit tests, we can kill them at that point!

For the general case - yeah I agree that there's no definite answer to when one should definitely be used over the other. I favour integration tests for "bang for buck", but unit tests are ideal in situations when there's a specific piece of logic that lives in a single place and is complex/risky/central to the system's successful functioning.

packages/indicators/src/__tests__/__integration/arithmeticAnalytics.fixtures.json
```diff
@@ -0,0 +1,447 @@
...
+          "endDate": "2019-01-31",
+          "organisationUnitCodes": ["GR"]
+        }
+      },
+      "expected": [
+        {
+          "dataElement": "FemaleGtMale",
+          "organisationUnit": "GR",
+          "period": "20190101",
+          "value": 0
```
edmofro, line 260: So `true` is `0`, and `false` `1`?

packages/indicators/src/__tests__/__integration/arithmeticAnalytics.fixtures.json
edmofro: Oh ok cool! Nice test case :-)

edmofro (commented):
This is an awesome setup for integration tests!! 🎉 

I'm not totally convinced one big JSON file is the most clear way to define them - personally I'd prefer being able to structure things a little more explicitly. E.g. naming the input fixture blocks individually, then pulling them all together, and having that separate to the test cases. However, overall I think this is a big win to allow us to have consistent and declarative creation of integration tests. Have you played around with the same sort of thing defined in a `.js` file?

Also, I'm not going to pretend I checked all the logic of the test cases against the data fixtures - in most cases (particularly within the integration tests) I just trusted the test case descriptions.

packages/indicators/src/types.ts
kael89: Yeah TypeScript is very sophisticated!

packages/indicators/src/__tests__/__integration/integration.test.ts
kael89: Correct! And I'm about to remove `findOrDefault` completely after implementing https://github.com/beyondessential/tupaia-backlog/issues/663, so I 'll update this after 663 ends in `dev`, which should be soon enough

packages/indicators/src/__tests__/Builder/ArithmeticBuilder/testBuildAnalytics.ts
kael89: Same here! The object format seems handy for passing analytics around, but array seems simpler for defining them!

packages/indicators/src/__tests__/__integration/arithmeticAnalytics.fixtures.json
kael89: Umm no actually this is an error in the test case description!

Perhaps following the feminist tradition in Tupaia I subconsciously wanted the `FemaleGtMale` statement to be `true` rather than false 😉

edmofro: Ah right! Haha nice, Elena would be proud

packages/database/src/testUtilities/upsertDummyRecord.js
```diff
@@ -74,3 +74,11 @@ export const findOrCreateDummyRecord = async (model, findCriteria, data) => {
   const generatedData = await generateDummyRecord(model, { ...findCriteria, ...data });
   return model.findOrCreate(findCriteria, generatedData);
 };
+
+export const findOrCreateRecords = async (models, recordsByType) => {
```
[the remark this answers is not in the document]
kael89: I've thought the same in the past, but when I tried to convert one to the other a few specific tests failed. One example (could be the only one!) is [GET /changes/count](https://github.com/beyondessential/tupaia/blob/dev/packages/meditrak-server/src/tests/routeTests/changes/get.js), where we want to make sure that a record has actually been upserted (and not "found"). Of course, we could ensure this in other ways (delete before insert/use a unique code etc).

In general, we have a lot of really handy test utils that have evolved a bit independently. Hopefully one day we'll bring all of them together in a standardised test DB setup system, monorepo-wide!

packages/database/src/testUtilities/upsertDummyRecord.js
```diff
@@ -82,3 +82,17 @@ export const findOrCreateRecords = async (models, recordsByType) => {
     }
   }
 };
+
+/**
+ * Generates test data, and stores it in the database. Uses test ids so that all can be cleanly
+ * wiped afterwards. Any missing fields on the records passed in are generated randomly or using
+ * sensible defaults, using the logic in upsertDummyRecord
+ */
+export const populateTestData = async (models, recordsByType) => {
```
edmofro, line 91: Thanks for moving this alongside the others 🙌

kael89:
> This is an awesome setup for integration tests!! 🎉
> 
> I'm not totally convinced one big JSON file is the most clear way to define them - personally I'd prefer being able to structure things a little more explicitly. E.g. naming the input fixture blocks individually, then pulling them all together, and having that separate to the test cases. However, overall I think this is a big win to allow us to have consistent and declarative creation of integration tests. Have you played around with the same sort of thing defined in a `.js` file?

Thanks @edmofro, good points about defining integration tests setup in multiple `.js` files. I tried it briefly and keen to give it a second shot, after all required code fixes are done.

edmofro (approved):
Would still be keen to check out the readability of js files instead, but that aside this is all approved 🎉

edmofro:
> > This is an awesome setup for integration tests!! 🎉
> > I'm not totally convinced one big JSON file is the most clear way to define them - personally I'd prefer being able to structure things a little more explicitly. E.g. naming the input fixture blocks individually, then pulling them all together, and having that separate to the test cases. However, overall I think this is a big win to allow us to have consistent and declarative creation of integration tests. Have you played around with the same sort of thing defined in a `.js` file?
> 
> Thanks @edmofro, good points about defining integration tests setup in multiple `.js` files! I tried it briefly and keen to give it a second shot, after all required code fixes are done.

Cools, and I actually wasn't meaning multiple `.js` files - even all in one file, but the js syntax could allow better grouping and visual discrimination of sections. Let me know how it goes :-)

PR #1924, opened 2021-01-20
1173 add govt surveys custom export wish fiji

PShoe:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1173 
Adding government surveys to custom export 

### Changes:
Add two new custom survey exports for wish fiji using sub catchments (unlike other custom exports which use village)

- Example

---

### Screenshots:
image


packages/database/src/migrations/20201208073839-WishFijiGovSurveysAssociateToCatchments-modifies-data.js
```diff
@@ -0,0 +1,89 @@
...
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+// [code, name]
+const oldCatchment = ['FJ_CATCH_Namosi', 'Namosi Catchment'];
+const newCatchment = ['FJ_CATCH_Upper_Navua', 'Upper Navua Catchment'];
+
+const originalEntityId = '5df8072d61f76a485c2eb9f3'; // CWMH Suva Hospital
```
kael89, line 23: Ideally we'd get this id dynamically using a query, since it can theoretically be different in another database.

Not a problem here, just for future reference! In practice, our instance dbs get cloned from prod overnight, so ids will match! 😃

packages/database/src/migrations/20201208073839-WishFijiGovSurveysAssociateToCatchments-modifies-data.js
```diff
@@ -0,0 +1,89 @@
...
+
+const updateCatchment = (db, code, updateCode, updateName) =>
+  db.runSql(`
+    update entity 
+    set code = '${updateCode}', name = '${updateName}' 
+    where code = '${code}';
+  `);
+
+const catchmentNameToCode = {
```
kael89, line 32: I guess no more catchments are expected to be added till this runs - in which case, this hardcoded map is fine!

Alternatively, we could use the database as the ultimate source of truth

kael89 (approved):
Excellent work @PShoe !! Loved the separation of code into multiple, single-responsibility functions, the sensible variable names and the easy to follow code flow! 💯

PShoe:
Changes

1. Delete 3 incomplete/empty surveys from production 
2. Update two surveys to use PrimaryEntity question with type: sub_catchment, over radio and free text question types


WFIGMMET4 | PrimaryEntity | Catchment name | Catchment   name |   | type: sub_catchment
-- | -- | -- | -- | -- | --

WFIGM5 | PrimaryEntity | Catchment name | Catchment   name |   | type: sub_catchment
-- | -- | -- | -- | -- | --

3. Update sort in custom export dashboard migration
4. Update associate to catchments migration for 'legacy' answers that need to be mapped to a primary entity. Delete these answers from answer table after update is complete.

kael89:
> @kael89 not sure if I ping the original reviewer, or pop this back into the review channel for anyone to pick up... if the former, ping!

When a PR has further changes we ask a re-review from the original reviewer(s) - github makes this easy with this convenient button: 

image

PShoe:
> > @kael89 not sure if I ping the original reviewer, or pop this back into the review channel for anyone to pick up... if the former, ping!
> 
> When a PR has further changes we ask a re-review from the original reviewer(s) - github makes this easy with this convenient button:
> 
> image

Ah, will use the re-review next time, thank you!!

packages/database/src/migrations/20201208073839-WishFijiGovSurveysAssociateToCatchments-modifies-data.js
```diff
@@ -0,0 +1,86 @@
...
+      inner join question q
+      on a.question_id = q.id
+      where q.code = 'WFIGM5' OR q.code = 'WFIGMMET4'
+    );
+  `);
+
+exports.up = async function (db) {
+  await updateCatchment(db, oldCatchment[0], newCatchment[0], newCatchment[1]);
+  const surveyResponses = (await fetchSurveyResponseCatchments(db)).rows;
+  await Promise.all([surveyResponses, updateSurveyResponseEntities(db, surveyResponses)]);
```
kael89, line 76: Probably that would be enough since you've already awaited for `surveyResponses` one line above

```suggestion
  await updateSurveyResponseEntities(db, surveyResponses);
```

packages/database/src/migrations/20201208073839-WishFijiGovSurveysAssociateToCatchments-modifies-data.js
```diff
@@ -0,0 +1,86 @@
...
+const updateSurveyResponseEntities = async (db, surveyResponses) =>
+  surveyResponses.forEach(async sr => {
+    await updateSurveyResponseRow(db, sr.id, catchmentNameToCode[sr.name]);
+  });
+
+const deleteDeprecatedAnswers = db =>
+  db.runSql(`
+    delete from answer a
+    where id in (
+      select a.id
```
kael89, line 65: The syntax here is 100% fine, another option (to avoid nesting) is [DELETE ... USING ...](https://www.postgresqltutorial.com/postgresql-delete-join/)

kael89 (approved):
Looks great @PShoe ! I left a couple of notes but nothing that needs addressing

PR #1927, opened 2021-01-20
2160: Missing data for daily survey

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2160


packages/database/src/migrations/20210120051444-ReSyncSubmissionsAfterRelease70-modifies-data.js
```diff
@@ -0,0 +1,121 @@
...
+    FROM answer, survey_response, survey
+    WHERE
+      record_type = 'answer' AND
+      dhis_sync_queue.type = 'update' AND
+      dhis_sync_queue.record_id = answer.id AND
+      answer.survey_response_id = survey_response.id AND
+      survey_response.survey_id = survey.id AND
+      survey_response.end_time > '${MIN_SURVEY_END_TIME}'`);
+
+exports.up = async function (db) {
```
kael89, line 96: This migration is based on [20200218215230-ResyncAllTongaData.js](https://github.com/beyondessential/tupaia/blob/dev/packages/database/src/migrations/20200218215230-ResyncAllTongaData.js)

kael89:
I've done a bit of testing and confirm this works, at least partially, as follows:

Feature instance: https://2160-missing-data-for-daily-survey.tupaia.org/
Aggregation instance: https://2160-aggregation.tupaia.or

1. Run `yarn migrate-down && yarn migrate`, so that `meditrak-server` can pick up the changes in this migration
2. Using a DB query, set priority from 1 => 2 for all current syncable records
3. Using a DB query, set priority from 2 => 1 for survey responses and answers for the survey `DS` since '2021-01-14'
4. Updated the analytics tables

packages/database/src/utilities/markRecordsForResync.js
```diff
@@ -0,0 +1,29 @@
+/**
```
edmofro, line 1: Great util!

packages/database/src/migrations/20210120051444-ResyncSubmissionsAfterRelease70-modifies-data.js
```diff
@@ -0,0 +1,99 @@
...
+  const { rows } = await db.runSql(`
+    SELECT answer.*
+    FROM answer
+    JOIN survey_response ON answer.survey_response_id = survey_response.id
+    JOIN survey ON survey_response.survey_id = survey.id
+    LEFT JOIN dhis_sync_queue ON answer.id = record_id
+    WHERE
+      survey_response.end_time > '${MIN_SURVEY_END_TIME}' AND
+      survey.can_repeat = false AND -- repeating surveys are event based, don't need answers in sync queue
+      survey.code NOT IN ('CD3a', 'CD3b') AND -- non org unit entity surveys, event based
```
edmofro, line 44: There are more of these now I believe

packages/database/src/migrations/20210120051444-ResyncSubmissionsAfterRelease70-modifies-data.js
```diff
@@ -0,0 +1,99 @@
...
+    // publish changes for every survey response and answer that isn't already in
+    // the sync queue since Release 70
+    // n.b. this requires a meditrak-server instance to be running and listening for the changes
+    const surveyResponses = await selectSurveyResponsesForResync(db);
+    await markRecordsForResync(changeChannel, 'survey_response', surveyResponses);
+    const answers = await selectAnswersForResync(db);
+    await markRecordsForResync(changeChannel, 'answer', answers);
+
+    await updateSyncQueueForSurveyResponses(db);
+    await updateSyncQueueForAnswers(db);
```
edmofro, line 87: I'm not across why these are required...?

packages/database/src/migrations/20210120051444-ResyncSubmissionsAfterRelease70-modifies-data.js
edmofro: Although it's fairly safe as marking for resync still follows normal protocol so they will be filtered out before making it to the queue. Just a potential optimisation I believe.

edmofro (commented):
Looks good - just one question about why we need to update the DHIS sync queue entries?

packages/database/src/utilities/markRecordsForResync.js
kael89: You wrote it actually - I just extracted it! 🙌

packages/database/src/migrations/20210120051444-ResyncSubmissionsAfterRelease70-modifies-data.js
kael89: See the comment in the actual functions:

> -- use priority 2 not to block up new data syncing through

So this is not strictly necessary, just a bit of prioritisation logic

packages/database/src/utilities/markRecordsForResync.js
edmofro: Nice extraction - it's a good idea as it will probably be used again in future :-)

kael89:
@edmofro fyi you may have missed this comment: https://github.com/beyondessential/tupaia/pull/1927#discussion_r560711578

packages/database/src/migrations/20210120051444-ResyncSubmissionsAfterRelease70-modifies-data.js
edmofro: Aaah right - makes sense. Sorry, I did read that bit but didn't connect the dots!

edmofro (approved):
Approved! Thanks for putting this together and testing, I know it can be gruelling work

Potential improvement to list out newer non-org-unit-primary-entity surveys, but very much approved as is :-)

kael89:
> Approved! Thanks for putting this together and testing, I know it can be gruelling work
> 
> Potential improvement to list out newer non-org-unit-primary-entity surveys, but very much approved as is :-)

Another imporevement could be filtering out surveys using tupaia as a data service, but I did a bit of a lookup at the very vast majority of responses in the previous week were for dhis surveys, so I didn't bother

kael89:
With a quick lookup, event based surveys are about 10% of last week's responses, so given the urgency of this I'll leave it as it

@edmofro this is currently a hotfix against master, are you fine with this? Also, are you comfortable with it being merged without further testing? Most on my time in this card was spent was on testing. I believe that in the worse case this won't do what it's intended for, rather than break something

edmofro:
Yes I feel the same - I'm fine with merging straight to master

PR #1931, opened 2021-01-21
Port old surveyScreenComponent edit handling to new system

EMcQ-BES:
### Changes:

- Really just copy pasted the old code into the new handler. Double checked the permissions between questions and surveyScreenComponents and anyone who has access to edit the ssc would also have access to edit the attached question, so I did not add any further checks.


edmofro (approved):
Huh, this is pretty breaking of the REST API model, but if it was already that way then what the hell...it probably makes the admin panel code nice and clean.

PR #1935, opened 2021-01-22
Add missing dependency

kael89:
### Issue #:
[No issue] Thanks ESLint for pointing out this issue! The existing code just happens to work in practice because `meditrak-server` is installed in the same location as other packages, which use `lodash`... But it wouldn't work if `meditrak-server` was a standalone package


chris-bes (approved):

packages/meditrak-server/package.json
```diff
@@ -23,12 +23,12 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
-    "@tupaia/expression-parser": "1.0.0",
     "@tupaia/access-policy": "3.0.0",
     "@tupaia/auth": "1.0.0",
     "@tupaia/data-broker": "1.0.0",
     "@tupaia/database": "1.0.0",
     "@tupaia/dhis-api": "1.0.0",
+    "@tupaia/expression-parser": "1.0.0",
```
chris-bes, line 31: alphabetical!

kael89:
Will merge this straight into `dev` after the build succeeds, and do a quick smoke test there

PR #1936, opened 2021-01-22
Migration to add Strive Entity Relations for Districts

nick-beyondessential:
### Changes:
- Set Strive to use entity_hierarchy for district definition so that when we add FEPT districts and sub_districts don't interfere with Strive.
- Remove sub_district from Hierarchy canonical_types 
- add FETP project canonical types
- Trim name and code of entity 'Milne Bay'
- Remove STRIVE dashbaords and mapOverlays from 'explore'


packages/database/src/migrations/20210121213045-MigrateStriveCanonicalToAlternateHeirarchy-modifies-data.js
```diff
@@ -0,0 +1,94 @@
...
+  const facilities = (await selectEntities(db, 'facility')).rows;
+  await addEntityRelations(db, facilities);
+
+  const villages = (await selectEntities(db, 'village')).rows;
+  await addEntityRelations(db, villages);
+
+  const cases = (await selectEntities(db, 'case')).rows;
+  await addEntityRelations(db, cases);
+
+  await updateValues(db, 'entity_hierarchy', { canonical_types: {} }, { id: striveHierarchyId });
```
nick-beyondessential, line 77: @edmofro is this correct - we ewant to remove canonical types from Strive when creating as alternate_heirarchy?

edmofro: I probably need to understand the context a bit more, but my initial reaction is "no".

`canonical_types` indicates which types should be automatically included in the hierarchy. If you remove that, how will new cases be linked in?

Further info: even if you add `case` into the `canonical_types`, by adding the cases explicitly as relations, you're telling the hierarchy traverser that this level in the hierarchy has been manually taken care of, so it won't bother looking to include any automatically.

Why is this being done?

Happy to have a call on this if you want?

nick-beyondessential: In summary from call:
All hierarchies contain a mixture of 'canonical' and 'alternate' relations. `entity_relation` entries entered on a type 'generational' basis where divergent from canonical rather than an all or nothing.
`canonical_types` should remain for Strive, removing `sub_district` which do not exist for Strive.
`entity_relation` entries should be added for the existing 8 districts county->district entities. The rest will continue to function as before.
When importing districts (Provinces) and sub_districts (Districts) for FETP project [#1671](https://github.com/beyondessential/tupaia-backlog/issues/1671) `canonical_types` for FETP hierarchy will be 'country,district,sub_district,individual` and project->country relation defined as `entity_relation`

billli0 (approved):
Looks good!

nick-beyondessential:
Merging into dev to test DHIS2 sync

PR #1937, opened 2021-01-22
1789 fix deep linking

tcaiger:
### Changes:

- Add permission check when loading dashboard groups


tcaiger:
@edmofro - can you please do another code review on the latest commit. Bindu found an issue where the landing modal was sometimes showing underneath the dashboard modal after logging in with a nested url that was part of the explore project but had a specific dashboard group. The fix addresses that issue.

tcaiger:
ps. I accidentally merged the first PR into dev before it was tested (after the release) which is why there is a new PR.

edmofro (approved):

PR #1939, opened 2021-01-23
2165: Added support for labels in Binary questions

chris-bes:
### Changes:

- Allow importing `optionLabels` for Binary questions. No options/values are expected, they will be `Yes/No`
- The labels have to be in order for `Yes/No`
- The options column in Question table for Binary questions will look like below:
```
{"{\"label\":\"ໄດ້ຮັບ Yes\"}","{\"label\":\"ບໍ່ໄດ້ຮັບ No\"}"}
```

or if no labels/colors defined, it will be:

```
{}
```


packages/meditrak-server/src/routes/importSurveys/importSurveys.js
```diff
@@ -389,7 +389,26 @@ async function processOptionSetName(models, name) {
...
+  }
+};
+
+const processBinaryOptions = (optionLabelsString, optionColorsString) => {
+  const optionLabels = splitOnNewLinesOrCommas(optionLabelsString);
+  const optionColors = splitOnNewLinesOrCommas(optionColorsString);
+  return optionLabels.map((label, i) => {
+    const color = optionColors[i];
+
+    return { label, color };
```
edmofro, line 407: What about adding `value` here, and making the code in meditrak-app simpler? That way instead of mapping through ['Yes', 'No'], you could just pass the options array created here directly.

packages/meditrak-server/src/routes/exportSurveys/RowBuilder.js
```diff
@@ -53,7 +53,6 @@ export class RowBuilder {
     const processedOptions = options.map((option, optionIndex) => {
       try {
         const { value, label, color } = JSON.parse(option);
-        if (value === undefined) return option; // May be JSON but still not a formatted option object
```
edmofro, line 56: Is this case no longer possible?

packages/meditrak-app/app/assessment/specificQuestions/BinaryQuestion.js
```diff
@@ -4,7 +4,26 @@
  */
 
 import React from 'react';
-
+import PropTypes from 'prop-types';
 import { RadioQuestion } from './RadioQuestion';
 
-export const BinaryQuestion = props => <RadioQuestion options={['Yes', 'No']} {...props} />;
+export const BinaryQuestion = ({ options, ...props }) => {
+  const binaryOptions = options.length
```
edmofro, line 11:
```suggestion
  const hasCustomOptions = !!options.length;
  const binaryOptions = hasCustomOptions
```
Could make things slightly more obvious

edmofro (changes requested):
A couple of minor questions for you

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
chris-bes: I was thinking of doing the same, but then I chose the current way because I kinda didn't want to store the `value` Yes/No in the `options` column (which can be changed through the admin panel). Is that a good enough reason or maybe I was just overthinking? But I agree your suggestion will make things much easier!

chris-bes: Never mind I was definitely overthinking, will follow your suggest!

edmofro: That's a good thought...probably best solution would be validation on editing options through the admin panel. Let's just leave it for now though, because that is sure to create conflicts with Ethan's work to add the api level permissions

packages/meditrak-server/src/routes/exportSurveys/RowBuilder.js
```diff
@@ -45,19 +45,28 @@ export class RowBuilder {
...
+    const processedOptions = [];
+    options.forEach((option, optionIndex) => {
       try {
         const { value, label, color } = JSON.parse(option);
-        optionLabels[optionIndex] = label;
-        optionColors[optionIndex] = color;
-        return value;
+
+        // Do not export any options for Binary questions because they have fixed options (Yes/No)
+        if (restOfRowData.type !== 'Binary') {
```
edmofro, line 57: Nice catch 👏

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
```diff
@@ -389,53 +389,44 @@ async function processOptionSetName(models, name) {
   return null;
 }
 
-const processOptions = (optionValuesString, optionLabelsString, optionColorsString, type) => {
+const processOptionValues = (optionValuesString, type) => {
   switch (type) {
     case 'Binary':
-      return processBinaryOptions(optionLabelsString, optionColorsString);
+      return ['Yes', 'No'];
```
edmofro, line 395: Ah nice, it even simplifies this bit!

packages/meditrak-server/src/routes/exportSurveys/RowBuilder.js
edmofro: Just this question pending @chris-bes

chris-bes: Ah sorry, I was going to make comments once I've fixed up all the reviews. Yep I don't think this is possible anymore. Had a search in the `question` database and there's no cases that an `option` is in JSON format but has no `value` property. So I think it's safe to remove it

edmofro: Great! Sorry, I jumped on this before you requested re-review - bad form.

edmofro (approved):

packages/data-api/src/TupaiaDataApi.js
```diff
@@ -179,6 +179,11 @@ export class TupaiaDataApi {
 
   buildOptionsMetadata = (options = [], type) => {
     const optionList = options.length === 0 && type === 'Binary' ? DEFAULT_BINARY_OPTIONS : options;
+
+    if (!optionList.length) {
+      return undefined;
+    }
+
```
chris-bes, line 186: Just to fix this bug here (https://github.com/beyondessential/tupaia-backlog/issues/1322#issuecomment-777075175). This is because in `tableOfDataValues` data builder, it doesn't check if `options` metadata is empty or not to get the value (below)
options

So I just return undefined here if there are no options, I think this also aligns with how DHIS2 return options metadata

edmofro (approved):

packages/database/src/migrations/20210124055714-ConvertDhisSurveysIntoTupaia-modifies-data.js
```diff
@@ -80,6 +87,15 @@ exports.up = async function (db) {
...
   }
 
+  // Have to convert these reports to use 'Yes' because the data elements are Condition questions
+  dashboardReportsChangeForValueOfInterest.forEach(async id => {
+    await db.runSql(`
+      update "dashboardReport" dr 
+      set "dataBuilderConfig" = regexp_replace(dr."dataBuilderConfig"::text, '\\"valueOfInterest\\"\\: 1','"valueOfInterest": "Yes"','g')::jsonb 
+      where id = '${id}'
+  `);
+  });
```
chris-bes, line 97: `RHS6UNFPA1354` and `RHS6UNFPA1355` have been converted from `Binary` to `Condition` questions last week (https://github.com/beyondessential/tupaia-backlog/issues/1359). Those 2 questions are now hidden and generate Yes/No automatically. 

Since they're no longer `Binary`, they have not been sanitized to `1/0`, causing this bug here (https://github.com/beyondessential/tupaia-backlog/issues/1322#issuecomment-778969387).

I feel that this is probably a special case and won't happen so often so it's fine to just modify the reports to use `Yes/No` instead of trying to find a way to also uniform them to be `1/0`

edmofro: Darn!

edmofro (approved):
lgtm :-)

PR #1941, opened 2021-01-24
576: Return binary data as 1/0

chris-bes:
### Changes:
- Sanitize internal binary and checkbox data to 1/0 instead of Yes/No
- Migrate all existing vizes using binary internal data as Yes/No to 1/0


packages/data-api/src/TupaiaDataApi.js
```diff
@@ -96,7 +96,7 @@ export class TupaiaDataApi {
     if (dataElementCodes && Array.isArray(dataElementCodes)) {
       const sqlQuery = await new SqlQuery(
         `
-        SELECT question.code, question.name, question.text
+        SELECT question.code, question.name, question.text, question.type
```
chris-bes, line 99: Return question type for Custom Export to check if the data is binary and convert it from `1/0` to `Yes/No` for the values in the exported excel

packages/database/src/migrations/20210123120226-MigrateLaosSchoolsMapOverlaysBinaryDataToZeroOne-modifies-data.js
```diff
@@ -0,0 +1,110 @@
...
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  // Modify config for Laos Schools map overlays, from <"Yes"> to <1> (I've run through all the configs and it is safe to do so)
+  await db.runSql(`
+    update "mapOverlay" mo
+    set "measureBuilderConfig" = regexp_replace(mo."measureBuilderConfig"::text, '"Yes"','1','g')::jsonb
```
chris-bes, line 78: I've checked the overlays above and there are many different variations using `Yes/No` in the config (eg: `valueOfInterest: Yes`, `value: Yes`, `filter: ['Yes',...]...`) So I think it's easier to just replace `"Yes"` with `1`. I've gone through each map overlay and can confirm that it is safe to do so (no special cases with "Yes" that we dont want to replace with `1`)

packages/database/src/migrations/20210123120226-MigrateLaosSchoolsMapOverlaysBinaryDataToZeroOne-modifies-data.js
```diff
@@ -0,0 +1,110 @@
...
+    where id in (${arrayToDbString(MAP_OVERLAY_IDS)})
+  `);
+
+  // Modify config for Laos Schools map overlays, from <"No"> to <0> (I've run through all the configs and it is safe to do so)
+  await db.runSql(`
+    update "mapOverlay" mo
+    set "measureBuilderConfig" = regexp_replace(mo."measureBuilderConfig"::text, '"No"','0','g')::jsonb
```
chris-bes, line 85: Same as the comment above. Can confirm it is safe to replace all `"No"` with `0`

packages/database/src/migrations/20210123120226-MigrateLaosSchoolsMapOverlaysBinaryDataToZeroOne-modifies-data.js
```diff
@@ -0,0 +1,110 @@
...
+    where id in (${arrayToDbString(MAP_OVERLAY_IDS)})
+  `);
+
+  // Modify presentation options for Laos Schools map overlays, from <"Yes"> to <1>
+  await db.runSql(`
+    update "mapOverlay" mo
+    set "presentationOptions" = regexp_replace(mo."presentationOptions"::text, '"value": "Yes"','"value": 1','g')::jsonb
```
chris-bes, line 92: Have gone through all the map overlays above and can confirm this regex replacement is safe

packages/database/src/migrations/20210123120226-MigrateLaosSchoolsMapOverlaysBinaryDataToZeroOne-modifies-data.js
```diff
@@ -0,0 +1,110 @@
...
+    where id in (${arrayToDbString(MAP_OVERLAY_IDS)})
+  `);
+
+  // Modify presentation options for Laos Schools map overlays, from <"No"> to <1>
+  await db.runSql(`
+    update "mapOverlay" mo
+    set "presentationOptions" = regexp_replace(mo."presentationOptions"::text, '"value": "No"','"value": 0','g')::jsonb
```
chris-bes, line 99: Have gone through all the map overlays above and can confirm this regex replacement is safe

packages/database/src/migrations/20210124013906-FixLaosSchoolsDevelopmentPartnerSupportMapOverlay-modifies-data.js
```diff
@@ -0,0 +1,42 @@
...
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`
+    update "mapOverlay" mo
+    set "measureBuilderConfig" = regexp_replace(mo."measureBuilderConfig"::text, '"valueToMatch": "Yes"','"condition": 1','g')::jsonb
```
chris-bes, line 27: These overlays actually have the wrong config so this is just to fix an existing bug. To apply conditions on entityAggregation, we need to use property `condition` instead of `valueToMatch`

packages/database/src/migrations/20210124014513-UpdateLaosSchoolsDevPartnersMapOverlays-modifies-data.js
```diff
@@ -0,0 +1,95 @@
...
+  'Laos_Schools_Dev_Partner_SchDP_HII',
+  'Laos_Schools_Dev_Partner_SchDP_Plan',
+  'Laos_Schools_Dev_Partner_SchDP_RtR',
+  'Laos_Schools_Dev_Partner_SchDP_UNIC',
+  'Laos_Schools_Dev_Partner_SchDP_WB',
+  'Laos_Schools_Dev_Partner_SchDP_WC',
+  'Laos_Schools_Dev_Partner_SchDP_WFP',
+  'Laos_Schools_Dev_Partner_SchDP_WR',
+  'Laos_Schools_Dev_Partner_SchDP_WV',
+];
```
chris-bes, line 21: It's easier to update `Yes/No` to `1/0` for these map overlays separately because the presentationOptions references `Yes/No` multiple times in different ways

packages/web-config-server/src/apiV1/dataBuilders/generic/dataDownload/rawDataValues.js
```diff
@@ -22,6 +22,15 @@ const expandSurveyCodes = surveys => {
...
 
+const convertValueByType = (type, value) => {
+  switch (type) {
+    case 'Binary':
+    case 'Checkbox':
+      return value === 1 ? 'Yes' : 'No';
+    default:
+      return value;
+  }
+}
```
chris-bes, line 33: Convert `1/0` to `Yes/No` because we actually want to show `Yes/No` values for Binary in custom export

packages/data-api/src/utils.js
```diff
@@ -4,10 +4,15 @@
...
-    const sanitizedValue = parseFloat(value);
-    return Number.isNaN(sanitizedValue) ? '' : sanitizedValue;
+  switch (type) {
+    case 'Number': {
+      const sanitizedValue = parseFloat(value);
+      return Number.isNaN(sanitizedValue) ? '' : sanitizedValue;
+    }
+    case 'Binary':
+    case 'Checkbox':
+      return value === 'Yes' ? 1 : 0;
```
chris-bes, line 14: Actual sanitization for internal data to change `Yes/No` to `1/0`

packages/web-config-server/src/tests/apiV1/dataBuilders/modules/covid-samoa/flight/flightAgeRanges.test.js
```diff
@@ -62,16 +62,16 @@ describe('flightAgeRanges', () => {
 
     flight.events = [
       {
-        dataValues: { QMIA031: 'Yes' },
+        dataValues: { QMIA031: 1 },
```
chris-bes, line 65: This is for Samoa visualisations

packages/data-api/src/TupaiaDataApi.js
edmofro: This creates a tighter coupling between our web-config-server client and the internal data-api. Our intention is that the data services would all be able to support the same response format, so that any client's code can be performed on results from any data service. So even though it's for the `rawDataDownloads` which is kind of related to meditrak, we would be best to avoid this. 

Could you check out how we map options to their labels with internal and/or DHIS2 data? That might reveal a more loosely coupled pattern

packages/web-config-server/src/apiV1/dataBuilders/generic/dataDownload/rawDataValues.js
edmofro: Would prefer to do this in a different way. I think the way we'd do it with DHIS2 at the moment is to look at the data element metadata, which can include an `options` field with each. Do a find on `includeOptions` as a starting point. Not saying this is perfect - but we'd be best to keep consistent where we can

packages/database/src/migrations/20210124013906-FixLaosSchoolsDevelopmentPartnerSupportMapOverlay-modifies-data.js
edmofro: Another great catch

packages/database/src/migrations/20210124020024-UpdateLaosSchoolsSpecialMapOverlays-modifies-data.js
```diff
@@ -0,0 +1,357 @@
+'use strict';
```
edmofro, line 1: Full disclosure: I didn't check this file thoroughly as it's pretty big

edmofro (approved):
LGTM! I don't know all of the Laos visualisations, but it looks like you've done a very thorough job here 🙌 

Can you confirm there aren't any Binary/Checkbox questions that use the `tupaia` data source outside of Laos?

packages/web-config-server/src/apiV1/dataBuilders/generic/dataDownload/rawDataValues.js
chris-bes: Thanks Edwin for pointing that out, I agree that we should keep it consistent!

Just checking if I'm in the right direction. So after having a look at how dhis2 does `includeOptions`, it returns something like below when we do fetchDataElements:
```
[
   {
     options: {
       Church: 'Church',
       Workplace: 'Workplace',
       School: 'School',
       Community: 'Community'
     },
     code: 'HP31n',
     name: 'HP NCD Risk Factor Screening: Setting Type',
     id: 'fvmaQAPl5bI'
   }
]
```

I think we can do the same and it can return something like (with actual value keyed by sanitized value) :
```
[
   {
     options: {
       1: 'Yes',
       0: 'No',
     },
     code: 'ABC',
     ...
   }
]
```

edmofro: Yes exactly. This will probably also result in the Laos ones becoming Laos text...but maybe that's a good thing!

edmofro: > With that, binary questions with no labels will also have to have their options defined as {Yes,No}

You mean by the code rather than in the database right? Like we can default to that in the data fetching logic?

chris-bes:
@edmofro 

> Can you confirm there aren't any Binary/Checkbox questions that use the tupaia data source outside of Laos?

Outside of Laos, there are some Binary/Checkbox question/data_sources using `tupaia` service type in:
-  WISH Fiji: We only have Raw data downloads for this so I've sanity checked and they are still working
- Covid-19 Samoa: I've updated the visualisations that belong to this as part of this PR and also done a sanity check
- Demo Land

Beside those mentioned above, there are 3 left (screenshot here: https://github.com/beyondessential/tupaia-backlog/issues/576#issuecomment-768747550). Not that I know of we have to do anything with them?

Also, PR is ready to reviewed again! I just need to fix some failing tests

packages/data-api/src/TupaiaDataApi.js
```diff
@@ -55,21 +56,23 @@ export class TupaiaDataApi {
...
-  async fetchDataElements(dataElementCodes) {
+  async fetchDataElements(dataElementCodes, { includeOptions = true }) {
     if (!dataElementCodes || !Array.isArray(dataElementCodes)) {
       throw new Error('Please provide an array of data element codes');
     }
-    return new SqlQuery(
+    const sqlQuery = new SqlQuery(
       `
-      SELECT code, name
+      SELECT code, name, type
```
edmofro, line 65: Does this need to fetch `options` as well?

edmofro (changes requested):
Unfortunately there are two sources of options:
- `question.options`
- `question.option_set_id`
Do you think we should support the second type as well? I know it's not required right now, but it could confuse someone in future...? This is really blowing out sorry Chris!

packages/data-api/src/TupaiaDataApi.js
chris-bes: oops...

chris-bes:
> Unfortunately there are two sources of options:
> 
> * `question.options`
> * `question.option_set_id`
>   Do you think we should support the second type as well? I know it's not required right now, but it could confuse someone in future...? This is really blowing out sorry Chris!

Ahh I should have thought of that. Nah all good! I don't think that's too much work, happy to support the 2nd type as well 😸

edmofro:
Will wait this time until you re-request review :-)

packages/data-api/src/TupaiaDataApi.js
```diff
@@ -129,30 +131,69 @@ export class TupaiaDataApi {
...
+      const optionSetIds = [
+        ...new Set(dataElementsMetadata.filter(d => !!d.option_set_id).map(d => d.option_set_id)),
+      ];
+      // Get all the options from the option sets and grouped by set ids.
+      const optionsGroupedBySetId = await this.getOptionsGroupedBySetId(optionSetIds);
+
+      return dataElementsMetadata.map(
+        ({ options, type, option_set_id: optionSetId, ...restOfMetadata }) => {
+          // In reality, the options can only come from either question.options or question.option_set_id
+          const allOptions = [...options, ...(optionsGroupedBySetId[optionSetId] || [])];
```
edmofro, line 146: Very succinct 👏

packages/data-api/src/TupaiaDataApi.js
```diff
@@ -129,30 +131,69 @@ export class TupaiaDataApi {
...
   buildOptionsMetadata = (options = [], type) => {
     const optionList = options.length === 0 && type === 'Binary' ? DEFAULT_BINARY_OPTIONS : options;
     const optionsMetadata = {};
+
     optionList.forEach(option => {
       try {
-        const { value, label } = JSON.parse(option);
+        // options coming from question.options are JSON format strings
+        // options coming from option_set are actual JSON objects
+        const optionObject = typeof option === 'string' ? JSON.parse(option) : option;
```
edmofro, line 188: This function doesn't otherwise know about the source of options, so I'd prefer to keep it ignorant. Could you instead (in the succinct line earlier) `map` and `parse` the options at the point they're gathered from each source?

edmofro (approved):
Happy as with this! One optional suggestion

PR #1942, opened 2021-01-24
1322: Added Custom Export for UNFPA

chris-bes:
### Changes:
- Added Custom Export for UNFPA
- Convert UNFPA surveys from dhis to tupaia
- Fixed up a few incorrect vizes

### DIFF
--- a/packages/aggregator/src/analytics/aggregateAnalytics/aggregations/entityAggregations.js
+++ b/packages/aggregator/src/analytics/aggregateAnalytics/aggregations/entityAggregations.js
@@ -3,7 +3,7 @@
  * Copyright (c) 2019 Beyond Essential Systems Pty Ltd
  */
 
-import { checkValueSatisfiesCondition } from '@tupaia/utils';
+import { checkValueSatisfiesCondition, convertToPeriod } from '@tupaia/utils';
 
 /**
  * Combines analytics together across org units
@@ -14,20 +14,25 @@ import { checkValueSatisfiesCondition } from '@tupaia/utils';
  * @param {Function} keyMapper
  */
 const valuePerOrgGroup = (analytics, aggregationConfig, valueMapper, keyMapper) => {
-  const { orgUnitMap = {}, condition } = aggregationConfig;
+  const { orgUnitMap = {}, condition, periodType } = aggregationConfig;
   const filterValueMapper = condition ? createFilterValueMapper(condition) : () => true;
 
   const analyticsByKey = {};
-  analytics.forEach(analytic => {
+  analytics.forEach(originalAnalytic => {
     const organisationUnit =
-      (orgUnitMap[analytic.organisationUnit] && orgUnitMap[analytic.organisationUnit].code) ||
-      analytic.organisationUnit;
-    const key = keyMapper(analytic, organisationUnit);
+      (orgUnitMap[originalAnalytic.organisationUnit] &&
+        orgUnitMap[originalAnalytic.organisationUnit].code) ||
+      originalAnalytic.organisationUnit;
+    const convertedPeriod = periodType
+      ? convertToPeriod(originalAnalytic.period, periodType)
+      : originalAnalytic.period;
+    const analytic = { ...originalAnalytic, organisationUnit, period: convertedPeriod };
+    const key = keyMapper(analytic);
     const value = filterValueMapper(analytic.value) ? valueMapper(analytic.value) : 0;
 
     // If there are no matching response elements already being returned, add it
     if (!analyticsByKey[key]) {
-      analyticsByKey[key] = { ...analytic, value, organisationUnit };
+      analyticsByKey[key] = { ...analytic, value };
     } else {
       analyticsByKey[key].value += value;
     }
@@ -78,8 +83,7 @@ const sumValueMapper = value => value;
 
 const countValueMapper = () => 1;
 
-const perOrgGroupKey = (analytic, organisationUnit) =>
-  `${analytic.dataElement}__${organisationUnit}`;
+const perOrgGroupKey = analytic => `${analytic.dataElement}__${analytic.organisationUnit}`;
 
-const perPeriodPerOrgGroupKey = (analytic, organisationUnit) =>
-  `${analytic.period}__${analytic.dataElement}__${organisationUnit}`;
+const perPeriodPerOrgGroupKey = analytic =>
+  `${analytic.period}__${analytic.dataElement}__${analytic.organisationUnit}`;
--- a/packages/aggregator/src/analytics/aggregateAnalytics/aggregations/entityAggregations.js
+++ b/packages/aggregator/src/analytics/aggregateAnalytics/aggregations/entityAggregations.js
@@ -3,7 +3,7 @@
  * Copyright (c) 2019 Beyond Essential Systems Pty Ltd
  */
 
-import { checkValueSatisfiesCondition, convertToPeriod } from '@tupaia/utils';
+import { checkValueSatisfiesCondition } from '@tupaia/utils';
 
 /**
  * Combines analytics together across org units
@@ -14,25 +14,20 @@ import { checkValueSatisfiesCondition, convertToPeriod } from '@tupaia/utils';
  * @param {Function} keyMapper
  */
 const valuePerOrgGroup = (analytics, aggregationConfig, valueMapper, keyMapper) => {
-  const { orgUnitMap = {}, condition, periodType } = aggregationConfig;
+  const { orgUnitMap = {}, condition } = aggregationConfig;
   const filterValueMapper = condition ? createFilterValueMapper(condition) : () => true;
 
   const analyticsByKey = {};
-  analytics.forEach(originalAnalytic => {
+  analytics.forEach(analytic => {
     const organisationUnit =
-      (orgUnitMap[originalAnalytic.organisationUnit] &&
-        orgUnitMap[originalAnalytic.organisationUnit].code) ||
-      originalAnalytic.organisationUnit;
-    const convertedPeriod = periodType
-      ? convertToPeriod(originalAnalytic.period, periodType)
-      : originalAnalytic.period;
-    const analytic = { ...originalAnalytic, organisationUnit, period: convertedPeriod };
-    const key = keyMapper(analytic);
+      (orgUnitMap[analytic.organisationUnit] && orgUnitMap[analytic.organisationUnit].code) ||
+      analytic.organisationUnit;
+    const key = keyMapper(analytic, organisationUnit);
     const value = filterValueMapper(analytic.value) ? valueMapper(analytic.value) : 0;
 
     // If there are no matching response elements already being returned, add it
     if (!analyticsByKey[key]) {
-      analyticsByKey[key] = { ...analytic, value };
+      analyticsByKey[key] = { ...analytic, value, organisationUnit };
     } else {
       analyticsByKey[key].value += value;
     }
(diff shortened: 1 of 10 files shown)


packages/aggregator/src/analytics/aggregateAnalytics/aggregations/entityAggregations.js
```diff
@@ -14,20 +14,25 @@ import { checkValueSatisfiesCondition } from '@tupaia/utils';
  * @param {Function} keyMapper
  */
 const valuePerOrgGroup = (analytics, aggregationConfig, valueMapper, keyMapper) => {
-  const { orgUnitMap = {}, condition } = aggregationConfig;
+  const { orgUnitMap = {}, condition, periodType } = aggregationConfig;
```
chris-bes, line 17: Added periodType to allow converting to desired periods when doing entityAggregation

packages/database/src/migrations/20210124073946-FixUNFPAPercentagesOfFacilitiesOfferingServicesReports-modifies-data.js
```diff
@@ -0,0 +1,79 @@
...
+
+const dashboardReportsConfig = {
+  ids: [
+    'UNFPA_Region_Percentage_Facilities_Offering_Services_ANC',
+    'UNFPA_Region_Percentage_Facilities_Offering_Services_Delivery',
+    'UNFPA_Region_Percentage_Facilities_Offering_Services_Family_Planning',
+    'UNFPA_Region_Percentage_Facilities_Offering_Services_PNC',
+  ],
+  oldDataBuilder: 'sumValuesPerQuarterByOrgUnit',
+  newDataBuilder: 'finalValuesPerQuarterByOrgUnit',
```
chris-bes, line 39:
```
'UNFPA_Region_Percentage_Facilities_Offering_Services_ANC',
'UNFPA_Region_Percentage_Facilities_Offering_Services_Delivery',
'UNFPA_Region_Percentage_Facilities_Offering_Services_Family_Planning',
'UNFPA_Region_Percentage_Facilities_Offering_Services_PNC',
```

These dashboards returns wrong data after converting to `tupaia` data. They should be taking final values per quarter for a facility instead of summing all the values during a quarter for a facility. (Original ticket for these dashboards: https://github.com/beyondessential/tupaia-config-server/issues/1166)

packages/web-config-server/src/export/ExportSurveyDataHandler.js
```diff
@@ -76,8 +77,10 @@ export class ExportSurveyDataHandler extends RouteHandler {
         origin: 'A2',
       });
 
-      sheetNames.push(surveyName);
-      sheets[surveyName] = sheet;
+      const sheetName = truncateString(surveyName, 31);
```
edmofro, line 80: Nice 🙌

packages/database/src/migrations/20210124055714-ConvertDhisSurveysIntoTupaia-modifies-data.js
```diff
@@ -0,0 +1,109 @@
...
+exports.down = async function (db) {
+  const { rows: dataSources } = await getDataSources(db, surveyCodes);
+  const serviceType = 'dhis';
+
+  for (const dataSource of dataSources) {
+    await switchServiceType(db, dataSource.code, serviceType);
+  }
+
+  for (const code of surveyCodes) {
+    await switchServiceType(db, code, serviceType);
```
edmofro, line 96: Safer to make sure the only ones being switched are of type `data_group`, in case there was e.g. a survey with the code `BCD` and a question from a different survey also with the code `BCD`

edmofro: Same goes for the other places above

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/entityAggregations.js
edmofro: Can we support this by using a series of aggregations instead? It seems unrelated to the entity aggregation process.

packages/database/src/migrations/20210124073946-FixUNFPAPercentagesOfFacilitiesOfferingServicesReports-modifies-data.js
edmofro: Great catch Chris

edmofro (changes requested):
Main question is about combining the two aggregations

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/entityAggregations.js
chris-bes: True it does seem unrelated to entity aggregation! So there are 2 aggregations performed in these dashboards (listed here: https://github.com/beyondessential/tupaia/pull/1942#discussion_r563420278):
- FINAL_EACH_QUARTER
- COUNT_PER_PERIOD_PER_ORG_GROUP (entity aggregation)

I just need the `period` of the analytics to be in `quarter` format before entityAggregation is performed, so that' why I tried to convert the periods by adding `periodType` here.. Perhaps I can do that in the `FINAL_EACH_QUARTER` aggregation instead?

edmofro: @kael89 you had some thoughts on converting periods to their correct formats during aggregation right? I believe doing it in `FINAL_EACH_QUARTER` would fit with what you were thinking.

kael89: Definitely, it seems that most existing aggregations are not concerned with changing period formats, although their name would imply so. It also a blocker for using most `FINAL_EACH...` aggregations with `indicators`.

Card: https://github.com/beyondessential/tupaia-backlog/issues/2150#issue-788619473 - 4th point ("Make periods returned...")

---

My only other comment is that `FINAL_EACH_WEEK` works correctly, since it **has** to convert periods to the special week format (eg 2021W3), before comparing them. This is not a requirement for `FINAL_EACH_DAY`, `FINAL_EACH_MONTH`, `FINAL_EACH_YEAR`, which are just an extension one another (2020 => 202001 => 20200101), and can use a single string comparison.

Based on the above observation, I would expect that `FINAL_EACH_QUARTER` also **has** to convert periods to its custom format (2021Q1), but I may be wrong.

chris-bes: > most existing aggregations are not concerned with changing period formats, although their name would imply so

I did have the same thought... I'm very tempted to make this global change for `FINAL_EACH...` aggregations. However, I'm also afraid of the amount of regression testing... (given that this card already requires a lot of regression test). In my opinion, Having a config seems like the best option for now?

edmofro: I reckon the fact that this card already requires regression test is an argument _for_ changing it now. I'm not sure how likely regressions are though, being unfamiliar with the data building code these days - do you know whether there are many places that assume the format of a period? @kael89, do you?

kael89: Possibly [getPreferredPeriod](https://github.com/beyondessential/tupaia/blob/dev/packages/aggregator/src/analytics/aggregateAnalytics/aggregations/utils/getPreferredPeriod.js)... I believe that there are a few reports where is supplied by both `mSpply` and `meditrak`. The one set of data uses a daily period, and the other a monthly. We prefer one over the other for this specific report.

If those reports use `FINAL_EACH` aggregations, and those aggregations change the period formats, the above comparison may no longer be possible

If this functionality is indeed desirable, I suggest adding an optional config option to `FINAL_EACH`, eg 
```js
{ useOriginalPeriodType: true }
// or
{ retainPeriodType: true }
```

chris-bes: Sorry for my late reply, I was interrupted with some other tasks :(. 

So I've had a search on the reports and found at least 1 report that uses both `mSupply` and `meditrak` data element codes and also use `FINAL_EACH_...` (id: `UNFPA_Delivery_Services_Stock`). Possibly in the future we can have more? So I'm keen to follow Kostas' suggestion

kael89: Thanks for the investigation @chris-bes , did you also confirm that this report's functionality depends on each source using different period types in analytics?

chris-bes: That's a good point... Now that I've checked, the data coming from `mSupply` does have daily period rather than monthly period that we expected. Could it be that the monthly period doesn't apply to all mSupply data elements?
msupply

chris-bes: Just had more investigations on this. Perhaps the reports that @kael89 was talking about are the ones that use `Medicines Availability Aggregation`. They use a preaggregated element, so that's why I failed to search for it in the first place. To pre aggregate the data elements, they use `FINAL_EACH_MONTH_PREFER_DAILY_PERIOD` aggregation. So I think this is the case...

kael89: That must be it @chris-bes !!

edmofro: It seems quite isolated - the meds availability preaggregation is the only place I know of where we use the difference in period types as information, and it seems it's built in to the aggregation type so will be able to happen before the conversion to uniform period type. Does this mean we can go with consistently converting to the expected period type, rather than using a config option? I'm nervous of the proliferation of config options!

kael89: I agree we should avoid adding an extra option only for a specific use case. Even if we leverage the difference of period types in the input, we can still sanitize all periods in the output according to the `FINAL_EACH....` period type.

The only problem would be if someone runs a `FINAL_EACH_MONTH` aggregation and then a `FINAL_EACH_MONTH_PREFER_DAILY_PERIOD` , but it doesn't seem to be the case in this report.

chris-bes: Thanks guys for your valuable inputs! I can have a try at making that global change (without using config to convert the period types) and will do a smoke test on the reports!

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -57,14 +57,13 @@ class FinalValueCache {
...
     this.cache = cache;
-    if (periodTypes.length > 1) {
-      // If multiple period types exist, make data consistent by converting
-      // all periods to the coarsest type (e.g. for day and month mix, convert to month)
-      this.ensurePeriodConsistency(findCoarsestPeriodType(periodTypes));
-    }
+
+    // Convert all the periods to a uniform aggregation period type
+    // eg: FINAL_EACH_MONTH will convert all periods to monthly periods
+    this.convertCachePeriods(aggregationPeriod);
```
chris-bes, line 63: This is where the period conversion happens. I've smoke test all the affected reports and preaggregators and they seem alright.

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/entityAggregations.js
```diff
@@ -83,7 +78,8 @@ const sumValueMapper = value => value;
...
 const countValueMapper = () => 1;
 
-const perOrgGroupKey = analytic => `${analytic.dataElement}__${analytic.organisationUnit}`;
+const perOrgGroupKey = (analytic, organisationUnit) =>
+  `${analytic.dataElement}__${organisationUnit}`;
 
-const perPeriodPerOrgGroupKey = analytic =>
-  `${analytic.period}__${analytic.dataElement}__${analytic.organisationUnit}`;
+const perPeriodPerOrgGroupKey = (analytic, organisationUnit) =>
+  `${analytic.period}__${analytic.dataElement}__${organisationUnit}`;
```
chris-bes: Just reverting this entityAggregation file back to before cause we don't need `periodType` conversion here anymore

edmofro (approved):
Wahoo! Nice

PR #1944, opened 2021-01-24
1796 samoa custom raw data download

billli0:
### Issue #:
[#1796 Samoa COVID customised raw data download](https://github.com/beyondessential/tupaia-backlog/issues/1796)
[#2201 Customised raw data download doesn't have date range](https://github.com/beyondessential/tupaia-backlog/issues/2201)
### Changes:
- Add new aggregation type `REPLACE_ENTITY_ANSWER_WITH_ENTITY_NAME` and test cases
- Migration
- New function `.getEntityAnswerWithEntityName()` in `Entity` Model
- Support data range for custom data download


packages/database/src/modelClasses/Entity.js
```diff
@@ -374,6 +374,38 @@ export class EntityModel extends DatabaseModel {
     );
   }
 
+  async getEntityAnswerWithEntityName(dataSourceEntities, questionCode) {
```
kael89, line 377: I don't think this method should live in the model, since its concern is very specific. It's better if we move it to a utility function.

Models should be focused on described the basic underlying "truths" about objects stored in the DB, and their relationships (the `model` of our application). The details that sit on top of that are part of the `business` logic, which can change more often and is more related to a specific client's needs (eg connecting entities to survey responses via `Entity`/`PrimaryEntity` questions)

packages/database/src/migrations/20210119053017-SomoaCovidCustomisedRawDataDownload-modifies-data.js
```diff
@@ -0,0 +1,94 @@
...
+    exportDataBuilder: {
+      dataBuilder: 'rawDataValues',
+      dataBuilderConfig: {
+        skipHeader: false,
+        surveysConfig: {
+          [surveyCode]: {
+            entityAggregation: {
+              aggregationType: 'REPLACE_ENTITY_ANSWER_WITH_ENTITY_NAME',
+              aggregationConfig: {
+                questionCodes: ['QMIA029'],
```
kael89, line 40: Instead of `questionCodes`, let's use `dataElementCodes` since it refers to the actual object that will be retrieved

Reasons for that:
* Data element and question codes currently match, but this is not a strict requirement in our code (eg a question code could be theoretically different from a data element code)
* A data element is not necessarily associated with a question: it can be linked to a 3rd party api (`weather` service), an `indicator` etc

packages/web-config-server/src/aggregator/buildAggregationOptions.js
```diff
@@ -84,3 +86,32 @@ const fetchEntityAggregationConfig = async (
     config: { ...entityAggregationConfig, orgUnitMap: entityToAncestorMap },
   };
 };
+
+const replaceEntityAnswerWithEntityNameConfig = async (
+  models,
+  dataSourceEntities,
```
kael89, line 92: I don't think this will work in general... `dataSourceEntities` are built based on the entities selected in the `web-config-server`, in other words the entity selected in the front end.

On the other hand, your code concerns `QMIA029`, which is an `Entity` question. In the general case, this entity could be anywhere in the world! For all we know, a Samoa survey may allow the user to select a city in Australia.

In conclusion, this **will** work but only under specific conditions. What would be more useful instead would be to map any entity id to its name, since we already have this information in the database.

kael89 (changes requested):
Hi @billli0 ,

I think we should use a different approach here. There is no need to use an `aggregation`: we just need a simple id to name mapping, which can happen in the data builder level.

If we want to make this more generic/reusable, we can define it as a util that the data builder uses. I have the impression that similar functionality already exists in data builders, eg for mapping entity codes to names. Who knows, you may even find an `id => name` util also already exists!

packages/database/src/migrations/20210119053017-SomoaCovidCustomisedRawDataDownload-modifies-data.js
kael89: Based on my [comment](https://github.com/beyondessential/tupaia/pull/1944#pullrequestreview-577820540), since this will be used in the `surveyDataExport` builder we probably don't event have to specify which question codes are for an `Entity`. Since this data builder deals with survey export, it should also be able to get information about which survey questions are of `Entity` type, and automatically translate all of them if a `dataBuilderConfig` field is set

packages/database/src/modelClasses/Entity.js
```diff
@@ -374,6 +374,38 @@ export class EntityModel extends DatabaseModel {
...
+    const eventId = 'event_id';
+    const entityAnswerIdToName = 'entity_answer_id_to_name';
+    const entityAnswerIdToNameMapping = await this.database.executeSql(
+      `
+        SELECT sr.id as ${eventId}, e2.name as ${entityAnswerIdToName}
```
kael89, line 382: There are some valid reasons for using raw queries: performance optimisations, specific operations that our `DatabaseModel` API does not support etc.

None of those reasons apply here, so we should use the `models` object which is passed around as an abstraction for Model operations - you'll find a lot of example for this in our codebase

kael89 (changes requested):

packages/database/src/migrations/20210119053017-SomoaCovidCustomisedRawDataDownload-modifies-data.js
billli0: After a bit more looking around, have found this data builder has a name (`SurveyDataExportBuilder`) which is similar to `survey export` , but it seems like it doesn't deal with the questions type.

billli0:
Thanks @kael89 , a lot of great comments. I agree `REPLACE_ENTITY_ANSWER_WITH_ENTITY_NAME` is for a specific purpose here and it could not be used in other cases. Inspired by your suggestion, from databuilders I've found `transformValue`, which can be extended and used for `id -> name`. I'll try to modify the raw data result using that.

packages/database/src/modelClasses/Entity.js
billli0: Thanks for sharing the specific purposes of this class.

packages/web-config-server/src/apiV1/dataBuilders/transform.js
```diff
@@ -3,15 +3,20 @@
  * Copyright (c) 2019 Beyond Essential Systems Pty Ltd
  */
 
-const orgUnitCodeToName = async (models, value) => {
-  const organisationUnit = await models.entity.findOne({ code: value });
-  const { name } = organisationUnit || {};
-
-  return name || value;
+const entityFindNameByCondition = async (models, condition) => {
```
kael89, line 6: Function names generally start with a verb, since they are "actions" taken. Let's also use `conditions` instead of `condition`:

```suggestion
const findEntityName = async (models, conditions) => {
```

packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js
```diff
@@ -80,6 +80,15 @@ export class DataBuilder {
       eventId,
       ...additionalQueryConfig,
     });
+    if (
+      additionalQueryConfig.entityAnswerIdToName &&
+      additionalQueryConfig.entityAnswerIdToName.dataElementCodes
```
kael89, line 85: We now support [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in our codebase:

```suggestion
      additionalQueryConfig?.entityAnswerIdToName?.dataElementCodes
```

packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js
```diff
@@ -168,4 +181,21 @@ export class DataBuilder {
   };
 
   areEventResults = results => !!(results[0] && results[0].event);
+
+  eventsDataValuesEntityIdToName = async (models, events, dataElementCodes) => {
```
kael89, line 185:
```suggestion
  translateEventEntityIdsToNames = async (models, events, dataElementCodes) => {
```

Also, this function doesn't use `this`, and its logic is independent from this class, so we should make it an independent utility defined in another file and used in `DataBuilder`. We can place on its own file in `@tupaia/web-config-server/src/apiV1/dataBuilderes/helpers`

packages/database/src/migrations/20210119053017-SomoaCovidCustomisedRawDataDownload-modifies-data.js
```diff
@@ -0,0 +1,91 @@
...
+    SET
+      "dashboardReports" = array_remove("dashboardReports", '${reportId}')
+    WHERE
+      "code" IN (${groupCodes.map(code => `'${code}'`).join(',')});
+  `);
+}
+
+exports.up = async function (db) {
+  // Associate missing hierarchy id to some data sources
+
```
kael89, line 79: (detail) This comment is probably a leftover?

```suggestion
```

packages/database/src/migrations/20210119053017-SomoaCovidCustomisedRawDataDownload-modifies-data.js
```diff
@@ -0,0 +1,91 @@
+'use strict';
+
+import { insertObject, addReportToGroups } from '../utilities/migration';
```
kael89, line 3: About `addReportToGroups` : We shouldn't import DB specific utilities from `migration.js`, but rather copy them in the migration - see [this comment](https://github.com/beyondessential/tupaia/blob/dev/packages/database/src/utilities/migration.js#L141)

packages/database/src/migrations/20210119053017-SomoaCovidCustomisedRawDataDownload-modifies-data.js
```diff
@@ -0,0 +1,91 @@
...
+    exportDataBuilder: {
+      dataBuilder: 'rawDataValues',
+      dataBuilderConfig: {
+        skipHeader: false,
+        surveysConfig: {
+          [surveyCode]: {
+            entityAggregation: {
+              dataSourceEntityType: 'case',
+            },
+            entityAnswerIdToName: { dataElementCodes: ['QMIA029'] },
```
kael89, line 40: We can use a flatter option here:

```suggestion
            entityIdToNameElements: ['QMIA029'],
```

kael89 (changes requested):
Looking much better @billli0 , great work! A few changes and it will be ready to go!

packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js
billli0: Thanks Kostas, love this shorter expression!

packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js
billli0: Thanks Kostas, I agree, it looks better. Last week I did the same but was having babel importing error when moving it to `helpers`, couldn't solve it but today it works fine.

packages/database/src/migrations/20210119053017-SomoaCovidCustomisedRawDataDownload-modifies-data.js
billli0: Can I leave this config as an object? I was thinking in the future we might need more configs related to the entity name, like adding spaces at front and behind, color or font size. But I'm happy to make it simple and sharp.

kael89: I had similar concerns when working with data builders... I found out that most of the time, configs were better being simple rather than allowing space for something that never came in the future.

This is a case by case decision though, happy to leave it to you here

packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js
billli0: The code works fine but It actually fail the test, see [codeship](https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/72f51d32-1997-4782-ab92-08f23e441e8b?component=step_validate-and-test_test_test-batch-1_test-web-config-server) . It pass the test when this function is under `dataBuilders` folder, perhaps that's the reason `transform.js` is there.

kael89: True, although the error seems to be related to `analyticsPerPeriod`?

Also, there are lots of other helpers defined in this folder and used across data builders, so I can't see why this wouldn't work. Could you please investigate?

packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js
```diff
@@ -6,6 +6,7 @@ import { getSortByKey, getSortByExtractedValue, getUniqueEntries } from '@tupaia
 
 import { NO_DATA_AVAILABLE } from '/apiV1/dataBuilders/constants';
 import { transformValue } from 'apiV1/dataBuilders/transform';
```
kael89, line 8: @billli0  Just a clue - I can see that some internal imports in `web-config-server` do not have an `'/` prefix. Perhaps those are messing up with path resolution during builds:

```suggestion
import { transformValue } from '/apiV1/dataBuilders/transform';
```

We use [babel-module-resolved[https://www.npmjs.com/package/babel-plugin-module-resolver) to use aliased imports, just in web-config-server

billli0: Here comes the interesting thing, it works with 

>import { translateEventEntityIdsToNames } from '/apiV1/dataBuilders/helpers/translateEventEntityIdsToNames';

but not 

> import { translateEventEntityIdsToNames } from '/apiV1/dataBuilders/helpers'

kael89: This is weird? Perhaps because of other internal imports lying around without a `/` prefix? Or it could be unrelated. Who knows! 🤷

packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js
```diff
@@ -69,7 +70,7 @@ export class DataBuilder {
     const { organisationUnitCode, startDate, endDate, trackedEntityInstance, eventId } = this.query;
     const eventsProgramCode = overridenProgramCode || programCode;
 
-    return this.aggregator.fetchEvents(eventsProgramCode, {
+    const rawEvents = await this.aggregator.fetchEvents(eventsProgramCode, {
```
kael89, line 73: No need to distinguish between "raw" and non-raw events in this function now

```suggestion
    const events = await this.aggregator.fetchEvents(eventsProgramCode, {
```

packages/web-config-server/src/apiV1/dataBuilders/generic/dataDownload/rawDataValues.js
```diff
@@ -83,10 +83,11 @@ class RawDataValuesBuilder extends DataBuilder {
             code => !excludeCodes.includes(code),
           );
         }
-        const { entityAggregation } = surveyConfig;
+        const { entityAggregation, entityIdToNameElements = {} } = surveyConfig;
```
kael89, line 86:
```suggestion
        const { entityAggregation, entityIdToNameElements } = surveyConfig;
```

We want to keep this undefined here - if not, [this check](https://github.com/beyondessential/tupaia/blob/1796-somoa-covid-rawdata-dowload/packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js#L85) will be true even if the field is not provided

kael89 (changes requested):
Almost there!

kael89 (approved):

PR #1949, opened 2021-01-24
Clear out dead code in permissions branch

EMcQ-BES:
### Changes:

- Removing code that is no longer used
- Move a few functions into CRUDHandler format


packages/meditrak-server/src/routes/index.js
```diff
@@ -181,13 +173,13 @@ export default {
   importUsers: catchAsyncErrors(importUsers),
   importOptionSets: catchAsyncErrors(importOptionSets),
   postChanges: catchAsyncErrors(postChanges),
-  pruneChanges: catchAsyncErrors(pruneChanges),
+  pruneChanges: catchAsyncErrors(pruneChanges), // TODO: Legacy endpoint?
```
edmofro, line 176: Yep - safe to delete

packages/meditrak-server/src/routes/optionSets/EditOptionSets.js
```diff
@@ -15,6 +15,10 @@ export class EditOptionSets extends EditHandler {
   }
 
   async editRecord() {
-    await this.updateRecord();
+    const optionSet = await this.models.optionSet.findById(this.recordId);
+    const originalData = await optionSet.getData();
+    const updatedModel = { ...originalData, ...this.updatedFields };
+
+    return this.models.optionSet.updateById(this.recordId, updatedModel);
```
edmofro, line 22: What's the benefit of this specific handling?

packages/meditrak-server/src/dataAccessors/index.js
```diff
@@ -4,10 +4,6 @@
  */
```
edmofro, line 4: Surprised there are still so many left! Where is e.g. `findSurveyScreenComponentsInSurvey` used?

edmofro (commented):
Will need another round of regression testing after this...sorry @Klaus-S 😞

packages/meditrak-server/src/routes/optionSets/EditOptionSets.js
EMcQ-BES: I thought it was odd, but it was one of the few endpoints with a specific handler in the dataAccessors folders so I thought there must have been a reason.

If you also don't understand/remember why it was done this way then maybe I'll just get rid of it.

edmofro (approved):

packages/meditrak-server/src/routes/optionSets/EditOptionSets.js
edmofro: The original commit was in [this PR](https://github.com/beyondessential/meditrak-server/pull/249/files), where there was an obvious reason for the `editOption` to be handled specifically, but less obvious for `editOptionSet`. I'd be keen to drop it and then highlight it as an area for retesting.

PR #1951, opened 2021-01-25
715 psss editable table validation

tcaiger:
### Changes:

- Add styled tooltip for displaying validation messages on editable cells
- Add validation to ensure input are digits

---

### Screenshots:
See issue

### DIFF
--- a/packages/psss/src/components/EditableTable.js
+++ b/packages/psss/src/components/EditableTable.js
@@ -42,7 +42,7 @@ const ReadOnlyTextField = styled(EditableTextField)`
   }
 `;
 
-const RedTooltip = styled(Tooltip)`
+const ErrorTooltip = styled(Tooltip)`
   & .MuiTooltip-tooltip {
     background: ${props => props.theme.palette.error.main};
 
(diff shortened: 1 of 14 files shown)


packages/psss/src/components/EditableTable.js
```diff
@@ -42,6 +42,16 @@ const ReadOnlyTextField = styled(EditableTextField)`
   }
 `;
 
+const RedTooltip = styled(Tooltip)`
```
edmofro, line 45: Is `ErrorTooltip` a better name, in case the theme changes the error colour (highly unlikely, but possible)?

packages/psss/src/components/EditableTable.js
```diff
@@ -52,16 +62,22 @@ export const EditableCell = React.memo(({ id, columnKey, ...props }) => {
...
+        <EditableTextField
+          id={id}
+          name={id}
+          defaultValue={defaultValue}
+          inputProps={{ 'aria-label': columnKey }}
+          error={!!errors[id]}
+          inputRef={register({
+            required: 'Required',
+            pattern: {
+              value: /^\d+$/,
```
edmofro, line 75: Any decimal points in PSSS?

edmofro (approved):
Questions, but nothing holding up approval :-)

packages/psss/src/components/EditableTable.js
tcaiger: A good question but definitely not at this point, as we are counting numbers of people, so I think this  is a pretty solid assumption for now.

PR #1957, opened 2021-01-26
2090 stop cloudwatch on startup

edmofro:
### Changes:
Couldn't find a setting for cloudwatch agent to avoid turning itself on every reboot, so instead this adds a step to the startup script to turn it off for feature instances.

Note that cloudwatch agent is also on for the aggregation servers, but we rarely make extra instances of them, so I think it's okay for those rare extra instances to report their metrics.


kael89 (approved):
Great!

edmofro:
I've tested this through feature deployments and watching cloudwatch, so will merge directly into dev

PR #1960, opened 2021-01-27
Add to git ignore

Alastair-L:
### Issue #: No issue

I know we have gone with postman for requests, but I still have some requests which I would like to keep, and it might be easier if they were in the tupaia monorepo.

This PR is just a suggestion, I can always just put any files in an adjacent directory that isn't checked in.

Please let me know what you think.

Note: I have checked that this won't ignore anything existing: 
image


edmofro (approved):
Happy with this, thanks @Alastair-L :-)

P.S. we haven't actually "gone" with Postman...we just haven't done the card to set it up in vscode and see which is preferred - that's https://github.com/beyondessential/tupaia-backlog/issues/1495

Alastair-L:
I'll just merge this without testing (after the automated one), I think it should be ok.

EMcQ-BES:
Just curious is there a particular reason you want this is the `.gitignore` instead of your local `.git/info/exclude` ?

Alastair-L:
I didn't know that existed! Thanks Ethan. I'm not sure it's worth reverting now though?

edmofro:
Pretty happy with this committed to set up a convention - could you share it at dev chat @Alastair-L?

PR #1965, opened 2021-01-28
Update test data dump

kael89:


edmofro (approved):
Lots of changes that happened quite a while ago - I guess that represents that our split of data and schema is allowing tests to function on an old testDataDump, which is positive :-)

What was the inspiration?

kael89:
> Lots of changes that happened quite a while ago - I guess that represents that our split of data and schema is allowing tests to function on an old testDataDump, which is positive :-)
> 
> What was the inspiration?

The removal of the `alert`, `alert_comment` tables

PR #1968, opened 2021-01-28
1404 create new strive map overlay

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1404

This creates a new map overlay group and map overlay for the STRIVE project

The easiest way I could think of to create the type of map overlay requested by that card was to create two map overlays and link them together into one using `linkedMeasures` - there could perhaps be a way to achieve this with a single map overlay but I couldn't figure this out

I also didn't include a down migration because I was having issues getting it to work,  I think because of a foreign key constraint or because the down migration wasn't remembering the `id` that was created in the up migration using `generateId()` (but probably a foreign key thing). I figured the down migrations aren't really necessary and are mostly just useful for development but let me know if I should include one

I included a screenshot of what the map overlay looks like here ( https://github.com/beyondessential/tupaia-backlog/issues/1404#issuecomment-768791837 ) if it's useful for context

### DIFF
--- a/packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
+++ b/packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
@@ -130,8 +130,27 @@ exports.up = async function(db) {
   await insertObject(db, 'map_overlay_group_relation', NEW_OVERLAY_GROUP_RELATION_FOR_GROUP);
 };
 
-exports.down = function(db) {
-  return null;
+exports.down = async function (db) {
+  await db.runSql(`
+    DELETE FROM "mapOverlay" WHERE "id" = '${LINKED_MEASURE_OVERLAY_ID}';
+  `);
+
+  await db.runSql(`
+    DELETE FROM "mapOverlay" WHERE "id" = '${MAIN_OVERLAY_ID}';
+  `);
+
+  await db.runSql(`
+    DELETE FROM "map_overlay_group_relation" WHERE "child_id" = '${MAIN_OVERLAY_ID}';
+  `);
+
+  await db.runSql(`
+    DELETE FROM "map_overlay_group_relation" WHERE "child_id" in
+    (SELECT "id" from "map_overlay_group" WHERE "code" = 'STRIVE_Molecular_Data');
+  `);
+
+  await db.runSql(`
+    DELETE FROM "map_overlay_group" WHERE "code" = 'STRIVE_Molecular_Data';
+  `);
 };
 
 exports._meta = {
--- a/packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
+++ b/packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
@@ -23,6 +23,8 @@ const NEW_OVERLAY_GROUP_ID = generateId();
 const NEW_OVERLAY_GROUP_RELATION_FOR_OVERLAY_ID = generateId();
 const NEW_OVERLAY_GROUP_RELATION_FOR_GROUP_ID = generateId();
 
+// the below object contains common properties across the two overlays that will be created
+
 const BASE_OVERLAY = {
   userGroup: 'STRIVE User',
   dataElementCode: 'value',
--- a/packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
+++ b/packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
@@ -1,0 +1,139 @@
+'use strict';
+
+import { insertObject } from '../utilities';
+import { generateId } from '../utilities/generateId';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
```diff
@@ -0,0 +1,139 @@
...
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const MAIN_OVERLAY_ID = 'PG_STRIVE_K13_C580Y_Positive';
+const LINKED_MEASURE_OVERLAY_ID = 'PG_STRIVE_K13_C580Y_Positive_Count';
+const NEW_OVERLAY_GROUP_ID = generateId();
```
Alastair-L, line 22: My understanding is that it is not common to execute much code at the file level, instead doing everything except variable declaration in functions. This is (maybe, partly?) because the code will run when the module is first imported (see https://javascript.info/modules-intro#a-module-code-is-evaluated-only-the-first-time-when-imported) rather than when the `up` function is called. This means, for instance, that even if up were never called, we would still have to do the calculation in `generateId()`.

However:
a) `db-migrate` is clever and doesn't import the file unless it needs it.
b) There is a precedent: `database/src/migrations/20200812043136-AddTongaHpuSettingTypeMapOverlay-modifies-data.js`
c) Doing it this way is very clean, and I can't articulate a reason as to why it shouldn't be done.

Therefore go ahead, I just wanted to mention it I suppose.

packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
```diff
@@ -0,0 +1,139 @@
...
+        hideFromLegend: true,
+        color: 'green',
+      },
+      {
+        value: '0',
+        hideFromLegend: true,
+        color: 'red',
+      },
+      {
+        value: null,
```
Alastair-L, line 63: What happens when you click on a grey pin?

packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
```diff
@@ -0,0 +1,139 @@
...
+
+exports.up = async function(db) {
+  await insertObject(db, 'mapOverlay', LINKED_MEASURE_OVERLAY);
+  await insertObject(db, 'mapOverlay', MAIN_OVERLAY);
+  await insertObject(db, 'map_overlay_group', NEW_OVERLAY_GROUP);
+  await insertObject(db, 'map_overlay_group_relation', NEW_OVERLAY_GROUP_RELATION_FOR_OVERLAY);
+  await insertObject(db, 'map_overlay_group_relation', NEW_OVERLAY_GROUP_RELATION_FOR_GROUP);
+};
+
+exports.down = function(db) {
```
Alastair-L, line 133: I definitely reckon you could write an `exports.down` for this. I think it should be as simple as deleting all the records you just inserted right?
(https://beyondessential.slack.com/archives/CDGS6P56K/p1590559208005500)

Alastair-L (commented):
Looks good to me! Just that one question about the grey pins.

packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
avaek: It just displays "No Data"

packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
avaek: I thought so too, however I was having issues with the `map_overlay_group_relation` and `map_overlay_group` deletions yesterday. I assumed it was foreign key constraints but I looked into it more this morning and it seems the issue is because I create their `id` using the `generateId()` function

For example, I set the `id` using `const NEW_OVERLAY_GROUP_ID = generateId();`

When I do it this way neither `DELETE FROM "map_overlay_group" WHERE "id" = '${NEW_OVERLAY_GROUP.id}'` or `DELETE FROM "map_overlay_group" WHERE "id" = '${NEW_OVERLAY_GROUP_ID}'` will work (the code will run, however these entries will not be deleted). However, if I change the `id` to:

`const NEW_OVERLAY_GROUP_ID = 'test1234'`

then it will work.

To solve this, I just changed the problematic DELETEs to not refer to the `id` & instead refer to something else unique - I've just pushed this code to this PR, let me know if it looks okay

Alastair-L: Good point! I think it is somewhat related to my first comment (that running code at file level can sometimes be confusing).

packages/database/src/migrations/20210126013657-AddK13C580YPositiveMapOverlay-modifies-data.js
Alastair-L: Awesome, perfect

Alastair-L (approved):
LGTM! Nice work Andrew! 

(The build is failing because there is a capital in the branch name, you'll need to make a new branch and PR)

avaek:
Thanks for that @Alastair-L !

I'll close this PR and make a new PR for a new branch (without a capital in the name) here (if you could quickly check it out / approve it if it still looks good): https://github.com/beyondessential/tupaia/pull/1973

PR #1971, opened 2021-01-28
Permissions bugfixes 1647

EMcQ-BES:
### Changes:

- Restructure GETHandlers so SQL filtering applies to both the record fetching and the record count
- Fixup /me endpoints to work with CRUDHandlers
- Allow BES admin to edit projects
- Change default join type for access requests


edmofro (approved):
Nice solution! I actually think this makes the code more nicely separated even for the find - seems a more contained responsibility for the functions to just build the permissions filter than build permissions + return the results

🎉

PR #1974, opened 2021-01-28
1783 psss unit tests

tcaiger:
### Changes:
- Update jest tests
- Add script to generate test users
- Fix cypress tests to work with new servers
- Move all tests to one folder to reuse fixtures


rohan-bes (approved):
Nice work, looks good 😄

PR #1976, opened 2021-01-29
1619 Heatmap - Home village of quarantine passengers 

billli0:
### Issue #:
[Map Overlay: Heatmap - Home village of quarantine passengers #1619](https://github.com/beyondessential/tupaia-backlog/issues/1619)

### Changes:
- migration -> Attach case to village parent
- migration -> Create heatmap map overlay


packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
```diff
@@ -0,0 +1,97 @@
...
+  if (!name.includes('/'))
+    return db.runSql(`
+  select * from entity
+  where name = '${cleanName}'
+`);
+  return null;
+}
+
+async function changeEntityCaseParentToVillage(db, caseParentMapping) {
+  Promise.all(
```
kael89, line 48:
```suggestion
  await Promise.all(
```

packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
```diff
@@ -0,0 +1,97 @@
...
+      const [caseCode, parentId] = Object.entries(cas)[0];
+      if (parentId) await updateValues(db, 'entity', { parent_id: parentId }, { code: caseCode });
+    }),
+  );
+}
+
+exports.up = async function (db) {
+  const cases = (await getAnswersByQuestionCode(db, caseQuestionCode)).rows;
+  const addressesIds = (await getAnswersByQuestionCode(db, addressIdQuestionCode)).rows;
+  const addressesTexts = (await getAnswersByQuestionCode(db, addressTextQuestionCode)).rows;
```
kael89, line 59:
```suggestion
  const addressIds = (await getAnswersByQuestionCode(db, addressIdQuestionCode)).rows;
  const addressTexts = (await getAnswersByQuestionCode(db, addressTextQuestionCode)).rows;
```

packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
```diff
@@ -0,0 +1,97 @@
...
+
+  const caseParentMapping = await Promise.all(
+    cases.map(async cas => {
```
kael89, line 62: One extra letter is worth the readability:

```suggestion
    cases.map(async case => {
```

packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
```diff
@@ -0,0 +1,97 @@
...
+        from answer a
+        inner join survey_response sr
+        on a.survey_response_id = sr.id
+        inner join question q
+        on a.question_id = q.id
+        where q.code in ('${code}');    
+  `);
+}
+
+async function findEntityIdByName(db, name) {
```
kael89, line 37:
```suggestion
async function findEntityByName(db, name) {
```

packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
```diff
@@ -0,0 +1,97 @@
...
+  const cleanName = name.replace("'", '').trim();
+  if (!name.includes('/'))
```
kael89, line 39: This check is not required: if a name includes unknown characters, an entity will not be found anyway

kael89 (changes requested):
Nice one! Just a few renames requested.

Question: what will we do with responses where a parent village is not found?

billli0:
> Nice one! Just a few renames requested.
> 
> Question: what will we do with responses where a parent village is not found?

Hi Kostas. We decided to ignore them for now, and cleaning these data is one of the to do list. see [comment](https://github.com/beyondessential/tupaia-backlog/issues/1717#issuecomment-769448971).

packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
billli0: Yeah It is confusing, this is for one special case mapping village name `V'ictoria` to `Victoria`(Can't remember the name so pick victoria)

packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
billli0: It will have conflict with `case` from `switch`. Perhaps `individual`?

kael89: Oh it's a reserved word!! Haha interesting. I'd suggest something like `currentCase`

packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
kael89: My comment was not about the `cleanName` function (this makes sense), it was about 

```javascript
if (!name.includes('/'))
```

kael89: It will work as is though, so it's not important

kael89 (approved):
Looks good!

packages/database/src/migrations/20210125042225-SamoaCovidSurveyHierarchyUpdate-modifies-data.js
billli0: This is also an interesting one, if query with:
>   select * from entity
  where name = 'abc/cde'

It will confuse postgre sql and cause error.

PR #1978, opened 2021-01-29
1789 fix deep linking

tcaiger:
### Changes:

- Clear the active dashboard from the url when the OrgUnit changes 
- Move setting the default dashboard from history middleware to fetchDashboard so that is happens in the correct sequence
- Tweak dashboard dropdown menu to set the top of the dropdown list as a fallback (currently it sets the bottom)

---


packages/web-frontend/src/sagas.js
```diff
@@ -690,11 +694,28 @@ function* fetchDashboard(action) {
...
+      if (dashboard[currentDashboardGroupCode]) {
+        yield put(setDashboardGroup(project.dashboardGroupName));
+      }
+    }
+
+    // Check if the user has permission to view the dashboard
+    if (currentDashboardCode && !(currentDashboardCode in dashboard)) {
+      yield call(handleInvalidPermission, { projectCode });
+      return;
+    }
```
rohan-bes, line 717: Would this mean that if a user accidentally mistypes or uses a broken link to see a dashboard they will get prompted with the insufficient permissions dialog? Or will such a request throw an error?

rohan-bes (approved):
Looks good to me 😄 Just one question about the case where the DashboardCode in the URL is invalid, but approving ahead of time if that turns out to be a non-issue.

PR #1979, opened 2021-01-29
2176: Added operational facilities overlay for Fiji

chris-bes:
### Changes:

- Added operational facilities overlay for Fiji
- Moved Services Provided group to the top to make the new `Operational Facilities` the default overlay when hitting Fiji

### DIFF
--- a/packages/database/src/migrations/20210129022440-AddFacilityTypeOverlayForFiji-modifies-data.js
+++ b/packages/database/src/migrations/20210129022440-AddFacilityTypeOverlayForFiji-modifies-data.js
@@ -23,7 +23,7 @@ const LINKED_OVERLAY_ID = 'Fiji_Operational_Facility_Type';
 
 const MAIN_OVERLAY = {
   id: MAIN_OVERLAY_ID,
-  name: 'Fiji operational facilities',
+  name: 'Operational facilities',
   userGroup: '',
   dataElementCode: 'BCD1',
   linkedMeasures: `{${LINKED_OVERLAY_ID}}`,
@@ -67,7 +67,7 @@ const MAIN_OVERLAY = {
 
 const LINKED_OVERLAY = {
   id: LINKED_OVERLAY_ID,
-  name: 'Fiji facilities',
+  name: 'Facility types',
   dataElementCode: 'facilityTypeCode',
   userGroup: 'Public',
   measureBuilderConfig: {
@@ -125,7 +125,7 @@ exports.up = async function (db) {
     map_overlay_group_id: await codeToId(db, 'map_overlay_group', 'Services_provided'),
     child_id: LINKED_OVERLAY_ID,
     child_type: 'mapOverlay',
-    sort_order: 10,
+    sort_order: 0, // doesn't matter because this overlay is hidden
   };
 
   await insertObject(db, 'mapOverlay', MAIN_OVERLAY);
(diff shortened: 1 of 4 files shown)


packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -93,14 +93,21 @@ const checkIfGroupedMapOverlaysAreEmpty = nestedMapOverlayGroups => {
...
-  //If all the relations have sort_order, sort by the sort_order. Otherwise by default sort by name.
-  return allRelationsHaveSortOrder(relations) ? sortBySortOrder : getSortByKey('name');
+  const sortedOverlaysByOrder = mapOverlayItems
+    .filter(m => childIdToSortOrder[m.id] !== null)
+    .sort((m1, m2) => childIdToSortOrder[m1.id] - childIdToSortOrder[m2.id]);
+  const sortedOverlaysAlphabetically = mapOverlayItems
+    .filter(m => childIdToSortOrder[m.id] === null)
+    .sort(getSortByKey('name'));
+
+  return [...sortedOverlaysByOrder, ...sortedOverlaysAlphabetically];
```
chris-bes, line 110: Before if any of the items in a list of map overlays or map overlay groups doesn't have a sort_order, the whole list will be sorted alphabetically. I've modified this to sort the ones that have sort_order first to allow more flexibility

IgorNadj (approved):
LGTM 🏌️

packages/database/src/migrations/20210129022440-AddFacilityTypeOverlayForFiji-modifies-data.js
```diff
@@ -0,0 +1,165 @@
...
+        value: 3,
+      },
+    ],
+    displayType: 'color',
+    hideFromMenu: true,
+    measureLevel: 'Facility',
+    displayedValueKey: 'facilityTypeName',
+  },
+  countryCodes: '{FJ}',
+  projectCodes: '{explore}',
```
edmofro, line 111: Should this show up for `wish` as well? @kujo4pm?

packages/database/src/migrations/20210129022440-AddFacilityTypeOverlayForFiji-modifies-data.js
```diff
@@ -0,0 +1,165 @@
...
+  id: LINKED_OVERLAY_ID,
+  name: 'Facility types',
+  dataElementCode: 'facilityTypeCode',
+  userGroup: 'Public',
+  measureBuilderConfig: {
+    facilityTypeCodeMetadataKey: 'type',
+  },
+  measureBuilder: 'valueForOrgGroup',
+  presentationOptions: {
+    values: [
```
edmofro, line 78: Was it necessary to specify these values? I would have expected using `displayedValueKey` would be enough, and make it less "stuck" to the config and more able to respond to whatever the facility types might get changed to in future?

chris-bes: As I understand (I might be wrong), even with `displayedValueKey` defined, we still have to specify the values for the overlay. Looking at (`updateLegendFromDisplayedValueKey` in `measureData.js`), it seems to me is what it does is it will try to override the labels/names specified in `values` with the defined `displayedValueKey`, given that the `values` still has to be specified with the correct `value`...?

edmofro: Aah cool, thanks for clarifying!

PR #1983, opened 2021-01-31
1400: Add Strive positive malaria cases bar graph + some core work

Alastair-L:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/1400

To add the indicators required for the bar graph, one change was required in `data-broker > DhisService`.
### Situation:
Fetching event analytics with the new api.
2 programs: `SQMAL` and `SPF`. 
2 dataElementCodes, one from each program (`STR_QMAL05` and `STR_PF05`, respectively)
### Existing behavior:
The existing system failed when trying to fetch dataElementCodes that weren't part of the program being fetched, and tried to fetch both dataElements for both api calls.
### Proposed solution (this PR):
Only ever fetch dataElementCodes that are associated with the program being fetched, as according to the `data_source_data_group` table. 

Note: currently not all dataElement-program combinations are represented in that table, but that should be fixed here: https://github.com/beyondessential/tupaia/pull/1995

### I would love to hear feedback on whether this is a viable solution, or if there is another option we should pursue


packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
```diff
@@ -275,23 +305,11 @@ export const testPullAnalytics = () => {
         });
 
         it('no program', async () => {
```
Alastair-L, line 307: I just realized that this change might suppress some errors. Just a note to self to revert the change

packages/indicators/src/builders/helpers.ts
```diff
@@ -40,7 +40,11 @@ export const fetchAnalytics = async (
   await Promise.all(
     Object.entries(aggregationJsonToCodes).map(async ([aggregationJson, codes]) => {
       const aggregations = JSON.parse(aggregationJson);
-      const { results } = await aggregator.fetchAnalytics(codes, fetchOptions, { aggregations });
+      const { results } = await aggregator.fetchAnalytics(
+        codes,
+        { ...fetchOptions, useDeprecatedApi: false },
```
Alastair-L, line 45: Will this have any unintended side affects?

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -226,27 +197,80 @@ export class DhisService extends Service {
...
+    },
+  })
+
+  pullAnalyticsFromEventsForApi = async (api, dataSources, options) => {
+    const { programCodes = [], period, startDate, endDate, organisationUnitCodes } = options;
+
+    const baseQuery = {
       programCodes,
-      dataElementCodes: dhisElementCodes,
+      organisationUnitCodes,
```
Alastair-L, line 244: Will this have any unintended side affects?

packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
```diff
@@ -225,6 +225,27 @@ export const testPullAnalytics = () => {
...
+            programCode: 'POP01',
+          }),
+        );
+        expect(dhisApi.getEventAnalytics).toHaveBeenCalledWith(
+          expect.objectContaining({
+            programCode: 'DIFF_GROUP',
+          }),
+        );
+
+        // Note: The api should NOT be called with programCode: CODE_WITH_NO_ELEMENTS
```
Alastair-L, line 244: Note to self: also could potentially hide errors?

packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
Alastair-L: Actually does this make sense to consider as a case?

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -226,27 +197,81 @@ export class DhisService extends Service {
     return buildAnalyticsFromEvents(translatedEvents, dataElements);
   };
 
-  pullAnalyticsFromEventsForApi = async (api, dataSources, options) => {
-    const { programCodes = [], period, startDate, endDate } = options;
+  translateDhisEventAnalytics = async (dhisAnalytics, dataSources) => {
```
Alastair-L, line 200: Do all these little, specific functions it easier to understand or less so?

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -226,27 +197,81 @@ export class DhisService extends Service {
...
-      eventAnalytics,
-      dataSources,
-    );
+    const allAnalytics = [];
 
-    return buildAnalyticsFromDhisEventAnalytics(translatedEventAnalytics, dataElementCodes);
+    const fetchAnalyticsForProgram = async programCode => {
+      const dataSourcesToFetch = await this.getDataSourcesInProgram(dataSources, programCode);
+
+      if (dataSourcesToFetch.length === 0) return;
```
Alastair-L, line 257: Is stopping the fetching process if there are no `dataSources` to fetch going to have any unintended side affects?

packages/database/src/migrations/20210128055252-AddStrivePositiveResultBarGraph-modifies-data.js
```diff
@@ -0,0 +1,114 @@
...
+  id: 'PG_Strive_PNG_Positive_Result_By_Type_Bar',
+  dataBuilder: 'analyticsPerPeriod',
+  dataBuilderConfig: {
+    series: [
+      { key: 'Pf +ve', dataElementCode: 'STRIVE_PERCENTAGE_PF_OF_TOTAL_WEEKLY' },
+      { key: 'Pv +ve', dataElementCode: 'STRIVE_PERCENTAGE_PV_OF_TOTAL_WEEKLY' },
+      { key: 'Pm +ve', dataElementCode: 'STRIVE_PERCENTAGE_PM_OF_TOTAL_WEEKLY' },
+      { key: 'Po +ve', dataElementCode: 'STRIVE_PERCENTAGE_PO_OF_TOTAL_WEEKLY' },
+    ],
+    entityAggregation: {
```
Alastair-L, line 43: Note: because of the `SUM_EACH_WEEK` aggregationType for each dataElement in the indicators, there should only be one analytic per indicator per week returned to the dataBuilder.

This means that the entityAggregation here is mostly for show, the org unit should NOT be important in building the data. 

However, I figure it makes the most sense to still have it, as it matches conceptually what we are doing here.

packages/indicators/src/builders/helpers.ts
kael89: I believe no! Good decision to use this option

packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
kael89: I believe so! Example use case: get a list of programs, filter based on a condition that depends on dynamic data, then fetch using the remaining programs. The result could sometimes be an empty program list.

My conclusions so far:
* It is generally beneficial to support empty lists/maps as input
  ✔ supports a more flexible/dynamic usage of the API
  ✔enhances robustness
  ✔reduces complexity by introducing less assumptions
* On the other hand, we shouldn't overuse default parameters (`programCodes = []`) unless it is for really optional params eg optional config. In other words, support empty inputs, but the client of the API is expected to explicitly provide them.
  ✔ simpler function signatures
  ✔ no need to worry about `null` values
  ✔ we essentially assert the type of the param

packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
```diff
@@ -356,6 +375,51 @@ export const testPullAnalytics = () => {
...
+            dataElementCodeToName: {
+              POP01: 'Population 1',
+              DIF01: 'Different 1',
+            },
+          },
+        };
+
+        buildAnalyticsMock
+          .mockReturnValueOnce(analyticsResponsePOP01)
+          .mockReturnValueOnce(analyticsResponseDIF01);
```
kael89, line 408: Nice api usage!

packages/data-broker/src/services/dhis/DhisService.js
kael89: I love them! Size does matter in function length: all other qualitative aspects being equal, the smaller the better!

A few resources on the subject:
* [Martin Fowler - FunctionLength](https://martinfowler.com/bliki/FunctionLength.html)
* [Rule of 30](https://dzone.com/articles/rule-30-%E2%80%93-when-method-class-or)

kael89 (approved):
This is an excellent piece of work @Alastair-L !! Very mature work and approach on refactoring. I loved this PR 🎉 

Approving this since all my suggestions are about small details, but let's make sure that a few questions that arose are answered before we proceed

packages/database/src/migrations/20210128055252-AddStrivePositiveResultBarGraph-modifies-data.js
```diff
@@ -0,0 +1,114 @@
...
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+
+const CODE_TO_NAME = { STR_PF05: 'PF', STR_PV05: 'PV', STR_PM05: 'PM', STR_PO05: 'PO' };
+
+const insertIndicator = (db, code) => {
+  const indicator = {
+    id: generateId(),
+    code: `STRIVE_PERCENTAGE_${CODE_TO_NAME[code]}_OF_TOTAL_WEEKLY`,
```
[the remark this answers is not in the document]
edmofro: I prefer 2 also:
- It makes the actual acronyms more obvious
- Just a bit friendlier on the eyes

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -5,6 +5,7 @@
 
 import keyBy from 'lodash.keyby';
```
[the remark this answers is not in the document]
nick-beyondessential: @kael89 started looking at this with @IgorNadj, I suspect the `Event analytics require at least one organisation unit id` error doesn't get triggered because aggregator catches it first?
```
  async fetchAnalytics(codeInput, fetchOptions, aggregationOptions = {}) {
    const { organisationUnitCode, organisationUnitCodes } = fetchOptions;
    if (!organisationUnitCode && (!organisationUnitCodes || !organisationUnitCodes.length)) {
      // No organisation unit code, return empty response
      return {
        results: [],
        metadata: {
          dataElementCodeToName: {},
        },
        period: periodFromAnalytics([], fetchOptions),
      };
    }
```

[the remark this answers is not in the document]
kael89: Yes, that makes sense! So an error will not get thrown, but the returned data will be always empty.

The next step would be to investigate existing functionality using `useDeprecatedApi: false` and confirm that it's broken - if it's not broken, then explain how it could work without org unit codes

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -226,27 +197,81 @@ export class DhisService extends Service {
...
+  mergeAnalytics = analytics => ({
+    results: analytics
+      .map(({ results }) => results)
+      .flat()
+      .sort(getSortByKey('period')),
+    metadata: {
+      dataElementCodeToName: analytics.reduce(
+        (dataElementCodeToName, { metadata }) => ({
+          ...dataElementCodeToName,
+          ...(metadata.dataElementCodeToName || {}),
```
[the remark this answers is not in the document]
Alastair-L: Wow! Interesting. In fact it looks like all non-iterables don't error! Thanks
image
image

packages/database/src/migrations/20210128055252-AddStrivePositiveResultBarGraph-modifies-data.js
```diff
@@ -0,0 +1,114 @@
...
+    builder: 'arithmetic',
+    config: {
+      formula: `${code} / STR_QMAL05`,
+      aggregation: {
+        [code]: 'SUM_EACH_WEEK',
+        STR_QMAL05: 'SUM_EACH_WEEK',
+      },
```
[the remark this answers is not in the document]
Alastair-L: Clean api 🎉

packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
Alastair-L: Fantastic! That makes a lot of sense.

packages/data-broker/src/services/dhis/DhisService.js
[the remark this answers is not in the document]
kael89: I don't know this applied to other non-iterables, nice find!

packages/data-broker/src/services/dhis/DhisService.js
Alastair-L: Awesome! I finally found time to read those sources and they seem pretty clear! I'm glad to hear that (although perhaps they should be even shorter (`mergeAnalyticResults`, `mergeAnalyticsMetadata`)

kael89 (approved):

packages/data-broker/src/services/dhis/DhisService.js
Alastair-L: See [Nick's notes](https://github.com/beyondessential/tupaia/pull/1983#discussion_r575935785) on why it currently works and [Igor's addendum](https://github.com/beyondessential/tupaia/pull/2024#discussion_r576462039) on what the plan is, resulting in [this card](https://github.com/beyondessential/tupaia-backlog/issues/2251) to investigate further. Currently it looks like that card is a prerequisite for this one. Is that correct @IgorNadj?

packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
Alastair-L: Taking a what would Kostas do approach, let's do what makes the code most closely match our intuitive separated concerns. Namely, that regardless of dataSource, you should only be able to pull for programCodes which are in the database. Making the logic part of the dhisApi doesn't make much sense as I'd expect that we want this functionality for all dataSources. So option 2 🎉

Alastair-L: Ended up only doing it in DhisService, as programCodes are only used in DHIS (which I didn't know)

Alastair-L:
I believe that this PR is now just blocked on this: https://github.com/beyondessential/tupaia/pull/1983#discussion_r578143942

packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
kael89: I guess after https://github.com/beyondessential/tupaia-backlog/issues/663 was implemented all elements are grouped under a data group, so we could use data groups/elements (and not surveys/questions)  for the `TupaiaService` too.

But we currently don't, so good call!

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -216,6 +216,17 @@ export class DhisService extends Service {
...
 
+  validateProgramCodesExist = async programCodes => {
+    const validateProgramCode = async programCode => {
+      const programDataSource = await this.models.dataSource.findOne({
+        code: programCode,
+        type: this.dataSourceTypes.DATA_GROUP,
+      });
+      if (!programDataSource) throw new Error(`Program not found: ${programCode}`);
+    };
+    return Promise.all(programCodes.map(programCode => validateProgramCode(programCode)));
```
kael89, line 227: It would be more efficient to fetch data sources using all `programCodes` with a single `find` query, then check whether any one is missing in the results

packages/data-broker/src/__tests__/services/dhis/DhisService/testPull/testPullAnalytics.js
```diff
@@ -207,6 +207,15 @@ export const testPullAnalytics = () => {
         );
       });
 
+      it('non-existent program', async () => {
+        await expect(
+          dhisService.pull([DATA_SOURCES.POP01, DATA_SOURCES.POP02], 'dataElement', {
+            programCodes: ['POP01', 'DIFF_GROUP', 'CODE_NOT_IN_SYSTEM'],
+            useDeprecatedApi: false,
+          }),
+        ).rejects.toThrowError(/.*CODE_NOT_IN_SYSTEM.*/g);
```
kael89, line 216: (No need to change this) you can also use our custom matcher:

```suggestion
        ).toBeRejectedWith(/.*CODE_NOT_IN_SYSTEM.*/g);
```

Those are defined in [/jest.setup.js](https://github.com/beyondessential/tupaia/blob/dev/jest.setup.js)

kael89 (approved):
Just a small change suggested - making this optional as the performance gain for 1-2 programs would be negligible

Alastair-L:
^ Small merge conflict with https://github.com/beyondessential/tupaia/pull/2124

kael89 (approved):

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -135,11 +135,7 @@ export class DhisService extends Service {
     const { programCodes } = options;
 
     if (programCodes) {
-      // TODO remove `useDeprecatedApi` option as soon as `pullAnalyticsFromEventsForApi_Deprecated()` is deleted
-      const { useDeprecatedApi = true } = options;
-      return useDeprecatedApi
-        ? this.pullAnalyticsFromEventsForApi_Deprecated
-        : this.pullAnalyticsFromEventsForApi;
+      return this.pullAnalyticsFromEventsForApi;
```
IgorNadj, line 138: This commit would be a great separate PR. It would also involve regression testing a whole bunch of visualisations right?

PShoe:
@Alastair-L or @IgorNadj checking on this PR. There's quite a bit to read in here and chasing where we landed. Still blocked on this [comment](https://github.com/beyondessential/tupaia/pull/1983#discussion_r578143942) ?

kael89:
@PShoe without being 100% sure, my last impression about this card was that it was ready to go. Then there was this change: https://github.com/beyondessential/tupaia/pull/1983#discussion_r591910337, which as @IgorNadj commented may involve a significant amount of regression testing + it may not be required in the context of this PR.

**Note:** we need to take into account whether the changes here will work after https://github.com/beyondessential/tupaia-backlog/issues/2177 is merged. 2177 could actually make all the dhis-related changes in this PR redundant, although we can still keep them for future usage.

packages/data-broker/src/services/dhis/DhisService.js
PShoe: @IgorNadj  how does your investigation work affect the above?

IgorNadj:
Discussed today, waiting on Alastair to have time to look into this with us

Alastair-L:
# Notes:
## In https://github.com/beyondessential/tupaia/pull/1983#discussion_r579956717
>I guess after beyondessential/tupaia-backlog#663 was implemented all elements are grouped under a data group, so we could use data groups/elements (and not surveys/questions) for the TupaiaService too.

The discussed card is now implemented so we can clean up if desired.
## Response to https://github.com/beyondessential/tupaia/pull/1983#discussion_r591910337 
>This commit would be a great separate PR. It would also involve regression testing a whole bunch of visualisations right?

According to your notes here: https://github.com/beyondessential/tupaia-backlog/issues/2251#issuecomment-783090514, you're totally right this would break a preaggregator (plus a bunch of tests) (which I didn't test at the time 😬). Definitely worth reverting that change, not sure what was going through my head at that stage...

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -158,100 +153,103 @@ export class DhisService extends Service {
     return events;
   };
 
-  fetchEventAnalyticsForPrograms = async (api, programCodes, query) => {
-    const allHeaders = [];
```
IgorNadj, line 162: We used to merge first then translate, so we had to update the width and height to be correct (the translator relies on it - probs). After these changes, we translate first then merge, so we don't need this logic anymore

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -158,100 +153,103 @@ export class DhisService extends Service {
...
-    let metaData = { items: {}, dimensions: {} };
-    let width = 0;
-    let height = 0;
-    const rows = [];
+  translateDhisEventAnalytics = async (dhisAnalytics, dataSources) => {
+    const dataElementCodes = dataSources.map(({ code }) => code);
 
-    const fetchAnalyticsForProgram = async programCode => {
```
IgorNadj, line 168: This method had been inlined, because the merging has been extracted into another method

packages/database/src/migrations/20210128055252-AddStrivePositiveResultBarGraph-modifies-data.js
```diff
@@ -0,0 +1,111 @@
...
+      { key: 'Pv +ve', dataElementCode: 'STRIVE_Percentage_PV_Of_Total_Weekly' },
+      { key: 'Pm +ve', dataElementCode: 'STRIVE_Percentage_PM_Of_Total_Weekly' },
+      { key: 'Po +ve', dataElementCode: 'STRIVE_Percentage_PO_Of_Total_Weekly' },
+    ],
+    entityAggregation: {
+      dataSourceEntityType: 'village',
+      aggregationEntityType: 'facility',
+      aggregationType: 'REPLACE_ORG_UNIT_WITH_ORG_GROUP',
+    },
+    programCodes: ['SQMAL', 'SPF', 'SPV', 'SPM', 'SPO'],
```
IgorNadj, line 45: Data Broker does something funny (in pullAnalyticsFromEventsForApi). It lets you call getAnalytics with multiple program codes. 

BEFORE: It used to be that it would just loop the program codes and for each program request all the DE codes.  

AFTER: After these changes, it loops the program codes and for each program it only requests the DE codes for that program.

-----

Even though pullAnalyticsFromEventsForApi lets you do this, no vis/data builder does this, or it would fall over. TODO: check if this anyone actually does this (look at mapOverlay FRIDGE_SOH_VALUE). ------ And then check with Kostas to see if we can just throw an exception if any data builder tries to request data from dhis getAnalytics with >1 programCode, >1 DE.  

`programCodes` -> `programCode`.
`if (programCodes typeof Array) throw 'no'`

packages/data-broker/src/services/dhis/DhisService.js
rohan-bes: This issue has been resolved a while ago, see: https://github.com/beyondessential/tupaia-backlog/issues/2251

rohan-bes:
Closing out in favour of: https://github.com/beyondessential/tupaia/pull/2624

PR #1986, opened 2021-02-01
1526 import org units

IgorNadj:
### Changes:

- Fix lat/lon of a city
- Copy districts and sub_districts from laos_schools

---

### DIFF
--- a/packages/database/src/migrations/20210201051217-SyncLaosEocEntitiesWithLaosSchools-modifies-data.js
+++ b/packages/database/src/migrations/20210201051217-SyncLaosEocEntitiesWithLaosSchools-modifies-data.js
@@ -1,0 +1,189 @@
+'use strict';
+
+import { generateId } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210201051217-SyncLaosEocEntitiesWithLaosSchools-modifies-data.js
```diff
@@ -0,0 +1,189 @@
...
+
+    const subDistrictId = await getEntityId(db, province.capital.subDistrictCode);
+
+    const cityId = await getEntityId(db, province.capital.code);
+
+    await db.runSql(`UPDATE entity SET parent_id = '${subDistrictId}' WHERE id = '${cityId}';`);
+
+    await db.runSql(`
+      INSERT INTO entity_relation (id, parent_id, child_id, entity_hierarchy_id)
+      VALUES (generate_object_id(), '${subDistrictId}', '${cityId}', '${laosEocEntityHierarchyId}');
```
kael89, line 178: I guess we also need to delete existing entity relations?

kael89 (commented):
Great work, your migration was a pleasure to read 😃 Just one question before I approve

packages/database/src/migrations/20210201051217-SyncLaosEocEntitiesWithLaosSchools-modifies-data.js
IgorNadj: Ah very good spotting! I was messing with data manually and missed that step. Updated

kael89 (approved):
🙌

PR #1992, opened 2021-02-02
Add FETP individual profile reports (migration)

nick-beyondessential:
### Changes:

- migration to add 4 new dashboard reports:  
FETP_PG_graduate_details, FETP_PG_graduate_area_of_expertise, FETP_PG_graduate_number_of_outbreaks, FETP_PG_graduate_number_of_outbreaks_lead
- and  dashboardGroup: PG_FETP_Individual_FETP_Graduates
- add noDataValue config to tableOfCalculatedValues
- update NoDataMessage component to check for existrence of 'noDataMessage' in viewContent not falsey check


packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfDataValues/tableOfCalculatedValues.js
```diff
@@ -32,11 +33,13 @@ class TableOfCalculatedValuesBuilder extends TableOfDataValuesBuilder {
 
   async buildValuesByCell() {
     const hierarchyId = await this.fetchEntityHierarchyId();
+    const noDataValue = this.config.noDataValue ?? NO_DATA_AVAILABLE;
```
kael89, line 36: Great usage of `??`!

packages/database/src/migrations/20210202061216-AddFetpGraduateIndividualProfiles-modifies-data.js
```diff
@@ -0,0 +1,209 @@
...
+
+  const dataBuilderConfig = {
+    rows: Object.values(report.elements),
+    cells: Object.keys(report.elements).map(mapElementsToCells),
+    columns: ['main'],
+    entityAggregation: {
+      aggregationType: 'RAW',
+      dataSourceEntityType: ['individual'],
+      aggregationEntityType: 'individual',
+    },
```
kael89, line 35: Not sure if this is required, since it maps the aggregation source type to itself? I'm not fully across `entityAggregation`, so the question is mainly for my benefit 😄

packages/database/src/migrations/20210202061216-AddFetpGraduateIndividualProfiles-modifies-data.js
```diff
@@ -0,0 +1,209 @@
...
+
+const buildSingle = report => {
+  const dataBuilderConfig = {
+    dataElementCodes: report.elements,
+  };
+  const viewJson = {
+    name: report.name,
+    type: 'view',
+    viewType: 'singleValue',
+    noDataMessage: ' ',
```
kael89, line 63: Does this need any `web-frontend` updates, or will it work out of the box? Just asking because you mentioned

> Add empty noDataMessage for SingleValue and MultiSingleValue reports

in the PR's description, so I assumed this would be related to `web-frontend` changes.

kael89 (commented):
Great work @nick-beyondessential !! A few questions and I'm happy to approve

packages/database/src/migrations/20210202061216-AddFetpGraduateIndividualProfiles-modifies-data.js
nick-beyondessential: Yes using existing functionality, I'll remove from description as it is confusing, I copied from commit messages.

nick-beyondessential: One 'hack' I had to  do was to set empty string with space (' ') because  '' was being falsey, but feel this was an ok solution.

kael89: Hm ideally the front-end code should be updated to check for `null`/`undefined` and not for all falsy values? Do you thing that would be a tricky change?

nick-beyondessential: No should be fine as long as it doesn't break existing frontend component config assumptions. (because it is changed in code rather than just a config setting) But I can check this.
```
export const NoDataMessage = ({ viewContent }) => {
  let message = 'No data';

  if (viewContent.noDataMessage) {
    message = viewContent.noDataMessage;
  } else if (viewContent.source === 'mSupply') {
    message = 'Requires mSupply';
  } else ...
  }
```

kael89: Yeah, `''` and `0` should both be valid no data messages

nick-beyondessential: Changed to `if ('noDataMessage' in viewContent)`

packages/database/src/migrations/20210202061216-AddFetpGraduateIndividualProfiles-modifies-data.js
nick-beyondessential: You are correct need `dataSourceEntityType` but not `aggregationEntityType` in this case

kael89: Good to know! Not a big deal but I guess the simpler the better

packages/web-frontend/src/components/View/NoDataMessage.js
```diff
@@ -22,7 +22,7 @@ const NoDataMessageText = styled.p`
 export const NoDataMessage = ({ viewContent }) => {
   let message = 'No data';
 
-  if (viewContent.noDataMessage) {
+  if ('noDataMessage' in viewContent) {
```
kael89, line 25: This will not work if `noDataMessage` is explicitly set to `null`/`undefined`: 
```js
const viewContent= { noDataMessage: undefined };
console.log('noDataMessage' in viewContent); // true
```

But it's probably fine for our purposes here.

kael89 (approved):
🙌

PR #1997, opened 2021-02-03
Reduce cache memory footprint

edmofro:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/2196

Rather than storing the overhead of the models etc. in the cache, we should just store the data, and construct the model instances again when we need them.

This was revealed when testing redis, because the models are a circular structure (each model references the other models, which then reference the original and around it goes), which can't be serialised to JSON and stored in redis. Seems it's not just a benefit in making the cache serialisable, but also means we save about 65% less memory in the cache for the Laos entity hierarchy alone.

See https://docs.google.com/spreadsheets/d/1KoewibyIxEJjpcjataeKi5svspPi41gbcG0LUtwIHZk/edit#gid=1369222111 for memory usage data (second tab)


kael89 (approved):
Great optimisation and benchmark data!

PR #1999, opened 2021-02-03
Use a bigger batch size for fetching ancestors

edmofro:
### Changes:
- Increase the max batch size after testing that 14000 Laos schools have no problems

---

### DIFF
--- a/packages/database/src/utilities/runDatabaseFunctionInBatches.js
+++ b/packages/database/src/utilities/runDatabaseFunctionInBatches.js
@@ -3,8 +3,6 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-import flatten from 'lodash.flatten';
-
 // some part of knex or node-pg struggles with too many bindings, so we batch them in some places
 // errors occurred at around 5000 bindings in initial testing, but more is needed - see issue #2197
 export const MAX_BINDINGS_PER_QUERY = 2500;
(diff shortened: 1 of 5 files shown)


packages/database/src/utilities/runDatabaseFunctionInBatches.js
```diff
@@ -3,18 +3,21 @@
...
-    const batch = arrayToBeBatched.slice(i, i + batchSize);
-    const resultForBatch = await databaseFunction(batch);
-    if (Array.isArray(resultForBatch)) {
-      result.push(...resultForBatch);
-    }
+    batches.push(arrayToBeBatched.slice(i, i + batchSize));
   }
-  return result;
+  const batchedResults = await Promise.all(batches.map(async b => databaseFunction(b)));
+  return flatten(batchedResults.filter(r => Array.isArray(r)));
```
kael89, line 22: This will work for 1 nesting level, which I guess is the case here... If not, we'd have to specify a max nesting level

```suggestion
  return batchedResults.filter(r => Array.isArray(r)).flat();
```

kael89 (changes requested):
Very interesting finding! Just a small suggestion

packages/database/src/utilities/runDatabaseFunctionInBatches.js
edmofro: Oh nice! Forgot about node's improvements 🙌 

Yep just one level. Will change that over and remove the dependency

kael89 (approved):

PR #2009, opened 2021-02-05
1398 facility level epicurve

Alastair-L:
### Changes: 

Added `useDeprecatedApi: false` when fetching all analytics. 
Possibly scope for a lot of regression, so perhaps we could just add it to `analyticsPerPeriod`.

### DIFF
--- a/packages/web-config-server/src/aggregator/Aggregator.js
+++ b/packages/web-config-server/src/aggregator/Aggregator.js
@@ -30,7 +30,11 @@ export class Aggregator extends BaseAggregator {
       hierarchyId,
     );
 
-    return super.fetchAnalytics(dataElementCodes, { ...fetchOptions, useDeprecatedApi: false }, aggregationOptions);
+    return super.fetchAnalytics(
+      dataElementCodes,
+      { ...fetchOptions, useDeprecatedApi: false },
+      aggregationOptions,
+    );
   }
 
   async fetchEvents(programCode, originalQuery, replacementValues) {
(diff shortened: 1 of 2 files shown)


IgorNadj (approved):
LGTM

Alastair-L:
Notes: I force pushed, but it should be the same code

Alastair-L:
^Force pushing to revert merging in dev.

packages/web-config-server/src/aggregator/Aggregator.js
```diff
@@ -30,7 +30,11 @@ export class Aggregator extends BaseAggregator {
       hierarchyId,
     );
 
-    return super.fetchAnalytics(dataElementCodes, fetchOptions, aggregationOptions);
+    return super.fetchAnalytics(
+      dataElementCodes,
+      { ...fetchOptions, useDeprecatedApi: false },
```
IgorNadj, line 35: This is a global change, but there may be nothing else affected, TODO: check

IgorNadj: Nope, just a reformat

chris-bes:
@IgorNadj Do we still need this PR?

packages/web-config-server/src/aggregator/Aggregator.js
IgorNadj: Hmm, looks like it's not just a reformat... `useDeprecatedApi: false` added?

IgorNadj:
This was reimplemented in #2470

PR #2030, opened 2021-02-09
Clean up structure of meditrak-server tests

edmofro:
No issue - just some cleanup to align the structure of our tests

This will be pretty difficult to review line by line, so I'd suggest checking it out and looking at how the tests directory fits together now. I haven't deleted any tests, just renamed, moved, and extracted a bit of boilerplate, so if tests pass and the structure looks ok, it shouldn't need closer review than that.

Changes:
- Moved all test utilities (e.g. getModels, TestableApp) into `/tests/testUtilities` rather than `/tests` root
- Moved and renamed all test suites in the root `/tests` or `/tests/routeTests` folders to instead sit within the directory structure matching the tested code's location
- Merged a couple of test suites together where there were already existing tests both within and outside the appropriate directory structure (`surveyResponse.test.js`, `importSurveyResponses.test.js`)
- Extracted a couple of minor boilerplate setup functions to be reused, including granting full access to work as a core part of `TestableApp`, and setting up the dummy sync queue to be simpler in the client code.


EMcQ-BES (approved):

edmofro:
Merging straight to dev as the tests passing is all the testing we need :-)

PR #2037, opened 2021-02-10
1149 keep timestamp values as strings

edmofro:
### Issue # 
https://github.com/beyondessential/tupaia-backlog/issues/1149 towards point 3

When pulling dates out of the database, our node postgres driver converts them to JS dates. Annoyingly, this means they get parsed using the default logic of "if no timezone is provided, assume local time". The parsed date is then converted to UTC to be stored, because JS dates are just a UTC timestamp with some formatting on top. 

This behaviour messed me up for a while. The solution here stops that whole parsing behaviour, and just returns strings for timestamp columns, which I think keeps things more "expected", and might make your life easier @EMcQ-BES

### DIFF
--- a/packages/admin-panel/src/pages/resources/SurveyResponsesPage.js
+++ b/packages/admin-panel/src/pages/resources/SurveyResponsesPage.js
@@ -27,7 +27,7 @@ const date = {
   Header: 'Date of Survey',
   source: 'end_time',
   type: 'tooltip',
-  accessor: row => moment(row.end_time).local().toString(),
+  accessor: row => moment(row.end_time).local().format('ddd, MMM Do YYYY, HH:mm:ss ZZ'),
   filterable: false,
   editable: false,
 };
@@ -36,10 +36,7 @@ const dateOfData = {
   Header: 'Date of Data',
   source: 'data_time',
   type: 'tooltip',
-  accessor: row =>
-    moment(row.data_time || row.end_time)
-      .local()
-      .toString(),
+  accessor: row => moment.parseZone(row.data_time).format('ddd, MMM Do YYYY, HH:mm:ss'),
   filterable: false,
   editConfig: {
     type: 'datetime-local',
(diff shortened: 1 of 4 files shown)


packages/admin-panel/src/pages/resources/SurveyResponsesPage.js
```diff
@@ -27,7 +27,7 @@ const date = {
   Header: 'Date of Survey',
   source: 'end_time',
   type: 'tooltip',
-  accessor: row => moment(row.end_time).local().toString(),
+  accessor: row => moment(row.end_time).local().format('ddd, MMM Do YYYY, HH:mm:ss ZZ'),
```
edmofro, line 30: This file just proof of concept, the main change is the next file

packages/database/src/TupaiaDatabase.js
```diff
@@ -51,6 +52,10 @@ const VALID_COMPARISON_TYPES = ['where', 'whereBetween', 'whereIn', 'orWhere'];
 // keeping all the tests passing
 const HANDLER_DEBOUNCE_DURATION = 250;
 
+// turn off parsing of timestamp (not timestamptz), so that it stays as a sort of "universal time"
+// string, independent of timezones, rather than being converted to local time
+pg.types.setTypeParser(1114, val => val);
```
EMcQ-BES, line 57: I don't love the magic number here, are these ids enumerated somewhere?

edmofro: Yep, in a more recent version of `node-pg` 😞 

I can check out what's required to upgrade

EMcQ-BES (approved):

packages/database/src/TupaiaDatabase.js
edmofro: I've looked through the breaking changes from pg 7 -> 8 and we're safe on all counts, so I'm going to upgrade and use the oid constants provided there.

PR #2035, opened 2021-02-10
1149 meditrak data time

edmofro:
### Issue #: 
https://github.com/beyondessential/tupaia-backlog/issues/1149, subtasks 10, 11, 13, 14, 18

PR #2040, opened 2021-02-10
1149 replace submission time reads

EMcQ-BES:
### Changes:

- Update submission_time reads to instead read data_time value
- Update unit tests to run correctly with data_time


packages/meditrak-server/src/tests/dhis/EventBuilder.test.js
```diff
@@ -126,49 +126,4 @@ describe('EventBuilder', () => {
       });
     });
   });
-
-  describe('Timezone awareness', () => {
```
EMcQ-BES, line 130: I've removed these tests because they're testing that the timezone aware value converts correctly, which we are in the process of removing

packages/meditrak-server/src/tests/hooks/questionHooks.test.js
```diff
@@ -161,7 +161,7 @@ describe('Question hooks', () => {
         expect(spy2.called).to.be.true;
       });
 
-      it('Should not run a hook for a backdated value', async () => {
+      xit('Should not run a hook for a backdated value', async () => {
```
EMcQ-BES, line 164: This specific test writes and then reads the data, so I have currently broken it with updating the reads and it will need to be re-enabled and passing when all the writes are updated.

edmofro (changes requested):
Some stuff inline, plus as discussed we need to check each place the date is being parsed and formatted, as well as each place it's being used as a query parameter, to make sure they maintain their timezone agnosticism

edmofro:
Regarding the point about maintaining timezone agnosticism, I tried a fix in the AggregateDataPusher and it worked. Better check each place pretty thoroughly, but we can at least know that:
- Using the strip timezone util I introduced in https://github.com/beyondessential/tupaia/pull/2035 works as expected
- When dates are then passed through as timezone-stripped strings as db query parameters, they are interpreted correctly as timezone agnostic at the query level. 

On that second point, I ran some tests, and it all looks good. The db query for AggregateDataPusher that checks for duplicate survey responses comes out as
```
select "survey_response"."id", "survey_response"."survey_id", "survey_response"."user_id", "survey_response"."assessor_name", "survey_response"."start_time", "survey_response"."end_time", "survey_response"."metadata", "survey_response"."submission_time", "survey_response"."timezone", "survey_response"."entity_id", "survey_response"."data_time" from "survey_response" where "id" != '5ffbadd580df3713ff6eb029' and "survey_id" = '5b889d7bf013d654c493ae70' and "entity_id" = '5d3f8844854d9431bf043d2e' and "data_time" between '2021-01-01T00:00:00.000' and '2021-01-31T23:59:59.999' and "end_time" > '2021-01-11 14:45:57.774' order by "data_time" DESC;
```
Which has the `"data_time" between` clause using the correct datetime with no offset suffix. I played around with adding duplicates just inside and just outside the boundaries, and it picked up those within, and left out those outside.

@EMcQ-BES I'll leave making that update and identifying what others are required to you, but feel free to ask me questions (I have a fair bit of the context in my head now)

edmofro:
Oh, and the fix looked like this:
Screen Shot 2021-02-12 at 1 46 49 PM

Obviously that's using a util that's not in this branch, so you might like to branch off `1149-meditrak-data-time` (https://github.com/beyondessential/tupaia/pull/2035)

packages/meditrak-server/src/ms1/startSyncWithMs1.js
```diff
@@ -98,7 +98,7 @@ export async function pushChange(models, change, ms1Api) {
 
   const surveyResponseId = surveyResponse.id;
 
-  const { assessor_name: sentBy, submission_time: sentAt } = surveyResponse;
+  const { assessor_name: sentBy, data_time: sentAt } = surveyResponse;
```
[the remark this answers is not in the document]
kujo4pm: Yeah definitely... especially given that we're presenting data on a map overlay - it will be easier to audit.

packages/meditrak-server/src/tests/routes/postChanges.test.js
```diff
@@ -460,6 +469,86 @@ describe('POST /changes', async () => {
...
+            {
+              action: 'SubmitSurveyResponse',
+              payload: surveyResponseObject,
+            },
+          ],
+        });
+
+        expect(response).to.have.property('statusCode', 400);
+      });
+    });
```
edmofro, line 551: Great new test cases, thanks

packages/meditrak-server/src/ms1/startSyncWithMs1.js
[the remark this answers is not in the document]
edmofro: Great :-)

packages/meditrak-server/src/tests/routes/surveyResponse.test.js
```diff
@@ -394,17 +395,17 @@ describe('surveyResponse endpoint', () => {
...
-        '2019-07-31T05:48:00+12:00',
-      ],
-    );
+    expect(moment.utc(dbResponse.data_time).tz(dbResponse.timezone).format()).to.be.oneOf([
+      '2019-07-31T06:48:00+13:00',
+      // While 'Antarctica/McMurdo' is in DST time, it will be selected as the timezone,
+      // however the timestamp is hardcoded to a date outside of DST time for McMurdo, so it will be
+      // parsed with it's non-DST +12 offset.
+      '2019-07-31T05:48:00+12:00',
+    ]);
```
edmofro, line 404: I find this a weird test case. I know you're trying to make the best of the old case, but I just don't think it works as well, or is even necessary, with `data_time`. Let's just remove it, unless you want to argue for its value?

edmofro (approved):
Looking really good! I find it hard to know whether all of the changes will work as expected, because timezones, but that just means we'll need solid testing (and testing notes) to be sure.

packages/meditrak-server/src/tests/routes/surveyResponse.test.js
```diff
@@ -369,45 +369,6 @@ describe('surveyResponse endpoint', () => {
...
-    const { surveyResponseId } = body.results[0];
-    const dbResponse = await models.surveyResponse.findOne({ id: surveyResponseId });
-    expect(moment.utc(dbResponse.data_time).tz(dbResponse.timezone).format()).to.be.oneOf([
-      '2019-07-31T06:48:00+13:00',
-      // While 'Antarctica/McMurdo' is in DST time, it will be selected as the timezone,
-      // however the timestamp is hardcoded to a date outside of DST time for McMurdo, so it will be
-      // parsed with it's non-DST +12 offset.
-      '2019-07-31T05:48:00+12:00',
-    ]);
-    expect(dbResponse.timezone).to.be.oneOf(timezones);
```
edmofro: Oh sorry, I actually _do_ like this test (adding a `timezone` string based on the `timestamp`'s offset), but don't think we need the `expect` above it (which is essentially just testing that `data_time` when converted to `utc` and then back to its timezone comes out as what it was before)

EMcQ-BES: Oh that makes sense, sorry for the confusion

edmofro (approved):

PR #2045, opened 2021-02-11
2226: Add sub_facility entity type

chris-bes:
### Changes:

- Add sub_facility entity type

### DIFF
--- a/packages/database/src/migrations/20210211041322-AddSubFacilityEntityType-modifies-schema.js
+++ b/packages/database/src/migrations/20210211041322-AddSubFacilityEntityType-modifies-schema.js
@@ -1,0 +1,33 @@
+'use strict';
+
+import { TupaiaDatabase } from '../TupaiaDatabase';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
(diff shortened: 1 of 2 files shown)


packages/database/src/migrations/20210211041322-AddSubFacilityEntityType-modifies-schema.js
```diff
@@ -0,0 +1,33 @@
...
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function () {
+  const db = new TupaiaDatabase();
+  await db.executeSql(`
+    ALTER TYPE public.entity_type ADD VALUE IF NOT EXISTS 'sub_facility';
+  `);
+  db.closeConnections();
```
rohan-bes, line 24: `db.closeConnections();` should we be doing this at the end of migrations? If so, woops I haven't been doing it, will start hehe

rohan-bes (approved):
Nice, looks good!

packages/database/src/migrations/20210211041322-AddSubFacilityEntityType-modifies-schema.js
chris-bes: This migration was actually just a copy paste from other migrations that adds new entity type (eg: https://github.com/beyondessential/tupaia/pull/568 or https://github.com/beyondessential/tupaia/pull/368). I'm also curious but I'm sure there was a reason for it so I didn't bother changing it 😄

PR #2046, opened 2021-02-11
2163 support number and percentage in excel export download

billli0:
### Issue #:
[2163 - inclusion of percentage in exported excel sheet](https://github.com/beyondessential/tupaia-backlog/issues/2163)

### Changes:
- A bit eslint formating in `valueAndPercentageByAgeByFlightDate.js`
- Support `numberAndPercentage` in excel file download


packages/web-config-server/src/export/excelFormatters/formatMatrixDataForExcel.js
```diff
@@ -124,12 +127,22 @@ const convertAoaToAoo = data => {
...
 
   return value;
-}
+};
+
+const numberAndPercentage = (value, { numerator, denominator }) => {
+  if (isNaN(value)) return value;
+  const percentage = (numerator / denominator) * 100;
+  return `${value} (${percentage === 0 ? 0 : percentage.toFixed(1)}%)`;
+};
```
chris-bes, line 148: I think it's better if we move this method to @tupaia/utils just in case we want to re use it.

Ideally I think we want to move all of the formatters in `formatters.js` from @tupaia/web-frontend to @tupaia/utils so we can reuse them in both web-frontend and web-config-server. But because of this issue (https://github.com/beyondessential/tupaia-backlog/issues/1798), web-frontend cannot use any internal dependency.

This also looks a bit simpler than the version in formatter.js as well...?

packages/web-config-server/src/export/excelFormatters/formatMatrixDataForExcel.js
```diff
@@ -124,12 +127,22 @@ const convertAoaToAoo = data => {
 
 const isEmpty = data => data === undefined || data === null;
 
-const addValueOrEmpty = value => {
+const addValueOrEmpty = (value, valueType) => {
   if (isEmpty(value)) return '';
 
   if (typeof value === 'object') {
+    if (valueType === VALUE_TYPES.NUMBER_AND_PERCENTAGE) {
```
chris-bes, line 134: Do you think we should make this similar to the pattern in `formatter.js` (check line 102)?

```
const VALUE_TYPE_TO_FORMATTER = {
  [VALUE_TYPES.NUMBER_AND_PERCENTAGE]: numberAndPercentage,
};
```

chris-bes (changes requested):
LGTM! Just 1 or 2 suggestions

packages/web-config-server/src/export/excelFormatters/formatMatrixDataForExcel.js
billli0: Thanks Chris for sharing the latest news!
 
I'm not pretty familar about `formatters.js` on frontend but I guess it is for formatting the value before visualizing. If so I guess the best practice might be sending the value to frontend, after formating it instead of formating them on frontend? That's the thing in my head when looking at `formatters.js`. Unless there is any special case or other usage on front end.

chris-bes: I think that would require a bit of refactor since we work out which formatter that we want to use based on `valueType` config in `viewJson`, so it should only apply in the front end

billli0: Yes you are right, here we also use `valueType` in `viewJson`. Now we send `value: 0, { numerator, denominator}` then we calculate them to `0 (12%)` in front end. I just think that if there is a better solution to send `0 (12%)` directly to front end. 

And yeah if we do so we should do rafactor and a huge testing... In conclusion I'm happy to move that to `utils` 😄

packages/utils/src/formatDataValueByType.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * Tupaia Web
+ * Copyright (c) 2019 - 2021 Beyond Essential Systems Pty Ltd.
+ * This source code is licensed under the AGPL-3.0 license
+ * found in the LICENSE file in the root directory of this source tree.
+ */
```
chris-bes, line 6: Since it's in @tupaia/utils, perhaps we can change it to:

```
/**
 * Tupaia
 */
```

chris-bes (approved):
LGTM! Just 1 minor suggestion so pre approve!

packages/utils/src/formatDataValueByType.js
billli0: Thanks Chris for focusing on details. 👍 I just copy and paste and didn't look at it closely. I will change it later.

PR #2047, opened 2021-02-11
2230: Allow import data service entity when importing entities in Admin Panel

chris-bes:
### Changes:

- Allow import data service entity when importing entities in Admin Panel


IgorNadj (approved):
LGTM

PR #2054, opened 2021-02-12
2190 project sticky region labels

tcaiger:
### Changes:

- Replaced `tile_sets` column with `config` in the projects table
- Updated project selectors to use config for tile sets
-  Configure Regional Labels (tooltips) based on the config

---

### Screenshots:
See issue

### DIFF
--- a/packages/admin-panel/src/pages/resources/ProjectsPage.js
+++ b/packages/admin-panel/src/pages/resources/ProjectsPage.js
@@ -51,10 +51,11 @@ const FIELDS = [
     type: 'tooltip',
   },
   {
-    Header: 'Tile Sets',
-    source: 'tile_sets',
-    secondaryLabel: 'Comma separated list (eg. osm,satellite,terrain).',
-    type: 'tooltip',
+    Header: 'Config',
+    source: 'config',
+    type: 'jsonTooltip',
+    editConfig: { type: 'jsonEditor' },
+    secondaryLabel: 'eg. { "tileSets": "osm,satellite,terrain", "permanentRegionLabels": true }',
   },
   {
     Header: 'Sort',
--- a/packages/web-frontend/src/containers/Map/ConnectedPolygon.js
+++ b/packages/web-frontend/src/containers/Map/ConnectedPolygon.js
@@ -21,6 +21,7 @@ import {
   selectAllMeasuresWithDisplayInfo,
   selectCurrentMeasureId,
   selectOrgUnitChildren,
+  selectAreRegionLabelsPermanent,
 } from '../../selectors';
 import ActivePolygon from './ActivePolygon';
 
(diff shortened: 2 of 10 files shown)


packages/admin-panel/src/pages/resources/ProjectsPage.js
```diff
@@ -51,10 +51,11 @@ const FIELDS = [
...
   {
-    Header: 'Tile Sets',
-    source: 'tile_sets',
-    secondaryLabel: 'Comma separated list (eg. osm,satellite,terrain).',
-    type: 'tooltip',
+    Header: 'Config',
+    source: 'config',
+    type: 'jsonTooltip',
+    editConfig: { type: 'jsonEditor' },
+    secondaryLabel: 'eg. {tileSets: osm,satellite,terrain, stickyRegionLabels: true}',
```
edmofro, line 58:
```suggestion
    secondaryLabel: 'eg. {"tileSets": ["osm","satellite","terrain"], "stickyRegionLabels": true}',
```

edmofro: Oh, looking at the tests perhaps it's not a fully formatted array - even more important to specify I reckon!

packages/database/src/migrations/20210211224521-AddConfigToProjects-modifies-schema.js
```diff
@@ -0,0 +1,37 @@
...
+};
+
+/**
+ *
+ *
+ * @param db
+ * @returns {{replacementParams: *, query: *}}
+ */
+exports.up = function (db) {
+  db.runSql(`ALTER TABLE project DROP COLUMN IF EXISTS tile_sets;`);
```
edmofro, line 24: Do we need to migrate the old `tile_sets` data across to `config`, for `laos_eoc`? Seems safer than relying on us remembering to add it back in through the admin panel

edmofro (changes requested):
Nice solution, and good call not making yet another column!

I wonder about `sticky` being the term we refer to this behaviour? `sticky` to me sounds like it would be _more_ permanent, rather than less...I don't mind the sticking-to-mouseness being tied to the permanence (though I do wonder if that was necessary or you thought it was just nicer?), but I reckon we could use a term that was more related to the primary aspect of permanence, e.g. `permanentRegionLabels` with a default of `true`, or `regionLabelsOnHover` with a default of `false`.

tcaiger:
@edmofro - ready for re-review. I updated the names to `permanentRegionLabels` with defaults of true and it is much clearer. I played around with `permanent` off and `sticky` off on the tooltips and you are 100% right that it works that way. I definitely prefer the UX of sticky though so I've stuck 😉  with that.

packages/web-frontend/src/containers/Map/ConnectedPolygon.js
```diff
@@ -219,8 +219,9 @@ const mapStateToProps = (state, givenProps) => {
   const orgUnit = selectOrgUnit(state, organisationUnitCode);
   const coordinates = orgUnit ? orgUnit.location.region : undefined;
 
+  console.log('selectAreRegionLabelsPermanent(state)', selectAreRegionLabelsPermanent(state));
```
edmofro, line 222:
```suggestion
```

edmofro (approved):
Stuck hehe

Nice, this is good to go once that console log is deleted :-)

PR #2056, opened 2021-02-14
2155 chart components part 1

tcaiger:
### Issue #: [2155 Chart Components](https://github.com/beyondessential/tupaia-backlog/issues/2155)

This is part 1 of 4 for adding chart components to the ui-components package.

### Changes:

- Make a copy of charts components in ui components (Pie, Line, Area & Bar charts)
- Refactor charts to be small functional components

Note: There are a lot of constants and helpers used in the charts components. I discussed with Edwin how to manage them in ui-components. We decided that the best solution for now is to simply duplicate them. Once lesmis is done and web-frontend is moved to use ui-components, we will probably be able to delete most of the duplicates from web-frontend. I've made [another card](https://github.com/beyondessential/tupaia-backlog/issues/2241) to pick that up later.

---

### Screenshots:
See Issue


packages/ui-components/src/components/Chart/PieChart.js
```diff
@@ -0,0 +1,256 @@
...
+
+  const getValidData = () => {
+    const { data, valueType } = viewContent;
+
+    return data
+      .filter(element => element.value > 0)
+      .map(item => {
+        const { name, ...otherKeyValues } = item;
+        // Map names to labels if available
+        let label = getPresentationOption(name, 'label');
```
nick-beyondessential, line 182:
```suggestion
        const label = getPresentationOption(name, 'label') ?? name;
```
Though ?? will treat falsey as valid.

packages/ui-components/src/components/Chart/utils.js
```diff
@@ -0,0 +1,134 @@
...
+  const place = 10 ** decimalPlace;
+  return Math.floor(number * place) / place;
+};
+
+// Note: will display 0 if passed undefined
+const currency = value => numeral(value).format('$0.00a');
+
+const fraction = (value, { total }) => {
+  if (isNaN(total)) return 'No data';
+  return `${String(value)}/${String(total)}`;
```
nick-beyondessential, line 29: Maybe this should return 0 if both value and total are 0?

nick-beyondessential (approved):
Looks good to me! Nice work.

PR #2057, opened 2021-02-14
2155 chart components part 2

tcaiger:
### Issue #: [2155 Chart Components](https://github.com/beyondessential/tupaia-backlog/issues/2155)

This is part 2 of 4 for adding chart components to the ui-components package.

### Changes:

- Copy and refactor Cartesian Chart component
- Copy and refactor supporting chart components (Tooltips etc). These components are still very similar to web-frontend. It is not very easy to improve them due to the idiosyncrasies of re-charts.

Note: There are a lot of constants and helpers used in the charts components. I discussed with Edwin how to manage them in ui-components. We decided that the best solution for now is to simply duplicate them. Once lesmis is done and web-frontend is moved to use ui-components, we will probably be able to delete most of the duplicates from web-frontend. I've made [another card](https://github.com/beyondessential/tupaia-backlog/issues/2241) to pick that up later.

---

### Screenshots:
See Issue


packages/ui-components/src/components/Chart/CartesianChart.js
```diff
@@ -0,0 +1,227 @@
...
+  const getIsActiveKey = legendDatakey =>
+    activeDataKeys.length === 0 ||
+    activeDataKeys.includes(legendDatakey) ||
+    legendDatakey === LEGEND_ALL_DATA_KEY;
+
+  const updateChartConfig = hasDisabledData => {
+    const newChartConfig = { ...chartConfig };
+
+    if (hasDisabledData && !chartConfig[LEGEND_ALL_DATA_KEY]) {
+      const allChartType = Object.values(chartConfig)[0].chartType || chartType || 'line';
```
nick-beyondessential, line 93: @tcaiger, where does the `chartType` variable here come from?

nick-beyondessential: Nevermind - found it defined on line 140.

packages/ui-components/src/components/Chart/Tooltip.js
```diff
@@ -0,0 +1,129 @@
...
+        formatLabelledValue(name, value, valueTypeForLabel, metadata)
+      )}
+    </TooltipContainer>
+  );
+};
+
+export const Tooltip = props => {
+  const { payload, active, presentationOptions } = props;
+
+  const data = payload || []; // This is to hancle when recharts overrides the payload as null
```
nick-beyondessential, line 102: handle

nick-beyondessential (approved):
Nice work Tom, looks great!

PR #2059, opened 2021-02-14
2155 chart components part 4

tcaiger:
### Changes:

- Add chart components to storybook
- Add a login form to storybook so that private dashboard report data can be used to test the charts

---

### Screenshots:
See Issue


packages/ui-components/stories/story-utils/LoginForm.js
```diff
@@ -0,0 +1,105 @@
...
+  padding-bottom: 1rem;
+`;
+
+const OpenLoginButton = styled(Button)`
+  position: absolute;
+  top: 30px;
+  right: 30px;
+`;
+
+const baseUrl = process.env.REACT_APP_CONFIG_SERVER_BASE_URL || 'http://localhost:8000/api/v1/';
```
nick-beyondessential, line 37: Should localhost be set as a env variable too?

nick-beyondessential (approved):
Looks good!

tcaiger:
@edmofro - in regards to your previous suggestions.

**Naming of Chart**
I was imagining that the component structure would end up being something like: Chart and Matrix as their own components and then maybe Report as a HOC above that. Looking at the props between Matrix and Chart they look difference enough to me to warrant different components and also I feel that they are conceptually different things. I think it will be easier to decide when I pull out the Matrices so how about we leave it as chart for now?

**Replace ViewId with ReportId** 
I have updated the charts to use `reportId` instead of `viewId` thanks for the suggestions.

PR #2060, opened 2021-02-14
[WIP] 2239 add geo data for laos districts sub districts

IgorNadj:
### Changes:

- add geo data for laos districts sub districts


IgorNadj:
Branch name too long, recreating

PR #2062, opened 2021-02-15
2239 add geo data for laos

IgorNadj:
### Changes:

- Add migration to clean up geographical_area for laos
- Entity importer can now import facilities at all levels
- Fix entity hierarchy for laos_eoc


nick-beyondessential (approved):
LGTM

PR #2065, opened 2021-02-15
Add raw data download to FETP

PShoe:
### Changes:
- Add dashboard group
- Add dashboard reports to Solomon Islands and PNG

Note: I suspect there is a less repetitive way of adding same report to two countries.

---

### Screenshots:

image


nick-beyondessential:
Also better practise to link screenshots to card rather than PR in case they contain sensitive data.

PShoe:
@nick-beyondessential back to you for review, switched to rawDataValues - chat to me about this tomorrow?

packages/database/src/migrations/20210214234630-AddDataDownloadNationalFETP-modifies-data.js
```diff
@@ -0,0 +1,93 @@
...
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const DATA_ELEMENT_CODE = 'FGDS';
```
nick-beyondessential, line 19: Bit pedantic but I'd rename this variable `SURVEY_CODE`

nick-beyondessential (approved):
Looks good, optional name suggestion

packages/database/src/migrations/20210214234630-AddDataDownloadNationalFETP-modifies-data.js
PShoe: done, thanks!

PR #2071, opened 2021-02-15
1799: Indicator optimisation fixes

kael89:
### Issue #:
Final part in https://github.com/beyondessential/tupaia-backlog/issues/1799. Addresses all pending issues identified in [previous PRs](https://github.com/beyondessential/tupaia/pull/1873#issuecomment-772936894)

### Commit overview

I suggest reviewing this PR one commit at a time

#### Part 1: Fix and refactor `aggregator` utils

bb54c0328 Extract compare utils to their own file
0a4ceb1f0 Move periodToDateString and add tests
5c57d41e9 Refactor buildAndInsert methods
5a125a28d Use util
3d5badfb9 Use arrayToAnalytics for test data
e8e2c5355 getDateRange methods should return date ranges, not periods!
37136999d Add keepAnalyticsMatchingFetchOptions util
13cdfe0fd Add adjustFetchOptions utils
a5e671fb0 Add test cases

* Fixed some existing issues. It was a coicidence that they hadn't surfaced till now (eg because we are using weekly, and not monthly or annual, period types in PSSS indicators)
* Introduced the `AggregationList` concept: an array of aggregations to be applied to a data element (the concept is implied in our codebase, this PR gives it a name)
* Added utils to adjust `fetchOptions` to aggregation lists

#### Part 2: Fix nested aggregation logic

3f802fa74 Fix nested aggregation building logic

A lot of things happen at the same time in this commit. I'll try to mitigate this by listing the main changes here:

* Removed references to a "populated" analytics repo. The fact that the repo must be populated before used is an implementation detail, which builders eg `ArithmeticBuilder` need not be aware of. `AnalyticsRepo` is responsible for throwing an error if its fields are accessed before data get fetched.
* Fixed the aggregation logic in `AnalyticsRepo.getAggregatedAnalytics`, as [discussed](https://github.com/beyondessential/tupaia/pull/1873#issuecomment-761892449)
* Removed the distinction between "root" and "nested" indicators 🎉 

#### Part 3: Update tests

459f1363d Minor fixture updates
0746213a7 Use .ts files for fixtures - just moving things around
734f8c07a Update unit test

I tried using `.ts` instead of `.json` files, and I think it's much better this way! Thanks for suggesting it 🙌


packages/indicators/src/Builder/Builder.ts
```diff
@@ -23,29 +20,24 @@ export abstract class Builder {
...
-      );
-    }
-
-    return this.buildAnalyticValues(
-      populatedAnalyticsRepo,
-      buildersByIndicator,
-      fetchOptions,
-    ).map(value => ({ ...value, dataElement: this.indicator.code }));
-  };
+    extraAggregations: Aggregation[] = [],
```
kael89, line 31: Those aggregations are passed top-down between nested indicators, similar to how `props` are propagated in React components.

I couldn't of a better name, but please let me know if you do!

kael89:
@edmofro The moment of truth... Benchmarks!

[Indicator Optimisation Benchmark.zip](https://github.com/beyondessential/tupaia/files/5984691/Indicator.Optimisation.Benchmark.zip)

My conclusions:
* The "optimised" indicators have a 2x-3x improved performance on raw `report-server` requests
* On my local PSSS, the performance gain is negligible
* On feature instances, the performance gain is noticeable from a subjective user's perspective, although not as impressive as in the raw `report-server` measurements

And my theory:
When loading PSSS pages we trigger multiple async requests. It seems that my local machine fares much better with parallelizing requests than our feature instances, even to the extend that `20 countries * 0.6 secs/country indicator => 0.6 secs page load time` (all parallel!). This is a bit similar with @rohan-bes 's recent findings when comparing `dev` DB queries with his local. 

The above hint that infrastructure optimisation may be highly beneficial for performance, even with a more "naive" code approach, in some cases.

packages/utils/src/period/getDefaultPeriod.js
```diff
@@ -6,7 +6,8 @@
 import { convertDateRangeToPeriods } from './convertDateRangeToPeriods';
 import { utcMoment } from '../datetime';
 
-export const EARLIEST_DATA_DATE = utcMoment('2017-01-01'); // Tupaia started in 2017
+export const EARLIEST_DATA_DATE_STRING = '2017-01-01'; // Tupaia started in 2017
```
edmofro, line 9: Interesting to be reminded of this...completely outside the scope of this PR but we do now have PSSS data stretching back to 2015, though I guess if we need to show it on tupaia.org @enunan will just need to create a ticket for extending the earliest date supported on tupaia :-)

packages/database/src/testUtilities/buildAndInsertSurveys.js
```diff
@@ -28,19 +28,23 @@ const buildAndInsertQuestion = async (
...
-const buildAndInsertDataGroup = async (models, fields) =>
-  findOrCreateDummyRecord(
+const buildAndInsertDataGroup = async (models, fields) => {
+  const { code, type, ...createFields } = fields;
+  return findOrCreateDummyRecord(
     models.dataSource,
-    { service_type: 'dhis', ...fields, type: 'dataGroup' },
-    { config: { isDataRegional: true } },
+    { code, type: 'dataGroup' },
+    { service_type: 'tupaia', ...createFields },
```
edmofro, line 36: Hooray for `tupaia` as the primary use case supported

packages/database/src/testUtilities/buildAndInsertSurveys.js
```diff
@@ -28,19 +28,23 @@ const buildAndInsertQuestion = async (
...
-const buildAndInsertDataElement = async (models, fields) =>
-  findOrCreateDummyRecord(
+const buildAndInsertDataElement = async (models, fields) => {
+  const { code, type, ...createFields } = fields;
+  return findOrCreateDummyRecord(
     models.dataSource,
-    { service_type: 'dhis', ...fields, type: 'dataElement' },
-    { config: { isDataRegional: true } },
+    { code, type: 'dataElement' },
+    { service_type: 'tupaia', createFields },
```
edmofro, line 45: `...createFields`

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/offsetPeriod.test.js
```diff
@@ -188,17 +187,35 @@ describe('offsetPeriod', () => {
...
-        [-2, { startDate: '20200101', endDate: '20221231' }],
-        [-1, { startDate: '20190101', endDate: '20211231' }],
-        [0, { startDate: '20180101', endDate: '20201231' }],
-        [1, { startDate: '20170101', endDate: '20191231' }],
-        [2, { startDate: '20160101', endDate: '20181231' }],
+        [-2, { startDate: '2020-01-01', endDate: '2022-12-31' }],
+        [-1, { startDate: '2019-01-01', endDate: '2021-12-31' }],
+        [+0, { startDate: '2018-01-01', endDate: '2020-12-31' }],
+        [+1, { startDate: '2017-01-01', endDate: '2019-12-31' }],
+        [+2, { startDate: '2016-01-01', endDate: '2018-12-31' }],
```
edmofro, line 200: Nicely aligned with that extra `+` 😁

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/offsetPeriod.js
```diff
@@ -24,31 +31,37 @@ const validateOffsetPeriodConfig = (config = {}) => {
...
+  // so that enough data will be available for its calculation
+  const offsetCompensation = config.offset * -1;
+
+  const getMomentWithOffset = dateString => {
+    const momentUnit = periodTypeToMomentUnit(periodType);
+    return utcMoment(dateString).add(offsetCompensation, momentUnit);
+  };
+
+  const startMoment = getMomentWithOffset(dateRange.startDate);
+  const endMoment = getMomentWithOffset(dateRange.endDate).endOf(periodType);
```
edmofro, line 61: Interested in what this does when going from e.g. `2020-01-31` with an offset of 1 so we're wanting to include data up to `2020-02-28`. Worried it would end up at `2020-03-31`. You've got a good edge case test for going from a smaller month to a larger one, but how about a larger to a smaller?

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/adjustFetchOptionsToAggregation.test.js
```diff
@@ -0,0 +1,109 @@
...
+      { startDate: '2020-02-01', endDate: '2020-03-31', period: '202002;202003' },
+    ],
+    [
+      'adjustments in the same direction',
+      [getOffsetMonthsAgg(-1), 'FINAL_EACH_MONTH', getOffsetMonthsAgg(-2)],
+      { startDate: '2020-04-01', endDate: '2020-05-31', period: '202004;202005' },
+    ],
+    [
+      'adjustments in different directions: they cancel each other',
+      [getOffsetMonthsAgg(-1), 'FINAL_EACH_MONTH', getOffsetMonthsAgg(1)],
```
edmofro, line 42: Oh woah cool!

packages/aggregator/src/analytics/aggregateAnalytics/adjustFetchOptionsToAggregation.js
```diff
@@ -0,0 +1,69 @@
+/**
```
edmofro, line 1: Filename should probably have `List` on the end

packages/indicators/src/__tests__/__integration/indicatorApi.fixtures.json
```diff
@@ -41,15 +53,51 @@
...
       }
+    ],
+    "surveyResponses": [
+      {
+        "surveyCode": "Births",
+        "entityCode": "AU",
+        "submission_time": "2020-12-10T15:00:00Z",
+        "answers": {
+          "Red": "10",
+          "Blue": "20"
```
edmofro: Red and Blue births sound a bit morbid... 😅

packages/indicators/src/__tests__/__integration/arithmeticAnalytics.fixtures.json
```diff
@@ -99,6 +95,64 @@
...
+                    { "type": "OFFSET_PERIOD", "config": { "offset": 1, "periodType": "QUARTER" } }
+                  ]
+                }
+              }
+            ],
+            "aggregation": "FINAL_EACH_QUARTER"
+          }
+        },
+        {
+          "code": "AverageOf3QuarterlyIncreases",
```
edmofro: This is a cool demonstration of the power of all of the pieces!

packages/utils/src/period/getDefaultPeriod.js
kael89: Good to know! Perhaps this kind of information should live in the DB (and be configurable via the admin panel)

packages/database/src/testUtilities/buildAndInsertSurveys.js
kael89: Oops!

packages/aggregator/src/analytics/aggregateAnalytics/adjustFetchOptionsToAggregation.js
kael89: I had it like that initially but then omitted it, since two methods are exported:
* `adjustFetchOptionsToAggregationList`
* `adjustFetchOptionsToAggregationLists`

Happy to use `adjustFetchOptionsToAggregationList` though!

packages/indicators/src/__tests__/__integration/indicatorApi.fixtures.json
kael89: Indeed! 😆 Fixed in https://github.com/beyondessential/tupaia/commit/0746213a7b99f994f34a8d670b698d53a84bd0f7 (generally a housekeeping commit, but this change sneaked in)

packages/indicators/src/__tests__/__integration/arithmeticAnalytics.fixtures.json
kael89: Indeed! It exposes a lot of aspects of nested aggregations

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/offsetPeriod.js
kael89: The behaviour going from larger to smaller is already tested in [offsetPeriod tests](https://github.com/beyondessential/tupaia/blob/dev/packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/offsetPeriod.test.js#L167). Based on [this comment](https://github.com/beyondessential/tupaia/blob/dev/packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/offsetPeriod.test.js#L193), I am only testing this edge case cause going from smaller to larger is different for periods:
`20190430 + 1 month` => `20190530`

and date strings:
`2019-04-30 + 1 month` => `2019-05-30`

I personally it's fine as is, but let me know if you think it's a bit unclear, and I can add a smaller to larger test case for `getDateRangeForOffsetPeriod` too

kael89: Hey but I just realised the two functions have now become separated, as they should (since the one uses periods, and the other date strings)!! So I'll update the tests here to include a complete period type suite. Nice catch!

kael89:
@edmofro All current comments have been addressed

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/offsetPeriod.js
```diff
@@ -51,14 +51,12 @@ export const getDateRangeForOffsetPeriod = (dateRange, config) => {
...
-    const momentUnit = periodTypeToMomentUnit(periodType);
-    return utcMoment(dateString).add(offsetCompensation, momentUnit);
-  };
-
-  const startMoment = getMomentWithOffset(dateRange.startDate);
-  const endMoment = getMomentWithOffset(dateRange.endDate).endOf(periodType);
+  const startMoment = utcMoment(dateRange.startDate).add(offsetCompensation, momentUnit);
+  const endMoment = utcMoment(dateRange.endDate)
+    .add(offsetCompensation, momentUnit)
+    .endOf(momentUnit);
```
kael89, line 59: The new tests revealed a small bug - I was using `periodType` instead `momentUnit` here

packages/utils/src/period/getDefaultPeriod.js
edmofro: Yeah that could work, like an earliest date per project setting. For now we're still ok with it as is, so will resolve this conversation

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/offsetPeriod.js
edmofro: Great catch by the tests

packages/indicators/src/IndicatorApi.ts
```diff
@@ -77,8 +77,8 @@ export class IndicatorApi {
       );
       currentIndicatorCodes = indicatorCodes;
       dataElementCodes.push(...nonIndicatorCodes);
-      const currentAggregations = getAggregationsForBuilders(currentBuilders);
-      aggregations.push(...currentAggregations);
+      const listMapsByBuilder = getAggregationListMapsByBuilder(currentBuilders);
+      aggregationListMap = addEntriesInAggregationListMap(aggregationListMap, listMapsByBuilder);
```
edmofro, line 81: Potentially could make better separation of concerns (see my comment about your comment indicating the behaviour of calling code from within the util), and possibly make things easier to understand if:
- instead of running the merge process with each iteration, we just build up an array of either all builders or all listMapsByBuilders
- at the end, we then send that whole array through to the util, which merges and flattens them all (meaning it has the understanding of order etc. within it)

What do you think?

edmofro (changes requested):
This review is for "the big commit" only.

The system is very smart! I like some of the renames (especially populate -> fetch). 

In all honesty though, I find it really, really hard to follow. I'm now at the point where I'm basically giving up on getting a full grasp - I understand the general flow of things, but find the specifics very tough to follow through. I think that's mostly just because it's doing a complex thing, but a couple of things that might make it slightly easier:
- It is quite agonising trying to parse the difference between AggregationListMap, AggregationListsMap, AggregationListMaps, AggregationListsMaps. I wonder if we could make them more distinctive somehow? I'm not sure how, but pulling out those `s`'s did my head in
- Trying to visualise the underlying data structure was even harder when flipping between those names, perhaps the odd example in comments could help

Nice work on wrapping your head around all of this!

packages/indicators/src/__tests__/__integration/fixtures/arithmeticAnalyticsFixtures.ts
```diff
@@ -0,0 +1,234 @@
...
+  setup: {
+    dbRecords: {
+      entity: ENTITIES,
+      indicator: entriesToArithmeticIndicators(ARITHMETIC_INDICATOR_ENTRIES),
+    },
+    surveys: SURVEYS,
+    surveyResponses: ARRAY_SURVEY_RESPONSES.map(arrayToSurveyResponse),
+  },
+  testCases: ARRAY_TEST_CASES.map(arrayToTestCase),
+};
```
edmofro, line 241: Oooh yes nice, using .ts does make things prettier

packages/indicators/src/__tests__/__integration/fixtures/arithmeticAnalyticsFixtures.ts
```diff
@@ -172,19 +179,19 @@ const ARRAY_TEST_CASES: ArrayTestCase[] = [
...
-    [['MonthlyBirthIncreaseRatio', 'GR', '20201230', 2]],
+    [['MonthlyBirthIncreaseRate', 'GR', '20201230', (14 - 7) / 7]],
   ],
   [
     'Nested indicator + inline indicator (parameter)',
-    ['AnnualBirthIncreaseRatio'],
+    ['AnnualBirthIncreaseRate'],
     ['2019-01-01', '2020-12-31', ['GR']],
-    [['AnnualBirthIncreaseRatio', 'GR', '2020', 7]],
+    [['AnnualBirthIncreaseRate', 'GR', '2020', (7 + 14 - 3) / 3]],
```
edmofro, line 194: I like this! Showing where the answer comes from rather than just the answer

packages/indicators/src/__tests__/__integration/fixtures/helpers.ts
```diff
@@ -9,8 +9,8 @@ import { arrayToAnalytics } from '@tupaia/data-broker';
 export type ArraySurveyResponse = [string, string, string, Record<string, string>];
 
 export const arrayToSurveyResponse = (arrayResponse: ArraySurveyResponse) => {
-  const [surveyCode, entityCode, submissionTime, answers] = arrayResponse;
-  return { surveyCode, entityCode, submission_time: submissionTime, answers };
+  const [surveyCode, entityCode, date, answers] = arrayResponse;
+  return { surveyCode, entityCode, submission_time: `${date}T15:00:00Z`, answers };
```
edmofro, line 13: Upcoming conflict @EMcQ-BES

packages/indicators/src/AnalyticsRepository.ts
```diff
@@ -35,94 +38,57 @@ interface Fields {
...
-    const { startDate, endDate } = fetchOptions;
-    const startPeriod = dateStringToPeriod(startDate);
-    const endPeriod = dateStringToPeriod(endDate);
-
-    return analytics.filter(
-      ({ period }) =>
-        comparePeriods(period, startPeriod) >= 0 && comparePeriods(period, endPeriod) <= 0,
+    const analyticsMatchingOptions = keepAnalyticsMatchingFetchOptions(analytics, adjustedOptions);
+    return this.aggregator.aggregateAnalytics(
+      analyticsMatchingOptions,
```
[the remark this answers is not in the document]
kael89: My impression is that here we use the exact range of analytics required. The aggregations for this specific data element are propagated from the top level down (via `extraAggregations`), forming the `aggregationList` which is passed to this function.
We then keep only the analytics that would have been fetched if we would use those exact aggregations, using `keepAnalyticsMatchingFetchOptions`.

But I may have missed something here but please let me know! 🙂

packages/indicators/src/IndicatorApi.ts
kael89: Good point, that's definitely doable!

packages/indicators/src/Builder/Builder.ts
kael89: I like it!

packages/indicators/src/mergeAggregationListsMaps.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const addEntry = ([entryCode, map]: [string, AggregationListsMap]) => {
+    const existingListsForEntry = newMap[entryCode];
+
+    Object.entries(map).forEach(([dataElement, elementLists]) => {
+      const listsToInsert = existingListsForEntry
+        ? // If there are existing lists for the current `entryCode`, we want to
+          // combine them with the new `elementLists`
+          existingListsForEntry
+            // Existing entry list was pushed before the new element list, so keep it in the start
```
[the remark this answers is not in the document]
kael89: Good point! 

I realise that the `entryCode` concept is totally confusing and not self-explanatory. I believe that that following [your suggestion](https://github.com/beyondessential/tupaia/pull/2071#discussion_r577258523) will make the logic much clearer!

packages/indicators/src/Builder/ArithmeticBuilder/ArithmeticBuilder.ts
```diff
@@ -102,41 +120,38 @@ export class ArithmeticBuilder extends Builder {
...
-    populatedAnalyticsRepo: AnalyticsRepository,
+    analyticsRepo: AnalyticsRepository,
     buildersByIndicator: Record<string, Builder>,
-    fetchOptions: FetchOptions,
+    extraAggregations: Aggregation[],
   ) =>
     this.getVariables()
       .map(variable => {
-        const analytics = this.getAnalyticsForVariable(
+        const aggregationList = [...this.config.aggregation[variable], ...extraAggregations];
```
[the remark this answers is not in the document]
kael89: We need to add them here so that `fetchOptions` are correctly adjusted in [AnalyticsRepository](https://github.com/beyondessential/tupaia/blob/1799-indicator-optimisation-fixes/packages/indicators/src/AnalyticsRepository.ts#L83). I believe this is related to [this question](https://github.com/beyondessential/tupaia/pull/2071#discussion_r577244511)

packages/indicators/src/AnalyticsRepository.ts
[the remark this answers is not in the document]
edmofro: I've played with an example and realise you're totally on top of it! This does pass through exactly the required analytics 🎉

kael89:
@edmofro merging this - all comments have been addressed, expect for improving `getAggregationListsForAnalyticsFetch` which will no longer be required

PR #2072, opened 2021-02-15
2240 hotfix survey submit

EMcQ-BES:
### Changes:

- When submitting a survey against a new entity, run the permissions check against the parent entity of the newly created entity


edmofro (approved):
lgtm

PR #2080, opened 2021-02-16
2247 update country codes for specific individual entities

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/2247


chris-bes (approved):
LGTM!

PR #2087, opened 2021-02-17
1799: Simplify indicator aggregation handling

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1799

This PR addresses all comments in 
https://github.com/beyondessential/tupaia/pull/2071#pullrequestreview-591625439. Hopefully this makes changes a bit easier to review, if not feel free to merge straight into the base branch and review there!

### DIFF
--- a/packages/indicators/src/getAggregationListsForAnalyticsFetch.ts
+++ b/packages/indicators/src/getAggregationListsForAnalyticsFetch.ts
@@ -3,12 +3,15 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { AggregationListsMap } from './types';
+import { Builder } from './Builder';
+import { AggregationList } from './types';
+
+type AggregationListsMap = Record<string, AggregationList[]>;
 
 /**
  * Merge aggregation list maps from `newEntries` in `targetMap`
  */
-export const addEntriesInAggregationListMap = (
+const addEntriesInAggregationListMap = (
   targetMap: AggregationListsMap,
   newEntries: Record<string, AggregationListsMap>,
 ): AggregationListsMap => {
@@ -37,4 +40,20 @@ export const addEntriesInAggregationListMap = (
 
   Object.entries(newEntries).forEach(addEntry);
   return newMap;
+};
+
+const getAggregationListMapsByBuilder = (builders: Builder[]) =>
+  Object.fromEntries(
+    builders.map(builder => [builder.getIndicator().code, builder.getAggregationListsByElement()]),
+  );
+
+export const getAggregationListsForAnalyticsFetch = (buildersByNestDepth: Builder[][]) => {
+  let aggregationListMap: AggregationListsMap = {};
+
+  buildersByNestDepth.forEach(builders => {
+    const listMapsByBuilder = getAggregationListMapsByBuilder(builders);
+    aggregationListMap = addEntriesInAggregationListMap(aggregationListMap, listMapsByBuilder);
+  });
+
+  return Object.values(aggregationListMap).flat();
 };
--- a/packages/indicators/src/AnalyticsRepository.ts
+++ b/packages/indicators/src/AnalyticsRepository.ts
@@ -11,7 +11,7 @@ import {
   Aggregator,
   keepAnalyticsMatchingFetchOptions,
 } from '@tupaia/aggregator';
-import { Aggregation, Analytic, FetchOptions } from './types';
+import { Aggregation, AggregationList, Analytic, FetchOptions } from './types';
 
 interface Fields {
   analyticsByElement: Record<string, Analytic[]>;
@@ -60,7 +60,7 @@ export class AnalyticsRepository {
    */
   async fetch(
     dataElementCodes: string[],
-    aggregationLists: Aggregation[][],
+    aggregationLists: AggregationList[],
     fetchOptions: FetchOptions,
   ) {
     const expandedOptions = adjustFetchOptionsToAggregationLists(fetchOptions, aggregationLists);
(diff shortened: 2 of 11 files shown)


kael89:
@edmofro the final step in this PR would be to clean, simplify and comment the beast called [getAggregationListsForAnalyticsFetch.ts](https://github.com/beyondessential/tupaia/blob/1799-simplify-aggregation-list-handling/packages/indicators/src/getAggregationListsForAnalyticsFetch.ts). But before I do so, here are my thoughts on this optimization attempt [Spoiler alert: I don't think it's worth it]. 

The only practical advantage at the moment would be just slightly faster PSSS load times, still not super fast. I have launched up the following instances so that you can have a look yourself:
* 
*

packages/indicators/src/getAggregationListsForAnalyticsFetch.ts
```diff
@@ -3,12 +3,15 @@
...
+import { AggregationList } from './types';
+
+type AggregationListsMap = Record<string, AggregationList[]>;
 
 /**
- * Merge aggregation list maps from `listMapEntries` in `targetMap`
+ * Merge aggregation list maps from `newEntries` in `targetMap`
  */
-export const addEntriesInAggregationListMap = (
+const addEntriesInAggregationListMap = (
```
edmofro, line 14:
```suggestion
const addEntriesInAggregationListsMap = (
```

edmofro:
@kael89 it is extremely noble that you would be willing to pull out some of your own work for the sake of the bigger picture. However, I am seeing quite a large performance difference between those two branches - both quite noticeable from the user perspective, and numerically significant:
image
image

The other consideration is that this optimisation should result in linear scaling of indicator nesting, and I can imagine that we will quickly want more and more deeply nested indicators.

I do hear your concerns that such complex code comes with a) the risk of logical holes, and b) significant maintenance overhead. For these reasons I agree that we should only pursue this further if we are convinced that the balance is right between these tradeoffs and the performance enhancement (which can't be found through simpler means/won't be made redundant by other upcoming performance enhancements).

So I'm interested in making that case...

Regarding logical holes:
- while it took you a long time to get the `AverageOf3QuarterlyIncreases` test passing, I imagine that flushed out a lot of cases that would come up across other indicators
- if we do keep this optimisation, we should make sure that any viz relying on an indicator has a thorough data audit during testing
- as you noticed, although there is more complex code now in place so more room for cracks in that complexity, actually by building a complex system you found and fixed several previously missed edge cases that weren't correctly handled pre-optimisation, so there may be some sort of reverse relationship whereby increased complexity leads us to take more care, test more thoroughly, and think through the logic more deeply 🤔 😄  (ok maybe this is a stretch)

Regarding maintenance 
- can you explain why the `count` indicator builder needs to work with events rather than analytics?

Regarding performance 
- it would be good to see whether your experience of relatively little performance benefit between branches is explained by something like caching which I haven't seen kick in (I have tried a couple of runs with similar results)
- if we can resolve that question, it would be good to investigate some more deeply nested (but still realistic) indicators and see if this has a better performance gain there

Keen for a call sometime today! Although I have found that the typing back and forth has (on my end) been conducive to more slow analysis, so I don't feel too bad that we didn't have a call yesterday

kael89:
@edmofro I did some tests on my local (used the same machine as for the original [benchmarks](https://github.com/beyondessential/tupaia/pull/2071#issuecomment-779499715). Results:

**Dev indicators:**

samoa_dev

**Optimised indicators:**

samoa_optimised

Both very fast and negligible performance difference!!

I also tried Fiji in dev, which I hadn't loaded in the optimised indicators, just to make sure that there is no obscure caching involved

edmofro:
Wow - blazingly fast! I had just been writing this up:

> We had a call on the above. The outcome is that @kael89 sees such negligible difference between the two branches on his local machine that we should really assess whether the added complexity is worth it, or if instead there is some difference between our AWS setup and his PC that we can learn from and apply instead, for even more performance improvement than this whole optimisation.
> 
> Steps forward from here:
> 1. @kael89 is benchmarking "open Samoa" on his local between the two branches, as it has a bigger discrepancy between the ...

So now 1. is complete, the next steps I was going to suggest were:
2. Identify what causes AWS to be so much slower. Does it just have much worse specs, or is it limiting DB connections, hitting CPU thresholds, hitting memory thresholds, or something else? 
3. See if there is still negligible difference with a more deeply nested indicator. I believe it's worth investigating this a bit further before we mothball this branch completely

That really is incredible how fast it is locally though!

kael89:
That all sounds good @edmofro and I agree on the next steps. However, I feel I do not have time to address 2 and 3 during this sprint. I am fully occupied with e2e tests and we have decided to priotise having an MVP for them complete by this sprint.

Perhaps @rohan-bes could help with 2 (although I am not sure about his availability), since he has made similar observations in the past? I would also appreciate if another dev takes care of 3. Indicators are used across the team now so it would be a good chance to spread the knowledge 🙂

edmofro:
Yes sure - agree that we should keep the e2e tests goal in mind now that this looks like a dead end. Thanks so much for your high degree of perspective on all of this!

kael89:
@edmofro merging this

kael89:
For the records, after a discussion with @edmofro we decided to revert analytic fetching to the current algorithm in `dev`. The main reason is that the single-fetch algorithm adds significant complexity, while it seems that with appropriate hardware setup the simpler algorithm can result in very satisfying performance.

We can still release all other changes related to this piece of work. Also, we should still do  **2** and **3** from [this comment](https://github.com/beyondessential/tupaia/pull/2087#issuecomment-780932853) at some stage

PR #2089, opened 2021-02-17
1674 Dashboard: Number of FETP Graduates by District, PNG

nick-beyondessential:
### Changes:
- Add FETP module databuilder countIndividualsByDistrict*
- migration for dashboardReports

* Struggled with trying to modify existing databuilders for required functionality. Counting individual entities which can have district or sub_district parent. If this is useful in future for other individual (or other entity types) could be generalised in future.


packages/database/src/migrations/20210216073210-AddFetpDashboardGradByDistrict-modifies-data.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+import { arrayToDbString, insertObject } from '../utilities/migration';
+
+var dbm;
+var type;
+var seed;
+
+const report = {
+  id: 'FETP_PG_graduate_by_district',
```
rohan-bes, line 10: Super nit picky, but don't we normally do dashboard ids in all caps? Or have the times changed?

rohan-bes (approved):
Man these databuilders that don't rely on data-elements always send me for a spin, I was so confused initially :P

Looks good to me 👍

packages/database/src/migrations/20210216073210-AddFetpDashboardGradByDistrict-modifies-data.js
nick-beyondessential: Not sure when that was a convention - but the overwhelming majority of ids seem to be mixed case. Tried to follw convention of project country code prefix.

PR #2096, opened 2021-02-18
Add permissions check to importOptionSets endpoint

EMcQ-BES:
### Changes:

- Add permissions check to importOptionsSets endpoint handler
- Move importOptionSets file under optionSets/ directory


chris-bes (approved):
LGTM!

edmofro:
As this is a hot fix I'm happy for you to merge it for testing on the release branch (a case of it already being too broken to care if we break it more)

PR #2099, opened 2021-02-18
[WIP] 2139: Perform aggregations during fetch for Tupaia Data API

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2139:
Also covers: https://github.com/beyondessential/tupaia-backlog/issues/2140 and https://github.com/beyondessential/tupaia-backlog/issues/2101

This PR has kind of become the whole epic for the above tickets. I may split it out into multiple PRs for the sake of convenience in the future...

The changes can be broken into 3 parts:
1. Include the git submodule, and relevant build/sql scripts to install pg-mv-fast-refresh in the tupaia database and build the analytics fast-refresh materialized view (referred to as the analytics table)
2. Changes to the tupaia data-api to be able to pull data from the analytics table, and to be able to perform aggregations on that data in the SQL requests based on the requested aggregation types
3. A couple of migrations to convert the Flu Tracker data to use Tupaia Data API, and editing the Flu Tracker dashboards to support the new data source

#### Part 1: The Analytics Table and pg-mv-fast-refresh
Currently when fetching data from the tupaia data api, we join a number of tables together to create the relevant sub-section of data required. This join can be a little costly for performance, but also makes more complex queries very tricky to perform. 

The goal of the analytics table was to have a single table where all the relevant data for fetching answers is housed. This allows for:
- Pre-aggregated columns for fast lookups of aggregate data
- Indexing of the table to further improve performance

However the analytics table needs to be able to be kept up to date with new answers in a fairly speedy fashion. In order to do this I made use pg-mv-fast-refresh. This library allows for creating tables from mutli-joined select statements, and will then can update that table whenever any inserts/deletes/updates are made to the base tables. The update doesn't happen automatically by default and should be called by a separate process.

More on pg-mv-fast-refresh can be found here:
https://github.com/CDLSoftware/pg-mv-fast-refresh
https://aws.amazon.com/blogs/database/building-fast-refresh-capability-in-amazon-rds-for-postgresql/

In order to setup pg-mv-fast-refresh and the analytics table from our current Tupaia setup, you need to perform the following:
```
- Create a password for the 'postgres' user
- Setup a .env file in @tupaia/data-api with the example.env fields filled in
- Run: yarn workspace @tupaia/data-api install-mv-refresh
- Run: yarn workspace @tupaia/data-api build-analytics-table
```

Once the table is built, it can be updated by running `SELECT mv$refreshMaterializedView('analytics', 'public', true)` against the database (will probably add a yarn task to do this soon).

#### Part 2: Aggregating data during the data-fetch
When doing performance testing of tupaia data-api, a cause of noticeable slowness was when we were pulling in a large number of records from the database. The SQL query itself ran in an acceptable speed, but it took node.js a long time to load in all those records. This caused individual requests to run slower, but also caused a general slowdown for the system.

A solution to this issue is for us to pull in fewer records, especially if the records we were pulling in were superfluous anyway. This often happens in the case where we intend to perform some aggregation that ignore a large amount of the data (eg. `MOST_RECENT`). The analytics table makes performing such aggregations at query time a lot easier, but tupaia data-api needs to have custom query logic based on the aggregations requested.

Given that the query building code is currently largely incomprehensible, here's an example of the query that returns just the most recent answers for AU state-level Flu Tracking data:
```
SELECT
  date AS "date",
  entity_code AS "entityCode",
  entity_name AS "entityName",
  data_element_code AS "dataElementCode",
  event_id AS "eventId",
  value AS "value",
  answer_type AS "type"
FROM (
  SELECT data_element_code, entity_code, entity_name
  FROM analytics
  INNER JOIN (
    VALUES ('FWV_003')
  INNER JOIN (
    VALUES ('AU_New South Wales'), ('AU_Northern Territory'), ('AU_Victoria'), ('AU_Queensland'), ('AU_South Australia'), ('AU_Tasmania'), ('AU_Australian Capital Territory'), ('AU_Western Australia')
  WHERE date > '2017-01-01T00:00:00.000Z'
  AND date < '2021-02-28T00:00:00.000Z'
  GROUP BY data_element_code, entity_code, entity_name
) as a1
CROSS JOIN LATERAL (
  SELECT date, event_id, value, answer_type
  FROM analytics
  WHERE entity_code = a1.entity_code
  AND data_element_code = a1.data_element_code
  order by date desc
  limit 1
) as a2
ORDER BY date;
``` 

The two `INNER JOIN ( VALUES ...)` blocks are a bit of a performance improvement on our previous `WHERE IN (...)` logic. It doesn't seem to make a huge difference but helps a bit for larger lists of entities/data-elements. See for more info: https://dba.stackexchange.com/questions/91247/optimizing-a-postgres-query-with-a-large-in

The `GROUP BY` and the `CROSS JOIN LATERAL` at the end is where the magic happens. The group by reduces the rows to unique per entity & data-element, and the cross join selects the first row in that group by descending date and adds its data.

With indexes in place this query performs very well, especially in the case where there are a few groups with a large number of entries. However, in the instance that there is a large number of groups with very few (1 namely) entries, the performance is worse than just a raw fetch. An example that highlights this can be seen in the performance testing rubric here: https://docs.google.com/spreadsheets/d/1KoewibyIxEJjpcjataeKi5svspPi41gbcG0LUtwIHZk/edit?ts=602d9d73&pli=1#gid=982594139&range=P7

Another option is to use `SELECT DISTINCT ON (data_element_code, entity_code)...` instead of the `CROSS JOIN`, which performs worse in the few groups many entries scenario, but outperforms in the large number of groups with 1 entry scenario (once the number of groups > ~5k-10k).

I think there's more room of design and customization here to come with the best solution...

#### Part 3: Migrating Flu Tracker data to Tupaia Data API
Not much to say here really, but this is what's being done by: 
- `packages/database/src/migrations/20201214060202-DeleteCovidauDhisSyncData-modifies-data.js`
- `packages/database/src/migrations/20201213231508-UseTupaiaAsServiceCovidauFluTracker-modifies-data.js`
- `packages/database/src/migrations/20201214034339-UpdateCovidReportsForTupaiaDataSource-modifies-data.js`


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "packages/data-api/scripts/pg-mv-fast-refresh"]
+	path = packages/data-api/scripts/pg-mv-fast-refresh
+	url = git@github.com:beyondessential/pg-mv-fast-refresh.git
```
rohan-bes, line 3: In case you've never used git submodules before (as I had) then what this allows is for git to keep a record of another repo which it will fetch to that directory if requested. This is pretty ideal for us in this situation as we only need to pull down that repo to install the pg-mv-refresh functions, so most of the time we don't actually need it in our project 👍

packages/data-api/src/SqlQuery.js
```diff
@@ -6,22 +6,51 @@
...
     this.parameters = baseParameters;
     this.orderByClause = null;
+    this.hasWhereClause = false;
+  }
+
+  loggableQuery() {
+    const replacementIterator = this.parameters
+      .map(param => param.replace(/'/g, "''"))
+      [Symbol.iterator]();
+    return this.query.replace(/\?/g, () => `'${replacementIterator.next().value}'`);
```
rohan-bes, line 22: This may be one of the more complex pieces of javascript I've ever written... purely for the purposes of logging 😂

edmofro:
Regarding the inner join on values part, did you try keeping the `where in` and just wrapping the rest of it in a `values`? @EMcQ-BES suggested that at one point during our earlier investigations, and it made a bit of a difference - just wondering if that's down to the `values` or the `join`, because having it in the `where` clause makes one less weird bit to think about.

packages/data-api/src/fetchData.js
```diff
@@ -6,63 +6,159 @@ import { utcMoment } from '@tupaia/utils';
 
 import { SqlQuery } from './SqlQuery';
 
-const generateBaseSqlQuery = ({ dataElementCodes, organisationUnitCodes, startDate, endDate }) => {
+const SUPPORTED_AGGREGATIONS = [
```
rohan-bes, line 9: Really not my cleanest code throughout this file 😅 will clean it up soon I promise

rohan-bes:
Closing this PR as it has been broke up into:
https://github.com/beyondessential/tupaia/pull/2118
https://github.com/beyondessential/tupaia/pull/2122
https://github.com/beyondessential/tupaia/pull/1826

PR #2107, opened 2021-02-19
2295 upgrade packages to allow node v14

IgorNadj:
### Changes:

- `pg`
  - Upgrade from `7.10` to `8.5`
  - Changelog: https://github.com/brianc/node-postgres/blob/master/CHANGELOG.md
  - Breaking changes that may affect us:
    - **pg@8.0.0** Change default behavior when not specifying rejectUnauthorized with the SSL connection parameters
      - See first item in https://node-postgres.com/announcements#2020-02-25
      - Local development seems to be ok, since we connect without SSL. I don't think anyone uses this (https://beyondessential.slack.com/archives/CDGS6P56K/p1614220174020900)
      - dev / feature branch deployments seem to be ok, since we connect without SSL. Tested on https://2295-upgrade-node-to-v14.tupaia.org/ and no issues with tupaia, no errors connecting to db in logs from any services.
      - Production seems to run a local database as well, so it should work the same. I can't see any DB_ENABLE_SSL or similar in the production .env configs, so it should behave the same. Not sure how to test this other than release to production. **TODO: is this right?** 
    - **pg-query-stream@3.0.0** Changed stream.close to stream.destroy
      - It doesn't look like we use this

- `pg-pubsub`
  - Upgrade from `0.3.0` to `0.6.1`
  - Changelog: https://github.com/voxpelli/node-pg-pubsub/blob/main/CHANGELOG.md
  - Breaking changes that may affect us:
    - **0.6.0 (2021-02-16)** - Breaking change: .addChannel() now returns a Promise rather than the PGPubsub instance
      - Used in 3 places, but we never use the return value, so no change is needed. Tested change handlers, they are triggered and caught as previously.


package.json
```diff
@@ -101,6 +101,6 @@
     "yargs": "^16.0.3"
   },
   "engines": {
-    "node": "^12.18.3"
+    "node": "^14.15.4"
```
edmofro, line 104: Is it possible to have two supported options here for a while?

packages/database/package.json
```diff
@@ -34,8 +34,8 @@
     "lodash.keyby": "^4.6.0",
     "moment": "^2.24.0",
     "os": "0.1.1",
-    "pg": "7.10.0",
-    "pg-pubsub": "0.3.0",
+    "pg": "8.5.1",
+    "pg-pubsub": "0.6.1",
```
edmofro, line 38: Happy with this :+1:

It's the kind of change we'd want to merge soon after a release so that it gets a full week of people using it in dev, just in case there's any regression, because it's so central (particularly `pg-pubsub` - I haven't looked through those release notes and they're below v1 so could have breaking changes)

package.json
edmofro: Doesn't look like it, but also looks like it should be ok to leave it on 12.18 but run locally (may get warnings, but we don't have `engine-strict` turned on

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines

edmofro: Oh, I was wrong - we do have `engine-strict` turned on! I guess that's why `yarn` fails on v14

image

IgorNadj: Updated to have required node version be 12 or 14

packages/database/package.json
IgorNadj: I have put breaking changes and testing notes in the description

IgorNadj:
@edmofro this is ready for re-review, the only thing I haven't tested is production, see description.

edmofro (approved):
Awesome, thanks for the thorough breaking changes investigation!

edmofro:
That's fine, there's no difference between production and dev in terms of db setup

PR #2114, opened 2021-02-22
[WIP] 1378: Implement 'count' indicator builder

chris-bes:
### Changes:

- Implemented countEvent indicator
- Renamed arithmetic indicators to arithmeticAnalytic

### Upcoming:

- Unit tests (after the overall design is approved)
- Renaming Arithmetic classes/folders/variables to new names (after we've finalised the name here: https://github.com/beyondessential/tupaia/pull/2114#discussion_r579966051)


packages/database/src/migrations/20210211051517-RenameArithmeticIndicators-modifies-data.js
```diff
@@ -0,0 +1,35 @@
...
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`
+    UPDATE indicator
+    SET builder = 'arithmeticAnalytic'
```
chris-bes, line 20: Now that indicator can fetch events or analytics as inputs, we want to make this more obvious in the indicator names. So we want to change `arithmetic` to something more specific. I'm not sure if this is a good name. I'm thinking between 3 options:
- arithmeticAnalytic
- analyticArithmetic
- runAnalyticArithmetic

I went with `arithmeticAnalytic` and `countEvent` to be more consitent with how we name data builders... But any suggestions would be appreciated!

packages/indicators/src/Builder/CountEventBuilder/config/types.ts
```diff
@@ -0,0 +1,19 @@
...
+export type DefaultValue = number | string | 'undefined';
+
+/**
+ * A flexible config schema that can be provided as an input to the indicator
+ */
+export type CountEventConfig = {
+  readonly formula: string;
+  readonly programCode: string;
+  readonly defaultValues?: Record<string, DefaultValue>;
+};
```
chris-bes, line 19: This is the spec of `countEvent` config. We don't have `aggregation` like `arithmetic` indicator because we usually don't aggregate event data. I've also left out `paramBuilders` because:

1. We don't need it at the moment
2. I feel that it's better to wait until there's specific use case for it so that we can come up with a more concise config format?

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -22,6 +22,14 @@ const average = (...argumentList) => {
   return sum / existingValues.length;
 };
 
+/**
+ * List of built in functions in math.js
+ * This list is here because when extracting variables for an expression by traversing the node tree,
+ * the built in functions are also considered as Symbol Node (which is similar to variables).
+ * So we want to exclude them when getting the variables.
+ */
+const BUILT_IN_FUNCTIONS = ['equalText'];
```
chris-bes, line 31: This will be used in a later PR

chris-bes:
Migrated all to this PR: https://github.com/beyondessential/tupaia/pull/2125

PR #2118, opened 2021-02-22
2101: Analytics Table Materialized View

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2101:

Currently when fetching data from the tupaia data api, we join a number of tables together to create the relevant sub-section of data required. This join can be a little costly for performance, but also makes more complex queries very tricky to perform. 

The goal of the analytics table was to have a single table where all the relevant data for fetching answers is housed. This allows for:
- Pre-aggregated columns for fast lookups of aggregate data
- Indexing of the table to further improve performance

However the analytics table needs to be able to be kept up to date with new answers in a fairly speedy fashion. In order to do this I made use pg-mv-fast-refresh. This library allows for creating tables from mutli-joined select statements, and will then can update that table whenever any inserts/deletes/updates are made to the base tables. The update process is handled by the meditrak-server via the `AnalyticsRefresher` class, which listens to changes to the answers table and triggers a fast refresh.

More on pg-mv-fast-refresh can be found here:
https://github.com/CDLSoftware/pg-mv-fast-refresh
https://aws.amazon.com/blogs/database/building-fast-refresh-capability-in-amazon-rds-for-postgresql/

In order to setup pg-mv-fast-refresh and the analytics table from our current Tupaia setup, you need to perform the following:
```
- Setup a .env file in @tupaia/data-api with the example.env fields filled in (see lastpass for details)
- Run: yarn workspace @tupaia/data-api install-mv-refresh
- Run: yarn workspace @tupaia/data-api build-analytics-table
```
^the above steps ought to be run automatically on deployment for aws instances, but may need to be run manually if setting up on a local branch.

Once the table is built, it can be updated by running `SELECT mv$refreshMaterializedView('analytics', 'public', true)` against the database (will probably add a yarn task to do this soon).


packages/database/src/analytics/AnalyticsRefresher.js
```diff
@@ -0,0 +1,51 @@
+/**
```
rohan-bes, line 1: Unit tests incoming for this class

packages/database/src/analytics/AnalyticsRefresher.js
```diff
@@ -0,0 +1,51 @@
...
+    // remove timeout so any jobs added now get scheduled anew
+    this.scheduledRefreshTimeout = null;
+
+    // get the subtrees to delete, then run the delete
+    const start = Date.now();
+    await this.database.executeSql(
+      `SELECT mv$refreshMaterializedView('analytics', 'public', true);`,
+    );
+    const end = Date.now();
+    console.log(`Refreshed analytics table, took: ${end - start}ms`);
```
rohan-bes, line 49: Considering leaving this console log in for tracking purposes? It may get rather noisy tho so leaning towards removing it

packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -55,6 +55,11 @@ for PACKAGE in ${PACKAGES[@]}; do
         # sync queues
         echo "Migrating the database"
         yarn migrate
+
+        # After running migrations it's good to ensure that the analytics table is fully built
+        echo "Building analytics table"
+        yarn workspace @tupaia/data-api install-mv-refresh
+        yarn workspace @tupaia/data-api build-analytics-table
```
rohan-bes, line 62: This only builds the analytics table if it doesn't already exist, otherwise it skips it

chris-bes (commented):
Looks great @rohan-bes , took me a while to set this up locally but worth it!

I also see that you're still pushing commits to this branch so should I wait a bit more?

packages/data-api/scripts/createAnalyticsMaterializedView.sql
```diff
@@ -0,0 +1,76 @@
...
+  iNewLogTableCount INTEGER := 0;
+  tStartTime TIMESTAMP;
+  source_table TEXT;
+  source_tables_array TEXT[] := array['answer', 'survey_response', 'entity', 'survey', 'question', 'data_source'];
+  pSqlStatement TEXT := '
+    SELECT
+      entity.code as entity_code,
+      entity.name as entity_name,
+      question.code as data_element_code,
+      survey.code as survey_code,
```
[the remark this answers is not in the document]
rohan-bes: Ah! Great point, will do 😄

packages/database/src/analytics/AnalyticsRefresher.js
```diff
@@ -0,0 +1,60 @@
...
+    this.database = database;
+    this.models = models;
+    this.scheduledRefreshTimeout = null;
+    this.scheduledRefreshPromise = null;
+    this.scheduledRefreshPromiseResolve = null;
+    this.changeHandlerCancellers = [];
+  }
+
+  listenForChanges() {
+    this.changeHandlerCancellers = [this.models.answer.addChangeHandler(this.handleAnswerChange)];
```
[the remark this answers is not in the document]
rohan-bes: Not a silly question at all, very relevant! I basically didn't wanna add too many listeners to avoid adding too much of a performance overhead. My logic is:
- If entities/questions/surveys/etc. are added, we don't need to update analytics until new answers are actually added to the system
- It's pretty rare for us to edit existing data_element_codes/entity_codes/etc. if so we can manually run a refresh? You should be able to do this via a migration

Valid point tho, what are your thoughts on my reasoning?

packages/data-api/scripts/createAnalyticsMaterializedView.sql
```diff
@@ -0,0 +1,76 @@
...
+      survey_response.id as event_id,
+      date_trunc(''day'', survey_response.end_time) as "day_period",
+      date_trunc(''week'', survey_response.end_time) as "week_period",
+      date_trunc(''month'', survey_response.end_time) as "month_period",
+      date_trunc(''year'', survey_response.end_time) as "year_period",
+      answer.text as value,
+      answer.type as answer_type,
+      survey_response.end_time as "date"
```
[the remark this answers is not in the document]
rohan-bes: Ah! You're right, great catch! I was playing around with this code a bit recently and I think I made a bug hehe.

I might hold off on this field for now as I think @EMcQ-BES is adding a `data_time` field, which is timezone adjusted and should be the one we eventually end up using?

rohan-bes:
> Looks great @rohan-bes , took me a while to set this up locally but worth it!
> 
> I also see that you're still pushing commits to this branch so should I wait a bit more?

I'm just pushing changes to try and get the tests working, the core code should stay the same more or less. But happy for you to hold off til the tests are green if you like 👍

packages/database/src/analytics/AnalyticsRefresher.js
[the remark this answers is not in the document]
chris-bes: I have to agree!

packages/data-api/scripts/createAnalyticsMaterializedView.sql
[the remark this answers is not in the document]
EMcQ-BES: Correct, that is being worked on [here](https://github.com/beyondessential/tupaia-backlog/issues/1149)

chris-bes:
Looks good! I still have 1 minor suggestion here https://github.com/beyondessential/tupaia/pull/2118/files#r584397794. But since it's minor feel free to do it whenever! Pre-approve

chris-bes (approved):

package.json
```diff
@@ -31,7 +31,7 @@
     "refresh-database": "node scripts/node/refreshDatabase.js",
     "test-all": "node ./scripts/node/testAll --silent",
     "update-codeship-env-vars": "./packages/devops/scripts/ci/updateCiEnvVars.sh",
-    "update-parameter-store-env-vars": "./scripts/bash/updateParameterStoreEnvVars.sh",
+    "download-parameter-store-env-vars": "./scripts/bash/syncDotEnvWithParameterStore.sh --download",
```
rohan-bes, line 34: Renamed/re-worked this script as it wasn't being used anywhere

packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -21,13 +21,9 @@ PACKAGES=("meditrak-server" "web-config-server" "psss-server" "report-server" "w
 # For each package, get the latest and deploy it
 for PACKAGE in ${PACKAGES[@]}; do
     # Set up .env to match the environment variables stored in SSM parameter store
+    yarn download-parameter-store-env-vars --package-name $PACKAGE --environment $ENVIRONMENT
+
     cd ${HOME_DIRECTORY}/packages/$PACKAGE
-    rm .env
-    echo "Checking out ${ENVIRONMENT} environment variables for ${PACKAGE}"
```
chris-bes, line 26: Perhaps we should still echo this line which can be helpful? Unless it's also done in the other script

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "packages/data-api/scripts/pg-mv-fast-refresh"]
 	path = packages/data-api/scripts/pg-mv-fast-refresh
-	url = git@github.com:beyondessential/pg-mv-fast-refresh.git
+	url = ../pg-mv-fast-refresh.git
```
chris-bes, line 3: Nice!

packages/devops/scripts/ci/updateCiEnvVars.sh
```diff
@@ -1,6 +1,41 @@
...
-for PACKAGE in "meditrak-server" "admin-panel" "web-frontend" "web-config-server" "devops" "database"; do
+rm -f old.env
+echo "Checking out previous ci-env-vars"
+git reset -q HEAD ci-env-vars.encrypted
+git checkout $COMPARE_BRANCH -- ci-env-vars.encrypted
+git reset -q HEAD ci-env-vars.encrypted
+echo "Decryting old ci-env-vars"
+jet decrypt ci-env-vars.encrypted old.env
+echo "Creating new ci-env-vars"
+for PACKAGE in "meditrak-server" "admin-panel" "web-frontend" "web-config-server" "devops" "data-api" "database"; do
```
chris-bes, line 30: Silly question but any reasons that we only want to update CI env vars for these packages?

chris-bes (approved):
Nice! Looks good to me. Up to you if you want to add that echo or not.

packages/devops/scripts/deployment/deployPackages.sh
rohan-bes: Yeah the other script does an echo:
`echo "Checking out environment variables under the prefix ${PARAMETER_STORE_PATH_PREFIX}"`

perhaps slightly less readable but I think it's fine 😄

.gitmodules
rohan-bes: This should fix the htts/ssh issue 😄

packages/devops/scripts/ci/updateCiEnvVars.sh
rohan-bes: So these are the only packages that require an `.env` file to run their tests (except for devops which I think uses an env var as part of the codeship script itself).

I suppose we could change it to use all packages (and update the script to handle when that .env file isn't present) however then we have a little bit more complexity ensuring there are no name clashes between the .env files (something that we should avoid anyway in fairness)

PR #2119, opened 2021-02-22
2318: Event push should wait for entity to be fully set up

edmofro:
### Issue #:
- https://github.com/beyondessential/tupaia-backlog/issues/2318

The event push has been quickly failing until it gets to priority 5, because the new case entity created as part of it hasn't been fully set up when the push is attempted. This PR changes it to wait "patiently" for two things to happen
- The tracked entity instance to be created on DHIS2, and its id recorded against the entity
- The entity relation table to be rebuilt with the new case entity inside it


packages/database/src/modelClasses/Entity.js
```diff
@@ -182,17 +191,21 @@ export class EntityType extends DatabaseType {
...
+            [QUERY_CONJUNCTIONS.OR]: {
+              descendant_id: this.id,
+            },
+          },
+          {
+            joinWith: TYPES.ANCESTOR_DESCENDANT_RELATION,
+            sort: ['entity_hierarchy.name ASC'],
+          },
+        ),
+      v => v.length > 0,
```
rohan-bes, line 208: I'm sure you would've considered this, but is it possible for us to break this out the same way that has been done with `getDhisId()`/`getDhisIdPatiently()`? Perhaps something like:
```
async fetchDefaultEntityHierarchyId() {
  ... // As previous
}

async fetchDefaultEntityHierarchyIdPatiently() {
  ... // New fetchPatientlyLogic
}
```
and then in `packages/meditrak-server/src/dhis/pushers/data/event/EventBuilder.js`
```
    // Need to wait patiently here to avoid events failing to sync due to entities not yet populating in DHIS
    const patientlyAwaitedHierarchyId = await entity.fetchDefaultEntityHierarchyIdPatiently(); 
    const { code: orgUnitCode } = await entity.fetchNearestOrgUnitAncestor(patientlyAwaitedHierarchyId);
```

I guess I just feel that this patient wait behaviour is kind of unusual, and given that it in 'failure scenario' (ie. requesting orgUnitAncestor for an orgUnit that isn't in a hierarchy) will wait 10 seconds before erroring out, I'd prefer we explicitly call it in the places we need? I know this failure scenario oughtn't occur, but still...

What are your thoughts on this? Am I being paranoid?

packages/utils/src/fetchPatiently.js
```diff
@@ -0,0 +1,25 @@
...
+  validReturnTest = v => !!v,
+  retries = DEFAULT_RETRIES,
+  retryPeriodMs = DEFAULT_RETRY_PERIOD_MS,
+) => {
+  const value = await fetchFn();
+  if (validReturnTest(value) || retries === 0) {
+    return value;
+  }
+  await sleep(retryPeriodMs);
+  return fetchPatiently(fetchFn, validReturnTest, retries - 1, retryPeriodMs);
```
rohan-bes, line 24: Recursion will never not be pretty to me 😍

rohan-bes (changes requested):
Requesting a change, but happy to get pushback on it if you feel it's unnecessary or will lead to too much code complexity

packages/database/src/modelClasses/Entity.js
edmofro: No I don't think that's paranoid - it's a solid suggestion. 

My only pushback would be that other places `fetchNearestOrgUnitAncestor` is used also have the potential to be safer if they wait for the entity hierarchy to be fully built, e.g. `TrackedEntityPusher` could be called soon after the entity is added to the database. I can have a look through though and try to identify where it might be really irrelevant, and if there are enough places I'll follow your suggestion :-)

edmofro: I've looked through, and all but one are basically things that respond to changes/handle sync to DHIS2, so I think it's appropriate for all of them to use the patient behaviour.

The only one that's a change responder is `feedScraper.getSurveyResponseFacilityData`, which uses it indirectly via `SurveyResponse.fetchOrganisationUnit()`. That's a periodic scraper, so less concerned with changes that might have just happened - but still it isn't a time sensitive operation, so there's no real harm in having it use the patient behaviour as well.

For reference, the places I found it to be used were
```
- DhisChangeDetailGenerator.generateSurveyResponseDetails - appropriate
- EventBuilder.enrollTrackedEntityAndGetTrackerEventFields and EventBuilder.fetchDefaultEventOrgUnit - appropriate
- TrackedEntityPusher.buildRecord - appropriate
- AggregateDataPusher.fetchOrganisationUnit - appropriate
- feedScraper.getSurveyResponseFacilityData - less appropriate but not time sensitive
```

What do you think? Happy to leave as is rather than adding the separate fetch in each of those places? I don't mind either way, I do think there's something nice about completely pulling the defaulting behaviour out of that function, which this would allow.

rohan-bes: Yeah if it's appropriate in all places then I'm not too fussed... Perhaps just renaming to `fetchDefaultEntityHierarchyIdPatiently()` as a compromise? I'm happy to let you make the call either way 😄

rohan-bes (approved):

packages/database/src/modelClasses/Entity.js
edmofro: Done :+1:

PR #2121, opened 2021-02-22
893 project bounds

tcaiger:
### Changes:

- fix bounds calculation
- use permission based bounds for project on the front end

---


packages/web-config-server/src/utils/geoJson.js
```diff
@@ -51,16 +51,10 @@ export const calculateBoundsFromEntities = entities => {
     return null;
   }
 
+  const points = bounds.reduce((values, bound) => [...values, ...bound.map(point => point)], []);
```
edmofro, line 54:
```suggestion
  const points = bounds.reduce((values, bound) => [...values, ...bound], []);
```

packages/web-frontend/src/sagas.js
```diff
@@ -594,7 +594,22 @@ function* fetchOrgUnitData(organisationUnitCode, projectCode) {
...
     };
     const requestResourceUrl = `organisationUnit?${queryString.stringify(urlParameters)}`;
-    const orgUnitData = yield call(request, requestResourceUrl);
+
+    let orgUnitData = yield call(request, requestResourceUrl);
+
+    const state = yield select();
+
+    // If it's a project, use the map bounds from the store that we already have as they are more precise
+    if (orgUnitData.type === 'Project' && state.map.position.bounds) {
```
edmofro, line 603: This feels pretty fragile to me. We're relying on the fact that the `state.map.position.bounds` has been set by the project, which assumes a specific order of events. 

If it were pulling from somewhere that wasn't writeable by things other than the project (e.g. `state.something.projectx.bounds`) I'd be more keen on it.

Would still prefer it came from the backend endpoint handler though - was that a no go?

edmofro (changes requested):
The geoJson bit is much simpler with the first map of points, nice one!

tcaiger:
> The geoJson bit is much simpler with the first map of points, nice one!

Yes and as well as being clearer, it changes the calculation to make sure we are getting the min and max values across all points. The previous calculation was assuming an order of points as it's input and sometimes miscalculating.

packages/web-frontend/src/sagas.js
tcaiger: Yes agreed. Have moved the logic to the endpoint handler

packages/web-config-server/src/apiV1/organisationUnit.js
```diff
@@ -12,12 +12,46 @@ const translateDescendantForFrontEnd = (descendant, childIdToParentId, entityIdT
   parent: entityIdToCode[childIdToParentId[descendant.id]],
 });
 
+async function fetchEntitiesWithProjectAccess(req, entities, userGroups) {
```
edmofro, line 15: Is this exactly the same as the function in the `project` endpoint handler? If so, can we make it a util?

packages/web-config-server/src/apiV1/organisationUnit.js
```diff
@@ -12,12 +12,46 @@ const translateDescendantForFrontEnd = (descendant, childIdToParentId, entityIdT
...
+      const accessByEntity = await Promise.all(
+        orgUnits.map(async orgUnit => {
+          const { id, name, code, bounds } = orgUnit;
+          return {
+            id,
+            name,
+            code,
+            bounds,
+            hasAccess: await this.checkUserHasEntityAccess(orgUnit),
+          };
```
edmofro, line 47: This is not a full entity, but it's also a lot more than we need (as far as I can tell). Could we go one way or the other? I'd opt for the minimal version, in which case the result would be more like `boundsOfEntitiesWithAccess`.

packages/web-config-server/src/apiV1/organisationUnit.js
```diff
@@ -12,12 +12,46 @@ const translateDescendantForFrontEnd = (descendant, childIdToParentId, entityIdT
...
 
   async buildResponse() {
     const { includeCountryData } = this.query;
     const project = await this.fetchProject();
+
+    // If this is a project entity, get entitiesWithAccess and set them on the entity
+    if (this.entity.isProject()) {
+      const orgUnits = await this.entity.getNearestOrgUnitDescendants();
+
+      const accessByEntity = await Promise.all(
```
edmofro, line 38: The `xByY` naming pattern is usually used for objects where `y` is the key and `x` is the value. This would be better named something like `entitiesWithAccessInfo` as it's an array.

This probably comes from the `project` endpoint, so I realise you aren't introducing the specific name here!

edmofro (changes requested):
Are we able to also remove the `bounds` from the `project` endpoint handler, given that we should be able to use the project's organisation unit on the front end instead? I totally get that you're eager to move on and back to LESMIS, but I'd rather not leave tech debt if this is the opportune moment to address it - the two pathways to setting bounds seem unnecessary and likely to cause future bugs.

edmofro:
Thanks though for putting the effort into moving the logic to the back end! I think we're on our way to a bit more resilience around bound handling.

packages/web-config-server/src/models/Entity.js
```diff
@@ -40,7 +41,11 @@ class EntityType extends CommonEntityType {
     return {
       type,
       point: translatePointForFrontend(point),
-      bounds: translateBoundsForFrontend(bounds),
+
+      // When possible return the bounds for an entity based on the enetities the user has access to
```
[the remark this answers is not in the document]
edmofro: Nit: this is marked resolved but hasn't been corrected

packages/web-config-server/src/apiV1/organisationUnit.js
```diff
@@ -18,6 +18,21 @@ export default class extends RouteHandler {
...
     const { includeCountryData } = this.query;
     const project = await this.fetchProject();
+
+    // If this is a project entity, get entitiesWithAccess and set them on the entity
+    if (this.entity.isProject()) {
+      const orgUnits = await this.entity.getNearestOrgUnitDescendants();
+
+      const entitiesWithAccessInfo = await Promise.all(
+        orgUnits.map(async orgUnit => ({
+          ...orgUnit,
```
edmofro, line 28: Did you consider that option of just storing the bounds, rather than full entity details? All good if you did and prefer the whole entity

edmofro (approved):
Stoked with this outcome! A couple of earlier suggestions I'm just following up on, but approved either way :-)

Thanks for putting in the mahi

packages/web-config-server/src/apiV1/organisationUnit.js
tcaiger: Yeah I did look at it but I decided I like this better as it made sense to me that way

edmofro: Sweet - just that typo to go!

edmofro (approved):

PR #2122, opened 2021-02-22
2140: Fetch internal data from Analytics table and support aggregation during data fetch

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2140:

When doing performance testing of tupaia data-api, a cause of noticeable slowness was when we were pulling in a large number of records from the database. The SQL query itself ran in an acceptable speed, but it took node.js a long time to load in all those records. This caused individual requests to run slower, but also caused a general slowdown for the system.

A solution to this issue is for us to pull in fewer records, especially if the records we were pulling in were superfluous anyway. This often happens in the case where we intend to perform some aggregation that ignore a large amount of the data (eg. `MOST_RECENT`). The analytics table makes performing such aggregations at query time a lot easier, but tupaia data-api needs to have custom query logic based on the aggregations requested.

Given that the query building code is currently largely incomprehensible, here's an example of the query that returns just the most recent answers for AU state-level Flu Tracking data:
```
SELECT
  date AS "date",
  entity_code AS "entityCode",
  entity_name AS "entityName",
  data_element_code AS "dataElementCode",
  event_id AS "eventId",
  value AS "value",
  answer_type AS "type"
FROM (
  SELECT data_element_code, entity_code, entity_name
  FROM analytics
  INNER JOIN (
    VALUES ('FWV_003')
  INNER JOIN (
    VALUES ('AU_New South Wales'), ('AU_Northern Territory'), ('AU_Victoria'), ('AU_Queensland'), ('AU_South Australia'), ('AU_Tasmania'), ('AU_Australian Capital Territory'), ('AU_Western Australia')
  WHERE date > '2017-01-01T00:00:00.000Z'
  AND date < '2021-02-28T00:00:00.000Z'
  GROUP BY data_element_code, entity_code, entity_name
) as a1
CROSS JOIN LATERAL (
  SELECT date, event_id, value, answer_type
  FROM analytics
  WHERE entity_code = a1.entity_code
  AND data_element_code = a1.data_element_code
  order by date desc
  limit 1
) as a2
ORDER BY date;
``` 

The two `INNER JOIN ( VALUES ...)` blocks are a bit of a performance improvement on our previous `WHERE IN (...)` logic. It doesn't seem to make a huge difference but helps a bit for larger lists of entities/data-elements. See for more info: https://dba.stackexchange.com/questions/91247/optimizing-a-postgres-query-with-a-large-in

The `GROUP BY` and the `CROSS JOIN LATERAL` at the end is where the magic happens. The group by reduces the rows to unique per entity & data-element, and the cross join selects the first row in that group by descending date and adds its data.

With indexes in place this query performs very well, especially in the case where there are a few groups with a large number of entries. However, in the instance that there is a large number of groups with very few (1 namely) entries, the performance is worse than just a raw fetch. An example that highlights this can be seen in the performance testing rubric here: https://docs.google.com/spreadsheets/d/1KoewibyIxEJjpcjataeKi5svspPi41gbcG0LUtwIHZk/edit?ts=602d9d73&pli=1#gid=982594139&range=P7

Another option is to use `SELECT DISTINCT ON (data_element_code, entity_code)...` instead of the `CROSS JOIN`, which performs worse in the few groups many entries scenario, but outperforms in the large number of groups with 1 entry scenario (once the number of groups > ~5k-10k).

I think there's more room of design and customization here to come with the best solution in the future, but for now the performance seems decent across the board.


packages/data-api/src/fetchData.js
```diff
@@ -6,63 +6,159 @@ import { utcMoment } from '@tupaia/utils';
...
+    SELECT 
+      date AS "date",
+      entity_code AS "entityCode",
+      entity_name AS "entityName",
+      data_element_code AS "dataElementCode",
+      event_id AS "eventId",
+      value AS "value",
+      answer_type AS "type"
+    FROM (
+      ${getA1Select(firstAggregationType)}
```
chris-bes, line 136: I think `getA1Select` returns different fields based on the `firstAggregationType`, for some aggregations it only returns the COMMON_FIELDS which are `data_element_code, entity_code, entity_name`. So it might not return all the columns asked in the outer select part above (listed below). Would that throw an error or something?

```
SELECT 
date AS "date",
entity_code AS "entityCode",
entity_name AS "entityName",
data_element_code AS "dataElementCode",
event_id AS "eventId",
value AS "value",
answer_type AS "type"
```

chris-bes (commented):
Maybe we can give

packages/data-api/src/fetchData.js
```diff
@@ -6,63 +6,159 @@ import { utcMoment } from '@tupaia/utils';
 
 import { SqlQuery } from './SqlQuery';
 
-const generateBaseSqlQuery = ({ dataElementCodes, organisationUnitCodes, startDate, endDate }) => {
+const SUPPORTED_AGGREGATIONS = [
+  'FINAL_EACH_DAY',
+  'FINAL_EACH_WEEK',
+  'FINAL_EACH_MONTH',
+  'FINAL_EACH_YEAR',
+  'MOST_RECENT',
```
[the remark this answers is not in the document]
rohan-bes: Ahh yes... I must admit that currently this file is not in a very pretty shape. I kinda rushed out getting the functionality working and didn't put much time into writing clean code. I'll take a look at it tomorrow and push up a refactor 😄

packages/data-api/src/fetchData.js
rohan-bes: So, the rest of those fields are returned by the `a2 select`, which should be joined if the relevant aggregation is being used. If there are fields missing though, you are right an SQL error would be thrown I believe.

chris-bes (approved):

packages/data-api/src/fetchData.js
```diff
@@ -121,7 +106,7 @@ const generateBaseSqlQuery = ({
   endDate,
   aggregations,
 }) => {
-  const firstAggregationType = aggregations && aggregations[0] && aggregations[0].type;
+  const firstAggregation = AGGREGATIONS[aggregations?.[0]?.type] || AGGREGATIONS.DEFAULT;
```
chris-bes, line 123: Just checking, I remember in some visualisations, we do want to perform the entityAggregation first before the input aggregations (by having a config `entityAggregationOrder: 'BEFORE' || 'AFTER'`) Will that still work in this case? Coz I think this is assuming that the input aggregation is the first one.

Maybe the worst case is just it will just do the input aggregations not on the fetchData level?

packages/data-api/src/fetchData.js
```diff
@@ -49,63 +73,24 @@ const getA1WhereClause = conditions => {
...
-    case 'FINAL_EACH_WEEK':
-      return 'AND week_period = a1.week_period';
-    case 'FINAL_EACH_MONTH':
-      return 'AND month_period = a1.month_period';
-    case 'FINAL_EACH_YEAR':
-      return 'AND year_period = a1.year_period';
-    default:
-      return '';
-  }
+const getA1GroupByClause = firstAggregation => {
```
chris-bes, line 77: Should we name this `aggregation` or `firstAggregation` in this method?

packages/data-api/src/__tests__/TupaiaDataApi.test.js
```diff
@@ -13,6 +13,7 @@ describe('TupaiaDataApi', () => {
   const models = getTestModels();
 
   beforeAll(async () => {
+    jest.setTimeout(30000); // Extend default jest timeout as these tests can take a while
```
chris-bes, line 16: I did run into the same problem when running the indicators test locally and had to do this as well (maybe because my machine is slow). Maybe we should move this to somewhere more global?

chris-bes (approved):
Looks nice! My main question is about the order of the aggregations. But I have a feeling that it's going to be okay so pre approve

packages/data-api/src/fetchData.js
rohan-bes: No you're right, we won't support doing aggregation in the fetchData level for that scenario, it will be done by the aggregator instead. Really we only support quite a specific scenario (which actually happens to be rather common) where the first aggregation in the chain is one of the supported ones. I think it will be too difficult to try and support multiple or chained aggregations in the fetchData level, though it may be possible if we wish (eg. We could support [MOST_RECENT, SUM], if we wrap the most recent request with another request for the sum).

Ensuring that we get the most utility out of aggregating on the dataFetch will require us to be mindful when creating dashboards/overlays of how the data is being processed throughout the pipeline. It puts a bit more onus on us to write our configs in a way that yields the best performance.

packages/data-api/src/__tests__/TupaiaDataApi.test.js
rohan-bes: Yeah I've been having a few issues with performance in jest and @kael89 mentioned he was having it too... Usually the codeship tests run fine despite it however I actually saw this suite timeout in codeship as well one time...

I suspect we haven't got jest configured quite properly... I'd prefer we got to the bottom of the mystery a bit more before globally extending the timeout?

packages/data-api/src/fetchData.js
rohan-bes: I've been calling it `firstAggregation` everywhere to indicate that it's just the first one in the aggregation list. I'm not too fussed either way though?

chris-bes: I'm happy with it too. Just a minor thought 😄

packages/data-api/src/__tests__/TupaiaDataApi.test.js
kael89: Jest is not slow anymore for me (was a local issue). However, often it does get stuck in local integration tests.

I dis some investigating a while ago. My initial conclusions were that it is related to database connection attempts and jest parallelization (we didn't have those issues with mocha, which runs tests serially).

Hope to get some time to investigate and fix this soon!

packages/data-api/src/fetchData.js
chris-bes: thanks for being patient with all my questions!

PR #2125, opened 2021-02-23
1378: Implement 'count' indicator builder

chris-bes:
### Changes:

- Implemented countEvent indicator
- Renamed arithmetic indicators to arithmeticAnalytic

### Upcoming:

- Unit tests (after the overall design is approved)
- Renaming Arithmetic classes/folders/variables to new names (after we've finalised the name here: https://github.com/beyondessential/tupaia/pull/2125#discussion_r580741280)


packages/database/src/migrations/20210211051517-RenameArithmeticIndicators-modifies-data.js
```diff
@@ -0,0 +1,35 @@
...
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`
+    UPDATE indicator
+    SET builder = 'arithmeticAnalytic'
+    WHERE builder = 'arithmetic';
+  `);
+};
```
chris-bes, line 23: Now that indicator can fetch events or analytics as inputs, we want to make this more obvious in the indicator names. So we want to change `arithmetic` to something more specific. I'm not sure if this is a good name. I'm thinking between 3 options:
- arithmeticAnalytic
- analyticArithmetic
- runAnalyticArithmetic

I went with `arithmeticAnalytic` and `countEvent` to be more consitent with how we name data builders... But any suggestions would be appreciated!

packages/indicators/src/Builder/CountEventBuilder/config/types.ts
```diff
@@ -0,0 +1,19 @@
...
+export type DefaultValue = number | string | 'undefined';
+
+/**
+ * A flexible config schema that can be provided as an input to the indicator
+ */
+export type CountEventConfig = {
+  readonly formula: string;
+  readonly programCode: string;
+  readonly defaultValues?: Record<string, DefaultValue>;
+};
```
chris-bes, line 19: This is the spec of `countEvent` config. We don't have `aggregation` like `arithmetic` indicator because we usually don't aggregate event data. I've also left `paramBuilders` out because:

1. We don't need it at the moment
2. I feel that it's better to wait until there's specific use case for it so that we can come up with a more concise config format?

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -22,6 +22,14 @@ const average = (...argumentList) => {
   return sum / existingValues.length;
 };
 
+/**
+ * List of built in functions in math.js
+ * This list is here because when extracting variables for an expression by traversing the node tree,
+ * the built in functions are also considered as Symbol Node (which is similar to variables).
+ * So we want to exclude them when getting the variables.
+ */
+const BUILT_IN_FUNCTIONS = ['equalText'];
```
chris-bes, line 31: This is needed in a later PR

kael89 (changes requested):
Great PR @chris-bes !! I love how you followed the existing structure for `ArithmeticBuilder`. It definitely made reviewing very pleasant, and I believe it will make code maintenance smoother! 🎉 

All suggestions are about minor things. One note, when you do add tests I highly recommend looking into the integration tests that were added for the arithmetic builder: 

* [integration.test.ts](https://github.com/beyondessential/tupaia/blob/1378-1799-testable-branch/packages/indicators/src/__tests__/__integration/integration.test.ts)
* [ArithmeticAnalyticsFixtures.ts](https://github.com/beyondessential/tupaia/blob/1378-1799-testable-branch/packages/indicators/src/__tests__/__integration/fixtures/arithmeticAnalyticsFixtures.ts)

I found integration tests extremely valuable. With unit tests, after architecture changes my tests were useless. On the other hand, I could use the same integration tests moving back and forth towards architectures. 30'' of running them is often enough to know your code basically works!

[Unit tests](https://github.com/beyondessential/tupaia/blob/1378-1799-testable-branch/packages/indicators/src/__tests__/Builder/ArithmeticBuilder/ArithmeticBuilder.test.ts) also have their place in features like `indicators`, but mainly to test edge cases. I'd say that with thorough integration tests in place, they are a "nice to have"

chris-bes:
@kael89 

Thanks so much for your throughout reviews. I have addressed your reviews and added integration tests for `eventCount` indicator.

I've broken them into 3 parts:
https://github.com/beyondessential/tupaia/pull/2125/commits/522cdb41e273431ce9e440061ca4e80aae0b05d2: For addressing your review comments
https://github.com/beyondessential/tupaia/pull/2125/commits/4f4647e314408564ff4fa340ff6f62d458e14bc6: For renaming the classes/variables to `analyticArithmetic`/`eventCount` like we decided + fixing failing tests
https://github.com/beyondessential/tupaia/pull/2125/commits/fef600b3b7d44e3421756df2cc269e5499b8fda8: For adding integration tests for `eventCount` indicators. 
Extra:
https://github.com/beyondessential/tupaia/pull/2125/commits/784f2f763be11ef180ba7de32039076912062a81 Just a few small fixes to remove a silly test code that I accidently committed in 1 of the commits above

Thanks for setting up the base of integration tests very nicely so I can reuse a lot of them. Regarding unit tests, I actually believe I should also write them. But at the moment I have some other urgent priorities so I would prefer leaving them for later when I have more time... (hope that's okay with you)

packages/indicators/jest.setup.js
```diff
@@ -10,3 +10,5 @@ afterAll(async () => {
   await clearTestData(database);
   await database.closeConnections();
 });
+
+jest.setTimeout(30000)
```
chris-bes: Please ignore this in this commit, I've removed it from the later commit https://github.com/beyondessential/tupaia/pull/2125/commits/784f2f763be11ef180ba7de32039076912062a81

packages/indicators/src/__tests__/__integration/fixtures/eventCount/eventCountFixtures.ts
```diff
@@ -0,0 +1,144 @@
...
+      { code: 'Quarantine_Place', type: 'Radio' },
+      { code: 'Positive', type: 'Binary' },
+      { code: 'Age', type: 'Number' },
+    ],
+  },
+];
+
+const ARRAY_SURVEY_RESPONSES: ArraySurveyResponse[] = [
+  // AU
+  ['Covid_19_Cases', 'AU', '2020-03-01', { Quarantine_Place: 'Hotel', Positive: 'Yes', Age: '20' }],
```
kael89, line 33: I see you're drawing some inspiration from recent events! 😎

packages/indicators/src/__tests__/__integration/fixtures/eventCount/eventCountFixtures.ts
```diff
@@ -0,0 +1,144 @@
...
+  ['Covid_19_Cases', 'AU', '2020-04-30', { Quarantine_Place: 'Home', Positive: 'No', Age: '25' }],
+  ['Covid_19_Cases', 'AU', '2020-05-29', { Quarantine_Place: 'Hotel', Positive: 'No', Age: '38' }],
+  ['Covid_19_Cases', 'AU', '2020-06-25', { Quarantine_Place: 'Home', Positive: 'Yes', Age: '40' }],
+  // NZ
+  ['Covid_19_Cases', 'NZ', '2020-03-01', { Quarantine_Place: 'Hotel', Positive: 'Yes', Age: '35' }],
+  ['Covid_19_Cases', 'NZ', '2020-03-30', { Quarantine_Place: 'Home', Positive: 'No', Age: '23' }],
+  ['Covid_19_Cases', 'NZ', '2020-04-30', { Quarantine_Place: 'Home', Age: '23' }], // Missing Quarantine_Place
+  // US
+  ['Covid_19_Cases', 'US', '2020-03-01', { Quarantine_Place: 'Hotel', Positive: 'Yes', Age: '22' }],
```
kael89, line 42: The `US` data are not used, so we could delete them.

There is definitely a point to defining fixtures for multiple entities and verifying that only the specified entities are used. We already check this by having data for the same periods in `AU` and `NZ`.

packages/indicators/src/types.ts
```diff
@@ -15,6 +15,17 @@ export interface Analytic extends AnalyticValue {
   readonly dataElement: string;
 }
 
+export type DataValues = Record<string, DbValue>;
```
kael89, line 18: Not sure if events values are always `string` or `string|number`?

```suggestion
export type DataValues = Record<string, string>;
```

kael89 (changes requested):
Great work Chris! A few more details and this is ready to go!

1. The purpose of the existing `indicatorApiFixtures` was to test generic `IndicatorApi` behaviour (regardless of builder specifics). So, I would suggest to use a structure like the following:
   ```
   __integration
     fixtures
       analyticArithmeticFixtures.ts
       eventCountFixtures.ts
       indicatorApiFixtures.ts
    ```

   We don't have to test all indicators in `indicatorApiFixtures.ts` (since they will be also tested independently). But I believe it would be valuable to add a test case for fetching multiple codes across multiple builders. Apart from that, we could use any builder for the rest of the `Indicator API` integration tests

2. A bit more general comment: now that I see the integration tests, I don't think that the count builder in its current form actually "counts' anything? Rather, it evaluates boolean expressions.

   Perhaps it would be valuable to change the implementation, so that after analytics are calculated (as is currently the case) we sum them per org unit across the specified date range. Aggregation may be useful here too (eg sum per month), but if it's not required we can skip it for now.

   Happy to discuss this in a call if you have more questions! Above all, we should keep in mind how this is going to be used in practical scenarios.

chris-bes:
> Great work Chris! A few more details and this is ready to go!
> 
> 1. The purpose of the existing `indicatorApiFixtures` was to test generic `IndicatorApi` behaviour (regardless of builder specifics). So, I would suggest to use a structure like the following:
>    ```
>    __integration
>      fixtures
>        analyticArithmeticFixtures.ts
>        eventCountFixtures.ts
>        indicatorApiFixtures.ts
>    ```
>    
>    
>    We don't have to test all indicators in `indicatorApiFixtures.ts` (since they will be also tested independently). But I believe it would be valuable to add a test case for fetching multiple codes across multiple builders. Apart from that, we could use any builder for the rest of the `Indicator API` integration tests
> 2. A bit more general comment: now that I see the integration tests, I don't think that the count builder in its current form actually "counts' anything? Rather, it evaluates boolean expressions.
>    Perhaps it would be valuable to change the implementation, so that after analytics are calculated (as is currently the case) we sum them per org unit across the specified date range. Aggregation may be useful here too (eg sum per month), but if it's not required we can skip it for now.
>    Happy to discuss this in a call if you have more questions! Above all, we should keep in mind how this is going to be used in practical scenarios.

Thanks Kostas, I agree with you on the 1st one

For the 2nd point, I might be wrong, but in my head, since the indicator will output **analytics**, I was thinking that it's better to keep them raw (in this case per orgunit/per period) so that we can perform whatever aggregations we want with the indicator outputs which can be more flexible. Perhaps the name is misleading? Maybe we should call it `eventCheckCondition` instead?

kael89:
@chris-bes About point 2: I agree, `eventCheckCondition` seems more reusable. So we have 3 options:
1. Rename the indicator builder
2. Do not rename it, but change the existing implementation to calculate the `count`
3. Expose both indicators

I believe you have created some data builders locally that use the new indicator builder, so it may be a good idea to confirm that option 1 makes sense in that context too.

Please let me know how this went, then this is ready to go! 🎉

packages/indicators/src/types.ts
chris-bes: I think there's number as well because we are now sanitizing internal binary data to 1/0

kael89 (approved):
I know I'm really frustrating now but how would you feel about  adding a final `-s` to `eventCheckCondition` (make it `eventCheckConditions`)?

Apart from that it looks great, pre-approving! 🙌

kael89 (approved):
Ready to go! 🎉

chris-bes:
I made a bit of mistake and forgot to commit the deleted files, so pushed a new commit for that!

PR #2129, opened 2021-02-23
1788 Add source at mapoverlay

billli0:
### Issue #:
[1788 Add source for weather map overlays data](https://github.com/beyondessential/tupaia-backlog/issues/1788)

### Changes:
- Add `ReferenceTooltip.js` on map overlay group and map overlay (`HierarchyItem.js`)
- Integrate reference info in `findAccessibleGroupedMapOverlays.js`, which will show reference against the group rather than each overlay if every map overlay in a group has the same source, see [comment](https://github.com/beyondessential/tupaia-backlog/issues/1788#issuecomment-780936014)
- Migration to add reference to every `mapOverlay` under group `Weather Observations/Forecast`


packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -54,31 +58,66 @@ const findNestedGroupedMapOverlays = async (
       const mapOverlayGroupRelation = mapOverlayGroupRelations[i];
       const name = mapOverlayGroupIdToName[mapOverlayGroupRelation.child_id];
       const childMapOverlayGroupRelations = await mapOverlayGroupRelation.findChildRelations();
-      const children = await findNestedGroupedMapOverlays(
+      const mapOverlays = await findNestedGroupedMapOverlays(
```
chris-bes, line 61: We do support nested map overlay groups (in laos schools: https://github.com/beyondessential/tupaia-backlog/issues/1788#issuecomment-784827175). So the original `children` here actually can have map overlay groups as well. So I think naming this `mapOverlays` may not be accurate

packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -184,11 +223,12 @@ export const findAccessibleGroupedMapOverlays = async (models, accessibleMapOver
 
     if (isNonEmptyMapOverlayGroup) {
       accessibleRelations.push(worldRelationById[groupId]);
-      accessibleOverlayGroups.push({
+      const accessibleOverlayGroup = integrateMapOverlayReference({
+        mapOverlays: nestedMapOverlayGroups,
```
chris-bes, line 227: `nestedMapOverlayGroups` is probably a list of map overlays in Laos EOC because we only have 1 level of nesting there. But if we do this in multi nested level groups like Laos Schools, `nestedMapOverlayGroups` will be a list of over lay groups I believe. Example screenshot here: https://github.com/beyondessential/tupaia-backlog/issues/1788#issuecomment-784827175

In addition, the current code also supports having group and overlays in the same group (no projects use this at the moment, but can be in the future). Example screenshot is here: https://github.com/beyondessential/tupaia-backlog/issues/194#issuecomment-679847182. So `nestedMapOverlayGroups` can actually contain both map overlays and overlay groups.

Fortunately I think you probably don't need to rewrite too much? (I might be a bit too naive). Perhaps the below code should be enough? If so the same has to be done in the other place that you call the same function

```suggestion
      const accessibleOverlayGroup = integrateMapOverlayReference({
        mapOverlays: nestedMapOverlayGroups, // do filter for map overlays here
```

It'd be great if you can test locally if it works with overlays in multi nested groups in Laos Schools as well.

chris-bes (changes requested):
Nice work @billli0 ! Just 1 concern about the multi nested groups/group that can contain both overlays and groups. Hopefully that's not too difficult to fix

packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
billli0: Thanks Chris for these sharing! I thought this new recursive function can deal with having group and overlays in the same group, but I didn't test it to be honest. 

I guess option 2 will have conflict when two sub groups has different source, so I agree options 1 makes sense to me and is easier to implement. I will go and implement this.

packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -54,31 +58,66 @@ const findNestedGroupedMapOverlays = async (
...
 };
 
+// Only display reference info on map overlay group if all map overlays have same reference.
+const integrateMapOverlayReference = ({ mapOverlays, id, groupName }) => {
+  const mapOverlayGroupResult = {
+    id,
+    name: groupName,
+  };
+
+  const getReference = mapOverlay => get(mapOverlay, REFERENCE_PATH);
```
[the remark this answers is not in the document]
billli0: Thanks, `get` used for other purpose before, but now it looks redundant, I will remove it.

packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -54,31 +58,66 @@ const findNestedGroupedMapOverlays = async (
...
+  const firstReference = getReference(mapOverlays[0]);
+  if (firstReference) {
+    const referencesAreNotTheSame = mapOverlays.find(
+      mapOverlay =>
+        !(getReference(mapOverlay) && isEqual(getReference(mapOverlay), firstReference)),
+    );
```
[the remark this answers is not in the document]
billli0: Thanks Chris. I thought `find` will stop the loop and `every` doesn't, but it seems like `every` can smartly stop the loop as well. Also for readability I agree we should change it to `every`.

packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -59,26 +61,62 @@ const findNestedGroupedMapOverlays = async (
...
   return sortedMapOverlayResults.map(item => {
-    //id was added only for sorting purposes, remove it because it is not required in the front end
+    // id was added only for sorting purposes, remove it because it is not required in the front end
     const { id, ...itemToReturn } = item;
     return itemToReturn;
   });
 };
 
+// Only display reference info on map overlay group if all map overlays have same reference.
+const integrateMapOverlayReference = ({ id, groupName, children }) => {
```
chris-bes, line 84: slightly nit picky but since `children` can contain both overlays and overlay groups, maybe it's better to change all the variables `mapOverlays`/`mapOverlay` here to `mapOverlayItems`/`mapOverlayItem`?

billli0: Good idea Chris 👍

chris-bes (approved):
Nice work @billli0 !

PR #2130, opened 2021-02-23
2263: Make sure entity updates are not pushed to Laos DHIS

chris-bes:
### Changes:

- Added push flag to metadata column of entity to prevent pushing to DHIS2
- Changed dhis.id to dhis.trackedEntityId for clarity

### DIFF
--- a/packages/database/src/migrations/20210222124639-UpdateLaosEntitiesPushMetadata-modifies-data.js
+++ b/packages/database/src/migrations/20210222124639-UpdateLaosEntitiesPushMetadata-modifies-data.js
@@ -1,0 +1,62 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`
+    -- We want to drop the entity trigger because we just want to add push: false to Laos entities, no actual data is changed, this shouldn't trigger any change listener (eg: for pushing)
+    DROP TRIGGER IF EXISTS entity_trigger
+    ON entity;
+
+    -- Add push = false to Laos entities
+    UPDATE entity
+    SET metadata = jsonb_set(metadata, '{dhis,push}', 'false')
+    WHERE country_code = 'LA'
+    AND metadata->'dhis' IS NOT NULL;
+
+    -- Also change the metadata in data column if dhis_sync_log to reflect the same push = false
+    UPDATE dhis_sync_log
+    SET data = jsonb_set(data::jsonb, '{metadata,dhis,push}', 'false')::text
+    WHERE record_type = 'entity'
+    AND record_id IN (SELECT id FROM entity WHERE country_code = 'LA' AND metadata->'dhis' IS NOT NULL);
+  `);
+};
+
+exports.down = async function (db) {
+  await db.runSql(`
+    DROP TRIGGER IF EXISTS entity_trigger
+    ON entity;
+
+    UPDATE entity
+    SET metadata = metadata #- '{dhis,push}'
+    WHERE country_code = 'LA'
+    AND metadata->'dhis'->'push' IS NOT NULL;
+
+    UPDATE dhis_sync_log
+    SET data = (data::jsonb #- '{metadata,dhis,push}')::text
+    WHERE record_type = 'entity'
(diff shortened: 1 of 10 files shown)


packages/database/src/migrations/20210222124639-UpdateLaosEntitiesPushMetadata-modifies-data.js
```diff
@@ -0,0 +1,55 @@
...
+    push: false,
+    isDataRegional: true,
+  },
+};
+
+exports.up = async function (db) {
+  await db.runSql(`
+    -- We want to drop the entity trigger because we just want to add push: false to Laos entities, no actual data is changed, this shouldn't trigger any change listener (eg: for pushing)
+    DROP TRIGGER IF EXISTS entity_trigger
+    ON entity;
```
chris-bes, line 28: We just want to add the `push` flag to metadata column, no actual data is changed. So dropping the trigger for entity within this transaction to prevent having 20000 entity change listeners fired

packages/database/src/migrations/20210222204800-UpdateEntityDhisIdToTrackEntityInstanceId-modifies-data.js
```diff
@@ -0,0 +1,66 @@
...
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`
+    -- We want to drop the entity trigger because we just want to change the id key -> trackedEntityId, no actual data is changed, this shouldn't trigger any change listener (eg: for pushing)
+    DROP TRIGGER IF EXISTS entity_trigger
+    ON entity;
```
chris-bes, line 21: Same like above, just want to update the key `dhis.id` to `dhis.trackedEntityId`, no need to fire change listeners

packages/meditrak-server/src/dhis/DhisChangeValidator.js
```diff
@@ -110,9 +110,17 @@ export class DhisChangeValidator extends ChangeValidator {
     return this.queryValidSurveyResponseIds(surveyResponseIds);
   };
 
+  getValidEntityUpdates = async updateChanges => {
+    const entityIds = this.getIdsFromChangesForModel(updateChanges, this.models.entity);
+    if (entityIds.length === 0) return [];
+
+    const entities = await this.models.entity.findManyById(entityIds);
+    return entities.filter(e => e.allowsPushingToDhis()).map(e => e.id);
```
chris-bes, line 118: This is where it actually filter the changes and prevent creating sync queue records for the updated entities that we don't want to push to DHIS2

packages/meditrak-server/src/routes/importEntities/updateCountryEntities.js
```diff
@@ -118,6 +118,14 @@ export async function updateCountryEntities(transactingModels, countryName, enti
         dataServiceEntityToUpsert,
       );
     }
+    const entity = await transactingModels.entity.findOne({ code });
+    const metadata =
+      entity && entity.metadata
+        ? entity.metadata // we don't want to override the metadata if the entity already exists
```
edmofro, line 124: Ooh good fix!

packages/database/src/migrations/20210222124639-UpdateLaosEntitiesPushMetadata-modifies-data.js
```diff
@@ -0,0 +1,55 @@
...
+    -- Add push = false to Laos entities
+    UPDATE entity
+    SET metadata = '${JSON.stringify(NEW_METADATA)}'
+    WHERE country_code = 'LA';
+
+    -- Also change the metadata in data column if dhis_sync_log to reflect the same push = false
+    UPDATE dhis_sync_log
+    SET data = jsonb_set(data::jsonb, '{metadata}', '${JSON.stringify(NEW_METADATA)}')::text
+    WHERE record_type = 'entity'
+    AND record_id IN (SELECT id FROM entity WHERE country_code = 'LA');
```
chris-bes, line 39: There are some existing sync_logs for Laos entities. I'm not sure if we can/should delete them...? So this is just to update the `data` column to also have the same `push` flag = false

edmofro (approved):
Looks good to me!

packages/database/src/migrations/20210222124639-UpdateLaosEntitiesPushMetadata-modifies-data.js
edmofro: I think it's fine either way to delete or leave. They're kind of just clutter, but it's also a very cluttered table anyway! They probably represent the sync to the regional dhis2 server, which is valid, but if we deleted the Laos entities from there (which we may want to), then we should definitely delete both the sync logs and sync queue records. For now, up to you.

PR #2132, opened 2021-02-24
826: Script to run e2e tests

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/826

Adds e2e tests in our CI/CD
* They are triggered on branches ending with `-e2e`
* Implements [the suggested e2e test flow](https://github.com/beyondessential/tupaia-backlog/issues/826#issuecomment-780307985), apart from snapshot pushing/pulling. This part of the flow is captured in another card (https://github.com/beyondessential/tupaia/issues/1534), so it will be implemented separately

Also a bit of refactoring/cleanup, sorry for that - hopefully it's not too messy!


kael89:
**Manual steps required** after this is merged in `dev` (and `prod`):
* Add `CYPRESS_RECORD_KEY` in `web-frontend/.env`. Environment: `e2e`

packages/utils/src/requireEnv.js
```diff
@@ -0,0 +1,19 @@
...
+
+export const requireEnv = variable => {
+  const value = process.env[variable];
+  if (value === undefined) {
+    throw new Error(
+      `Could not load env variable '${variable}', required in ${getCurrentPackage()} `,
+    );
+  }
+  return value;
+};
```
rohan-bes, line 19: This is really handy, nice 👍

packages/web-frontend/cypress/scripts/generateConfig/index.js
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env node
```
rohan-bes, line 1: I this required to call this script from `yarn`?

packages/web-frontend/cypress/support/actions.js
```diff
@@ -4,14 +4,14 @@
...
 
 import { TEST_USER } from '../constants';
-import { TestUserPasswordUndefinedError } from './helpers';
 
-const getTestUserPassword = () => {
-  const password = Cypress.env('TEST_USER_PASSWORD');
-  if (!password) {
-    throw new TestUserPasswordUndefinedError();
+const requireCyEnv = variable => {
+  const unprefixedVariable = variable.replace(/^CYPRESS_/, '');
```
rohan-bes, line 9: Slightly confused by this, so we are removing the `CYPRESS_` from the start of the variable? But later on we are calling it without a `CYPRESS_` prefix? Is this just to clean it? My final point of confusion is that aren't the variables prefixed with `CYPRESS_` in the `.env` file? Does Cyprus remove those prefixes when loading in variables?

rohan-bes (approved):
Looks good! Couple of confusions/questions but nothing major.

packages/web-frontend/cypress/scripts/generateConfig/index.js
kael89: I believe it is, since the shebang isn't there, it will run it as a bash script by default

kael89: Actually, they are not required since we invoke those with `babel-node` or (`node`)... This is a problem only if the package is a field itself: see https://docs.npmjs.com/cli/v7/configuring-npm/package-json#bin

I recently encountered this issue in a personal project, that's why I updated those... I believe it's still good practice to include the shebang to indicate the script runner, what do you think?

packages/web-frontend/cypress/support/actions.js
kael89: Cypress doesn't automatically load environment variables files from `.env` files (it uses its own config files)...We use [cypress-dotenv](https://www.npmjs.com/package/cypress-dotenv) for that.

`cypress-dotenv` only loads variables prefixed with `CYPRESS_` by default. Then, to actually retrieve the variables you have to omit the prefix... To be honest, I just realised there is a flag that you can pass to make it load all variables instead of just `CYPRESS_` ones.

I guess this is a standard practice when something is exposed to the client, probably in order to protect other sensitive data from being exposed, for example:
* [Gatsby env variables](https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/#client-side-javascript)
* [Create React app env variables](https://create-react-app.dev/docs/adding-custom-environment-variables/)

But this shouldn't be a concern here since the tests run on the server.

I still find it handy to prefix those variables with `CYPRESS_`, although I do understand that removing the prefix afterwards may seem a bit weird... Perhaps I should add a comment?

kael89: Correction: it seems that the `CYPRESS_` prefix **can** (not **must**, as I thought) be ommitted. So both of the following work:

1. `Cypress.env(`CYPRESS_TEST_USER_PASSWORD`);
2.`Cypress.env(`TEST_USER_PASSWORD`);

The current implementation of the function supports both use cases, by removing the (optional) prefix before getting the value, and adding it when printing the name of the missing variable. It thus assumes that variables **must** be prefixed where they are actually defined

PR #2134, opened 2021-02-24
2237 Raw data download request time out

billli0:
### Issue #:
[2237 Network request timed out error](https://github.com/beyondessential/tupaia-backlog/issues/2237)
### Changes:
- Fetch from db in group rather than in sequence


packages/meditrak-server/src/routes/exportSurveyResponses/exportSurveyResponses.js
```diff
@@ -143,45 +149,54 @@ export async function exportSurveyResponses(req, res) {
...
+        // Fetch all answers of all survey responses for further use in 'processSurveyResponse()'.
+        const surveyResponseIds = surveyResponses.map(response => response.id);
+        const answers = await findAnswersInSurveyResponse(
+          models,
+          surveyResponseIds,
+          {},
+          { columns: [{ [`${TYPES.SURVEY_RESPONSE}.id`]: 'survey_response.id' }], sort: [] },
+        );
+        const nestedAnswers = groupBy(answers, 'survey_response.id');
+        console.log(nestedAnswers);
```
kael89, line 193:
```suggestion
```

packages/meditrak-server/src/routes/exportSurveyResponses/exportSurveyResponses.js
```diff
@@ -143,45 +149,54 @@ export async function exportSurveyResponses(req, res) {
...
-            surveyResponseIndex < surveyResponses.length;
-            surveyResponseIndex++
-          ) {
-            await processSurveyResponse(surveyResponses[surveyResponseIndex], entity);
-          }
+        const surveyResponses = await models.surveyResponse.find(
+          surveyResponseFindConditions,
+          sortAndLimitSurveyResponses,
+        );
+        const nestedEntitiesById = keyBy(entities, 'id');
```
kael89, line 183: `keyBy` maps an item to another item, so I wouldn't think that entities are nested here

```suggestion
        const entitiesById = keyBy(entities, 'id');
```

packages/meditrak-server/src/routes/exportSurveyResponses/exportSurveyResponses.js
```diff
@@ -143,45 +149,54 @@ export async function exportSurveyResponses(req, res) {
...
+        // Fetch all answers of all survey responses for further use in 'processSurveyResponse()'.
+        const surveyResponseIds = surveyResponses.map(response => response.id);
+        const answers = await findAnswersInSurveyResponse(
+          models,
+          surveyResponseIds,
+          {},
+          { columns: [{ [`${TYPES.SURVEY_RESPONSE}.id`]: 'survey_response.id' }], sort: [] },
+        );
+        const nestedAnswers = groupBy(answers, 'survey_response.id');
```
kael89, line 192: `nested` is usually used when something is an array of arrays (or object of objects etc)... Here, we can make the grouping context a bit more obvious:

```suggestion
        const answersByResponseId = groupBy(answers, 'survey_response.id');
```

packages/meditrak-server/src/routes/exportSurveyResponses/exportSurveyResponses.js
```diff
@@ -143,45 +149,54 @@ export async function exportSurveyResponses(req, res) {
...
+        console.log(nestedAnswers);
+        for (const response of surveyResponses) {
+          await processSurveyResponse(
+            response,
+            nestedEntitiesById[response.entity_id],
+            nestedAnswers,
```
kael89, line 198: I assume we can pass the answers for the current survey directly (similarly to how we directly pass the entity one line above). In this case, you'll have to refactor `processSurveyResponse` a bit

```suggestion
            answersByResponseId[response.id],
```

packages/meditrak-server/src/routes/exportSurveyResponses/exportSurveyResponses.js
```diff
@@ -123,15 +126,18 @@ export async function exportSurveyResponses(req, res) {
...
         const dateString = moment(surveyDate).format(EXPORT_DATE_FORMAT);
         if (!easyReadingMode) exportData[0].push(currentSurveyResponse.id);
         exportData[1].push(currentEntity.code);
         exportData[2].push(responseName);
         exportData[3].push(dateString);
-        const answers = await findAnswersInSurveyResponse(models, currentSurveyResponse.id);
+        const answers =
+          nestedAnswers && nestedAnswers[currentSurveyResponse.id]
+            ? nestedAnswers[currentSurveyResponse.id]
+            : await findAnswersInSurveyResponse(models, currentSurveyResponse.id);
```
kael89, line 140: Why would we want to refetch answers here? I would expect that all of them have been fetched already

If this is true, we can simply used the answers in this function's input, and remove `async` from this function's signature

kael89 (changes requested):
Good initiative making this code faster @billli0 ! Let's make a few aspects of it simpler too

packages/meditrak-server/src/routes/exportSurveyResponses/exportSurveyResponses.js
billli0: Thanks Kostas, `answersByResponseId` looks good 👍

packages/meditrak-server/src/routes/exportSurveyResponses/exportSurveyResponses.js
billli0: Yeah because we use `processSurveyResponse` at another place here, but I agree we can refactor that as well, then it will look better.

packages/meditrak-server/src/routes/exportSurveyResponses/exportSurveyResponses.js
billli0: Will change 👍

kael89 (approved):
Nice work! Let's make sure that both branches of the `if condition` in line 148 are tested:
1. Single survey response specified
2. We are looking for multiple responses

PR #2141, opened 2021-02-25
2343: Prevent dhisApi from pushing changes to Laos DHIS2 server

chris-bes:
### Changes:

- Throw error in dhis api if we attempt to push data for laos dhis.

### DIFF
--- a/packages/dhis-api/src/DhisApi.js
+++ b/packages/dhis-api/src/DhisApi.js
@@ -34,9 +34,10 @@ const {
 const LATEST_LOOKBACK_PERIOD = '600d';
 
 export class DhisApi {
-  constructor(serverName, serverUrl) {
+  constructor(serverName, serverUrl, serverPushEnable = true) {
     this.serverName = serverName;
     this.serverUrl = serverUrl;
+    this.serverPushEnable = serverPushEnable;
     this.fetcher = new DhisFetcher(serverName, serverUrl, this.constructError);
     this.deleteEvent = this.deleteEvent.bind(this);
   }
@@ -132,6 +133,10 @@ export class DhisApi {
   }
 
   async post(endpoint, data, queryParameters) {
+    if (!this.serverPushEnable) {
+      throw new Error(`Attempted to push data to ${this.serverName} that should not be pushed`);
+    }
+
     return this.fetch(endpoint, queryParameters, {
       body: JSON.stringify(data),
       method: 'POST',
@@ -139,6 +144,10 @@ export class DhisApi {
   }
 
   async put(endpoint, data, queryParameters) {
+    if (!this.serverPushEnable) {
+      throw new Error(`Attempted to push data to ${this.serverName} that should not be pushed`);
+    }
+
     return this.fetch(endpoint, queryParameters, {
       body: JSON.stringify(data),
       method: 'PUT',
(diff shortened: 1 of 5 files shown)


packages/dhis-api/src/DhisApi.js
```diff
@@ -132,13 +133,21 @@ export class DhisApi {
...
   }
 
   async put(endpoint, data, queryParameters) {
+    if (!this.serverPushEnable) {
+      throw new Error(`Attempted to push data to ${this.serverName} that should not be pushed`);
+    }
+
     return this.fetch(endpoint, queryParameters, {
       body: JSON.stringify(data),
       method: 'PUT',
```
IgorNadj, line 153: Can we add it to ` delete()` as well. I think that's all of them that mutate data

packages/dhis-api/src/DhisApi.js
```diff
@@ -132,20 +133,32 @@ export class DhisApi {
   }
 
   async post(endpoint, data, queryParameters) {
+    if (!this.serverSyncEnable) {
+      throw new Error(`Attempted to push data to ${this.serverName} that should not be pushed`);
+    }
```
chris-bes, line 138: @edmofro Just want to run this through you.

@IgorNadj and I have discussed and we decided to add this in `DhisApi` as the final gate to prevent any changes go into Laos DHIS2 server. Basically we'll just need to add `serverSyncEnable` env variable for Laos EOC DHIS2.

Ideally it should never reach here for Laos DHIS2 because:
- We will have an option to disable syncing when importing NEW entities by this task (https://github.com/beyondessential/tupaia-backlog/issues/2301)
- We already disabled pushing for all EXISTING Laos entities here (https://github.com/beyondessential/tupaia/pull/2130)

So we put this as the last guard. This is just us being very risk adverse. What do you reckon?

edmofro: Yeah I think it's a good idea! Will review

packages/utils/src/dhis.js
```diff
@@ -67,6 +70,8 @@ export const getDhisConfig = ({
   }
   const serverName = serverNameInput || getServerName(entityCode, isDataRegional);
   const serverUrl = getServerUrlFromName(serverName);
+  const serverSyncEnableParam = getServerVariable(serverName, 'DHIS_SYNC_ENABLE');
+  const serverSyncEnable = serverSyncEnableParam ? serverSyncEnableParam === 'true' : true;
```
edmofro, line 74: This will mean the default is `false`, so we'll need to add environment variables to turn it on for the existing servers. Would prefer to flip the default.

Perhaps the variable can be `serverReadOnly` instead, which flips everything?

packages/utils/src/dhis.js
```diff
@@ -41,6 +41,9 @@ const getServerName = (entityCode, isDataRegional) => {
     : REGIONAL_SERVER_NAME; // If the country does not have a dhis2 server, use the regional server
 };
 
+const getServerVariable = (serverName, variableName) =>
+    process.env[`${serverName.toUpperCase()}_${variableName}`] || process.env[variableName];
```
edmofro, line 45: Might be easier to just hard code it up by `SERVER_NAMES` given it's not going to change across deployments, so we can save ourselves from managing an environment variable?

edmofro (changes requested):
I like the setup of going through `getDhisConfig`. A couple of suggestions

edmofro (approved):
🎉

PR #2142, opened 2021-02-25
2301: Added option disable push entity to dhis2 when importing new entities in Admin Panel

chris-bes:
### Changes:

- Added option disable push entity to dhis2 server when importing new entities in Admin Panel

### Screenshot:
https://github.com/beyondessential/tupaia-backlog/issues/2301#issuecomment-786368352


edmofro (approved):
LGTM :-)

PR #2144, opened 2021-02-25
2235 add lesmis package

tcaiger:
### Issue #: [2235](https://github.com/beyondessential/tupaia-backlog/issues/2235) Set up LESMIS backend package

Questions
- Is the name of tupaia-server package ok?
- How much to move to tupaia-server?
  - Should we add meditrak-connection to it?
  - Should we add login and logout routes to it?
  - Should it return a create app function instead of just base classes? 
- How best to handle types between parent and child classes?


packages/lesmis-server/src/routes/Route.ts
```diff
@@ -0,0 +1,42 @@
...
+  reportConnection?: ReportConnection;
+
+  async handle() {
+    // All routes will be wrapped with an error catcher that simply passes the error to the next()
+    // function, causing error handling middleware to be fired. Otherwise, async errors will be
+    // swallowed.
+    try {
+      const session = await this.verifyAuth();
+      this.meditrakConnection = new MeditrakConnection(session);
+      this.reportConnection = new ReportConnection(session);
```
edmofro, line 24: I would keep just these in the lesmis-server `Route` class, and have the parent include `handle()` with a call to `this.setupConnections` (another abstract method that can be implemented at this level)

edmofro (changes requested):
I realise this is a draft PR but as I'm familiar with the code I went ahead and did a full review. I'd suggest we just keep the one PR rather than split it - I can keep reviewing the changes from here.

It's looking great to me. Regarding your questions:
- I prefer something like `server-boilerplate` to `tupaia-server` (which I think people might confuse as an actual service, probably with the responsibilities meditrak-server currently has)
- I'm on the fence with all of the questions about how much to move! There are arguments either way, but
  - I marginally favour `MeditrakConnection` et al. living in the `lesmis-server` package at the moment - handy because e.g. LESMIS might want to get all `meditrak-server` responses as camel case where other services might not, etc. If we get more server implementations and they're all using really similar api connection layers, that might change, but that's where I'm at right now.
  - I think yes to Login and Logout routes being in it. Given that it holds `AuthConnection` and that all tupaia servers should authenticate in the same way, that makes sense to me.
  - `createApp` function instead of base classes... My gut reaction was initially "yes", but after reading through I can also see it making it hard if we have a specific need for some express middleware and it's hard to see where to put it. Let's stick with the base classes for now. Sometimes trying too hard not to repeat yourself leads to confusion later, and I have the feeling this could be one of those places.
- Types between parent and child classes: could you give more context here?

packages/tupaia-server/src/Route.ts
```diff
@@ -0,0 +1,71 @@
...
+    const session: SessionType = await this.sessionModel.findById(sessionId);
+    if (!session) {
+      throw new UnauthenticatedError('Session not found in database');
+    }
+
+    return session;
+  }
+
+  async verifyAuth(): Promise<SessionType> {
+    return this.getSession();
```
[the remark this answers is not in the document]
tcaiger: I went for making the whole class abstract as well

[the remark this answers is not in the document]
tcaiger: Actually then I decided not to 😬

tcaiger:
@edmofro - this is ready for another review. Happy to discuss if it's easier when we are in the office. Regarding the types, I managed to answer my own question, so don't worry about that.

packages/server-boilerplate/README.md
```diff
@@ -0,0 +1,33 @@
...
+See lesmis-server for a typical example of implementation.
+
+## App bootstrapping
+
+#### Attaching the session model
+- The session model needs to be added to the app. This can be done using the `sessionModel` model and `attachSessionModel` util which are found in this package.
+eg. `app.use(attachSessionModel(sessionModel));`
+ 
+#### addRoutesToApp
+- When adding the routes to the app, you will need to wrap the routes with the the `handleRoute` util for the routes to work properly.
```
edmofro, line 14:
```suggestion
- When adding the routes to the app, you will need to wrap the routes with the `handleWith` util for the routes to work properly.
```

packages/server-boilerplate/README.md
```diff
@@ -0,0 +1,33 @@
...
+eg. `app.get('/v1/test', handleWith(TestRoute));`
+
+- You also need to add the error hanlder as the fallback route.
```
edmofro, line 17:
```suggestion
- You also need to add the error handler as the fallback route.
```

packages/server-boilerplate/README.md
```diff
@@ -0,0 +1,33 @@
...
+eg. `app.use(handleError);`
+
+## Routes
+- Includes Route which is base class with useful functionality for handling responses. Route is an abstract class which must be extended to custom routes.
```
edmofro, line 21:
```suggestion
- Includes `Route` which is base class with useful functionality for handling responses. `Route` is an abstract class which must be extended to custom routes.
```

edmofro (approved):
Looks great! Went commit by commit, and definitely agree the with the call to return `LoginRoute` to extending `UnauthenticatedRoute` and using a different name for the verification step 👌 

Just a couple typos, but fully approved 🎉

PR #2148, opened 2021-02-26
2298 remove Laos oxygen concentrators project

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/2298#issuecomment-785467065

This migration assumes that all the survey responses associated with the Laos facilities being deleted have been backed up / deleted on the admin panel first (cos it's easier to do it this way)
- This hasn't happened yet but will happen before testing / releasing this migration

This migration essentially removes Laos from being accessible from Explore mode (and also removes some map overlay / dashboard report associations related to this) and deletes the old facilities from the oxygen concentrators project (as per the comment linked above)

### DIFF
--- a/packages/database/src/migrations/20210225033329-RemoveLaosOxygenConcentratorsProject-modifies-data.js
+++ b/packages/database/src/migrations/20210225033329-RemoveLaosOxygenConcentratorsProject-modifies-data.js
@@ -1,0 +1,135 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const FACILITIES_TO_REMOVE = [
+  'LA_XAXIE',
+  'LA_VCSAN',
+  'LA_VCPAR',
+  'LA_PHSAM',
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210225033329-RemoveLaosOxygenConcentratorsProject-modifies-data.js
```diff
@@ -0,0 +1,135 @@
...
+    "id"='9';
+  `);
+
+  await db.runSql(`
+  UPDATE
+    "mapOverlay"
+  SET
+    "countryCodes"='{NR,NU,PW,TV,AS,GU,PF,NC,MP,PI,WF,CK,PG,DL,SB,TK,PH,VE,MH,WS,FM,KI,CI,TO,VU}'
+  WHERE
+    "id"='126';
```
edmofro, line 107: So we don't use these two map overlays in any other projects for Laos? E.g. Laos Schools, Laos EOC?

Helpful to put a lil comment next to the id so reviewer can tell what the overlay is without looking in the db 🙏

edmofro (commented):
Code looks great 🎉 

Just a question about those map overlays

packages/database/src/migrations/20210225033329-RemoveLaosOxygenConcentratorsProject-modifies-data.js
avaek: Thanks @edmofro - yup, those two map overlays only exist in Laos for the "Explore" project (which this migration removes)

edmofro (approved):
lgtm~

PR #2149, opened 2021-02-28
1675 fetp graduate locations matrix

nick-beyondessential:
### Issue #: [1675](https://github.com/beyondessential/tupaia-backlog/issues/1675) FETP Graduate locations matrix (National and Provincial)

This report counts the number of graduates (individuals) at the National (PG_National Capital), Provincal (district) and District (sub_district) levels grouped by their Work Category (`FETPNG20Data_009`.) It was decided in refinement to us entity hierarchy instead of answers to `FETPNG20Data_034` and `FETPNG20Data_033` - for project consistency.
Using `tableOfCalculatedValues` databuilder seemed a less painful route than adding to aggregator functionality (Group by datavalue coupled with adding ORG_GROUP aggregation on most direct ancestor (district,sub_district and quasi national)
but happy to discuss if you think it would be a better option. This seems to work well for the 100 or so indiviuals in this project. 

### Changes:
- Added parent type and parent code filter for aggregators in `calculateOperationForAnalytics`
- Added COUNT_CONDITION type for `calculateOperationForAnalytics`
- migration fro national and district level dashboards reports


packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
```diff
@@ -41,20 +41,72 @@ const performSingleAnalyticOperation = (analytics, config) => {
...
+    if (
+      organisationUnitParentType &&
+      !checkOrganisationUnitParentValue(
+        organisationUnitParentMap,
+        a.organisationUnit,
+        'type',
+        organisationUnitParentType,
+      )
+    )
+      return false;
```
kael89, line 93:
```suggestion
    const parentSatisfiesConditions = Object.entries(parent).every(([field, parentFilter]) =>
      checkValueSatisfiesCondition(parent[field], parentFilter),
    );
    if (!parentSatisfiesConditions) {
      return false;
    }
```

You can then delete the whole `checkOrganisationUnitParentValue` function!

kael89: I don't check for `parent` existence above - if it's `undefined` an error will be thrown. I believe it's a fair assumption since `World` should be the only org unit with no parent

kael89 (changes requested):
First part of my review, to make things easier for both of us!

I have reviewed everything except for `buildOrganisationUnitParentValue`, which I'll do in the second part

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
```diff
@@ -63,35 +115,87 @@ const sumDataValues = (analytics, dataValues, filter = {}) => {
...
+  const orgUnitCodes = [...new Set(analytics.map(a => a.organisationUnit))];
+  const parentIdMap = {};
+  const orgUnitParentMap = {};
+
+  const orgUnits = await models.entity.find({ code: orgUnitCodes });
+  const orgUnitParentIds = [...new Set(orgUnits.map(o => o.parent_id))];
+  const orgUnitParents = await models.entity.find({ id: orgUnitParentIds });
+  orgUnitParents.forEach(p => {
+    parentIdMap[p.id] = { code: p.code, type: p.type };
+  });
```
kael89, line 140: Better to keep all parent fields - allows for more filtering options in the future:

```suggestion
  const parentIdMap = keyBy(orgUnitParents, 'id');
```

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
```diff
@@ -63,35 +115,87 @@ const sumDataValues = (analytics, dataValues, filter = {}) => {
...
       ...a,
       value: checkValueSatisfiesCondition(a.value, config?.countCondition || '*') ? 1 : 0,
     })),
     dataValues,
     filter,
+    models,
   );
 };
 
+const buildOrganisationUnitParentValue = async (analytics, models) => {
```
kael89, line 130: In general, feel free to use `orgUnit` instead of `organisationUnit` - common abbreviations help making the code simpler, and we have decided that we prefer `orgUnit` over the full version

```suggestion
const buildOrgUnitParentMapForAnalytics = async (analytics, models) => {
```

kael89 (changes requested):
I am a bit concerned about using entity related logic in data builders:

1. As you point out in the PR description, we already have aggregations that map org units to their ancestors
2. We use some elaborate org unit retrieval logic (projects hierarchies etc) - see all the entity related logic in [web-config-server/Aggregator](https://github.com/beyondessential/tupaia/blob/dev/packages/web-config-server/src/aggregator/Aggregator.js). This knowledge is effectively lost here

After some thinking, I believe it's fine to go with the approach you followed here:
1. For good or for bad, I believe we already use the `entity` model to retrieve org unit hierarchy data in data builders. So, it's not like you broke this window 😉 Let's just make sure that this is the last resort!
2. Upcoming `entity-server` work will hopefully make retrieving entity related data both decoupled and easy!

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
```diff
@@ -212,10 +332,10 @@ export const getDataElementsFromCalculateOperationConfig = config =>
 export const calculateOperationForAnalytics = async (models, analytics, config) => {
   const { operator } = config;
   if (SINGLE_ANALYTIC_OPERATORS.includes(operator)) {
-    return performSingleAnalyticOperation(analytics, config);
+    return performSingleAnalyticOperation(analytics, config, models);
   }
   if (ARITHMETIC_OPERATORS.includes(operator)) {
-    return performArithmeticOperation(analytics, config);
+    return performArithmeticOperation(analytics, config, models);
```
kael89, line 338: One last thing: it's a common convention to pass dependencies first, then data. We also follow this convention with `models` - they are the first argument in most cases.

Let's also follow this convention here! There will be a very small regression risk for functionality in line 341 (`PERATORS[operator](analytics, config, models)`), but should be very easy to test

kael89 (changes requested):

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
nick-beyondessential: I hear you. The other consideration using aggregation on ancestors was that we want only immediate ancestors (parents) for aggregation. Which could be achieved by adding a parameter or creating a new aggregation type? Coupled with the aggregating on data values as well, climbing through the broken window seemed path of least pain.

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
```diff
@@ -41,20 +41,72 @@ const performSingleAnalyticOperation = (analytics, config) => {
...
+  return OPERATORS[operator](filteredAnalytics[0].value, config, models);
 };
 
-const sumDataValues = (analytics, dataValues, filter = {}) => {
-  let sum; // Keep sum undefined so that if there's no data values then we can distinguish between No data and 0
-  const { organisationUnit: organisationUnitFilter } = filter;
+const filterAnalytics = async (analytics, filter, models) => {
+  const {
+    organisationUnit: organisationUnitFilter,
+    parentType: organisationUnitParentType,
```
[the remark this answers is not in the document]
nick-beyondessential: Yes this makes sense and I should have organised it more generally, thanks for clearer suggestion.

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
```diff
@@ -41,20 +43,53 @@ const performSingleAnalyticOperation = (analytics, config) => {
...
-  analytics.forEach(({ dataElement, value, organisationUnit }) => {
     if (
       organisationUnitFilter &&
-      !checkValueSatisfiesCondition(organisationUnit, organisationUnitFilter)
+      !(
+        isPlainObject(organisationUnitFilter) &&
+        Object.keys(organisationUnitFilter).length === 1 &&
+        Object.keys(organisationUnitFilter)[0] === 'parent'
+      ) &&
+      !checkValueSatisfiesCondition(a.organisationUnit, organisationUnitFilter)
```
nick-beyondessential, line 80: Filtering out parent filter from organisationUnitFilter else missing 'operator' when it tries to check as parent filter is a child of organisationUnitFilter. Maybe a cleaner way to do this, could check on simple value (other reports currently only using simple code) but this allows eg exclude code

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
nick-beyondessential: Except in this case not all operators use models so models is an optional argument here?

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
nick-beyondessential: Maybe filter out to check function would be better i.e. `!isParentFilter(organisationUnitFilter)`

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
kael89: Oh I see. I'd probably still pass the models as the first argument and then just not use it in the operator utils, but up to you - I'm fine with it staying as is

packages/web-config-server/src/apiV1/dataBuilders/helpers/calculateOperationForAnalytics.js
kael89: There is:

```js
 const { parent: parentFilter, orgUnitFilter } = filter.organisationUnit || {};
```

And then you can simply check for existence of `parentFilter` and `orgUnitFilter` before the corresponding `checkValueSatisfiesCondition` calls. 

We could probably make this even simpler by refactoring `checkValueSatisfiesCondition` to skip checking if `condition` is `undefined`, but this involves a bit of regression risk so we can skip it for now.

nick-beyondessential: Ah yes, so parent: parentFilter does the job of filtering out parent from the filter. I like it.

kael89: (I just updated my comment above, I forgot the `...` on `orgUnitFilter`

kael89: > condition is undefined worries me a bit if it is a legitimate error.

Good point... Probably the reason we haven't gone this way

nick-beyondessential: `const { parent: parentFilter, ...orgUnitFilter } = filter.organisationUnit || {};` splits organisationUnit into an array of char if string. May have to break it out into a couple of steps.

kael89: Oh yes, you can change the approach according to the type of `filter.organisationUnit`

nick-beyondessential: Added check for empty condition object in `checkValueSatisfiesCondition` I think that is safe and a fair assumption.

kael89: Very good call, will simplify things here!

packages/utils/src/checkValueSatisfiesCondition.js
```diff
@@ -31,6 +32,9 @@ export const checkValueSatisfiesCondition = (value, condition) => {
     return (condition === ANY_VALUE_CONDITION && value !== '') || value === condition;
   }
 
+  // Handle empty condition Object {}
+  if (Object.keys(condition).length < 1) return true;
```
kael89, line 36: This module has unit tests, let's add a test case for this behaviour

kael89 (approved):
Looks good @nick-beyondessential, approved! I added a suggestion for a small unit test case addition

PR #2153, opened 2021-03-01
Make PG public and add FETP dahsboards to SB

nick-beyondessential:
### Changes:

- add PG to public access country list
- copy FETP dashboards to SB 
- remove STRIVE mapOverlays from explore


packages/database/src/migrations/20210301123426-AddFetpDashboardsToSolomonIs-modifies-data.js
```diff
@@ -0,0 +1,71 @@
...
+    await db.runSql(`
+    select * from "dashboardGroup" where 'fetp' = ANY("projectCodes") and "organisationUnitCode" IN ('PG', 'SB');
+  `)
+  ).rows;
+
+  const sbDashboardGroups = dashboardGroups
+    .filter(r => r.organisationUnitCode === 'SB')
+    .map(dg => dg.code);
+
+  const pgDashboardGroups = dashboardGroups.filter(r => r.organisationUnitCode === 'PG');
```
edmofro, line 32: Confusing that `pgDashboardGroups` contains the record, whereas `sbDashboardGroups` contains the codes. Would suggest aligning both to have the same data structure, or rename one

packages/database/src/migrations/20210301123426-AddFetpDashboardsToSolomonIs-modifies-data.js
```diff
@@ -0,0 +1,71 @@
...
+          'dashboardGroup',
+          { dashboardReports: dg.dashboardReports },
+          { code: sbCodeToCheck },
+        );
+
+      return insertObject(db, 'dashboardGroup', {
+        organisationLevel: dg.organisationLevel,
+        userGroup: dg.userGroup,
+        organisationUnitCode: 'SB',
+        dashboardReports: `{${dg.dashboardReports.join(',')}}`,
```
edmofro, line 49: Why is the `.join` required here but not in the `updateValues`?

packages/database/src/migrations/20210301123426-AddFetpDashboardsToSolomonIs-modifies-data.js
```diff
@@ -0,0 +1,71 @@
...
+        code: sbCodeToCheck,
+        projectCodes: `{${dg.projectCodes.join(',')}}`,
+      });
+    }),
+  );
+};
+
+exports.down = function (db) {
+  return db.runSql(`
+    update "dashboardGroup" set "dashboardReports" = '{}' where "code" in (${arrayToDbString(
```
edmofro, line 60: It confuses me that we'd leave dashboard groups with no reports at all in them, but given this is the `down` that won't be run, I'm not too worried

edmofro (approved):
Looks like it does what we want - just a couple of comments for clarity but nothing holding back approval :-)

packages/database/src/migrations/20210301123426-AddFetpDashboardsToSolomonIs-modifies-data.js
nick-beyondessential: I don't know why but `updateValues` seems to handle the js array conversion but `insertObject` was throughing a psql malformed array error.

packages/database/src/migrations/20210301123426-AddFetpDashboardsToSolomonIs-modifies-data.js
nick-beyondessential: Yes just reverting it to the pre-migration state, used mostly while I dev and test and a mild sense of completion.

packages/database/src/migrations/20210301123426-AddFetpDashboardsToSolomonIs-modifies-data.js
nick-beyondessential: Yep fair call, renamed to sbDashboardGroupCodes

packages/database/src/migrations/20210301123426-AddFetpDashboardsToSolomonIs-modifies-data.js
edmofro: And the current state is that the groups exist with no reports inside? Weird!

edmofro (approved):

packages/database/src/migrations/20210301123426-AddFetpDashboardsToSolomonIs-modifies-data.js
nick-beyondessential: Added when the FETP project was created before any reports created.

PR #2161, opened 2021-03-02
1534: Persist e2e test snapshots

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1534

See the card of a discussion on the snapshot handling logic


packages/web-frontend/cypress/scripts/testE2e/pushSnapshots.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const pushSnapshots = async (snapshotRepo, branch, snapshots) => {
+  const date = Date.now();
+  const updateBranch = `tupaiabot/snapshot-update/${branch}_${date}`;
+  await snapshotRepo.createBranchIfNotExists(branch);
+  await snapshotRepo.createBranch(updateBranch, branch);
+
+  const commitFiles = [{ path: SNAPSHOTS.pathInRepo, content: snapshots.toString() }];
+  await snapshotRepo.commitFilesToBranch(updateBranch, commitFiles, 'Update snapshots');
+  const prTitle = `${branch} snapshot update (${moment(date).utc().format()})`;
```
edmofro, line 18: Thanks for adding the branch to the PR title

packages/web-frontend/cypress/scripts/testE2e/pullSnapshots.js
```diff
@@ -0,0 +1,27 @@
...
+
+const selectPullBranch = async (snapshotRepo, branch) => {
+  const logger = getLoggerInstance();
+
+  if (await snapshotRepo.hasBranch(branch)) {
+    logger.info(`A matching snapshot branch was found, using '${branch}'`);
+    return branch;
+  }
+
+  const { default_branch: defaultBranch } = await snapshotRepo.getDetails();
```
edmofro, line 18: Oooh, clever having the default set by the repo

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
```diff
@@ -36,7 +37,8 @@ describe('Dashboard reports', () => {
 
       cy.visit(url);
       cy.wait('@report');
-      cy.findByTestId('enlarged-dialog').snapshotHtml({ name: 'html' });
+      cy.findByTestId('enlarged-dialog').snapshotHtml({ name: SNAPSHOTS.key });
+      cy.findByTestId('enlarged-dialog').snapshotHtml({ name: SNAPSHOTS.newKey });
```
edmofro, line 41: I don't really understand this, but maybe by the time I'm finished the review I will

packages/web-frontend/cypress/scripts/testE2e/Snapshots.js
```diff
@@ -0,0 +1,97 @@
...
+   * @param {Object<string, any>} snapshotTree
+   */
+  constructor(snapshotTree) {
+    this.snapshotTree = snapshotTree;
+  }
+
+  static import = path => {
+    const contents = fs.readFileSync(path, { encoding: 'utf-8' });
+    // Note: JS parsing is a bit naive here, ideally we should use a code parsing library
+    const snapshotJson = isJs(path) ? contents.replace('module.exports = ', '') : contents;
```
kael89, line 47: I don't expect that using a code parser would be too complicated here, but I skipped it because of work load. Happy to have another look in the near future

packages/devops/scripts/ci/validateBranchName.sh
```diff
@@ -22,6 +23,17 @@ function get_branch_name() {
     echo $branch_name
 }
 
+function check_name_is_not_reserved() {
```
kael89, line 26: @edmofro also sneaked this change in - not strictly related to the scope of this PR, more related to your recent work.

I thought we wouldn't want people accidentally updating the `e2e` instance by pushes, since it should be an auto-generated isntance

packages/web-frontend/cypress/scripts/testE2e/Snapshots.js
```diff
@@ -0,0 +1,97 @@
...
+      throw new Error('"Snapshots.equals" expects a Snapshots instance');
+    }
+    return this.toString() === snapshots.toString();
+  };
+
+  isEmpty = () => Object.keys(this.snapshotTree).length === 0;
+
+  extractSnapshotsByKey = (key, options = {}) => {
+    const extractedNodePaths = [];
+    this.extractionConfig = {
```
edmofro, line 70: Can we pass through `extractionConfig` as a third argument, rather than storing it on the instance?

packages/web-frontend/cypress/scripts/testE2e/testE2e.js
```diff
@@ -0,0 +1,90 @@
...
+  const gitHub = new GithubApi({ token: requireEnv('GITHUB_ACCESS_TOKEN') });
+  return RemoteGitRepo.connect(gitHub, SNAPSHOTS.repoUrl);
+};
+
+/**
+ * Runs e2e tests using the following steps:
+ * 1. Pull snapshots
+ * 2. Generate config
+ * 3. Run tests using the pulled snapshots
+ * 4. If new snapshots were captured during tests, push them
```
edmofro, line 53:
```suggestion
 * 4. If new snapshots were captured during tests, or differences in output found, push them
```

edmofro: From what I can see, this is correct? Generally we hope they will just be new snapshots, but sometimes they will be differences that are either legit or not legit, in which case they will be pushed for manual review right?

edmofro (changes requested):
All looks really good! The interactions with GitHub are particularly clean 🎉 

I have run out of time to work out why we are capturing the snapshots twice and why the renaming of the snapshot key needs to happen - could you explain that part?

packages/devops/scripts/ci/validateBranchName.sh
```diff
@@ -22,6 +23,17 @@ function get_branch_name() {
...
+    local $branch_name=$1
+
+    for reserved_name in ${RESERVED_NAMES[@]}; do
+        if [[ "$branch_name" == "$reserved_name" ]]; then
+            log_error "❌ Branch name is reserved"
```
edmofro, line 31: Might as well include `$branch_name` in the log

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
kael89: We use snapshots under the "standard" key `SNAPSHOTS.key` for comparison with existing snapshots, since they will be stored in the snapshot repo (and pulled) using the same key. Snapshot comparison is done on a nested key basis

At the same time, we want to store the new snapshots somewhere avoiding a comparison... That's why we use `SNAPSHOTS.newKey`, which serves as a "temp" variable. Eventually we will extract those snapshots after the tests run, to create the snapshot PR.

The cleanest and simplest way would be to store the new snapshots in a `newSnapshots.js` file, but unfortunately the library we use [@cypress-snapshot](https://www.npmjs.com/package/@cypress/snapshot) does not allow that.

kael89: I just realised I should capture the snapshots in reverse order!! We want to capture "new" snapshots first, since the "existing" snapshots may throw a error because of mismatch. I had done so in a past local commit... 🤔 

Also, I guess I should add a comment to make what's happening clearer

packages/web-frontend/cypress/scripts/testE2e/Snapshots.js
kael89: We definitely can! It's a matter of preference. I thought it would be redundant to just pass the same value over and over in all recursions.

I'm personally in the "a little state doesn't harm" school of thought, although I guess it's a bit unconventional these days. Very happy to change that here if you prefer!

packages/devops/scripts/ci/validateBranchName.sh
kael89: Good call! Certainly won't harm

packages/web-frontend/cypress/scripts/testE2e/testE2e.js
kael89: Most definitely! The truth is, "new" is probably a misnomer in this PR, a more accurate definition would be "snapshotsCapturedDuringLastTestRun". But couldn't find a short and elegant description for that 😆 

I should definitely update this comment though. Please let me know if you come up with a better term than "new" snapshtos!

packages/web-frontend/cypress/scripts/testE2e/Snapshots.js
kael89: A bit more explanation on my reasoning: when I say "redundant", I mean in relation to code simplicity (in function signatures), not to performance toll (which would negligible, if any). The tradeoff is: more complexity in the class state level for less complexity in the recursion logic (signature). Recursions are generally more complicated than average, so I thought the tradeoff is worth it.

kael89:
@edmofro All changes have been implemented in https://github.com/beyondessential/tupaia/pull/2161/commits/27972daeea7b4879ef5908b6417b8ef70ef5c1a9 apart from https://github.com/beyondessential/tupaia/pull/2161#discussion_r587089632 (pending your answer). I also added a boolean `push` option for the `testE2e` script in https://github.com/beyondessential/tupaia/pull/2161/commits/7e334043bc2de0827c55e2e301ba8f4e554c51a1 - found it handy at least locally

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
```diff
@@ -34,11 +34,16 @@ describe('Dashboard reports', () => {
...
       cy.wait('@report');
-      cy.findByTestId('enlarged-dialog').snapshotHtml({ name: SNAPSHOTS.key });
-      cy.findByTestId('enlarged-dialog').snapshotHtml({ name: SNAPSHOTS.newKey });
+
+      cy.findByTestId('enlarged-dialog').as('enlargedDialog');
+      // Capture and store the snapshot using the "new" key, to avoid comparison with existing snapshots.
+      // We want to store the new snapshots no matter what: a failed comparison would prevent that
+      cy.get('@enlargedDialog').snapshotHtml({ name: SNAPSHOTS.newKey });
+      // Then, use the "standard" key to trigger a comparison with existing snapshots.
+      // This way we check for regression
```
edmofro, line 45: Right, I get it now - cypress automatically does the comparison, and will throw away or mess with the snapshot taken under the existing key right?

packages/utils/src/script.js
```diff
@@ -0,0 +1,76 @@
...
+    .finally(() => {
+      if (cleanup) {
+        cleanup();
+      }
+    });
+};
+
+/**
+ * @typedef {Object} ScriptConfig
+ * @property {string} [command] Use '*' or '$0' to refer to the default command
```
kael89, line 52: The diff in this file is a bit funny: the changes essentially start here + a copyright header and an import at the top of the file (compare with `runScript.js`)

packages/web-frontend/cypress/scripts/testE2e/Snapshots.js
edmofro: I find it unexpected, so confusing because I have to put the pieces together. I agree that recursion is confusing, but I think this nice static config parameter shouldn't add much mental load, whereas understanding the method-specific state did for me. It may be a personal preference thing though, so will leave it up to you!

edmofro (approved):
LGTM!

packages/utils/src/script.js
edmofro: Yep I compared against the original :+1:

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
kael89: That's exactly right! The snapshot library could be a bit more flexible with were (and how) snapshots are stored - sound like an open source contribution chance 😉

packages/web-frontend/cypress/scripts/testE2e/Snapshots.js
kael89: No worries, I'll change! Your suggestion probably adheres to the principle of least surprise in the context of our codebase - and is certainly clean!

That was a bit of an experimental approach so I mainly wanted to understand your viewpoint 😉

edmofro (approved):

PR #2075, opened 2021-02-15
1726 FETP Graduate Area of Expertise map overlays

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1726

The requirements of this map overlay have been changed since this card was originally made, you can read through the comments for context if you like but long story short the type of map overlay this now creates is screenshotted in this comment (district level map overlay which sums answers to various questions submitted by `individual`'s in that district): https://github.com/beyondessential/tupaia-backlog/issues/1726#issuecomment-778932323

**Note:**
- This branch is branched from https://github.com/beyondessential/tupaia/pull/1939 because my map overlays use `binary` internal data (which has been changed by that branch so I branched off it so these new map overlays will work once that code is released)

PR #2168, opened 2021-03-03
1726 fetp grad area of expertise

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/1726

The requirements of this map overlay have been changed since this card was originally made, you can read through the comments for context if you like but long story short the type of map overlay this now creates is screenshotted in this comment (district level map overlay which sums answers to various questions submitted by `individual`'s in that district): https://github.com/beyondessential/tupaia-backlog/issues/1726#issuecomment-778932323

**Note:**
- I've made this PR before [here](https://github.com/beyondessential/tupaia/pull/2075) but closed it / deleted that branch because of merge conflicts that were much easier to solve by just making a new branch off dev and copying the migration over


chris-bes (approved):
LGTM!

packages/database/src/migrations/20210303025836-AddFETPGradAreaOfExpertiseOverlays-modifies-data.js
```diff
@@ -24,14 +24,12 @@ const BASE_OVERLAY = {
   userGroup: 'FETP Graduates',
   isDataRegional: true,
   measureBuilderConfig: {
-    aggregationType: 'SUM',
     entityAggregation: {
-      aggregationType: 'REPLACE_ORG_UNIT_WITH_ORG_GROUP',
       dataSourceEntityType: 'individual',
       aggregationEntityType: 'district',
```
chris-bes, line 29: Very minor but I think if we're removing `aggregationType` then we no longer need `aggregationEntityType` as well

chris-bes (approved):
LGTM! I just have 1 minor suggestion so pre approving

packages/database/src/migrations/20210303025836-AddFETPGradAreaOfExpertiseOverlays-modifies-data.js
avaek: Thanks @chris-bes! Oddly it seems to be required even with the `aggregationType` removed - the overlays don't work properly without it (or at least they don't work properly without it the way I've currently set them up)

chris-bes: Ahh sorry, you're right. I didn't realize that this overlay is shaded polygons by districts. All good!

PR #2178, opened 2021-03-04
2188 e2e instance data

edmofro:
### Issue #
https://github.com/beyondessential/tupaia-backlog/issues/2188

Manual steps required on release:
-[ ] Open `crontab -e` and add the following line, which will dump `survey_response` and `answer` to S3 every night at 12:30pm UTC (currently 11:30pm in Melb, 10:30pm after daylight savings):
```
30 12 * * * /home/ubuntu/tupaia/packages/devops/scripts/dumpE2eDataToS3.sh | while IFS= read -r line; do printf '\%s \%s\n' "$(date)" "$line"; done >> /home/ubuntu/logs/dump_e2e_data.txt
```

### DIFF
--- a/packages/devops/scripts/deployment/dumpE2eDataToS3.sh
+++ b/packages/devops/scripts/deployment/dumpE2eDataToS3.sh
@@ -14,10 +14,10 @@ if [[ "$STAGE" == "e2e" ]]; then
   mkdir -p dumps
 
   echo "Dumping survey_response"
-  pg_dump -U tupaia --column-inserts --data-only --table=survey_response tupaia > dumps/survey_response.sql
+  psql -U tupaia tupaia -c "\copy survey_response to 'dumps/survey_response.csv' csv header"
 
   echo "Dumping answer"
-  pg_dump -U tupaia --column-inserts --data-only --table=answer tupaia > dumps/answer.sql
+  psql -U tupaia tupaia -c "\copy answer to 'dumps/answer.csv' csv header"
 
   echo "Compressing dumps"
   tar cfz dumps.tgz dumps/*
--- a/packages/devops/scripts/deployment/dumpE2eDataToS3.sh
+++ b/packages/devops/scripts/deployment/dumpE2eDataToS3.sh
@@ -14,10 +14,10 @@ if [[ "$STAGE" == "e2e" ]]; then
   mkdir -p dumps
 
   echo "Dumping survey_response"
-  psql -U tupaia tupaia -c "\copy survey_response to 'dumps/survey_response.csv' csv header"
+  psql -U tupaia tupaia -c "\copy survey_response to 'dumps/survey_response.csv' delimiter ',' csv header"
 
   echo "Dumping answer"
-  psql -U tupaia tupaia -c "\copy answer to 'dumps/answer.csv' csv header"
+  psql -U tupaia tupaia -c "\copy answer to 'dumps/answer.csv' delimiter ',' csv header"
 
   echo "Compressing dumps"
   tar cfz dumps.tgz dumps/*
--- a/packages/devops/scripts/deployment/dumpE2eDataToS3.sh
+++ b/packages/devops/scripts/deployment/dumpE2eDataToS3.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+# Run directly from a nightly cron job, so needs full setup of path etc.
+
 # Set PATH to include depencencies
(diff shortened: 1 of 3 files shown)


packages/devops/scripts/deployment/dumpE2eDataToS3.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
edmofro, line 1: This will be run every night, to push data from the baseline e2e instance to S3

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
edmofro, line 1: This will be run on startup, to restore e2e test data for any deployment with a branch ending in `e2e`

packages/devops/scripts/deployment/dumpE2eDataToS3.sh
```diff
@@ -0,0 +1,35 @@
...
+# Set the branch based on STAGE
+if [[ "$STAGE" == "e2e" ]]; then
+  echo "Dumping E2E data from E2E baseline instance to S3"
+  mkdir -p dumps
+
+  echo "Dumping survey_response"
+  psql -U tupaia tupaia -c "\copy survey_response to 'dumps/survey_response.csv' delimiter ',' csv header"
+
+  echo "Dumping answer"
+  psql -U tupaia tupaia -c "\copy answer to 'dumps/answer.csv' delimiter ',' csv header"
```
kael89, line 20: Perhaps we should also add `dhis_sync_queue`, `dhis_sync_log`?

kael89 (changes requested):
Amazing!! 🎉 Just one suggestion and one change

Also, we can change the base branch of this to `e2e-tests-mvp`

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo "Restoring survey responses"
+psql -U tupaia tupaia -c "\copy survey_response from 'dumps/survey_response.csv' delimiter ',' csv header"
+
+echo "Restoring answers"
+psql -U tupaia tupaia -c "\copy answer from 'dumps/answer.csv' delimiter ',' csv header"
+
+echo "Re-enabling notification triggers"
+psql -U tupaia tupaia -c "ALTER TABLE survey_response DISABLE TRIGGER survey_response_trigger;"
+psql -U tupaia tupaia -c "ALTER TABLE answer DISABLE TRIGGER answer_trigger;"
```
kael89, line 26:
```suggestion
psql -U tupaia tupaia -c "ALTER TABLE survey_response ENABLE TRIGGER survey_response_trigger;"
psql -U tupaia tupaia -c "ALTER TABLE answer ENABLE TRIGGER answer_trigger;"
```

kael89 (changes requested):

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
edmofro: 😬 
good catch!!

packages/devops/scripts/deployment/dumpE2eDataToS3.sh
edmofro: Yeah great point! I did think about doing this, but then wondered if there were other tables (like `ms1_sync_queue` etc.) and decided it might get out of hand for no real benefit. It will result in a somewhat corrupted db state, but only tables that won't be used for e2e at this stage, and keeps the process simple. What do you think?

kael89: I agree with the overall approach! Just for those tables in specific, there may be data sync leftovers in prod from last night - I guess we wouldn't want both `dev` and `e2e` trying to sync those

edmofro: No but we actually already have something in place to prevent that:
```
      // don't sync changes that happened before the server was started, to avoid multiple
      // deployments syncing identical changes that came from the overnight snapshot
      criteria.change_time = {
        comparator: '>',
        comparisonValue: SERVER_START_TIME,
      };
```

kael89: Great!!!

kael89 (approved):
🚀

edmofro:
@kael89 can I leave coordination of merging and applying that manual change over to you? Note that there is an addition to `startup.sh` which won't take effect until it has been in production for a day (that old problem)

PR #2180, opened 2021-03-04
1707: Entity-Server scaffolding

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1707:

Basic scaffolding for entity-server. Requirements:
- Basic test route which returns hello world
- User authentication via auth token
- CI/CD processes in place (can deploy a feature instance and have entity server running and accessible)
- Basic test framework

Currently I've done all of the steps outlined here to be completed prior to releasing: https://docs.beyondessential.com.au/books/dev-ops/page/steps-for-releasing-a-new-primary-package

Will perform release steps when releasing next week


packages/entity-server/src/@types/express/index.d.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Authenticator } from '@tupaia/auth';
+
+declare global {
+  namespace Express {
+    export interface Request {
+      accessPolicy: AccessPolicy;
+      authenticator: Authenticator;
+    }
+  }
+}
```
rohan-bes, line 15: This is how we can use declaration merging to add tupaia specific fields to the express `Request` type

packages/entity-server/src/auth/middleware.ts
```diff
@@ -0,0 +1,68 @@
+/**
```
rohan-bes, line 1: Still a lot of duplication in this file with `packages/report-server/src/auth/middleware.ts`

Personally I think it acceptable as most of this code is sort of boilerplate and there isn't much complex logic that is being duplicated, however happy to get push back on this.

packages/entity-server/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "forceConsistentCasingInFileNames": true,
+    "module": "CommonJS",
+    "noEmit": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "ES2019",
+    "typeRoots": ["src/@types", "node_modules/@types"]
+  },
+  "exclude": ["node_modules", "src/@types"]
+}
```
rohan-bes, line 14: Will migrate this into common `tsconfig-ts.dependency.json` as part of https://github.com/beyondessential/tupaia-backlog/issues/2400

Didn't want to make too many cross package changes in this PR

packages/entity-server/src/app/addRoutesToApp.ts
```diff
@@ -0,0 +1,49 @@
...
+) => {
+  // use default if response is already being sent and there are connection problems
+  if (res && res.headersSent) {
+    next(err);
+    return;
+  }
+
+  const error = 'respond' in err ? err : new InternalServerError(err);
+  error.respond(res);
+};
```
edmofro, line 35: The functions above might get included in a new package `@tupaia/server-boilerplate` that @tcaiger has been working on in https://github.com/beyondessential/tupaia/pull/2144

Which ever gets in first - can you guys coordinate to make sure these are ultimately pulled in from the shared util?

packages/entity-server/src/auth/middleware.ts
edmofro: Could move into `@tupaia/server-boilerplate` once @tcaiger's PR hits dev

packages/report-server/src/auth/middleware.ts
```diff
@@ -40,14 +40,7 @@ const authenticateBearerAuthHeader = async (authHeader: string) => {
...
 const authenticateBasicAuthHeader = async (req: ReportsRequest, authHeader: string) => {
-  const usernamePassword = Buffer.from(authHeader.split(' ')[1], 'base64').toString();
-  if (!usernamePassword.includes(':')) {
-    throw new UnauthenticatedError('Invalid Basic auth credentials');
-  }
-
-  // Split on first occurrence because password can contain ':'
-  const username = usernamePassword.split(':')[0];
-  const password = usernamePassword.substring(username.length + 1, usernamePassword.length);
+  const { username, password } = getUserAndPassFromBasicAuth(authHeader);
```
edmofro, line 43: The other option would be to use `basic-auth` package like we do in `meditrak-server` (see `clientAuth.js`)

edmofro: But I'm kinda into rolling our own for something so straightforward, so actually how about instead updating `meditrak-server` to use yours?

edmofro (approved):
All looking good! Would be great if you can work with @tcaiger to identify which boilerplate can be pulled out, but that can happen later as it might be tricky to sequence things

packages/entity-server/src/auth/middleware.ts
rohan-bes: I agree!

edmofro (approved):
Nice tests too (especially the `:` one)

rohan-bes:
@edmofro wondering if there'd be any problems with me merging this to `dev`? Based on what's written in https://docs.beyondessential.com.au/books/dev-ops/page/steps-for-releasing-a-new-primary-package I think it should be safe? Or do I need to ssh into `dev` while the build is running and manually set up the pm2 process?

In terms of testing I think this is largely 'No Testing Required'. The one exception being regression testing the meditrak-server api-clients to ensure their basic auth credentials are still valid. In the database we have a couple of test entries and then entries for:
- psss-sever: I checked this, works correctly 👍 
- josh@sussol.net: Not sure what/who this is? Is this client still in use?

edmofro:
- josh@sussol is the mSupply client, which pushes immunisation inventory and fridge temperature data to us

You shouldn't need to manually set up the pm2 process no, but I may be missing something without reading through that doc on Tamanu day, so just see how you go ;-)

PR #2182, opened 2021-03-04
only load user page urls when no project is being loaded

tcaiger:
### Issue #: [2402](https://github.com/beyondessential/tupaia-backlog/issues/2402) Fix user page urls

This is a hotfix but I think it needs to be tested before being released so I have made it from the dev branch and not the release branch.

### Changes:

- Fix user page urls

---


rohan-bes (approved):
Looks good!

PR #2184, opened 2021-03-04
2113: Update psss weekly cases indicators to support daily data

chris-bes:
### Changes:

- Update psss weekly cases indicators to support daily data


kael89 (approved):
Perfection! I absolutely love this PR! 😍

PR #1959, opened 2021-01-27
885-2143: Hotfix missing HP05 and SCRF data

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/885
https://github.com/beyondessential/tupaia-backlog/issues/2143

PR #2187, opened 2021-03-05
2257 add missing cases to sync queue

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/2257

Note: This just adds the missing entities to the sync queue
- It might be worth changing the priority of the survey responses associated with these cases in this migration too

This is based on a somewhat similar migration that @kael89 did [here](https://github.com/beyondessential/tupaia/pull/1959/files)
- One difference to note - the linked PR changes the priority of the records it adds to the sync queue (to avoid clogging the queue) whereas mine does not. The reason I chose not to do this is because there are only 129 records being added which wouldn't take too long to get through


edmofro (approved):
lgtm, nice one! 

I think it's reasonable to just leave the survey responses to make their way through the queue a few days after release, given they've been missing for months. However, if you're interested and not too busy, selecting the `survey_response` records with `entity_id` matching the ids of the entities being resynced shouldn't be too difficult.

avaek:
Thanks @edmofro - I've updated the migration to also change the priority for the relevant survey responses

packages/database/src/migrations/20210305010124-AddMissingCasesToSyncQueue-modifies-data.js
```diff
@@ -29,18 +29,51 @@ const createSelectCasesQuery = columns => {
 };
 
 const selectCasesForResync = async db => {
-  const query = createSelectCasesQuery('*');
+  const query = createSelectCasesQuery();
```
edmofro, line 32: Kostas' version used a separate function to create the query because he was reusing parts of the query in multiple places. In your PR, you're using each query just once, so just put it here rather than calling a function to create it.

FYI if you were using it multiple places, but the function took no parameters (as it does now) and immediately returned the string, you'd declare that string as a `const` instead.

packages/database/src/migrations/20210305010124-AddMissingCasesToSyncQueue-modifies-data.js
```diff
@@ -0,0 +1,88 @@
...
+      type = 'update' AND
+      record_id IN (${arrayToDbString(surveyResponseIds)})`);
+
+exports.up = async function (db) {
+  const changeChannel = new DatabaseChangeChannel();
+
+  try {
+    // n.b. this requires a meditrak-server instance to be running and listening for the changes
+    const cases = await selectCasesForResync(db);
+    const surveyResponses = await selectSurveyResponsesForResync(db);
```
edmofro, line 73:
```suggestion
    const surveyResponses = await selectSurveyResponsesForResync(db, cases.map(c => c.id));
```

packages/database/src/migrations/20210305010124-AddMissingCasesToSyncQueue-modifies-data.js
```diff
@@ -0,0 +1,88 @@
...
+        e.type = 'case'
+    );
+`;
+};
+
+const selectSurveyResponsesForResync = async db => {
+  const query = createSelectSurveyResponsesQuery();
+  const { rows } = await db.runSql(query);
+  return rows;
+};
```
edmofro, line 55:
```suggestion
const selectSurveyResponsesForResync = async (db, entityIds) => {
  const query = `
    SELECT * FROM survey_response
    WHERE entity_id in ${arrayToDbString(entityIds)}
`;
  const { rows } = await db.runSql(query, entityIds);
  return rows;
};
```

edmofro (changes requested):
Nice, all logically makes sense! Suggestions just to simplify and optimise it - you've already fetched the entity ids from the db, so no need to make poor old postgres do it again ;-)

avaek:
Thanks for the optimisations @edmofro, very clean & elegant - I've applied them & pushed the changes to this PR (although I guess I should have made the changes individually & committed them to each of your suggestion comments to resolve them, I haven't used that feature before & only noticed it once I'd already committed the changes)

edmofro (approved):
No no need to apply the suggestions individually, you did it right! And you might not have corrected my errors if you'd used the Github suggestion commit function 😁

PR #2191, opened 2021-03-05
Hot fix laos dashboards crash

chris-bes:
### Changes:

- Return data empty array


kael89 (approved):
Classic!!

kael89:
@chris-bes @edmofro  I believe we should refactor the front-end to not crash when `data` is not included in the response... I've seen this error happening too many times, and doesn't seem like a good "coal mine canary" - in all the cases that I bumped into this, it was errors like the one here - people forgetting to include `data` in the response. None of them pointed to an error in the core data builder logic - where a crash would be actually valuable.


Let's discuss this next week!

PR #2192, opened 2021-03-05
Fix: Remove tupaia_read from testDataDump

kael89:
### Issue #:
[No issue]


kael89:
@IgorNadj We don't need  the `tupaia_read`  role in the test db...

IgorNadj (approved):

PR #2197, opened 2021-03-09
1535: Use specific date range for e2e tests

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1535


packages/web-frontend/cypress/scripts/generateConfig/generateReportConfig.js
```diff
@@ -69,7 +100,7 @@ const selectOrgUnitCode = async (database, orgUnitCodes, entityConditions) => {
 };
 
 const selectUrlParams = async (database, report, dashboardGroups) => {
-  const { viewJson } = report;
+  const viewJson = report.viewJson || {};
```
edmofro, line 103: In case you're interested, the following is also possible, though I find it unclear and prefer your syntax
```suggestion
  const { viewJson = {} } = report;
```

kael89: I am familiar with this syntax (and I generally like it 😃 ) I didn't go this way here, because `viewJson` may be `null`

edmofro: Ah true!

edmofro (approved):
All lgtm!

PR #2172, opened 2021-03-03
1622: Added samoa confirmed cases overlay

chris-bes:
### Changes:
- Added samoa confirmed cases overlay

PR #2199, opened 2021-03-09
1622: Added samoa confirmed cases overlay

chris-bes:
### Changes:
- Added samoa confirmed cases overlay


chris-bes:
Made from https://github.com/beyondessential/tupaia/pull/2172 because the old one had a lot of conflicts with dev

packages/database/src/migrations/20210219055420-SamoaHeatmapHomeVillageOfConfirmedCases-modifies-data.js
```diff
@@ -24,6 +24,13 @@ const indicator = {
     formula:
       "WS_COVID19_Clinsurv_137 == 1 or equalText(WS_COVID19_Clinsurv_75, '1. Positive') or equalText(WS_COVID19_Clinsurv_79, '1. Positive') or equalText(WS_COVID19_Clinsurv_83, '1. IgG positive only') or equalText(WS_COVID19_Clinsurv_83, '2. IgM positive only') or equalText(WS_COVID19_Clinsurv_83, '3. IgM / IgG positive') or equalText(WS_COVID19_Clinsurv_87, '1. Positive')",
     programCode: 'SC1CS',
+    defaultValues: {
+      WS_COVID19_Clinsurv_137: 0,
+      WS_COVID19_Clinsurv_75: 'undefined',
+      WS_COVID19_Clinsurv_79: 'undefined',
+      WS_COVID19_Clinsurv_83: 'undefined',
+      WS_COVID19_Clinsurv_87: 'undefined',
+    },
```
chris-bes, line 33: The formula to count the cases is:
```
WS_COVID19_Clinsurv_137 (Clinical criteria met) = Yes
OR
WS_COVID19_Clinsurv_75 (NP swab result) = Positive
OR
WS_COVID19_Clinsurv_79 (OP swab result) = Positive
OR
WS_COVID19_Clinsurv_83 (Blood test result) = 1., 2. or 3.
OR
WS_COVID19_Clinsurv_87 (Sputum sample result) = Positive
```

The way the indicator builder work is if not all data values are provided, the event will be skipped. So I think we have to provide default values for all of them, just so that the event will not be skipped in case only 1 of the questions is answered (eg: only `WS_COVID19_Clinsurv_137` is answered with Yes and the other questions are skipped by the user)

PR #2200, opened 2021-03-09
1537: e2e tests should throw for empty reports

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1537

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -41,24 +41,7 @@
     },
     // Packages using jest
     {
-      "files": [
-        "packages/access-policy/**",
-        "packages/aggregator/**",
-        "packages/auth/**",
-        "packages/data-api/**",
-        "packages/data-broker/**",
-        "packages/dhis-api/**",
-        "packages/expression-parser/**",
-        "packages/entity-server/**",
-        "packages/indicators/**",
-        "packages/meditrak-app/**",
-        "packages/psss/**",
-        "packages/psss-server/**",
-        "packages/report-server/**",
-        "packages/ui-components/**",
-        "packages/utils/**",
-        "packages/web-frontend/**"
-      ],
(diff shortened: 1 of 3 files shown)


packages/web-frontend/cypress.json
```diff
@@ -5,6 +5,7 @@
   "projectId": "pty5cc",
   "requestTimeout": 30000,
   "responseTimeout": 60000,
+  "tupaia_requireNonEmptyVisualisations": false,
```
kael89, line 8: I used the `tupaia_` prefix to namespace our cypress options

kael89: As you can see, this option is set to `false`. The plan is:

1. Do not use this option initially, since a lot of reports will be empty
2. Manually update this option on a server and invoke the `test-e2e` script. Use the results to submit test data
3. After all reports non-empty, set this flag to `true`, making it the default behaviour moving forward

I could devise a more use friendly way of setting this off and on, but I decided the above plan is simpler and meets our current requirements.

.eslintrc
```diff
@@ -41,24 +41,7 @@
...
-        "packages/indicators/**",
-        "packages/meditrak-app/**",
-        "packages/psss/**",
-        "packages/psss-server/**",
-        "packages/report-server/**",
-        "packages/ui-components/**",
-        "packages/utils/**",
-        "packages/web-frontend/**"
-      ],
+      "files": ["packages/**/__tests__/**", "packages/web-frontend/src/tests/**"],
```
edmofro, line 44: Great simplification!

edmofro (approved):
Also a good example of validating expected behaviour outside of snapshot comparison :-)

lgtm, bring on e2e!!!!

PR #2205, opened 2021-03-10
2153-map-overlay-components-part-1

tcaiger:
### Changes:

- Add MapContainer and TileLayer components

---


packages/ui-components/src/components/Map/constants.js
```diff
@@ -0,0 +1,32 @@
...
+// Note: There's a little bit of a hack going on here, the bounds[0] for explore are actually [6.5, 110]
+// However in order to trigger the map to re-render we set them slightly adjusted as [6.5001, 110]
+// See: https://github.com/beyondessential/tupaia-backlog/issues/540#issuecomment-631314721
+export const DEFAULT_BOUNDS = [
+  [6.5001, 110],
+  [-40, 204.5],
+];
+
+export const MAPBOX_TOKEN =
+  'pk.eyJ1Ijoic3Vzc29sIiwiYSI6ImNqNHMwOW02MzFhaGIycXRjMnZ1dXFlN2gifQ.1sAg5w7hYU7e3LtJM0-hSg';
```
rohan-bes, line 16: Initially I thought that this may be a big security issue, however I took a look at the Mapbox docs and it seems it's fine (afterall a user can just find this token by inspecting the outbound requests from tupaia.org to mapbox).

Their docs did however mention that it would be good to make use of their token rotation feature. I'm not suggesting  we do that now but might be worth putting plans in place for it?

https://docs.mapbox.com/help/troubleshooting/how-to-use-mapbox-securely/#access-tokens (under Rotate Tokens)

rohan-bes (approved):
Looks good to me 😄

packages/ui-components/src/components/Map/constants.js
tcaiger: hmmm that's a good point. I'll make a card to capture that.

PR #2208, opened 2021-03-10
2386: Architecture for entity-server

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2386 and https://github.com/beyondessential/tupaia-backlog/issues/2334:
- Implemented basic `/:projectCode/:entityCode[?fields=[code,country_code,etc.]]` endpoint
- Made Route class generic to allow specifying custom Request and Response objects for each route
- Added middleware to handle checking permissions and converting entity models into response objects

*Unit tests incoming!!*


packages/entity-server/src/app/addRoutesToApp.ts
```diff
@@ -5,15 +5,14 @@
...
 import { authenticationMiddleware } from '../auth';
+import { attachEntityToRequest, attachFormatterToResponse } from '../entity/middleware';
 
-const handleWith = (RouteClass: typeof Route) => (
-  req: Request,
-  res: Response,
-  next: NextFunction,
-) => {
+const handleWith = <Req extends Request, T extends Route<Req>>(
+  RouteClass: new (req: Req, res: Response, next: NextFunction) => T,
```
rohan-bes, line 14: Please don't be scared by this syntax, it means that the RouteClass will be a function that if you call `new` upon with the arguments provided then it will return an object of type `T`.

Speaking of type `T`, what are your thoughts @kael89 on using `T` here as the type alias (as opposed to something like `RouteType` or even `R`)?

packages/entity-server/src/entity/Entity.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { EntityModel as BaseEntityModel } from '@tupaia/database';
+
+export interface EntityFields {
+  readonly code: string;
+  readonly country_code: string;
+}
+
+export interface EntityModel extends BaseEntityModel, EntityFields {
+  isProject: () => boolean;
```
rohan-bes, line 15: Unfortunately typescript can't infer the `isProject()`, etc. functions on the `EntityModel` class... I think because they're actually attached to the `EntityType` class the database model manually exposes them... oh well 🤷

packages/entity-server/src/entity/middleware.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const mapEntitiesToFields = (fields: (keyof EntityFields)[]) => (model: EntityModel) => {
+  const mappedEntity: { -readonly [field in keyof EntityFields]?: EntityFields[field] } = {};
+  fields.forEach(field => {
+    mappedEntity[field] = model[field];
+  });
+  return mappedEntity;
+};
+
+const validFields: (keyof EntityFields)[] = ['code', 'country_code'];
```
rohan-bes, line 81: This is rather handly as TS will throw an error here if we try to add a string that isn't a key of `EntityFields` 👍 

Technically there's another syntax that will produce an array of strings of all the keys of `EntityFields` however I opted not to use it for a couple reasons:
- It doesn't read as clearly
- We may wish to only support a subset of the `EntityFields` (I suspect this is likely)

packages/entity-server/src/index.ts
```diff
@@ -19,7 +20,7 @@ dotenv.config(); // Load the environment variables into process.env
 /**
  * Set up models for Authenticator
  */
-const models = new ModelRegistry(new TupaiaDatabase());
+const models = new ModelRegistry(new TupaiaDatabase()) as EntityServerModelRegistry;
```
rohan-bes, line 23: Unfortunately resorted to typecasting here...

I tried instantiating a class that extended `ModelRegistry` and only had the `entity` field declared however that ended up with `entity` being `undefined`... not exactly sure what went wrong there, perhaps worthy of more investigation...

packages/entity-server/src/routes/Route.ts
```diff
@@ -5,14 +5,16 @@
 import { Request, NextFunction, Response } from 'express';
 import { respond } from '@tupaia/utils';
 
-export class Route {
-  req: Request;
+type ResBody<Type> = Type extends Response<infer Body> ? Body : null; // Infers body of response type
 
-  res: Response;
+export class Route<Req extends Request = Request, Res extends Response = Response> {
```
rohan-bes, line 10: Made the base `Route` class generic so that we can specify what Request/Response we expect from each route

packages/entity-server/src/routes/Route.ts
```diff
@@ -34,7 +36,7 @@ export class Route {
     }
   }
 
-  async buildResponse(): Promise<Record<string, unknown>> {
+  async buildResponse(): Promise<ResBody<Res>> {
```
rohan-bes, line 39: This is pretty cool, it means that Typescript can infer the return type of this function based on the Response type for each route, now we get proper errors if we return an unexpected response object 👍

packages/entity-server/src/routes/requests.ts
```diff
@@ -0,0 +1,31 @@
...
+interface FetchEntityQuery extends Query {
+  fields?: string;
+}
+
+export type EntityResponseObject = {
+  [field in keyof EntityFields]?: EntityFields[field];
+};
+
+// eslint-disable-next-line @typescript-eslint/comma-dangle
+export interface FetchEntityRequest<
```
rohan-bes, line 24: Might rename  some of these in the future to have more generic names...

packages/entity-server/src/routes/requests.ts
```diff
@@ -0,0 +1,31 @@
...
+  P = FetchEntityParams,
+  ResBody = EntityResponseObject,
+  ReqBody = Record<string, never>,
+  ReqQuery = FetchEntityQuery
+> extends Request<P, ResBody, ReqBody, ReqQuery> {
+  context: Context<{ entity: EntityModel }>;
```
rohan-bes, line 30: @kael89 This is the solution I found to having the properties of the context object being forced to be optional...
By making `Context` a generic type, each request can specify the context it expects to work with 😄

packages/entity-server/src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export { EntityModel, EntityFields } from './entity';
+export { FetchEntityRequest, FetchEntityResponse, EntityResponseObject } from './routes';
+
+export interface EntityServerModelRegistry extends ModelRegistry {
+  readonly entity: EntityModel;
+}
+
+export type Context<T> = {
+  [field in keyof T]: T[field];
```
edmofro, line 18: The result of this looks like a type that is just the same as `T`?

packages/entity-server/src/routes/Route.ts
edmofro: Nice, that is quite cool. I must say the Typescript syntax is taking some getting used to though! It's powerful, but that power comes through abstractions that I still need to learn

packages/entity-server/src/entity/Entity.ts
edmofro: Hmmm...I don't think the model _does_ have `isProject()` etc. functions! 

Basically the model is more of a factory, and produces instances of `EntityType` - _those_ are the bit with the instance specific functions.

Do you want to talk about this on a quick call?

packages/entity-server/src/app/addRoutesToApp.ts
kael89: While I'm still not decided on the single vs multi letter generics dilemma, I think I'm leaning towards the single letter ones now (could be a matter of habit).

I remember I felt a bit confused when I was once reading `jest` types and they were using `E`, `T`, `R` generics. But after you get familiar with the context (`E` => expected, `R` => received etc), single letter generics probably make things simpler.

Overall, `T` seems like a good choice. Probably  better than `R` in this case, for the reason that both `Request` and `Response` start with `R`

packages/entity-server/src/entity/middleware.ts
```diff
@@ -0,0 +1,101 @@
...
+  accessPolicy: AccessPolicy,
+) => {
+  // Assume user always has access to all world items.
+  if (entity.code === 'World') {
+    return true;
+  }
+  // Timor-Leste is temporarily turned off
+  if (entity.country_code === 'TL') {
+    return false;
+  }
```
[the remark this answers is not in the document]
edmofro: For sure, I'd prefer to remove both!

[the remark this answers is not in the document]
edmofro: `TL` actually may not be redundant, but certainly should be handled in Tupaia rather than here. Basically no one should even know that `TL` is in our system, no matter what their permissions, for political reasons, but we still wanted to hold on to all of the data for it. That's gone on for long enough that I'd prefer now to take a backup and get rid of the specific handling. Could you make a card for that please 🙏

packages/entity-server/src/entity/middleware.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (validateField(requestedField)) {
+      fields.add(requestedField);
+    } else {
+      throw new Error(`Unknown field requested: ${requestedField}`);
+    }
+  });
+  return Array.from(fields);
+};
+
+const validateField = (field: string): field is keyof EntityFields => {
```
edmofro, line 99: Mostly we try to declare functions before they're used, even if that isn't technically required

packages/entity-server/src/routes/requests.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
edmofro, line 1: The grouping of `requests` and `responses` might make more sense when there are more, and it's more obvious what the shared elements are, but I would have instinctively grouped them together, and put that in a file alongside the specific route they related to. So you'd have a directory structure more like `routes/FetchEntity/FetchEntityRoute.ts`, `routes/FetchEntity/interfaces.ts`. What do you think about that? You've probably worked through other routes and the way you've got it comes out better?

edmofro (changes requested):
Looking really good, and some super clever ts stuff in there! Too clever for me to understand some of the time 😂  I get there in the end though...

I'm thinking now would be the best time to pull in the (now in dev) `server-boilerplate` stuff, because otherwise, with the changes you're making, I see them diverging a lot and becoming hard to pull together later 🙏

packages/entity-server/src/app/addRoutesToApp.ts
kael89: I'd go with `T`

edmofro: Yeah sorry I added that comment before you responded, but only just submitted the review as a whole - your thoughts above are great!

packages/entity-server/src/types.ts
rohan-bes: You have indeed, I'm worried I'm doing something a bit silly here so if you can see a better way please let me know.

Ideally what we'd be doing here is making context's type itself be a generic parameter of the Request type, ie:
`Request<P = ParamsDict, ..., Context = any>` 

However, you can't extend the number of generic type parameters of an existing type when using declaration merging (which is what you're meant to do to add fields to the Express.Request type).

My solution was to use generics with the `Context` type so that it basically becomes whatever you pass into it, a bit funky admittedly but seems to get the job done.

packages/entity-server/src/app/createApp.ts
```diff
@@ -9,14 +9,14 @@ import bodyParser from 'body-parser';
...
-import { ModelRegistry } from '@tupaia/database';
+import { EntityServerModelRegistry } from '../types';
 
 import { addRoutesToApp } from './addRoutesToApp';
 
 /**
  * Set up express server with middleware,
  */
-export function createApp(models: ModelRegistry) {
+export function createApp(models: EntityServerModelRegistry) {
```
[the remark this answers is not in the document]
rohan-bes: Ah good point, I think you're right I'll change that 👍

packages/entity-server/src/entity/middleware.ts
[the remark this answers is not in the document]
rohan-bes: Created: https://github.com/beyondessential/tupaia-backlog/issues/2456

packages/entity-server/src/entity/middleware.ts
rohan-bes: Ah good point, will fix

packages/entity-server/src/routes/Route.ts
rohan-bes: Yeah I feel a bit of a similar way, most of the syntax is pretty unfamiliar to me and it's a bit daunting. The plus side is that when you finally get it right, vscode plays so nicely with it that I find it really re-assuring 😄

packages/entity-server/src/routes/requests.ts
rohan-bes: Tbh most of the design for this I put in place assuming that we'd have a few different types of routes/requests (as we had originally planned with the `/:hierarchyCode/ancestors` and `/:hierarchyCode/descendants` endpoints).

Now that we've consolidated our endpoint design into a single structure (which I love btw) I might refactor all this to be a bit more straightforward. I think we still want a different class for each route, so I'll probs go with either:
defining them inline in `routes/FetchEntity/FetchEntityRoute.ts` (renamed to `routes/hierarchy/SingleEntityRoute.ts`)
defining them in a `types.ts` file under `routes/hierarchy/`

rohan-bes:
@edmofro I just pushed up some changes which addressed a number of the issues you raised in your first review. Right now I'm working on refactoring `entity-server` to use `server-boilerplate`. There may be a bunch of code changes to come out of that... I'm planning on making a separate PR for that

packages/entity-server/src/app/createApp.ts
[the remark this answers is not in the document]
rohan-bes: In the end I went with leaving the full registry on as we need at least the `Project` model as well as the `Entity` model, and I thought it might be nice to have access to all the utility methods on the `ModelRegistry` class (tho I'm not sure if we'll be actually using them...)

packages/entity-server/src/models/Project.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Model } from './types';
+
+export interface ProjectFields {
+  readonly code: string;
+  readonly entity_hierarchy_id: string;
+}
+
+export interface ProjectType extends BaseProjectType, ProjectFields {}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
```
rohan-bes, line 16: I keep adding this but I should just make it a rule in our linting rules. I think it's very common to do this when you want to create a specific subtype with set generic arguments, `Express` seems to do it a lot...

packages/entity-server/src/app/createApp.ts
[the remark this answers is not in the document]
edmofro: Sweet, fair enough!

packages/entity-server/src/models/Project.ts
edmofro: Yeah please add the rule

packages/entity-server/src/models/Entity.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Model } from './types';
+
+export interface EntityFields {
+  readonly code: string;
+  readonly country_code: string;
+}
+
+export interface EntityType extends BaseEntityType, EntityFields {
+  getChildren: (hierarchyId: string) => Promise<EntityType[]>;
+  getChildrenViaHierarchy: (hierarchyId: string) => Promise<EntityType[]>;
```
edmofro, line 16: I've just looked at this second method, and I'm not sure it has any place in modern Tupaia? Is there any distinction between the two? I can see that the second would exclude any canonically connected descendants, but I'm not sure that's a use case that's important?

packages/entity-server/src/routes/hierarchy/middleware.ts
```diff
@@ -46,39 +52,44 @@ export const attachFormatterToResponse = async (
...
   const mappedEntity: { -readonly [field in keyof EntityFields]?: EntityFields[field] } = {};
   fields.forEach(field => {
-    mappedEntity[field] = model[field];
+    mappedEntity[field] = entity[field];
   });
   return mappedEntity;
 };
 
 const validFields: (keyof EntityFields)[] = ['code', 'country_code'];
+const validateField = (field: string): field is keyof EntityFields => {
```
edmofro, line 89: Thanks for moving this one, but it was actually a general comment for throughout this file 🙏

packages/entity-server/src/routes/hierarchy/types.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export type EntityResponseObject = {
+  [field in keyof EntityFields]?: EntityFields[field];
+};
+
+// eslint-disable-next-line @typescript-eslint/comma-dangle
+export interface HierarchyRequest<
+  P = HierarchyRequestParams,
+  ResBody = EntityResponseObject,
+  ReqBody = Record<string, never>,
```
edmofro, line 28: What's the `never` about? I looked it up, but still can't fathom its use here

edmofro: Is it because a `HierarchyRequest` is always a `GET`, so shouldn't have a body...?

edmofro (changes requested):
I have a question about the inclusion of project here. In our designs, we'd made the endpoints take the `hierarchyCode`. From that I'd assumed that the client would deal with projects, and the `entity-server` would just deal with hierarchies. 

Did you decide it would be easier on clients to give the `entity-server` knowledge of projects? 

From the LESMIS end, there are no projects, so that client could certainly just use the hierarchy directly rather than thinking about a project code, but maybe most other uses of the entity server would have to do an extra step to get the hierarchy from the project, so it could be justifiable. Keen to hear your thinking!

Note: I see now it's `entity_hierarchy.name` so probably should have been `hierarchyName` rather than `hierarchyCode`, but same diff

packages/entity-server/src/routes/hierarchy/types.ts
rohan-bes: Yeah it was a bit of a weird thing to do hehe. I googled around and this is what was recommended to me when you want to suggest that a type should be empty. I'm happy to go with anything really since this is just the default setting, and ideally we'd be explicitly setting what the ReqBody should look like per route

packages/entity-server/src/routes/hierarchy/middleware.ts
rohan-bes: Oh I see. I think my desire to have exported members at the top was conflicting with this. Will fix the rest of the file 👍

packages/entity-server/src/models/Entity.ts
rohan-bes: I'll have to re-test this, at the time I added it since it seems the common way that we fetch children of project entities throughout Tupaia. I was failing to get the expected results using `getChildren()` but I think that was actually due to me incorrectly using `hierarchyName` instead of `hierarchyId`. Good pickup, I'd prefer not to use this one if possible...

edmofro (approved):
Looks great!

PR #2210, opened 2021-03-10
2283 lesmis breadcrumbs

tcaiger:
### Changes:

- add lesmis breadcrumbs component
- update page routes
- add placeholder fetching of orgunit data from webconfig server

---

### Screenshots:
see issue


packages/lesmis/src/api/queries/useOrgUnitData.js
```diff
@@ -0,0 +1,37 @@
+/*
```
tcaiger, line 1: This is a place holder query until EntityServer is ready and so is outside the scope of this review.

packages/lesmis/src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,81 @@
...
+    return { ...query, segments: [] };
+  }
+
+  const { countryHierarchy, organisationUnitCode } = query.data;
+  const segments = getSegments(countryHierarchy, organisationUnitCode);
+
+  return { ...query, segments };
+};
+
+export const Breadcrumbs = () => {
```
tcaiger, line 63: Since the ui component is basically just an implementation of material-ui and this component is more about mapping data to the url, I decided that it's better to make this a component in LESMIS and not a re-usable component in the ui-components package.

packages/lesmis/src/routes/PageRoutes.js
```diff
@@ -6,17 +6,15 @@
...
 import { ProfileView } from '../views/ProfileView';
 import { AboutView } from '../views/AboutView';
 import { ContactView } from '../views/ContactView';
+import { EntityView } from '../views/EntityView';
 import { NotFoundView } from '../views/NotFoundView';
 
 /**
  * Main Page Routes
- * eg. /vientiane-capital/mayparkngum/mayparkngum
+ * eg. /en/TO/dashboard
```
tcaiger, line 17: internationalisation is to come later

packages/lesmis/src/routes/PageRoutes.js
```diff
@@ -32,14 +30,8 @@ export const PageRoutes = React.memo(() => (
     <Route path="/contact">
       <ContactView />
     </Route>
-    <Route exact path="/:province">
```
tcaiger, line 35: Now that we have a url design, I am updating the routes to match. URL design was done in [this](https://app.zenhub.com/workspaces/active-sprints-5eea9d3de8519e0019186490/issues/beyondessential/tupaia-backlog/2274) task

packages/lesmis/src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,81 @@
...
+const Link = props => <MuiLink color="inherit" {...props} component={RouterLink} />;
+
+const getSegments = (orgUnits, orgUnitCode, heirarchy = []) => {
+  const orgUnit = orgUnits.find(entity => entity.organisationUnitCode === orgUnitCode);
+
+  if (!orgUnit) {
+    return [];
+  }
+
+  const newHeirarchy = [...heirarchy, orgUnit];
```
edmofro, line 42:
```suggestion
  const newHierarchy = [...hierarchy, orgUnit];
```
Better do a find/replace on this

edmofro: Also
```suggestion
  const newHeirarchy = [orgUnit, ...heirarchy];
```
Would avoid the `reverse` (no biggy either way - whatever you find easier to think about)

packages/lesmis/src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,81 @@
...
+
+  if (orgUnit.type === 'Country') {
+    return newHeirarchy.reverse();
+  }
+  return getSegments(orgUnits, orgUnit.parent, newHeirarchy);
+};
+
+const useBreadcrumbs = () => {
+  const query = useCurrentOrgUnitData({ includeCountryData: true });
```
edmofro, line 51:
```suggestion
  const orgUnitResponse = useCurrentOrgUnitData({ includeCountryData: true });
```

packages/lesmis/src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,81 @@
...
+
+  if (!query.data) {
+    return { ...query, segments: [] };
+  }
+
+  const { countryHierarchy, organisationUnitCode } = query.data;
+  const segments = getSegments(countryHierarchy, organisationUnitCode);
+
+  return { ...query, segments };
```
edmofro, line 60: Why are we returning more than just the `segments` here? Do you want to have some breadcrumb error handling?

edmofro (approved):
Really getting used to and loving the custom hooks vibe

packages/lesmis/src/components/Breadcrumbs.js
tcaiger: 💯

packages/lesmis/src/components/Breadcrumbs.js
tcaiger: yeah exactly for a loading state or error state. I have added a loader as might as well do that now. For error handling on get requests, I don't really want to have ui for that everywhere so I will think of a generic ui for that some how (maybe error boundaries)

PR #2211, opened 2021-03-11
2444: Added default values for daily cases data

chris-bes:
### Changes:

- Added default values for daily cases data


kael89 (approved):

packages/database/src/migrations/20210311002222-AddDefaultValueForDailyCasesInPSSSWeeklyIndicators-modifies-data.js
```diff
@@ -0,0 +1,171 @@
+'use strict';
```
chris-bes, line 1: This migration can be written better but I just wanted to fix the bug quickly

PR #2214, opened 2021-03-11
2153 map overlay components part 3

tcaiger:
### Changes:

- Add Polygon, PolygonLayer and AreaTooltips

---


rohan-bes (approved):
Not gonna lie I did get a bit of a scare when I saw +10k lines hehe

This is really good, much cleaner than our existing polygon layer code 👍

PR #2216, opened 2021-03-11
2445: Cleanup e2e tests MVP

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2445


packages/web-frontend/cypress/constants.js
```diff
@@ -5,7 +5,6 @@
 
 import { integrationFolder } from '../cypress.json';
 
-export const CONFIG_ROOT = 'cypress/config';
```
kael89, line 8: Initially I thought that it would would be useful to encapsulate where config is stored. In practice, we use a lot of hard coded `config/...` imports, so no reason to pretend that this is configurable.

packages/web-frontend/cypress/scripts/generateConfig/generateReportConfig.js
```diff
@@ -123,10 +124,17 @@ const selectUrlParams = async (database, report, dashboardGroups) => {
     }
   }
 
-  return undefined;
+  // No compatible entry found, throw error
+  throw new Error(
```
kael89, line 128: It's probably better to throw an error here than just skip this report, which is the existing functionality.

That way we'll be aware of the need to add new `orgUnitMaps.json` entries - which is a simple task really and shouldn't happen too often.

packages/web-frontend/cypress/support/helpers.js
```diff
@@ -3,26 +3,6 @@
...
-export const getConfigPath = fileName => `${CONFIG_ROOT}/${fileName}.json`;
-
-const getConfigExamplePath = fileName => `${CONFIG_ROOT}/${fileName}.example.jsonc`;
-
-export class EmptyConfigError extends Error {
-  constructor(fileName) {
-    const path = getConfigPath(fileName);
-    const examplePath = getConfigExamplePath(fileName);
-    const generationCommand = '`yarn workspace @tupaia/web-frontend cypress:generate-config`';
-    const message = `Test config error: ${path} contains no test data. Either use ${generationCommand} to populate this file, or populate it with manual data. See ${examplePath} for an example`;
```
kael89, line 19: Let's keep things simple - the users of our code should know (or ask) what to do if they get an "empty config" error. Plus, I added config documentation in this PR.

packages/web-frontend/package.json
```diff
@@ -16,12 +16,12 @@
...
     "cypress:generate-config": "yarn run:babel cypress/scripts/generateConfig",
-    "cypress:open": "cypress open",
-    "cypress:run": "cypress run",
+    "cypress:open": "cross-env NODE_ENV=test cypress open",
+    "cypress:run": "cross-env NODE_ENV=test cypress run",
     "lint": "eslint --ignore-path ../../.gitignore .",
     "lint:fix": "yarn lint --fix",
     "run:babel": "babel-node --config-file \"../../babel.config.json\"",
-    "start": "PORT=8088 node scripts/start.js",
+    "start": "cross-env PORT=8088 node scripts/start.js",
```
kael89, line 24: Pretty sure that this will be fine, but we should test that a feature branch is successfully deployed after this change

edmofro (approved):
All lgtm 🤓

PR #2218, opened 2021-03-11
2153 map overlay components part 4

tcaiger:
### Issue #: [2153](https://github.com/beyondessential/tupaia-backlog/issues/2153) Map Overlay Components Part 1

This PR is for the marker map components. It also has the marker layer component which is just a convenient wrapper around the markers.

### Changes:

- Add marker components
- Add marker layer

---


packages/ui-components/src/components/Map/Markers/markerColors.js
```diff
@@ -0,0 +1,223 @@
...
+  blue: '#00A0FA',
+  orange: '#FA7850',
+  magenta: '#FA78FA',
+  cyan: '#14D2DC',
+  green: '#0AB45A',
+  red: '#AA0A3C',
+  lime: '#A0FA82',
+};
+
+export const TUPAIA_ORANGE = '#EE6230';
```
tcaiger, line 25: There is a card in the backlog to update the colors to be derived from a material-ui theme.https://github.com/beyondessential/tupaia-backlog/issues/2213

packages/ui-components/stories/story-utils/api/queries/useMeasureData.js
```diff
@@ -0,0 +1,128 @@
+/*
```
tcaiger, line 1: The `useData` hooks are designed to be the data integration between the api and the components. I have deliberately kept these outside of the components so that the components are as little aware of the api as possible.

rohan-bes (approved):
I think it looks good to me. PR's like this make me really wish there was some option to check for copy/paste changes in github. I checked out the branch and ran: `git diff --find-copies-harder 2153-map-overlay-components-part-1a` and it seemed okay to me 😄

rohan-bes:
Weird test failure... possibly the `cyclone.svg` file got broken somehow when it was copied???

PR #2223, opened 2021-03-12
2153 map overlay components part 1a

tcaiger:
### Changes:

- Update formatDataValueByType in tupaia utils to support all data types (essentially just move from ui-components utils to tupaia-utils)
- Refactor ui-component charts to use the util

In the short term map-overlays and in the long term web-frontend will also need to use this util.

---


packages/utils/src/formatDataValueByType.js
```diff
@@ -3,19 +3,97 @@
...
+const truncateDecimalToPlace = decimalPlace => number => {
+  const place = 10 ** decimalPlace;
+  return Math.floor(number * place) / place;
+};
+
+// Note: will display 0 if passed undefined
+const currency = value => numeral(value).format('$0.00a');
+
+const fraction = (value, { total }) => {
+  if (isNaN(total)) return 'No data';
```
kael89, line 23: Since this method has now moved in the more generic `utils` package, I'd suggest that it also returns a more generic value (`null`/`undefined`). Then the consumer can decide on how to interpret it (display `No data` or other).

This is just an example of the general issue of how we handle `No data` values. I believe we currently do so in a bit of a hard coded way, which is not ideal. Since it's a general issue and may be tricky to handle correctly, I'm happy for this to stay as is for now - but please keep it in mind for a possible future refactor!

packages/utils/src/constant.js
```diff
@@ -4,5 +4,13 @@
  */
 
 export const VALUE_TYPES = {
+  BOOLEAN: 'boolean',
```
kael89, line 7: I'd suggest to move this constant in `formatDataValueByType.js` and export it from there (I know you didn't add this file, but its existence doesn't make sense to me)

kael89 (changes requested):
Great work @tcaiger ! One small change and it's ready to go

It would also be great to add some unit tests for these at some point 😃

packages/utils/src/formatDataValueByType.js
tcaiger: Yes that's a good point. The no data string is a presentation concern and not a util concern. As you say I'm going to leave this one for now because it might be better to refactor the no data cases all at once or something.

packages/utils/src/constant.js
tcaiger: Yes good suggestion!

tcaiger:
@kael89 - Good reminder about the tests. There are some in web-frontend relating to this so I have moved them across and refactored them to fit the new util.

kael89 (approved):
Nice one!

PR #2239, opened 2021-03-15
Hotfix remove geograhicalArea import contraint

nick-beyondessential:
### Issue #: [2482](https://github.com/beyondessential/tupaia-backlog/issues/2482) Issue with importing entities

- remove contraint on parents having geographicalAreas recently added for country level facilities

### DIFF
--- a/packages/meditrak-server/src/routes/importEntities/getOrCreateParentEntity.js
+++ b/packages/meditrak-server/src/routes/importEntities/getOrCreateParentEntity.js
@@ -112,10 +112,8 @@ export async function getOrCreateParentEntity(transactingModels, entityObject, c
       parentEntity,
       transactingModels,
     );
-    if (!parentGeographicalArea)
-      throw new Error(
-        `Parent entity must have geographical area, parent entity code: ${parentCode}`,
-      );
+    if (!parentGeographicalArea) return { parentEntity };
+
     return { parentEntity, parentGeographicalArea };
   }
   // no explicit parent code provided, use either subdistrict or district as parent entity
--- a/packages/meditrak-server/src/routes/importEntities/updateCountryEntities.js
+++ b/packages/meditrak-server/src/routes/importEntities/updateCountryEntities.js
@@ -31,7 +31,7 @@ const getEntityMetadata = async (transactingModels, code, pushToDhis) => {
   const defaultMetadata = {
     dhis: { isDataRegional: true },
   };
-  
+
   // Only assign push = false, by default all entities will be pushed to dhis
   if (!pushToDhis) {
     defaultMetadata.dhis.push = pushToDhis;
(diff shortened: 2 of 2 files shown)


packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -97,7 +97,7 @@ export class DhisInputSchemeResolvingApiProxy {
 
     for (const dataElementCode of dataElementCodes) {
       const dataElement = dataElements.find(d => d.code === dataElementCode);
-      if (!dataElement.config.dhisId) {
+      if (!dataElement?.config?.dhisId) {
```
edmofro, line 100: I think you've got the wrong base branch @nick-beyondessential?

nick-beyondessential: yes, just changed

packages/meditrak-server/src/routes/importEntities/getOrCreateParentEntity.js
```diff
@@ -112,10 +112,8 @@ export async function getOrCreateParentEntity(transactingModels, entityObject, c
       parentEntity,
       transactingModels,
     );
-    if (!parentGeographicalArea)
-      throw new Error(
-        `Parent entity must have geographical area, parent entity code: ${parentCode}`,
-      );
+    if (!parentGeographicalArea) return { parentEntity };
```
edmofro, line 115: Should still throw the error if entity is of type facility

edmofro (changes requested):

packages/meditrak-server/src/routes/importEntities/getOrCreateParentEntity.js
edmofro: If you look at the calling code, this is the only place that parent geographical area is used
```
const { parentGeographicalArea, parentEntity } = await getOrCreateParentEntity(
      transactingModels,
      entityObject,
      country,
    );
    if (entityType === transactingModels.entity.types.FACILITY) {
      const defaultTypeDetails = getDefaultTypeDetails(facilityType);
      const facilityToUpsert = {
        type: facilityType,
        type_name: typeName || undefined, // Ensure empty string is treated as undefined
        category_code: categoryCode || undefined, // Ensure empty string is treated as undefined
        code,
        name,
        geographical_area_id: parentGeographicalArea.id,
      };
```

The validation could also be done there instead.

edmofro (approved):
Thanks (and glad that linting got caught)

packages/meditrak-server/src/routes/importEntities/updateCountryEntities.js
```diff
@@ -31,19 +31,24 @@ const getEntityMetadata = async (transactingModels, code, pushToDhis) => {
   const defaultMetadata = {
     dhis: { isDataRegional: true },
   };
-  
+
```
nick-beyondessential, line 34: caught a bunch of eslint auto-tidyups

nick-beyondessential:
Thanks for the review despite your busy schedule @edmofro, appreciate it!

PR #2240, opened 2021-03-15
2365: Added dengue commodities table

chris-bes:
### Changes:

- Added dengue commodities table


chris-bes:
On hold since the data elements are not accessible

PR #2246, opened 2021-03-16
2484 delete fiji entities

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/2484

### DIFF
--- a/packages/database/src/migrations/20210316003128-DeleteCertainFijiEntities-modifies-data.js
+++ b/packages/database/src/migrations/20210316003128-DeleteCertainFijiEntities-modifies-data.js
@@ -1,0 +1,51 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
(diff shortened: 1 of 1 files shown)


avaek:
Closing this PR as this is going to be released as a hotfix rather than included with that release branch

packages/database/src/migrations/20210316003128-DeleteCertainFijiEntities-modifies-data.js
```diff
@@ -0,0 +1,51 @@
...
+  'FJ_242',
+];
+
+exports.up = async function (db) {
+  FACILITIES_TO_REMOVE.forEach(async facility => {
+    await db.runSql(`
+      DELETE FROM "clinic" WHERE "code"='${facility}';
+      DELETE FROM "entity" WHERE "code"='${facility}';
+    `);
+  });
```
nick-beyondessential, line 42: Looks fine but fyi I don't think the loop will wait and you could do: 
```suggestion
 await db.runSql(`
      DELETE FROM "clinic" WHERE "code" in (${arrayToDbString(FACILITIES_TO_REMOVE)});
      DELETE FROM "entity" WHERE "code" in (${arrayToDbString(FACILITIES_TO_REMOVE)});
    `);
```

PR #2247, opened 2021-03-16
2301: Allow disable syncing subdistrict and district when importing facilities

chris-bes:
### Changes:

- Allow disable syncing subdistrict and district when importing facilities. Addressing the bug found here: https://github.com/beyondessential/tupaia-backlog/issues/2301#issuecomment-799054228


nick-beyondessential (approved):
Looks good Chris.

PR #2252, opened 2021-03-16
Hotfix bind params too long

IgorNadj:

### DIFF
--- a/packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
+++ b/packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
@@ -222,12 +222,19 @@ export class DhisInputSchemeResolvingApiProxy {
 
     const dhisIds = response.rows.map(row => row[orgUnitIdIndex]);
 
-    const mappings = await this.models.dataServiceEntity.find({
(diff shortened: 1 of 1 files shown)


packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -222,12 +222,19 @@ export class DhisInputSchemeResolvingApiProxy {
...
     const dhisIds = response.rows.map(row => row[orgUnitIdIndex]);
 
-    const mappings = await this.models.dataServiceEntity.find({
-      [QUERY_CONJUNCTIONS.RAW]: {
-        sql: `config->>'dhis_id' in (${dhisIds.map(() => '?')})`,
-        parameters: dhisIds,
-      },
-    });
+    let mappings = [];
+    // need to batch this query because we can get an SQL error if we do too many
```
edmofro, line 226: Happy for this not to happen now as I know it's an urgent hot fix, but please come back to it

```suggestion
    // need to batch this query because we can get an SQL error if we do too many
    // TODO convert to using `runDatabaseFunctionInBatches` or one of its wrappers
```

edmofro (approved):

nick-beyondessential (approved):
Looks good to me.

PR #2254, opened 2021-03-16
Fix countDisasterAffectedFacilitiesByStatus query params

nick-beyondessential:
### Changes:
- add missing query params to fetch analytics in countDisasterAffectedFacilitiesByStatus


kael89 (approved):
Good fix!

PR #2265, opened 2021-03-18
2312: Entity-Server - Fetch Descendants Endpoint

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2312:

`/hiearachy/:hierarchyName/:entityCode/descendants` endpoint

A few tricky things going on here:
- Had to develop branching logic to handle getting parent/child codes for many entities vs a single entity. The code flow is smoother for a single entity (see: `packages/entity-server/src/routes/hierarchy/middleware/fields.ts#mapEntityToFields`), but the performance is much worse, so had to implement `packages/entity-server/src/routes/hierarchy/middleware/fields.ts#mapEntitiesToFields`
- Wanted to perform a join twice on the same table for the AncestorDescendantRelationModel,  which wasn't working in our ORM code without a few changes, so I guess there's a regression risk there...
- Little bit of refactoring going on, don't be too put off by the size of the diffs 👍


packages/database/src/modelClasses/AncestorDescendantRelation.js
```diff
@@ -3,25 +3,59 @@
...
-  async getChildIdToParentId(hierarchyId) {
-    const relationRecords = await this.find({
+  async getImmediateRelations(hierarchyId) {
+    return this.find({
       entity_hierarchy_id: hierarchyId,
       generational_distance: 1,
     });
+  }
+
+  async getChildIdToParentId(hierarchyId) {
```
rohan-bes, line 42: Will probably refactor out usages of this method in `web-config-server` as I believe they always actually want the `code` instead of `id` anyway...

packages/database/src/modelClasses/Entity.js
```diff
@@ -152,6 +152,11 @@ export class EntityType extends DatabaseType {
     return this.parent_id ? this.model.findById(this.parent_id) : undefined;
   }
 
+  async getParent(hierarchyId) {
+    const ancestors = await this.getAncestors(hierarchyId, { generational_distance: 1 });
+    return ancestors && ancestors.length > 0 ? ancestors[0] : undefined;
+  }
```
rohan-bes, line 158: Hope the naming here isn't too clashing with the existing `parent()` method...
Tbh I wonder if we can remove that method as it seems a bit archaic...

packages/entity-server/examples.http
```diff
@@ -1,9 +1,24 @@
+@hostname = localhost
+@port = 8050
+@host = {{hostname}}:{{port}}
+@version = v1
+@contentType = application/json
+
+# In order to setup authorization, please set 'email' and 'password' in your restClient environement variables 
+# see: https://marketplace.visualstudio.com/items?itemName=humao.rest-client#environment-variables
+@authorization  = Basic {{email}}:{{password}}
```
rohan-bes, line 9: This is super handy and avoid someone accidentally committing their credentials

packages/entity-server/src/app/createApp.ts
```diff
@@ -38,8 +38,11 @@ export function createApp(models: EntityServerModelRegistry) {
   app.use((req: Request, res: Response, next: NextFunction) => {
     req.models = models;
     req.authenticator = authenticator;
-    req.context = {};
-    res.context = {};
+
+    const context = {}; // context is shared between request and response
+    req.context = context;
+    res.context = context;
```
rohan-bes, line 44: Decided to go with one shared context between request and response. My feeling was that the distinction between what belonged where was rather arbitrary... I like this way but can understand if others find it a little unexpected

packages/entity-server/src/routes/hierarchy/EntityDescendantsRoute.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export class EntityDescendantsRoute extends Route<
+  HierarchyRequest<HierarchyRequestParams, EntityResponseObject[]>,
+  HierarchyResponse<EntityResponseObject[]>
+> {
+  async buildResponse() {
+    return this.res.context.formatEntitiesForResponse(
+      [this.req.context.entity].concat(
+        await this.req.context.entity.getDescendants(this.req.context.hierarchyId),
+      ),
```
rohan-bes, line 22: Attaching the requested entity to it's descendant tree here... is this expected? It's inline with what we currently do in `web-config-server` and it does have the upside of saving the client from doing multiple requests to get a root entity and it's descendants (which is quite common) but I can see how it may be unanticipated...

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
```diff
@@ -0,0 +1,119 @@
...
+      (mappedEntity[field] as string | string[] | undefined) = await extendedFieldFunctions[field](
+        entity,
+        context,
+      );
+    }
+  }
+  return mappedEntity;
+};
+
+export const mapEntitiesToFields = (
```
rohan-bes, line 68: This function is right on the borderline for me of unreadability (switch statement within an if/else within a loop...) but I'm okay with how it is for now as atleast it encapsulates a lot of related funky logic in the one place. Welcome for pushback on it though

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
```diff
@@ -0,0 +1,119 @@
...
+          entities.forEach((entity, index) => {
+            mappedEntities[index].children_codes = getChildrenCodesBulk(
+              entity,
+              parentToChildrenMap,
+            );
+          });
+          break;
+        }
+        default: {
+          throw new Error(`Unknown extended entity field requested: ${field}`);
```
rohan-bes, line 113: Ideally we won't be throwing an error here but instead just blindly calling `await extendedFieldFunctions[field](entity, context);` like we do in `mapEntityToFields()`. However Typescript is being a little too smart here and knows that if the field isn't either `parent_code` or `children_codes` then it's invalid and forces us to throw an error.

If we add more `extendedFieldFunctions` in the future that doesn't require any special logic when performing on multiple entities (region data, etc.) then we can replace throw statement with the behaviour above 👍

packages/utils/src/object.js
```diff
@@ -99,13 +105,38 @@ export const flattenToObject = objectCollection => {
...
+ * Available field (key/value) mapper types:
+ * * `string`: uses the provided string as the field key
+ * * `Function`, will receive the object as its input eg object => object.value * 2
+ *
+ * @param {ObjectCollection} objectCollection
+ * @param {string|Function} keyMapper
+ * @param {string|Function} valueMapper
+ * @return {Object<string, string[]>}
+ */
+export const reduceToArrayDictionary = (objectCollection, keyMapper, valueMapper) => {
```
rohan-bes, line 129: Added this function for the scenario that we want to use `reduceToDictionary` but we know there will be collisions (ie. matching a parentCode to it's children)

packages/entity-server/src/routes/hierarchy/extendedFieldFunctions.ts
```diff
@@ -0,0 +1,50 @@
...
+  return (await entity.getParent(hierarchyId))?.code;
+};
+
+/**
+ * Faster than getParentCode() but requires pre-calculating the childToParentMap
+ * Use when performing getParentCode() for many entities
+ */
+export const getParentCodeBulk = (entity: EntityType, childToParentMap: Record<string, string>) => {
+  return childToParentMap[entity.code];
+};
```
edmofro, line 24: Honestly this seems like a function that adds more confusion than it solves? The content is super simple, and could be done and read inline. I would expect `getParentCodeBulk` to take a bulk list of entities and return all their parent codes.

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
```diff
@@ -0,0 +1,119 @@
...
+    return allFields; // Display all fields if none specifically requested
+  }
+
+  const requestedFields = queryFields.split(',');
+  const fields = new Set<keyof ExtendedEntityFields>();
+  requestedFields.forEach(requestedField => {
+    if (validateField(requestedField)) {
+      fields.add(requestedField);
+    } else {
+      throw new Error(`Unknown field requested: ${requestedField}`);
```
edmofro, line 42: Will this error end up being returned as the query response? McLean has been promoting this idea of the "robustness principle" over in Tamanu, which is to be conservative with what you send but _liberal with what you accept_. That would suggest we should just ignore invalid fields here. I'm not totally sold on it when we control the clients as well as the server, but I'm interested to have it in mind as we watch people interact more with these micro-services

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
edmofro: I'm actually fine with it - it reads pretty clearly to me tbh (though I do have the context of the rest of the PR in my head)

edmofro (approved):
Awesome work! Love the effort you've put into supporting `children_codes` but keeping things as fast as possible 🙌 

Actually just writing that out, I think it should be `child_codes`...but otherwise I only have optional suggestions so approved!

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
```diff
@@ -0,0 +1,119 @@
...
+  ExtendedEntityFields,
+  EntityResponseObject,
+} from '../types';
+import {
+  extendedFieldFunctions,
+  getParentCodeBulk,
+  getChildrenCodesBulk,
+} from '../extendedFieldFunctions';
+
+const validFields: (keyof EntityFields)[] = ['code', 'country_code'];
```
tcaiger, line 20: @rohan-bes - is there any reason why `name` `image_url` and `type` fields are not included as valid fields? I also checked in ExtendedEntityFields and didn't see them.

packages/entity-server/src/routes/hierarchy/EntityDescendantsRoute.ts
rohan-bes: I like this suggestion :)

rohan-bes:
> Awesome work! Love the effort you've put into supporting `children_codes` but keeping things as fast as possible 🙌
> 
> Actually just writing that out, I think it should be `child_codes`...but otherwise I only have optional suggestions so approved!

Thanks for the review!

Hahaha I knew something sounded wrong to me when I kept writing `children_codes` out 😂 Will fix!

packages/database/src/modelClasses/AncestorDescendantRelation.js
```diff
@@ -3,25 +3,59 @@
...
+    const relationRecords = await this.getImmediateRelations(hierarchyId);
     return reduceToDictionary(relationRecords, 'descendant_id', 'ancestor_id');
   }
+
+  async getChildCodeToParentCode(hierarchyId) {
+    const relationRecords = await this.getImmediateRelations(hierarchyId);
+    return reduceToDictionary(relationRecords, 'descendant_code', 'ancestor_code');
+  }
+
+  async getParentIdToChildIds(hierarchyId) {
```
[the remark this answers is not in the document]
rohan-bes: There were a few iterations of work getting this going. Initially I was pulling just the `ids` in and then mapping them to codes based on the entity objects returned from the `getDesendants()` fetch. This is how we do it currently in `/organisationUnit` in `web-config-server`. However the drawback there is that:
a) You can't show the parent/child of entities not in that fetch
b) The logic is just a bit more awkward and convoluted

So I switched to doing the join in the db layer and fetching the codes directly, however thought it might be best to leave a few of these functions around that aren't doing any harm?

packages/entity-server/src/routes/hierarchy/extendedFieldFunctions.ts
rohan-bes: Yeah, you're right. Originally it contained a lot more logic here and made sense to be here but I might just switch to doing it all inline for these `bulk` operations 👍

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
rohan-bes: Those are coming in the second PR: https://github.com/beyondessential/tupaia/pull/2266
👍

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
rohan-bes: Yes, it will return an error in the query response. The idea of tolerating 'invalid' fields is a bit of an interesting one with the way I've written the typescript here (where the keys of `extendedFieldFunctions` are a type that we use commonly). It could be handed, though I may need to do some refactoring...

As to the principal in general, I see the value but I dunno if see a strong reason to adopt it when as you say we are in control of the clients as well...

Might leave it as is for now?

edmofro: Yes sorry please do leave it as is for now! This one was more of a thought bubble cross-pollination idea than a suggested change for this PR :-)

packages/database/src/modelClasses/AncestorDescendantRelation.js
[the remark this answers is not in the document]
edmofro: Yeah that's fair enough. I'm tempted to remove this one, just because it's one less bit of clutter, but it's honestly no big deal either way

packages/web-config-server/src/apiV1/dataBuilders/generic/orgUnit/basicDataVillage.js
```diff
@@ -2,7 +2,7 @@ import { DataBuilder } from '/apiV1/dataBuilders/DataBuilder';
 
 export class BasicDataVillageBuilder extends DataBuilder {
   async build() {
-    const facility = await this.entity.parent();
+    const facility = await this.entity.getParent(await this.fetchEntityHierarchyId());
```
edmofro, line 5: Nice :+1:

edmofro (approved):

packages/database/src/modelClasses/AncestorDescendantRelation.js
rohan-bes: Might hold off on doing this refactor for now as it slightly changes the code logic and there's a bit of complexity there (crossing between `web-config-server` and `web-frontend packages`)

edmofro: No worries, that's sensible

PR #2266, opened 2021-03-18
2312: Part 2 - Added the remaining fields to the entity-server

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2312:
- Added the remaining fields to the entity-server
- Fixed bug in getDescendants and children_codes where results were not being filtered by countries the user has access to
- Refactored geojson functions in common @tupaia/utils
- Updated typescript to 4.2.3 to use newer syntax


packages/entity-server/src/models/types.ts
```diff
@@ -5,7 +5,13 @@
 
 import { DatabaseModel, DatabaseType } from '@tupaia/database';
 
+export type DbConditional<T> = { [field in keyof T]?: T[field] | T[field][] };
+
+export type Joined<T, U extends string> = {
+  [field in keyof T as field extends string ? `${U}.${field}` : never]: T[field];
+};
```
rohan-bes, line 12: This is really cool, it basically remaps the keys of one type to be prefixed with the given string, eg:
```
Joined<EntityFields, 'ancestor'> =>
{
  'ancestor.id': string;
  'ancestor.code': string;
  'ancestor.name': string;
  ...
}
```

packages/utils/src/geoJson.js
```diff
@@ -0,0 +1,86 @@
...
+
+  return coords.map(flipLatLngRecursive);
+}
+
+/**
+ * Convert point geojson string into [lat, lng]
+ * @param {string | null} pointGeoJsonString
+ * @returns {Point | null} [lat, lng]
+ */
+export function translatePoint(pointGeoJsonString) {
```
rohan-bes, line 33: Happy with these names or should they be a bit more verbose?
I feel the entire function signature pretty clearly explains what it does, however when it's imported and used inline it can be a little vague, eg:
```
    case 'coordinates': {
      const [lat, long] = translatePoint(entity.point);
      return `${lat}, ${long}`;
    }
```

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
```diff
@@ -46,20 +49,41 @@ export const extractFieldsFromQuery = (queryFields?: string) => {
 };
 
 type Writable<T> = { -readonly [field in keyof T]?: T[field] };
+
+const assignExtendedFieldToResponseObject = async (
```
edmofro, line 53: It's amazing how functionally oriented you've become considering your Java background! This is totes cool, but my instinct would be to create some sort of a `EntitiesResponse` class that held on to `mappedEntities` and had `assignField` and `assignExtendedField` methods. At the end you could call `getBuiltResponse`. I just don't like passing variables in to functions with the idea that they'll be mutated by that function - feels off.

packages/entity-server/src/routes/hierarchy/middleware/middleware.ts
```diff
@@ -34,16 +22,33 @@ export const attachContext = async (req: HierarchyRequest, res: Response, next:
 
     const hierarchy = await req.models.entityHierarchy.findOne({ name: hierarchyName });
     const entity = await req.models.entity.findOne({ code: entityCode });
+    if (!hierarchy || !entity) {
+      throwNoAccessError(hierarchyName, entityCode);
+    }
+
+    const rootEntity = entity.isProject()
+      ? entity
+      : await req.models.entity.findOne({ code: hierarchy.name });
```
edmofro, line 31: Hmmm - this would be adding an assumption/constraint that we don't currently have. I'd prefer to avoid binding projects <-> hierarchies in a one to one relationship. You might have e.g. a "Fiji Water Catchment Hierarchy" that is used by the "WISH Fiji" project and a future "Fiji Flood Risk" project.

packages/entity-server/src/routes/hierarchy/middleware/middleware.ts
```diff
@@ -34,16 +22,33 @@ export const attachContext = async (req: HierarchyRequest, res: Response, next:
...
+
+    const allowedCountries = (await rootEntity.getChildren(hierarchy.id))
+      .filter(e => req.accessPolicy.allows(e.country_code || undefined))
+      .map(e => e.country_code)
+      .filter(notNull);
```
edmofro, line 36: Might be cool to de-dupe here as well, in case the children of the project aren't countries themselves, and we end up with like 100 `FJ` codes. Not necessary as I think the assumption that project children are countries is baked in elsewhere (and also there's no broken logic as a result of having duplicates in the allowed countries list)

packages/utils/src/geoJson.js
edmofro: Inline that does look a bit vague. Nice and concise though, and could always look at the import to know it was coming from `geoJson`, which already gives you more info. 

I wonder if we should put helpers onto `EntityType`, so you can go `entity.getCoordinates();` or `entity.getBounds();` and it'll give you the nice format? That way there aren't many vague inline places

edmofro (changes requested):
Just one required change, which is to not assume that there's a one to one mapping between projects and hierarchies. Otherwise looking great :-)

tcaiger:
@rohan-bes - this is looking really cool. The typescript is blowing my mind. One small thing that would be nice for future integrations is a list of required env variables, either in an .env.example file or in the readme.

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
rohan-bes: You're totally right, that's a much cleaner pattern!

packages/entity-server/src/routes/hierarchy/middleware/middleware.ts
rohan-bes: Ooh good pickup!

rohan-bes:
> @rohan-bes - this is looking really cool. The typescript is blowing my mind. One small thing that would be nice for future integrations is a list of required env variables, either in an .env.example file or in the readme.

@tcaiger good call on the `.env.example` file, added it in the base PR 👍

packages/entity-server/src/routes/hierarchy/middleware/middleware.ts
rohan-bes: Wanna have a quick chat about this one? I actually considered this as I was writing the code but I guess my understanding is that we are already tied into that relationship due to there only being able to be one root entity per entity_hierarchy? I also feel it's not quite what we want, so would be happy to discuss how we can avoid this 😄

edmofro: Yep let's have a chat - I'll call you in 15

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
rohan-bes: I'm so glad that you requested this! Since refactoring to use a class I've been able to finally crack the generics problem I was facing in this code area that was forcing me to do a really ugly explicit typecast previously. Tbh I'm still not entirely sure why it works now but not previously, but I am very happy it does 🎉

edmofro (commented):
All looking great on re-review :-)

PR #2272, opened 2021-03-23
2182: Display loading circle when clicking retry

EMcQ-BES:
### Changes:

- When fetching data, clear the view contents (when a view has no content it displays the loading circle)


tcaiger (approved):
Good fix. I like that it's in the reducer, as it's nice and clean.

PR #2274, opened 2021-03-23
2485 fix duplicate y axis values

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/2485

It seems that when the `valueType` for a report is `number` that it can cause a bug with the Y Axis if the values are less than 5 (shown below, the value there is 1 and the Y Axis is displaying "duplicate" values)
- I think what is happening is each tick in the Y Axis is really a decimal however they're being rounded up/down to whole numbers because of the number type (so it seems like duplicate values but really it's just the distinct decimals are being rounded up/down)
- Note: this report only has `"valueType": "number"` because I changed it to this for testing purposes locally, normally it does not

Screen Shot 2021-03-23 at 3 07 55 PM

A solution I've found to this is just not allowing decimals when the `valueType` is `number`
- This solution changes the above report to look like this:

Screen Shot 2021-03-23 at 3 09 31 PM


edmofro:
Did you look in to allowing the decimals to show with a higher rounding precision if ticks have less than 1 full number's interval between them? I think that could be a nicer solution if it were possible

avaek:
> Did you look in to allowing the decimals to show with a higher rounding precision if ticks have less than 1 full number's interval between them? I think that could be a nicer solution if it were possible

I didn't but I can

packages/ui-components/src/components/Chart/YAxes.js
```diff
@@ -69,6 +74,7 @@ const YAxis = ({ config = {}, viewContent, isExporting }) => {
     <YAxisComponent
       key={yAxisId}
       ticks={ticks}
+      allowDecimals={displayDecimals(valueType)}
```
kael89, line 77:
```suggestion
      allowDecimals={valueType !== NUMBER}
```

(`NUMBER` can be exported from the `VALUE_TYPES` "enum" in line 12

We can then delete `displayDecimals`. Similarly for `CartesianChart`

kael89 (changes requested):
I agree with [Edwin's comment](https://github.com/beyondessential/tupaia/pull/2274#issuecomment-804535755) that ideally we should allow decimals ticks. If eg our values are 0.02, 0.04, forcing integer ticks would make chart bars unnecessarily small. 

My suggestions are for the existing implementation, in case tey are relevant to the final implementation

avaek:
> Did you look in to allowing the decimals to show with a higher rounding precision if ticks have less than 1 full number's interval between them? I think that could be a nicer solution if it were possible

There are only 16 reports on Tupaia that explicitly have a `number` `valueType` - most reports that display numbers have an `undefined` `valueType` (and therefore use the default formatter, which displays numbers at 2 decimal places where relevant (meaning that they aren't affected by the bug that this PR fixes))
- The easiest solution seems to be to just force the reports that explicitly have a `number` `valueType` to also use the default formatter
- I briefly tested this locally and couldn't find anywhere where this change caused any display issues (although further testing by the testers is required)

This change means the example report I showed in the initial description on this PR now looks like this:

Screen Shot 2021-03-24 at 10 51 08 AM

kael89 (approved):
Nice solution @avaek !

PR #2276, opened 2021-03-23
add lesmis searchbar

tcaiger:
### Changes:

- Add LESMIS SearchBar on home page and navbar

---

### Screenshots:
see issue


packages/lesmis/src/api/queries/useOrgUnitData.js
```diff
@@ -35,3 +35,6 @@ export const useCurrentOrgUnitData = ({ includeCountryData = false } = {}) => {
   const { organisationUnitCode } = useUrlParams();
   return useOrgUnitData({ organisationUnitCode, includeCountryData });
 };
+
+export const useCountryHeirarchyData = () =>
```
tcaiger, line 39: This will need to be updated soon once the entity server descendants endpoint is ready.

packages/lesmis/src/components/EntityMenu.js
```diff
@@ -0,0 +1,55 @@
+/*
```
tcaiger, line 1: There is another story for the EntityMenu [here](https://app.zenhub.com/workspaces/active-sprints-5eea9d3de8519e0019186490/issues/beyondessential/tupaia-backlog/2496). This is just a placeholder

packages/lesmis/src/components/SearchBar/utils.js
```diff
@@ -0,0 +1,46 @@
...
+export const getOptionText = (option, countryHeirarchy, hierarchy = []) => {
+  const { type, name, parent } = option;
+
+  const newHierarchy = [...hierarchy, name];
+
+  if (type === 'Country') {
+    return newHierarchy.join(', ');
+  }
+
+  const parentOrgUnit = countryHeirarchy.find(entity => entity.organisationUnitCode === parent);
```
edmofro, line 39: As an optimisation, it could be worth passing in an object or Map that has entities keyed by code. I think we'll end up with this being useful in other places, so it could be worth creating it from the array right back in the `useCountryHierarchyData` hook so that consumers can use either data structure
```
const { entities, entityByCode } = useCountryHierarchyData;
```

edmofro (changes requested):
This is awesome, I love the overall design and the beautiful use of hooks to compose utils out of material-ui and our own logic

One comment inline for an optimisation, plus I'd suggest some general tightening up of terms.
- We use a mix of "entity" and "organisationUnit" throughout Tupaia to mean pretty much the same thing (if you want to know the distinction, the best summary is that "an organisation unit is just an entity that has a permanent fixed location"). Let's try to get away from using both in LESMIS, as it's unnecessary and confusing. My preference is `entity`. Where the response from web-config-server includes `organisationUnitCode`, I'd suggest we use the orchestration server to map that to `entityCode` for LESMIS front end
- The array being referred to as `countryHierarchy` isn't actually structured as a hierarchy, and the `country` bit isn't super relevant either. I'd prefer we just thought of that as "the list of all entities", and called it something like `entities` or `allEntities`. Yes, it has parent codes in each entity, but that's just a detail and doesn't reflect how the data is actually structured
- I think `includeCountryData` is _possibly_ the right name for the query parameter in Tupaia, but here it just seems confusing. From the LESMIS app's perspective, it's not really data about the country you're asking for, it's just all entities. I'd either hide that query parameter inside the orchestration server if it can be always on, or at least map it to something else like `includeAllEntities`. Potentially the orchestration server can just support different endpoints that have that turned on or off in the call across to web-config-server, so that LESMIS is just requesting either `/entities` (lesmis-server turns `includeCountryData` on) or `/entity/:entityCode` (lesmis-server turns `includeCountryData` off). This whole last point may be completely irrelevant when integrating with entity-server, so ignore it for now, but keep in mind when we're doing that integration.

edmofro:
Oh also - please find + replace `heirarchy` -> `hierarchy`

packages/lesmis/src/components/SearchBar/useAutocomplete.js
```diff
@@ -0,0 +1,67 @@
...
+      if (ignoreCase) {
+        candidate = candidate.toLowerCase();
+      }
+
+      return candidate.substring(1).indexOf(input) > -1;
+    });
+
+    const filteredOptions = [...fromStartMatches, ...anyMatches];
+
+    return filteredOptions.slice(0, limit);
```
rohan-bes, line 43: Hi Tom, sorry I started the review for this yesterday and never got to finish it 😢 

@edmofro has touched on pretty much everything else I was gonna say, just one more thing of note.

More of a nice to have I think, but could be nice to optimise this code a little bit by returning early when we reach the limit, eg something like:
```
const anyMatches = [];
const filteredOptions = [];

for (let i = 0; i < options.length && filteredOptions.length <= limit; i++) {
  const option = options[i];
  let candidate = (stringify || getOptionLabel)(option);
  if (ignoreCase) {
    candidate = candidate.toLowerCase();
  }
  
  if (candidate.indexOf(input) === 0) {
    filteredOptions.push(option); // Matches start
  } else if (candidate.substring(1).indexOf(input) > -1)
    anyMatches.push(option); // Matches anywhere
  }
}

if (filteredOptions.length === limit) {
  return filteredOptions;
}

return [...filteredOptions, ...anyMatches].slice(0, limit);
```

tcaiger:
@edmofro and @rohan-bes - This is ready for re-review. In this PR I have added optimisation for the filter function as well as some basic tests on that function. Thanks Rohan for the really good suggestion on that. I have created [another PR](https://github.com/beyondessential/tupaia/pull/2294) for the entity-server integration which includes the other changes requested such as renaming orgUnit to entityCode and fixing spelling.

packages/lesmis/src/components/SearchBar/utils.js
tcaiger: I had a play with doing this and if it's ok with you I've decided to not do it but keep it in mind. I'm weary of adding another variable to memory which is in the order of 3mb when it seems to be performing well as it is and doing the find in a couple of places is very easy. Just going for the library object is also not an option because the material autocomplete component expects an array.

packages/lesmis/src/components/SearchBar/__tests__/createFilterOptions.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+    expect(zeResults.every(result => zeExampleResults.includes(result.name))).toBe(true);
+    expect(zeResults.length).toBe(6);
+  });
+
+  it('Filters with macrons', () => {
+    expect(getFilterResults('Åland')[0].name).toBe('Åland Islands');
+  });
+
+  it('Returns options that start with the serach first', () => {
```
edmofro, line 51:
```suggestion
  it('Returns options that start with the search first', () => {
```

packages/lesmis/src/components/SearchBar/utils.js
edmofro: Fair enough, no need to optimise early if it's performing fine :-)

There would be other optimisation options as well, like caching the result so that if the same `option` was passed in, the result was returned immediately rather than looking through again (because at the moment the same results might be running through this loop each time the search term changes)

edmofro (approved):
lgtm :-)

PR #2278, opened 2021-03-23
Remove unpermitted Laos Eoc Facilities

nick-beyondessential:
### Changes:
- migration to remove Laos EOC facilities without access permission


chris-bes (approved):
Looks good! I've tested locally as well. (screenshot: https://github.com/beyondessential/tupaia-backlog/issues/2535#issuecomment-804646243)

Do you know how urgent do we have to fix this? (maybe a question for Kurt/Erin). I was under the impression that PSI will fix the permissions last week but seems like they still haven't.

I think we can merge this straight to dev and test there

nick-beyondessential:
Thanks @chris-bes sounds good.

PR #2282, opened 2021-03-23
Hotfix: Merge master to dev

nick-beyondessential:
### Changes:
merge back into dev


chris-bes (approved):
LGTM! I believe we should still use `Create a merge commit` to merge master back to dev

PR #2283, opened 2021-03-24
1149 internal data fetch respect timezone

EMcQ-BES:
### Changes:

- Create new column in survey responses `data_time`
- Replace uses of `submission_time` with `data_time`


packages/admin-panel/src/pages/resources/SurveyResponsesPage.js
```diff
@@ -36,10 +36,10 @@ const dateOfData = {
   Header: 'Date of Data',
   source: 'data_time',
   type: 'tooltip',
-  accessor: row => moment.parseZone(row.data_time).format('ddd, MMM Do YYYY, HH:mm:ss'),
+  accessor: row => moment.parseZone(row.data_time).toString(),
```
edmofro, line 39: This is the only change I don't get. Is the default `toString` the same as what we were using?

edmofro (approved):
All looks good if you can explain the `toString` to me

PR #2286, opened 2021-03-24
Add Supply Chain Fiji Project

nick-beyondessential:
### Changes:
- new project migration


PShoe (approved):

billli0:
Do we need to do this here: 

>   await insertObject(db, 'dashboardGroup', {
    organisationLevel: 'Project',
    userGroup: USER_GROUP,
    organisationUnitCode: PROJECT_CODE,
    dashboardReports: '{project_details}',
    name: DASHBOARDGROUP_NAME,
    code: DASHBOARDGROUP_CODE,
    projectCodes: `{${PROJECT_CODE}}`,
  });

nick-beyondessential:
Good catch @billli0 I thought i had removed it.

nick-beyondessential:
Oh you are asking if we need to add it? I don't think so - being only one country makes sense to jump to that country.

billli0:
Thanks @nick-beyondessential , looks good to me!

PR #2287, opened 2021-03-24
2177: Use tupaia service in Strive

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2177

This PR depends on the following features for correct functionality:
1. [1149: Correct internal data fetching time constraints to respect timezone](https://github.com/beyondessential/tupaia-backlog/issues/1149)
2. [2101: Analytics table](https://github.com/beyondessential/tupaia-backlog/issues/2101)
3. [2100: Use data_time field in analytics table](https://github.com/beyondessential/tupaia-backlog/issues/2100)
4. support for empty `dataElementCodes` when fetching tupaia events

All the above are included in the base branch. 1 and 2 are currently under testing. 3 and 4 depend on 1 and 2 so they will be reviewed and tested after their dependencies pass testing.

One can still review this PR, since none of its prerequisites affect the core migration logic.

### DIFF
--- a/packages/database/src/migrations/20210318031303-UseTupaiaServiceInStrive-modifies-data.js
+++ b/packages/database/src/migrations/20210318031303-UseTupaiaServiceInStrive-modifies-data.js
@@ -1,0 +1,279 @@
+'use strict';
+
+import { arrayToDbString } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const ENTITY_AGGREGATION_CONFIG = {
+  reports: {
+    PG_Strive_PNG_Weekly_Febrile_Cases_By_Village: {
+      dataSourceEntityType: 'case',
+      aggregationEntityType: 'village',
+    },
+    PG_Strive_PNG_Weekly_Number_of_Febrile_Cases: { dataSourceEntityType: 'case' },
+    PG_Strive_PNG_Weekly_Reported_Cases: { dataSourceEntityType: 'case' },
+    PG_Strive_PNG_RDT_Tests_Total_And_Percent_Positive: { dataSourceEntityType: 'facility' },
+    PG_Strive_PNG_Weekly_Percentage_of_Positive_Consultations: { dataSourceEntityType: 'facility' },
+    PG_Strive_PNG_Weekly_Number_of_Consultations: { dataSourceEntityType: 'facility' },
+    PG_Strive_PNG_Weekly_Percentage_of_Positive_Malaria_Against_Consultations: {
+      dataSourceEntityType: 'facility',
+    },
+    PG_Strive_PNG_Febrile_Cases_By_Age: { dataSourceEntityType: 'case' },
+    PG_Strive_PNG_Febrile_Cases_By_Sex: { dataSourceEntityType: 'case' },
+    PG_Strive_PNG_Febrile_Cases_By_Week: { dataSourceEntityType: 'case' },
+    PG_Strive_PNG_Case_Report_Form_Export: { dataSourceEntityType: 'case' },
+    PG_Strive_PNG_Weekly_Febrile_Illness_RDT_Positive_By_Facility_National: {
+      dataSourceEntityType: 'case',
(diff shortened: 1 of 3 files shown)


packages/database/src/migrations/20210318031303-UseTupaiaServiceInStrive-modifies-data.js
```diff
@@ -0,0 +1,279 @@
...
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const ENTITY_AGGREGATION_CONFIG = {
```
kael89, line 19: I placed all "standard" tasks that need to be done when converting a project to `tupaia` service in this single migration file, to make future conversions easier

packages/database/src/migrations/20210318031303-UseTupaiaServiceInStrive-modifies-data.js
```diff
@@ -0,0 +1,279 @@
...
+    },
+    STRIVE_FIS_Village_Percent_mRDT_Positive_PF_In_Week: {
+      dataSourceEntityType: 'case',
+      aggregationEntityType: 'village',
+    },
+    STRIVE_WTF_Consultations_Radius: { dataSourceEntityType: 'facility' },
+    STRIVE_WTF_mRDT_Tests_Radius: { dataSourceEntityType: 'facility' },
+    STRIVE_WTF_Positive: { dataSourceEntityType: 'facility' },
+  },
+};
```
chris-bes, line 126: I didn't really go through this list to check of the `dataSourceEntityType` or `aggregationEntityType` are correct, but I assume you've already smoke tested them 👍

chris-bes (approved):
LGTM!

packages/database/src/migrations/20210318031303-UseTupaiaServiceInStrive-modifies-data.js
kael89: Good call I believe, yes I've smoke tested them, and we'll do some thorough regression testing which will hopefully catch anything that may be wrong!

kael89:
@chris-bes I added a [new commit](https://github.com/beyondessential/tupaia/pull/2287/commits/7c13941881a53255aee7ee2586a2feb57ab89bd3), could you please re-review?

There are branches for all the dependencies mentioned in this PR's description now, so I rebased the branches in this PR accordingly.

packages/database/src/migrations/20210322030033-FixStriveAfterTupaiaServiceConversion-modifies-data.js
```diff
@@ -14,7 +14,17 @@ exports.setup = function (options, seedLink) {
...
-const cleanDhisMetadataForStriveCases = async db => {
+const deleteDhisSyncTableRecordsForStriveCases = async (db, table) =>
+  db.runSql(`
+    DELETE FROM ${table}
+    USING survey_response sr
+    JOIN survey s on s.id = sr.survey_id
+    WHERE
+      sr.entity_id = ${table}.record_id AND
+      s.code = 'SCRF';
+`);
```
chris-bes, line 25: Sorry if I'm being a bit pedantic, but I think not all survey responses are attached to case entity in `SCRF` survey (not sure why, but I think it could be a mistake somehow)...? 

When I run this query (converted from your delete statement):
```
SELECT distinct e.type from dhis_sync_log d
JOIN survey_response sr on sr.entity_id = d.record_id
JOIN survey s on s.id = sr.survey_id
JOIN entity e on e.id = sr.entity_id
WHERE s.code = 'SCRF'
```

It gives the result (expected only case, but not sure how a facility slipped through):
image

Tracing the facility entity sync records, it is only 1 facility:
facility

So I guess we don't want to delete the sync record of the facility entities If I'm correct... If so we can add a where clause `e.type = 'case'` to the delete statement?

kael89: Not pedantic at all, on the contrary thanks for giving this a thorough look! Yeah not sure why a `facility` slipped through, but it's good to keep it safe and add the extra condition you mentioned

chris-bes (approved):
Looks good!

kael89:
Merging in the base branch so that  [2636](https://github.com/beyondessential/tupaia-backlog/issues/2636) and [2177](https://github.com/beyondessential/tupaia-backlog/issues/2177) can be tested together

PR #2290, opened 2021-03-24
1149 widget parse formatted utc

EMcQ-BES:
### Changes:

- Pass a format through to the datetime-utc widget for parsing
- Revert previous change to date of data formatting on survey responses page


edmofro (changes requested):
As discussed, let's try to support an `editConfig` specific `accessor`, that will take precedence over the general `accessor` when grabbing the value in `Editor.js` for the `InputField`

edmofro (approved):

PR #2294, opened 2021-03-25
2494 lesmis entity server integration

tcaiger:
### Changes:

- Added Entity and Entities endpoint to Lesmis-Server
- Refactored Lesmis FrontEnd to use entity data
- renamed orgUnit everywhere to be entityCode

---


packages/lesmis/src/components/Breadcrumbs.js
```diff
@@ -36,38 +36,30 @@ const Link = props => <MuiLink color="inherit" {...props} component={RouterLink}
 
 const Loader = () => (
   <Skeleton animation="wave">
-    <MuiLink>Loading</MuiLink>
+    <MuiLink>breadcrumbsLoading</MuiLink>
```
tcaiger, line 39: this string just mocks the length of the skeleton loader. The text doesn't actually show.

packages/lesmis-server/src/connections/EntityConnection.ts
```diff
@@ -0,0 +1,26 @@
...
+import camelcaseKeys from 'camelcase-keys';
+import { ApiConnection } from '@tupaia/server-boilerplate';
+import { LESMIS_PROJECT_NAME, LESMIS_COUNTRY_ENTITY_CODE } from '../constants';
+
+const { ENTITY_SERVER_API_URL = 'http://localhost:8050/v1' } = process.env;
+
+export class EntityConnection extends ApiConnection {
+  baseUrl = ENTITY_SERVER_API_URL;
+
+  async getEntities() {
```
tcaiger, line 15: We are just getting all the data at the moment rather than specifying fields. This may need to change later depending on how the front end caching works out. Will add support for fields on an as needed basis.

packages/lesmis-server/src/connections/EntityConnection.ts
```diff
@@ -0,0 +1,26 @@
...
+    const response = await this.get(
+      `hierarchy/${LESMIS_PROJECT_NAME}/${LESMIS_COUNTRY_ENTITY_CODE}/descendants`,
```
edmofro, line 17:
```suggestion
      `hierarchy/${LESMIS_PROJECT_NAME}/${LESMIS_COUNTRY_ENTITY_CODE}/descendants?includeRootEntity=true`,
```
New addition in Rohan's PR. Needed here or it won't include the "Laos" country entity itself. Alternative would be to pass the lesmis project name as the entity as well, as then it will be the root entity and return everything underneath (which would include the country)

edmofro (changes requested):
Just one change request, otherwise this is getting tight as 👏

packages/lesmis-server/src/connections/EntityConnection.ts
tcaiger: cool. I have a slight preference for the project as the entity as it fits well conceptually for me. ie get all the entities for the lesmis project

edmofro (approved):

PR #2295, opened 2021-03-25
Allow project config of frontend exclude types 

nick-beyondessential:
### Changes:
- Check for project config frontendExcludedTypes else use entity default list `entity.typesExcludedFromWebFrontend` as before.
- Add migration to individual to frontendExcludedTypes for Covid Samoa


IgorNadj (approved):
LGTM

PR #2298, opened 2021-03-26
2489 lesmis login flow

tcaiger:
### Changes:

- Convert LESMIS login page to a dialog (design change)
- Handle cache clearing on login and logout

---


packages/lesmis-server/src/routes/LoginRoute.ts
```diff
@@ -9,7 +9,7 @@ import { LoginRoute as BaseLoginRoute } from '@tupaia/server-boilerplate';
 import { LESMIS_PERMISSION_GROUP } from '../constants';
 
 export class LoginRoute extends BaseLoginRoute {
-  verifyAuth(accessPolicy) {
+  verifyLoginAuth(accessPolicy) {
```
tcaiger, line 12: bug fix related to a missed server-boilerplate change

packages/lesmis/src/api/mutations/useLogin.js
```diff
@@ -21,6 +21,8 @@ export const useLogin = () => {
     {
       onSuccess: () => {
         queryClient.invalidateQueries('user');
+        queryClient.invalidateQueries('entity');
+        queryClient.invalidateQueries('entities');
       },
```
tcaiger, line 26: Clear the cached data for these endpoints after login so that the data relating to the new permissions is picked up.

packages/lesmis/src/components/ProfileButton.js
```diff
@@ -18,15 +19,17 @@ const StyledProfileButton = styled(BaseProfileButton)`
...
 export const ProfileButton = () => {
-  const { data: user } = useUser();
+  const { data: user, isSuccess } = useUser();
   const queryClient = useQueryClient();
 
   const handleLogout = async () => {
     await post('logout');
     queryClient.invalidateQueries('user');
+    queryClient.invalidateQueries('entity');
+    queryClient.invalidateQueries('entities');
```
edmofro, line 29: Because it's used in two places quite far apart, perhaps there should be some place that this group of "queries to invalidate on auth change" is held. Either just a static constant with the queries listed out 
```
const QUERIES_INVALIDATED_ON_LOGIN_CHANGE = ['user', 'entity', 'entities'];
```
or a util function called something like  `invalidateQueriesOnLoginChange`

It may even be a smell that `handleLogout` should move to `mutations` so they're grouped together

packages/lesmis/src/components/LoginDialog.js
```diff
@@ -0,0 +1,113 @@
...
+  border-radius: 2.5rem;
+`;
+
+const CloseButton = styled(IconButton)`
+  position: absolute;
+  top: 0;
+  right: 0;
+`;
+
+const requestAnAccountUrl = 'https://info.tupaia.org/contact';
```
edmofro, line 74: Good call!

edmofro (approved):
LGTM! One little suggestion but over to you on whether and how to follow up

packages/lesmis/src/components/ProfileButton.js
tcaiger: yeah good call, I added a useLogout mutation to capture the api logic in the api directory. I may add a list of constants later if the invalidation gets more complex.

PR #2306, opened 2021-03-29
2550: Cleaned up sql bind params in DhisApiProxy

chris-bes:
### Changes:

- Cleaned up sql bind params.


nick-beyondessential (approved):
Nice work Chris

PR #2311, opened 2021-03-29
2407: Refactor Entity-Server to use Server-Boilerplate

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2407:

Bit of a major refactor here, feel free to hit me up if you have any questions.

I've basically broken the `server-boilerplate` package into 2 main sections:
- micro-service: Boilerplate code to create tupaia micro-services (entity-server, report-server, etc.)
- orchestrator: Boilerplate to create tupaia orchestrator servers (psss-server, lesmis-server, web-config-server, etc.)

These two types of server seem to be a common pattern emerging in our codebase and have distinct qualities so I thought it best to separate them. 

Examples to how to use server-boilerplate to create servers:
micro-service:
```
import { MicroServiceApiBuilder, handleWith } from '@tupaia/server-boilerplate';

new MicroServiceApiBuilder(new TupaiaDatabase())
    .useBasicBearerAuth('entity-server') // Specifies auth type as Basic and Bearer (standard for our micro-services)
    .use<HierarchyRequest>('/v1/hierarchy/:hierarchyName/:entityCode', attachContext) // Add a middleware
    .get<HierarchyRequest>(
      '/v1/hierarchy/:hierarchyName/:entityCode',
      handleWith(SingleEntityRoute),
    ) // Adds a route
    .build();
    
// Note: the use of 'HierarchyRequest' as a generic parameter here allows us to cast what type of 
// request each route/middleware will be handing (and thus what params/query/etc. are available to it)
```

orchestrator:
```
import { OrchestratorApiBuilder, handleWith } from '@tupaia/server-boilerplate';

const database = new TupaiaDatabase();
return new OrchestratorApiBuilder(database)
    .useSessionModel(new LesmisSessionModel(database)) // Basically just sets which table to use as the session model table
    .verifyLogin(verifyLogin) // Function to check the user is a valid login, can be replaced by middleware for various routes
    .get('/v1/user', handleWith(UserRoute))
    .get<EntityRequest>('/v1/entity/:entityCode', handleWith(EntityRoute))
    .build();
```
The OrchestratorApiBuilder adds the login/logout routes, and handles all the session logic such that any middleware/route added to it will automatically have the session object attached to the request (or error thrown if no session exists).


packages/entity-server/src/@types/express/index.d.ts
```diff
@@ -3,20 +3,19 @@
...
       accessPolicy: AccessPolicy;
-      authenticator: Authenticator;
       models: EntityServerModelRegistry;
-      context: Context;
+      ctx: Context;
     }
 
     export interface Response {
-      context: Context;
+      ctx: Context;
```
rohan-bes, line 18: We need these `index.d.ts` declaration files to extend the fields available on the standard request object. Using standard `interface OurRequest extends Request { ourField: string};` runs in to a bunch of problems when trying to use express library functions.

I tried really hard to see if we could share a single `index.d.ts` between various packages but it just seemed impossible 😢 

... but fear not! While it does add a bit of extra boilerplate (the whole thing we've been trying to get away from hehe) it's actually proven to be really neat to be able to specify these in a per-package kind of way. For instance, the `models` field on requests is actually used by both the `server-boilerplate` package and the `entity-server` package. But these two packages are interested in different types of models. By being able to type cast this object per package, we can restrict the scope of the models object that each package is aware of to just the database tables that matter to them 👍

packages/entity-server/src/app/createApp.ts
```diff
@@ -3,53 +3,30 @@
...
-  /**
-   * Add all routes to the app
-   */
-  addRoutesToApp(app);
-
-  return app;
+export function createApp() {
+  return new MicroServiceApiBuilder(new TupaiaDatabase())
+    .useBasicBearerAuth('entity-server')
+    .use<HierarchyRequest>('/v1/hierarchy/:hierarchyName/:entityCode', attachContext)
```
rohan-bes, line 22: Technically it may be possible to do away with the explicit typecasting here because we ought to be able to infer that `attachContext` is using a `HierarchyRequest` type request... however we're already doing a lot of fancy things here and I couldn't find a simple way to do it... 

Personally I don't mind it so much as this reads nicely to me as: Use this request, for this path, handled by this logic

packages/entity-server/src/routes/hierarchy/EntityDescendantsRoute.ts
```diff
@@ -13,23 +13,24 @@ import {
...
-    HierarchyRequestQuery & { includeRootEntity?: boolean }
-  >,
+  DescendantsRequest,
   HierarchyResponse<EntityResponseObject[]>
 > {
   async buildResponse() {
     const { includeRootEntity = false } = this.req.query;
-    const descendants = await this.req.context.entity.getDescendants(this.req.context.hierarchyId, {
-      country_code: this.req.context.allowedCountries,
+    const descendants = await this.req.ctx.entity.getDescendants(this.req.ctx.hierarchyId, {
```
rohan-bes, line 28: I blanket renamed `context` to `ctx` cos I've seen it around and it's quicker to write 😋 Does it read easy to you?

packages/lesmis-server/examples.http
```diff
@@ -0,0 +1,30 @@
...
+content-type: {{contentType}}
+
+{
+    "emailAddress": "{{email}}",
+    "password": "{{password}}"
+}
+
+###
+
+@sessionCookie = {{login.response.headers.Set-Cookie}}
```
rohan-bes, line 24: This is cool, it allows u to use the `login` request to get the session cookie and the pass it on in other requests 👍

packages/server-boilerplate/src/routes/Route.ts
```diff
@@ -1,36 +1,37 @@
...
+import { respond } from '@tupaia/utils';
+
+// Infers type arguments from request type
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export type Params<Type> = Type extends Request<infer P, any, any, any> ? P : never;
+export type ResBody<Type> = Type extends Request<any, infer sBody, any, any> ? sBody : never;
+export type ReqBody<Type> = Type extends Request<any, any, infer qBody, any> ? qBody : never;
+export type Query<Type> = Type extends Request<any, any, any, infer Q> ? Q : never;
+export type MatchingRequest<Req> = Request<Params<Req>, ResBody<Req>, ReqBody<Req>, Query<Req>>;
+export type MatchingResponse<Req> = Response<ResBody<Req>>;
```
rohan-bes, line 15: This is the generics hell that Express has put me through. I've tried my best to find a way to define `Route` as being able to take a type that extends `Request` but shares `Request`'s generic parameters. It feels odd to me that typescript doesn't have a good syntax to support this... Atleast all this complexity is now encapsulated in the `server-boilerplate package` 👍

packages/server-boilerplate/src/types.ts
```diff
@@ -3,76 +3,10 @@
...
-export interface TupaiaResponseBody {
-  accessPolicy?: AccessPolicyObject;
-  error?: string;
-}
-
-export interface Credentials {
-  emailAddress: string;
-  password: string;
-  deviceName: string;
-}
```
rohan-bes, line 36: Most of these types either got moved to being defined inline with the code that consumes them, or removed all together

packages/server-boilerplate/src/utils/handleWith.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ */
+import { Response, NextFunction } from 'express';
+import { Route } from '../routes';
+
+type ReqOfRoute<T> = T extends Route<infer Req> ? Req : never;
+// type ResOfRoute<T> = T extends Route<any, infer Res> ? Res : never;
+export const handleWith = <T extends { handle: () => Promise<void> }>(
+  RouteClass: new (req: ReqOfRoute<T>, res: Response, next: NextFunction) => T,
+) => (req: ReqOfRoute<T>, res: Response, next: NextFunction) => {
```
rohan-bes, line 13: More generics madness 😝 I think I have a problem...

`T` here is a Route subtype, eg. `LoginRoute`. The `new (...) => T` syntax is pretty magical, it implies that the parameter is a class which when instantiated creates a Route of type `T`.

tsconfig-ts.dependency.json
```diff
@@ -0,0 +1,20 @@
...
+    "module": "CommonJS",
+    "noEmit": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "ES2019",
+    "resolveJsonModule": true,
+    "sourceMap": true,
+    "baseUrl": "./packages",
+    "paths": {
+      "@tupaia/*": ["*/src"]
```
rohan-bes, line 15: This single config has changed my life ❤️ it allows our ts packages to do source mapping on our internal packages without requiring a rebuild. Super handy!

packages/lesmis-server/src/app/createApp.ts
```diff
@@ -3,38 +3,22 @@
...
-  return app;
+export function createApp() {
+  const database = new TupaiaDatabase();
+  return new OrchestratorApiBuilder(database)
+    .useSessionModel(new LesmisSessionModel(database))
+    .verifyLogin(verifyLogin)
+    .get('/v1/user', handleWith(UserRoute))
+    .get('/v1/entities', handleWith(EntitiesRoute))
+    .get<EntityRequest>('/v1/entity/:entityCode', handleWith(EntityRoute))
+    .build();
```
rohan-bes, line 23: @tcaiger I think you might be interested in this. So we no longer use `addRoutesToApp()` and instead we define all the routes here.

tcaiger: Hey Rohan, I have had quick look through these changes and it looks awesome! I'm glad to see a backend expert improving the server-boilerplate package.

tcaiger: Is my understanding correct that we can do .use() in this chain to extend the functionality of the server when using it? And do you know if it would work to re-use an already used module. ie `.use(bodyParser.json({ limit: '10mb' }))`?

tcaiger: Also this is probably a mis-informed question, but do we need to apply the handleWith wrapper here or could that be done in the boilerplate code now?

rohan-bes: > Is my understanding correct that we can do .use() in this chain to extend the functionality of the server when using it?
Yup, that's correct, `use()` will add a middleware layer to the specified path. As to whether you can overwrite an existing module (ie. bodyParser), I don't think so... If we wanna make this configurable I reckon we ought to break it out of the constructor into its own function, eg.
```
setBodyParserLimit(limit: string) {
  this.app.use(bodyParser.json({ limit }));
  return this;
}
```

> Also this is probably a mis-informed question, but do we need to apply the handleWith wrapper here or could that be done in the boilerplate code now?
Not mis-informed at all, you're totally right that would be ideal 😄. However, we are playing with some pretty complex generics here and I simply was able to find a way to coalesce the method signature of `get/use/etc.` to accept a `Route` class type and then apply the `handleWith` wrapper itself. I'm inclined to go back and give it another shot at some point since it definitely should be possible, it's really just cos my typescript skills are still pretty basic

packages/entity-server/src/routes/hierarchy/EntityDescendantsRoute.ts
kael89: They are fine either way! With short names, I've found that it's largely a matter of getting used to them (although they become hard to read when everything is shortened)

packages/server-boilerplate/src/routes/Route.ts
kael89: Yeap at least it's good that you moved this to a central location! Even as is, the verbosity is worth the extra type safety

packages/entity-server/src/app/createApp.ts
kael89: Agreed! We can always refactor it in the future if either out TS skills or TS itself gets enhanced!

packages/entity-server/src/@types/express/index.d.ts
kael89: Luckily it's not much boilerplate, since we can still load types for each field (eg `ModelRegistry` from `@tupaia/database` for `Request.models`)! So all good 😃

packages/server-boilerplate/src/utils/handleWith.ts
kael89: Nice one! This syntax seems pretty handy. Thanks for the explanation!

tsconfig-ts.dependency.json
kael89: Awesome find man!!!

kael89: It's also super neat that we just had to update this in only place

packages/server-boilerplate/src/micro-service/api/ApiBuilder.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  useBasicBearerAuth(apiName: string) {
+    const authenticator = new Authenticator(this.models);
+    this.app.use(buildBasicBearerAuthMiddleware(apiName, authenticator));
+    return this;
+  }
+
+  use<T extends MatchingRequest<T> = Request>(
+    path: string,
+    middleware: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>,
```
kael89, line 67: I guess that's what you've mentioned in one of your comments: it's not possible to deduce `T` from the passed in params (in this case, middleware), or rather it's too complicated/verbose?

it's fine to use explicit generics (`use<T>`), just trying to get a better understanding of our options 😃

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -0,0 +1,124 @@
...
+import express, { Express, Request, Response, NextFunction, RequestHandler } from 'express';
+import cors from 'cors';
+import bodyParser from 'body-parser';
+import errorHandler from 'api-error-handler';
+
+import { ModelRegistry, TupaiaDatabase } from '@tupaia/database';
+import { AccessPolicy } from '@tupaia/access-policy';
+import { UnauthenticatedError } from '@tupaia/utils';
+
+import { handleWith, handleError } from '../../utils';
```
kael89, line 15: By the way, I like your separation of imports in 3rd party modules/@tupaia packages/package modules. I've been combining the last 2 in my code, but perhaps I should start doing what you do here!

packages/server-boilerplate/src/orchestrator/models/Session.ts
```diff
@@ -8,11 +8,43 @@
...
+  refresh_token: string;
+}
+
+const extractTokenDetails = (decodedToken: string | Record<string, unknown> | null) => {
+  if (decodedToken === null || typeof decodedToken === 'string') {
+    throw new Error('Got unexpected result from decoded jwt token');
+  }
+
+  if (!('exp' in decodedToken) || !('iat' in decodedToken)) {
+    throw new Error('Decoded jwt token missing expected fields');
```
kael89, line 36: Nice validation!

packages/server-boilerplate/src/orchestrator/routes/LogoutRoute.ts
```diff
@@ -0,0 +1,26 @@
...
+  async buildResponse() {
+    const { sessionCookie } = this.req;
+    const sessionId = sessionCookie?.id;
+
+    if (sessionId && this.req.sessionModel) {
+      await this.req.sessionModel.deleteById(sessionId);
+    }
+
+    if (sessionCookie !== undefined) {
+      sessionCookie.reset?.();
```
kael89, line 21: You keep pulling nice tricks from your JS/TS hat Rohan 😄 Does this mean that `reset` will be called, only if it is defined **and** a function?

packages/server-boilerplate/src/routes/Route.ts
```diff
@@ -1,36 +1,37 @@
...
+/* eslint-enable @typescript-eslint/no-explicit-any */
+
+export class Route<
+  Req extends MatchingRequest<Req> = Request,
```
kael89, line 19: My understanding is that we use `MatchingRequest`/`MatchingResponse` to match the passed in types to `express` types (although I may be wrong).

If that's indeed the case, perhaps we should mention `Express` in the types to make our intentions clearer, eg `ExpressRequest`/`ExpressResponse`

kael89 (approved):
Everything looks awesome @rohan-bes !! Separating `orchestration` and `microService` logic was very inspired. Also, you've put some really nice work with TS types, I learned a lot syntax from this PR! 👏 

Moreover, I love the declarativeness of the new `ApiBuilder` classes - I am a big fan of "fluent interfaces" (chained methods) ❤️ 

I have a few questions and suggestions but nothing that really blocks this card from being merged.

One generic comment is about imports: you seem to avoid the `*` syntax (either purposefully or not). It has its pros and cons, but overall I find myself to tend on using it more and more. If you are concerned about name clashing: JS will throw an error if two imports included in `*` have the same name, and it's usually easy to spot those.

packages/server-boilerplate/src/types.ts
```diff
@@ -3,76 +3,10 @@
...
-  id: string;
-  email: string;
-  access_policy: AccessPolicyObject;
-  access_token: string;
-  access_token_expiry: number;
-  refresh_token: string;
-}
+export type Context<T> = {
+  [field in keyof T]: T[field];
+};
```
kael89, line 12: I would assume that we could just use `Record<string, unknown>` directly where we currently use `Context`, or perhaps just define it as `type Context = Record<string, unknown`.

But it may be that you do here is required for proper type inheritance - could you please enlighten me? 🤓

kael89 (approved):
A few more questions/comments

tsconfig-ts.dependency.json
```diff
@@ -0,0 +1,20 @@
+{
```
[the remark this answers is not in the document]
rohan-bes: Yup, I agree with all those point! 👍

tsconfig-ts.dependency.json
```diff
@@ -0,0 +1,20 @@
...
+    "target": "ES2019",
+    "resolveJsonModule": true,
+    "sourceMap": true,
+    "baseUrl": "./packages",
+    "paths": {
+      "@tupaia/*": ["*/src"]
+    }
+  },
+
+  "exclude": ["node_modules", "dist"]
```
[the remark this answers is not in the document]
rohan-bes: Good point, will remove

packages/server-boilerplate/src/index.ts
```diff
@@ -4,7 +4,14 @@
...
 
-export * from './connections';
-export * from './models';
-export * from './routes';
-export * from './utils';
+export { ApiConnection } from './connections';
+export { Route } from './routes';
+export { handleWith } from './utils';
+export { Context } from './types';
+export { ApiBuilder as MicroServiceApiBuilder } from './micro-service';
```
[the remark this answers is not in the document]
rohan-bes: Hmmm I don't mind changing to `microService`... agree its not too much of a big deal but I think you are right to it should be changed

packages/server-boilerplate/src/micro-service/api/ApiBuilder.ts
rohan-bes: Yes, it ought to be possible. What really makes this whole generics code so complicated is because we are working with generics of generics (eg. A generic route, which in turn takes a generic request as a parameter). It seems that typescript doesn't seem to do any kind of 'deep infer' logic, so you need to pass through the nested generic parameters all the way through... it gets very complicated and typescript complains about how complex the type signatures are getting...
But yeah, if we invest the time there should be a way to do it nicely

packages/server-boilerplate/src/orchestrator/routes/LogoutRoute.ts
rohan-bes: Technically I don't think it checks if it's a function, just if `sessionCookie` has a property called `reset`. Should be safe tho 😄

packages/server-boilerplate/src/routes/Route.ts
rohan-bes: Good point, I think that is a bit more clear 😄

packages/lesmis-server/src/app/createApp.ts
```diff
@@ -3,38 +3,22 @@
...
-  addRoutesToApp(app);
-
-  return app;
+export function createApp() {
+  const database = new TupaiaDatabase();
+  return new OrchestratorApiBuilder(database)
+    .useSessionModel(new LesmisSessionModel(database))
+    .verifyLogin(verifyLogin)
+    .get('/v1/user', handleWith(UserRoute))
+    .get('/v1/entities', handleWith(EntitiesRoute))
```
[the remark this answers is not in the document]
rohan-bes: Nope, since those routes use the default `Request` type and I'm using defaults in most of the generics here

packages/entity-server/src/app/createApp.ts
```diff
@@ -3,53 +3,30 @@
...
+    .get<HierarchyRequest>(
+      '/v1/hierarchy/:hierarchyName/:entityCode',
+      handleWith(SingleEntityRoute),
+    )
+    .get<DescendantsRequest>(
```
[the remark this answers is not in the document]
rohan-bes: The middleware will actually match all routes under the `/v1/hierarchy/:hierarchyName/:entityCode` path, so it is being used here 👍

packages/server-boilerplate/src/types.ts
rohan-bes: Tbh I'm not too sure, I'll check since that would be cleaner... 👍

tsconfig-ts.dependency.json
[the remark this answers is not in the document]
kael89: You may also have to add them in the package `tsconfig.json` files

packages/server-boilerplate/src/index.ts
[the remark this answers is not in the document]
kael89: Let's go for it then - it should be easy to do a simple rename in a single commit

packages/server-boilerplate/src/micro-service/api/ApiBuilder.ts
kael89: Sounds good, thanks for the explanation! It's already nice as is 😃

packages/server-boilerplate/src/orchestrator/routes/LogoutRoute.ts
kael89: You are correct! At a first glance, I initially expected a `sessionCookie?.reset()` syntax for that, but I guess I would be wrong, since

```js
const a = {};
console.log(a.test); // undefined - no error thrown
console.log(a.test()); // throws error: a.test is not a function
```

So the syntax I expected would not help if `reset` itself is not defined, since calling an `undefined` function is invalid!

[Mozilla docs - Optional chaining with function calls](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

PR #2312, opened 2021-03-29
UI Components Bundling

tcaiger:
### Changes:

- Replaced webpack with rollup to enable tree shaking
- Split ui components build into 3 smaller bundles

Tree shaking isn't as effective as I thought because the bundlers are not able to 100% determine which parts of the code are dead/unused. When I added the Chart and Map components the total bundle that the apps produce was 5 MB+. So as a work around I have split ui components into the following 3 bundles to ensure large amounts of dead code are not bundled with the apps.

- core ui components
- map components
- chart components

The core ui components can be imported as:
`import { Buttton } from @tupaia/ui-components`

The map and chart components are inside a lib directory and need to be imported like this:
`import { Legend } from @tupaia/ui-components/lib/map`
`import { Chart } from @tupaia/ui-components/lib/chart`

One question is whether it is worth splitting out the Map and Chart components into their own package in the mono-repo? Or maybe just pulling them out of src into another directory in the ui-components package?

### Testing Results
After some experimentation I figured out that the best way to test is to look at the bundles from the apps that consume the components. The size of the bundle from ui-components its self doesn't tell the whole story because there is tree shaking, code elimination and dependency injection downstream.

Here are the results from the 2 packages that currently use ui-components. The results include Chart components but not Map Components. It is about a 25% decrease in size.

  | Webpack Build | New Rollup Build
-- | -- | --
Admin Panel | 4.8 MB | 3.65 MB
PSSS | 5.9 MB | 4.8 MB


packages/ui-components/rollup.config.js
```diff
@@ -0,0 +1,67 @@
...
+  }),
+  commonjs(),
+  analyze({ summaryOnly: true }),
+];
+
+const externals = [
+  ...Object.keys(pkg.devDependencies || {}),
+  ...Object.keys(pkg.dependencies || {}),
+  /@material-ui/,
+  /date-fns/,
```
edmofro, line 36: Why do these two need to be separately listed?

edmofro (approved):
I don't love importing from `/lib`, but it's a good enough solution for now. Nice work getting through rollup hell!

packages/ui-components/rollup.config.js
tcaiger: Yeah I don't love it either - I'm open to future improvements for sure! Those patters for @material-ui etc are needed to make sure that imports that don't match the name from dependencies exactly are picked up. Eg. `@material-ui/core/button` or `date-fns/locale/la` and so on.

edmofro: Cool makes sense!

PR #2313, opened 2021-03-29
Add lesmis front end deploy steps

edmofro:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2582

As well as making the changes in code here, I followed some manual steps:
- Bought url for lesmis.la via Crazy Domains
- Added a Hosted Zone on Route53 for lesmis.la, with two A records pointing to the tupaia.org IP
- Forwarded DNS hosting to Route53 by setting the Name Servers on Crazy Domains 
- Altered the certbot-tupaia user through AWS IAM to allow it access to the lesmis.la Hosted Zone
- Obtained an SSL certificate by ssh'ing into production and running `AWS_PROFILE=certbot certbot-auto certonly -d *.lesmis.la -d lesmis.la --dns-route53 -m edwin@beyondessential.com.au --agree-tos --non-interactive --server https://acme-v02.api.letsencrypt.org/directory`
- Added environment variables to AWS Parameter store for both dev and production, as per https://docs.beyondessential.com.au/books/dev-ops/page/updating-environment-variables
- Added `lesmis` to the subdomains tag against the production and development EC2 instances

I think that should get LESMIS all set up for hosting on both lesmis.tupaia.org and lesmis.la


packages/devops/scripts/ci/validateBranchName.sh
```diff
@@ -24,7 +24,7 @@ function get_branch_name() {
 }
 
 function check_name_is_not_reserved() {
-    local $branch_name=$1
+    local branch_name=$1
```
edmofro, line 27: Noticed this when testing my changes

tcaiger (approved):
Looks good to me. My only comment is do we need to add "lesmis" to the updateCiEndVars.sh script?

edmofro:
Good question but nope, only stuff that codeship needs should go there. Codeship isn't currently running any tests against lesmis so doesn't need any related env vars :-)

PR #2316, opened 2021-03-30
2584: Make all Laos Schools public

EMcQ-BES:
### Changes:

- Add migration


edmofro (approved):
Very tidy migration!

One little convention we've developed - we generally pass `db`/`models`/other "big" objects as the first parameter(s). But it's not at all gonna hold this up :-)

PR #2319, opened 2021-03-30
Lesmis content pages

tcaiger:
### Changes:

- Refactor breadcrumbs
- Add Content Pages

---

### Screenshots:
See issue


packages/lesmis/src/views/PageView.js
```diff
@@ -0,0 +1,61 @@
...
+  font-size: 28px;
+  line-height: 42px;
+`;
+
+export const PageView = ({ content }) => {
+  const { title, body, urlSegment } = content;
+  const breadcrumbs = [{ name: title, code: urlSegment }];
+
+  return (
+    <MuiThemeProvider theme={contentPageTheme}>
```
tcaiger, line 41: In the designs the typography has it's own set of styles for content pages so I have added a theme to model that.

packages/lesmis/src/constants/page.js
```diff
@@ -0,0 +1,44 @@
+/*
```
tcaiger, line 1: This is obviously pretty basic and wouldn't scale that well if the complexity of the pages were to increase but given the simplicity of the current content pages, I think it works pretty well.

packages/lesmis/src/theme/contentPageTheme.js
```diff
@@ -0,0 +1,75 @@
+/*
```
edmofro, line 1: Would be nice if there were a way to "merge modifications" from one theme into a base theme (this has some overlap with the main LESMIS theme)

packages/lesmis/src/views/PageView.js
edmofro: Good call, though I wonder if intentional by @David-G-UX?

packages/lesmis/src/views/PageView.js
```diff
@@ -0,0 +1,61 @@
...
+
+const Title = styled(Typography)`
+  font-weight: 600;
+  font-size: 1.75rem;
+  line-height: 2.6rem;
+`;
+
+export const PageView = ({ content }) => {
+  const { title, body, urlSegment } = content;
+  const breadcrumbs = [{ name: title, code: urlSegment }];
```
edmofro, line 38: Would remove the `code` here, I don't think it's used for anything? Will just need to add a condition check in the breadcrumbs component to not show an entity link if it's the last in the list _or_ has no code

edmofro (changes requested):
Just one slightly awkward feeling about:
- The breadcrumbs referring to "entity" all over the show where they're now being used for non-entity pages
- The urlSegment for '/about' and '/contact' being repeated in two quite distant places

But I think just allowing the Breadcrumbs to take in "title only" pages, that don't include an entity code, will solve both of those qualms

tcaiger:
@edmofro - This is ready for re-review. I have refactored the `breadcrumbs` to use name and urlSegment and removed the concept of entityCode from them and the `useBreadcrumbs` hook us now called `useEntityBreadcrumbs` to reflect their entity context. I also tightened up the `PageContentTheme` a bit to re-use all the properties from the main them apart from typography settings.

packages/lesmis/src/components/Breadcrumbs.js
```diff
@@ -40,49 +40,36 @@ const Loader = () => (
...
+    <Link to="/">Home</Link>
+    {isLoading ? (
+      <Loader />
+    ) : (
+      breadcrumbs.map(({ name, urlSegment }, index) => {
+        const last = index === breadcrumbs.length - 1;
+        return last ? (
+          <ActiveSegment key={urlSegment}>{name}</ActiveSegment>
+        ) : (
+          <Link to={makeEntityLink(urlSegment)} key={urlSegment}>
```
edmofro, line 54: This is the last entity specific part of breadcrumbs - might as well get rid of it now while you're at it?

tcaiger: @edmofro - done

edmofro (approved):

PR #2321, opened 2021-03-30
2311: Added /relations route for entity-server

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2311:

Added the `/relations` endpoint to the entity-server to allow for fetching for ancestor->descendant maps (primarily used in entity-aggregation).

Few things to note:
- The response object can be either grouped by `ancestor` or `descendant` based on a query parameter
- The response object can be flattened down to a map of codes (commonly this is all we are interested in, and it can yield an up to 30% performance boost in some scenarios)

Performance seems acceptable, taking ~0.5 seconds to load all sub_district -> school relations for Laos Schools once the hierarchy has been cached


packages/entity-server/src/routes/hierarchy/EntityDescendantsRoute.ts
```diff
@@ -19,10 +18,7 @@ export type DescendantsRequest = HierarchyRequest<
   HierarchyRequestBody,
   HierarchyRequestQuery & { includeRootEntity?: boolean }
 >;
-export class EntityDescendantsRoute extends Route<
-  DescendantsRequest,
-  HierarchyResponse<EntityResponseObject[]>
-> {
```
rohan-bes, line 25: Turns out that specifying the Response type was unnecessary (as it was inferred from the Request type anyway) so removing for simplicity 👍

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
```diff
@@ -0,0 +1,150 @@
...
+  | (EntityResponseObject & { descendants: EntityResponseObject[] })[] // groupBy: ancestor, flattenToCode: false
+  | (EntityResponseObject & { ancestor: EntityResponseObject })[] // groupBy: descendant, flattenToCode: false
+  | Record<string, string[]> // groupBy: ancestor, flattenToCode: true
+  | Record<string, string>, // groupBy: descendant, flattenToCode: true
+  HierarchyRequestBody,
+  HierarchyRequestQuery & {
+    ancestorType?: string;
+    descendantType?: string;
+    groupBy?: 'ancestor' | 'descendant';
+    flattenToCode?: 'true' | 'false';
```
rohan-bes, line 28: Thoughts on the naming of these params `flattenToCode` and `groupBy`? Couldn't come up with anything better but they don't feel quite right to me either...

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const ancestorCodeToDescendantCodes = reduceToArrayDictionary(
+      await this.buildDescendantAncestorPairs(descendants, ancestorType),
+      'ancestor',
+      'descendant',
+    );
+
+    if (flattenToCode) {
+      return ancestorCodeToDescendantCodes;
+    }
```
rohan-bes, line 72: Arguably the flattenToCode logic could be used in other endpoints as well, but I don't think it's as relevant for them... we can add it if it turns out to be useful 👍

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
kael89: We could rename `flattenToCode` to just `flat`, whose default behaviour is using `code` fields. This allows for other fields to be used when flatting in the future - with the addition of an extra param, eg `field`/`flatField`

We may never need to do that, but it's good to have the option!

kael89: I agree that `groupBy` doesn't feel exactly right, although it's not too bad either. I'll read the code and try to think of other options!

packages/database/src/modelClasses/Entity.js
```diff
@@ -363,6 +363,13 @@ export class EntityModel extends DatabaseModel {
     );
   }
 
+  /**
+   * Fetches descendant => ancestor map in given hierarchy
+   * @param {string[]} descendantCodes
+   * @param {string} hierarchyId
+   * @param {string} ancestorType
+   * @returns {Promise<Record<string, { code: string, name: string }>>} Map of descendant code to ancestor (code, name)
```
kael89, line 371: Unfortunately I don't think that JSDocs support the `Record` type. But we can use `Object` instead:

```suggestion
   * @returns {Promise<Object<string, { code: string, name: string }>>} Map of descendant code to ancestor (code, name)
```

I believe I checked this the other day, but it would be good to confirm that this is indeed the case - using `Record` would be handy!

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
```diff
@@ -0,0 +1,150 @@
...
+          });
+
+    const formattedEntitiesByCode = await this.buildFormattedEntitiesByCode(ancestors, descendants);
+    return descendants.map(descendant => ({
+      ...formattedEntitiesByCode[descendant.code],
+      ancestor: formattedEntitiesByCode[descendantCodeToAncestorCode[descendant.code]],
+    }));
+  }
+
+  async buildResponse() {
```
kael89, line 125: I find it convenient to declare class methods top/down: the base calling method first, then the ones it calls etc. Based on this, `buildResponse` would be placed on the top of this class

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
```diff
@@ -0,0 +1,150 @@
...
+    const {
+      ancestorType,
+      descendantType,
+      groupBy = 'ancestor',
+      flattenToCode: flattenToCodeQueryParam,
```
kael89, line 130: A convention I personally use is adding an `Input` suffix, so here this would be `flattenToCodeInput`.

This is not standard across our codebase though, so up to you!

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
```diff
@@ -0,0 +1,150 @@
...
+    );
+
+    return Object.entries(descendantAncestorMapping).map(([descendant, { code: ancestor }]) => ({
+      descendant,
+      ancestor,
+    }));
+  }
+
+  async buildFormattedEntitiesByCode(ancestors: EntityType[], descendants: EntityType[]) {
+    const allEntities = [...ancestors, ...descendants];
```
kael89, line 48: Combining ancestors and descendants should be the caller's responsibility:

```suggestion
  async buildFormattedEntitiesByCode(entities: EntityType[]) {
```

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
```diff
@@ -0,0 +1,150 @@
...
+    const formattedEntities = await this.req.ctx.formatEntitiesForResponse(allEntities);
+    const formattedEntitiesByCode: Record<string, EntityResponseObject> = {};
+    allEntities.forEach((unformattedEntity, index) => {
+      formattedEntitiesByCode[unformattedEntity.code] = formattedEntities[index];
+    });
+    return formattedEntitiesByCode;
```
kael89, line 54: Using [lodash.keyBy](https://lodash.com/docs/4.17.15#keyBy):

```suggestion
    return keyBy(formattedEntities, 'code');
```

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    const ancestorsWithDescendantsCodes = Object.values(descendantCodeToAncestorCode);
+    const ancestors =
+      entity.type === ancestorType
+        ? [entity]
+        : await entity.getDescendants(hierarchyId, {
+            code: ancestorsWithDescendantsCodes,
+            country_code: allowedCountries,
+          });
```
kael89, line 116: This logic is the same in `groupByAncestorResponse`, so you could probably extract it to a `getRequestedAncestors` method or similar

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
kael89, line 28: Probably worth extracting some of those long types to new local types

kael89 (changes requested):
Great work @rohan-bes !! I can see how this will be useful in heaps of places 🚀 

A few changes requested, but nothing major

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
kael89: Couldn't think of anything better as of yet - but `groupBy` doesn't sound too bad!

packages/database/src/modelClasses/Entity.js
rohan-bes: Using `Record` worked fine for me 😄

kael89: Great!!

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
kael89: This is unresolved but not a big deal - a detail really.

packages/entity-server/src/types.ts
```diff
@@ -12,4 +12,31 @@ export interface EntityServerModelRegistry extends ModelRegistry {
   readonly entityHierarchy: EntityHierarchyModel;
 }
 
-export type Resolved<T> = T extends Promise<infer R> ? R : T; // Returns resolved type if type is promise
+// Returns resolved type if type is promise
+export type Resolved<T> = T extends Promise<infer R> ? R : T;
+
+// Extracts keys that have object-like values from type T
+export type ObjectLikeKeys<T> = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kael89, line 20: Pending suggestion: use `unknown` instead of `any` (if possible)

packages/entity-server/src/routes/hierarchy/relations/types.ts
```diff
@@ -0,0 +1,44 @@
...
+export type RelationsSubQuery = Omit<HierarchyRequestQuery, 'fields'>;
+export type RelationsQuery = RelationsSubQuery & {
+  groupBy?: 'ancestor' | 'descendant';
+} & Partial<Prefix<RelationsSubQuery, 'ancestor'>> &
+  Partial<Prefix<RelationsSubQuery, 'descendant'>>;
+
+type RelationsSubContext = {
+  filter: EntityFilter;
+  field: keyof FlattableEntityFields;
+  type: string;
```
kael89, line 30: Should `type` be included in `filter`?

kael89 (approved):
Some great code ready to be merged! I learned a lot by reviewing your PRs, both about TypeScript and about api design!! 💯 👏

kael89:
@rohan-bes I left a few comments both nothing worth blocking this feature's merge

packages/entity-server/src/routes/hierarchy/EntityRelationsRoute.ts
rohan-bes: Funnily I actually got a contradicting recommendation from @edmofro that we prefer to put all functions called by another function above that function (admittedly that was not for class but a standard script file). I don't think I have too much of a preference either way 🙂

kael89: I have discussed this with Edwin in the past, and I believe the conclusion was: top-down structure for classes, bottom-up for functions, right @edmofro ?

Perhaps worth quickly mentioning in dev-chat to get it across the team

edmofro: That's right @kael89!

And yes, I think worth bringing up at dev chat

PR #2324, opened 2021-03-31
2496 lesmis entity browse

tcaiger:
### Changes:

- Add Entity Browser Menu
- Add Loading States to Search, Entity Browse and Home
- Update login and logout cache logic

Note: I haven't implemented the skeleton loader as the way the search currently loads it isn't possible to get to the Entity Browser Menu in a loading state.

---

### Screenshots:
See issue


packages/lesmis/src/api/mutations/useLogin.js
```diff
@@ -20,9 +20,9 @@ export const useLogin = () => {
       }),
     {
       onSuccess: () => {
-        queryClient.invalidateQueries('user');
-        queryClient.invalidateQueries('entity');
-        queryClient.invalidateQueries('entities');
+        queryClient.resetQueries('user');
```
tcaiger, line 23: This makes sure that the cached data against the query is not temporarily served.

packages/lesmis/src/components/SearchBar/SearchBar.js
```diff
@@ -180,51 +177,53 @@ export const SearchBar = ({ linkType, className }) => {
   const showNoResults = popupOpen && inputValue && focused && !dirty;
 
   return (
-    <SearchContainer className={className}>
```
tcaiger, line 183: All that's happening here is the search component getting wrapped in the Fade component. No other change

packages/lesmis/src/components/EntityMenu.js
```diff
@@ -21,13 +31,131 @@ const TextButton = styled(MuiButton)`
...
+  }
+`;
+
+const ListItemLink = props => <ListItemText component={RouterLink} {...props} />;
+
+const getEntitiesByCodes = (entities, codes) => entities.filter(e => codes.includes(e.code));
+
+const ListItemComponent = React.memo(({ entities, entity, onMenuClose }) => {
+  const [open, setOpen] = useState(false);
+  const hasChildren = Array.isArray(entity.childCodes);
```
rohan-bes, line 103: Are we doing any translation of the entity object when pulling from the `entity-server`? Otherwise I think this might be: `entity.child_codes`?

rohan-bes (approved):
All looks good to me, just one q about the formatting of the entity object but approving anyway 👍

packages/lesmis/src/components/EntityMenu.js
tcaiger: Yes we are doing translation to camel case in lesmis-server connections. If you look in EntityConnection for example, you should see the translation there where we are leveraging the `camelcase-keys` library.

PR #2334, opened 2021-04-01
Remove ProgramIndicators from special data flow

IgorNadj:


IgorNadj:
Don't merge until the data side is fixed in DHIS

chris-bes (approved):
Looks good!

PR #2335, opened 2021-04-05
2311: pt2 - Added common `field` query param which will flatten entities to the requested field

rohan-bes:
### Changes:

Specify `field` url parameter to get entity returned as just that field, eg:
`http://{{host}}/{{version}}/hierarchy/explore/TO&field=code => 'TO'`


packages/entity-server/src/routes/hierarchy/EntityDescendantsRoute.ts
```diff
@@ -4,29 +4,32 @@
...
     });
-    const responseEntities = includeRootEntity
-      ? [this.req.ctx.entity].concat(descendants)
-      : descendants;
-    return this.res.ctx.formatEntitiesForResponse(responseEntities);
+    const responseEntities = includeRootEntity ? [entity].concat(descendants) : descendants;
+    if (flat) {
+      return formatEntitiesForResponse(this.req.models, this.req.ctx, responseEntities, flat);
+    }
+    return formatEntitiesForResponse(this.req.models, this.req.ctx, responseEntities, fields);
```
rohan-bes, line 33: I really wanna be able to specify `formatEntitiesForResponse(this.req.models, this.req.ctx, responseEntities, flat || fields)` here but typescript really doesn't seem happy with that as it can't safely infer the type of `flat || fields` 😢

packages/entity-server/src/routes/hierarchy/format/formatEntityForResponse.ts
```diff
@@ -0,0 +1,123 @@
...
+  ctx: HierarchyContext,
+  entities: EntityType[],
+  fields: (keyof ExtendedEntityFields)[],
+): Promise<EntityResponseObject[]>;
+export async function formatEntitiesForResponse(
+  models: EntityServerModelRegistry,
+  ctx: HierarchyContext,
+  entities: EntityType[],
+  flat: keyof FlattableEntityFields,
+): Promise<EntityType[typeof flat][]>;
```
rohan-bes, line 61: I find this function overloading syntax so ugly, yet also quite useful... hope its not too verbose...

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
```diff
@@ -40,81 +37,26 @@ export const extractFieldsFromQuery = (queryFields?: string) => {
...
-) => {
-  const responseBuilder = new EntityResponseObjectBuilder();
-  for (let i = 0; i < fields.length; i++) {
-    const field = fields[i];
-    if (isExtendedField(field)) {
-      responseBuilder.set(field, await extendedFieldFunctions[field](entity, context));
-    } else {
-      responseBuilder.set(field, entity[field]);
-    }
+const flattableFields: (keyof FlattableEntityFields)[] = ['id', 'code', 'name'];
```
rohan-bes, line 46: Only valid fields are the `NOT NULL` fields here.

packages/entity-server/src/routes/hierarchy/types.ts
```diff
@@ -15,8 +15,27 @@ export interface HierarchyRequestParams {
...
+}[keyof T];
+
+type EntityFilterQuery = Partial<
+  Omit<EntityFields, ObjectLikeKeys<EntityFields>> &
+    FlattenAndPrefix<EntityFields, ObjectLikeKeys<EntityFields>>
+>;
+
+export interface HierarchyRequestQuery extends EntityFilterQuery {
   fields?: string;
+  field?: string;
```
rohan-bes, line 38: Hope these params aren't too confusing... `field` is for getting the flat response, whereas `fields` will return an object with listed fields. I thought about just using `fields` and if just 1 is supplied we flatten it? My issue there would be it could be unexpected for clients to get a different data shape based on the number of requested fields...

Also, while the param in the query is `field` I'm calling it `flat` on the context object and all throughout the code. I did this since I fould field and fields to be rather confusing but I am happy to get pushback on it

tupaia-packages.code-workspace
```diff
@@ -110,6 +110,7 @@
     "cSpell.words": [
       "Codeship",
       "DHIS",
+      "Flattable",
```
rohan-bes, line 113: Apparently Flattable isn't a word... 🙄

packages/entity-server/src/routes/hierarchy/middleware/fields.ts
```diff
@@ -40,81 +37,26 @@ export const extractFieldsFromQuery = (queryFields?: string) => {
...
-                field,
-                await extendedFieldFunctions[field](entity, context),
-              );
-            }),
-          );
-        }
-      }
-    } else {
-      entities.forEach((entity, index) => responseBuilders[index].set(field, entity[field]));
-    }
```
rohan-bes, line 117: Didn't change any functionality here really, just moved to `packages/entity-server/src/routes/hierarchy/format/formatEntityForResponse.ts`

packages/entity-server/src/routes/hierarchy/types.ts
```diff
@@ -29,31 +48,36 @@ export type ExtendedFieldFunctions = Readonly<
 
 type ExcludeCommonFields<T, U> = Omit<T, Extract<keyof T, keyof U>>;
 
+export type FlattableEntityFields = Pick<EntityFields, SimpleFieldKeys<EntityFields>>;
```
kael89, line 51: Perhaps we could redefine `SimpleFieldKeys`as `SimpleFields` (or `PickSimple`), by removing `[keyof T]` in that type, and then define this as 

```suggestion
export type FlattableEntityFields = SimpleFields<EntityFields>;
```

kael89 (changes requested):
Great work @rohan-bes! A few changes requested, but I love the general structure!

packages/entity-server/src/routes/hierarchy/types.ts
```diff
@@ -29,31 +48,36 @@ export type ExtendedFieldFunctions = Readonly<
...
 export type ExtendedEntityFields = ExcludeCommonFields<EntityFields, ExtendedFieldFunctions> &
   ExtendedFieldFunctions;
 
 export type EntityResponseObject = {
   [field in keyof ExtendedEntityFields]?: ExtendedEntityFields[field];
 };
 
+export type EntityResponse =
+  | EntityResponseObject
+  | FlattableEntityFields[keyof FlattableEntityFields];
```
kael89, line 62: `FlattableEntityFields[keyof FlattableEntityFields]` appears several times in this PR, so it may be worth defining it as a new type (eg `EntityResponseFlat`, `EntityResponsePrimitive`)

packages/entity-server/src/routes/hierarchy/types.ts
rohan-bes: You know what you are right actually, and given that we are using typescript and they have different type signatures I don't think it'll be too confusing. I'll change flat->field internally as well 👍

kael89 (changes requested):

packages/entity-server/src/routes/hierarchy/types.ts
rohan-bes: I did this in pt4 👍

PR #2336, opened 2021-04-05
2311: pt3 - Added common filtering functionality

rohan-bes:
### Changes:
Specify a value or values for a field to filter requested entities by, eg:
`http://{{host}}/{{version}}/hierarchy/explore/TO/descendants?fields=type,code&type=facility => Facility descendants of Tonga`


packages/data-broker/src/services/dhis/DhisInputSchemeResolvingApiProxy.js
```diff
@@ -229,10 +229,7 @@ export class DhisInputSchemeResolvingApiProxy {
     for (let i = 0; i < dhisIds.length; i += SQL_BIND_CHUNK_SIZE) {
       const chunkDhisIds = dhisIds.slice(i, i + SQL_BIND_CHUNK_SIZE);
       const chunkMappings = await this.models.dataServiceEntity.find({
-        [QUERY_CONJUNCTIONS.RAW]: {
-          sql: `config->>'dhis_id' in (${chunkDhisIds.map(() => '?')})`,
-          parameters: chunkDhisIds,
-        },
+        'config->>dhis_id': chunkDhisIds,
```
rohan-bes, line 232: I'm fixing the unit tests I broke after changing this, but @IgorNadj you may be interested in this change. I've basically made it possible to reference jsonb field via a similar `->>` notation to the postgres one

packages/database/src/TupaiaDatabase.js
```diff
@@ -581,7 +581,10 @@ function addWhereClause(connection, baseQuery, where) {
     if (!VALID_COMPARISON_TYPES.includes(comparisonType)) {
       throw new Error(`Cannot compare using ${comparisonType}`);
     }
-    const columnSelector = castAs ? connection.raw(`??::${castAs}`, [key]) : key;
+
+    const columnKey = key.includes('->>') ? connection.raw(`??->>?`, key.split('->>')) : key;
+    const columnSelector = castAs ? connection.raw(`??::${castAs}`, [columnKey]) : columnKey;
```
rohan-bes, line 586: Still need to test to ensure that that `castAs` doesn't break when key includes `->>`

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -0,0 +1,63 @@
...
+const isNestedField = (field: keyof EntityFilterQuery): field is keyof NestedFilterQueryFields =>
+  (nestedFields as (keyof EntityFilterQuery)[]).includes(field);
+
+type JsonBKey<T extends keyof NestedFilterQueryFields> = T extends `${infer Field}_${infer Key}`
+  ? `${Field}->>${Key}`
+  : T;
+
+const toJsonBKey = <T extends keyof NestedFilterQueryFields>(nestedField: T): JsonBKey<T> => {
+  const [field, value] = nestedField.split('_');
+  return [field, value].join('->>') as JsonBKey<T>;
```
rohan-bes, line 32: Have to cast explicitly here unfortunately as `string.split()` returns a generic `string` type. It's only a matter of time before I write a custom `split()` function that can return the correctly typed array of a string literal type 😝

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -0,0 +1,63 @@
...
+  'attributes_type',
+];
+const isFilterableField = (field: string): field is keyof EntityFilterQuery =>
+  (filterableFields as string[]).includes(field);
+
+const nestedFields: (keyof NestedFilterQueryFields)[] = ['attributes_type'];
+const isNestedField = (field: keyof EntityFilterQuery): field is keyof NestedFilterQueryFields =>
+  (nestedFields as (keyof EntityFilterQuery)[]).includes(field);
+
+type JsonBKey<T extends keyof NestedFilterQueryFields> = T extends `${infer Field}_${infer Key}`
```
kael89, line 26: `JsonBKey` is a cool type that can be generalised, perhaps we could use something more generic than `NestedFilterQueryFields` and also move it to `entity-server/src/types`.

Don't worry about it if it's too inconvenient here

kael89 (commented):
Awesome work with those types @rohan-bes !! I had a few minor suggestions, but most of my comments are just giving kudos 😄 

On a more general note: I'm a bit torn between using flat params for filtering, eg
```js
/hierarchy/laos_schools/LA/descendants?type=school&attributes_type=primary,secondary
```

or nesting them under a `filter` param:
```js
/hierarchy/laos_schools/LAdescendants?filter=type:school;attributes_type:primary,secondary
```

**Flat approach**:
* (Probably) more readable

**Nested approach**
* No risk of query param clashes
* Makes filter parsing code simpler

I'm leaning towards the 2nd approach, especially given that this is a pretty generic functionality that is enabled across routes (and not eg. an explicit `code` query param enabled in specific routes). @edmofro, I'd like your opinion - perhaps I'm just influenced by DHIS2 here!

edmofro:
I'd support nesting the filters under a `filter` key. I think it makes things simpler as soon as you have any non-filter query params, like `fields` etc., and also avoids potential clashes as you say @kael89. Also, we may want filters to have more complex behaviour e.g. `or` vs. `and`, which would be much easier to support if all the filters were under one query param in a custom format

PR #2345, opened 2021-04-07
2583 basic auth lesmis server

edmofro:
### Issue #:

### Changes:
Session now optional. If there is none attached to the request, will use default credentials for the LESMIS Server Laos Public user.

### Manual steps required on release:
- [ ] Update env vars per LastPass


packages/lesmis-server/src/app/createApp.ts
```diff
@@ -7,15 +7,14 @@ import { TupaiaDatabase } from '@tupaia/database';
...
+import { attachSession } from '../auth';
 
 /**
  * Set up express server with middleware,
  */
 export function createApp() {
-  return new OrchestratorApiBuilder(new TupaiaDatabase())
+  return new OrchestratorApiBuilder(new TupaiaDatabase(), attachSession)
     .useSessionModel(LesmisSessionModel)
-    .verifyLogin(verifyLogin)
```
edmofro, line 18: Removed verifyLogin check as we allow anyone to access the lesmis-server endpoints, whether they're logged in or not

packages/lesmis-server/src/connections/SessionHandlingApiConnection.ts
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable max-classes-per-file */
```
rohan-bes, line 1: Personally I'd love for us to just disable this rule entirely? I think creating one class per file is just standard practice and I don't really need a rule to remind me of this... and in the scenario that we choose to break this rule we know what we're doing and why... How do you feel about it?

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -24,13 +29,16 @@ export class ApiBuilder {
...
   private readonly database: TupaiaDatabase;
 
+  private readonly attachSession: (req: Request, res: Response, next: NextFunction) => void;
+
   private attachVerifyLogin?: (req: LoginRequest, res: Response, next: NextFunction) => void;
 
   private verifyLoginMiddleware?: (req: Request, res: Response, next: NextFunction) => void;
 
-  constructor(transactingConnection: TupaiaDatabase) {
+  constructor(transactingConnection: TupaiaDatabase, attachSession = defaultAttachSession) {
```
rohan-bes, line 38: I think I have a slight preference towards setting the `attachSession` on the builder via something like:

```
  private attachSession = defaultAttachSession;
  
  ...
  
  useAttachSession(attachSession: (req: Request, res: Response, next: NextFunction) => void) {
    this.attachSession = attachSession;
  }
```

I appreciate that this could possibly create issues for someone using the `ApiBuilder` who didn't realise you need to specify the custom `attachSession` prior to adding any of their routes, but I feel keeping the constructor cleaner is worth the tradeoff?

packages/utils/src/createBasicHeader.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+export const createBasicHeader = (username, password) =>
+  `Basic ${Buffer.from(`${username}:${password}`).toString('base64')}`;
```
rohan-bes, line 7: Love this!! Shoulda done this a long time ago ❤️

packages/lesmis-server/src/auth/attachSession.ts
```diff
@@ -0,0 +1,32 @@
...
+import { UnauthenticatedError } from '@tupaia/utils';
+
+export const attachSession = async (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const sessionId = req.sessionCookie?.id;
+    if (!sessionId) {
+      // within lesmis-server, not having a session means the route handlers will use the default
+      // public lesmis server user auth header
+      next();
+      return;
```
rohan-bes, line 17: Given that the session is optional within the context of the `lesmis-server` we should probably change the `Request` type in the global namespace to have session be optional. In `lesmis-server/src/@types/express/index.d.ts`:
```
    export interface Request {
      session?: LesmisSessionType;
      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      ctx: any;
    }
```

packages/lesmis-server/src/auth/index.ts
```diff
@@ -3,4 +3,4 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-export { verifyLogin } from './verifyLogin';
+export { attachSession } from './attachSession';
```
rohan-bes, line 6: Very nitpicky, so totally optional, but I guess I feel this package ought to be renamed to `session` instead of `auth`?

rohan-bes (changes requested):
Couple of changes requested, but it's looking really good!! Nice one 👍

packages/lesmis-server/src/connections/SessionHandlingApiConnection.ts
edmofro: Yeah I'm with you there! @kael89 how about you?

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
edmofro: Ooh yeah I like that actually, and gets in early with the pattern before we have a sprawling list of constructor arguments to support configuration options

packages/lesmis-server/src/auth/attachSession.ts
edmofro: Good point

packages/lesmis-server/src/auth/index.ts
edmofro: Valid!

packages/lesmis-server/src/connections/SessionHandlingApiConnection.ts
edmofro: I'm going to leave this for now and raise it at dev chat

kael89: I agree with you, let's just disable it!

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -14,23 +14,33 @@ import { UnauthenticatedError } from '@tupaia/utils';
...
+  LoginRoute,
+  LoginRequest,
+  LogoutRoute,
+  attachSession as defaultAttachSession,
+} from '../routes';
 import { ExpressRequest, Params, ReqBody, ResBody, Query } from '../../routes/Route';
 import { sessionCookie } from './sessionCookie';
 import { SessionModel } from '../models';
 
+type Middleware = (req: Request, res: Response, next: NextFunction) => void;
```
rohan-bes, line 27: Nice!

rohan-bes (approved):
Awesome, looks great!

PR #2346, opened 2021-04-07
Allow basic auth on any web-config-server endpoint

edmofro:
### Issue #:


packages/web-config-server/src/authSession/authSession.js
```diff
@@ -1,45 +1,46 @@
...
+    req.accessPolicy = await getAccessPolicyForUser(req.models, basicAuthUser.fullName);
     next();
-    // else check if this is the first request after user logged out
-    // and send 440 or authenticate as public user
-  } else if (req.lastuser && req.lastuser.userName && req.lastuser.userName !== PUBLIC_USER_NAME) {
+    return;
+  }
+
+  // if logged in or logging in continue
+  const userName = req.session?.userJson?.userName;
```
rohan-bes, line 21: Love this syntax, so handy

packages/web-config-server/src/authSession/authSession.js
```diff
@@ -1,45 +1,46 @@
...
+  if (!!userName || checkAllowedUnauthRoutes(req)) {
+    req.accessPolicy = req.accessPolicy || (await getAccessPolicyForUser(req.models, userName));
+    next();
+    return;
+  }
+
+  // check if this is the first request after user logged out and send 440
+  if (req.lastuser?.userName !== PUBLIC_USER_NAME) {
```
rohan-bes, line 29: I think the control flow is slightly changed here? Previously if `req.lastUser.userName = undefined` we would fail this if check whereas now we succeed? Not sure if that's at all a valid case though...

rohan-bes (approved):
Looks good to me, just one question about the change in control flow in `auth()` but as long as that's confirmed not to be an issue then it's all good, so I'm pre-approving 👍

packages/web-config-server/src/authSession/authSession.js
edmofro: I think it should still fail, because `PUBLIC_USER_NAME` is not `undefined` :-)

edmofro: Oooops... 🤦

edmofro: Yep you're totally right, I mucked that up!

edmofro: Fixed!

rohan-bes (approved):
Awesome, perfect!

edmofro:
Merged via https://github.com/beyondessential/tupaia/pull/2375

PR #2356, opened 2021-04-08
Update PSSS to use data_time

EMcQ-BES:
### Changes:

- Missed this because I only searched `*.js` files for references to `submission_time`
- This appears to be the last remaining reference to `submission_time`

### DIFF
--- a/packages/psss-server/src/connections/MeditrakConnection.ts
+++ b/packages/psss-server/src/connections/MeditrakConnection.ts
@@ -4,7 +4,7 @@
  */
 
 import { generateId } from '@tupaia/database';
-import { convertPeriodStringToDateRange } from '@tupaia/utils';
+import { convertPeriodStringToDateRange, stripTimezoneFromDate } from '@tupaia/utils';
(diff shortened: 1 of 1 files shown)


packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -46,17 +46,17 @@ export class MeditrakConnection extends ApiConnection {
...
       pageSize: '1',
-      columns: `["entity.code","survey.code","submission_time","id"]`,
+      columns: `["entity.code","survey.code","data_time","id"]`,
       filter: JSON.stringify({
         'survey.code': { comparisonValue: surveyCode },
         'entity.code': { comparisonValue: orgUnitCode },
-        submission_time: {
+        data_time: {
           comparator: 'BETWEEN',
           comparisonValue: [startDate, endDate],
```
EMcQ-BES, line 55: `convertPeriodStringToDateRange` returns these values as date format `YYYY-MM-DD` so shouldn't need to have the timezone stripped

edmofro (approved):

PR #2359, opened 2021-04-08
2351 lesmis server report endpoint

EMcQ-BES:
### Changes:

- Add WebConfigConnection
- Report endpoint acts as a very dumb middle-man forwarding to either the report server or the web config server


packages/lesmis-server/src/routes/ReportRoute.ts
```diff
@@ -0,0 +1,35 @@
...
+    const { reportCode } = this.req.params;
+    const { type } = this.req.query;
+    switch(type) {
+      case 'view':
+        return this.webConfigConnection.getDashboardReport({ viewId: reportCode, ...this.req.query });
+      case 'measureData':
+        return this.webConfigConnection.getMapOverlay({ measureId: reportCode, ...this.req.query });
+      case 'report':
+      default:
+        return this.reportConnection.forwardReportQuery(reportCode, this.req.query);
```
edmofro, line 32: Nice defaulting - I'd suggest we don't even add the explicit query parameter for those going to `report`, with the idea that eventually they all will be

packages/lesmis-server/src/connections/WebConfigConnection.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { ApiConnection, QueryParameters } from '@tupaia/server-boilerplate';
+const { WEB_CONFIG_API_URL = 'http://localhost:8000/api/v1' } = process.env;
+
+export class WebConfigConnection extends ApiConnection {
+  baseUrl = WEB_CONFIG_API_URL;
+
+  async getDashboardReport(query: QueryParameters) {
```
edmofro, line 12: We have this semi-convention of using `fetch` when a getter is async.
```suggestion
  async fetchDashboardReport(query: QueryParameters) {
```
(same below)

packages/lesmis-server/src/connections/ReportConnection.ts
```diff
@@ -27,4 +28,8 @@ export class ReportConnection extends SessionHandlingApiConnection {
       period: periods.join(';'),
     });
   }
+
+  async forwardReportQuery(reportCode: string, query: QueryParameters) {
```
edmofro, line 32: Is anything else using the more opinionated one above? I'd say we shouldn't have both. Either replace the `fetchReport` completely (and probably keep that name) or make the report route use the opinionated one

packages/lesmis-server/src/routes/ReportRoute.ts
```diff
@@ -0,0 +1,35 @@
...
+    super(req, res, next);
+
+    this.reportConnection = new ReportConnection(req.session);
+    this.webConfigConnection = new WebConfigConnection(req.session);
+  }
+
+  async buildResponse() {
+    const { reportCode } = this.req.params;
+    const { type } = this.req.query;
+    switch(type) {
```
edmofro, line 25: Are there differences in the parameters required by the three? Ideally that's the domain of the orchestration server, so the lesmis front end doesn't have to deal with constructing different types of url. However also happy to keep it as is if @tcaiger has already made lesmis front end do exactly that!

edmofro (changes requested):
Nice work!

packages/lesmis-server/src/routes/ReportRoute.ts
EMcQ-BES: Removed

packages/lesmis-server/src/connections/WebConfigConnection.ts
EMcQ-BES: Ah I was curious why we were inconsistent about the language around fetch/get, that makes sense

edmofro (approved):

packages/lesmis-server/src/connections/ReportConnection.ts
EMcQ-BES: The more opinionated `fetchReport` is not being used by `lesmis-server`, looks like this file was originally copy/pasted from PSSS where that was actually used. I've replaced it with the new less opinionated function.

packages/lesmis-server/src/routes/ReportRoute.ts
EMcQ-BES: There are different parameters required, I wrote it dumb because I thought we already had those urls generated on the front end. Will get Tom to confirm, and will add smarts if needed.

EMcQ-BES: After a chat with Tom I've added some extra handling to account for some of the shared parameters between these endpoints, the front end may still pass through extra query parameters in cases where they're needed

packages/lesmis-server/src/routes/ReportRoute.ts
```diff
@@ -20,15 +21,30 @@ export class ReportRoute extends Route {
   }
 
   async buildResponse() {
-    const { reportCode } = this.req.params;
-    const { type } = this.req.query;
+    const { entityCode, reportCode } = this.req.params;
+    const {type} = this.req.query;
```
edmofro, line 25: I think prettier wants the spaces

edmofro (approved):

packages/lesmis-server/src/routes/ReportRoute.ts
EMcQ-BES: 🤦‍♂️🤦‍♂️ I did this as an obvious mistake to check if prettier was running after I adjusted my lint settings, and then failed to run it correctly from CLI

PR #2369, opened 2021-04-12
Shade out laos 

billli0:
### Issue #:
[1766](https://github.com/beyondessential/tupaia-backlog/issues/1766)
### Changes:
1. Add shaded map urls into frontend for Laos EOC 
2. Migration to change `laos_eoc` project maps config to shaded map


packages/web-frontend/src/selectors/projectSelectors.js
```diff
@@ -48,7 +50,7 @@ export const selectActiveTileSetKey = state => state.map.activeTileSetKey;
 export const selectActiveTileSet = createSelector(
   [selectTileSets, selectActiveTileSetKey],
   (tileSets, activeTileSetKey) => {
-    return tileSets.find(tileSet => tileSet.key === activeTileSetKey);
+    return tileSets.find(tileSet => tileSet.key === activeTileSetKey) || tileSets[0];
```
billli0, line 53: When rendering the page, the initial `tileSets` here were an array with only two items(osm and satellite), and I couldn't find where they were set. This will cause error because with new changes, `activeTileSetKey` here is `laosOpenStreets`.

packages/web-frontend/src/selectors/projectSelectors.js
```diff
@@ -37,7 +37,9 @@ export const selectTileSets = createSelector(selectCurrentProject, project => {
 
   if (project.config && project.config.tileSets) {
     const customSetKeys = project.config.tileSets.split(',').map(item => item.trim());
-    tileSetKeys = [...tileSetKeys, ...customSetKeys];
+    tileSetKeys = customSetKeys.includes('laosOpenStreets', 'laosSatellite')
```
IgorNadj, line 40: I can see the intent here, to have laos_eoc only have the laos tilesets. But, this is putting laos specific logic into a function that should work for all projects equally.

My suggestion is to instead add a property to the project config, `includeDefaultTilesets` (default true). This can be set to false for laos_eoc.

The code then becomes: `tileSetKeys = includeDefaultTilesets ? [...tileSetKeys, ...customSetKeys] : customSetKeys`.

The defaults `osm, satellite` are defined above on line 36.

billli0: Thanks Igor, this looks great!
Additionally I suggest we use **not**IncludeDefaultTilesets, if not, in other projects `includeDefaultTilesets` is missing and false, and that will not include default tile sets.

IgorNadj: Ah yep, I was thinking if it was `includeDefaultTilesets` you can leave it out of the others, and then in this code here you can give it a default of true e.g.

```
const { includeDefaultTilesets = true } = project.config;
```

Both ways make sense, I just like to use positive variable names as they are usually easier for me to read and understand quickly.

billli0: That's a smart way to play with code. I agree using positive variable name is better. Will implement this suggestion.

IgorNadj (approved):
LGTM 🏌️

PR #2370, opened 2021-04-12
2248 entity clean up

billli0:
### Issue #:
[2248](https://github.com/beyondessential/tupaia-backlog/issues/2248)


chris-bes:
Looks good! But don't we need to change some entity codes as well?

chris-bes (commented):
Looks good! But don't we need to change some entity codes as well?

packages/database/src/migrations/20210412055533-LaosEocEntitycleanUp-modifies-data.js
```diff
@@ -0,0 +1,64 @@
...
+  for (const entityCode of entityCodes) {
+    const trimmedEntityCode = entityCode.trim();
+    await db.runSql(`
+      UPDATE entity 
+      SET code = '${trimmedEntityCode}'
+      where code = '${entityCode}';
+
+      UPDATE clinic 
+      SET code = '${trimmedEntityCode}'
+      where code = '${entityCode}';
```
chris-bes, line 41: Just a bit nit picky but not all the provided entity codes are facilities. I think this will just not update if it doesn't exist but up to you to change it or not

billli0:
Thanks Chris, I've added a loop to change these entity codes

packages/database/src/migrations/20210412055533-LaosEocEntitycleanUp-modifies-data.js
```diff
@@ -20,16 +20,218 @@ exports.up = async function (db) {
...
-  for (const entityCode of entityCodes) {
+  for (let i = 0; i < originalEntityCodes.length; i++) {
+    await db.runSql(`
+    UPDATE entity 
+    SET code = '${newEntityCodes(i)}'
+    where code = '${originalEntityCodes(i)}';
+
+    UPDATE clinic 
+    SET code = '${newEntityCodes(i)}'
+    where code = '${originalEntityCodes(i)}';
```
chris-bes, line 228:
```suggestion
```
I think those are all individuals so we don't need to update clinic table

chris-bes (changes requested):
Thanks Biao, almost there!

packages/database/src/migrations/20210412055533-LaosEocEntitycleanUp-modifies-data.js
billli0: Yeah you are right. Thanks Chris, I will delete these codes.

chris-bes (approved):
Looks good!

PR #2371, opened 2021-04-12
Allow bearer auth against web-config-server

edmofro:
### Issue #:

### Changes:
Adds support for bearer auth

### Manual deploy steps required
- [ ] Add JWT_SECRET to the .env on dev and production


packages/web-config-server/src/authSession/getUserFromAuthHeader.js
```diff
@@ -0,0 +1,47 @@
...
+  if (authHeader.startsWith('Basic ')) {
+    return getUserFromBasicAuth(authenticator, authHeader);
+  }
+
+  if (authHeader.startsWith('Bearer ')) {
+    return getUserFromBearerAuth(models, authHeader);
+  }
+
+  return null;
+};
```
rohan-bes, line 47: Would love to migrate web-config-server over to new server-boilerplate whenever there's time 😄 (not suggesting we do this now btw!)

packages/lesmis-server/examples.http
```diff
@@ -41,3 +41,9 @@ Content-Type: {{contentType}}
 GET http://{{host}}/{{version}}/user HTTP/1.1
 content-type: {{contentType}}
 Cookie: {{sessionCookie}}
+
+
+### Logged in reports fetch
+GET http://{{host}}/{{version}}/reportData/LA_Phonhong%20District/Laos_Schools_Number_Of_Schools_By_Type_Table?type=view&dashboardGroupId=337 HTTP/1.1
+content-type: {{contentType}}
+Cookie: {{sessionCookie}}
```
rohan-bes, line 49: Love these! 👍

rohan-bes (approved):
Looks good!

packages/lesmis-server/examples.http
edmofro: Thanks! Turns out you don't need to explicitly set the Cookie, the api client will take care of that just like a browser when the response from login sets it. Next PR removes them.

packages/web-config-server/src/authSession/getUserFromAuthHeader.js
edmofro: Yeah that'd be so nice... 😍

edmofro:
Merged via https://github.com/beyondessential/tupaia/pull/2375

PR #2375, opened 2021-04-13
2583 accept api client secret

edmofro:
### Issue #:

Probably easiest reviewed commit-by-commit (they're hopefully quite orderly)

### Changes
- Added api client secret check to server-boilerplate auth middleware (with failover still falling through to regular user basic auth check)
- Moved report-server to use server-boilerplate auth middleware (had all the same existing content, so this is the best way to add that same support for api client secret checking)
- Changed web-config-server basic auth checking to _expect_ api client secret, rather than regular username and password (the latter only being introduced in the first PR of this series, and not yet on dev - regular username and password checking via basic auth header is not a requirement of web-config-server)

### Manual release steps
- [ ] Update env vars of web-config-server, report-server, and entity-server to include API_CLIENT_SALT
- [ ] Update env vars of lesmis-server to swap out MICROSERVICE_CLIENT_PASSWORD for MICROSERVICE_CLIENT_SECRET and add MICROSERVICE_CLIENT_USERNAME


packages/report-server/src/app/createApp.ts
```diff
@@ -38,6 +39,11 @@ export function createApp(database: TupaiaDatabase, models: ModelRegistry) {
     next();
```
rohan-bes, line 39: I think we may now be able to remove setting the authenticator on the request object above?

packages/web-config-server/src/authSession/getUserFromAuthHeader.js
```diff
@@ -2,21 +2,18 @@
...
+const getApiClientUserFromBasicAuth = async (models, authHeader) => {
+  const { username, password: secretKey } = getUserAndPassFromBasicAuth(authHeader);
+
+  // first attempt to authenticate as an api client, in case a secret key was used in the auth header
+  const secretKeyHash = encryptPassword(secretKey, process.env.API_CLIENT_SALT);
+  const apiClient = await models.apiClient.findOne({
+    username,
+    secret_key_hash: secretKeyHash,
+  });
+  return apiClient && apiClient.getUser();
```
rohan-bes, line 16:
```suggestion
  return apiClient?.getUser();
```

Nitpicky so no need if you don't want.

rohan-bes (approved):
Minor requests but nothing major. Looks good to me 😄

packages/report-server/src/app/createApp.ts
edmofro: Good point, that cleans things up :-)

packages/web-config-server/src/authSession/getUserFromAuthHeader.js
edmofro: Oh na I'm keen as, I love this syntax and just need to get in the habit - thanks!

PR #2376, opened 2021-04-13
1641 psss vis country level dashboard 2y

IgorNadj:
### Changes:

- Add PsssVisCountryLevelDashboard2y migration
- Limit percentage chart to 100%

---

### Screenshots:
https://github.com/beyondessential/tupaia-backlog/issues/1641#issuecomment-818342080


billli0:
Do we need to add different color for different line chart?

IgorNadj:
> Do we need to add different color for different line chart?

I think it's just for the bars, I asked Erin but haven't got a reply yet. I might merge and if we need to, we can create a new PR to change colours

PR #2378, opened 2021-04-13
1709: pt1 - Multi Entity Routes in entity-server

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1709:

Added ability to omit the '/:entityCode' part of an entity-server route and instead pass in a url param 'entities' with a list of entity codes. Routes will return all data for listed entities. Currently just supported for /relations and /descendants as that was all that was needed for this ticket

Also made the ancestor_filter=type url parameter optional for the relations route. The new default if this is omitted is to find the relations of the descendants to the requested entity.


packages/entity-server/src/app/createApp.ts
```diff
@@ -22,8 +31,28 @@ import { attachRelationsContext } from '../routes/hierarchy/relations/middleware
...
   return new MicroServiceApiBuilder(new TupaiaDatabase())
     .useBasicBearerAuth('entity-server')
-    .use<HierarchyRequest>('/v1/hierarchy/:hierarchyName/:entityCode', attachContext)
-    .get<HierarchyRequest>(
+    .use<SingleEntityRequest | MultiEntityRequest>(
+      '/v1/hierarchy/:hierarchyName',
+      attachCommonContext,
+    )
+    .use<MultiEntityRequest>(
+      ['/v1/hierarchy/:hierarchyName/relations', '/v1/hierarchy/:hierarchyName/descendants'],
```
rohan-bes, line 39: Have to be a bit explicit here as matching against `/v1/hierarchy/:hierarchyName` will match all routes which isn't what we want. We probably ought to switch to using regexes throughout this whole file to make the routes very explicit but I couldn't quite be bothered doing that at the moment

packages/entity-server/src/routes/hierarchy/types.ts
```diff
@@ -48,24 +56,36 @@ export type FlattenedEntity = FlattableEntityFields[keyof FlattableEntityFields]
...
+  entity: EntityType;
+}
+
+export interface MultiEntityContext extends CommonContext {
+  entities: EntityType[];
 }
 
-export interface HierarchyRequest<
-  P = HierarchyRequestParams,
+export interface SingleEntityRequest<
```
rohan-bes, line 75: Bulk renamed HierarchyRequest => SingleEntityRequest

Sorry for all the deltas...

packages/entity-server/src/routes/hierarchy/middleware/attachCommonContext.ts
```diff
@@ -0,0 +1,37 @@
+/**
```
rohan-bes, line 1: All of this is copied out from the previously named `packages/entity-server/src/routes/hierarchy/middleware/middleware.ts`

packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
```diff
@@ -0,0 +1,110 @@
+/**
```
rohan-bes, line 1: Logic all copied from `packages/entity-server/src/routes/hierarchy/middleware/middleware.ts` with some internal code restructuring changes

packages/entity-server/src/routes/hierarchy/relations/EntityRelationsRoute.ts
```diff
@@ -4,122 +4,15 @@
...
 import { RelationsRequest } from './types';
-
-type Pair = {
-  descendant: string;
-  ancestor: string;
-};
+import { RelationsResponseBuilder } from './RelationsResponseBuilder';
 
 export class EntityRelationsRoute extends Route<RelationsRequest> {
-  async buildAncestorCodesAndPairs(descendants: EntityType[]): Promise<[string[], Pair[]]> {
```
rohan-bes, line 18: All logic moved to `packages/entity-server/src/routes/hierarchy/relations/RelationsResponseBuilder.ts` so as to be re-usable with `MultiEntityRelationsRoute`

Originally I tried to have the multi entity route call this route directly, however that was proving difficult for typescript and generally I found it a bit of an ugly approach...

packages/entity-server/src/routes/hierarchy/MultiEntityDescendantsRoute.ts
```diff
@@ -0,0 +1,42 @@
...
+          ...filter,
+        });
+        const entitiesToUse = includeRootEntity ? [entity].concat(descendants) : descendants;
+        responseEntities.push(...entitiesToUse);
+      }),
+    );
+    if (field) {
+      return formatEntitiesForResponse(this.req.models, this.req.ctx, responseEntities, field);
+    }
+    return formatEntitiesForResponse(this.req.models, this.req.ctx, responseEntities, fields);
```
kael89, line 40:
```suggestion

  return formatEntitiesForResponse(this.req.models, this.req.ctx, responseEntities, field || fields);
```

packages/entity-server/src/app/createApp.ts
kael89: No worries!

kael89: The combination of generics + many-to-many mapping between `use` and `get` makes this section a bit harder to read now, but I'm not sure whether there is a better way...

I sense that it would be advantageous to see the supported routes with one glance,. Perhaps we will attempt to infer the request context in the future? For the time being, your approach seems the way to go.

One idea is grouping `use` and `get` statements together, although this may not actually improve things - still, probably worth a quick try.

packages/entity-server/src/routes/hierarchy/relations/types.ts
```diff
@@ -18,27 +20,47 @@ export type Prefix<T, P extends string> = {
...
   extends Request<
-    HierarchyRequestParams,
-    | Record<FlattenedEntity, EntityResponse> // groupBy: descendant
-    | Record<FlattenedEntity, EntityResponse[]>, // groupBy: ancestor
-    HierarchyRequestBody,
-    RelationsQuery
+    MultiEntityRequestParams,
+    RelationsResponseBody,
+    RequestBody,
+    RelationsQuery & { entities?: string }
```
kael89, line 60: Are `entities` here optional?

kael89 (changes requested):
Excellent work @rohan-bes ! I left a few suggestions but they are all minor

packages/entity-server/src/app/createApp.ts
rohan-bes: Could do something like this if you find it more readable?
```
.use<MultiEntityRequest>('/v1/hierarchy/:hierarchyName/relations', attachMultiEntityContext)
.use<MultiEntityRequest>('/v1/hierarchy/:hierarchyName/descendants', attachMultiEntityContext)
...
```

Alternatively, we could scrap the whole multi-entity/single-entity distinction and just make `entityCode` an optional param that overrides the `entities` param if it's present. That would achieve the same behaviour basically. We could even consolidate the routes then if we wish? I'm in two minds about such an approach; on the one hand it would remove complexity and reduce code duplication, on the other hand my gut does feel a bit that we want to treat the single entity and multi-entity cases as separate...

Thoughts on this?

kael89: My gut also feels that the routes should be different, even if they share they same functionality under the hood... Also I think I find your suggestion above more readable. Finally, it's also a formatting thing... I wish that all those could fit in one line 😃

packages/entity-server/src/routes/hierarchy/MultiEntityDescendantsRoute.ts
rohan-bes: I'll need to add another function overload to support this, but I think it's worth it, will do 👍

kael89: How come? Since you already call this method with one or the other, based on the `if` predicate

packages/entity-server/src/routes/hierarchy/relations/RelationsResponseBuilder.ts
```diff
@@ -0,0 +1,154 @@
...
+    models: EntityServerModelRegistry,
+    ctx: RelationsContext,
+    groupBy: 'ancestor' | 'descendant' = 'ancestor',
+  ) {
+    this.models = models;
+    this.groupBy = groupBy;
+
+    const ancestorType = ctx.ancestor.type || ctx.entity.type;
+    if (ancestorType === null) {
+      throw new Error('No explicit ancestorType provided and entity type is null');
```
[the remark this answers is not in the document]
rohan-bes: Yeah, no, you're right. Usually with this kind of builder pattern you have no (or minimal) constructor params and then set properties before building. It's usually the `build()` function that does all the error checking before building. It can be very clean that way since you can have the various setters interacting with properties but it only does validation once.

That being said, it makes the typescript slightly more wordy to support that. You basically need to make the properties optional, and then validate that they're all present. I kinda wanted to rely on the constructor to be responsible for ensuring the ancestorType property is present so that the rest of the class can operate without having to check it's present all the time. I know it's not the cleanest OOP, but in this instance I think I prefer it 😄

packages/entity-server/src/routes/hierarchy/relations/types.ts
rohan-bes: Technically yes. All URL query parameters are optional in the sense that the user's request will still hit the middleware/route even if they didn't add the parameter. I actually like making all query parameters optional in typescript for this reason, as it forces my middleware/route to do validation/error handling for the case that the user forgot to add the right parameters 😄

We set the `entities` on the `MultiEntityContext` as non-optional since by the time we are setting the context we've ensured the url parameter is present 👍

packages/entity-server/src/routes/hierarchy/relations/RelationsResponseBuilder.ts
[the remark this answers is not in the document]
kael89: I understand, thanks for the explanation, certainly happy to leave this as is! 🙌

packages/entity-server/src/app/createApp.ts
rohan-bes: I think I'll add a `setVersion` property to the api-builder so that it can prefix each endpoint with the version which should shave off a whole 4 chars per line 😝

kael89: Haha, honestly, don't worry too much about it 😛

kael89: I mean, if we can find ways to make this very easy to parse, it would be awesome, but not a biggie

rohan-bes: Also, could omit the `Entity` part of `MultiEntity...Request`/`MultiEntity...Route`, eg. `MultiDescendantsRequest`? Doesn't quite make as much sense to me tho...

kael89: Better to leave `Entity` there, I reckon... Again, no stress about making this more succinct

packages/entity-server/src/routes/hierarchy/MultiEntityDescendantsRoute.ts
rohan-bes: If you're using function overloads, typescript will only allow you to call a function with one of the explicitly defined overloads. I do kind of like this behaviour tbh, even if it seems a bit weird, since otherwise it's hard to keep your overloads explicit, eg.
```
function combine(nums: number[]): number;
function combine(words: string[]): string;
function combine(values: number[] | string[]) {
  if (isNumber(values[0]) {
    return sum(values);
  }
  
  return values.join(' ');
}

const numberCombiner = (nums: number[]) => combine(nums); // valid
const wordCombiner = (words: string[]) => combine(words); // valid
const valueCombiner = (values: number[] | string[]) => combine(values); // Error: The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.
```
But if you add:
```

File: packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts [hunk 9]
```diff
@@ -0,0 +1,110 @@
...
+    const { filter } = req.query;
+    req.ctx.filter = extractFilterFromQuery(allowedCountries, filter);
+
+    next();
+  } catch (error) {
+    next(error);
+  }
+};
+
+export const attachMultiEntityContext = async (
```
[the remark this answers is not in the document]
rohan-bes: Good point, will do

packages/entity-server/src/routes/hierarchy/MultiEntityDescendantsRoute.ts
kael89: Interesting, thanks!

packages/entity-server/src/app/createApp.ts
```diff
@@ -31,41 +31,27 @@ import { attachRelationsContext } from '../routes/hierarchy/relations/middleware
...
     .useBasicBearerAuth('entity-server')
-    .use<SingleEntityRequest | MultiEntityRequest>(
-      '/v1/hierarchy/:hierarchyName',
-      attachCommonContext,
-    )
-    .use<MultiEntityRequest>(
-      ['/v1/hierarchy/:hierarchyName/relations', '/v1/hierarchy/:hierarchyName/descendants'],
-      attachMultiEntityContext,
-    )
+    .use<SingleEntityRequest | MultiEntityRequest>('hierarchy/:hierarchyName', attachCommonContext)
```
kael89, line 34: Looks  better now - thanks! 🙏

packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
```diff
@@ -36,30 +58,11 @@ export const attachSingleEntityContext = async (
...
-      throwNoAccessError(entityCode);
-    }
-
-    req.ctx.entity = entity;
-    req.ctx.allowedCountries = allowedCountries;
+    const context = await validateEntitiesAndBuildContext(req, [entityCode]);
 
-    const { filter } = req.query;
-    req.ctx.filter = extractFilterFromQuery(allowedCountries, filter);
+    [req.ctx.entity] = context.entities;
```
kael89, line 63: I didn't know you could use array destructuring without a `const` or `let` - nice!

kael89 (approved):
Great work @rohan-bes !

packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
rohan-bes: I actually didn't know this either, but our linter was yelling at me to `use array destructuring` and I had to keep trying things until it was happy 😝

kael89: Ah, the benevolent dictatorship of linters... 😆

kael89:
@rohan-bes we also need to update the [server-boilerplate README](https://github.com/beyondessential/tupaia/blob/psss-phase-2/packages/server-boilerplate/README.md) (`setVersion()` changes)

PR #2380, opened 2021-04-13
1709: pt2 - Entity Aggregation via Aggregator package

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1709:

- Added 'context' object on the databroker which holds a userSession object which can be shared throughout the pipeline to ensure requests to various micro-services are performed under the correct user.
- Added EntityConnection to aggregator package which queries the entity-server to attain relations map for performing entity aggregations
- Added hierarchy to the fetchOptions in a request
- Added logic to replace requested entity codes with data source entity codes in aggregator


packages/aggregator/package.json
```diff
@@ -11,7 +11,7 @@
   "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
   "main": "dist/index.js",
   "scripts": {
-    "build": "babel src --out-dir dist --source-maps --ignore \"src/__tests__/**\" --config-file \"../../babel.config.json\"",
+    "build": "rm -rf dist && babel src --out-dir dist --source-maps --ignore \"src/__tests__/**\" --config-file \"../../babel.config.json\"",
```
rohan-bes, line 14: Not actually sure if this was needed but I added when my build step wasn't working as expected and I don't mind keeping it.

packages/aggregator/src/Aggregator.js
```diff
@@ -17,6 +17,7 @@ export class Aggregator {
 
   constructor(dataBroker) {
     this.dataBroker = dataBroker;
+    this.context = dataBroker.context;
```
rohan-bes, line 20: Data broker is the single source of truth for the context

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/adjustOptionsToAggregationList.test.js
```diff
@@ -0,0 +1,66 @@
...
+    [
+      'adjustments in different directions: they cancel each other',
+      { aggregations: [getOffsetMonthsAgg(-1), 'FINAL_EACH_MONTH', getOffsetMonthsAgg(1)] },
+      fetchOptions,
+    ],
+    [
+      'adjustments in different directions: one of them is stronger',
+      { aggregations: [getOffsetMonthsAgg(1), 'FINAL_EACH_MONTH', getOffsetMonthsAgg(-3)] },
+      { ...fetchOptions, startDate: '2020-03-01', endDate: '2020-04-30', period: '202003;202004' },
+    ],
```
rohan-bes, line 50: Still need to add tests here for the new functionality, however that would require a fair amount of mocking and I'm kind of feeling the time pressure atm...

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -0,0 +1,60 @@
...
+const createRelationsUrl = (
+  hierarchyName,
+  entityCodes,
+  aggregationEntityType,
+  dataSourceEntityType,
+) =>
+  `hierarchy/${hierarchyName}/relations?entities=${entityCodes.join(
+    ',',
+  )}&field=code&groupBy=descendant&${
+    aggregationEntityType === 'requested' ? '' : `ancestor_filter=type:${aggregationEntityType}`
```
rohan-bes, line 24: This is a new supported aggregationEntityType, `requested`. It will aggregate dataSourceEntities to the requested entities. It is very useful but I feel it should be more clearly signified somewhere...

packages/psss-server/src/types.ts
```diff
@@ -17,7 +17,7 @@ export interface SessionCookie {
 
 type PsssRequestBody = Record<string, unknown>;
 
-export interface PsssRequest<>extends Request<ParamsDictionary, unknown, PsssRequestBody, Query> {
+export interface PsssRequest extends Request<ParamsDictionary, unknown, PsssRequestBody, Query> {
```
rohan-bes, line 20: The previous syntax was invalid and the new version of typescript was throwing an error at this...

packages/report-server/src/reportBuilder/fetch/fetch.ts
```diff
@@ -25,11 +25,13 @@ const buildParams = (params: unknown): FetchParams => {
     throw new Error(`Expected object but got ${params}`);
   }
 
-  if (Object.keys(params).length > 1) {
+  const { aggregation, ...fetchParams } = params;
+
+  if (Object.keys(fetchParams).length > 1) {
```
rohan-bes, line 30: Truth be told I'm not at all happy with the validation going on here, however I've been meaning to re-write report-server to use the new server-boilerplate and once that's been done I think we can take a fresh look at the validation logic...

packages/report-server/src/routes/fetchReport.ts
```diff
@@ -11,13 +11,10 @@ import { ReportBuilder } from '../reportBuilder';
...
-  private aggregator: Aggregator;
-
-  constructor() {
-    this.aggregator = createAggregator(Aggregator);
-  }
-
   fetchReport = async (req: ReportsRequest, res: Response): Promise<void> => {
+    const aggregator = createAggregator(Aggregator, {
+      session: { getAuthHeader: () => req.headers.authorization },
+    });
```
rohan-bes, line 17: Now need a new aggregator per request, so we can associate the user session with it.

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -0,0 +1,134 @@
...
+export const adjustOptionsToAggregationList = async (context, fetchOptions, aggregationOptions) => {
+  const { aggregations: aggregationList = [] } = aggregationOptions;
+  if (aggregationList.length === 0) {
+    return [fetchOptions, aggregationOptions];
+  }
+
+  const { startDate, endDate } = getAdjustedDateRange(fetchOptions, aggregationList);
+  const period = getAdjustedPeriod(fetchOptions, { startDate, endDate });
+
+  const [organisationUnitCodes, aggregations] = await getAdjustedOrganisationUnitsAndAggregations(
```
kael89, line 124: I'd expect to see an object returned here, but you know what - I actually don't mind the array, since `getAdjustedOrganisationUnitsAndAggregations` is local to this file.

I find that although not often used in JavaScript, array return values can make things simpler in some specific cases. So, I'm personally happy with this one 😃

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+/**
+ * Each aggregation in the list may require the original fetch option dimensions (dates, org units etc)
+ * to be adjusted, eg when we need to take into account past/future data
+ *
+ * @param {Object[]} aggregationList
+ */
+export const adjustOptionsToAggregationList = async (context, fetchOptions, aggregationOptions) => {
+  const { aggregations: aggregationList = [] } = aggregationOptions;
```
kael89, line 116: At some point, we should refactor `Aggregator` to expect that
```ts
type AggregationOptions = Aggregation[];
```

While currently, it is
```ts
type AggregationOptions = { aggregation: Aggregation[] };
```

probably because in the past `AggregationOptions` could also be a simple string (or something like that).

There are some utils in `indicators` which allow for polymorphic definition of aggregation. The client can then specify aggregation in various formats (even a simple string), which would be "expanded" to their full `AggregationOptions` format in the main entry to `Aggregator`.

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -0,0 +1,134 @@
...
+    return getDateRange ? getDateRange(currentRange, aggregation.config) : currentRange;
+  }, dateRange);
+
+const shouldFetchDataSourceEntities = ({ config: { dataSourceEntityType } = {} } = {}) =>
+  dataSourceEntityType;
+
+const shouldFetchRelations = ({
+  type,
+  config: { aggregationEntityType, dataSourceEntityType } = {},
+} = {}) => aggregationEntityType && dataSourceEntityType && !(type === AGGREGATION_TYPES.RAW);
```
kael89, line 44: Similarly (and also optionally) here, you can consider optional chaining if it makes things clearer

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -0,0 +1,60 @@
...
+    dataSourceEntityFilter = {}, // TODO: Add support for dataSourceEntityFilter https://github.com/beyondessential/tupaia-backlog/issues/2660
+  ) {
+    const response = await this.get(
+      createRelationsUrl(hierarchyName, entityCodes, aggregationEntityType, dataSourceEntityType),
+    );
+    const formattedRelations = {};
+    Object.entries(response).forEach(([descendant, ancestor]) => {
+      formattedRelations[descendant] = { code: ancestor };
+    });
+    return [Object.keys(formattedRelations), formattedRelations];
```
kael89, line 58: We can just return `formattedRelations` here, then the client can use its `keys` if it wants

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -0,0 +1,134 @@
...
+        dataSourceEntityType,
+        dataSourceEntityFilter,
+      );
+
+      if (!adjustedOrganisationUnitCodes) {
+        adjustedOrganisationUnitCodes = dataSourceEntities;
+      }
+      adjustedAggregations.push({
+        ...aggregation,
+        config: { ...aggregation.config, orgUnitMap: relations },
```
kael89, line 101: Great!!! So much simpler for the `orgUnitMap` to be built here!! 👏 👏

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -0,0 +1,134 @@
...
+      ] = await entityConnection.getDataSourceEntitiesAndRelations(
+        hierarchy,
+        organisationUnitCodes,
+        aggregationEntityType,
+        dataSourceEntityType,
+        dataSourceEntityFilter,
+      );
+
+      if (!adjustedOrganisationUnitCodes) {
+        adjustedOrganisationUnitCodes = dataSourceEntities;
```
kael89, line 97: I see that here (and in line 70 above) `adjustedOrganisationUnitCodes` are set, not appended... Intuitively, I would expect them to be a list of items that can build of with each iteration.

On the other hand, I don't know if this would actually make _sense_: in the same way, I haven't thought whether chaining multiple org unit aggregations makes sense in general, or is even used currently (probably not).

Happy to leave this as is, and not make a decision right now, just something to flag for the future - perhaps we should also add a comment here ("chaining org unit aggregations has unexpected behaviour"). We can also throw an error, in this scenario - or skip this for now, to avoid unnecessary regression risk

packages/aggregator/src/createAggregator.js
```diff
@@ -6,7 +6,7 @@
 import { DataBroker } from '@tupaia/data-broker';
 import { Aggregator } from './Aggregator';
 
-export const createAggregator = (AggregatorClass = Aggregator, ...additionalParams) => {
-  const dataBroker = new DataBroker();
+export const createAggregator = (AggregatorClass = Aggregator, context, ...additionalParams) => {
```
kael89, line 9: This will break existing usages of `createAggregator`, eg in [web-config-server](https://github.com/beyondessential/tupaia/blob/dev/packages/web-config-server/src/apiV1/DataAggregatingRouteHandler.js#L18)

kael89 (commented):
Great work @rohan-bes !! That will makes things much cleaner for `web-config-server`. I also love the addition of `requested`!

I just have one question: do we plan to also migrate `web-config-server` to use `context` in a subsequent PR? Because as is, I believe that it won't work. I personally feel that we may have to take a slightly "pragmatic" (read: hacky) approach, where we check if a `context` object is passed: if yes, establish an `EntityConnection` and create the org unit maps; if not, do nothing (the client should build and provide the maps)

kael89:
@rohan-bes another, perhaps cleaner way to solve this issue:

* Refactor [createAggregator](https://github.com/beyondessential/tupaia/blob/dev/packages/aggregator/src/createAggregator.js) to `const createAggregator = (AggregatorClass = Aggregator, constructorParams, context)` where `constructorParams` is an array
* In `web-config-server`, call it this way: `createAggregator(Aggregator, [this.models, this])`
* As mentioned, build org unit maps in `aggregator` only if a `context` has been provided

Side note: to be honest, I'm not really sure why we have to use `createAggregator` and not just instantiate `Aggregator` objects ourselves. There would be a motivation for establishing this factory pattern, but I can't remember now... Just creating `Aggregator` instances would probably make things simpler!

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/adjustOptionsToAggregationList.test.js
kael89: That's a pity... We should definitely add thorough unit tests for this functionality when we start using it in `web-config-server`!

packages/report-server/src/aggregator/Aggregator.ts
```diff
@@ -17,19 +17,22 @@ export class Aggregator extends BaseAggregator {
   async fetchAnalytics(
     dataElementCodes: string[],
     organisationUnitCodes: string,
+    hierarchy: string | undefined,
     periodParams: PeriodParams,
+    aggregationOptions: unknown,
```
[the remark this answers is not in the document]
rohan-bes: You're right, would totally love to refactor this whole area of code in the report-server. I had a much worse idea of the true expected input to the aggregator back when I wrote it, now I'd probably not even bother having a wrapper class for it and just call the aggregator directly.

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
rohan-bes: Yeah I agree, will do

packages/aggregator/src/connections/EntityConnection.js
rohan-bes: I guess I kinda felt that _knowing_ that the keys of the relations map are the dataSourceEntities was a bit of a concern of the entityConnection... Does that make sense to you? Not too fussed either way so happy to change

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
rohan-bes: Definitely much simpler than the code that builds it in web-config-server 😛

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
rohan-bes: The whole `dataSourceEntities` thing is a bit of a confusing thing in aggregation imo. Truth be told, it's a fetchOption, not an aggregation option. You're saying, fetch me data against all the entities of this type under the requested entity. We even replace the `orgUnitCodes` in the fetch option rather than the aggregation option!

My preferences for how we ought to design it would be:
a) dataSourceEntityType is a fetchOption, which we use to replace the requested org units with as we do here
b) dataSourceEntityType is a top-level property on the aggregation options, which we use to replace the requested org units with as we do here
c) what we currently do, where it's nested inside an aggregation config, and it only makes sense for us to fetch the entities for the very first aggregation which specifies the dataSourceEntityType...


Btw I actually think chaining entity aggregations makes sense and would work with the implementation I have here, eg:
```
fetchOptions = { organisationUnitCodes: ['TO'], hierarchy: 'explore' }

aggregations = [
{
  type: 'FINAL_PER_ORG_GROUP', 
  config: { dataSourceEntityType: 'village', aggregationEntityType: 'facility'}
}, 
{
  type: 'SUM_PER_ORG_GROUP', 
  config: { dataSourceEntityType: 'facility', aggregationEntityType: 'country'}
}]

// After running adjustOptionsToAggregationList

fetchOptions = { organisationUnitCodes: [<all_villages_in_tonga>], hierarchy: 'explore' }

aggregations = [
{
  type: 'FINAL_PER_ORG_GROUP', 
  config: { dataSourceEntityType: 'village', aggregationEntityType: 'facility', orgUnitMap: <village->facility>}
}, 
{
  type: 'SUM_PER_ORG_GROUP', 
  config: { dataSourceEntityType: 'facility', aggregationEntityType: 'country', orgUnitMap: <facility->country>}
}]
``` 
I believe this should correctly compute the sum of the final villlage per facility per country 😄

packages/aggregator/src/createAggregator.js
rohan-bes: You are right, how would you feel about merging this code in to the feature branch even if it breaks web-config-server, with a plan to refactor web-config-server over to using this type of entity-aggregation before we merge the feature branch into `dev`?

>Side note: to be honest, I'm not really sure why we have to use createAggregator and not just instantiate Aggregator objects ourselves. There would be a motivation for establishing this factory pattern, but I can't remember now... Just creating Aggregator instances would probably make things simpler!

I totally agree with this and would like to do it as part of the above mentioned refactor?

packages/aggregator/src/connections/EntityConnection.js
kael89: It makes sense! Let's leave this as is then - just wanted to confirm that it was a conscious decision 🙌

packages/aggregator/src/createAggregator.js
kael89: Agreed! Although I believe we should not migrate `web-config-server` to use the new "entity server in aggregator" functionality as part of a PSSS release. This would unnecessarily increase the regression testing scope.

Fine with `web-config-server` being broken in `psss-phase-2` for now

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
rohan-bes: Yeah I know what you mean... personally I'm a bit torn since I actually quite like our polymorphic format for aggregations (it's nice to be able to pass in either a single or array or either strings or objects). 

I feel maybe Aggregator should be responsible for handling all those different type of input but formatting them into a consistent shape immediately and we use that throughout the rest of Aggregator?

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
rohan-bes: I've been using tuples a lot lately. Personally I've been enjoying them, but I can totally see the downsides...

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -0,0 +1,60 @@
...
+ *
+ */
+import { ApiConnection } from '@tupaia/server-boilerplate';
+
+const { ENTITY_SERVER_API_URL = 'http://localhost:8050/v1' } = process.env;
+
+const createDescendantsUrl = (hierarchyName, entityCodes, dataSourceEntityType) =>
+  `hierarchy/${hierarchyName}/descendants?entities=${entityCodes.join(
+    ',',
+  )}&field=code&descendant_filter=type:${dataSourceEntityType}`;
```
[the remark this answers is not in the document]
rohan-bes: I'm glad you brought this up, the `get()` function in the `ApiConnection` actually already does this and I was just too silly to think to use it!! Thank you!!

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
kael89: They are also convenient as test input, and TypeScript makes them easier to use - since you can define/see the type for each element, which adds context to them.

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
kael89: I definitely like polymorphic input as an option! I actually spend some time making indicators work with various aggregation formats. As you say, we should "expand" them once in Aggregator, or once in each package (since `indicators` also benefit from a consistent internal format)

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
kael89: It should! It seems to depend on the aggregation order, which is a reasonable requirement in my opinion - after all, the client should be aware that aggregations are performed in order.

Great point about `dataSourceEntities` being a fetch option in essence! We may find that we can improve this even further - moving all logic here is a the great, big step! 🚀

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -7,22 +7,12 @@ import { ApiConnection } from '@tupaia/server-boilerplate';
...
-  `hierarchy/${hierarchyName}/relations?entities=${entityCodes.join(
-    ',',
-  )}&field=code&groupBy=descendant&${
-    aggregationEntityType === 'requested' ? '' : `ancestor_filter=type:${aggregationEntityType}`
-  }&descendant_filter=type:${dataSourceEntityType}`;
+const queryParams = (entityCodes, dataSourceEntityType) => ({
+  entities: entityCodes.join(','),
+  descendant_filter: `type:${dataSourceEntityType}`,
+  field: 'code',
+  groupBy: 'descendant',
```
kael89, line 14: `groupBy` will be also used in the `descendants` query, I guess that's fine?

rohan-bes: Ah good point! That is technically incorrect, but practically fine... I should probs remove it tho since it's a bit confusing

kael89 (approved):
Great stuff, I'm glad this could be implemented with clean architectural changes! 💯

PR #2382, opened 2021-04-13
1500: PSSS app endpoint - GET Country metadata

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1500


packages/psss-server/src/connections/EntityConnection.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { PSSS_ENTITY, PSSS_HIERARCHY } from '../constants';
+import { ApiConnection } from './ApiConnection';
+
+const { ENTITY_URL = 'http://localhost:8050/v1' } = process.env;
```
rohan-bes, line 9:
```suggestion
const { ENTITY_SERVER_API_URL = 'http://localhost:8050/v1' } = process.env;
```
I think that's the convention we have for naming env vars for the server urls

packages/psss-server/src/connections/ReportConnection.ts
```diff
@@ -5,7 +5,7 @@
 
 import { ApiConnection } from './ApiConnection';
 
-const { REPORT_API_URL = 'http://localhost:8030/v2' } = process.env;
+const { REPORT_API_URL = 'http://localhost:8030/v1' } = process.env;
```
rohan-bes, line 8: Nice catch!

packages/psss-server/examples.http
```diff
@@ -39,7 +39,7 @@ content-type: application/json
 DELETE http://localhost:8040/v1/weeklyReport/TO?week=2021W07 HTTP/2.0
 
 ### Fetch Confirmed Weekly Report Data
-GET http://localhost:8040/v1/confirmedWeeklyReport?startWeek=2021W07&endWeek=2021W07 HTTP/2.0
+GET http://localhost:8040/v1/confirmedWeeklyReport?startWeek=2021W07&endWeek=2021W07&orgUnitCodes=TO&orgUnitCodes=FJ HTTP/2.0
```
kael89, line 42: We could also pass `orgUnitCodes` as a comma-separated string, which would make the URL easier to read.

On the other hand, this would make invoking the API a tiny bit more complicated since we'd need to stringify an array parameter. I decided to not go this way and keep the programmatic invocation of the API simple, but let me know if you think that it would be preferable

rohan-bes, line 42: While I get that this works out of the box, I do find it a tad ugly. We have a bit of a convention of providing a comma separated list if we want an array of org units?

packages/psss-server/src/connections/EntityConnection.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type EntityObject = {
+  id: string;
+  code: string;
+  name: string;
+};
+
+interface EntityFilter {
+  type?: string | string[];
+}
```
kael89, line 19: @rohan-bes We could use some TS magic here by importing the fuller types defined in `entity-server`, but then again we'd make `psss-server` dependent on `entity-server`, if only for `dev` purposes.

Keen to hear about ways to share types between orchestration servers and microservices! For the time being, I thought that defining what we need should be enough here.

packages/psss-server/src/connections/EntityConnection.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const buildFilter = (fields: EntityFilter) =>
+  Object.entries(fields)
+    .map(([key, value]) => `${key}:${value}`)
+    .join(';');
```
kael89, line 24: @rohan-bes similarly here: this repeats logic existing in `entity-server`, but I didn't want to add that package as a dependency of `psss-server`

rohan-bes (changes requested):
Just a few small suggestions but overall looks good 😄

packages/psss-server/src/routes/country/FetchCountries.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export class FetchCountries extends Route {
+  async buildResponse() {
+    const countries = await this.entityConnection.fetchCountries();
+    const { accessPolicy } = await this.getSession();
+    const allowedEntities = accessPolicy.getEntitiesAllowed(PSSS_PERMISSION_GROUP);
+
+    const countryData = countries
+      .filter(c => allowedEntities.includes(c.code))
+      .sort(getSortByKey('name'));
```
kael89, line 18: Perhaps `entity-server` could support ordering, in the meantime I believe it's fine to do so here

packages/psss-server/examples.http
kael89: Just commented on that in parallel - happy to change this!

packages/psss-server/src/connections/EntityConnection.ts
kael89: Good point (it seems that it's actually `ENTITY_API_URL`)

packages/psss-server/src/connections/EntityConnection.ts
rohan-bes: Hmmm it's a good question... I think for now it's fine for us to keep them decoupled... Perhaps in the future it might be nice for each micro-service to define some exposed interface types that clients can use to correctly format requests... I think we still need some time to see how things evolve before trying to tightly couple everything 😄

packages/psss-server/src/routes/country/FetchCountries.ts
rohan-bes: I think @tcaiger created a ticket for us to add this functionality: https://app.zenhub.com/workspaces/active-sprints-5eea9d3de8519e0019186490/issues/beyondessential/tupaia-backlog/2658

packages/psss-server/src/connections/EntityConnection.ts
kael89: I agree!

packages/psss-server/.env.example
```diff
@@ -2,6 +2,7 @@ DB_NAME=
 DB_PASSWORD=
 DB_URL=
 DB_USER=
+ENTITY_API_URL=
```
rohan-bes, line 5: Hmm it's definitely called the `ENTITY_SERVER_API_URL` in the `lesmis-server` however I actually think I prefer your naming convention better... Perhaps if you or @tcaiger could change it there?

rohan-bes (approved):
Just one last comment about the naming of the entity server url env var, but won't hold up approving 😄

packages/psss-server/.env.example
kael89: I am following the existing convention in [MeditrakConnection](https://github.com/beyondessential/tupaia/blob/dev/packages/psss-server/src/connections/MeditrakConnection.ts#L10), [ReportConnection](https://github.com/beyondessential/tupaia/blob/dev/packages/psss-server/src/connections/ReportConnection.ts#L8)

kael89: @rohan-bes I'm gonna merge this in the epic branch, and we can resolve the naming convention later

PR #2385, opened 2021-04-13
Make /me endpoint permission free

edmofro:
The `/me` endpoint has self-enforcing permissions: so long as we are looking for the user based on the one attached to the auth header, there's no issue (someone should always be able to see their own info)

This removes permission checking from `/me` (and makes it a little more consistent with other handlers too)


tcaiger:
@edmofro - I was just testing this and noticed that it returns the api user in the response. Is it possible to check on the server if it is an api user and return a special public user response in a similar way to web-config-server? If it's not easy then I'm not too bothered by this from a front end point of view.

edmofro:
I think it'd be best to put that switch in lesmis-server itself. I'll work on that now? So you'd like a null response if the user hasn't logged in, right?

tcaiger:
Yep that would be good.

edmofro:
Done!

EMcQ-BES (approved):
LGTM, love how clean that `GETUserForMe` class is

edmofro:
Given it is only LESMIS (and commented out sections of PSSS) that use the `/me` endpoint, I'm going to merge this for testing in dev

PR #2388, opened 2021-04-14
2515 lesmis map overlays

tcaiger:
### Issue #: [2515](https://github.com/beyondessential/tupaia-backlog/issues/2515) lesmis map overlay selector

This PR is for adding the measures route to lesmis server which is a pre-requisite for adding the map overlay selector front end component. The PR for the will follow.

### Changes:

- Add `map-overlays` route to lesmis server
- Add MapOverlaySelector component

---


packages/lesmis-server/src/connections/WebConfigConnection.ts
```diff
@@ -11,10 +11,14 @@ export class WebConfigConnection extends SessionHandlingApiConnection {
...
+    return this.get('view', query);
   }
 
   async fetchMapOverlay(query: QueryParameters) {
-    return this.get(`measureData`, query);
+    return this.get('measureData', query);
+  }
+
+  async fetchMeasures(query: QueryParameters) {
+    return this.get('measures', query);
```
edmofro, line 22: These two could become `fetchMapOverlayData` and `fetchMapOverlays`

edmofro (changes requested):
All the code and structure looks great. However, I would like to reduce the plurality of terms in LESMIS - no need to propagate across the confusion of an organically grown system from Tupaia. 

I think I've mentioned this before with `dashboardReport` and `view` - in LESMIS we shouldn't use `view` anywhere in my opinion, outside of probably the `WebConfigConnection` class that does the `/view` fetch. 

Similarly, in this case I don't see any reason to maintain the two names `mapOverlays` and `measures` - let's just use `mapOverlays` everywhere.

tcaiger:
Yes good points about the naming. I have renamed to `mapOverlays` everywhere and removed any references to `view` which are outside of web config connection.

edmofro (approved):
Thanks, lgtm!

tcaiger:
Closing as have made a new PR with these changes for the MVP [here](https://github.com/beyondessential/tupaia/pull/2462)

PR #2389, opened 2021-04-14
2515 lesmis map overlay selector part 2

tcaiger:
### Changes:

- Add measure panel component
- Add query to lesmis server for map overlay data
- Small breadcrumbs fix
---


packages/lesmis/src/components/MapOverlaysPanel/index.js
```diff
@@ -0,0 +1,7 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ *
+ */
+export * from './MapOverlaysPanel';
+export * from './MapOverlaysPanelContainer';
```
tcaiger, line 7: The container is used in another place later for the map section which is why it is it's own component

packages/lesmis/src/components/MapOverlaysPanel/MapOverlaysPanel.js
```diff
@@ -0,0 +1,54 @@
...
+import { MapOverlayGroup } from './MapOverlayGroup';
+import { MapOverlaysLoader } from './MapOverlaysLoader';
+import { MapOverlaysPanelContainer as Container } from './MapOverlaysPanelContainer';
+
+const Body = styled.div`
+  padding: 0.2rem 0 2rem;
+`;
+
+export const MapOverlaysPanel = ({ overlays, isLoading, overlayId, setOverylayId }) => {
+  const [selectedPath, setSelectedPath] = useState(null);
```
tcaiger, line 18: Part of the design is to highlight the path to the selected overlay. This is the state that stores that path so that the nested components can be highlighted accordingly. example path eg. [0, 0, 1]

packages/lesmis/src/components/MapOverlaysPanel/MapOverlayGroup.js
```diff
@@ -0,0 +1,368 @@
...
+  .MuiSvgIcon-root {
+    font-size: 1.5rem;
+  }
+`;
+
+/**
+ * Calculates whether the map overlay group is the selected one or is part of the
+ * path to the selected one
+ */
+const getActiveClass = (path, selectedPath) => {
```
edmofro, line 254: My main suggestion out of this PR is to change the path to hold a list of codes rather than indexes. A couple of advantages to that:
- Allow easier recreation of selected path from a deep link
- Avoid any bugs where the components rerender and for some reason change order
- A little easier to interpret when debugging, if you print out the selected path

Fortunately, I don't think there is anything about your current setup that precludes that from happening! Hopefully it's just swapping out the setting of the path from using indexes to, ideally, `option.code`*, but if you want to avoid any back end changes, `option.name` for now. 

*The map overlay groups don't currently include their code, though it would be easy enough to change that and would be non-breaking on web-config-server. Also, the overlays use `measureId`, but I feel that could be renamed to just `code` on LESMIS.

edmofro (changes requested):
This is super nice and clean structurally! I just have one significant request, but mostly all looking great

packages/lesmis/src/views/MapView.js
```diff
@@ -19,10 +23,21 @@ const Main = styled.div`
...
+  const { entityCode } = useUrlParams();
+  const [overlayId, setOverylayId] = useState(null);
+
+  const { data: overlaysData, isLoading } = useMapOverlaysData({ entityCode });
+
+  return (
+    <Container>
+      <MapOverlaysPanel
+        isLoading={isLoading}
+        overlays={overlaysData?.measures}
```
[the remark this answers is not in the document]
tcaiger: Good idea - done

packages/lesmis/src/utils/useEntityBreadcrumbs.js
```diff
@@ -7,24 +7,24 @@ import { useUrlParams } from './useUrlParams';
 import { useEntitiesData } from '../api';
 import { makeEntityLink } from './makeEntityLink';
 
-const getHierarchy = (entities, entityCode, hierarchy = []) => {
+const getHierarchy = (entities, entityCode, view, hierarchy = []) => {
```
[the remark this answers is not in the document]
tcaiger: `view` here refers to the url param which holds the `dashboard` vs `map` toggle state. eg `Route path="/:entityCode/:view?"`

packages/lesmis/src/components/MapOverlaysPanel/MapOverlayGroup.js
```diff
@@ -0,0 +1,368 @@
...
+                overlayId={overlayId}
+                setOverylayId={setOverylayId}
+                selectedPath={selectedPath}
+                setSelectedPath={setSelectedPath}
+                path={[...path, index]}
+              />
+            ) : (
+              <FormControlLabel
+                control={<Radio />}
+                key={option.measureId}
```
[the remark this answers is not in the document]
tcaiger: Yes I agree that is confusing. I'm not that keen to update it in `lesmis-server` as it involves iterating through a nested structure which is a bit of work and is a small hit for performance. I think this data structure is temporary anyway right? As a short term fix to keep lesmis front end clean I have added code to the response from `meditrak-server`.

packages/lesmis/src/components/MapOverlaysPanel/MapOverlayGroup.js
tcaiger: I agree that it would be better if possible to use a code rather than an index. However I started implementing it and although it was easy enough to add the code to the meditrak response it actually doesn't easily work with the algorithm to get the active class because we need to know where or not any given node is before or after the selected node to highlight the path down to it. It would work if the overlay list was alphabetical but it is not.

tcaiger: Regarding the 3 points above. 
- I think the first one is a valid point, it could be done at the moment but it would be a bit awkward
- I think point 2 is quite hypothetical and at least as likely to happen with the codes
- I think point 3 is negligible difference

packages/lesmis/src/utils/useEntityBreadcrumbs.js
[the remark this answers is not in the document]
edmofro: Oooh perfect! That's exactly what `view` should mean 🙌

edmofro (approved):
Good chats, agree this is the best solution given that tricky requirement of having to highlight the nodes on the way down!

Updates all look good :-)

PR #2344, opened 2021-04-07
2636: Allow fetching tupaia events without data elements

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2636

PR #2390, opened 2021-04-14
2177-2636: Convert STRIVE data to be internal

kael89:
Includes commits already approved in the following PRs:

* https://github.com/beyondessential/tupaia/pull/2118: first 37 commits
  * 2101: Added Materialized View Analytics Table
  * ...
  * Remove unused import

* https://github.com/beyondessential/tupaia/pull/2344: commits 38 and 39
  * Allow empty dataElementCodes in tupaia event fetch
  * 2177: Use tupaia service in Strive

The rest are `dev` merges. The commit history is a bit messy because we already squashed the first 37 commits in `dev`


packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js
```diff
@@ -59,6 +65,7 @@ export class DataBuilder {
     };
 
     return this.aggregator.fetchAnalytics(dataElementCodes, fetchOptions, this.query, {
+      aggregations,
```
kael89, line 68: This could break existing functionality. I checked existing reports,overlays, data and measure builders and none of them seems to have an `aggregation` field in their config - so I guess we're fine!

`aggregations` will be used if defined, if not, `aggregationType`/`aggregationConfig` - see [buildAggregationOptions](https://github.com/beyondessential/tupaia/blob/dev/packages/web-config-server/src/aggregator/buildAggregationOptions.js#L32)

We should certainly refactor how we handle aggregation in data builders - it's highly inconsistent!

kael89:
@chris-bes every commit here is reviewed already (see description) apart from the last 2 which are related to a review you've done, could you please have a look? 🙏

chris-bes (approved):
LGTM!

PR #2393, opened 2021-04-14
2489 lesmis login integration

tcaiger:
### Issue #: [2489](https://github.com/beyondessential/tupaia-backlog/issues/2489) lesmis login integration

Various changes related to the integration of lesmis frontend and lesmis server.

### Changes:

- Delete private route from lesmis frontend as it's not used
- Add getUser check to login flow
- Tweak public user response to be empty object
- Split up verifyLogin in server-boilerplate
- Add login verification to lesmis-server

---


packages/lesmis-server/src/utils/verifyLoginAccess.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ *
+ */
+import { AccessPolicy } from '@tupaia/access-policy';
+import { PermissionsError } from '@tupaia/utils';
+import { LESMIS_PERMISSION_GROUP } from '../constants';
+
+export const verifyLoginAccess = (accessPolicy: AccessPolicy) => {
+  const hasAccess = new AccessPolicy(accessPolicy.policy).allowsAnywhere(LESMIS_PERMISSION_GROUP);
```
edmofro, line 11: Is it this, or is it "has at least Public access to Laos"? I guess this is better in terms of actually being able to log in. Yep, I'm on board.

packages/lesmis/src/api/mutations/useLogin.js
```diff
@@ -27,5 +28,21 @@ export const useLogin = () => {
...
+
+  if (loginQuery.isSuccess) {
+    return { ...loginQuery, ...userQuery };
+  }
+
+  /**
+   * If the login was not successful simply return the errors and fetching state related
+   * to the login query
+   */
+  return loginQuery;
```
edmofro, line 47: Could this go before the `useUser` as an early return? Or do you reckon the narrative reads better the way you've got it?

edmofro: Oh, is it because `useUser` needs to have `enabled: false` passed through so that future usage of it in other parts of the app will get some sort of disabled message?? Does the `enabled: false` persist to other calls until `enabled: true` is passed through?

edmofro (approved):
LGTM! One minor suggestion but keen to learn if there's even more magic to react-query than I realise

packages/lesmis/src/api/mutations/useLogin.js
tcaiger: No no magic - as far as I'm aware the enabled prop won't get persisted or anything. The reason I didn't add the return before the useUser hook is because of one of the react [rules of hooks](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level) which is "Only Call Hooks at the Top Level".

edmofro: Aaah right! Of course, that makes sense. Thanks :-)

tcaiger:
Hey, would you mind quickly checking that last commit makes sense. Just moved the verifyLogin check to above when the session gets created so that the user isn't logged in if verification throws the error.

packages/lesmis-server/src/utils/verifyLoginAccess.ts
```diff
@@ -7,8 +7,8 @@ import { AccessPolicy } from '@tupaia/access-policy';
 import { PermissionsError } from '@tupaia/utils';
 import { LESMIS_PERMISSION_GROUP } from '../constants';
 
-export const verifyLoginAccess = (accessPolicy: AccessPolicy) => {
-  const hasAccess = new AccessPolicy(accessPolicy.policy).allowsAnywhere(LESMIS_PERMISSION_GROUP);
+export const verifyLoginAccess = (accessPolicy: any) => {
```
edmofro, line 10: We should try to be explicit with types where we can. I'm not sure if this is a `string` or a POJO (which we generally use `Record<string, any>` for), but either way it would be worth marking as such

edmofro (approved):
Yep makes a lot of sense to me, nice!

edmofro (approved):

PR #2397, opened 2021-04-15
E2e test fixes

kael89:
Addressing issues that were identified during the first e2e test workshop. See the commit for an overview of the changes


packages/devops/scripts/ci/testE2e.sh
```diff
@@ -4,7 +4,7 @@ DIR=$(dirname "$0")
 DEPLOYMENT_URL=$(${DIR}/determineDeploymentUrl.sh)
 if curl --output /dev/null --silent --head --fail $DEPLOYMENT_URL; then
     echo "Deployment for ${CI_BRANCH} exists, running e2e tests"
-    ssh ubuntu@$DEPLOYMENT_URL "cd tupaia && yarn workspace @tupaia/${PACKAGE} test-e2e"
+    ssh ubuntu@$DEPLOYMENT_URL "cd tupaia && yarn workspace @tupaia/${PACKAGE} test-e2e --ciBuildId ${CI_BUILD_ID}"
```
edmofro, line 7: Is it cypress or codeship that provides `CI_BUILD_ID` as an env var? If it's codeship, convenient that it matches cypress terminology!

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
```diff
@@ -3,7 +3,7 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-import config from '../config/dashboardReports.json';
+import reportUrls from '../config/dashboardReports.json';
```
edmofro, line 6: Much more descriptive!

packages/web-frontend/cypress.json
```diff
@@ -3,8 +3,7 @@
   "testFiles": "*.spec.js",
   "integrationFolder": "cypress/e2e",
   "projectId": "pty5cc",
-  "requestTimeout": 30000,
-  "responseTimeout": 60000,
```
edmofro, line 7: No responseTimeout? I guess requestTimeout covers all we need?

kael89 (commented):
Sorry @edmofro - I had some comments ready but didn't submit them

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
```diff
@@ -27,17 +28,23 @@ tar xf "$dump_file"
 
 echo "Disabling notification triggers"
 psql -U tupaia tupaia -c "ALTER TABLE survey_response DISABLE TRIGGER survey_response_trigger;"
```
edmofro, line 30: You could instead use `all` as per https://beyondessential.slack.com/archives/C0191QHH3QE/p1616128907004600

packages/web-frontend/cypress.json
kael89: The [default responseTimeout](https://docs.cypress.io/guides/references/configuration#Timeouts) of 30secs will be used, it seems that this is enough

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
```diff
@@ -27,17 +28,23 @@ tar xf "$dump_file"
...
+psql -U tupaia tupaia -c "ALTER TABLE survey_response DISABLE TRIGGER trig\$_survey_response;"
 psql -U tupaia tupaia -c "ALTER TABLE answer DISABLE TRIGGER answer_trigger;"
+psql -U tupaia tupaia -c "ALTER TABLE answer DISABLE TRIGGER trig\$_answer;"
 
-echo "Clearing survey response and answer tables"
+echo "Clearing tables: survey_response, answer, analytisc"
```
edmofro, line 35: typo 👀

packages/devops/scripts/ci/testE2e.sh
kael89: It's Codeship! Cypress devs must had Codeship Pro in mind - it's the only [CI/CD platform](https://docs.cypress.io/guides/guides/parallelization#CI-Build-ID-environment-variables-by-provider) whose name matches the cypress option! 😆

packages/devops/scripts/deployment/pushE2eDumpLogToS3.sh
```diff
@@ -0,0 +1,27 @@
...
+    force=true
+fi
+
+local_log="/home/ubuntu/logs/dump_e2e_data.log"
+s3_folder_uri="s3://tupaia/dumps/e2e"
+
+if [[ "$STAGE" == "e2e" || $force == true ]]; then
+    s3_log=$(sed -En "s|.*$s3_folder_uri/(.*).tgz|\1.log|p" $local_log)
+
+    echo "Pushing $remote_log to S3"
```
edmofro, line 22: $remote_log  -> $s3_log ?

packages/devops/scripts/deployment/pushE2eDumpLogToS3.sh
```diff
@@ -0,0 +1,27 @@
...
+
+if [[ $1 == "--force" ]]; then
+    force=true
+fi
+
+local_log="/home/ubuntu/logs/dump_e2e_data.log"
+s3_folder_uri="s3://tupaia/dumps/e2e"
+
+if [[ "$STAGE" == "e2e" || $force == true ]]; then
+    s3_log=$(sed -En "s|.*$s3_folder_uri/(.*).tgz|\1.log|p" $local_log)
```
edmofro, line 20: I can't parse this so am just trusting it!

packages/devops/scripts/deployment/dumpE2eDataToS3.sh
```diff
@@ -7,29 +7,40 @@ export PATH=/home/ubuntu/.local/bin:/home/ubuntu/.yarn/bin:/home/ubuntu/.config/
...
 
-  echo "Dumping survey_response"
-  psql -U tupaia tupaia -c "\copy survey_response to 'dumps/survey_response.csv' delimiter ',' csv header"
+  dump_table survey_response
+  dump_table answer
+  dump_table analytics
 
-  echo "Dumping answer"
-  psql -U tupaia tupaia -c "\copy answer to 'dumps/answer.csv' delimiter ',' csv header"
+  dump_file="${STAGE}_$(date '+%s')_dumps.tgz"
```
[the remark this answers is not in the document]
edmofro: S3 is also pretty bloody cheap, so there's no rush

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
kael89: Good idea, will do!

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
```diff
@@ -1,12 +1,29 @@
...
 echo "Restoring E2E data from S3"
 
-echo "Pulling dumps from S3"
-aws s3 cp s3://tupaia/dumps/e2e/dumps.tgz .
+echo "Pulling latest dump from S3"
+s3_dump_path=$(
+  aws s3api list-objects-v2 \
+    --bucket tupaia \
+    --prefix $s3_folder/ \
+    --query 'reverse(sort_by(Contents[?ends_with(Key, `.tgz`)], &LastModified))[:1].Key' \
```
edmofro, line 20: Hopefully this query works better than the UI, which only sorts the first page of data rather than all

edmofro (approved):
Looks good to me! Just a few comments which I've dotted around the place

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
kael89: Haha really?! That's unexpected 😆

packages/devops/scripts/ci/testE2e.sh
edmofro: Ha, nice!

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
edmofro: So unexpected I feel it's worth double checking this query option doesn't do the same...

packages/devops/scripts/deployment/pushE2eDumpLogToS3.sh
kael89: Oops correct! It seems that `aws s3` also prints this info, so I'll delete the statement

packages/devops/scripts/deployment/pushE2eDumpLogToS3.sh
kael89: Haha let me try to explain then - I'll tell you this line was more complicated on my first iteration:

* `-E` enables extended regex - that way we avoid escaping `()` in the regex
* Did you know you can use separators other than `/` in `sed`? I didn't! It's very convenient here, since we avoid escaping all the `/` in `$s3_folder_uri`. I used `|` as a separator, which is an invalid path char
* `\1` is the first matching group in the regex

So in summary: find a line which includes `$s3_folder_uri/` => `s3://tupaia/dumps/e2e/`, then grab everything after it and until `.tgz` - this will be the base name of the generated file. Finally, append `.log` to it. Eg
Input:  `(...other stuff) e2e-test-fixes_1618393495_dumps.tgz`
Output:  `e2e-test-fixes_1618393495_dumps.log`

edmofro: Nice!

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
kael89: Tried `psql -U tupaia tupaia -c "ALTER TABLE entity DISABLE TRIGGER all"` and got:

> ERROR:  permission denied: "RI_ConstraintTrigger_a_109476" is a system trigger

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
edmofro: I've just checked @kael89 and I do think this will break if we end up with more than one page of dumps...

"Client-side filtering is supported by the AWS CLI client using the --query parameter. This parameter has capabilities the server-side filtering may not have." ~ https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-filter.html

Perhaps the best option is to just disable pagination, and hope that we don't blow things up:
"You can disable pagination by providing the --no-paginate argument" ~ https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
edmofro: Oh damn 😞

kael89: Same in my local

packages/devops/scripts/deployment/restoreE2eDataFromS3.sh
kael89: Thanks for checking! "Special capabilities"... ahaha

edmofro (approved):

PR #2398, opened 2021-04-15
2412: Implement DHIS API fetch of indicator values

chris-bes:
### Issue: https://github.com/beyondessential/tupaia-backlog/issues/2412

# Changes:
- Implemented fetching indicator values
- Removed hacky code that fetches indicator values directly from DhisApi from some data builders

# Context:
Tupaia uses `analytics/rawData` to fetch raw analytics and aggregate data from our end. This has been working well until we need to fetch indicator values for Laos DHIS2. Unfortunately, `analytics/rawData` does not return any data for indicators (more explanations here: https://github.com/beyondessential/tupaia-backlog/issues/2412#issuecomment-806250541). 

Even more unfortunate, as I've checked the documentation and also dived into DHIS2 source code, **the only endpoint that can return indicator values is `analytics.json`**. This endpoint returns aggregated data instead of raw data:
- Data will be aggregated (default aggregation is the data element's aggregation type) for all organisation units below the requested organisation unit in the hierarchy. Eg: If you request for Laos, and there are data in the school level, instead of sending back data from each school org unit like `analytics/rawData`, it only sends back 1 data object with value of the total sum (or whatever aggregation) of all schools under Laos.
- Data will be aggregated for the periods as well. Eg: if data is submitted daily (eg: `20200401`), and you request monthly periods (eg: `202004`), all daily data will be aggregated (eg: SUM) to the month period granularity. Or if you request with startDate and endDate, it will also aggregate data within that date range into 1 value object.

_So, in general, to use `analytics.json` endpoint, but still make it fit into our system (there are a lot of useful custom aggregations/entity aggregations we've written on our side that we can't get rid of), we have to request DHIS2 with the correct period granularity and data source org unit level to prevent the aggregation_. 

Eg: If the data is daily at school level, and we want to get the data from 2017 to 2020, we have to request with:
- periods: `20170101;20170102;20170103;20170104;...;20201231` (at the moment, with rawData endpoint, we can just use startDate, endDate, or monthly period will return all raw data within that month/range of startDate endDate)
- orgUnits: `SchoolA;SchoolB;SchoolC;SchoolD` (at the moment, with rawData endpoint, we can just use Laos/DistrictA/... and it will return all raw data within that entity)
# Solutions:
We have discussed in the refinement and a few solutions were proposed:

## Plan A:
Change the method to pull analytics from `analytics/rawData` to `analytics.json` (so will globally applied to data elements). 2/3 of the time I've spent was to investigating on this solution.

Pros:

- No distinction between fetching indicator values or dataelement values. It will just work transparently

Cons:

- Likely to bump into a lot of hidden problems since we're switching to a new endpoint, a lot of regression testing (basically most of the vizes).

### What I've tried:
- Added `dataSourceEntityType` to all the visualisations request for org units at the exact data level (similar to internal data).
- Converted requested periods to the period granularity of the data (eg: if the data is submitted daily, always request for daily periods `20200101;...`

### Results:
Hitting a few bumps with this solution:

**Issue 1**: For visualisations that have time selector, tupaia is smart enough to convert them into the correct period ranges (eg: if time selector is `month` or `one_month_at_a_time`, startDate, endDate are converted to periods `201801;201802;...`). However, there are a lot of visualisations that do not have time selector, especially map overlays. With those visualisations that have no time selector, tupaia cannot work out which period type it should convert to and will just request with default monthly periods from 2017 til now. This works well with `analytics/rawData` endpoint, but doesn't for `analytics.json` endpoint because if the data is daily, it will aggregate to months. Also, if the data is monthly and we request with daily periods, it will not return any data. So we have to somehow get all the data element's submitted period types, and config them to all the dhis data sources or visualisations.

**Issue 2**: Performance becomes worse because we are requesting with a lot more periods. Batching is used to prevent long URLs but will take 2 or 3 more requests if the data is daily and visualisation is showing data within a long time range.

My general feeling after experimenting with this solution is, we are stretching the aggregate `analytics.json` endpoint to work like `rawData` even though the endpoint is definitely not designed for fetching raw data. Because of this, even if we can solve all the problems I've found above, my gut feeling is there are more hidden problems that I just haven't found out... So to fetch indicator values, I'm leaning towards using the `analytics.json` endpoint only for fetching indicator values (because that's the only way that we can fetch indicator values so be it).

## Plan B:
Have a flag in the data_source so that we know if one is an indicator. If the data_source is an indicator, use aggregate `analytics.json` endpoint instead. 

### What I've tried:
- In data_source config, add `dhisDataType: Indicator` if it is an indicator, the type column is still `dataElement`. If data-broker sees this flag, it will request analytics using the aggregate `analytics.json` endpoint instead, otherwise always use `analytics/rawData`. I chose this way instead of setting type column `indicator` because I feel that indicator on DHIS2 is not much different from a data element (it still returns analytics format) except that we cannot delete/push data to it. If we set type column = `indicator`, I feel that it is going to create unnecessary distinction between them in a some places.
- Introduce a generic config in `dashboardReport.dataBuilderConfig` and `mapOverlay.measureBuilderConfig` which is `dataPeriodType`. This will help force the `QueryBuilder` to convert the period ranges into the correct period granularity that if the visualisation does not have any time selector and need to fetch indicator values using aggregated `analytics.json` endpoint (to solve **issue 1** in Plan A).

### Results:
- Even though aggregate `analytics.json` endpoint is not what we want to use in my opinion (but we have to as it's the only way to get indicator values), I feel that it is safer to use it only for fetching indicator values instead of making it a global change.  So this PR is for implementing Plan B.


packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -133,7 +138,7 @@ export class DhisService extends Service {
   }
 
   getPullAnalyticsForApiMethod = options => {
-    const { programCodes } = options;
+    const { programCodes, dhisDataType } = options;
```
chris-bes, line 141: `dhisDataType: 'Indicator'` is specified in the config of the data_source. I chose this way instead of setting type column indicator because I feel that indicator on DHIS2 is not much different from a data element (it still returns analytics format) except that we cannot delete/push data to it. If we set type column = indicator, I feel that it is going to create unnecessary distinguishment between them in a some places.

packages/web-config-server/src/tests/apiV1/dataBuilders/generic/table/tableOfDataValues/tableOfDataValues.test.js
```diff
@@ -2,27 +2,13 @@
  * Tupaia Config Server
```
chris-bes, line 2: Removing hacky code

packages/web-config-server/src/apiV1/utils/fetchIndicatorValues/checkAllDataElementsAreDhisIndicators.js
```diff
@@ -9,14 +9,13 @@
...
     code: dataElementCodes,
-    type: 'dataElement',
   });
 
   for (const dataElementCode of dataElementCodes) {
     const dataElement = dataElements.find(d => d.code === dataElementCode);
     if (
       !dataElement.config.dhisId ||
-      !['Indicator', 'ProgramIndicator'].includes(dataElement.config.dhisDataType)
+      !['ProgramIndicator'].includes(dataElement.config.dhisDataType)
```
chris-bes, line 18: Removing hacky code. Will still need to keep the hacky code for fetching `ProgramIndicator`. But it's already handled here https://github.com/beyondessential/tupaia/pull/2334

packages/web-config-server/src/apiV1/measureBuilders/checkMultiConditionsByOrgUnit.js
```diff
@@ -7,10 +7,6 @@ import groupBy from 'lodash.groupby';
```
chris-bes, line 7: Removing hacky code

packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfDataValues/tableOfDataValues.js
```diff
@@ -21,10 +21,6 @@ import {
   buildCategoryData,
```
chris-bes, line 21: Removing hacky code

packages/web-config-server/src/apiV1/dataBuilders/generic/percentage/percentagesOfValueCounts.js
```diff
@@ -11,10 +11,6 @@ import {
   countAnalyticsThatSatisfyConditions,
```
chris-bes, line 11: Removing hacky code

edmofro:
Wonderful write up in the description @chris-bes! Looks like you've put in a lot of work and tried several things, and I would tend to agree with all of your conclusions.

Just to get the thought out of my head, one final possibility would be to _ask DHIS2_ if a given code relates to an indicator or a data element, but I'm honestly really happy with that being hard coded into the `data_source` table and being part of the data stewards' responsibility when describing new data sources.

I haven't looked at the code yet, but just wanted to say thanks for all the hard work and clear briefing notes!

packages/web-config-server/src/apiV1/utils/fetchIndicatorValues/checkAllDataElementsAreDhisIndicators.js
```diff
@@ -9,14 +9,13 @@
 export const checkAllDataElementsAreDhisIndicators = async (models, dataElementCodes) => {
   const dataElements = await models.dataSource.find({
     code: dataElementCodes,
-    type: 'dataElement',
```
edmofro, line 12: Any reason for removing this? I would have thought some codes could be duplicated between data elements and data groups, so better to be safe?

packages/web-config-server/src/aggregator/QueryBuilder.js
```diff
@@ -61,9 +61,18 @@ export class QueryBuilder {
...
-    if (this.replacementValues.period) {
+
+    // 'dataPeriodType' is used when you want to force converting the periods to a specific period type. 
+    // Eg, if 'dataPeriodType' = 'MONTH', force converting to MONTHLY periods '201701;201702;201703;201704;...'
+    
+    // This is useful in the scenario when you are fetching indicator values which has to request the aggregate `analytics.json` endpoint
+    // and we have to send the correct period type that the data is submitted. Normally this is taken care of if the visualisation has the correct time selector,
+    // but if it does not, we have to manually convert the periods to the right type so that `analytics.json` can return the correct data without aggregation.
+    if (this.query.startDate && this.query.endDate && dataPeriodType) {
+      this.query.period = convertDateRangeToPeriodQueryString(this.query.startDate, this.query.endDate, dataPeriodType);
```
edmofro, line 74: I wonder if this is the right place for that? Other clients of data broker may want to fetch DHIS2 indicators, in which case they have to repeat this logic. An alternative would be pass through the query parameter `dataPeriodType` and have data broker pass that to DhisApi, and have it convert there if it is an indicator. I can't remember whether it's easy to pass config all the way through though...

edmofro (commented):
I've just had a 👀  rather than a full review, but I agree with everything I've seen! Great job @chris-bes 🙏 

Only two comments inline, and also a general nervousness that the more complexity we introduce, the more performance hit we might see. E.g. we seem to be fetching data sources from the db quite a few times, which is an extra bit of lag with each fetch, but I think this has at least been kept to conditionally happen only in the new, more complex code paths. In future if we find this is an issue, we could pass around a shared data sources object or something, but for now it does keep the code as clean as possible!

packages/web-config-server/src/aggregator/QueryBuilder.js
```diff
@@ -61,9 +61,18 @@ export class QueryBuilder {
     // Make standard replacements if required
     this.query.startDate = this.getQueryParameter('startDate');
     this.query.endDate = this.getQueryParameter('endDate');
-
+    const dataPeriodType = this.getQueryParameter('dataPeriodType');
```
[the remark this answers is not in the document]
IgorNadj: Discussed, scenario: two indicators in the same query, one monthly one weekly

packages/data-broker/src/services/dhis/InboundAggregateDataTranslator.js
```diff
@@ -103,7 +103,12 @@ export class InboundAggregateDataTranslator {
       this.dataSources,
       ({ dataElementCode }) => dataElementCode,
     );
-
+    // We can't always guarantee that dhis data elements can have codes (eg: Laos DHIS).
```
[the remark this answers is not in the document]
IgorNadj: Discussed: still lost, would be great to have a comment for why this is different

packages/web-config-server/src/apiV1/utils/fetchIndicatorValues/checkAllDataElementsAreDhisIndicators.js
chris-bes: Sorry I think that was a mistake, will reverted!

packages/web-config-server/src/aggregator/QueryBuilder.js
chris-bes: Yep that's a good point about other clients may want to fetch DHIS2 indicators, Igor also raised a similar concern so I'll try to do that in DhisApi instead. My only justification is that I thought DhisApi shouldn't need to have any logic to modify the query parameters, meaning that the clients should be responsible for passing in the right query parameters.

packages/data-broker/src/services/dhis/InboundAggregateDataTranslator.js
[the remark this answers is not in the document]
chris-bes: It's confusing. There are 2 places that translate the analytics response from DHIS2, `DhisInputSchemeResolvingProxy,js` and `InboundAggregateDataTranslator.js`.

For example: we have a Laos DHIS2 raw analytics response without data element code (because Laos DHIS2 doesn't have codes for data elements):

```
"metaData": {
	"items": {
		"eKWzHRGV4Vy": {
			"uid": "eKWzHRGV4Vy",
			"name": "Data Element ABC",
			"dimensionItemType": "DATA_ELEMENT",
			"valueType": "NUMBER",
		}
	}
}
``` 

The translate response code in `DhisInputSchemeResolvingProxy,js` will translate the `**key**` of the items map from UID to our internal data element code in tupaia. Usually, in a metadata item object, it will also have `code: ABC`, but since Laos DHIS2 doesn't have codes, item object will not have that property

```
"metaData": {
	"items": {
		"Tupaia_Data_Element_Code_ABC": {
			"uid": "eKWzHRGV4Vy",
			"name": "Data Element ABC",
			"dimensionItemType": "DATA_ELEMENT",
			"valueType": "NUMBER",
		}
	}
}
```

And then when we reach the InboundAggregateDataTranslator.js, it will normally try to grab the code from the item object (since we iterate by Object.values(items) and we don't care about the key):
```
     const { uid, code: dhisCode, name } = item;
```

But since metadata item doesn't have any code, `dhisCode` will be undefined and throws an error here.
I guess we can grab the code from the metadata item key instead. But my only concern is the item key is not always the code (eg: if we fetch data from our regional DHIS2, it won't go through DhisInputSchemeResolvingProxy,js` pipeline, so the metadata item key will be uid in that case)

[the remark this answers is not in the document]
IgorNadj: >maybe just fill in the missing metadata item.code
Yeah that's what I was thinking too. I think the only reason it doesn't already is because of my incomplete original implementation... sorry about that.

packages/web-config-server/src/aggregator/QueryBuilder.js
edmofro: Ah yep that's a good point. However, I think what we're doing here is making DhisApi behave as expected (i.e. in line with the other endpoints) given certain query parameters, so it doesn't matter so much that it's altering those query parameters as they go through to DHIS2. I do see that it might be surprising for people if they expect DhisApi to act as per the DHIS2 docs...but this use case seems worth it.

chris-bes:
Just realized tests in DhisService are broken because of the pull logic split refactor. Will fix it up

packages/data-broker/src/services/dhis/pullers/AnalyticsPuller.js
```diff
@@ -0,0 +1,283 @@
...
+        },
+      };
+    };
+    const pullForApi = async api => {
+      if (dhisDataType === this.dataSourceModel.getDhisDataTypes().INDICATOR) {
+        // Multiple DHIS Indicators may have different period types,
+        // so we have to group them by their period types and fetch them separately
+        const groupedDataSourcesByPeriodType = this.groupIndicatorDataSourcesByPeriodType(
+          dataSources,
+        );
```
chris-bes, line 88: This is the logic to handle fetching multiple indicators with different submitted period types

packages/data-broker/src/services/dhis/pullers/AnalyticsPuller.js
```diff
@@ -0,0 +1,283 @@
+/**
```
chris-bes, line 1: I've splitted the pull logic into AnalyticsPuller/EventsPuller

packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.js
```diff
@@ -0,0 +1,46 @@
+/**
```
chris-bes, line 1: Also split pull data element metadata from `DhisService` into this class. The reason is `AnalyticsPuller` also need to pull metadata of data elements, so I pulled it into a class so that it can also be called separately from `AnalyticsPuller` and `DhisService`

packages/data-broker/src/services/dhis/pullers/EventsPuller.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { buildEventsFromDhisEventAnalytics } from '../buildAnalytics';
+
+export class EventsPuller {
```
IgorNadj, line 8: Looks good!

packages/database/src/migrations/20210414015925-UpdateLaosEOCIndicatorDataSources-modifies-data.js
```diff
@@ -0,0 +1,51 @@
...
+    UPDATE data_source
+    SET config = config || '{"indicator": { "dataPeriodType": "MONTH"}}'
+    WHERE code IN (${arrayToDbString(DATA_SOURCE_CODES)});
+  `);
+};
+
+exports.down = function (db) {
+  return db.runSql(`
+    UPDATE data_source
+    SET config = config - 'indicator'
```
IgorNadj, line 44: Never knew you could do this, wow

packages/dhis-api/src/buildAnalyticsQuery.js
```diff
@@ -75,6 +75,26 @@ export const buildDataValueAnalyticsQueries = queryInput => {
   return queries;
 };
 
+export const buildAggregatedDataValueAnalyticsQueries = queryInput => {
+  const newQueryInput = { ...queryInput };
+
```
IgorNadj, line 80: Suggest comment:

All data in dhis has a date granularity, e.g. you can submit an event against a full date, or just December 2018. 

When we are fetching raw data, we can ignore this, because if we request data for 2015-2021 it will return all data, including an event at December 2018 because it is within that range.

But when we fetch aggregated data, the API behaves differently. If we request data for 2015-2021 it will NOT return an event at December 2018. To get the event, we have to request data for 201810 specifically.

This is true for both fetching data element values and fetching indicator values, however, we use the raw data API endpoint most of the time, and we only use the aggregated API endpoint for indicators. So practically, this is an indicator problem, but technically it's for both.

packages/data-broker/src/services/dhis/pullers/AnalyticsPuller.js
```diff
@@ -0,0 +1,286 @@
...
+    this.dataElementsMetadataPuller = dataElementsMetadataPuller;
+  }
+
+  groupDataSourcesByDhisDataType = dataSources =>
+    groupBy(
+      dataSources,
+      d => d.config?.dhisDataType || this.dataSourceModel.getDhisDataTypes().DATA_ELEMENT,
+    );
+
+  groupIndicatorDataSourcesByPeriodType = dataSources => {
```
IgorNadj, line 28: I've also put a comment in buildAggregatedDataValueAnalyticsQueries which I think is correct, read that first and let me know if I'm wrong. But I think this is not just an indicator problem. Technically, someone could come along and fetch data element values through our aggregated method... they would be crazy to do so but they could. Do we want to rename the method here to remove the mention of Indicator? Happy to leave as is, just a thought.

IgorNadj (approved):
LGTM! I've left a few comments on minor things, and there's one thing I should check but yeah looks great

PR #2399, opened 2021-04-15
Clean up timeouts after fetch resolves

edmofro:
If you make a lot of fetches in a row, the timeouts build up and consume memory. Better to clear them as soon as the fetch resolves.

I found this because it was causing out of memory errors in my similar implementation over on Tamanu Mobile, if I ran sync super fast. There's a remote possibility that this has been contributing to our out of memory errors on web-config-server. That's a much bigger memory environment, so it's most likely not a significant factor, but great to clean it up anyway.

### DIFF
--- a/packages/meditrak-app/app/api/TupaiaApi.js
+++ b/packages/meditrak-app/app/api/TupaiaApi.js
@@ -37,24 +37,20 @@ export class TupaiaApi {
   }
 
   async reauthenticate(loginCredentials) {
-    try {
-      const response = await this.post(
-        AUTH_API_ENDPOINT,
-        null,
-        JSON.stringify(loginCredentials),
-        CLIENT_BASIC_AUTH_HEADER,
-        false,
-      );
-      if (response.error) return response;
-      const { accessToken, refreshToken, user } = response;
-      if (!accessToken || !refreshToken || !user) {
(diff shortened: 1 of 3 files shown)


packages/meditrak-app/app/api/TupaiaApi.js
```diff
@@ -230,59 +222,67 @@ export class TupaiaApi {
...
+      reject(new Error('Network request timed out'));
     }, TIMEOUT_INTERVAL);
+    cleanup = () => {
+      clearTimeout(id);
+      resolve();
+    };
   });
+  return { promise, cleanup };
+};
+export const fetchWithTimeout = async (url, config) => {
```
kael89, line 280: This function seems identical to the one in `@tupaia/utils`... I guess we have reasons for not including that package in `meditrak-app`?

kael89 (approved):
A valuable change, looks good!

packages/meditrak-app/app/api/TupaiaApi.js
edmofro: Yes unfortunately it breaks things, because it imports all this node stuff that's not supported in the react-native environment

edmofro:
Merged straight into dev as this will be "regression tested by default" over the next week

PR #2403, opened 2021-04-15
2502 lesmis map overlays part 2

tcaiger:
### Issue #: [2502](https://github.com/beyondessential/tupaia-backlog/issues/2502) LESMIS Map Overlays

This PR is for adding the map and map overlays to lesmis frontend.

### Changes:

- Add `useMapOverlayReportData` to fetch and parse map overlay report data
- Update mapView to use map components

---


packages/lesmis/src/views/MapView.js
```diff
@@ -17,27 +33,137 @@ const Container = styled.div`
...
+      )}
+      <Main>
+        <Map
+          bounds={entityData ? entityData.bounds : null}
+          onLocationChange={handleLocationChange}
+          dragging
+        >
+          <TileLayer tileSetUrl={activeTileSet.url} />
+          {isCountry ? (
+            <PolygonLayer entities={provicesData} Polygon={EntityPolygonLink} />
```
tcaiger, line 143: If it's the country entity, then display the clickable regions rather than all the overlays because the amount of overlays at laos country level (thousands) jams rendering. This imitates behaviour on tupaia.org

packages/lesmis/src/views/MapView.js
```diff
@@ -17,27 +33,137 @@ const Container = styled.div`
...
+
+  const { data: overlayReportData } = useMapOverlayReportData({
+    entityCode,
+    mapOverlayCode: selectedOverlay,
+    hiddenMeasures,
+  });
+
   const { data: overlaysData, isLoading } = useMapOverlaysData({ entityCode });
 
+  const { data: provicesData } = useProvinceData();
```
edmofro, line 100:
```suggestion
  const { data: provincesData } = useProvinceData();
```

edmofro: And maybe also 
```suggestion
  const { data: provincesData } = useProvincesData();
```
so they have consistent pluralisation

packages/lesmis/src/views/MapView.js
```diff
@@ -17,27 +33,137 @@ const Container = styled.div`
...
+  z-index: 9999;
+  pointer-events: none;
+`;
+
+const useProvinceData = () => {
+  const query = useEntitiesData();
+  let provinces = [];
+
+  if (query.data) {
+    provinces = query.data.filter(entity => entity.parentCode === 'LA');
```
edmofro, line 72:
```suggestion
    provinces = query.data.filter(entity => entity.type === 'district');
```
Weirdly, there are some facilities that have Laos as their parent

Screen Shot 2021-04-16 at 2 10 40 PM

edmofro: I guess it's a bit strange to use `district` as the comparison value when we're looking for provinces, so maybe a variable called `BACKEND_PROVINCE_TYPE` or something to make sure people don't think it's a bug/typo

packages/lesmis/src/views/MapView.js
edmofro: Also, considering we're not using the loading state at all, it could even be 
```suggestion
  const provinces = useProvinces();
```
which would also make the hook itself simpler.

On that note - _should_ we be using the loading state of any of these other queries?

packages/lesmis/src/views/MapView.js
```diff
@@ -17,27 +33,137 @@ const Container = styled.div`
...
   const { entityCode } = useUrlParams();
+  const [activeTileSetKey, setActiveTileSetKey] = useState(TILE_SETS[0].key);
+  const [hiddenMeasures, setHiddenMeasures] = useState([]);
   const [selectedOverlay, setSelectedOverlay] = useState(null);
 
+  useEffect(() => {
+    setHiddenMeasures([]);
+  }, [setHiddenMeasures, selectedOverlay]);
+
+  const { data: entityData } = useEntityData({
```
edmofro, line 88: It might make the component a little simpler to pull these four hooks out into one `useMapViewData` that has a combined loading state along with each of the pieces of data as keys in the return object

packages/lesmis/src/views/MapView.js
```diff
@@ -17,27 +33,137 @@ const Container = styled.div`
...
+    mapOverlayCode: selectedOverlay,
+    hiddenMeasures,
+  });
+
   const { data: overlaysData, isLoading } = useMapOverlaysData({ entityCode });
 
+  const { data: provicesData } = useProvinceData();
+
+  const handleLocationChange = (map, bounds) => {
+    const mapBoxBounds = L.latLngBounds(bounds);
```
edmofro, line 103:
```suggestion
    const mapBoxBounds = Leaflet.latLngBounds(bounds);
```
I had to check what `L` referred to - wouldn't have to with the full word

packages/lesmis/src/views/MapView.js
```diff
@@ -17,27 +33,137 @@ const Container = styled.div`
...
+        </Map>
+        <TilePicker
+          tileSets={TILE_SETS}
+          activeTileSet={activeTileSet}
+          onChange={setActiveTileSetKey}
+        />
+        {!isCountry && (
+          <Legend
+            measureOptions={overlayReportData?.measureOptions}
+            setHiddenMeasures={handleSetHiddenMeasures}
```
edmofro: I've stopped insisting that we should use `useCallback` absolutely everywhere, but I think this function gets passed down through a long enough chain of components that forcing them to re-render each render cycle of this parent component isn't worth it. Let's convert `handleSetHiddenMeasures` to a `useCallback` plz

packages/lesmis/src/views/MapView.js
```diff
@@ -17,27 +33,137 @@ const Container = styled.div`
...
+          isLoading={isLoading}
+          overlays={overlaysData}
+          selectedOverlay={selectedOverlay}
+          setSelectedOverlay={setSelectedOverlay}
+        />
+      )}
+      <Main>
+        <Map
+          bounds={entityData ? entityData.bounds : null}
+          onLocationChange={handleLocationChange}
```
edmofro, line 134: And probably `handleLocationChange` should be wrapped in a `useCallback` too

packages/lesmis/src/api/queries/useMapOverlayReportData.js
```diff
@@ -0,0 +1,129 @@
...
+      return {
+        ...entity,
+        ...measure,
+        coordinates: entity.point,
+        region: entity.region,
+        color: displayInfo.color,
+        icon: displayInfo.icon,
+        originalValue: displayInfo.originalValue,
+      };
+    });
```
edmofro: By creating an `entitiesByCode` structure, we can cut out three inner loops. Inner loops are always good things to avoid for performance reasons, but particularly when we're dealing with an array that's 10s of thousands of entities long.

I expect the same structure might be handy other places, so I recommend we create it in `useEntitiesData` so that it can return `{ entities, entityByCode, isLoading }`

```suggestion
    entitiesByCode,
  measureOptions,
  mapOverlayCode,
  hiddenValues,
}) {
  return measureData
    .filter(measure => !hiddenValues.includes(measure[mapOverlayCode]))
    .map(measure => {
      const displayInfo = getMeasureDisplayInfo(measure, measureOptions);
      const entity = entitiesByCode[measure.organisationUnitCode];

      return {
        ...entity,
        ...measure,
        coordinates: entity.point,
        region: entity.region,
        color: displayInfo.color,
        icon: displayInfo.icon,
        originalValue: displayInfo.originalValue,
      };
    });
```

edmofro (changes requested):
I've got a bit hard on this one - let me know if you want to talk about any of it

One big question...does this deal with multi-measure map overlays? I wasn't able to see where, but it is pretty complex stuff so maybe I'm just not following

packages/lesmis/src/views/MapView.js
tcaiger: In the end I didn't pull all the hooks together but I did reduce it down to just 2 - one for the overlay selector and one for the overlays

edmofro (changes requested):

packages/lesmis/src/api/queries/useEntitiesData.js
```diff
@@ -3,15 +3,31 @@
...
+    () => get(`entities/${entityCode}`, { params }),
+    {
+      staleTime: 1000 * 60 * 60 * 1,
+      refetchOnWindowFocus: false,
+      retry: 1,
+    },
+  );
+
+  const entitiesByCode = keyBy(query.data, 'code');
+  return { ...query, entitiesByCode };
```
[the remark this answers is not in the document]
tcaiger: Unfortunately we can't use `entitesByCode` for pre-populating data for useEntityData as entitesByCode data is not actually part of the cached data

packages/lesmis/src/api/queries/useMapOverlayReportData.js
```diff
@@ -0,0 +1,181 @@
...
+    }, {});
+
+    setHiddenValues(hiddenByDefault);
+  }, [setHiddenValues, measureData]);
+
+  const handleSetHiddenValues = useCallback(
+    (key, value, hidden) => {
+      setHiddenValues(state => ({
+        ...state,
+        [value]: hidden,
```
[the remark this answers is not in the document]
tcaiger: I ended up going for the same data structure as tupaia.org which is and object for each measure series.

edmofro (approved):

edmofro:
Looking great! Ready for test!!!

PR #2410, opened 2021-04-16
2678: Analytics table refresh is broken after running migrations if Tupaia User has Super User privileges

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2678:

Added logic to prevent installing the notificationt trigger on the analytics and log tables

### DIFF
--- a/packages/database/src/analytics/AnalyticsRefresher.js
+++ b/packages/database/src/analytics/AnalyticsRefresher.js
@@ -3,6 +3,8 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
+import winston from 'winston';
+
 const REFRESH_DEBOUNCE_TIME = 1000; // wait 1 second after changes before refreshing, to avoid double-up
 
 export class AnalyticsRefresher {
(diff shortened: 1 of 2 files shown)


packages/database/src/analytics/AnalyticsRefresher.js
```diff
@@ -57,6 +59,13 @@ export class AnalyticsRefresher {
   };
 
   static async executeRefresh(database) {
-    await database.executeSql(`SELECT mv$refreshMaterializedView('analytics', 'public', true);`);
+    try {
+      const start = Date.now();
+      await database.executeSql(`SELECT mv$refreshMaterializedView('analytics', 'public', true);`);
+      const end = Date.now();
+      winston.info(`Analytics table refresh took: ${end - start}ms`);
```
rohan-bes, line 66: This could potentially end up rather noisy, however it may also help us keep track analytics refreshes and how long they are ending up taking. I'm leaning towards keeping it but happy to hear other opinions

packages/database/src/runPostMigration.js
```diff
@@ -35,6 +36,7 @@ const getSelectForTablesWithoutTriggers = triggerSuffix => `
     AND privileges.privilege_type = 'TRIGGER'
     AND t.table_schema = 'public'
     AND t.table_type != 'VIEW'
+    AND t.table_name NOT LIKE 'log$_%'
```
rohan-bes, line 39: Ignore all log tables with this

edmofro (approved):

PR #2411, opened 2021-04-16
1337: Site level derived data in Total Cases indicators for PSSS

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1337:

Added support for site level derived data in Total Cases indicators for PSSS

### DIFF
--- a/packages/database/src/migrations/20210415110921-UpdatePSSSWeeklyIndicatorsSupportSiteData-modifies-data.js
+++ b/packages/database/src/migrations/20210415110921-UpdatePSSSWeeklyIndicatorsSupportSiteData-modifies-data.js
@@ -1,0 +1,122 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const oldConfig = syndromeCode => ({
+  formula: `firstExistingValue(PSSS_${syndromeCode}_Cases, PSSS_${syndromeCode}_Daily_Cases)`,
+  aggregation: {
+    [`PSSS_${syndromeCode}_Cases`]: 'FINAL_EACH_WEEK',
+    [`PSSS_${syndromeCode}_Daily_Cases`]: ['SUM_EACH_WEEK', 'FINAL_EACH_WEEK'],
+  },
+  defaultValues: {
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210415110921-UpdatePSSSWeeklyIndicatorsSupportSiteData-modifies-data.js
```diff
@@ -0,0 +1,122 @@
...
+    code: 'sumSiteWeeklyCases',
+    config: {
+      formula: `PSSS_${syndromeCode}_Cases`,
+      aggregation: [
+        'FINAL_EACH_WEEK',
+        {
+          type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+          config: {
+            dataSourceEntityType: 'facility',
+            aggregationEntityType: 'country',
```
rohan-bes, line 40: One might argue that we ought to use `requested` here, but given that we currently only care about aggregating to the country level I'd rather be explicit and we can change it later if needs be

packages/database/src/migrations/20210415110921-UpdatePSSSWeeklyIndicatorsSupportSiteData-modifies-data.js
```diff
@@ -0,0 +1,122 @@
...
+    [`PSSS_${syndromeCode}_Daily_Cases`]: ['SUM_EACH_WEEK', 'FINAL_EACH_WEEK'],
+    sumSiteWeeklyCases: 'RAW',
+    sumSiteDailyCases: 'RAW',
+  },
+  defaultValues: {
+    [`PSSS_${syndromeCode}_Cases`]: 'undefined',
+    [`PSSS_${syndromeCode}_Daily_Cases`]: 'undefined',
+    sumSiteWeeklyCases: 'undefined',
+    sumSiteDailyCases: 'undefined',
+  },
```
kael89, line 81: We should add support for a single string (expanded for all formula variables) for `defaultValues` at some point... We already support this for `aggregation`

kael89 (approved):
Nice work! So happy it worked smoothly out of the box!! 😌

packages/database/src/migrations/20210415110921-UpdatePSSSWeeklyIndicatorsSupportSiteData-modifies-data.js
rohan-bes: Or even a default default value?

kael89: Very meta, haha! We probably don't want to use `default` values by default - we have to use it here to overcome the "all defined or not calculate" logic in indicators

PR #2418, opened 2021-04-19
Refresh analytics during runPostMigrations

rohan-bes:


packages/database/src/runPostMigration.js
```diff
@@ -86,6 +86,12 @@ export const runPostMigration = async driver => {
     ),
   );
 
+  // Refresh analytics in case they've been impacted by migrations
+  const start = Date.now();
+  await driver.runSql(`SELECT mv$refreshMaterializedView('analytics', 'public', true);`);
```
rohan-bes, line 91: I wanted to use `AnalyticsRefresher`'s static invokation to do this, however it relies on a `TupaiaDatabase` object and the `driver` object we have here doesn't support the `executeSql` function. I didn't really wanna change things in the `AnalyticsRefresher` so we have a small amount of code duplication here instead.

rohan-bes:
@kael89 I believe this should fix the issue you're seeing on your strive branch, but I'm testing locally in any case

rohan-bes:
> @kael89 I believe this should fix the issue you're seeing on your strive branch, but I'm testing locally in any case

Tested this locally and it works (takes a whole minute to refresh the analytics tho, I believe this is because of just how many answers are being added)

kael89 (approved):
Nice fix!

PR #2421, opened 2021-04-19
1411: Handled case where entity aggregation config returned no relations

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1411:

The decision with this ticket is just to support fetching the PSSS Weekly Report and PSSS Weekly Cases when passing in the facility codes as orgUnitCodes. In order to support this a small change needs to be made to the entity aggregation logic to support aggregating to the requested entity level.

### DIFF
--- a/packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
+++ b/packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
@@ -72,8 +72,7 @@ const getAdjustedOrganisationUnitsAndAggregations = async (
           dataSourceEntityType,
           dataSourceEntityFilter,
         );
-        adjustedOrganisationUnitCodes =
-          dataSourceEntities.length > 0 ? dataSourceEntities : organisationUnitCodes;
+        adjustedOrganisationUnitCodes = dataSourceEntities;
         adjustedAggregations.push(aggregation);
         return;
       }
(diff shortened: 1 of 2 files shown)


packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -93,12 +95,15 @@ const getAdjustedOrganisationUnitsAndAggregations = async (
...
       if (!adjustedOrganisationUnitCodes) {
-        adjustedOrganisationUnitCodes = dataSourceEntities;
+        adjustedOrganisationUnitCodes =
+          dataSourceEntities.length > 0 ? dataSourceEntities : organisationUnitCodes;
+      }
+      if (Object.keys(relations).length > 0) {
+        adjustedAggregations.push({
+          ...aggregation,
+          config: { ...aggregation.config, orgUnitMap: relations },
+        });
```
rohan-bes, line 105: I think this is a safe assumption to make. If there are no relations, we don't need to include this aggregation as each analytic would just map to itself.

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -66,12 +66,13 @@ const getAdjustedOrganisationUnitsAndAggregations = async (
 
       if (!adjustedOrganisationUnitCodes && !shouldFetchRelations(aggregation)) {
         const { dataSourceEntityType, dataSourceEntityFilter } = aggregation.config;
-        adjustedOrganisationUnitCodes = await entityConnection.getDataSourceEntities(
+        const dataSourceEntities = await entityConnection.getDataSourceEntities(
```
kael89, line 69: The changes in this file can probably be completely reverted

kael89 (approved):
Clean solution!

PR #2430, opened 2021-04-20
fix broken migration

billli0:


billli0:
Same as the previous fix.

nick-beyondessential (approved):

PR #2431, opened 2021-04-20
1336: Support deriving PSSS Total Sites indicator from number of reporting sites

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1336:

Added 2 new indicators to determine if a site has completed a Weekly Survey or a Daily Survey for a given week

Updated Total Cases indicator to preferentially use: 
- National recorded answer for Sites Reported
- Number of Sites that submitted a weekly survey
- Number of sites that completed a daily survey

Note: After adding these indicators the performance of PSSS has dropped substantially! (~15 seconds to load the Weekly Report in my local). Subsequent work must be done to improve performance but that's out of the scope for this PR.

### DIFF
--- a/packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
+++ b/packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
@@ -116,7 +116,9 @@ const getAdjustedOrganisationUnitsAndAggregations = async (
 export const adjustOptionsToAggregationList = async (context, fetchOptions, aggregationOptions) => {
   const { aggregations: aggregationList = [] } = aggregationOptions;
   if (aggregationList.length === 0) {
-    return [fetchOptions, aggregationOptions];
+    // Trim off any pre-existing aggregations from fetch options
+    const { aggregations, ...restOfFetchOptions } = fetchOptions;
+    return [restOfFetchOptions, aggregationOptions];
   }
 
   const { startDate, endDate } = getAdjustedDateRange(fetchOptions, aggregationList);
--- a/packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
+++ b/packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
@@ -22,10 +22,17 @@ const oldConfig = {
 };
 
 const newConfig = {
-  formula: `firstExistingValue(PSSS_Sites_Reported, PSSS_Site_Reported)`,
+  formula: `firstExistingValue(PSSS_Sites_Reported, PSSS_Site_Weekly_Survey_Completed, PSSS_Site_Daily_Survey_Completed)`,
   aggregation: {
     PSSS_Sites_Reported: 'FINAL_EACH_WEEK',
-    PSSS_Site_Reported: [
+    PSSS_Site_Weekly_Survey_Completed: [
+      'FINAL_EACH_WEEK',
+      {
+        type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+        config: { dataSourceEntityType: 'facility', aggregationEntityType: 'country' },
+      },
+    ],
+    PSSS_Site_Daily_Survey_Completed: [
       'FINAL_EACH_WEEK',
       {
         type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
@@ -35,16 +42,26 @@ const newConfig = {
   },
   defaultValues: {
     PSSS_Sites_Reported: 'undefined',
-    PSSS_Site_Reported: 'undefined',
+    PSSS_Site_Weekly_Survey_Completed: 'undefined',
+    PSSS_Site_Daily_Survey_Completed: 'undefined',
   },
 };
 
-const PSSS_SITE_REPORTED = {
-  code: 'PSSS_Site_Reported',
+const PSSS_SITE_REPORTED_WEEKLY = {
+  code: 'PSSS_Site_Weekly_Survey_Completed',
   builder: 'eventCheckConditions',
   config: {
     formula: 'true',
     programCode: 'PSSS_WSR',
+  },
+};
+
+const PSSS_SITE_REPORTED_DAILY = {
+  code: 'PSSS_Site_Daily_Survey_Completed',
+  builder: 'eventCheckConditions',
+  config: {
+    formula: 'true',
+    programCode: 'PSSS_DSR',
   },
 };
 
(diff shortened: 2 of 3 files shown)


packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -116,7 +116,9 @@ const getAdjustedOrganisationUnitsAndAggregations = async (
 export const adjustOptionsToAggregationList = async (context, fetchOptions, aggregationOptions) => {
   const { aggregations: aggregationList = [] } = aggregationOptions;
   if (aggregationList.length === 0) {
-    return [fetchOptions, aggregationOptions];
+    // Trim off any pre-existing aggregations from fetch options
+    const { aggregations, ...restOfFetchOptions } = fetchOptions;
+    return [restOfFetchOptions, aggregationOptions];
```
rohan-bes, line 121: Bit of a hacky fix here, the issue that was occurring is because we pass through the fetch options to the IndicatorApi which in turn passes them through to each required dataFetch/parameterFetch. Since we add the aggregation options list earlier in the chain they persist here even if no aggregations are requested.

I think the ideal fix here would be for each consumer of the fetch options to rebuild the options containing only the fields that make sense to them (ie. the IndicatorApi itself would end up dropping the aggregations list from the fetch options).

However I think there's still room for brainstorming on the best way to do that, and this is an acceptable solution imo

packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
```diff
@@ -0,0 +1,114 @@
...
+};
+
+const PSSS_SITE_REPORTED_WEEKLY = {
+  code: 'PSSS_Site_Weekly_Survey_Completed',
+  builder: 'eventCheckConditions',
+  config: {
+    formula: 'true',
+    programCode: 'PSSS_WSR',
+  },
+};
```
rohan-bes, line 57: Technically these indicators could be parameters, however there's currently a bug in indicator parameters that the fetch options don't get mutated as required per the aggregation options. I wasn't entirely sure if I ought to go ahead and make that change, since it more tightly couples the aggregator and indicator logic and seemed pretty messy... I think we can come up with a better design with time.

For now this seems acceptable to me 😄

packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
```diff
@@ -0,0 +1,114 @@
...
+        type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+        config: { dataSourceEntityType: 'facility', aggregationEntityType: 'country' },
+      },
+    ],
+  },
+  defaultValues: {
+    PSSS_Sites_Reported: 'undefined',
+    PSSS_Site_Weekly_Survey_Completed: 'undefined',
+    PSSS_Site_Daily_Survey_Completed: 'undefined',
+  },
```
kael89, line 47: I created [#2696: Indicators - Allow string to be used for defaultValues](https://github.com/beyondessential/tupaia-backlog/issues/2696) to make `defaultValues` definition in cases like this simpler

packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
kael89: Good and pragmatic approach, I agree!

packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
```diff
@@ -0,0 +1,114 @@
...
+    PSSS_Sites_Reported: 'FINAL_EACH_WEEK',
+    PSSS_Site_Weekly_Survey_Completed: [
+      'FINAL_EACH_WEEK',
+      {
+        type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+        config: { dataSourceEntityType: 'facility', aggregationEntityType: 'country' },
+      },
+    ],
+    PSSS_Site_Daily_Survey_Completed: [
+      'FINAL_EACH_WEEK',
```
kael89, line 36: Should this be `FINAL_EACH_DAY`? (sum data for all days in a week)

packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
```diff
@@ -0,0 +1,114 @@
...
+  config: {
+    formula: 'true',
+    programCode: 'PSSS_DSR',
+  },
+};
+
+const insertIndicator = async (db, indicator) => {
+  const { code } = indicator;
+
+  await insertObject(db, 'data_source', {
```
kael89, line 71: Did you know that `meditrak-server` listens to indicator changes and updates data sources accordingly?? I added this logic a while back but I forgot! See the [Indicator model](https://github.com/beyondessential/tupaia/blob/dev/packages/database/src/modelClasses/Indicator.js#L34)

I guess it's fine in practice since the data source will be created first, and then the trigger will probably just fail because of a `code/type` constraint

kael89 (approved):
It's very satisfying to see that indicators can be used in so many different ways... And everything relies on a couple of fundamental indicator builders!

About performance: more specific indicator builders may have been able to deal with performance concerns in a simple way, but as a longer-term strategy we should try our best to optimise the fundamental builders first. Looking forward to an optimisation discussion!

Approved with questions

packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
rohan-bes: Oh wow! I didn't know! I still think explicitly creating them in migrations might be the way to go cos often when running locally you don't have meditrak-server up

packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
rohan-bes: Nope, since this aggregation will just select a single analytic per org unit per week which is what we want (multiple daily submissions from the same site in on week should only count as 1)

kael89: Oh true, since this is for counting "how many submitted in a week" rather than summing any results... Thanks!

packages/database/src/migrations/20210419225536-UpdatePSSSTotalSitesReportToUseSiteData-modifies-data.js
kael89: True!

PR #2434, opened 2021-04-21
2538 lesmis dashboard reports

tcaiger:
### Issue #: [2538](https://github.com/beyondessential/tupaia-backlog/issues/2538) lesmis dashboard reports

This PR is for the changes to ui-components charts to make the api simpler and to update the styles to match the lesmis designs. This is a pre-requisite for adding charts to lesmis front end.

### Changes:

- update chart api to just use viewContent
- remove data fetching from storybook
- make new styled legends
- make new styled tooltips
- style chart axes
- setup storybook to handle themes
- add new chart stories for with light and dark themes for all the chart types

---


packages/ui-components/src/components/Chart/Legend.js
```diff
@@ -4,36 +4,74 @@
...
+
+const Text = styled.span`
+  line-height: 1.4;
+`;
+
+const getDisplayValue = (chartConfig, value) => {
+  if (value in chartConfig && 'label' in chartConfig[value]) {
+    return chartConfig[value].label;
+  }
+  return value;
```
edmofro, line 46:
```
  return chartConfig[value]?.label || value;
```

or, if you like having the obvious difference between conditions

```
  if (chartConfig[value]?.label) {
    return chartConfig[value].label;
  }
  return value;
```

packages/ui-components/src/components/Chart/Legend.js
```diff
@@ -4,36 +4,74 @@
...
 };
+
+export const getPieLegend = () => ({ payload }) => (
+  <LegendContainer style={{ padding: 0, marginBottom: -10 }}>
+    {payload.map(({ color, value }) => (
+      <LegendItem key={value} disabled>
+        <Box style={{ background: color }} />
+        <Text>{value}</Text>
```
edmofro, line 54: We don't want `getDisplayValue` here?

packages/ui-components/src/components/Chart/PieChart.js
```diff
@@ -26,136 +26,70 @@
...
-    </g>
-  );
+const chartColorAtIndex = (colorArray, index) => {
+  return colorArray[index % colorArray.length];
 };
 
 export const PieChart = ({ viewContent, isExporting, isEnlarged, onItemClick }) => {
   const [activeIndex, setActiveIndex] = useState(-1);
 
   // Disable tapping charts to reveal legend labels on mobile as they do not fit and are awkwardly cropped
```
edmofro, line 91: Considering we've stripped out all the other support for mobile and tupaiay stuff, as well as changing the hover behaviour, let's remove the mobile check here too

edmofro (changes requested):
Really good call to pull out the data fetching IMO. 

The stories are super clean with the only difference being the view content!

I am a little nervous about losing some of the peculiarities Tupaia has built up over the years, but it does make things clean. Just means we'll have to put in extra work and be more careful when integrating back into Tupaia down the track, but who knows when that might happen and at that point it might be good to review what Tupaia still wants (in terms of legends, mobile support/responsiveness, etc.)

packages/ui-components/src/components/Chart/Legend.js
tcaiger: We don't support this on tupaia yet but might as well since we are here.

packages/ui-components/src/components/Chart/PieChart.js
tcaiger: Yeah it's a fine balance of refactoring and updating stuff while not loosing any key functionality from tupaia. I may over step the line sometimes so it's good to ask the question. I'm keen to phase out the isMobile stuff in favour of a responsive approach but I've tried to keep it contained to just the interactions on the chart and the legend here. So yes good suggestion but I'm going to leave in the other mobile related bits of these components for now.

packages/ui-components/src/components/Chart/CartesianChart.js
```diff
@@ -147,7 +145,8 @@ export const CartesianChart = ({ viewContent, isEnlarged, isExporting }) => {
   } = viewContent;
 
   const hasDataSeries = chartConfig && Object.keys(chartConfig).length > 1;
-  const aspect = !isEnlarged && !isMobile() && !isExporting ? 1.6 : undefined;
+  // const aspect = !isEnlarged && !isMobile() && !isExporting ? 1.6 : undefined;
+  const aspect = undefined;
```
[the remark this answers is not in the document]
tcaiger: Good catch!

edmofro (approved):
Approving based on the changes added in part-2

PR #2435, opened 2021-04-21
Fix tableOfEvents

kael89:


packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfEvents.js
```diff
@@ -19,14 +19,15 @@ import {
 const DATE_FORMAT = 'DD-MM-YYYY';
 const TOTAL_KEY = 'Total';
 
-// TODO event.dataValues can be (array|object)
```
edmofro, line 22: What happened to the idea that it might be an array?

kael89: [tableOfEVents](https://github.com/beyondessential/tupaia/blob/dev/packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfEvents.js#L81)

It can theoretically be an array, but it's not in this data builder. I added the code out of memory with a note to test it, but forgot

edmofro: Ok cool! Happy for this to go straight to dev, but could you give the testers a full list of reports that use this builder, so that they get regression tested during this week's regression window?

edmofro (approved):

packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfEvents.js
kael89: Sure, I will let them know

PR #2437, opened 2021-04-21
Entity vitals

EMcQ-BES:
### Changes:

- Add reports for lesmis


packages/database/src/migrations/20210412210749-AddLesmisEntityVitalsReports-modifies-data.js
```diff
@@ -0,0 +1,216 @@
...
+      },
+      transform: [
+        'keyValueByDataElementName',
+        'mostRecentValuePerOrgUnit',
+        {
+          transform: 'select',
+          '...': ['organisationUnit'],
+          "'Photo'": '$row.SchFDpho',
+          "'ContactNumber'": '$row.SchCVD026b',
+          "'NumberOfStudents'":
```
edmofro, line 74: I reckon this should be an indicator

packages/database/src/migrations/20210412210749-AddLesmisEntityVitalsReports-modifies-data.js
```diff
@@ -0,0 +1,216 @@
...
+          'SchCVD026b',
+          'SchFDpho',
+        ],
+      },
+      transform: [
+        'keyValueByDataElementName',
+        'mostRecentValuePerOrgUnit',
+        {
+          transform: 'select',
+          '...': ['organisationUnit'],
```
edmofro, line 71: Remind me what this does?

edmofro (changes requested):
lgtm other than making the sum of students an indicator

packages/database/src/migrations/20210412210749-AddLesmisEntityVitalsReports-modifies-data.js
EMcQ-BES: This just includes the passed in org unit in the results of the report, probably not entirely necessary for our usage

edmofro: Aha :+1:

edmofro (approved):
All looking good to me now! So long as you've tested those suggested changes I think this is good to go :-)

EMcQ-BES:
Yep 👍  Ran a quick test in postman requesting one of the school vitals reports

image

PR #2439, opened 2021-04-21
Add partner logos to entity vitals view

EMcQ-BES:
### Changes:

- Add image files for partner logos
- Add function to generate logos from vitals data


packages/lesmis/src/components/getPartnerSupportItems.js
```diff
@@ -0,0 +1,44 @@
...
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+const StyledLogo = styled.div`
+  width: 55px;
+  height: 55px;
+  background-repeat: no-repeat;
+  background-position: center center;
+  background-size: contain;
+  border: 1px solid #dedee0;
```
tcaiger, line 16:
```suggestion
border: ${props => props.theme.palette.grey['400']};
```
Sorry I forgot to explain this but there is a material-ui theme set up for the project and we try to use it when it makes sense. Those times are really just when using brand colors and a couple of very commonly used colors such as this border one.

packages/lesmis/src/components/getPartnerSupportItems.js
```diff
@@ -0,0 +1,44 @@
...
+  WV: 'WV.png',
+};
+
+const Logo = image => (
+  <StyledLogo style={{ backgroundImage: `url('/images/partnerLogos/${image}')` }} />
+);
+
+export const getPartnerSupportItems = vitals => {
+  return Object.keys(LOGO_MAP).map(partner => (vitals[partner] ? Logo(LOGO_MAP[partner]) : null));
+};
```
tcaiger, line 44: I would prefer to make this file a react component for the PartnerLogo itsself that just takes in a code and returns the correct Logo. eg. `<PartnerLogo code={vital.code} />`. I think that would be more flexible and re-usable. Then we could move the map into the DashboardView and iterate from the vitals as opposed to the logos.

packages/lesmis/src/views/DashboardView.js
```diff
@@ -9,7 +9,7 @@ import MuiContainer from '@material-ui/core/Container';
 import Typography from '@material-ui/core/Typography';
 import { useUrlParams } from '../utils';
 import { useVitalsData } from '../api';
-import { EntityVitalsItem } from '../components';
+import { EntityVitalsItem, getPartnerSupportItems } from '../components';
```
tcaiger, line 12: Since we are going to change this import to "PartnerLogo" this comment is just for future reference. It is conventional in react to always return a react component as opposed to a function from a component file.  If you want to return a component which has a list of components it would looks something like: `<PartnerSupportList vitals={vitals} />`

tcaiger (changes requested):
Looking good. I've just requested the component structure be tweaked a bit. Let me know if you have any questions.

packages/lesmis/src/views/DashboardView.js
```diff
@@ -59,7 +59,14 @@ export const DashboardView = () => {
...
             <FlexRow>
-              <Typography>Development Partner Support</Typography>
+              <TitleContainer>
+                <Typography>Development Partner Support</Typography>
+              </TitleContainer>
+              {Object.entries(vitals)
+                .filter(([, value]) => value === true)
+                .map(([key]) => (
+                  <PartnerLogo code={key} />
+                ))}
```
EMcQ-BES, line 69: I want to make this smarter by having vitals return a `vitals.DevelopmentPartners` object, but that involves adjusting the report so I'd prefer to do that after I consolidate my branches

packages/lesmis/src/components/PartnerLogo.js
```diff
@@ -0,0 +1,46 @@
+/*
```
tcaiger, line 1: Nice - this is looking really clean 👍

tcaiger (approved):
Looks good

PR #2441, opened 2021-04-21
2538 lesmis dashboard reports part 2

tcaiger:
### Issue #: [2538](https://github.com/beyondessential/tupaia-backlog/issues/2541) Entity Dashboard Groups

This PR is for adding the dashboard groups to lesmis front end

### Changes:

- fetch dashboard data
- add dashboard group tabs

---

### DIFF
--- a/packages/lesmis/src/components/Tabs.js
+++ b/packages/lesmis/src/components/Tabs.js
@@ -97,8 +97,8 @@ const PanelComponent = styled.div`
   margin-bottom: 2rem;
 `;
 
-export const TabPanel = React.memo(({ children, value, dataKey, Panel }) => {
-  if (value !== dataKey) {
+export const TabPanel = React.memo(({ children, isSelected, Panel }) => {
+  if (!isSelected) {
     return null;
   }
   return <Panel>{children}</Panel>;
@@ -106,11 +106,11 @@ export const TabPanel = React.memo(({ children, value, dataKey, Panel }) => {
 
 TabPanel.propTypes = {
   children: PropTypes.node.isRequired,
-  value: PropTypes.string.isRequired,
-  dataKey: PropTypes.string.isRequired,
+  isSelected: PropTypes.bool,
   Panel: PropTypes.any,
 };
 
 TabPanel.defaultProps = {
   Panel: PanelComponent,
+  isSelected: false,
 };
--- a/packages/lesmis/src/components/Tabs.js
+++ b/packages/lesmis/src/components/Tabs.js
@@ -1,0 +1,116 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ *
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import MuiTabs from '@material-ui/core/Tabs';
+import Skeleton from '@material-ui/lab/Skeleton';
+import MuiTab from '@material-ui/core/Tab';
+import * as COLORS from '../constants';
+import { FlexStart } from './Layout';
+
+export const TabBar = styled(FlexStart)`
+  min-height: 5.5rem;
+  border-top: 1px solid ${props => props.theme.palette.grey['400']};
+  border-bottom: 1px solid ${props => props.theme.palette.grey['400']};
+`;
+
+const LoadingTab = () => <Skeleton width={100} height={20} style={{ marginLeft: 30 }} />;
+
+export const TabBarLoader = () => (
+  <>
+    <LoadingTab />
+    <LoadingTab />
+    <LoadingTab />
+    <LoadingTab />
+  </>
+);
+
+export const TabBarTabs = styled(MuiTabs)`
+  .MuiTabs-indicator {
+    height: 5px;
+    background-color: ${props => props.theme.palette.primary.main};
+  }
+`;
+
+export const TabBarTab = styled(MuiTab)`
+  min-width: auto;
+  padding: 2rem 0;
+  margin: 0 1rem;
+  letter-spacing: 0;
+  color: ${props => props.theme.palette.text.secondary};
+  opacity: 1;
+
+  &.Mui-selected {
+    color: ${props => props.theme.palette.primary.main};
+  }
+`;
+
+export const Tabs = styled(MuiTabs)`
+  overflow: visible;
+  background: ${COLORS.GREY_F9};
+
+  .MuiTabs-scroller {
+    overflow: visible !important; // overwrite the mui element style
+  }
+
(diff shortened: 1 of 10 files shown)


packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -0,0 +1,92 @@
...
+      <DashboardSection>
+        <FetchLoader isLoading={isLoading} isError={isError} error={error}>
+          {data &&
+            Object.entries(data).map(([heading, dashboardGroup]) => (
+              <TabPanel key={heading} dataKey={heading} value={selectedDashboard}>
+                {Object.entries(dashboardGroup).map(([groupName, groupValue]) => {
+                  const views = groupValue.views.filter(report => report.type === 'chart');
+                  return views.length > 0 ? (
+                    views.map(report => (
+                    // Todo: add report component
```
tcaiger, line 70: Other PR coming soon.

packages/lesmis/src/views/DashboardView.js
```diff
@@ -3,21 +3,86 @@
...
-  <TemplateBody>
-    <MuiContainer>
-      <Typography>Dashboard View</Typography>
-    </MuiContainer>
-  </TemplateBody>
+const YEARS = ['2015', '2016', '2017', '2018', '2019', '2020', '2021'];
+
+const EarlyChildhoodTab = () => (
+  <MuiBox p={5} minHeight={500}>
+    ESSDP Early Childhood Education Tab
```
tcaiger, line 28: Placeholder content.

packages/lesmis/src/views/DashboardView.js
```diff
@@ -3,21 +3,86 @@
...
+        ))}
+      </Tabs>
+      {TABS.map(({ key, Component }) => (
+        <TabPanel key={key} dataKey={key} value={selectedTab} Panel={React.Fragment}>
+          <Component
+            entityCode={entityCode}
+            selectedDashboard={selectedDashboard}
+            setSelectedDashboard={setSelectedDashboard}
+            selectedYear={selectedYear}
+            setSelectedYear={setSelectedYear}
```
edmofro, line 82: Would suggest the component itself maintain the state of selected dashboard group...and possibly also selected year, though I'm less sure on that just because of where the year selector sits

packages/lesmis/src/components/Tabs.js
```diff
@@ -0,0 +1,116 @@
...
+const PanelComponent = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 2rem;
+  margin-bottom: 2rem;
+`;
+
+export const TabPanel = React.memo(({ children, value, dataKey, Panel }) => {
+  if (value !== dataKey) {
```
edmofro, line 101: Why use a separate `dataKey` rather than `key`?
```suggestion
export const TabPanel = React.memo(({ children, value, key, Panel }) => {
  if (value !== key) {
```

packages/lesmis/src/views/DashboardView.js
```diff
@@ -3,21 +3,86 @@
...
+          <Typography>Dashboard View</Typography>
+        </MuiContainer>
+      </TemplateBody>
+      <Tabs value={selectedTab} onChange={handleChangeTab}>
+        {TABS.map(({ getLabel, key }) => (
+          <Tab key={key} label={getLabel(entityData?.type)} value={key} />
+        ))}
+      </Tabs>
+      {TABS.map(({ key, Component }) => (
+        <TabPanel key={key} dataKey={key} value={selectedTab} Panel={React.Fragment}>
```
edmofro, line 76: I feel the "value" of a tab panel would be more related to its content than to which one of its siblings is selected. Possibly a rename to

```suggestion
        <TabPanel key={key} dataKey={key} selectedKey={selectedTab} Panel={React.Fragment}>
```

Or possibly even pass the boolean directly
```suggestion
        <TabPanel key={key} selected={key === selectedTab} Panel={React.Fragment}>
```

packages/lesmis/src/views/DashboardView.js
edmofro: Or is this done here so that we don't have to manage the url state too many places?

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -0,0 +1,92 @@
...
+          </>
+        )}
+      </TabBar>
+      <DashboardSection>
+        <FetchLoader isLoading={isLoading} isError={isError} error={error}>
+          {data &&
+            Object.entries(data).map(([heading, dashboardGroup]) => (
+              <TabPanel key={heading} dataKey={heading} value={selectedDashboard}>
+                {Object.entries(dashboardGroup).map(([groupName, groupValue]) => {
+                  const views = groupValue.views.filter(report => report.type === 'chart');
```
edmofro, line 67:
```suggestion
                  const dashboardReports = groupValue.views.filter(report => report.type === 'chart'); // TODO support other report types
```
Two points there, the rename and the todo comment

edmofro (changes requested):
Another nicely contained PR!

Main point is supporting the dropdown rather than the tabs for the overall dashboard super-group selector

A couple of other questions about where to manage state, renaming the immortal `view` -> `dashboardReport` as far upstream as we can manage, and other misc.

packages/lesmis/src/views/DashboardView.js
tcaiger: Yes that's right the state is lifted up so that the selected year and dashboard group are remembered when clicking between the outer tabs. That seems like better behaviour at this point than resetting to the default. Yeah when we do the deep linking then it may make sense to move that state down or remove it completely.

packages/lesmis/src/components/Tabs.js
tcaiger: As far as I can tell 'key' gets filtered out by react internal code so you can't use it. Anyway I like your suggestion below to just pass in a `isSselected` boolean 👍

packages/ui-components/src/components/Chart/PieChart.js
```diff
@@ -90,15 +90,11 @@ export const PieChart = ({ viewContent, isExporting, isEnlarged, onItemClick })
 
   // Disable tapping charts to reveal legend labels on mobile as they do not fit and are awkwardly cropped
```
edmofro, line 91: I think we can remove this comment now too

edmofro (approved):
One unneeded comment to delete but otherwise good to go 🎉

packages/lesmis/src/components/Tabs.js
edmofro: Oh right, that's a bit annoying!

PR #2447, opened 2021-04-22
2476 update strive k13 map overlay

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/2476

Basically this adds one new map overlay & edits two existing ones


billli0 (approved):
LGTM 👍

PR #2449, opened 2021-04-22
add psss-con-total-cases indicator

billli0:
### Issue #:
[Palau Country Level Dashboard: Weekly Cases Trend Graph #2109](https://github.com/beyondessential/tupaia-backlog/issues/2109)
### Changes:
- Add `Psss-Con-Total-Cases` indicator


rohan-bes:
Hmmm I'm not sure this is the right approach here @billli0. I think it's best to create the indicator in its own migration, and it should be branched off and be merging into `dev`. 

There are already `PSSS_<syndrome>_Total_Cases` indicators in `dev` already so you don't need to be working off the psss-phase-2 branch

billli0:
Thanks @rohan-bes , I thought we need to update the old config. If not I will create this indicator with the old config.

PR #2451, opened 2021-04-22
Move specific API to EntityConnection

kael89:
Addressing https://github.com/beyondessential/tupaia/pull/2436#discussion_r617990322 - let me know what you think!

### DIFF
--- a/packages/psss-server/src/connections/EntityConnection.ts
+++ b/packages/psss-server/src/connections/EntityConnection.ts
@@ -3,7 +3,8 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { PSSS_HIERARCHY } from '../constants';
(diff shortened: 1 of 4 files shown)


packages/psss-server/src/connections/EntityConnection.ts
```diff
@@ -74,15 +75,52 @@ const getRelationParams = (options: RelationOptions) => {
 export class EntityConnection extends ApiConnection {
   baseUrl = ENTITY_API_URL;
 
-  fetchDescendants = async <T extends EntityOptions>(
+  public fetchCountries = async () =>
```
rohan-bes, line 78: I need to remember to use public/private in my classes 🤦 We should really make this a linting rule. Nice work!

rohan-bes (approved):
Looks great, love it!

PR #2452, opened 2021-04-22
2677: Added check for `session` when adjusting aggregation options

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2677:

Fairly simple hack that allows for web-config-server entity aggregation and aggregator entity aggregation logic to co-exist. Assumption is that if constructor of Aggregator doesn't provide a session in the context, then they are responsible for creating their own entity aggregation options.


kael89 (approved):
Cute little hack!

rohan-bes:
@kael89 build is only failing cos I forgot about the branch naming rules. So am gonna merge regardless

PR #2454, opened 2021-04-22
2286 lesmis dashboard year selector

tcaiger:
### Changes:

- Add year selector to dashboard reports

---

### Screenshots:

See Issue

### DIFF
--- a/packages/lesmis/src/api/queries/useDashboardReportData.js
+++ b/packages/lesmis/src/api/queries/useDashboardReportData.js
@@ -4,17 +4,15 @@
  *
  */
 import { useQuery } from 'react-query';
-import { getDefaultDates, formatDateForApi } from '@tupaia/ui-components/lib/chart';
+import { roundStartEndDates, formatDateForApi } from '@tupaia/ui-components/lib/chart';
 import { get } from '../api';
+import { SINGLE_YEAR_GRANULARITY, MIN_DATA_YEAR } from '../../constants';
 
-export const useDashboardReportData = ({
-  entityCode,
-  dashboardGroupId,
-  reportId,
-  periodGranularity,
-  defaultTimePeriod,
-}) => {
-  const { startDate, endDate } = getDefaultDates({ periodGranularity, defaultTimePeriod });
+export const useDashboardReportData = ({ entityCode, dashboardGroupId, year, reportId }) => {
+  const currentYear = new Date().getFullYear().toString();
+  const startYear = !year || year === 'all' ? MIN_DATA_YEAR : year;
+  const endYear = !year || year === 'all' ? currentYear : year;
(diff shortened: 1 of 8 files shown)


packages/lesmis/src/api/queries/useDashboardReportData.js
```diff
@@ -4,17 +4,15 @@
...
-  reportId,
-  periodGranularity,
-  defaultTimePeriod,
-}) => {
-  const { startDate, endDate } = getDefaultDates({ periodGranularity, defaultTimePeriod });
+export const useDashboardReportData = ({ entityCode, dashboardGroupId, year, reportId }) => {
+  const currentYear = new Date().getFullYear().toString();
+  const startYear = !year || year === 'all' ? MIN_DATA_YEAR : year;
+  const endYear = !year || year === 'all' ? currentYear : year;
+  const { startDate, endDate } = roundStartEndDates(SINGLE_YEAR_GRANULARITY, startYear, endYear);
```
tcaiger, line 15: This is a suggestion to have an all data option in the year selector. Happy to remove it if it doesn't make sense. It could be useful for business acceptance testing if nothing else.

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -19,26 +19,62 @@ import {
...
+
+const getYearOptions = () => {
+  const currentYear = new Date().getFullYear();
+  const years = [
+    {
+      label: 'All Data',
+      value: 'all',
+    },
+  ];
+  let year = MIN_DATA_YEAR;
```
edmofro, line 41: Should `MIN_DATA_YEAR` be a number rather than a string, given the way it's used here? Clearly it must work, but it feels a bit type-loose!

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -19,26 +19,62 @@ import {
...
+  while (year <= currentYear) {
```
edmofro, line 42: I wonder if we should go the other way around, so the most recent (probably most relevant) is closest to the top and to the default? Is this in Dave's design?

edmofro (approved):
Love it!

packages/lesmis/src/views/DashboardReportTabView.js
tcaiger: Yes that's a good point - incrementing on the string is a bit dodgy. Love the heightened type awareness hangover from typescript.

packages/lesmis/src/views/DashboardReportTabView.js
tcaiger: The designs doesn't show the menu when it's open but I think you are spot on that the most relevant should be first.

tcaiger:
I'll merge this now and address the couple of tweaks in the next PR for simplicity.

PR #2455, opened 2021-04-23
2516-lesmis-map-overlay-year-selector

tcaiger:
### Changes:

- Make re-usable year selector component
- Add year selector to map overlays panel
- Remove footer on map view (as requested by Dave)

---

### Screenshots:

See Issue


packages/lesmis/src/components/YearSelector.js
```diff
@@ -0,0 +1,71 @@
...
+    {
+      label: 'All Data',
+      value: ALL_DATES_VALUE,
+    },
+  ];
+
+  const startYear = parseInt(MIN_DATA_YEAR);
+  let endYear = new Date().getFullYear();
+
+  while (startYear <= endYear) {
```
edmofro, line 23:
```suggestion
  while (year >= startYear) {
```
Two suggestions:
- use `year` as variable name, because after the first iteration it's no longer the final year
- flip the order of the condition - this one's just personal preference, I always expect to see the variable that's changing on the left of the condition in a loop

packages/lesmis/src/views/DashboardView.js
```diff
@@ -50,18 +50,41 @@ const makeTabOptions = entityType => [
...
   },
-  { value: 'primary', label: 'ESSDP Primary', Component: TabTemplate, Body: 'ESSDP Primary' },
   {
-    value: 'indicatorSelection',
-    label: 'Free Indicator Selection',
+    value: 'international',
+    label: 'International reporting on SDGs',
     Component: TabTemplate,
-    Body: 'Free Indicator Selection',
+    Body: 'International reporting on SDGs',
```
edmofro, line 87: Oh I think we need the indicator selection on top of the ones Erin listed thanks

edmofro (approved):
Beautifully structured! I really like the way you've pulled apart the business logic of converting to the api format from the component

Some nice other tweaks in there too - setting up nicely for map overlay loading indicator ;-)

PR #2457, opened 2021-04-23
Add 'psss' hierarchy to report request

rohan-bes:


packages/report-server/.env.example
```diff
@@ -10,3 +10,4 @@ DHIS_USERNAME=
 JWT_SECRET=
 PORT=
 API_CLIENT_SALT=
+ENTITY_SERVER_API_URL=
```
kael89, line 13: I didn't change the name to `ENTITY_SERVER_API_URL` to be consistent with the existing convention in `psss-server`. Ideally we'd make all such variables consistent across packages!

```suggestion
ENTITY_API_URL=
```

kael89 (approved):
Looks good (with one small change but pre-approving)!

My only comment is that I can still fetch reports in the current version of `psss-server`, without having to pass a `hierarchy` - no error is thrown.

kael89:
> My only comment is that I can still fetch reports in the current version of `psss-server`, without having to pass a `hierarchy` - no error is thrown.

I actually got the error now after running the latest migrations 🙌

packages/report-server/.env.example
rohan-bes: Yup, I'll change this but it's also `ENTITY_SERVER_API_URL` in `lesmis-server` as well...

kael89: True, as discussed in https://github.com/beyondessential/tupaia/pull/2382#discussion_r612213598

For now being consistent across the package should be fine, the next step would be ensuring cross-package consistency

PR #2458, opened 2021-04-23
Report Server fetch via POST

EMcQ-BES:
### Changes:

- Allow report server to process report fetches from POST requests
- Allow lesmis server to forward to the report server using POST requests
- Send POST requests from lesmis

### DIFF
--- a/packages/report-server/src/types.ts
+++ b/packages/report-server/src/types.ts
@@ -8,14 +8,21 @@ import { ParamsDictionary, Query } from 'express-serve-static-core';
 import { AccessPolicy } from '@tupaia/access-policy';
 import { TupaiaDatabase, ModelRegistry } from '@tupaia/database';
 
-export interface FetchReportQuery extends Query {
+export interface FetchReportFilter {
   organisationUnitCodes: string;
   period?: string;
   startDate?: string;
   endDate?: string;
 }
 
-export interface FetchReportParams extends ParamsDictionary {
+export interface FetchReportQuery extends Query, FetchReportFilter {}
+
+interface ReportsRequestBody extends FetchReportFilter {
+  testConfig?: ReportConfig;
+  testData?: Record<string, string | number>[];
+}
+
+interface FetchReportUrlParams extends ParamsDictionary {
   reportCode: string;
 }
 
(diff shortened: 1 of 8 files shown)


packages/report-server/src/types.ts
```diff
@@ -30,6 +30,7 @@ export interface ReportConfig {
 interface ReportsRequestBody {
   testConfig?: ReportConfig;
   testData?: Record<string, string | number>[];
+  queryOverride?: FetchReportQuery;
```
rohan-bes, line 33: I think I just had a chat with Edwin about this and he mentioned he might be thinking of changing this, but I feel it might be a little nicer if the query options were just present on the root level of this type rather than nested with the `queryOverride` field?

For what it's worth the plan is to move the `testConfig` and `testData` fields out sometime soon, I wanna create an entirely separate route for the 'test' logic

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -132,6 +132,18 @@ export class ApiBuilder {
...
+  post<T extends ExpressRequest<T> = Request>(
+    path: string,
+    handler: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>,
+  ) {
+    if (this.verifyAuthMiddleware) {
+      this.app.post(path, this.attachSession, this.verifyAuthMiddleware, handler);
+    } else {
+      this.app.post(path, this.attachSession, handler);
+    }
+    return this;
```
rohan-bes, line 144: I guess it might be nice to consolidate this logic into a single function as it will probably be the same for `PUT` and `DELETE` as well.

Perhaps something like:
```
  get<T extends ExpressRequest<T> = Request>(
    path: string,
    handler: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>,
  ) {
    return addRoute('get', path, handler);
  }
    
  post<T extends ExpressRequest<T> = Request>(
    path: string,
    handler: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>,
  ) {
    return this.addRoute('post', path, handler);
  }
    
  addRoute<T extends ExpressRequest<T> = Request>(
    method: 'get' | 'post',
    path: string,
    handler: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>,
  ) {
    if (this.verifyAuthMiddleware) {
      this.app[method](path, this.attachSession, this.verifyAuthMiddleware, handler);
    } else {
      this.app[method](path, this.attachSession, handler);
    }
    return this;
  }
```

rohan-bes (changes requested):
Just a couple comments, but looks good overall 😄 

Really glad this worked and we didn't need to do anything more complicated to get this working 👍

edmofro:
Thanks for the comments @rohan-bes - I'm also pro having the query params to the root of the body :-)

Have made changes to Ethan's PR to respond to those comments as he's busy with other PRs, so ready to re-review

rohan-bes (approved):
Looks great!

PR #2460, opened 2021-04-23
2606 add unknown entities to Samoa and exclude from explore

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/2606

Basically this migration creates "Unknown" entities for Samoa for the `district`, `facility` and `village` types (e.g. "Unknown District", "Unknown Village" etc) and excludes them from showing up in the Explore project by strictly defining which districts should be visible in Explore in the `entity_relation` table (e.g. only the two existing districts & not the new "Unknown" one)


packages/database/src/migrations/20210422030246-CreateUnknownEntitiesForSamoa-modifies-data.js
```diff
@@ -0,0 +1,104 @@
...
+  name: 'Unknown Village',
+  type: 'village',
+  country_code: 'WS',
+  metadata: '{"dhis": {"isDataRegional": true}}',
+  attributes: '{}',
+};
+
+const ENTITY_RELATION_UPOLU_EXPLORE = {
+  id: generateId(),
+  parent_id: '5df1b88c61f76a485cd1ca09',
```
edmofro, line 55: I think I've said this in an earlier PR review for you - please don't just have ids with no explanation.

Ideal way: Add an extra query that fetches the id based on the entity code

Not so ideal way: Hard code the id with a comment indicating the code it relates to

Not allowed way: Just the id

Would be great to do this in the ideal way - search `codeToId` in `migrations` for examples

Screen Shot 2021-04-27 at 9 43 15 AM

edmofro (changes requested):
Clever solution! Just a minor point to fix up to make it more readable and robust

avaek:
Thanks for the review @edmofro - makes perfect sense! I've made those changes, they can perhaps be done in a cleaner way but this way works & I felt like I was wasting too much time trying to figure out how to do it more cleanly

packages/database/src/migrations/20210422030246-CreateUnknownEntitiesForSamoa-modifies-data.js
```diff
@@ -50,40 +49,51 @@ const NEW_UNKNOWN_VILLAGE_ENTITY = {
...
-const ENTITY_RELATION_SAVAII_EXPLORE = {
-  id: generateId(),
-  parent_id: '5df1b88c61f76a485cd1ca09',
-  child_id: '5df1b88c61f76a485ce4e6b1',
-  entity_hierarchy_id: '5e9d06e261f76a30c400001b',
-};
-
 exports.up = async function(db) {
-  await insertObject(db, 'entity', NEW_UNKNOWN_DISTRICT_ENTITY);
+  const SELECT_EXPLORE_ENTITY_HIERARCHY = await db.runSql(`
```
edmofro, line 53: consts inside the function generally use camelCase rather than UPPER_SNAKE_CASE

packages/database/src/migrations/20210422030246-CreateUnknownEntitiesForSamoa-modifies-data.js
```diff
@@ -50,40 +49,51 @@ const NEW_UNKNOWN_VILLAGE_ENTITY = {
...
+    SELECT id FROM entity_hierarchy WHERE name = 'explore';
+  `);
+  const [EXPLORE_ENTITY_HIERARCHY] = SELECT_EXPLORE_ENTITY_HIERARCHY.rows;
```
edmofro, line 56: Clever destructuring of the first array item!!

edmofro (changes requested):
Yeah the more DRY way to do this would be to pull out a util function called `nameToId` which just contains the code you've repeated in the up and down. But this is all good - please just change the case on those variable names

avaek:
> Yeah the more DRY way to do this would be to pull out a util function called nameToId which just contains the code you've repeated in the up and down. But this is all good - please just change the case on those variable names

Ohh yup that makes sense, thanks! I've fixed up the variable names now

edmofro (approved):
Nice work @avaek!

PR #2461, opened 2021-04-23
Bugfix - Increase bar chart preview base padding to 6

nick-beyondessential:
### Changes:
- increase the base value for the bar chart preview x-axis padding from 4 to 6. This seems to fix the overflow issue with the chart preview when only two period values are present without dramatically altering other preview of more or less bars.
- The root cause maybe deeper in the workings or rechart and happy to hear an alternative strategy else this seems like a quick win.


kael89 (approved):
Fix seems good to me! Let's just add some regression testing notes for a quick checking that other bar chart reports still render nicely

PR #2422, opened 2021-04-19
Datasource code cleanup

billli0:
### Issue #:
[#2552](https://github.com/beyondessential/tupaia-backlog/issues/2552)

### Changes:
- Migration to change specific data source code to follow naming convention  
- Use updated data source code in visuals

PR #2464, opened 2021-04-26
2607 cleanup program indicator path

IgorNadj:

### DIFF
--- a/packages/database/src/migrations/20210419032448-MarkProgramIndicatorAsBorked-modifies-data.js
+++ b/packages/database/src/migrations/20210419032448-MarkProgramIndicatorAsBorked-modifies-data.js
@@ -5,23 +5,27 @@ var type;
 var seed;
 
 /**
-  * We receive the dbmigrate dependency from dbmigrate initially.
-  * This enables us to not have to rely on NODE_PATH.
-  */
-exports.setup = function(options, seedLink) {
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
   dbm = options.dbmigrate;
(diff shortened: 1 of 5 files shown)


packages/database/src/migrations/20210419032448-MarkProgramIndicatorAsBorked-modifies-data.js
```diff
@@ -0,0 +1,27 @@
...
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function(db) {
+  await db.runSql(`update data_source SET config = config || '{"dhisIndicatorWrappingBug": true}' WHERE code = 'Total_Positive_Malaria_Cases';`);
```
chris-bes, line 18: Nice idea putting this in data_source config. The only concern I have is @billli0 is consolidating the data source codes in this PR (https://github.com/beyondessential/tupaia/pull/2422/files#diff-55fe0a1da2982d3cc5a26b9246ea67ba2e6b8af5ddb2971ceaf15e5f9dbb617aR29), and `Total_Positive_Malaria_Cases` is included. If you're confident that this is going to be merged in dev first then it's fine, but if not maybe we can change filtering by `code` to `id`?

chris-bes (approved):
Looks good! Will leave you to decide on what I suggested

packages/database/src/migrations/20210419032448-MarkProgramIndicatorAsBorked-modifies-data.js
IgorNadj: Ah nice catch, changed it to use id

PR #2465, opened 2021-04-26
2714: Improve performance of 'PSSS_Total_Sites_Reported' indicator

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2714:

Altered events index on analytics table to perform well even if no data element code is specified

Arguably I should have created a migration to change this index on the existing analytics table, however that would require creating a mechanism for performing migrations as a user other than the tupaia user and that seemed rather out of scope (the analytics table is owned by the mvrefresh user)

- [ ] PR Task: Drop old index and create new one


kael89 (approved):
So glad this turned out to be as easy as adding a new `index`. Still, finding the root cause wouldn't have been easy - great investigation! 👏 🏇

rohan-bes:
@kael89 Yup, we will need to manually drop and re-create the index in dev/prod. Will add a PR task for this

PR #2467, opened 2021-04-26
2643 dengue bar district

nick-beyondessential:
### Changes:
- migration to add Laos EOC District (sub_district) level Dengue Cases bar report


rohan-bes (approved):
Looks great!

PR #2470, opened 2021-04-27
1398: Added Strive Facility EpiCurve Dashboard

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1398:

Pretty straight-forward dashboard 👍


packages/database/src/migrations/20210426232045-AddStriveFacilityEpiCurveDashboard-modifies-data.js
```diff
@@ -0,0 +1,81 @@
...
+    ],
+    aggregationType: 'SUM_EACH_WEEK',
+    entityAggregation: {
+      dataSourceEntityType: 'case',
+      aggregationEntityType: 'facility',
+      aggregationType: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+    },
+  },
+  viewJson: {
+    name: 'Lab Confirmed Positive Results, Line Graph',
```
billli0, line 33:
```suggestion
    name: 'Lab Confirmed Positive Results',
```

Do you suggest we can remove this in report title?

billli0 (approved):
LGTM🕺🏻! Just a minor suggestion on the report title.

packages/database/src/migrations/20210426232045-AddStriveFacilityEpiCurveDashboard-modifies-data.js
rohan-bes: I might leave as is for now, as there is another dashboard that is a 'Lab Confirmed Positive Results' dashboard as well.

It's possible this dashboard ought to be called "Epi Curve" (I'm really not sure I think I'd need Erin's input) but this is the kind of thing that can easily be edited via the admin panel post-release 👍

PR #2471, opened 2021-04-27
2109 weekly case trend graph palau

billli0:
### Issue #:
[#2111 PSSS Palau Facility Level Weekly Case Number Trend Graph](https://github.com/beyondessential/tupaia-backlog/issues/2111)
[#2109 PSSS Palau Country Level Weekly Case Number Trend Graph](https://github.com/beyondessential/tupaia-backlog/issues/2109)

### Changes:
- New aggregation type <`FINAL_EACH_WEEK_FILL_EMPTY_WEEKS `> , and new config <`fillEmptyPeriodsWith`>, with that we can specify `null` for empty period, so that we can show the gap in line chart. See [screenshot](https://github.com/beyondessential/tupaia-backlog/issues/2111#issuecomment-827349592).
- Minor change in `getContinuousPeriodsForAnalytics.js`
- Migration to add two dashboard reports
- Change previous dashboard reports code for id consistency, and move them into diff group


packages/aggregator/src/analytics/aggregateAnalytics/aggregations/utils/getContinuousPeriodsForAnalytics.js
```diff
@@ -21,13 +21,14 @@ export const getContinuousPeriodsForAnalytics = (
   const periodsInAnalytics = analytics.map(analytic =>
     convertToPeriod(analytic.period, aggregationPeriod),
   );
+  const sortedPeriodsInAnalytics = periodsInAnalytics.sort(compareAsc);
   const endPeriod =
     !continueTilCurrentPeriod && periodsInAnalytics.length
-      ? Math.max(...periodsInAnalytics).toString()
+      ? sortedPeriodsInAnalytics[sortedPeriodsInAnalytics.length - 1].toString() // Max
       : getCurrentPeriod(aggregationPeriod);
```
billli0, line 28: For weekly period we have string items in periodsInAnalytics (e.g. ['2021W22', '2021W19', '2021W25', ...]), which couldn't play with Math.max() and Math.min()

packages/database/src/migrations/20210416052750-DailyCasesTrendGrahphPalauFacilityLevel-modifies-data.js
```diff
@@ -19,30 +19,30 @@ exports.setup = function (options, seedLink) {
 const DISEASE_DATA_ELEMENTS = {
   AFR: {
     dataElement: 'PSSS_AFR_Daily_Cases',
-    name: 'Acute fever and rash',
+    name: 'Acute Fever and Rash',
```
rohan-bes, line 22: I'm guessing this migration exists only on the epic branch you're merging in to? In general I am somewhat against editing existing migrations as it can cause issues both on feature/epic branches and in the user's local env, but on the other hand it keeps our migrations themselves cleaner. I think as long as you're conscious about ensure this updated migration get run then I'm okay with it in this instance 👍

packages/database/src/migrations/20210420031825-WeeklyCasesTrendGraphForPalauCountry-modifies-data.js
```diff
@@ -0,0 +1,120 @@
...
+        unit: 'week',
+        offset: 0,
+      },
+      start: {
+        unit: 'week',
+        offset: -52,
+      },
+    },
+    presentationOptions: {
+      periodTickFormat: '[W]w',
```
rohan-bes, line 93: Ahh I've been wondering how to do this, looks awesome will try it out!

rohan-bes (changes requested):
Couple of requested changes, but looking good overall!

packages/aggregator/src/analytics/aggregateAnalytics/aggregateAnalytics.js
```diff
@@ -82,6 +82,8 @@ export const aggregateAnalytics = (
         },
         QUARTER,
       );
+    case AGGREGATION_TYPES.FILL_EMPTY_WEEK_WITH_NULL:
+      return getFinalValuePerPeriod(analytics, { defaultValue: null, ...aggregationConfig }, WEEK);
```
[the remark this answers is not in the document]
billli0: Yeah you are right we should follow the aggregation type naming convention. It is more readable and extendable in the future. 

Another thing I would like to discuss is to put `WITH_NULL` at the end of `FINAL_EACH_WEEK_FILL_EMPTY_WEEKS` to be `FINAL_EACH_WEEK_FILL_EMPTY_WEEKS_WITH_NULL`. To me `...FILL_EMPTY...` convention is to use `.getContinuousValues()`, which will fill previous value, while here we want to fill it with null. 

What do you think?

packages/database/src/migrations/20210416052750-DailyCasesTrendGrahphPalauFacilityLevel-modifies-data.js
billli0: Thanks for reminding, I guess that is a trade off but I didn't consider that will also effect in the teammates' local dev.

packages/database/src/migrations/20210420031825-WeeklyCasesTrendGraphForPalauCountry-modifies-data.js
billli0: I thought that was a new functionality as well.

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/utils/getContinuousPeriodsForAnalytics.js
```diff
@@ -21,13 +21,14 @@ export const getContinuousPeriodsForAnalytics = (
   const periodsInAnalytics = analytics.map(analytic =>
     convertToPeriod(analytic.period, aggregationPeriod),
   );
+  const sortedPeriodsInAnalytics = periodsInAnalytics.sort(compareAsc);
   const endPeriod =
     !continueTilCurrentPeriod && periodsInAnalytics.length
-      ? Math.max(...periodsInAnalytics).toString()
+      ? sortedPeriodsInAnalytics[sortedPeriodsInAnalytics.length - 1].toString() // Max
```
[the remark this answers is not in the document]
billli0: I really love your suggestion, pretty handy 👍 And yeah I agree we should find some dashboard/map overlay that use this functions for testers.

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -106,25 +106,26 @@ class FinalValueAggregator {
...
-        }
-        if (mostRecentValue !== undefined) {
-          values.push({
-            ...mostRecentValue,
-            period,
-          });
-        }
+        const valueForPeriodFromCache = organisationUnitCache[period];
+        values.push({
+          ...(checkIfFillingContinuousValues && mostRecentValue),
```
billli0, line 124: Here we couldn't specify the default value because we don't have dataElement and organisationUnit like others do. But a period with nothing, the frontend will know there should ba a gap. Values example: 

> [
    {
  dataElement: 'POP21AND23',
  organisationUnit: 'TO_Vainihc',
  period: '201911',
  value: 398
},
{
  period: '201912'
} // show gap
]

rohan-bes: We wanna return:
 ```
 {
dataElement: 'POP21AND23',
organisationUnit: 'TO_Vainihc',
period: '201911',
value: null
}
```

File: packages/aggregator/src/aggregationTypes.js [hunk 7]
```diff
@@ -13,6 +13,7 @@ export const AGGREGATION_TYPES = {
   FINAL_EACH_MONTH: 'FINAL_EACH_MONTH',
   FINAL_EACH_QUARTER_FILL_EMPTY_QUARTERS: 'FINAL_EACH_QUARTER_FILL_EMPTY_QUARTERS',
   FINAL_EACH_QUARTER: 'FINAL_EACH_QUARTER',
+  FINAL_EACH_WEEK_FILL_EMPTY_WEEKS_WITH_NULL: 'FINAL_EACH_WEEK_FILL_EMPTY_WEEKS_WITH_NULL',
```
rohan-bes, line 16: As discussed, we might wanna just make this `FINAL_EACH_WEEK_FILL_EMPTY_WEEKS` and override the default of `previous` with `null` in the aggregation config

packages/aggregator/src/analytics/aggregateAnalytics/aggregateAnalytics.js
```diff
@@ -48,7 +48,7 @@ export const aggregateAnalytics = (
         analytics,
         {
           ...aggregationConfig,
-          fillEmptyPeriodsTilNow: true,
+          fillEmptyPeriodsWith: 'previous',
```
rohan-bes, line 51: Can we flip the order of these so that the config can override the default of `previous`

rohan-bes (changes requested):

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -106,28 +107,60 @@ class FinalValueAggregator {
...
+      aggregationPeriod,
+      checkIfFillingContinuousValues,
+    );
 
+    let values;
+    this.cache.iterateOrganisationUnitCache(organisationUnitCache => {
+      values = checkIfFillingContinuousValues
+        ? this.getContinuousValues(organisationUnitCache, periods)
+        : this.fillWithPresetValues(organisationUnitCache, periods, fillEmptyPeriodsWith);
+    });
```
rohan-bes, line 163: Hmmm I think something isn't working here, on my local I see incorrect data for a COVID AU dashboard that uses this aggregation:
image


Tbh I think we can take a step back here and use much simpler logic based on the original code to handle this logic. The logic should be:
- Where we were using the `mostRecentValue`, now use the `mostRecentValue` if `fillEmptyPeriodsWith` = `previous` otherwise use `fillEmptyPeriodsWith`

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -106,28 +107,60 @@ class FinalValueAggregator {
...
     });
+    return values;
+  }
+
+  getFilledValues(analytics, aggregationPeriod, fillEmptyPeriodsWith) {
+    const checkIfFillingContinuousValues = fillEmptyPeriodsWith === 'previous';
+    const periods = getContinuousPeriodsForAnalytics(
+      analytics,
+      aggregationPeriod,
+      checkIfFillingContinuousValues,
```
rohan-bes, line 155: I'm not entirely sure if `checkIfFillingContinuousValues` is what we want to be using to determine whether we get periods up until the current period or not... It seems to be that this ought to be another config option though I also feel it's hard to make the argument that we should do this...

Perhaps let's just leave a comment here clarifying why this decision was made?

packages/database/src/migrations/20210420031825-WeeklyCasesTrendGraphForPalauCountry-modifies-data.js
```diff
@@ -0,0 +1,123 @@
...
+      {
+        type: 'FINAL_EACH_WEEK_FILL_EMPTY_WEEKS',
+        config: {
+          fillEmptyPeriodsWith: null,
+        },
+      },
+    ],
+    entityAggregation: {
+      aggregationOrder: 'BEFORE',
+      dataSourceEntityType: 'facility',
```
rohan-bes, line 72: The `dataSourceEntityType` here should be `country`, and there's no need for an `aggregationEntityType`. The indicators will be responsible for converting facility level data to the country level data so the dashboards don't need to worry about that. This is true for other PSSS country level dashboards as well.

rohan-bes (changes requested):
I tested this code locally it was really cool to see the dashboard lines working correctly `null` was being passed! However I also noticed that a tooltip was still showing and that it had `value: 0` in it for the null weeks. I tracked this down to being an issue in `formatDataValueByType({ value: null }) => 0`. It's part of what prompted this discussion in slack which I think we should follow in order to decide the best way to handle this: https://beyondessential.slack.com/archives/C0191QHH3QE/p1620027327003400

There is a simple (slightly hacky) way to fix this though if you do to `web-frontend/src/components/View/ChartWrapper/Tooltip.js` and change line 98 of to be:
```
  const filteredPayload = props.payload.filter(
    ({ value }) => value !== undefined && value !== null,
  );
```

A few more general changes requested, please message me if you don't understand any of the points 👍

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
billli0: I agree that makes sense to be another config option, but with that we will have more configs in db:
> aggregations: [
      {
        type: 'FINAL_EACH_WEEK_FILL_EMPTY_WEEKS',
        config: {
          fillEmptyPeriodsWith: null,
          continueTilCurrentPeriod: false,
        },
      }

So that is a trade off. TBH I prefer to have more configs because that makes code easier to read. What do you think?

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
billli0: This one works very well with less changes. Nice one!

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
rohan-bes: Hmm yeah I'm inclined to agree, perhaps something like:
```
{
   fillEmptyPeriodsWith: <any value> (default: 'previous')
   fillEmptyPeriodsTilNow: <boolean> (default: false)
}
```
I'm inclined to make the `fillEmptyPeriodsTilNow` be false by default as imo it does feel a little un-exepected... 

However, that would mean changing the config of all other existing dashboards/overlays that use this aggregation as I think the previous behaviour was to `fillEmptyPeriodsTilNow`... If you're happy to do this then I guess this would be the most complete solution.

Ultimately though, I'm also okay with us using the code you have currently (if fillling `previous` data, then fill til now, otherwise don't) as long as we add a comment explaining that this could/should be expanded to a config in the future? That way the next time that a dev needs to implement divergent logic (either a `previous` fill where they **don't** want to fill til now, or a non-`previous` fill where they **do** want to fill til now) they are suggested to do this 😄

rohan-bes:
@billli0 One last request btw, could we please add some unit tests for the new aggregation logic? Basically a unit test suite for `getFinalValuePerPeriod.js`. Happy for this card to go into testing before the unit tests are written btw 👍

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
billli0: Yeah I agree making it be false as default would make more sense. 
I dig in a bit and found this config is used in losts of visuals already. It is doable but I would suggest we can do it in a separate card or leave it for a dev next time.

billli0:
> @billli0 One last request btw, could we please add some unit tests for the new aggregation logic? Basically a unit test suite for `getFinalValuePerPeriod.js`. Happy for this card to go into testing before the unit tests are written btw 👍

Sounds good! That would be helpful for further extension.

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -106,12 +106,19 @@ class FinalValueAggregator {
...
+      aggregationPeriod,
+      checkIfFillingContinuousValues, // To get a continuous period till now, we need to set true for this option config (continueTilCurrentPeriod). This can be a new option config in the future.
+    );
 
     const values = [];
     this.cache.iterateOrganisationUnitCache(organisationUnitCache => {
-      let mostRecentValue;
+      let mostRecentValue = checkIfFillingContinuousValues
+        ? undefined
+        : Object.values(organisationUnitCache)[0];
```
billli0, line 121: Here we want to fill empty period with value 'test' in two orgUnits, see this test case as example. Screenshot: 
11620283114_ pic_hd copy

In this case `period` is '20210101 - 20210107'. We need a preset `mostRecentValue`, if not, it will start filling from `20210104` where org2 has value.

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.test.js
```diff
@@ -0,0 +1,128 @@
...
+        // org2
+        { dataElement: 'element1', organisationUnit: 'org2', period: '20210101', value: 'test' },
+        { dataElement: 'element1', organisationUnit: 'org2', period: '20210102', value: 'test' },
+        { dataElement: 'element1', organisationUnit: 'org2', period: '20210103', value: 'test' },
+        { dataElement: 'element1', organisationUnit: 'org2', period: '20210104', value: 'test' },
+        { dataElement: 'element1', organisationUnit: 'org2', period: '20210105', value: 1 },
+        { dataElement: 'element1', organisationUnit: 'org2', period: '20210106', value: 'test' },
+        { dataElement: 'element1', organisationUnit: 'org2', period: '20210107', value: 3 },
+      ]);
+    });
```
rohan-bes, line 126: @billli0 I don't think this is correct. I believe we should only by filling missing values within found values **per** org unit. eg:
```suggestion
    it(`Filling preset value by 'fillEmptyPeriodsWith'`, () => {
      const testAnalytics = [
        { dataElement: 'element1', organisationUnit: 'org1', period: '20210101', value: 1 },
        { dataElement: 'element1', organisationUnit: 'org1', period: '20210104', value: 2 },
        { dataElement: 'element1', organisationUnit: 'org2', period: '20210105', value: 1 },
        { dataElement: 'element1', organisationUnit: 'org2', period: '20210107', value: 3 },
      ];
      const aggregationPeriod = 'DAY';
      const aggregationConfig = {
        fillEmptyPeriodsWith: 'test',
      };
      expect(
        getFinalValuePerPeriod(testAnalytics, aggregationConfig, aggregationPeriod),
      ).toStrictEqual([
        // org1
        { dataElement: 'element1', organisationUnit: 'org1', period: '20210101', value: 1 },
        { dataElement: 'element1', organisationUnit: 'org1', period: '20210102', value: 'test' },
        { dataElement: 'element1', organisationUnit: 'org1', period: '20210103', value: 'test' },
        { dataElement: 'element1', organisationUnit: 'org1', period: '20210104', value: 2 },
        // org2
        { dataElement: 'element1', organisationUnit: 'org2', period: '20210105', value: 1 },
        { dataElement: 'element1', organisationUnit: 'org2', period: '20210106', value: 'test' },
        { dataElement: 'element1', organisationUnit: 'org2', period: '20210107', value: 3 },
      ]);
    });
```

Is there a reason you want the behaviour to be as you have written in this test?

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.test.js
```diff
@@ -0,0 +1,128 @@
...
+        period,
+      }));
+    };
+
+    it('Filling continuous values', () => {
+      const defaultVariables = { dataElement: 'element1', organisationUnit: 'org1', value: 1 };
+
+      const startDate = '20210101';
+      const aggregationPeriod = 'WEEK';
+      const testAnalytics = [{ ...defaultVariables, period: startDate }];
```
rohan-bes, line 53: Probably worth having more than just 1 analytic here, to properly test that `previous` logic is working correctly

rohan-bes (changes requested):

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.test.js
billli0: Hi @rohan-bes, I agree we should fill in missing value only for the gap, but when we get the period from [getContinuousPeriodsForAnalytics()](https://github.com/beyondessential/tupaia/blob/e6b5847088de0a5c677b586889cec2657e1ac794/packages/aggregator/src/analytics/aggregateAnalytics/aggregations/utils/getContinuousPeriodsForAnalytics.js#L16), we will get the earliest and the latest date from analytics, which doesn't count on the orgUnit and dataElement. 

If that's the right behaviour, do you suggest we should make it count on the orgUnit and dataElement? I guess it should be simple.

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.test.js
billli0: Good idea! Will add that to the test.

packages/aggregator/src/__tests__/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.test.js
rohan-bes: Ahh I see your point. Yeah I'd say it'd be best then to recalculate the `periods` using `getContinuousPeriodsForAnalytics()` per orgUnit and dataElement.

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -106,12 +106,19 @@ class FinalValueAggregator {
...
+  getFilledValues(analytics, aggregationPeriod, fillEmptyPeriodsWith) {
+    const checkIfFillingContinuousValues = fillEmptyPeriodsWith === 'previous';
+    const periods = getContinuousPeriodsForAnalytics(
+      analytics,
+      aggregationPeriod,
+      checkIfFillingContinuousValues, // To get a continuous period till now, we need to set true for this option config (continueTilCurrentPeriod). This can be a new option config in the future.
+    );
 
     const values = [];
     this.cache.iterateOrganisationUnitCache(organisationUnitCache => {
```
rohan-bes, line 118:
```suggestion
    const values = [];
    this.cache.iterateOrganisationUnitCache(organisationUnitCache => {
      const periods = getContinuousPeriodsForAnalytics(
        Object.values(organisationUnitCache),
        aggregationPeriod,
        checkIfFillingContinuousValues, // To get a continuous period till now, we need to set true for this option config (continueTilCurrentPeriod). This can be a new option config in the future.
```

Would something like this work to address your comment above? https://github.com/beyondessential/tupaia/pull/2471/files#r627825280

billli0: Do you mean test cases can explain the logic and we can remove this comment?

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -106,12 +106,18 @@ class FinalValueAggregator {
...
 
-  getContinuousValues(analytics, aggregationPeriod) {
-    const periods = getContinuousPeriodsForAnalytics(analytics, aggregationPeriod, true);
+  getFilledValues(aggregationPeriod, fillEmptyPeriodsWith) {
+    const checkIfFillingContinuousValues = fillEmptyPeriodsWith === 'previous';
 
     const values = [];
     this.cache.iterateOrganisationUnitCache(organisationUnitCache => {
-      let mostRecentValue;
+      const analyticsPerOrgUnit = Object.values(organisationUnitCache);
```
rohan-bes, line 114: Nice choice to break this out into a variable, the naming makes it very clear what's going on 👍

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -120,10 +126,13 @@ class FinalValueAggregator {
...
         }
         if (mostRecentValue !== undefined) {
-          values.push({
+          const valueForOneOrgUnit = {
             ...mostRecentValue,
             period,
-          });
+          };
+          if (!checkIfFillingContinuousValues && !valueForPeriod)
+            valueForOneOrgUnit.value = fillEmptyPeriodsWith;
```
rohan-bes, line 134:
```suggestion
          if (!checkIfFillingContinuousValues && !valueForPeriod) {
            valueForOneOrgUnit.value = fillEmptyPeriodsWith;
          }
```

Minor thing, but I think we always use brackets around code blocks in our codebase 👍

packages/aggregator/src/analytics/aggregateAnalytics/aggregations/getFinalValuePerPeriod.js
```diff
@@ -106,12 +106,18 @@ class FinalValueAggregator {
     this.cache = cache;
   }
 
-  getContinuousValues(analytics, aggregationPeriod) {
-    const periods = getContinuousPeriodsForAnalytics(analytics, aggregationPeriod, true);
+  getFilledValues(aggregationPeriod, fillEmptyPeriodsWith) {
+    const checkIfFillingContinuousValues = fillEmptyPeriodsWith === 'previous';
```
rohan-bes, line 110:
```suggestion
    const fillingPreviousValues = fillEmptyPeriodsWith === 'previous';
```
Minor suggestion but reads a bit more clearly to me

packages/web-config-server/src/apiV1/dataBuilders/generic/analytics/analyticsPerPeriod.js
```diff
@@ -30,6 +30,7 @@ class AnalyticsPerPeriodBuilder extends DataBuilder {
   }
 
   getResultsPerPeriod = results => {
+    if (!results) return [];
```
rohan-bes, line 33: Curious about this change, what was the reason for it? I think at the least we ought be to returning `{}` no? as that seems to be the shape of `resultsPerPeriod`?

Might be easier to do:
```suggestion
  getResultsPerPeriod = (results = [])=> {
```

rohan-bes (approved):
A few small changes requested, but approving in advance as they're nothing major. Nice work @billli0, looks really good and the unit tests are awesome!

packages/web-config-server/src/apiV1/dataBuilders/generic/analytics/analyticsPerPeriod.js
billli0: Yes you are right, it should return {}. 

The reason to add this is that, when frontend requests data, it will request twice (similar behaviour in [this card](https://github.com/beyondessential/tupaia-backlog/issues/2524). When config server is managing the first response, result could be `undefined` from aggregator if there is no data, which will cause error when performing `.forEach()` later, with that, error will be sent to frontend, which is not what we want...

I'm happy to make a new card and investigate it.

billli0: Sorry I couldn't reproduce now, I guess we have data coming in (a few days ago) and it fixed this issue. I will still create a new card and mention it.

PR #2473, opened 2021-04-28
1553 Replace PEHS matrix placeholder

billli0:
### Issue #:
[#1553](https://github.com/beyondessential/tupaia-backlog/issues/1553)

### Screenshots:
[Images in comment](https://github.com/beyondessential/tupaia-backlog/issues/1553#issuecomment-828855004)


packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
```diff
@@ -154,6 +157,12 @@ const buildMatrixDataFromViewContent = viewContent => {
   };
 };
 
+const getPlaceholder = (presentationOptions, categoryPresentationOptions) => {
+  if (!presentationOptions && !categoryPresentationOptions) return textOnlyMatrixPlaceholder;
+  if (presentationOptions?.applyLocation?.columnIndexes) return mixMatrixPlaceholder;
```
billli0, line 162: Logic from [code](https://github.com/beyondessential/tupaia/blob/954a73e64d3a5df1b0da147d78b109daf0d85d92/packages/web-frontend/src/components/View/MatrixWrapper/components/Row.js#L119)

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
```diff
@@ -154,6 +157,12 @@ const buildMatrixDataFromViewContent = viewContent => {
   };
 };
 
+const getPlaceholder = (presentationOptions, categoryPresentationOptions) => {
+  if (!presentationOptions && !categoryPresentationOptions) return textOnlyMatrixPlaceholder;
```
billli0, line 161: Logic from [code](https://github.com/beyondessential/tupaia/blob/954a73e64d3a5df1b0da147d78b109daf0d85d92/packages/web-frontend/src/components/View/MatrixWrapper/components/Matrix.js#L142)

edmofro: Would be better to pull this out into a named boolean, or a named function (like the place you got it from does). That makes the condition a lot easier to read inline, because it just looks like 
```suggestion
  if (!isUsingDots) return textOnlyMatrixPlaceholder;
```

And then the logic for working out `isUsingDots` is clearly defined on another line/in another place.

Ideally, this logical block would be shared across both components. A couple of options there:
- pull out a util function that is common to both
- work it out in the wrapper and pass the result down to the child

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
edmofro: Same comment

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
```diff
@@ -296,8 +305,9 @@ export class MatrixWrapper extends Component {
   }
 
   render() {
-    const { isEnlarged, isExporting } = this.props;
-
+    const { isEnlarged, isExporting, viewContent } = this.props;
+    const { presentationOptions, categoryPresentationOptions } = viewContent;
+    const matrixPlaceholder = getPlaceholder(presentationOptions, categoryPresentationOptions);
```
edmofro, line 310: I think it's more the responsibility of the function to work out which parts of viewContent it needs - those two aren't otherwise needed or known about within this outer render function, so it's too detailed for its scope.
```suggestion
    const matrixPlaceholder = getPlaceholder(viewContent);
```

edmofro (changes requested):
Screenshots look great! Just a couple of comments

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
billli0: Thanks Edwin, I agree having a named boolean is mush easier to read, definitely will put it in. 👍 
And I was thinking to use `isUsingDots` here then pass in to children component, but there is different `isUsingDots` logic for `Row` and `RowGroup` component, which in `Row` it is a dynamic variable that can be customed to show text for special index.

edmofro: Cool, in that case perhaps pull out a util function that can be called from both places :-)

packages/web-frontend/src/components/View/MatrixWrapper/components/Matrix.js
```diff
@@ -139,24 +140,20 @@ export class Matrix extends PureComponent {
...
     const conditions = this.props.presentationOptions.conditions || [];
     return conditions.filter(condition => !!condition.legendLabel).length > 0;
   }
 
-  setRowRef(rowElement) {
-    if (!this.rowElements.includes(rowElement)) {
-      this.rowElements.push(rowElement);
-    }
-  }
-
```
billli0, line 159: move up .setRowRef() to meet eslint requirement

edmofro (approved):
lgtm!

PR #2474, opened 2021-04-28
1221 Allow buildInternalDependencies to build only deps of a specific package

edmofro:
### Issue #: 
https://github.com/beyondessential/tupaia-backlog/issues/1221

### Changes:
buildInternalDependencies now optionally takes in --packageJsonPath to indicate the package.json of the package to build internal dependencies of

### DIFF
--- a/scripts/bash/backendStartDev.sh
+++ b/scripts/bash/backendStartDev.sh
@@ -55,5 +55,5 @@ else
     # add the watch flags to the server start process, as well as a 1 second delay to debounce the
     # many restarts that otherwise happen during the initial build of internal dependencies
     start_server="${start_server} --delay 1 ${watch_flags}"
-    yarn concurrently "${DIR}/buildInternalDependencies.sh --watch --withTypes" "eval ${start_server}"
+    yarn concurrently "${DIR}/buildInternalDependencies.sh --watch --packagePath ." "eval ${start_server}"
 fi
--- a/scripts/bash/backendStartDev.sh
+++ b/scripts/bash/backendStartDev.sh
@@ -49,7 +49,7 @@ if [[ ${skip_internal} == true ]]; then
     eval ${start_server}
 else
     echo "Internal dependencies are under watch for hot reload (use --skip-internal or -s for faster startup times)"
-    for PACKAGE in $(${DIR}/getInternalDependencies.sh ${PWD}); do
+    for PACKAGE in $(${DIR}/getInternalDependencies.sh .); do
         watch_flags="${watch_flags} --watch ../${PACKAGE}/dist"
     done
     # add the watch flags to the server start process, as well as a 1 second delay to debounce the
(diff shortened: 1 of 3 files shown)


scripts/bash/backendStartDev.sh
```diff
@@ -49,11 +49,11 @@ if [[ ${skip_internal} == true ]]; then
...
 else
     echo "Internal dependencies are under watch for hot reload (use --skip-internal or -s for faster startup times)"
-    for PACKAGE in $(${DIR}/getInternalDependencies.sh); do
+    for PACKAGE in $(${DIR}/getInternalDependencies.sh .); do
         watch_flags="${watch_flags} --watch ../${PACKAGE}/dist"
     done
     # add the watch flags to the server start process, as well as a 1 second delay to debounce the
     # many restarts that otherwise happen during the initial build of internal dependencies
     start_server="${start_server} --delay 1 ${watch_flags}"
-    yarn concurrently "${DIR}/buildInternalDependencies.sh --watch --withTypes" "eval ${start_server}"
```
edmofro, line 58: You'll notice this no longer builds types. I personally feel it's better to have something that is fast enough for people to use than comprehensive enough to cover all situations, but I might be missing the importance of building types all the time?

scripts/bash/getInternalDependencies.sh
```diff
@@ -1,2 +1,67 @@
...
+done
+for i in "${!internal_dependencies[@]}"; do
+    array_without_gaps+=( "${internal_dependencies[i]}" )
+done
+internal_dependencies=("${array_without_gaps[@]}")
+unset array_without_gaps
+
+# echo out result for calling script to pick up
+echo ${internal_dependencies[@]}
+exit 0
```
rohan-bes, line 66: Honestly the complexity of this script does daunt me somewhat wondering if it might be better as a Node Script (tupaia/scripts/node/Script.js)?

Don't wanna push this too hard though as I know we have a number of more complex scripts and refactoring this script may create more problems than it sovles...

rohan-bes (approved):
Approving, though with the suggestion that if it feels appropriate we could migrate this script to a node script.

Super excited about having this though, I suspect I'll be using it soon enough as it'll save me heaps of time developing micro-services. Also it could prove super useful if we can improve our build system to use something like there (where it looks at the changed packages and only runs tests for those packages and their dependants).

scripts/bash/getInternalDependencies.sh
edmofro: Yeah you know after finishing it, I absolutely agree, and wish I'd gone for that from the start. But, because of that same complexity I'd prefer not to jump back in and potentially break things again now. Maybe next time this needs tweaking we port it to a node script?

rohan-bes: I think it would be nice to create a separate node script that determines dependencies of any given package, and then this script can call that? The recursive logic is the only thing that I find particularly advanced for a bash script. I can think of a few ways we might wanna determine dependency trees between our packages so I think it'd come in handy 👍

edmofro: That'd be a good way to separate it, yeah. For now though, I've had enough of internal dependencies! 😂

PR #2480, opened 2021-04-29
2538 lesmis dashboard report fixes

tcaiger:
### Changes:

- Add error boundary to dashboard reports
- Fix data formatting parsing for number types
- Add laos link to the top of the entity browse menu
- Fix search icon entity type mapping
- Fix dashboard group tab defaults
- Fix Dashboard super tab Select padding

---

### DIFF
--- a/packages/lesmis/src/components/FetchLoader.js
+++ b/packages/lesmis/src/components/FetchLoader.js
@@ -7,6 +7,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import CircularProgress from '@material-ui/core/CircularProgress';
 import { SmallAlert } from '@tupaia/ui-components';
+import { ErrorBoundary } from './ErrorBoundary';
 import { FlexCenter } from './Layout';
 
 const SpinningLoader = () => (
@@ -30,7 +31,7 @@ export const FetchLoader = ({ isLoading, isError, error, children, Loader }) => 
     );
   }
 
-  return children;
+  return <ErrorBoundary>{children}</ErrorBoundary>;
 };
 
 FetchLoader.propTypes = {
--- a/packages/lesmis/src/views/DashboardReportTabView.js
+++ b/packages/lesmis/src/views/DashboardReportTabView.js
@@ -3,7 +3,7 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  *
  */
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { SmallAlert } from '@tupaia/ui-components';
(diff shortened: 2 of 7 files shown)


packages/lesmis/src/components/FetchLoader.js
```diff
@@ -30,7 +31,7 @@ export const FetchLoader = ({ isLoading, isError, error, children, Loader }) =>
     );
   }
 
-  return children;
+  return <ErrorBoundary>{children}</ErrorBoundary>;
```
tcaiger, line 34: This will show an error message instead of crashing the app if there are any errors rending in the children

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -52,7 +68,7 @@ export const DashboardReportTabView = ({ entityCode, TabSelector }) => {
               variant="scrollable"
               scrollButtons="auto"
             >
-              {Object.entries(data).map(([heading]) => (
+              {Object.keys(data).map(heading => (
```
edmofro, line 71: Good cleanup

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -26,11 +26,27 @@ const DashboardSection = styled(FlexCenter)`
...
+  if (dashboardNames.includes(DEFAULT_DASHBOARD_GROUP)) {
+    setSelectedDashboard(DEFAULT_DASHBOARD_GROUP);
+  } else {
+    setSelectedDashboard(dashboardNames[0]);
+  }
+};
+
 export const DashboardReportTabView = ({ entityCode, TabSelector }) => {
   const [selectedYear, setSelectedYear] = useState(ALL_DATES_VALUE);
   const [selectedDashboard, setSelectedDashboard] = useState(DEFAULT_DASHBOARD_GROUP);
```
edmofro, line 41: Perhaps this should start empty, now that the effect is taking care of setting the default if it exists

packages/utils/src/formatDataValueByType.js
```diff
@@ -81,8 +81,8 @@ const percentage = value => {
   return `${Math.round(percentageValue * floatNormalizer) / floatNormalizer}%`;
 };
 
-const number = (value, { presentationOptions = {} }) => {
-  const { valueFormat = '0,0' } = presentationOptions;
+const number = (value, { presentationOptions }) => {
+  const valueFormat = presentationOptions?.dataValueFormat ? valueFormat : '0,0';
```
edmofro, line 85: Unless I'm missing something, `valueFormat` isn't defined yet?
```suggestion
  const valueFormat = presentationOptions?.dataValueFormat || '0,0';
```

edmofro (changes requested):

packages/utils/src/formatDataValueByType.js
tcaiger: woops you are totally right. I've fixed it now.

packages/lesmis/src/views/DashboardReportTabView.js
edmofro: Optional suggestion here @tcaiger

edmofro (approved):

packages/lesmis/src/views/DashboardReportTabView.js
tcaiger: I did do this initially and noticed a small visual delay before it sets the default. Since 9 times out of 10 the default is in the set, I though I might as well set it initially and have it slightly more seamless. I don't really mind though if you'd rather I unset it?

edmofro: No that makes sense! When we have other dashboards it might be worth revisiting, but for now that's perfectly reasonable :-)

PR #2483, opened 2021-04-30
2501 lesmis more entity vitals

EMcQ-BES:
### Changes:

- Add Loading State
- Clean up messy code
- Add report for village vitals
- Various adjustments for looks

---

### Screenshots:

There are screenshots on the original issue


packages/lesmis/src/views/DashboardView.js
```diff
@@ -39,108 +61,341 @@ const RedTitle = styled(Typography)`
...
+const MapPlaceholder = () => (
+  <PlaceholderBox>
+    <Typography>TODO: Map</Typography>
+  </PlaceholderBox>
+);
+
+const CountryView = ({ vitals }) => {
+  return (
+    <Container maxWidth={false}>
+      <VitalsSection>
```
tcaiger, line 97:
```suggestion
const CountryView = ({ vitals }) => (
      <>
```

packages/lesmis/src/views/DashboardView.js
```diff
@@ -39,108 +61,341 @@ const RedTitle = styled(Typography)`
...
+`;
+const MapPlaceholder = () => (
+  <PlaceholderBox>
+    <Typography>TODO: Map</Typography>
+  </PlaceholderBox>
+);
+
+const CountryView = ({ vitals }) => {
+  return (
+    <Container maxWidth={false}>
```
tcaiger, line 96:
```suggestion
const CountryView = ({ vitals }) =>  (
    <>
```

packages/lesmis/src/views/DashboardView.js
```diff
@@ -39,108 +61,341 @@ const RedTitle = styled(Typography)`
...
+        <TitleContainer>
+          <GreyTitle>Development Partner Support</GreyTitle>
+        </TitleContainer>
+        {Object.entries(vitals).map(([key, value]) =>
+          value === true ? <PartnerLogo code={key} key={key} /> : null,
+        )}
+      </PartnersContainer>
+    </Container>
+  );
+};
```
tcaiger, line 125:
```suggestion
    </>
  );
```

packages/lesmis/src/views/DashboardView.js
```diff
@@ -39,108 +61,341 @@ const RedTitle = styled(Typography)`
...
-        <Wrapper>
-          <MuiContainer>
-            <Typography>Dashboard View</Typography>
-          </MuiContainer>
-        </Wrapper>
-      );
-  }
+
+  return (
+    <Wrapper>
```
tcaiger, line 366:
```suggestion
    <Wrapper>
        <Container maxWidth={false}>
```

packages/lesmis/src/api/queries/useVitalsData.js
```diff
@@ -8,100 +8,210 @@ import { get, post } from '../api';
...
 
-const useSchoolInformation = entityCode =>
+const useReport = (entity, reportName, options, enabled) =>
   useQuery(
-    ['vitals', 'school', entityCode],
-    () =>
-      get(`reportData/${entityCode}/LESMIS_school_vitals`, {
-        params: { endDate: '2021-01-01' },
-      }),
+    [reportName, entity?.code],
```
tcaiger, line 40:
```suggestion
    [reportName, entity?.code, options],
```

packages/lesmis/src/views/DashboardView.js
tcaiger: Might as well pull the container out to the main component to save a few lines of code

packages/lesmis/src/api/queries/useVitalsData.js
tcaiger: Need to include the options, otherwise we will get cached results even when the options are different.

tcaiger (changes requested):
Looking really good. Just need to add the options to the `useReport` hook and the small tweak to the container component and it's good to go

packages/lesmis/src/views/DashboardView.js
EMcQ-BES: Ah cool, I tried splitting it out but it kept complaining that I needed to return a single component. Using an empty component was the concept I was missing.

edmofro:
One extra visual thing: need some margin between elements in the vitals section

image

packages/lesmis/src/api/queries/useVitalsData.js
EMcQ-BES: Also needed to swap out the formatting of `endTime` just because I had been passing it as the current `Date()` which would include the current time and therefore never actually return a cached value if we passed it as the option.

EMcQ-BES:
> One extra visual thing: need some horizontal margin between elements in the vitals section

Added slight margin, we can fiddle with the amount if necessary

image

edmofro:
Looks good! Can you screenshot one with a `...` as well?

EMcQ-BES:
image

tcaiger (approved):

PR #2484, opened 2021-05-02
Add mock for yargs to storybook build

edmofro:
Fixes storybook not working because it can't load yargs


tcaiger (approved):
Great - I thought there must be a storybook side solution. It has the added benefit of handling future additions of yargs as well which is nice.

PR #2486, opened 2021-05-02
2746 legend too far down

edmofro:
### Issue #: 
https://github.com/beyondessential/tupaia-backlog/issues/2746

### Changes:
I didn't entirely remove the min height because below a certain point it seems better to have the legend disappear than to have the map become unusable

See issue for what it looks like at 350px


tcaiger (approved):
Nice - good move keeping in a small min-height. Cheers

PR #2489, opened 2021-05-03
2752 lesmis public permission group

EMcQ-BES:
### Changes:

- Swap `Laos Schools User` to `LESMIS Public`
- Update `mapOverlays` and `dashboardReports` to use `LESMIS Public`
- Update reference to `Laos Schools User` in `lesmis-server` constants file


edmofro (approved):
lgtm :-)

PR #2490, opened 2021-05-03
Change LESMIS reports to use LESMIS Public permission group

EMcQ-BES:
### Changes:

- Migrate LESMIS reports to use `LESMIS Public` user group


packages/database/src/migrations/20210503020225-LESMISReportsNewPermissionGroup-modifies-data.js
```diff
@@ -0,0 +1,44 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const permissionGroupNameToId = async (db, name) => {
+  const record = await db.runSql(`SELECT id FROM permission_group WHERE name = '${name}'`);
```
edmofro, line 18:
```suggestion
  const result = await db.runSql(`SELECT id FROM permission_group WHERE name = '${name}'`);
```
It's not really a record at this point

packages/database/src/migrations/20210503020225-LESMISReportsNewPermissionGroup-modifies-data.js
```diff
@@ -0,0 +1,44 @@
...
+  return record.rows[0] && record.rows[0].id;
```
edmofro, line 19: I'm not 100% confident on this because I'm not sure what transpiler the migrations go through, but try
```suggestion
  return record.rows[0]?.id;
```

edmofro (approved):
A couple of very minor things - honestly don't mind if you don't follow up given it's in a migration that probably won't be read nearly as many times as code in other places!

PR #2498, opened 2021-05-04
Remove the all data options from the year selectors

tcaiger:
### Changes:

- remove the all data option from dashboard and map overlay year selector
- set 2020 as the default year


EMcQ-BES (approved):
lgtm

PR #2499, opened 2021-05-04
2745 lesmis dashboard headings

EMcQ-BES:
### Changes:

- Rename dashboardGroups for LESMIS
- Add new empty dashboardGroups when necessary

### DIFF
--- a/packages/database/src/migrations/20210503020203-RenameDashboardHeadings-modifies-data.js
+++ b/packages/database/src/migrations/20210503020203-RenameDashboardHeadings-modifies-data.js
@@ -26,11 +26,12 @@ const renameDashboardGroups = async function (db, orgLevel, from, to) {
 
 const createEmptyDashboardGroup = async function (db, orgLevel, name) {
   await db.runSql(`
-    INSERT INTO "dashboardGroup"("organisationLevel","userGroup","organisationUnitCode","dashboardReports","name","code")
+    INSERT INTO "dashboardGroup"("organisationLevel","userGroup","organisationUnitCode","dashboardReports","name","code","projectCodes")
     VALUES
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210503020203-RenameDashboardHeadings-modifies-data.js
```diff
@@ -0,0 +1,211 @@
...
+  `);
+};
+
+const createEmptyDashboardGroup = async function (db, orgLevel, name) {
+  await db.runSql(`
+    INSERT INTO "dashboardGroup"("organisationLevel","userGroup","organisationUnitCode","dashboardReports","name","code","projectCodes")
+    VALUES
+    ('${orgLevel}','LESMIS Public','LA','{}','${name}','LA_${name
+    .split(' ')
+    .join('_')}_${orgLevel}_LESMIS_Public',
```
edmofro, line 33: Nice work removing the spaces

edmofro (commented):
Do we need to do anything to order the dashboard groups correctly?

EMcQ-BES:
> Do we need to do anything to order the dashboard groups correctly?

Yes we need to do something, but I don't have an exact answer of what.

Currently I believe they're just displayed in the order they appear in the database, which means that these newly created empty groups are always going to be displayed at the end. There's not a great way of rearranging them in the database without swapping around their ids which seems messy.

I'm not 100% sure whether we wanted to do all that in the scope of [2745](https://github.com/beyondessential/tupaia-backlog/issues/2745) or [2538](https://github.com/beyondessential/tupaia-backlog/issues/2538) where it was originally mentioned.

edmofro (approved):
Ok I'm gonna say the ordering thing is out of scope for this PR, in which case:

Very nice migration!

PR #2507, opened 2021-05-05
2748 sticky dashboard headers

edmofro:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2748


packages/lesmis/src/components/NavBar.js
```diff
@@ -18,7 +18,8 @@ import { useUser } from '../api';
 import { useUrlParams } from '../utils';
 
 const Container = styled.nav`
-  position: relative;
+  position: sticky;
+  top: 0;
```
edmofro, line 22: Not strictly included in the ticket definition, but seemed to be true from Dave's video that the nav bar is sticky

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -39,8 +62,38 @@ const setDefaultDashboard = (data, setSelectedDashboard) => {
...
+    return () => window.removeEventListener('scroll', detectScrolledPastTop);
+  }, [stickyBarsHeight]);
+
+  const scrollToTop = useCallback(() => {
+    // if the top of the dashboards container is above the sticky dashboard header, scroll to the top
+    if (isScrolledPastTop) {
+      topRef.current.scrollIntoView();
+      window.scrollTo({ top: window.pageYOffset - stickyBarsHeight, behavior: 'smooth' });
+    }
+  }, [isScrolledPastTop, stickyBarsHeight]);
```
edmofro, line 95: A bunch of this stuff may need to be pulled into a more generic location when we have multiple dashboards

packages/lesmis/src/components/Tabs.js
```diff
@@ -3,25 +3,46 @@
...
 import MuiTabs from '@material-ui/core/Tabs';
 import Skeleton from '@material-ui/lab/Skeleton';
 import MuiTab from '@material-ui/core/Tab';
+import { NAVBAR_HEIGHT } from '../constants';
 import { FlexStart } from './Layout';
 
 const LoadingTab = () => <Skeleton width={100} height={20} style={{ marginLeft: 30 }} />;
 
-export const TabBar = styled(MuiContainer)`
+const TabBarOuterContainer = styled.div`
```
tcaiger, line 18: As far as I can tell this component is not needed and the styles could just be combined with `StickyTabBarOuterContainer`

packages/lesmis/src/components/Tabs.js
```diff
@@ -3,25 +3,46 @@
...
+const TabBarInnerContainer = styled(MuiContainer)`
   display: flex;
   align-items: center;
   justify-content: flex-start;
-  border-top: 1px solid ${props => props.theme.palette.grey['400']};
-  border-bottom: 1px solid ${props => props.theme.palette.grey['400']};
 `;
 
+// sticks on scroll, sitting below the nav bar
+export const StickyTabBar = forwardRef(({ children }, ref) => (
```
tcaiger, line 37: I would prefer to get rid of this component and move the 2 containers inline in `DashboardReportTabView`. It makes sense if we had an encapsulated sticky tab bar component but at the moment most of the code for it is in the tab view any way. This keeps the tabs components free of the sticky logic, removes the forwardRef which could be confusing and removes the need to look between files for parts of the behaviour.

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -14,18 +14,41 @@ import {
...
+`;
+
+const ScrollToTopButtonImg = styled.img`
+  width: 40px;
+  height: 40px;
+`;
+
+const ScrollToTopButton = React.memo(({ onClick }) => (
+  <ScrollToTopButtonContainer onClick={onClick}>
+    <ScrollToTopButtonImg src="/images/up-arrow.svg" alt="back to top" />
```
tcaiger, line 45: I would usually prefer to use the [material-ui icons](https://material-ui.com/components/material-icons/#material-icons) rather than adding a new image. We will just need to add a box around the icon to make it look the same as the design. An extra advantage of this is then you can control the icon on interactions (eg the color on hover)

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -14,18 +14,41 @@ import {
...
   YearSelector,
 } from '../components';
-import { DEFAULT_DASHBOARD_GROUP, DEFAULT_DATA_YEAR } from '../constants';
+import { DEFAULT_DASHBOARD_GROUP, DEFAULT_DATA_YEAR, NAVBAR_HEIGHT_INT } from '../constants';
 
 const DashboardSection = styled(FlexCenter)`
   min-height: 31rem;
 `;
 
+const ScrollToTopButtonContainer = styled.div`
```
tcaiger, line 29: Sorry I should have picked this up when we chatted. I'm not that familiar with position: sticky yet. Position sticky makes sense for the navbar and toolbar but I think position fixed makes more sense here. The advantage of position sticky is that it turns on and off position fixed which we are doing manually ourselves in this instance. And the advantage of using position fixed is that we can set the position in the x-axis (eg. right: 32px) and avoid having to add the flex container etc.

packages/lesmis/src/views/DashboardReportTabView.js
tcaiger: Yeah that sounds good. It makes sense to keep the code in here at the moment since it's custom behaviour relating to this component.

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -14,18 +14,41 @@ import {
...
+
+const ScrollToTopButton = React.memo(({ onClick }) => (
+  <ScrollToTopButtonContainer onClick={onClick}>
+    <ScrollToTopButtonImg src="/images/up-arrow.svg" alt="back to top" />
+  </ScrollToTopButtonContainer>
+));
+ScrollToTopButton.propTypes = {
+  onClick: PropTypes.func.isRequired,
+};
+
```
tcaiger, line 51:
```suggestion
const useRefHeight = () => {
  const [tabBarHeight, setTabBarHeight] = useState(0);

  const measureTabBarHeight = useCallback(tabBarNode => {
    if (tabBarNode !== null) {
      setTabBarHeight(tabBarNode.getBoundingClientRect().height);
    }
  }, []);

  return [tabBarHeight, measureTabBarHeight];
};

Even though it makes sense for the behaviour to be in with this component, it might still be nice to pull out the code for getting the height of the toolbar into a hook above just for readability reasons. I've put an example here but it would need to change a bit if the forward reffing was removed.
```

tcaiger: Even though it makes sense for the behaviour to be in with this component, it might still be nice to pull out the code for getting the height of the toolbar into a hook above just for readability reasons. I've put an example here but it would need to change a bit if the forward reffing was removed.

tcaiger (changes requested):
Nice front end action! I played around with it locally and seems to work perfectly. I've made a few suggestions mostly relating to organisation of code. I don't feel strongly about most of them tough so feel free to push back. Happy to discuss anything as well.

packages/lesmis/src/components/Tabs.js
edmofro: Yep, I was just aiming to keep the styling separate in case tabs are used elsewhere and don't want to be sticky at the bottom of the nav bar

edmofro: Happy to change if you think it's a case of YAGNI

edmofro: But your other suggestion is a much better way to do it - pull the sticky stuff fully out

packages/lesmis/src/components/Tabs.js
edmofro: This was a great idea - it's heaps nicer now :-)

packages/lesmis/src/views/DashboardReportTabView.js
edmofro: Good call - changed :-)

packages/lesmis/src/views/DashboardReportTabView.js
edmofro: Oh yeah that was a lot easier haha

packages/lesmis/src/views/DashboardReportTabView.js
edmofro: Agreed - it was getting a bit too complex in that component and this was a nice option for pulling out. Done.

tcaiger (approved):
Lookn good!

PR #2508, opened 2021-05-05
2501 bugfixes entity vitals

EMcQ-BES:
### Changes:

- Default value for sub district report
- Hard code country vitals


packages/lesmis/src/views/DashboardView.js
```diff
@@ -182,13 +182,13 @@ const CountryView = ({ vitals }) => {
         </TitleContainer>
         <EntityVitalsItem
           name="No. Schools"
-          value={vitals.NumberOfSchools}
+          value="13849"
```
edmofro, line 185: Could you add todo comments next to these referencing the new tickets

edmofro (approved):
Looks good - just some comments to add but approved

PR #2509, opened 2021-05-05
Minimap mvp

edmofro:
### Issue #:
Very MVP version of https://github.com/beyondessential/tupaia-backlog/issues/2741

Not as nice as Dave's shadow-background design, and isn't very helpful at school level

Screenshots https://github.com/beyondessential/tupaia-backlog/issues/2741#issuecomment-832402392


packages/lesmis/src/components/index.js
```diff
@@ -3,22 +3,23 @@
...
 export * from './Breadcrumbs';
-export * from './MapOverlaysPanel';
-export * from './LocationHeader';
-export * from './SearchBar';
 export * from './EntityMenu';
-export * from './EntityVitalsItem';
-export * from './PartnerLogo';
 export * from './EntityPolygonLink';
-export * from './Tabs';
+export * from './EntityVitalsItem';
```
tcaiger, line 9: 🙏

packages/lesmis/src/components/MiniMap.js
```diff
@@ -0,0 +1,38 @@
...
+import { useEntityData } from '../api';
+
+const TILE_SET_URL = TILE_SETS.find(t => t.key === 'satellite').url;
+
+const Map = styled(MapContainer)`
+  width: 510px;
+  height: 370px;
+`;
+
+export const MiniMap = ({ entityCode, theme }) => {
```
tcaiger, line 25: Just need to add theme to prop types

packages/lesmis/src/components/MiniMap.js
```diff
@@ -0,0 +1,38 @@
...
+  const { data: entityData, isFetching: isLoadingEntityData } = useEntityData(entityCode);
+
+  return isLoadingEntityData ? null : (
+    <Map bounds={entityData?.bounds} dragging={false} zoomControl={false}>
+      <TileLayer tileSetUrl={TILE_SET_URL} />
+      {entityData?.region && <InversePolygonMask region={entityData?.region} />}
+      {entityData?.point && <IconMarker coordinates={entityData?.point} color={RED} />}
```
tcaiger, line 32: Nice! I like that we have both the pin and mask options

packages/lesmis/src/components/MiniMap.js
```diff
@@ -0,0 +1,38 @@
...
+
+const TILE_SET_URL = TILE_SETS.find(t => t.key === 'satellite').url;
+
+const Map = styled(MapContainer)`
+  width: 510px;
+  height: 370px;
+`;
+
+export const MiniMap = ({ entityCode, theme }) => {
+  const { data: entityData, isFetching: isLoadingEntityData } = useEntityData(entityCode);
```
tcaiger, line 26: Is there a reason why you went for `isFetching` rather than `isLoading`. I would have chosen isLoading because we only really want to show null if there is absolutely not data yet while between fetches (which could happen for instance when clicking breadcrumbs) it might be nicer to keep the map in place and just change the data (I think that would result in a nice map animation)

packages/ui-components/src/components/Map/InversePolygonMask.js
```diff
@@ -0,0 +1,45 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
```
tcaiger, line 6: Great component - love it

packages/lesmis/src/api/queries/useEntityData.js
```diff
@@ -8,11 +8,19 @@ import { get } from '../api';
...
+      return get(`entity/${entityCode}`);
+    },
+    {
+      staleTime: 1000 * 60 * 60 * 1,
+      refetchOnWindowFocus: false,
+      // pre-populate the query data from the cached entities data if it's available
+      initialData: () =>
+        queryClient
+          .getQueryData('entities')
+          ?.find(entity => entity.code === entityCode && !!entity.bounds), // check the entity exists and isn't "lean"/geo-data free
```
tcaiger, line 23: Good catch - I remembered this yesterday and was going to mention it. It would be better if possible to solve this by relying on the cache keys. It would avoid us having to add this kind of logic in a few places and also it would ensure maximum use of the cache. In this instance, we are missing the cache benefit in places that just want the simple data (eg. the breadcrumbs)

tcaiger (changes requested):
Super quick work and it looks great. If you can add the propType and have a look at the loading state then I am happy to approve it. It would be great to have a look at the caching but it might make sense to split that out for later if we don't want to hold this up.

tcaiger (approved):

packages/lesmis/src/components/MiniMap.js
edmofro: Sorry this was just me not knowing my way around react-query! That's actually quite a cool distinction

packages/lesmis/src/components/MiniMap.js
```diff
@@ -19,7 +19,8 @@ const TILE_SET_URL = TILE_SETS.find(t => t.key === 'satellite').url;
 
 const Map = styled(MapContainer)`
   width: 510px;
-  height: 370px;
+  min-height: 370px;
+  height: auto;
```
edmofro, line 23: Oh nice!

edmofro:
Thanks for polishing this off @tcaiger 🙏

PR #2510, opened 2021-05-05
2685 covid samoa percentage reports

Alastair-L:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/2685

Added dashboard reports at 3 org unit levels.

Using the dataBuilder `calcPerSeries` really doesn't match the mental model of what is going on here (there is no series!), but I didn't want to add a new dataBuilder if I could avoid it...


rohan-bes:
> doesn't match the mental model of what is going on here (there is no series!), but I didn't want to add a new dataBuilder if I could avoid it...

I've been there too buddy 😢

rohan-bes (approved):
Nice work, really appreciate those comments btw, nice to be clear about such things.

packages/database/src/migrations/20210505101731-AddSamoaCovidPercentageDashboards-modifies-data.js
```diff
@@ -0,0 +1,110 @@
...
+const getDataBuilderConfig = doseNum => ({
+  series: {
+    value: { // The value of this key (value) IS important
+      xyz: { // The value of this key (xyz) is not important
+        operands: [
+          {
+            dataValues: [doseNum === 1 ? 'COVIDVac4' : 'COVIDVac8'],
+          },
+          {
+            dataValues: ['population_WS001'],
```
Alastair-L, line 34: Just wanted to note: this will sum all data for this dataElement (but it shouldn't matter as long as that is kept in mind if the population is updated)

PR #2513, opened 2021-05-06
2707 Dashboard Number of passengers per flight

billli0:
### Issue #:
[2707 Dashboard Number of passengers per flight](https://github.com/beyondessential/tupaia-backlog/issues/2707)
### Changes:
- Auto eslint formating 
- New custom dataBuilder `totalNumOfPassengerByFlightDate`
- Migration

### Screenshots:
[Screenshots](https://github.com/beyondessential/tupaia-backlog/issues/2707#issuecomment-833171991)


packages/web-config-server/src/apiV1/dataBuilders/modules/covid-samoa/valueAndPercentageByDataValueByFlightDate.js
```diff
@@ -27,70 +27,78 @@ export class ValueAndPercentageByDataValueByFlightDate extends DataBuilder {
...
+    flights.sort((a, b) => (a.date < b.date ? -1 : 1));
 
     for (const flight of flights) {
-      const formattedDate = moment(flight.date).format('MMM D') + ' Flight';
+      const formattedDate = moment(flight.date).format('MMM D');
       columns.push({
         key: flight.key,
-        title: `Number (%): ${formattedDate}`,
+        title: `Number (%): ${formattedDate} Flight`,
       });
```
billli0, line 48: Do a bit manual formating

packages/web-config-server/src/apiV1/dataBuilders/modules/covid-samoa/totalNumOfPassengerByFlightDate.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+import moment from 'moment';
+import { DataBuilder } from '/apiV1/dataBuilders/DataBuilder';
+import { Flight, getTotalNumPassengers, FLIGHT_DATE } from './flight';
+
+export class TotalNumOfPassengerByFlightDate extends DataBuilder {
+  async build() {
+    const dataElementCodes = [FLIGHT_DATE];
+    const { startDate, endDate } = this.query;
+    this.query.startDate = undefined;
```
nick-beyondessential, line 13: Out of curiosity why are these being set to undefined?

billli0: Sorry I forgot to comment that, usually when fetching result, `.fetchEvents()` will use startDate and endDate as a period to fetch events.

But here is a special case that, we want all events sending back as result, and then filter by selected period on `FLIGHT_DATE`, because `FLIGHT_DATE` is not the response time, is a data element.

nick-beyondessential: Fair enough, assigning unassigned feels strange to me, but as this is a custom builder in the covid model should be ok if it works.

nick-beyondessential: Maybe add a comment?

nick-beyondessential (approved):
Looks good to me, nice work @billli0

packages/web-config-server/src/apiV1/dataBuilders/modules/covid-samoa/totalNumOfPassengerByFlightDate.js
billli0: Yeah good point, will leave a comment there.

PR #2515, opened 2021-05-06
Mini-map updates

edmofro:
- Fixes entity polygon not matching Dave's design
- Fixes flicker between map and then photo for facilities with photos
- Improves pin used throughout LESMIS to not have the "health" cross
- Cleans up some peripherally related code


packages/lesmis/src/components/MiniMap.js
```diff
@@ -24,14 +25,38 @@ const Map = styled(MapContainer)`
   height: auto;
 `;
 
+const BasicPolygon = styled(Polygon)`
+  fill: ${RED};
```
tcaiger, line 29:
```suggestion
fill: ${props => props.theme.palette.primary.main};
```

This is totally an optional change. Just thought I'd point out that it is possible to use the main theme color.

packages/lesmis/src/components/MiniMap.js
```diff
@@ -24,14 +25,38 @@ const Map = styled(MapContainer)`
...
+const PointMarker = ({ entityData }) =>
+  entityData?.point && <IconMarker coordinates={entityData?.point} color={RED} scale={1.5} />;
+/* eslint-enable react/prop-types */
+
 export const MiniMap = ({ entityCode }) => {
+  const { data: countryData, isLoading: isLoadingCountryData } = useEntityData(COUNTRY_CODE);
   const { data: entityData, isLoading: isLoadingEntityData } = useEntityData(entityCode);
 
-  return isLoadingEntityData ? null : (
+  return isLoadingCountryData || isLoadingEntityData ? null : (
```
tcaiger, line 54: It would be nice if you wanted to, to return a loading state instead of null. Maybe a very light grey box with the same minimum dimensions as the map or a skeleton loader. It would be nice to improve the height change that's happening when the data comes in.

edmofro: Well! Interesting you say that...I played around with this, but it seems all of the nice entity vitals loading skeletons have been broken by some change. It seems that nothing shows up (not even loaders) until useEntityData finishes loading...or something like that. So I pulled my mini-map skeleton loader back out as I couldn't easily tell if it was doing anything. @EMcQ-BES could you look at whether the individual vitals items loading states is still visible?

Admittedly I gave up pretty quickly as it's been hard to fit in dev time today!

packages/lesmis/src/views/DashboardView.js
```diff
@@ -178,6 +178,14 @@ const VerticalDivider = styled(MuiDivider)`
   height: 7rem;
 `;
 
+const PhotoOrMap = ({ vitals }) => {
+  if (vitals.isLoading) return null;
```
tcaiger, line 182: Maybe better to place the loader here instead of the mini-map (or as well as in the mini-map)

packages/lesmis/src/views/DashboardView.js
```diff
@@ -360,7 +368,7 @@ const VillageView = ({ vitals }) => {
           isLoading={vitals.isLoading}
         />
       </VitalsSection>
-      <MiniMap entityCode={vitals.code} />
+      <PhotoOrMap vitals={vitals} />
```
tcaiger, line 371: Is it right that we will always show the photo over the map if it exists? Even for districts?

edmofro: I reckon so! I thought about this and decided...if someone's decided that any given entity, of any type, deserves a photo, we should show it

packages/lesmis/src/views/DashboardView.js
edmofro: Yep that's actually where I put it

edmofro: (when I had it)

packages/ui-components/src/components/Map/index.js
```diff
@@ -3,13 +3,14 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  *
  */
+export { Polygon } from 'react-leaflet';
```
tcaiger, line 6: Unless I'm missing something, we can delete either this line or the Polygon.js file. I don't mind which one goes.

packages/lesmis/src/components/MiniMap.js
EMcQ-BES: I would presume that nothing shows up because we use `entity.type` to pick which dashboard/vitals we're going to be displaying

edmofro: Right, that'll be it! So I think that's enough motivation to re-look at changing the useEntityData call to specify whether geo data is required, and if not, initialise from the full entities list. @EMcQ-BES could you make a card?

packages/ui-components/src/components/Map/index.js
edmofro: Oh so this was weird...I tried importing `Polygon` from `react-leaflet` directly inside lesmis and everything broke! I could only find one other example of that being done and it turned out the file wasn't being used (which is why I deleted it)

edmofro: Again, something I could put more time into, but it also seemed kinda nice to not have lesmis directly import `react-leaflet` when it's getting all the other map stuff from `ui-components`

packages/ui-components/src/components/Map/Markers/markerIcons.js
```diff
@@ -285,15 +301,29 @@ export const LEGEND_SHADING_ICON = 'square';
 export const LEGEND_RADIUS_ICON = 'radius';
 export const HIDDEN_ICON = 'hidden';
 
+const scAnchor = (anchor = [0.5 * ICON_BASE_SIZE, 0.5 * ICON_BASE_SIZE], scale = 1) => {
```
tcaiger, line 304: I can't find where this function is being used? Is it still needed?

edmofro: Ooops 👻

packages/lesmis/src/components/MiniMap.js
```diff
@@ -24,14 +25,38 @@ const Map = styled(MapContainer)`
...
+
+const RegionPolygon = ({ entityData }) => {
+  if (!entityData?.region) return null;
+  if (entityData?.type === 'country') return null; // country is fine without, as it has the mask
+
+  return <BasicPolygon positions={entityData?.region} interactive={false} />;
+};
+
+const PointMarker = ({ entityData }) =>
+  entityData?.point && <IconMarker coordinates={entityData?.point} color={RED} scale={1.5} />;
```
tcaiger, line 47: I'm not sure if you know, but it is possible to scale the icons on the map using css by targeting the mapbox svg and using `transform: scale(10)`. There is also a `transform-origin` property to let you control the point of scaling. It still may not work for all the scaling scenarios you added to ui-components but if it was just here it should work.

packages/lesmis/src/components/MiniMap.js
```diff
@@ -24,14 +25,38 @@ const Map = styled(MapContainer)`
...
+/* eslint-enable react/prop-types */
+
 export const MiniMap = ({ entityCode }) => {
+  const { data: countryData, isLoading: isLoadingCountryData } = useEntityData(COUNTRY_CODE);
   const { data: entityData, isLoading: isLoadingEntityData } = useEntityData(entityCode);
 
-  return isLoadingEntityData ? null : (
+  return isLoadingCountryData || isLoadingEntityData ? null : (
     <Map bounds={entityData?.bounds} dragging={false} zoomControl={false}>
       <TileLayer tileSetUrl={TILE_SET_URL} />
```
tcaiger, line 56: I would vote for not-showing the attribution and mapbox link on the mini-map because I think it looks messy. I feel that since we are showing it on the main map we could get away with it. It could be turned off by a hidden prop or we could just use the leaflet tilelayer directly which doesn't have it.

packages/lesmis/src/components/MiniMap.js
tcaiger: I think it could be solved without that with a bit of shuffling around. So could we make the ticket to polish the loading states and we could try both options. @EMcQ-BES maybe we could pair programme on it?

EMcQ-BES: https://github.com/beyondessential/tupaia-backlog/issues/2772

packages/lesmis/src/components/MiniMap.js
edmofro: Oh right! I did not know that. I figure given that the `scale` prop exists on `IconMarker`, we should make it work - but perhaps even within ui-components I could simplify how it's working. Might take a look at that on Monday.

packages/ui-components/src/components/Map/index.js
tcaiger: I agree that's a good call keeping leaflet out of lesmis (for now anyway) and thanks for deleting that file. I still think that `ui-components/Map/Polygon.js` could be deleted right?

packages/lesmis/src/components/MiniMap.js
EMcQ-BES: @tcaiger Sure we can look into that! I don't have full context for this, I just answered the question I was asked 😅

tcaiger (approved):
Great to see the new pins in lesmis instead of the medical ones! I'm happy for you to decide on whether to show the attribution or not. Otherwise I think everything else has been covered off 👍

packages/ui-components/src/components/Map/index.js
edmofro: Oh shoot I left that in huh...I see what you mean now! Sorry

packages/lesmis/src/components/MiniMap.js
edmofro: It does look messy!

edmofro: Removed, though it needed css applied at the MapContainer level because of the way leaflet structures the DOM.

PR #2516, opened 2021-05-06
2655 create covid samoa percentage measures

Alastair-L:


packages/database/src/migrations/20210506052033-AddSamoaCovidPercentageVaccinatedMeasures-modifies-data.js
```diff
@@ -0,0 +1,127 @@
...
+        dataElementCodes: [dataElementCode],
+        entityAggregation: {
+          dataSourceEntityType: 'village',
+        },
+      },
+    },
+    denominator: {
+      measureBuilder: 'sumAllPerOrgUnit',
+      measureBuilderConfig: {
+        dataElementCodes: ['population_WS001'],
```
Alastair-L, line 38: Just wanted to note: this will sum all data for this dataElement (but it shouldn't matter as long as that is kept in mind if the population is updated)

packages/database/src/migrations/20210506052033-AddSamoaCovidPercentageVaccinatedMeasures-modifies-data.js
```diff
@@ -0,0 +1,127 @@
...
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const getMeasureBuilderConfig = dataElementCode => ({
+  aggregations: [
+    {
+      type: 'FINAL_EACH_DAY',
+    },
+  ],
```
chris-bes, line 24: I might be wrong, but from looking at `composePercentagePerOrgUnit` measure builder and `fetchComposedData`, I think this `aggregations` config will be ignored if you put it at the top level config like this (see `fetchComposedData()` in measureBuilders/helpers.js`. Instead, I think that you have to put this `aggregations` config in each of the sub measure builder config (numerator and denominator)

chris-bes:
LGTM apart from what I mentioned!

PR #2517, opened 2021-05-06
Add global loading indicator to tupaia

EMcQ-BES:
### Changes:

- Add `LoadingScreen` to `RootScreen` of `desktop`
- Move `LoadingScreen` out of `mobile` because it's now shared between `mobile` and `desktop`


packages/web-frontend/src/screens/desktop/RootScreen/index.js
```diff
@@ -27,10 +28,11 @@ import { OverlayContainer } from '../../../utils';
 import { TOP_BAR_HEIGHT } from '../../../styles';
 import './desktop-styles.css';
 
-export const RootScreen = ({ enlargedDialogIsVisible }) => {
+export const RootScreen = ({ enlargedDialogIsVisible, isLoading }) => {
   return (
     <div>
       {/* The order here matters, Map must be added to the DOM body after FlexContainer */}
+      <LoadingScreen isLoading={isLoading} />
```
edmofro, line 35: I'd suggest we put it further down, so that it is later in the render stack and would obscure the overlay/modal etc.

edmofro (approved):
One little suggestion, but looking good :-)

packages/web-frontend/src/screens/desktop/RootScreen/index.js
EMcQ-BES: I can move it if you like but the loading screen style has `zIndex: Number.MAX_SAFE_INTEGER` so it's always rendered on top regardless. I put it at the top to make that layout consistent between `mobile` and `desktop`

image

edmofro: Hmmm...I still think it's nice to telegraph the fact it'll be over the top of the modal to other developers

edmofro (approved):
lgtm :-)

PR #1423, opened 2020-09-30
Fix webpack config

kael89:
CI/CD picked that

PR #2522, opened 2021-05-07
1423 Put waiviva under right parent Fiji

billli0:
### Issue #:
[#1423](https://github.com/beyondessential/tupaia-backlog/issues/1423)
### Changes:
- Just a simple migration


packages/database/src/migrations/20210507014219-PutWaiviviaToTheRightPlace-modifies-data.js
```diff
@@ -0,0 +1,63 @@
...
+const EASTERN_CODE = 'FJ_Eastern';
+const projectCode = 'wish';
+
+exports.up = async function (db) {
+  const WAIVIVIA_ID = await codeToId(db, 'entity', WAIVIVIA_CODE);
+  const BURETA_ID = await codeToId(db, 'entity', BURETA_CODE);
+
+  await db.runSql(`
+    UPDATE entity 
+    SET parent_id = '${BURETA_ID}'
```
nick-beyondessential, line 35: I believe that parent_id here should be set to Fiji to match other Fiji villages?

nick-beyondessential: I.e all villages are directly under Fiji for the historical reason mentioned here: https://github.com/beyondessential/tupaia-backlog/issues/1423#issuecomment-722877076
And set placed in a catchment for wish vis the entity_relation below.
This will keep them all consistent for when updated later.

packages/database/src/migrations/20210507014219-PutWaiviviaToTheRightPlace-modifies-data.js
```diff
@@ -0,0 +1,63 @@
...
+    WHERE code = '${WAIVIVIA_CODE}';
+  `);
+
+  await insertObject(db, 'entity_relation', {
```
nick-beyondessential, line 39: Any tiny bit of pickiness, if hierarchyNameToId doesn't find the names for whatever reason then this will insert a row with nulls or turn an error. Not a big deal if you know the entries are there but just thought Id point it out.

packages/database/src/migrations/20210507014219-PutWaiviviaToTheRightPlace-modifies-data.js
billli0: Thanks Nick for this catching! I see what you mean. I will change the parent_id to Fiji.

packages/database/src/migrations/20210507014219-PutWaiviviaToTheRightPlace-modifies-data.js
billli0: Yeah I agree, will add an IF statement for safer inserting.

packages/database/src/migrations/20210507014219-PutWaiviviaToTheRightPlace-modifies-data.js
```diff
@@ -0,0 +1,68 @@
...
+const BURETA_CODE = 'FJ_CATCH_Bureta';
+const FIJI_CODE = 'FJ';
+const EASTERN_CODE = 'FJ_Eastern';
+const projectCode = 'wish';
+
+exports.up = async function (db) {
+  const WAIVIVIA_ID = await codeToId(db, 'entity', WAIVIVIA_CODE);
+  const BURETA_ID = await codeToId(db, 'entity', BURETA_CODE);
+  const FIJI_ID = await codeToId(db, 'entity', FIJI_CODE);
+
```
nick-beyondessential, line 34: I was more thinking that the village and catchment were key to the input below if not found it would insert a row to the entity_relation table with child_id: null for example do something like this:

```suggestion
if 
throw new Error('MapOverlayGroup not found');
```

billli0: Good idea. My only concern is that, does throwing error will stop running the following migrations? If so, instead of throwing error, I would skip the sql.insert if they are null id.

nick-beyondessential: I will leave up to you. My thinking is that if the parent isn't found then the migration has failed and would be good to know that rather than a silent fail. But as this is a small change and easily checked i don't want to be too nick picky :)

nick-beyondessential (approved):
LGTM

packages/database/src/migrations/20210507014219-PutWaiviviaToTheRightPlace-modifies-data.js
billli0: Thanks @nick-beyondessential for raising this question in dev channel, I didn't think of it before. Throwing error in migrations makes more sense to me now.

PR #2527, opened 2021-05-10
Allow Legacy Options to be edited as a json text array 2671

EMcQ-BES:
### Changes:

- Allow the `JsonEditor` to pass its data as either a string or a JSON object
- Allow `Legacy Options` to be edited as a json text array

See the linked issue for recording of this working


packages/admin-panel/src/widgets/InputField.js
```diff
@@ -100,6 +100,18 @@ export const InputField = ({
         />
       );
       break;
+    case 'textArray':
```
edmofro, line 103: Doesn't really matter that the array contents are text, does it? Maybe it does... But couple of options if it could work for, e.g., `[1, 2, 3]`
```suggestion
    case 'array':
```
```suggestion
    case 'jsonArray':
```

edmofro (approved):
Solution lgtm! Just a question on a name, but approved.

packages/admin-panel/src/widgets/InputField.js
EMcQ-BES: Hmm, lemme do a quick experiment with other types and I'll update this if they work correctly

EMcQ-BES: Yep this works for other types, I've changed the named to `jsonArray`

edmofro (approved):

PR #2539, opened 2021-05-11
Allow a deployment to be "protected" from lambda

edmofro:
@tcaiger had the idea that we should make sure no one can accidentally delete production using lambda. I've attached the "Protected": "true" tags to the three key production EC2 instances, and already deployed the code in this PR to lambda


kael89 (approved):
Looks good!

PR #2540, opened 2021-05-11
New dashboard structure and endpoint

EMcQ-BES:
### Changes:

- Add migration for new dashboard table structure


packages/database/src/migrations/20210511024543-CreateDashboardItemsTableLayout-modifies-schema.js
```diff
@@ -0,0 +1,66 @@
...
+
+exports.up = async function (db) {
+  await db.runSql(`
+    CREATE TABLE dashboard_item (
+      id TEXT PRIMARY KEY,
+      code TEXT NOT NULL UNIQUE,
+      config JSONB NOT NULL DEFAULT '{}',
+      permission_groups TEXT[] NOT NULL CONSTRAINT permission_groups_not_empty CHECK (permission_groups <> '{}'),
+      report_code TEXT NOT NULL,
+      FOREIGN KEY (report_code) REFERENCES report (code) ON UPDATE CASCADE ON DELETE RESTRICT
```
EMcQ-BES, line 25: Self admittedly these column constraints are perhaps overly restrictive, I'm not sure if we have a particular way of deciding on column constraints, but this is built around my understanding of the intended usage

packages/database/src/migrations/20210511042757-MoveLESMISVisualsToNewDashboards-modifies-data.js
```diff
@@ -0,0 +1,100 @@
...
+                  THEN 'sub_district'
+                  ELSE LOWER("organisationLevel")
+                END) AS entity_types,
+      array_cat_agg(distinct "projectCodes") AS project_codes,
+      array_agg(distinct "userGroup") AS base_permission_groups,
+      concat("organisationUnitCode", '_', REPLACE(name, ' ', '_'), '_', REPLACE((array_agg(distinct "userGroup"))[1], ' ', '_')) as code
+    FROM "dashboardGroup"
+    WHERE 'laos_schools' = ANY("projectCodes")
+    GROUP BY name, "organisationUnitCode";
+  `);
```
EMcQ-BES, line 51: These SQL commands are a bit complicated because they're performing the full data conversion (aside from generating an id)

I've added screenshots to the ticket showing the result of running this migration just to show that these complicated commands are working.

packages/lesmis/src/components/EntityVitalsItem.js
```diff
@@ -102,7 +102,7 @@ export const EntityVitalsItem = ({ name, value, icon, isLoading }) => (
 
 EntityVitalsItem.propTypes = {
   name: PropTypes.string.isRequired,
-  value: PropTypes.string,
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
EMcQ-BES, line 105: This just fixes a console error in lesmis, just sneaking it in alongside other work

packages/lesmis/src/views/DashboardView.js
```diff
@@ -489,7 +489,7 @@ export const DashboardView = () => {
               case 'school':
                 return <SchoolView vitals={vitals} />;
               default:
-                return null;
+                return <></>;
```
EMcQ-BES, line 492: This also just fixes a console error

packages/database/src/migrations/20210511024543-CreateDashboardItemsTableLayout-modifies-schema.js
```diff
@@ -0,0 +1,70 @@
...
+    );
+  `);
+  await db.runSql(`
+    CREATE TABLE dashboard_relation (
+      id TEXT PRIMARY KEY,
+      dashboard_id TEXT NOT NULL,
+      child_id TEXT NOT NULL,
+      sort_order INTEGER,
+      FOREIGN KEY (dashboard_id) REFERENCES dashboard (id) ON UPDATE CASCADE ON DELETE RESTRICT,
+      FOREIGN KEY (child_id) REFERENCES dashboard_item (id) ON UPDATE CASCADE ON DELETE RESTRICT
```
edmofro, line 54:
```suggestion
      FOREIGN KEY (child_id) REFERENCES dashboard_item (id) ON UPDATE CASCADE ON DELETE CASCADE
```
If deleting a dashboard item, I think it's fine to just wipe all its relations rather than having to check whether it's been used anywhere first

packages/database/src/modelClasses/DashboardItem.js
```diff
@@ -0,0 +1,35 @@
...
+    return DashboardItemType;
+  }
+
+  async fetchItemsInDashboard(dashboardId, userGroups, criteria) {
+    return this.find(
+      {
+        ...criteria,
+        dashboard_id: dashboardId,
+        permission_groups: {
+          comparator: '&&', // User has ANY of the permission groups
```
edmofro, line 25: Helpful comment, thanks!

packages/database/src/utilities/migration.js
```diff
@@ -15,7 +15,12 @@ class RequiredParameterError extends Error {
 }
 
 export const insertObject = async (db, table, data, onError) =>
-  db.insert(table, Object.keys(data), Object.values(data), onError);
+  db.insert(
+    table,
+    Object.keys(data),
+    Object.values(data).map(x => (Array.isArray(x) ? `{${arrayToDoubleQuotedDbString(x)}}` : x)),
```
edmofro, line 21: Please check no previous migration is passing an array in (maybe that's automatically checked because it just wouldn't have worked)

packages/web-config-server/src/apiV1/legacyReport.js
```diff
@@ -0,0 +1,35 @@
...
+    if (drillDownLevel) {
+      const drillDownCode = baseDashboardItem.config.drillDown?.itemCodes[drillDownLevel];
+      if (!drillDownCode) {
+        throw new Error(`No drill down exists for level ${drillDownLevel}`);
+      }
+      dashboardItem = await this.models.dashboardItem.findOne({
+        code: drillDownCode,
+      });
+    }
+    const legacyReport = await this.models.legacyReport.findOne({ id: dashboardItem.report_code });
```
edmofro, line 27: Could do with error messages if either the base dashboard item or the legacy report don't exist (or does that happen in the permissions checker?)

packages/web-config-server/src/apiV1/dashboardses.js
```diff
@@ -0,0 +1,49 @@
...
+    const hierarchyId = await this.fetchHierarchyId();
+    const returnArray = [];
+    // Fetch dashboards
+    const dashboards = await this.models.dashboard.getDashboards(
+      entity,
+      hierarchyId,
+      query.projectCode,
+    );
+    // Fetch dashboard items
+    await Promise.all(
```
edmofro, line 25:
```suggestion
    const returnArray = await Promise.all(
```
And then you can just return inside the map function rather than assign

edmofro: Also, descriptive variable names ftw
```suggestion
    const dashboardsAndItems = await Promise.all(
```

packages/web-config-server/src/apiV1/legacyReport.js
```diff
@@ -0,0 +1,35 @@
...
+  static PermissionsChecker = DashboardItemPermissionChecker;
+
+  fetchReport = async ({ itemCode, drillDownLevel }) => {
+    const baseDashboardItem = await this.models.dashboardItem.findOne({
+      code: itemCode,
+      legacy: true,
+    });
+    let dashboardItem = baseDashboardItem;
+    if (drillDownLevel) {
+      const drillDownCode = baseDashboardItem.config.drillDown?.itemCodes[drillDownLevel];
```
edmofro, line 19: This is a smart way of continuing to support the current front end behaviour! Love the way you've structured it with the inheritance too. 

My plan with drill down support was that the front end should change so that it just requests a different report, based on the drill down item code in the base config. I think that's a little cleaner than having it work the drill down out here while still using the base item code. 

The implication there is a slightly different structure for the migration, where instead of putting the item codes into an array on the base config, each config has just one item code it can drill down to (so the second layer in has the code for the third layer in)

We don't need to put together the front end support for this now, but we probably do need to change the migration and delete this code. What do you think about that? Can you think of any downsides?

packages/database/src/migrations/20210511042757-MoveLESMISVisualsToNewDashboards-modifies-data.js
```diff
@@ -0,0 +1,177 @@
+'use strict';
```
edmofro, line 1: Haven't reviewed the SQL in here because it's fairly dense. In general, we prefer to pull the data out, manipulate it in javascript (which most of us are more familiar/comfortable with), and then put it back in. Because migrations are one-time, the performance isn't such an issue.

packages/lesmis/src/components/Report.js
```diff
@@ -125,6 +129,11 @@ export const Report = React.memo(
             dashboardGroupId={dashboardGroupId}
             reportId={reportId}
             periodGranularity={periodGranularity}
+<<<<<<< HEAD
```
edmofro, line 132: Some merge cruft in here

edmofro (commented):
Have done a draft/lightweight review across this. Awesome work @EMcQ-BES, it's all looking really good! And thanks again for putting up a draft PR :-)

packages/lesmis/src/components/Report.js
EMcQ-BES: Whoops 😳

packages/database/src/utilities/migration.js
EMcQ-BES: So looking back at this there _are_ a couple of situations where we pass an array in previous migrations. These are situations where the database column is a `jsonb` type, and therefore the stringified `[ x, y ,z ]` format is valid. But that format of course isn't valid for a column that is actually an array type `something[]` and that's when we need to convert to `{ x, y, z }` format.

This was a really good catch that I absolutely didn't consider. I think the best option is to just revert this change and run the `arrayToDoubleQuotedDbString` conversion in my individual migration.

edmofro: Sounds good, thanks for checking it out!

packages/database/src/modelClasses/Dashboard.js
```diff
@@ -0,0 +1,41 @@
...
+  static databaseType = TYPES.DASHBOARD;
+}
+
+export class DashboardModel extends DatabaseModel {
+  get DatabaseTypeClass() {
+    return DashboardType;
+  }
+
+  async getDashboards(entity, hierarchyId, projectCode, params) {
+    const ancestorCodes = await entity.getAncestorCodes(hierarchyId);
```
edmofro, line 20: This raises an interesting question about whether we want to continue allowing _any_ entity relationship access to exist outside of entity-server, or force this information to be fetched. It would probably fit better with our architecture to have the requester pass in the ancestors, rather than having the endpoint work that out. That does mean putting complexity in multiple clients rather than this one spot though.

This model is also structured really nicely to take care of accounting for the multiple semi-complicated filter rules in getting dashboards, which is exactly what a model should do.

Let's leave it how you've got it for now, but @rohan-bes it's a point for us to keep in mind about whether the entity-server should be the sole point of contact for entity relationship stuff. The benefit of that would be that we could pull it out into a proper micro-service that potentially even had its own database.

packages/web-config-server/src/apiV1/dashboardses.js
```diff
@@ -0,0 +1,63 @@
...
+  return [...sortedOutput, ...array];
+}
+
+export default class extends RouteHandler {
+  static PermissionsChecker = NoPermissionRequiredChecker;
+
+  buildResponse = async () => {
+    const { entity, query } = this;
+    const { code: entityCode } = entity;
+    const userGroups = await this.req.getUserGroups(entityCode);
```
edmofro, line 27:
```suggestion
    const permissionGroups = await this.req.getUserGroups(entityCode);
```
It might be too hard to make the `this.req.getUserGroups` consistent, but at least we can use `permissionGroups` everywhere new

packages/web-config-server/src/apiV1/dashboardses.js
```diff
@@ -0,0 +1,63 @@
...
+  // Sorts an array of objects by properties in a prioritized order
+  let sortedOutput = [];
+  columns.forEach(columnName => {
+    sortedOutput = sortedOutput.concat(array
+      .filter(item => !!item[columnName])
+      .sort((itemA, itemB) => itemA[columnName] > itemB[columnName] ? 1 : -1));
+    array = array.filter(item => !item[columnName]);
+  });
+  return [...sortedOutput, ...array];
+}
```
edmofro, line 19: A few things to fix up with this function:

1. Even if you know how it's being used outside the function, you should almost never mutate a function's parameters. It causes bugs because variables change opaquely from the caller's perspective
2. Try to use more descriptive names than `array`. What's in the array? 
3. It's called `sortByDbColumns`, but could work on properties other than db columns, so maybe a more generic name to reflect that
4. I haven't tested the logic properly, but reading through it looks like it filters all with one property defined first, then those with the secondary property defined second and so on, but ideally it would also do conflict resolution using the second property  (I suspect we'll end up with multiple dashboards at sort order `4`, for example)
5. There's no semi-colon on the end here, which indicates it hasn't been linted

packages/web-config-server/src/apiV1/dashboardses.js
```diff
@@ -0,0 +1,63 @@
...
+        const sortedDashboardItems = sortByDbColumns(dashboardItems, ['sort_order', 'code']);
+        return {
+          dashboardName: dashboard.name,
+          dashboardId: dashboard.id,
+          dashboardCode: dashboard.code,
+          entityType: entity.type,
+          entityCode,
+          entityName: entity.name,
+          items: Object.values(sortedDashboardItems).map(item => ({
+            itemCode: item.code,
```
edmofro, line 53:
```suggestion
            code: item.code,
```
I'd prefer to drop `item` prefix given it's a property _of_ an item.

packages/web-config-server/src/apiV1/dashboardses.js
```diff
@@ -0,0 +1,63 @@
...
+            legacy: item.legacy,
+            reportCode: item.report_code,
+            ...item.config,
```
edmofro, line 56: Hmmm...interested in the choice to spread here rather than just keep it inside a `config` key? It's probably the right way to go, but I'd like to hear your thoughts!

packages/database/src/migrations/20210511024543-CreateDashboardItemsTableLayout-modifies-schema.js
```diff
@@ -0,0 +1,71 @@
...
+      report_code TEXT NOT NULL,
+      legacy BOOLEAN NOT NULL DEFAULT 'false'
+    );
+  `);
+  await db.runSql(`
+    CREATE TABLE legacy_report (
+      id TEXT PRIMARY KEY,
+      "dataBuilder" TEXT,
+      "dataBuilderConfig" JSONB,
+      "dataServices" JSONB DEFAULT '[{"isDataRegional": true}]'
```
edmofro, line 33: I'd like this to be a cleaner table too, so let's add `code` and have `id` do its `uuid` thing (already mentioned), but also change the columns to snake case for consistency

packages/database/src/migrations/20210511042757-MoveLESMISVisualsToNewDashboards-modifies-data.js
edmofro: Still haven't reviewed this haha, will review after the other changes as some of them imply some migration updates

edmofro (changes requested):
Heaps of work in one PR! I'm pumped for these changes to get in, and then for Tupaia to update to them too, but have a few changes requested. Happy to talk through any of them

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -146,32 +146,36 @@ export const DashboardReportTabView = ({ entityCode, TabSelector }) => {
...
           {data &&
-            Object.entries(data).map(([heading, dashboardGroup]) => (
-              <TabPanel key={heading} isSelected={heading === selectedDashboard}>
-                {Object.entries(dashboardGroup).map(([groupName, groupValue]) => {
+            data.map(dashboard => (
+              <TabPanel
+                key={dashboard.dashboardName}
+                isSelected={dashboard.dashboardName === selectedDashboard}
+              >
+                {(() => {
```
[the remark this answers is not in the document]
EMcQ-BES: Inline js is expected to be a single resolvable statement, so since it's doing two steps (a filter followed by a ternary return) it needs to be wrapped. Previously it was wrapped in the `map` statement, but the changes in the layout of the `data` object mean that `map` is no longer necessary.

packages/web-config-server/src/apiV1/dashboardses.js
EMcQ-BES: This was mostly just to have the final output more consistent with the previous implementation.

packages/database/src/modelClasses/Dashboard.js
rohan-bes: Hmmm that's a good question @edmofro. So far, I've largely endeavoured to keep the 'entity relation' type logic that report-server deals with in the `EntityModel` (or `AncestorDescendantRelation` model) so that the entity-server itself isn't introducing new business rules about what relations exist between entities.

As such I think for now it's still fairly safe for code that has access to the `EntityModel` to query for relations and maintain consistent behaviour with the entity-server. This mightn't always be the case though...

The entity-server also handles permissions which the entity model does not... This may become even more apparent with our new 'Public data per application' system... I guess calls to `getAncestors()` should be safe as if you have a descendant already you should have access to all its ancestors, but the inverse is not true...

packages/web-config-server/src/apiV1/dashboardses.js
EMcQ-BES: I've just swapped this whole thing with a `lodash.orderby` because as I made it smarter I realised I was reinventing the wheel.

packages/lesmis/src/components/Report.js
```diff
@@ -109,7 +113,7 @@ export const Report = React.memo(
         </Header>
         <Body>
           <ChartTable
-            viewContent={viewContent}
+            viewContent={{ ...viewConfig, data: viewContent }}
```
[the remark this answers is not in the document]
EMcQ-BES: https://github.com/beyondessential/tupaia-backlog/issues/2923

chris-bes:
Hi @EMcQ-BES,

I'm just snooping into this PR a bit to get prepared for https://github.com/beyondessential/tupaia-backlog/issues/2814. All looking good and I'm very excited for this change!

I have only 1 concern. Before, with the same dashboard name (say `ICT Facilities`), in the database, we actually have multiple dashboard group records for different org unit levels. With the new dashboard structure, as I understand, we only have 1 dashboard group record with a column `entity_types` that can contain multiple org unit levels. So the dashboard will be shown at those specified entity levels in `entity_types` column with the child dashboard items.

However, we have a common scenario in Tupaia (not sure about LESMIS) when there are some dashboard items that should only be available at certain organisation unit level (eg: Dashboard Item `Laos_Schools_ICT_Facilities` of dashboard group `ICT Facilities` is only available at School level). Maybe I'm missing something but as I've gone through the code, I'm not sure if the new dashboard structure can handle the scenario. I think the endpoint will always return all the child dashboard items. So if you open `ICT Facilities` dashboard group at Country level Laos, the dashboard item `Laos_Schools_ICT_Facilities` that was only meant for School level will also be shown.

So if the scenario is not supported yet and if I can suggest, I think we should move the `entity_types` (and maybe also `project_codes`) columns from `dashboard` to `dashboard_item`. That way I think we can support the scenario above. Also, when users don't have access to any dashboard items, the endpoint just won't return the dashboard to the client.

edmofro:
Great snooping and find @chris-bes, and excellent suggestion too 🙏 

@EMcQ-BES could you follow up on Chris' suggestion to move the entity types and project codes to dashboard item? 

@chris-bes are you blocked in the meantime?

edmofro:
Thanks @chris-bes. 

@EMcQ-BES an update after chatting with Chris on Slack - he'll make two branches off this, one to update the entity types etc. to live on the dashboard item, and the other (off that) to update Tupaia to use this endpoint. Hopefully you don't have too many pending local changes!

We can then test all of the requirements together on a deployment based on Chris' final branch

Let us know when this is ready for re-review, but Chris will go ahead with implementing stuff off the end of it now anyway.

chris-bes:
Sorry @edmofro and @EMcQ-BES , I've just realized another scenario when starting the implementation today.

There are some generic dashboard reports that are used in multiple countries (especially those in `General` dashboard). And some of them are available in different organisation unit levels between countries (Real example: `Average Availability of Medicines across Operational Facilities` dashboard report is available in Kiribati at Country and SubDistrict levels, but only available in Tonga at Country level).

So having the permission columns (entity_types, project_codes, permission_groups) in `dashboard_item` is not ideal anymore because it won't be able to support such case above (because dashboard item's entity types will be applied the same to every dashboard it is in). With such case, I think dashboard_relation is now the best place to store the permission info.

edmofro:
Darn, good catch @chris-bes 

Storing the filter on the `dashboard_relation` record sounds like the most flexible solution. If it adds too much complexity though, feel free to just duplicate dashboard items in those few places where we have them available at different entity types per country, or even align those places where there are inconsistencies (they're probably a mistake anyway!)

packages/database/src/modelClasses/DashboardItem.js
```diff
@@ -0,0 +1,35 @@
...
+class DashboardItemType extends DatabaseType {
+  static databaseType = TYPES.DASHBOARD_ITEM;
+}
+
+export class DashboardItemModel extends DatabaseModel {
+  get DatabaseTypeClass() {
+    return DashboardItemType;
+  }
+
+  async fetchItemsInDashboard(dashboardId, userGroups, criteria) {
```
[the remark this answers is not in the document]
edmofro: I did check that note to self btw, but forgot to remove it from the PR comments

packages/database/src/migrations/20210511042757-MoveLESMISVisualsToNewDashboards-modifies-data.js
```diff
@@ -0,0 +1,181 @@
...
+      ON dashboard.name = "dashboardGroup".name
+    WHERE 'laos_schools' = ANY("projectCodes")
+    GROUP BY dashboard.id, child_id;
+  `);
+
+  for (const currentDashboardRelation of dashboardRelations.rows) {
+    const childId = await db.runSql(`
+      SELECT id FROM dashboard_item
+      WHERE code = '${currentDashboardRelation.child_id}'
+    `);
```
edmofro, line 129:
```suggestion
    const { id: childId } = (await db.runSql(`
      SELECT id FROM dashboard_item
      WHERE code = '${currentDashboardRelation.child_id}'
    `))[0];
```
This is the sort of thing I'd be happy to pull out into a migration util called `findOne`, which can be imported (because it doesn't make any assumptions about our database structure, so shouldn't change over time and affect past migrations)

edmofro: Oh and my reasoning here is I prefer variables to be named with no surprises, and I was surprised that something called `childId` held a structure like `[{ id: 'xxx', ... }]`

packages/database/src/migrations/20210511042757-MoveLESMISVisualsToNewDashboards-modifies-data.js
```diff
@@ -0,0 +1,181 @@
...
+  }
+};
+
+exports.up = async function (db) {
+  await createDashboardsFromDashboardGroups(db);
+  await createLegacyDashboardItems(db);
+  await createDashboardRelationsFromDashboardGroups(db);
+};
+
+exports.down = async function (db) {
```
edmofro, line 144: This is quite a prudent `down`, nicely done!

edmofro (changes requested):
Have now reviewed that dense sql migration, so just a couple of requests there

edmofro:
@EMcQ-BES could you also update from dev and resolve those conflicts? Once you've done that and made those changes, this is a good PR for a "walk and talk" before we hand it over to Chris :-)

chris-bes:
So sorry @EMcQ-BES @edmofro , I didn't know the migration hasn't been reviewed. With the plan of moving the filters to `dashboard_relation` and also processing drillDownLevel, I'm likely going write a new migration that pull the data out and manipulate with javascript which will be easier for me. So I think you don't have to update the migration @EMcQ-BES

packages/web-config-server/src/apiV1/legacyReport.js
```diff
@@ -0,0 +1,33 @@
...
+      throw new Error(`No dashboard item exists with the code ${itemCode}`);
+    }
+    const legacyReport = await this.models.legacyReport.findOne({
+      code: dashboardItem.report_code,
+    });
+    if (!legacyReport) {
+      throw new Error(`No legacy report exists with id ${dashboardItem.report_code}`);
+    }
+    return {
+      viewJson: dashboardItem.config,
```
edmofro, line 27: @chris-bes when you get to integrating this with Tupaia, if we can remove the need for the front end to receive `viewJson` at this point (as it should already have it from the first `/dashboards` request), then we can change this lookup to use the `reportCode` rather than the `itemCode`, and remove all need for the `/legacyReport` endpoint to query the `dashboardItem` table (which would more appropriately restrict its responsibility).

@EMcQ-BES has kept the `viewJson` in the response here in case there's too much to unpick to remove it.

edmofro (approved):
This PR is approved, nice one @EMcQ-BES!

It shouldn't be tested merged though until we have work from @chris-bes updating where some of the filter config sits and proving it can integrate with the Tupaia front end

packages/web-config-server/src/apiV1/legacyReport.js
chris-bes: That's a great idea @edmofro ! I agree it'd be better not to return `viewJson` here. 

However, I just have 1 concern that even if we don't need to return `viewJson`, we still need to send in the `itemCode` because if we only restrict it to `legacy_report` look up here, we won't be able to check for permission since all the permission columns are in the `dashboard_relation`. And it may not be possible to map the `report_code` back to a unique `dashboard_relation` because 1 `legacy_report` may be used in multiple `dashboard_item`? So I feel it's still necessary to pass the `itemCode` (and also `dashboardCode`) to this endpoint purely for permission check purpose. I can think of 2 options:

1. Have only 1 endpoint `dashboardItem/:itemCode/report?dashboardCode=xxx` which will check for the right permissions and also fetch either legacy report or new report based on `legacy`. So the front end just need to call this endpoint.
Or
2. At least add another `permission_group` column into `legacy_report` (similar to `report` table) so that we can check the permission (it will not be as strict as how we check in dashboard_relation)

I'm open to any other suggestions as well!

edmofro: Oh right I see the issue. I don't think we should go for option 2., as it seems like unnecessary extra work for a legacy table. So how about another option:

3. Consistent endpoint used by both, `/report/:reportCode?itemCode=yyy&legacy=true`

Points of difference:
- Term `reportCode` rather than `dashboardCode` (dashboards are now what we formerly called dashboard groups)
- `reportCode` moved to endpoint segment rather than url param (it's mandatory when fetching from that endpoint)
- `itemCode` moved to query param, as it's only required for legacy reports so could one day be dropped
- `legacy` status passed through from front end in query param, so that the back end route handler could process a non-legacy report without having to go to the database at all (doesn't have to re-look up the dashboard item, just passes request straight off to report server like a well behaved orchestration server)

What do you think?

chris-bes: > Term reportCode rather than dashboardCode (dashboards are now what we formerly called dashboard groups)

Oh I actually did mean `dashboardCode` in this case, because a `dashboard_item` can be linked to multiple dashboards, so just `itemCode` itself won't be sufficient to fetch a unique `dashboard_relation`.

Agreed with all other points! The only reason why I chose passing only `itemCode` but not `reportCode` was we can get the `reportCode` from `dashboard_item`, and it might mean we're sending redundant information. Also to prevent the case that the passed in `itemCode` and `reportCode` don't actually map together (ie: `itemCode` is a total different one that don't use the passed in `report_code`, but maybe we can also validate this in PermissionChecker?). So I was kinda leaning that `dashboard_item` is the source of truth. But if you think we don't have to worry about that then happy with your suggestion 👍

edmofro: Aaah right, sorry I didn't take enough time understanding the reason for `dashboardCode`, that makes sense.

But yes confirming I'm happy for report data fetches to skip the validation of the dashboard item, given that reports have a permission check on them themselves it's not a security issue (perhaps could catch a front end bug, but it'd be a pretty dumb bug ;-)). The legacy reports should have a permission check that does use the dashboard item code and dashboard code, but hopefully one day the legacy endpoint can be dropped entirely (maybe in 2035)

tcaiger:
closing as all changes are merged into: `2814-integrate-new-dashboard-tupaia`

edmofro:
@chris-bes this says it was reopened by you, but I don't think that was intentional? Closing

PR #2541, opened 2021-05-11
1506 1506 Active and archived alerts reports

chris-bes:
### Issue:
https://github.com/beyondessential/tupaia-backlog/issues/1506
https://github.com/beyondessential/tupaia-backlog/issues/1509

### Changes:

- Allow fetching event data values in report-server
- Added a few functions for report-server
- Added Active and archived alerts reports

### Responses:
I'm skipping `sitesReported` column and `totalCases` for now since I think they're out of scope. For `sitesReported`, as I understand, we can only create alerts for countries, not sentinel sites. So I'm not sure if that makes sense...

## Active alerts report:

```
{
    "results": [
        {
            "id": "609a07505da1e6102c00007d",
            "syndrome": "DIA",
            "organisationUnit": "FJ",
            "period": "2021W16"
        },
        {
            "id": "609a07505da1e6102c00007f",
            "syndrome": "DLI",
            "organisationUnit": "FJ",
            "period": "2021W16"
        }
    ]
}
```

## Archived alerts report:
```
{
    "results": [
        {
            "id": "609a12025da1e6102c000111",
            "syndrome": "AFR",
            "diagnosis": "A80",
            "organisationUnit": "PW",
            "outbreakStartDate": "20210505",
            "period": "2021W18"
        }
    ]
}
```


kael89:
@chris-bes thank you for the response samples!

> I'm skipping `sitesReported` and `totalCases` column for now since I think they're out of scope. For `sitesReported`, as I understand, we can only create alerts for countries, not sentinel sites. So I'm not sure if that makes sense...

I think that this data shouldn't come from the alert reports anyway, but rather from https://github.com/beyondessential/tupaia-backlog/issues/1507... So we are all good 😃 

---

By the way, perhaps `sitesReported` is just the equivalent of "Sites Reported"  in the from the Weekly Reports page (which is not directly related to alerts):

image

My guess is that this data is useful to display, in order to assess the spread of symptoms which lead to an alert

packages/report-server/src/reportBuilder/transform/aliases/periodConversionAliases.ts
```diff
@@ -13,3 +13,13 @@ export const convertPeriodToWeek = () =>
       '...': '*',
     },
   ]);
+
+export const convertEventDateToWeek = () =>
+  buildTransform([
+    {
+      transform: 'select',
+      "'period'": "convertToPeriod($row.eventDate, 'WEEK')",
```
rohan-bes, line 21: Man, it's kinda annoying that data-broker returns event with this `eventDate` field instead of `period` like `analytics` do huh? Oh well, that's too big of an issue to deal with right now 😝

packages/database/src/migrations/20210510061852-AddPSSSActiveAlertsReport-modifies-data.js
```diff
@@ -0,0 +1,66 @@
...
+    config: {
+      fetch: {
+        dataGroups: ['PSSS_Alert'],
+        dataElements: ['PSSS_Alert_Syndrome', 'PSSS_Alert_Archived'],
+      },
+      transform: [
+        {
+          // filter active alerts
+          transform: 'filter',
+          where: "eq($row['PSSS_Alert_Archived'], 0) or notExists($row['PSSS_Alert_Archived'])",
```
rohan-bes, line 39:
```suggestion
          where: "eq($row['PSSS_Alert_Archived'], 0) or not exists($row['PSSS_Alert_Archived'])",
```
Totally optional, but it's pretty neat that mathjs has such boolean logic built into it so we don't have to implement inverse functions if we don't want 😄

packages/database/src/migrations/20210511050324-AddPSSSArchivedAlertsReport-modifies-data.js
```diff
@@ -0,0 +1,78 @@
...
+      },
+      transform: [
+        {
+          // filter archived alerts
+          transform: 'filter',
+          where: "eq($row['PSSS_Alert_Archived'], 1)",
+        },
+        'convertEventDateToWeek',
+        {
+          // convert outbreakStartDate to correct DAY format
```
rohan-bes, line 48: Nice comments! Wonder if it would be possible to add comments directly into the config itself... 🤔

packages/report-server/src/reportBuilder/fetch/functions/dataGroup.ts
```diff
@@ -7,22 +7,33 @@ import { Aggregator } from '../../../aggregator';
 import { FetchReportQuery } from '../../../types';
 import { FetchResponse } from '../types';
 
+type DataGroupParams = {
+  dataGroups: unknown;
+  dataElements: undefined;
```
rohan-bes, line 12:
```suggestion
  dataElements: unknown;
```
I think `unknown` is the super type here, we don't technically know that the field is undefined just as yet 👍

packages/report-server/src/reportBuilder/functions/basic.ts
```diff
@@ -25,10 +25,18 @@ export const eq = (value1, value2): boolean => {
   return value1 === value2;
 };
 
+export const notEq = (value1, value2): boolean => {
+  return value1 !== value2;
+};
```
rohan-bes, line 30: As per my other comments I guess we don't necessarily need these functions, but happy to keep them in as they'll probably provide useful 😄 

Also worth considering that maybe we wanna migrate some of these functions into the expression parser package? Food for thought for the future 👍

rohan-bes (approved):
Looks really good @chris-bes, great reports!

Few minor suggestions but approving anyway.

kael89:
@chris-bes another thing I just realised... I believe we should use a day period here, not a week period. The reason is that an alert can be created during the week, and I believe we want to capture the exact day

chris-bes:
> @chris-bes another thing I just realised... I believe we should use a day period here, not a week period. The reason is that an alert can be created during the week, and I believe we want to capture the exact day

Can we or do we meant to be able to create an alert during the week? As I've checked the survey, the alert triggered date is (which is the period) is meant to be Weekly
Image from iOS

Also the mock up shows weekly period and `AlertTable` use `WeekAndDateCell`. But maybe we need to confirm the business rule with someone

kael89:
> Can we or do we meant to be able to create an alert during the week? As I've checked the survey, the alert triggered date is (which is the period) is meant to be Weekly

Oh I see, my bad... I had the impression that we'd need to know the exact day, but all the evidence suggests otherwise 😃

PR #2545, opened 2021-05-12
1510: PSSS - Show active alerts

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1510

### Screenshots:

psss_active_alerts

Please note that some functionality is missing, and should be implemented in future PRs:
* Table sorting
* Selecting a date range instead of a single week, but I want to check with the product owner first
* You'll see in the screenshot there is a "fetchData is not a function" error on the right, I think we need a new card for this ("Get alert data by id"). It seems similar to https://github.com/beyondessential/tupaia-backlog/issues/1507, but not exactly given its current specs


packages/ui-components/src/components/Table/TableRow.js
```diff
@@ -70,7 +70,7 @@ export const StyledTableRow = styled(MuiTableRow)`
 `;
 
 export const TableRow = React.memo(({ columns, rowData, className, onRowClick }) => (
-  <StyledTableRow className={className} onClick={onRowClick}>
+  <StyledTableRow className={className} onClick={e => onRowClick(e, rowData)}>
```
kael89, line 73: Hi @tcaiger , I'd like to check this change through you... Although I could find other ways to manage state related to a clicked row (eg using a Provider), I believe that the above behaviour is reasonable, and perhaps expected. Keen to hear your thoughts 🙂

Also, this change _should_ be safe since `onClick` is natively called with just one argument. So unless if a client does something unexpected, the new `rowData` parameter shouldn't shadow any argument in existing usages of `onRowClick`

packages/psss/src/containers/Panels/AlertsPanel.js
```diff
@@ -70,60 +69,72 @@ const menuOptions = [
...
 
-export const AlertsPanel = React.memo(({ isOpen, handleClose }) => {
-  const [isModalOpen, setIsModalOpen] = useState(false);
-  const [activeIndex, setActiveIndex] = useState(0);
-  const sitesState = useFetch(getAffectedSites);
-  const notesState = useFetch(getAlertsMessages);
-  const activityState = useFetch(getActivityFeed);
-  const { countryCode } = useParams();
-  const countryName = useSelector(state => getCountryName(state, countryCode));
+export const AlertsPanel = React.memo(
```
kael89, line 72: Apologies for the diff in this file - a lot of it is due to nesting change by `prettier`

packages/ui-components/src/components/Table/TableRow.js
kael89: Speaking of safe changes - unit tests saved my bacon! https://github.com/beyondessential/tupaia/pull/2545/commits/d46ae910d6dd771889e0323fd44d4365f85b863a

packages/psss/src/api/queries/useActiveAlerts.js
```diff
@@ -0,0 +1,23 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { MIN_DATE, SYNDROMES } from '../../constants';
+import { getPeriodByDate } from '../../utils';
+import { useReport } from './helpers';
+
+export const useActiveAlerts = (period, orgUnitCodes) => {
```
tcaiger, line 10: This is looking tight! The meshing of the server response with front end display data (syndomeName) is really clean .

packages/psss/src/components/Table/CellComponents/WeekAndDateCell.js
```diff
@@ -9,17 +9,17 @@ import styled from 'styled-components';
 import * as COLORS from '../../../constants/colors';
 import { getDisplayDatesByPeriod, getWeekNumberByPeriod } from '../../../utils';
 
-const CountrySummaryTitle = styled.div`
+const WeekAndDateTitle = styled.div`
```
tcaiger, line 12: Good renaming!

packages/psss/src/constants/syndromes.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+export const SYNDROMES = {
```
tcaiger, line 6: Great call pulling this into a seperate file - I can see that it will be useful in lots of places.

packages/psss/src/views/Tabs/AlertsTabView.js
```diff
@@ -19,41 +31,62 @@ const DateSubtitle = styled(Typography)`
...
+        onChange={newDate => setPeriod(getPeriodByDate(newDate))}
+        value={getDateByPeriod(period)}
+        isOpen={isPickerOpen}
+        onClose={() => setIsPickerOpen(false)}
+      />
+    </>
+  );
+
   return (
     <div style={{ position: 'relative' }}>
```
tcaiger, line 67: I think this div could be removed. I'd say it was just to hold the placeholder coming soon banner

packages/ui-components/src/components/Table/TableRow.js
tcaiger: Yes I think this is a great idea!

packages/psss/src/containers/Tables/AlertsTable.js
```diff
@@ -53,34 +57,47 @@ const createColumns = isForMultipleCountries => [
   },
 ];
 
-export const AlertsTable = React.memo(({ handlePanelOpen, countryCode }) => {
-  const { isLoading, isFetching, error, data, order, orderBy, handleChangeOrderBy } = useTableQuery(
-    'alerts',
+const getColumns = isSingleCountry =>
```
tcaiger, line 60: Good refactor - much easier to read!

packages/psss/src/containers/Panels/AlertsPanel.js
```diff
@@ -70,60 +69,72 @@ const menuOptions = [
...
+  ({ countryCode, period, syndromeName, isOpen, handleClose }) => {
+    const [isModalOpen, setIsModalOpen] = useState(false);
+    const [activeIndex, setActiveIndex] = useState(0);
+    const sitesState = useFetch(getAffectedSites);
+    const notesState = useFetch(getAlertsMessages);
```
tcaiger, line 77: I assume that none of the functionality of the tabs in the panel is supposed to work yet. As you can see the data for the tabs is just stubbed out using the fake api data. Also that useFetchHook is from a time before react-query was added to the project so it should be considered legacy and replaced with a react-query strategy when the real data is available.

packages/psss/src/views/Tabs/AlertsTabView.js
```diff
@@ -19,41 +31,62 @@ const DateSubtitle = styled(Typography)`
 
 export const AlertsTabView = React.memo(() => {
   const [isPanelOpen, setIsPanelOpen] = useState(false);
-  const { countryCode } = useParams();
+  const [isPickerOpen, setIsPickerOpen] = useState(false);
+  const [rowData, setRowData] = useState({});
+  const [period, setPeriod] = useState(getCurrentPeriod());
+  const { organisationUnit: countryCode, syndromeName } = rowData;
```
tcaiger, line 37: As far as I can tell, the `countryCode` wont exist in the rowData for single country users. I'm having trouble with my migrations so I can't test it. Does it work for you?

tcaiger: Also I think it would be nice to get the period from the rowData if it's easy. I think it would be more robust and less likely to potentially cause bugs as the row clicked on is always the data used. ie. in the original designs it was possible to see all the alerts in the table so there could be multiple alerts from a single country.

tcaiger (changes requested):
Looks really nice. Just left one question about the alert table data displaying on singleCountry pages. I might have missed something and if so then just let me know.

packages/ui-components/src/components/Table/TableRow.js
kael89: Eventually I will use `Context` (for other reasons). So I'll revert this change in another PR, but let's keep it in mind in case it would be useful in the future!

packages/psss/src/containers/Panels/AlertsPanel.js
kael89: This is my assumption too!

packages/psss/src/views/Tabs/AlertsTabView.js
kael89: Great point!

packages/psss/src/views/Tabs/AlertsTabView.js
kael89: Both good points! I'm going to pass the period from `rowData` in now, and if you don't mind, I'll fix the `countryCode` issue in the PR for https://github.com/beyondessential/tupaia-backlog/issues/1511 - there is a bit of params refactoring that needs to happen for  that one

tcaiger: Sweet - sounds good

kael89:
Hi @tcaiger , I made the suggested changed in the PR based on top of this one (https://github.com/beyondessential/tupaia/pull/2548 - see last commit), could you please re-review?

tcaiger (approved):

PR #2546, opened 2021-05-12
2655: Fix bug in samoa covid percentage map overlay

chris-bes:
### Changes:

- Fix bug in samoa covid percentage map overlay


billli0 (approved):
LGTM

chris-bes:
Merging to fix an existing issue in dev

PR #2548, opened 2021-05-12
1514: PSSS - Show archived alerts

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/1514

### Screenshots:
image

Please note that:
* table sorting will be added in a future PR
* data is currently non-paginated, and there is no way to filter them (eg select date). There is no mockup for this screen, so I'd like to discuss with the product owner before implementing something


packages/psss-server/src/routes/reports/FetchArchivedAlertsReportRoute.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { Route } from '../Route';
+
+export class FetchArchivedAlertsReportRoute extends Route {
```
tcaiger, line 8: I notice that this route is very similar to the `FetchActiveAlertsReportRoute` route. Did you consider having one endpoint with a query param to set which type of data is being requested?

packages/psss/src/api/queries/useArchivedAlerts.js
```diff
@@ -0,0 +1,26 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { MIN_DATE, SYNDROMES } from '../../constants';
+import { getPeriodByDate } from '../../utils';
+import { useReport } from './helpers';
+
+export const useArchivedAlerts = orgUnitCodes => {
```
tcaiger, line 10: again it seems like this could be drier. I don't believe in 100% dry all the time but this is a good candidate. I guess the main reason not combine them would be if we anticipate different requirements in the future between the archive data and the active data which may well be the case.

packages/psss/src/containers/Tables/ArchiveTable.js
```diff
@@ -61,36 +65,45 @@ const createColumns = isForMultipleCountries => [
...
 ];
 
-export const ArchiveTable = React.memo(({ countryCode }) => {
-  const { isLoading, isFetching, error, data, order, orderBy, handleChangeOrderBy } = useTableQuery(
-    'archive',
-  );
+const getColumns = isSingleCountry =>
+  isSingleCountry ? archiveColumns : [countryColumn, ...archiveColumns];
+
+const ArchiveTableComponent = React.memo(({ countryCode, countryCodes }) => {
```
tcaiger, line 76: I foresee the ArchiveTable, ArchiveTabView etc being very different from the AlertsTabView etc so I think it's great that these are seperate templates.

tcaiger (approved):
Nice work - looks tight. I just left one question about combining the Active and Archive data fetching but I'm happy to approve it because I assume that it was a deliberate decision not to combine them.

packages/psss/src/api/queries/useArchivedAlerts.js
kael89: Yeap, that was my line of thought too. There already exist some subtle differences (start end end dates), and the queries are simple enough that re-defining them probably makes them simpler than abstracting them.

packages/psss/src/containers/Tables/ArchiveTable.js
kael89: Agreed! Any similarities could be extracted in shared components. They have some obvious similarities (eg shared columns - but not always column widths), but extracting partial similarities would probably make things more complicated than simplify, in this case.

packages/psss-server/src/routes/reports/FetchArchivedAlertsReportRoute.ts
kael89: I did, but again, these endpoints are just `input => fetch => output` - the only "complexity" is converting the `orgUnitCodes` param, so I thought that it would be more maintainable to define all logic in each file

packages/psss/src/api/queries/useArchivedAlerts.js
kael89: You know what, you are right - and actually that would reflect the original endpoint design:

Active alerts: `/alerts`
Archived alerts: `/alerts/archived`

Thanks for pointing this out - will change that!

packages/psss-server/src/routes/reports/FetchArchivedAlertsReportRoute.ts
kael89: You know what, you are right - and actually that would reflect the original endpoint design:

Active alerts: /alerts
Archived alerts: /alerts/archived

Thanks for pointing this out - will change that!

PR #2550, opened 2021-05-12
1507: PSSS Confirmed data for alerts reports

chris-bes:
### Changes:

- Alert data per syndrome report
- This is the report for the Main alert tab https://github.com/beyondessential/tupaia-backlog/issues/1507

### Responses:
Request URL: `{{URL}}/fetchReport/PSSS_DIA_Confirmed_Report?period=2021W15;2021W16;2021W17&hierarchy=psss&organisationUnitCodes=PW`

Not returning sitesReported as it is descoped for Phase 2

```
{
    "results": [
        {
            "totalCases": 15
        }
    ]
}
```


packages/database/src/migrations/20210512001037-AddPSSSAccumulatedConfirmedCasesPerSyndromeIndicators-modifies-data.js
```diff
@@ -0,0 +1,52 @@
...
+        formula: `PSSS_Confirmed_${syndrome}_Cases`,
+        aggregation: ['FINAL_EACH_WEEK', 'SUM_PER_ORG_GROUP'],
+      },
+    };
+
+    await insertObject(db, 'indicator', indicator);
+  }
+
+  return null;
+};
```
rohan-bes, line 39: Hmmm so this is a good case of us to start asking the question, when to create an `indicator` or not?
- On the one hand, this is a valid concept within our PSSS system, and may prove to be re-used in other place
- On the other hand, this is just an aggregation on an existing indicator. Perhaps it would be best just to add support for aggregations in the report-server? (something we are planning on doing anyway)

I think personally I am leaning towards the later, however I'm still not entirely sure how best to structure the `aggregation` config in report-server.

If you're happy to go down that route, perhaps you, I, and @kael89 if he's interested could have a design discussion on the config?

packages/database/src/migrations/20210512004319-AddPSSSConfirmedDataPerSyndromeReport-modifies-data.js
```diff
@@ -0,0 +1,74 @@
...
+        'mostRecentValuePerOrgUnit',
+        {
+          // change key names
+          transform: 'select',
+          "'totalCases'": `$row.${totalConfirmedCasesIndicator}`,
+        },
+      ],
+    },
+    permission_group_id: permissionGroupId,
+  };
```
rohan-bes, line 52: So it was decided that we won't support a `# sites reported` accumulated value? It was gonna be tricky to do so I'm somewhat glad (but I was kinda curious to see if we could pull it off 😉)

rohan-bes (commented):
Just leaving as comment here since I think some further discussion is in order. Let me know what you think about my suggestion to scrap the indicator and use aggregation in report-server instead?

packages/database/src/migrations/20210512001037-AddPSSSAccumulatedConfirmedCasesPerSyndromeIndicators-modifies-data.js
chris-bes: Yep I totally agree the latter option will definitely be more useful in the future, so I'm happy to go down that route. The only concern I have is time constraint given that we have to deliver phase 2 by end of this week, and I'm not too familiar with report-server to estimate how long it would take to add support for aggregation. If you think it's achievable then count me in 👍

packages/database/src/migrations/20210512004319-AddPSSSConfirmedDataPerSyndromeReport-modifies-data.js
chris-bes: Yeap Erin has confirmed this!

packages/database/src/migrations/20210512001037-AddPSSSAccumulatedConfirmedCasesPerSyndromeIndicators-modifies-data.js
chris-bes: As discussed with Rohan on Slack. Ultimately, we want to achieve the aggregation pattern used in indicator package, which is to have separate aggregation per data element. But for now, we can keep it simple and just specify a single aggregation for all data elements as a default config, which is all required for this report and some upcoming STRIVE dashboards for solutions team. So for now the config could look like:

```
{
  fetch: {
    dataElements: [...],
    aggregation: 'SUM_PER_PERIOD'
  },
  transform: [...]
}
```

kael89: Great call guys, I also agree! We could pull the aggregator config parsing out of `indicators` and into a common place (eg the `aggregator` package)

chris-bes:
@rohan-bes my apology for the force pushes. The last commit should be ready for re-review

packages/report-server/src/aggregator/Aggregator.ts
```diff
@@ -14,13 +14,22 @@ type PeriodParams = {
 };
 
 export class Aggregator extends BaseAggregator {
+  aggregationToAggregationConfig = (aggregation: string ) => ({
+    type: aggregation
+  });
+
   async fetchAnalytics(
     dataElementCodes: string[],
+    aggregationList: string[] | undefined,
```
rohan-bes, line 23: Tbh I suspect I will refactor this soon as we will need to be able to pass in aggregation config in the config itself, but that can be done when it's needed 👍

rohan-bes (approved):
@chris-bes Looks awesome, thanks for adding the aggregation config 👍 

One request, can you please _**not**_ squash and merge this PR? Just regular merge. There's a good chance I'm gonna end up cherry-picking that commit to add the aggregation config and if it's squashed it will bring in the migration for the reports as well

packages/report-server/src/aggregator/Aggregator.ts
chris-bes: Yeah I agree, I was initially allowing object here for that. But I don't want to push too much if it's not needed now (I don't think those STRIVE vizes for solutions will need it anyway)

PR #2551, opened 2021-05-12
Entity aggregation in the db

edmofro:
### Issue 

https://github.com/beyondessential/tupaia-backlog/issues/2244

### Changes:

- Built off @rohan-bes's proof of concept [here](https://github.com/beyondessential/tupaia/commit/0ea9b4c5bdf2f9828a0e130255fc867a2910db3b#)
- Because we keep introducing more complexity into the analytics/aggregated fetch, I split out the event fetching into a separate class, with only a little common code in a parent class. I find it easier to reason about rather than having to consider both cases at once.
- In splitting out these classes, I was initially trying to be as DRY as possible, but following recent conversations I realised it was actually more straightforward if there was some repeated code in order to make assembling the query in your mind a lot easier, without having to jump in and out of the parent class
- The other benefit of the classes is that we can build up the list of query parameters for substitution in a field, rather than have a function mutate a variable that gets passed around everywhere, which I personally found unexpected


edmofro:
Closing in favour of https://github.com/beyondessential/tupaia/pull/2573

PR #2561, opened 2021-05-14
2663 async issues psss

IgorNadj:
### Changes:

See each commit for changes, summary:
- Add opt-in flag for survey response submissions that blocks the response and waits for analytics table rebuild
- Allow submit survey responses with a data_time without a tz
- Fix frontend query cache invalidation to properly refetch data after mutation

---

### Screenshots:
https://github.com/beyondessential/tupaia-backlog/issues/2663#issuecomment-841042918

### DIFF
--- a/packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
+++ b/packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
@@ -4,7 +4,7 @@
  */
 
 import momentTimezone from 'moment-timezone';
-import { DatabaseError, UploadError, stripTimezoneFromDate } from '@tupaia/utils';
+import { DatabaseError, UploadError, stripTimezoneFromDate, reformatDateStringWithoutTz, ValidationError } from '@tupaia/utils';
 import { uploadImage } from '../s3';
 import { BUCKET_PATH, getImageFilePath } from '../s3/constants';
 import { DEFAULT_DATABASE_TIMEZONE, getEntityIdFromClinicId } from '../database';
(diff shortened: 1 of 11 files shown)


packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
```diff
@@ -134,3 +122,49 @@ const createEntities = async (models, entitiesCreated, surveyId) => {
...
+      return stripTimezoneFromDate(
+        momentTimezone(suppliedDataTime, suppliedTimezone).format(),
+      );
+    }
+
+    // No timezone specified. We are submitting the data_time explicitly without a tz.
+    //
+    // If the input is in a known format like ISO8601 without tz, we can use this directly, we just
+    // reformat it into our specific format without tz.
+    const reformattedDataTime = reformatDateStringWithoutTz(suppliedDataTime);
```
IgorNadj, line 152: With this change, I wanted to tackle the problematic business logic of silently defaulting to Pacific/Auckland. 

It took me a couple days of sleuthing to figure out what was going on, and I am very familiar with the codebase, it might take the next person a week or more. Having it default to Pacific/Auckland is not obvious. Also, it introduces inconsistency because I would not have run into the problem if I were developing in NZ, but I just so happened to be in Melb, in a different timezone.

The first option would be to throw an error if timezone is missing, but this doesn't make sense as an interface. Our clients e.g. psss-server already have the data_time in the correct format (iso8601 without tz), they shouldn't be inventing a timezone, which will then be stripped.

So our second option, and the one I settled on, was to change the behaviour to only default to Pacific/Auckland for older clients as a fallback. Modern clients which use data_time will never have the timezone defaulted, they can either specify it, or specify a valid known data_time string in format ISO_8601 without tz.

IgorNadj: @EMcQ-BES Would be great to get your input on this, as you worked on this recently

IgorNadj:
Closing, invalid branch name. New PR: https://github.com/beyondessential/tupaia/pull/2562

PR #2562, opened 2021-05-14
2663 async issues psss (2)

IgorNadj:
### Changes:

See each commit for changes, summary:
- Add opt-in flag for survey response submissions that blocks the response and waits for analytics table rebuild
- Allow submit survey responses with a data_time without a tz
- Fix frontend query cache invalidation to properly refetch data after mutation

---

### Screenshots:
https://github.com/beyondessential/tupaia-backlog/issues/2663#issuecomment-841042918

### DIFF
--- a/packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
+++ b/packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
@@ -4,7 +4,7 @@
  */
 
 import momentTimezone from 'moment-timezone';
-import { DatabaseError, UploadError, stripTimezoneFromDate } from '@tupaia/utils';
+import { DatabaseError, UploadError, stripTimezoneFromDate, reformatDateStringWithoutTz, ValidationError } from '@tupaia/utils';
 import { uploadImage } from '../s3';
 import { BUCKET_PATH, getImageFilePath } from '../s3/constants';
 import { DEFAULT_DATABASE_TIMEZONE, getEntityIdFromClinicId } from '../database';
@@ -66,19 +66,7 @@ export async function updateOrCreateSurveyResponse(models, surveyResponseObject)
     }
 
     // Ensure data_time is populated, supporting legacy versions of MediTrak
-    const {
-      data_time: suppliedDataTime,
-      submission_time: submissionTime, // v1.7.87 to v1.9.110 (inclusive) uses submission_time
-      end_time: endTime, // prior to v1.7.87 fall back to end_time
-      timezone = DEFAULT_DATABASE_TIMEZONE, // if no timezone provided, use db default
-    } = surveyResponseObject;
-    const dataTime = suppliedDataTime || submissionTime || endTime;
-
-    // Convert to the original timezone, then strip timezone suffix, so it ends up in db as it
-    // appeared to the original survey submitter
-    surveyResponseProperties.data_time = stripTimezoneFromDate(
-      momentTimezone(dataTime).tz(timezone).format(),
-    );
+    surveyResponseProperties.data_time = getDataTime(surveyResponseObject);
 
     surveyResponse = await models.surveyResponse.updateOrCreate(
       {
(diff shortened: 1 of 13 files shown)


packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
```diff
@@ -66,19 +66,7 @@ export async function updateOrCreateSurveyResponse(models, surveyResponseObject)
...
-      timezone = DEFAULT_DATABASE_TIMEZONE, // if no timezone provided, use db default
-    } = surveyResponseObject;
-    const dataTime = suppliedDataTime || submissionTime || endTime;
-
-    // Convert to the original timezone, then strip timezone suffix, so it ends up in db as it
-    // appeared to the original survey submitter
-    surveyResponseProperties.data_time = stripTimezoneFromDate(
-      momentTimezone(dataTime).tz(timezone).format(),
-    );
+    surveyResponseProperties.data_time = getDataTime(surveyResponseObject);
```
IgorNadj, line 69: With this change, I wanted to tackle the problematic business logic of silently defaulting to Pacific/Auckland. 

It took me a couple days of sleuthing to figure out what was going on, and I am very familiar with the codebase, it might take the next person a week or more. Having it default to Pacific/Auckland is not obvious. Also, it introduces inconsistency because I would not have run into the problem if I were developing in NZ, but I just so happened to be in Melb, in a different timezone.

The first option would be to throw an error if timezone is missing, but this doesn't make sense as an interface. Our clients e.g. psss-server already have the data_time in the correct format (iso8601 without tz), they shouldn't be inventing a timezone, which will then be stripped.

So our second option, and the one I settled on, was to change the behaviour to only default to Pacific/Auckland for older clients as a fallback. Modern clients which use data_time will never have the timezone defaulted, they can either specify it, or specify a valid known data_time string in format ISO_8601 without tz.

IgorNadj: @EMcQ-BES Would be great to get your input on this, as you worked on this recently

EMcQ-BES: This all definitely **looks** like a very reasonable update to the `data_time` logic, removing an assumption of the original design and making it more generic is always a good call.

I'm definitely interested in seeing some regression testing since it's always a little difficult to wrap your head around timezone logic in code.

EMcQ-BES (approved):
LGTM!

packages/meditrak-server/src/routes/postChanges.js
```diff
@@ -64,8 +65,17 @@ export async function postChanges(req, res) {
...
 
-  for (const { action, translatedPayload } of translatedChanges) {
+  for (const { action, translatedPayload, ...rest } of translatedChanges) {
     await ACTION_HANDLERS[action](models, translatedPayload);
+
+    if (action === SUBMIT_SURVEY_RESPONSE) {
+      const { waitForAnalyticsRebuild } = rest;
+      if (waitForAnalyticsRebuild) {
+        const { database } = models;
+        await AnalyticsRefresher.executeRefresh(database);
```
chris-bes, line 75: Just minor but perhaps we can do this outside of the for loop, just in case in the future there are multiple changes that need to refresh the analytics table. Just a thought so no need to change it now

packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
chris-bes: I like this change too! I think at the moment we only have meditrak-app and psss-server clients for this endpoint. So hopefully regression testing is not going to be too much...

chris-bes (approved):
Looks good to me too!

packages/meditrak-server/src/dataAccessors/updateOrCreateSurveyResponse.js
chris-bes: As discussed we'll also bring the same change to `importSurveyResponses` endpoint

packages/meditrak-server/src/routes/importSurveyResponses/importSurveyResponses.js
```diff
@@ -39,7 +39,12 @@ export async function importSurveyResponses(req, res) {
       throw new UploadError();
     }
     const surveyNames = getArrayQueryParameter(req?.query?.surveyNames);
-    const timeZone = req?.query?.timeZone || DEFAULT_DATABASE_TIMEZONE;
+    const timeZone = req?.query?.timeZone;
```
IgorNadj, line 42: The change here is simpler, we just require the timezone which then gets stripped. Since it's a 1:1 mapping of endpoint and client, we don't need the endpoint to be as generic as postChanges. We should clean this up in the future perhaps.

PR #2564, opened 2021-05-16
2680 multi measure map overlays

tcaiger:
### Issue #: [2680](https://github.com/beyondessential/tupaia-backlog/issues/2680) Support Multi Measure Map Overlays

The scope was initially to fix support for multi-measure in the map components and in lesmis. I did quite a bit of testing and found a bug with the legend for radius map overlays which was crashing the app. Once I fixed this the multi-measure map overlays worked as expected.

While testing I took the change to add support for light theme to the map components.

### Changes:

- fix radius legend item bug
- add support for light theme to map components

---

### Screenshots:
see issue

### DIFF
--- a/packages/ui-components/src/components/Map/LeafletStyles.js
+++ b/packages/ui-components/src/components/Map/LeafletStyles.js
@@ -59,4 +59,10 @@ export const LeafletStyles = css`
   .leaflet-tooltip-right:before {
     border-right-color: #555;
   }
+  .leaflet-bottom.leaflet-right {
+    display: none;
+  }
+  .leaflet-control.leaflet-control-attribution {
+    background: none;
+  }
 `;
--- a/packages/ui-components/src/components/Map/LeafletStyles.js
+++ b/packages/ui-components/src/components/Map/LeafletStyles.js
@@ -59,7 +59,7 @@ export const LeafletStyles = css`
   .leaflet-tooltip-right:before {
     border-right-color: #555;
   }
-  .leaflet-bottom.leaflet-right {
+  .leaflet-bottom.leaflet-right .leaflet-control-attribution {
     display: none;
   }
   .leaflet-control.leaflet-control-attribution {
--- a/packages/ui-components/src/components/Map/LeafletStyles.js
+++ b/packages/ui-components/src/components/Map/LeafletStyles.js
@@ -60,6 +60,8 @@ export const LeafletStyles = css`
     border-right-color: #555;
   }
   .leaflet-bottom.leaflet-right .leaflet-control-attribution {
+    // Leaflet attribution is manually added in the TileLayer along side
+    // the mapbox attribution to keep all the attributions in one place
     display: none;
   }
   .leaflet-control.leaflet-control-attribution {
--- a/packages/ui-components/src/components/Map/Legend/LegendEntry.js
+++ b/packages/ui-components/src/components/Map/Legend/LegendEntry.js
@@ -6,12 +6,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import MuiButton from '@material-ui/core/Button';
 
-const Key = styled.div`
+const Button = styled(MuiButton)`
   display: flex;
   flex-direction: row;
   align-items: center;
-  padding: 0.3rem 0.3rem;
+  padding: 0.2rem 0.3rem;
+  font-weight: 400;
+  letter-spacing: 0;
   pointer-events: auto;
   margin-bottom: 0.1rem;
   cursor: pointer;
@@ -33,10 +36,10 @@ export const LegendEntry = React.memo(
     };
 
     return (
-      <Key onClick={handleClick} hidden={hidden}>
+      <Button onClick={handleClick} hidden={hidden}>
         {marker}
         <Label>{label}</Label>
-      </Key>
+      </Button>
     );
   },
 );
--- a/packages/ui-components/src/components/Map/Legend/LegendEntry.js
+++ b/packages/ui-components/src/components/Map/Legend/LegendEntry.js
@@ -3,18 +3,23 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-import PropTypes from 'prop-types';
 import React from 'react';
+import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
 const Key = styled.div`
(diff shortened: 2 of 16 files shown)


packages/ui-components/src/components/Map/Legend/LegendEntry.js
```diff
@@ -3,18 +3,26 @@
...
   align-items: center;
-  padding: 0.3rem;
+  padding: 0.2rem 0.3rem;
+  font-weight: 400;
+  letter-spacing: 0;
   pointer-events: auto;
+  margin-bottom: 0.1rem;
   cursor: pointer;
-  ${p => (p.hidden ? 'opacity: 0.5;' : '')}
+  opacity: ${props => (props.hidden ? '0.5' : '1')};
```
edmofro, line 21: Good improvement going for `props`, but it does make me wonder why `hidden` goes to `0.5` rather than `0` opacity

packages/ui-components/src/components/Map/Markers/markerIcons.js
```diff
@@ -43,21 +43,31 @@ const IconContainer = ({ children, scale, ...props }) => (
...
   return (
-    <IconContainer fill={color} scale={scale} viewBox="0 0 24 33">
-      <circle cx="12" cy="12" r="7" fill="#F9F9F9" />
+    <IconContainer fill={color} scale={scale} viewBox="0 0 27 39">
       <path
-        d="M12 0C5.37429 0 0 5.17275 0 11.55C0 20.2125 12 33 12 33C12 33 24 20.2125 24 11.55C24 5.17275 18.6257 0 12 0ZM12 15.675C9.63429 15.675 7.71429 13.827 7.71429 11.55C7.71429 9.273 9.63429 7.425 12 7.425C14.3657 7.425 16.2857 9.273 16.2857 11.55C16.2857 13.827 14.3657 15.675 12 15.675Z"
+        d="M13.5 0.631592C6.04607 0.631592 0 6.64584 0 14.0605C0 24.1322 13.5 39 13.5 39C13.5 39 27 24.1322 27 14.0605C27 6.64584 20.9539 0.631592 13.5 0.631592Z"
         fill={color}
       />
+      <circle cx="13.5" cy="14.1316" r="6.39474" fill="white" fillOpacity="0.2" />
```
edmofro, line 70: I guess it's this `fillOpacity` that's making it appear darker than it did before. What was the motivation for this change (both the restructure of the icon svg and the opacity tweak)?

edmofro: I was also thinking it might look good with a full white border around the pin, so that all colours show up nicely against the dark satellite view

packages/ui-components/src/components/Map/LeafletStyles.js
```diff
@@ -59,4 +59,10 @@ export const LeafletStyles = css`
   .leaflet-tooltip-right:before {
     border-right-color: #555;
   }
+  .leaflet-bottom.leaflet-right .leaflet-control-attribution {
+    display: none;
+  }
+  .leaflet-control.leaflet-control-attribution {
+    background: none;
+  }
```
edmofro, line 67: Can you add a comment around this stuff to indicate that we're hiding it because we manually display it on the left instead? It just looks a bit dodgy in our central Map styles otherwise

edmofro (changes requested):
lgtm other than the pin's inner circle

packages/ui-components/src/components/Map/Legend/LegendEntry.js
tcaiger: perhaps hidden is a bad name. The map overlay markers that are hidden and the legend button is 'greyed out' with 50% opacity.

edmofro: Aah yeah that makes sense, the prop name relates more to the thing the legend entry controls than the legend entry itself. I'm fine to leave it because I can't think of anything better that captures the idea across both the legend entry and the effect it's having

packages/ui-components/src/components/Map/Markers/markerIcons.js
tcaiger: The pin update was design direction from Dave ([here](https://www.figma.com/file/HSyuh072zJaSUeGRZ0k5dq/LESMIS-Design-System?node-id=1577%3A0)). He thought that the white inner was a bit too jarring. He actually designed dark inner circles but when I implemented it, I thought it looked better with semi-white circles.

edmofro: Oh right. I'm not loving it, let's see if we can get a quick look from Dave today. If not, happy to approve at the end of the day and make any small tweak on Thursday.

packages/ui-components/src/components/Map/utils/markerFormats.js
```diff
@@ -274,26 +310,10 @@ export function getMeasureDisplayInfo(
...
-        ...hiddenValues[key],
-      });
+  serieses.forEach(
+    ({ key, type, valueMapping, noDataColour, scaleType, scaleColorScheme, min, max }) => {
+      const valueInfo = getValueInfo(measureData[key], valueMapping);
 
-      if (valueInfo.isHidden) {
-        displayInfo.isHidden = true;
-      }
       displayInfo.originalValue = measureData.originalValue;
```
edmofro, line 306: It seems like this `originalValue` is a bit dodgy too...it'll be the value of the final series. We don't currently use it in any meaningful way in `ui-components` or `lesmis` though, so it's probably fine until we integrate with tupaia.org

packages/ui-components/src/components/Map/utils/markerFormats.js
```diff
@@ -190,13 +190,36 @@ export function flattenMeasureHierarchy(measureHierarchy) {
   return results;
 }
 
-function getValueInfo(value, valueMapping, hiddenValues = {}) {
+const getIsHidden = (measureData, serieses, allHiddenValues) =>
+  serieses
+    .map(({ key, valueMapping, hideByDefault }) => {
```
edmofro, line 195: Most of the code in the block below this is somewhat repeating the logic in `getValueInfo`...but it's just different enough that after staring for a while it's hard to see how to nicely DRY it out. I'm happy to approve it, but I reckon if the rest of this file got a refactor at some point it could all be cleaned up a fair bit.

edmofro (approved):

PR #2565, opened 2021-05-16
1511-1515: PSSS - Get alerts per country

kael89:
### Issue #:
[#1511](https://github.com/beyondessential/tupaia-backlog/issues/1511)
[#1515](https://github.com/beyondessential/tupaia-backlog/issues/1515)

### Screenshots:

**Active alerts for a country:**
active_alerts_for_country

**Archived alerts for a country:**
archived_alerts_for_country

We should also add a **Back* button to return to the international alerts page, similar to the one existing in Weekly Reports currently. It is not a part of the existing UI so I will check with the product owner first and then do it in a later PR


packages/psss-server/src/app/addRoutesToApp.ts
```diff
@@ -57,7 +57,7 @@ export function addRoutesToApp(app: Express) {
    * GET routes
    */
   app.get('/v1/test', handleWith(TestRoute));
-  app.get('/v1/alerts/:status', handleWith(FetchAlertsRoute));
+  app.get('/v1/alerts/:category', handleWith(FetchAlertsRoute));
```
tcaiger, line 60: Good rename

packages/psss/src/components/Table/CellComponents/AlertMenuCell.js
```diff
@@ -2,12 +2,30 @@
...
+  const { setIsOpen, setData } = useContext(AlertsPanelContext);
+  const data = { organisationUnit, period, syndromeName };
+
+  const handleClick = useCallback(() => {
+    setData(data);
+    setIsOpen(true);
+  }, [data, setData, setIsOpen]);
+
+  return (
+    <IconButton onClick={handleClick}>
```
tcaiger, line 21: I'm a bit confused here. In the design this button is for opening a menu with options to "Archive Alert" or "Create Outbreak". If that is no longer the plan then I think we should just remove this button but I don't think it is a good icon and placement for the opening of the side panel.

packages/psss/src/containers/Tables/AlertsTable/getColumns.js
```diff
@@ -0,0 +1,118 @@
...
+  CountryLinkCell,
+  StartDateCell,
+  SyndromeCell,
+  WeekAndDateCell,
+} from '../../../components';
+
+const createCountryColumn = alertCategory => {
+  const CellComponent = ({ organisationUnit }) => (
+    <CountryLinkCell
+      target={`/alerts/${alertCategory}/${organisationUnit}`}
```
tcaiger, line 18: As far as I remember there wasn't a requirement to do this. I actually think it's a nice feature but does make some other things like the date and the navigation to open the panel confusing. Might be worth touching base with Dave.

packages/psss/src/components/Table/CellComponents/AlertMenuCell.js
```diff
@@ -2,12 +2,30 @@
...
+import { AlertsPanelContext } from '../../../containers';
 
-export const AlertMenuCell = () => (
-  <IconButton>
-    <MoreVertIcon />
-  </IconButton>
-);
+export const AlertMenuCell = React.memo(({ organisationUnit, period, syndromeName }) => {
+  const { setIsOpen, setData } = useContext(AlertsPanelContext);
+  const data = { organisationUnit, period, syndromeName };
```
tcaiger, line 13: I'm curious about whether there is a unique id or code for an alert? My understanding is that it is just a survey response but it might still be nice to have the survey response id on the front end.

packages/psss/src/containers/Panels/AlertsPanel.js
```diff
@@ -67,74 +67,81 @@ const menuOptions = [
...
-AlertsPanel.defaultProps = {
-  period: '',
-  countryCode: '',
-  syndromeName: '',
+AlertsPanelProvider.propTypes = {
+  children: PropTypes.node.isRequired,
 };
+
+export const AlertsPanel = React.memo(() => {
+  const { data: panelData, isOpen, setIsOpen } = useContext(AlertsPanelContext);
```
tcaiger, line 90: The context is a nice solution to making the button which opens the panel and the panel share the state of the alert. Hopefully the actual data for the panel can be just fetched using react-query and so it doesn't have to be passed through context. It's a bit hard to see until it's finished if panel data is for the reference to the alert data or the actual alert data.

packages/psss/src/containers/NavBar.js
```diff
@@ -53,11 +36,13 @@ const makeLinks = state => {
     {
       label: 'Weekly Reports',
       to: multiCountry ? '/' : `/weekly-reports/${countryCodes[0]}`,
+      isActive: constructIsActive(['', 'weekly-reports']),
```
tcaiger, line 39: This is a great improvement!

packages/psss/src/routes/PageRoutes.js
```diff
@@ -4,64 +4,50 @@
...
-  </Switch>
-));
-
-PageRoutesComponent.propTypes = {
-  countryCodes: PropTypes.arrayOf(PropTypes.string).isRequired,
-};
-
-const mapStateToProps = state => ({
-  countryCodes: getCountryCodes(state),
+      <PrivateRoute path="/alerts/:category" authCheck={checkCountries}>
```
tcaiger, line 42: I like the new urls. They are more consistent which is nice.

packages/psss/src/views/AlertsOutbreaksView.js
```diff
@@ -56,7 +64,7 @@ export const AlertsOutbreaksView = () => {
     <>
       <Header Title={Title} ExportModal={ExportModal} />
       <TabsToolbar links={links} />
-      <AlertsRoutes />
+      <AlertsTabView />
```
tcaiger, line 67: In the designs the Archive view is quite different to the Alerts view. There are different columns in the table, different actions in the table, different page layout (full width vs. 2 columns), filters etc. At the moment the pages are the same. Is this just temporary or has there been a design change? I still lean to the idea of different templates if the views are different but I am open to combining them if you think the pros out weight the costs.

packages/psss/src/views/Tabs/AlertsTabView.js
```diff
@@ -63,28 +57,32 @@ export const AlertsTabView = React.memo(() => {
...
+          <AlertsPanelProvider>
+            {/* TODO Outbreak and Alert components could be merged when outbreaks are implemented */}
+            {alertsCategory === 'outbreaks' ? <OutbreaksTable /> : <AlertsTable period={period} />}
+            {alertsCategory === 'outbreaks' ? <OutbreaksPanel /> : <AlertsPanel />}
+          </AlertsPanelProvider>
+        </Main>
+        {alertsCategory !== 'outbreaks' && (
+          <Sidebar>
+            <Card variant="outlined">
+              <CardHeader color="primary" title="Selected Week" label={DatePicker} />
```
tcaiger, line 75: I know this is from a previous change but I missed it first time around. My understanding of the card in the side of the page was that it just showed the current iso week and then all the alerts were shown in the table. Is there any reason to not show all the alerts? One would imagine the list will never be very long at one time and navigating to alerts in the past doesn't seem like a realistic use case. Perhaps it's a question for a PO or designer?

tcaiger (changes requested):
Overall looking great. There are a bunch of great improvements to my original code. I have left a few pesky questions which would be good to cover. Thanks

packages/psss/src/components/Table/CellComponents/AlertMenuCell.js
kael89: Hey Tom, thanks for pointing this out! I indeed assumed that his button was used for opening the side panel, but I guess I got it wrong.

Perhaps there is a newer mockup but that's the one I am using: https://www.figma.com/file/N1AqYzfAXsTv4FHvAyb8a7/PSSS-Admin---Desktop?node-id=205%3A630 which is quite different from the existing components as far as alerts are regarded, so it often couldn't be used as a UI guide for alerts.

Let's have a chat so that you can explain me in more detail!

packages/psss/src/views/Tabs/AlertsTabView.js
kael89: I believe I got this wrong... I see what you mean. Let's also discuss about this

kael89:
New screenshots:

Active alerts - multiple countries
new_active_alerts

Archived alerts - multiple countries
new_archived_alerts

Active alerts - single country
new_active_alerts_single_country

Archived alerts - single country
new_archived_alerts_single_country

packages/psss/src/views/Tabs/AlertsTabView.js
```diff
@@ -30,53 +18,23 @@ const DateSubtitle = styled(Typography)`
...
+          <AlertsTable period={period} />
+          <AlertsPanel />
+        </AlertsPanelProvider>
       </Main>
       <Sidebar>
         <Card variant="outlined">
-          <CardHeader color="primary" title="Current Week" label={DatePicker} />
+          <CardHeader
+            color="primary"
+            title="Selected Week"
```
tcaiger, line 35:
```suggestion
            title="Current Week"
```

tcaiger (approved):
Looks good to me. Much easier to review now that we are on the same page :)

PR #2567, opened 2021-05-17
2686: Integrate Tupaia to be able to fetch data via Report-Server

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2686:
- Added reportServer databuilder which fetches data from the report-server
- Added new `access_token_expiry` field to `userSession` table to allow for `web-config-server` to manage it's own token refresh logic

Release tasks:
- [x] Create tupaia-server@tupaia.org api-client user
- [x] Create `REPORT_API_URL`, `MICROSERVICE_CLIENT_USERNAME`, `MICROSERVICE_CLIENT_SECRET` env vars


packages/database/src/migrations/20210520020331-AddAccessTokenExpiryFieldToTupaiaUserSessionTable-modifies-schema.js
```diff
@@ -0,0 +1,33 @@
...
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`
+    ALTER TABLE "userSession" 
+    ADD COLUMN access_token_expiry BIGINT NOT NULL DEFAULT 0;
```
rohan-bes, line 20: Bit of an odd decision for me to choose the default of 0 here. Basically, didn't wanna just make it `NOT NULL` as then I'd have to track down and refactor everywhere that these sessions get created. This way, sessions can be created without an expiry and then whenever a piece of code that uses the expiry logic gets called it immediately refreshes and sets the expiry for the first time.

packages/psss-server/src/models/PsssSession.ts
```diff
@@ -3,21 +3,13 @@
...
 import { AccessPolicyObject, SessionFields, SessionDetails } from '../types';
 import { AuthConnection } from '../connections/AuthConnection';
 
-const getTokenExpiry = (accessToken: string) => {
-  const { exp: expiryAuthServerClock, iat: issuedAtAuthServerClock } = jwt.decode(accessToken);
-  // subtract 3 seconds to account for latency since generation on the auth server
-  const validForSeconds = expiryAuthServerClock - issuedAtAuthServerClock - 3;
-  const expiryPsssServerClock = Date.now() + validForSeconds * 1000;
-  return expiryPsssServerClock;
-};
```
rohan-bes, line 19: Moved to `utils`

packages/server-boilerplate/src/orchestrator/models/Session.ts
```diff
@@ -26,30 +26,6 @@ interface SessionFields {
...
-const getTokenExpiry = (accessToken: string) => {
-  const [expiryAuthServerClock, issuedAtAuthServerClock] = extractTokenDetails(
-    jwt.decode(accessToken),
-  );
-
-  // subtract 3 seconds to account for latency since generation on the auth server
-  const validForSeconds = expiryAuthServerClock - issuedAtAuthServerClock - 3;
-  const expiryServerClock = Date.now() + validForSeconds * 1000;
-  return expiryServerClock;
-};
```
rohan-bes, line 51: Moved to `utils`

packages/report-server/src/aggregator/Aggregator.ts
```diff
@@ -69,10 +69,10 @@ const buildPeriodQueryParams = ({ period, startDate, endDate }: PeriodParams) =>
     builtEndDate = endDate;
   } else if (!period && !startDate && !endDate) {
     builtPeriod = getDefaultPeriod();
-    [builtStartDate, builtEndDate] = convertPeriodStringToDateRange(period);
+    [builtStartDate, builtEndDate] = convertPeriodStringToDateRange(builtPeriod);
```
rohan-bes, line 72: Was throwing error here when passing in undefined period previously

packages/web-config-server/src/tests/apiV1/dataBuilders/generic/reportServer/reportServerDataBuilder.test.js
```diff
@@ -0,0 +1,139 @@
...
+const reportConnection = { fetchReport };
+const models = {
+  project: { findOne: findProject },
+  entityHierarchy: { findById: findEntityHierarchyById },
+};
+
+describe('ReportServerDataBuilder', () => {
+  const assertBuilderResponseIsCorrect = async ({ config, query, entity }, expectedResponse) => {
+    const dataBuilder = new ReportServerBuilder(req, models, config, query, entity);
+    dataBuilder.reportConnection = reportConnection;
```
rohan-bes, line 82: Being quite heavy handed with the stubbing and mocking here. In truth I suspect there's a way to do this _correctly_ using sinon however I've really struggled with mocking microservices before. I'm not the most concerned when it comes to doing mocking the right way provided it works correctly, but am happy to put the time in to get this right if reviewer has a different opinion

packages/database/src/migrations/20210520020331-AddAccessTokenExpiryFieldToTupaiaUserSessionTable-modifies-schema.js
edmofro: Nice :+1:

packages/web-config-server/src/tests/apiV1/dataBuilders/generic/reportServer/reportServerDataBuilder.test.js
edmofro: There are a couple of things you could change if you wanted to:
- There's no huge benefit to having data come back in the response, unless I'm missing a processing step that's required in the data builder. Instead you could just have it respond with nothing but use the spy that sinon creates to check that it was called with the right parameters. It would make things a bunch simpler, but only do it if you are interested in learning about spies (and I'm happy to talk to you more about them on a call)
- Another potential change would be to move the test data into a `.fixtures` file and/or the stubs into a `.stubs` file. That can tidy up the tests file a fair bit, though it's definitely not necessary when neither are being used across multiple tests.

However, I'm also 100% fine with the approach you've taken, and the tests themselves are good. Also, my personal philosophy is that tests should be the gentlest part of a PR review, given that push back on tests will make adding tests feel like even more work, so make it less likely that people will do them. It's a bit like my Mum's philosophy of "never criticise the way someone does the vacuuming" ;-)

edmofro (approved):
Only one very minor required change (tweaking the wording of a comment), the rest are optional, so approved in advance :-)

packages/web-config-server/src/tests/apiV1/dataBuilders/generic/reportServer/reportServerDataBuilder.test.js
rohan-bes: > There's no huge benefit to having data come back in the response, unless I'm missing a processing step that's required in the data builder.
You're right, and originally I just thought about spying to see if the correct URL was used. However, technically there's the conversion of `{ results: [...], ... }` to `{ data: [...] }` being performed. While totally trivial, there's actually so little going on in this databuilder that it amounts to a reasonable percentage of the logic 😂 I guess I always wanted to validate the output as it will force someone to update the tests if they decide to extend/alter the shape of the response (ie. adding 'metadata' or 'period') fields that other databuilders produce.

rohan-bes: > "never criticise the way someone does the vacuuming"

I love this ❤️

PR #2572, opened 2021-05-18
2775-1600: Create alerts

chris-bes:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2775
https://github.com/beyondessential/tupaia-backlog/issues/1600

### Changes:

- Create alerts on confirming data. 
- Show alert created message


packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
```diff
@@ -37,12 +51,68 @@ export class ConfirmCountryWeeklyReportRoute extends Route {
...
+      }
+    }
+
+    if (previousAlertResponses && previousAlertResponses.length) {
+      // delete the old alerts in case the weekly data is
+      // re-confirmed and there have been some alerts created before
+      for (const alertResponse of previousAlertResponses) {
+        await this.meditrakConnection?.deleteSurveyResponse(alertResponse);
+      }
+    }
```
chris-bes, line 112: Not sure if this is the best way..

For Alert survey responses, they are no longer unique by the combo of surveyCode/orgUnit/period like weekly reports. The unique combo for the alert survey responses is should be surveyCode/syndrome/orgUnit/period. So in the scenario of creating alerts when **reconfirming data**, to deal with old alerts, the easiest way I've found is to delete them and recreate new ones again.

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
```diff
@@ -37,12 +51,68 @@ export class ConfirmCountryWeeklyReportRoute extends Route {
...
+
+        response.createdAlerts.push({
+          id: surveyResponseId,
+          title: syndromeCode,
+        });
+      }
+    }
+
+    if (previousAlertResponses?.length) {
+      // delete the old alerts in case the weekly data is
```
IgorNadj, line 107: Can you reconfirm? I thought it was a one off. Either way, good check

packages/psss/src/components/Modal/AlertCreatedModal.js
```diff
@@ -55,9 +55,11 @@ export const AlertCreatedModal = ({ isOpen, alerts, handleClose }) => (
         <Heading variant="h6">Alert has been created</Heading>
         <Copy>Syndrome is above the threshold and an Alert has been created.</Copy>
         {alerts.map(alert => (
-          <Link key={alert.id} to="alerts">
-            View alert: {alert.title}
-          </Link>
+          <div>
+            <Link key={alert.id} to="/alerts">
```
IgorNadj, line 59: Ah yep good solution

IgorNadj (approved):

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
chris-bes: I think if you have confirmed the data, and the weekly data changed, you can actually reconfirm it (I guess also in case you make a mistake then you can also fix it). But confirming with @rohan-bes

rohan-bes: @chris-bes @IgorNadj Yep, you can re-confirm. Useful for when a new site enters data. Also useful if you make a mistake 😄

rohan-bes: @chris-bes however, I think it's possible we may not want to 'delete' existing alerts when reconfirming... Let's have a chat about this?

chris-bes: sure, let's do that after your refinement meeting

chris-bes:
Hi @kael89 
https://github.com/beyondessential/tupaia/pull/2572/commits/8f6a40fe940d593d48308a8f66f311e5b3eb07ee: This commit should take care of a few more scenarios when creating alerts:
**Scenario 1:** There's already confirmed data for for Week 8 and AFR crosses the threshold so an alert has been triggered in Week 8. Then data for Week 8 is changed. When we are reconfirming the changed data in Week 8, if AFR no longer crosses the threshold => we should archive the alert (Erin's comment here: https://github.com/beyondessential/tupaia-backlog/issues/2775#issuecomment-842877567)

**Scenario 2:** When we're confirming data for Week 8 which has AFR crosses the threshold, if there's already an active alert triggered since Week 3 for AFR, we don't want to create an alert for AFR (Erin's comment here: https://github.com/beyondessential/tupaia-backlog/issues/2775#issuecomment-842878373)

**Scenario 3:** If there's an active alert for AFR created in Week 8 2021, then in Week 13 the threshold for AFR is no longer crossed, then we don't archive the alert automatically but let the users archive it themselves (Erin's comment here: https://github.com/beyondessential/tupaia-backlog/issues/2775#issuecomment-842987139)

packages/database/src/migrations/20210510061852-AddPSSSActiveAlertsReport-modifies-data.js
```diff
@@ -30,13 +30,14 @@ exports.up = async function (db) {
...
-        dataElements: ['PSSS_Alert_Syndrome', 'PSSS_Alert_Archived'],
+        dataElements: ['PSSS_Alert_Syndrome', 'PSSS_Alert_Archived', 'PSSS_Outbreak_Date'],
       },
       transform: [
         {
           // filter active alerts
           transform: 'filter',
-          where: "eq($row['PSSS_Alert_Archived'], 0) or notExists($row['PSSS_Alert_Archived'])",
+          where:
+            "(eq($row['PSSS_Alert_Archived'], 0) or notExists($row['PSSS_Alert_Archived'])) and notExists($row['PSSS_Outbreak_Date'])",
```
chris-bes, line 40: Fixed something that I missed, active alerts shouldn't return outbreaks

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
```diff
@@ -37,12 +56,93 @@ export class ConfirmCountryWeeklyReportRoute extends Route {
...
     return this.meditrakConnection?.updateOrCreateSurveyResponse(
-      SURVEY_CODE,
+      CONFIRMED_WEEKLY_SURVEY_COUNTRY,
       countryCode,
       week,
       answers,
     );
   }
+
+  async createAlerts(countryCode: string, week: string) {
```
rohan-bes, line 66: Love to see this renamed to something like `updateAlerts()` as we are both creating and arching existing alerts. Bonus points if you refactor out the internal logic of creating an alert or archiving one so it can read something like:
```
updateAlerts() {
  
  // ... setup ...

  if (!syndromeAlerts && result[`${syndromeCode} Threshold Crossed`] === true) {
    const newAlert = createAlert();
    response.createdAlerts.push(newAlert);
  }
  
  if (currentWeekSyndromeAlert && result[`${syndromeCode} Threshold Crossed`] === false) {
    archiveAlert();
  }
}
```

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
```diff
@@ -37,12 +56,93 @@ export class ConfirmCountryWeeklyReportRoute extends Route {
...
+        500,
+      );
+    }
+
+    const [result] = report.results;
+    const response: any = {
+      createdAlerts: [],
+    };
+    const startWeek = dateStringToPeriod(MIN_DATE, 'WEEK');
+    const activeAlertsData = await this.reportConnection?.fetchReport(
```
rohan-bes, line 85: I suppose we don't definitely have to decide how to deal with this case for now, as Outbreaks are out of scope, however are we sure that we should be creating a new alert if there is an existing (non-archived) Outbreak with that syndrome? I'd expect we don't (so this report would also need to return Outbreaks) but we can confirm with Erin on this.

Regardless not gonna suggest we bother creating a new report for this as of yet since Outbreaks are out of scope...

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
```diff
@@ -37,12 +56,93 @@ export class ConfirmCountryWeeklyReportRoute extends Route {
...
+      const currentWeekSyndromeAlert =
+        syndromeAlerts && syndromeAlerts.find(a => a.period === week);
+
+      // If there is an existing alert triggered in the selected week,
+      // and now for the selected week, the threshold is no longer crossed (because of reconfirming changed data),
+      // archive the existing alert triggered in the selected week
+      if (currentWeekSyndromeAlert && result[`${syndromeCode} Threshold Crossed`] === false) {
+        const alertSurveyResponse = await this.meditrakConnection?.findSurveyResponseById(
+          currentWeekSyndromeAlert.id,
+        );
```
rohan-bes, line 137:
```suggestion
```
I think this may be a bit of an un-needed fetch here. To get the `SurveyResponseObject` required by `updateSurveyResponse()` we only need `id`, `entityCode`, `dataTime` and `surveyCode`. All of these I believe we should have access to without requiring a fetch from meditrak-server correct?

We have a couple of options of how we could handle this:
a) Contruct a `SurveyResponseObject` inline here and pass it in
b) Refactor `updateSurveyResponse` to instead take `id`, `entityCode`, `dataTime` and `surveyCode`

I have a slight preference for b), how do you feel about it?

rohan-bes (changes requested):
Few changes requested, feel free to ask me about them if any are unclear 👍 

Looking good though!!

packages/psss-server/src/connections/MeditrakConnection.ts
```diff
@@ -103,7 +120,8 @@ export class MeditrakConnection extends ApiConnection {
     surveyCode: string,
     organisationUnitCode: string,
     period: string,
-    answers: Record<string, number>,
+    answers: Answers,
+    surveyResponseId?: string | undefined,
```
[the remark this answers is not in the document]
chris-bes: Yeah the `changes` endpoint unfortunately only returns a success message. What you suggested also works too. I did this because just in case in the future there's something useful that we want to get from the response of `changes` endpoint. Or we can return both the generated id and the response of `changes` endpoint

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
```diff
@@ -37,12 +56,93 @@ export class ConfirmCountryWeeklyReportRoute extends Route {
...
+          },
+          surveyResponseId,
+        );
+
+        response.createdAlerts.push({
+          id: surveyResponseId,
+          title: syndromeCode,
+        });
+
+        continue;
```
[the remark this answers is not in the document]
chris-bes: My bad, I removed some extra code after this line but forgot to remove `continue` 🤦

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
chris-bes: I've actually confirmed this with Erin and I think she said we still should create a new alert if there's already an existing active outbreak.

This is from her:
```
Thus the only way a new one is created when there is already an alert for that syndrome in recent weeks is if the existing alert was archived or turned into an outbreak
```

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
chris-bes: I think it will be useful to have both? It would be good if TypeScript can do method overloading. But if picking 1 then I agree b is the better option.

rohan-bes: TS can actually do overloading 👍 great idea, both is good! https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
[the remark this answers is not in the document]
chris-bes: Actually I just realized it's not the end of the loop. So this `continue` should stay

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
chris-bes: After playing with overloading in TS, I generally feel it is a bit awkward (at least for my case) and it's quite different than how other programming languages do overloading. So I decided to have just different method names for both options. We need a different method that takes `id`, `entityCode`, `period` and `surveyCode` anyway because we need to do timezone stripping for the period, and for the method that takes surveyResponseObject, we don't need to because data_time timezone should already be stripped.

We can discuss this further in the office tomorrow

rohan-bes: Yeah that's all good, I also am not the biggest fan of function overloads in `TS`. Happy with your latest code 👍

packages/psss-server/src/routes/reports/ConfirmCountryWeeklyReportRoute.ts
rohan-bes: Tbh I'm still not 100% sure that logic is valid, but I'll discuss this with Erin when we want to implement outbreaks 👍

rohan-bes (approved):
Looks awesome man, great work @chris-bes !

PR #2573, opened 2021-05-18
2244 Entity aggregation in the database

edmofro:
### Issue 

https://github.com/beyondessential/tupaia-backlog/issues/2244

### Changes:

- Built off @rohan-bes's proof of concept [here](https://github.com/beyondessential/tupaia/commit/0ea9b4c5bdf2f9828a0e130255fc867a2910db3b#)
- Because we keep introducing more complexity into the analytics/aggregated fetch, I split out the event fetching into a separate class, with only a little common code in a parent class. I find it easier to reason about rather than having to consider both cases at once.
- In splitting out these classes, I was initially trying to be as DRY as possible, but following recent conversations I realised it was actually more straightforward if there was some repeated code in order to make assembling the query in your mind a lot easier, without having to jump in and out of the parent class
- The other benefit of the classes is that we can build up the list of query parameters for substitution in a field, rather than have a function mutate a variable that gets passed around everywhere, which I personally found unexpected

### Notes

This actually doesn't affect many dashboard reports at all! In order to benefit from this change, one of the newly supported aggregation types needs to be the first aggregation on the list. It turns out that's not the case very often. Currently most dashboards that could benefit actually use `entityAggregation` plus an assumed default aggregation of `MOST_RECENT` that runs inside web-config-server. I've updated one such dashboard report as a proof of concept, and was planning to update more but decided to isolate the regression risk here and allow incrementally improving others, especially given most of the Laos Schools ones that posed the original performance issue are going to be thrown away and replaced with new report-server based dashboard items. For newly built dashboards, which will use `aggregations` rather than `entityAggregation`, we should see this be useful more often.


packages/aggregator/src/__tests__/Aggregator/Aggregator.test.js
```diff
@@ -40,7 +40,7 @@ const aggregationOptions = {
   aggregations: [
     {
       type: 'MOST_RECENT',
-      config: { orgUnitToGroupKeys: [], requestedPeriod: '20200214;20200215' },
+      config: { orgUnitMap: {}, requestedPeriod: '20200214;20200215' },
```
edmofro, line 43: Bit of a tangential change, cleaning up a couple of places that used a different key name. `orgUnitMap` might not be ideal, but at least it's consistent

packages/database/src/migrations/20210524212900-FixEntityAggregationsLaosSchools-modifies-data.js
```diff
@@ -0,0 +1,60 @@
...
+  `);
+  return dashboardReports[0] || null;
+}
+
+// Update data builder configuration for a report
+async function updateBuilderConfigByReportId(db, newConfig, reportId) {
+  return updateValues(db, 'dashboardReport', { dataBuilderConfig: newConfig }, { id: reportId });
+}
+
+const ENTITY_AGGREGATIONS_BY_REPORT_ID = {
```
edmofro, line 32: This migration is a lot more generic than it needs to be as I expected to move more than one dashboard report to using entity agg in the db, but as mentioned in the PR description, decided against at this point

packages/web-config-server/examples.http
```diff
@@ -1,29 +1,39 @@
-@hostname = localhost
-@port = 8000
-@host = {{hostname}}:{{port}}
-@version = v1
+@localhost = http://localhost:8000
```
edmofro, line 1: Some changes in here that allow it to be easier to swap between local and dev for making comparisons (as well as making some things harder to improve simplicity, like swapping api version, which we never do)

packages/web-config-server/src/aggregator/buildAggregationOptions.js
```diff
@@ -28,12 +28,14 @@ export const buildAggregationOptions = async (
     aggregationOrder: entityAggregationOrder = DEFAULT_ENTITY_AGGREGATION_ORDER,
   } = entityAggregationOptions;
 
-  // Note aggregationType and aggregationConfig might be undefined
-  const inputAggregations = aggregations || [{ type: aggregationType, config: aggregationConfig }];
+  // Note aggregationType and aggregationConfig might be undefined, in which case it will default
+  // to MOST_RECENT if no multi-aggregations are provided
```
edmofro, line 32: I had a logic change in here originally that filtered out `undefined` single aggregations, but it turns out that was required so that it would default to `MOST_RECENT`. Have updated the comment to clarify, and these changes should now be superficial only

packages/web-config-server/examples.http
```diff
@@ -1,29 +1,39 @@
...
     "emailAddress": "{{email}}",
     "password": "{{password}}"
 }
 
-###
-
+### Logout
 # @name logout
-POST http://{{host}}/api/{{version}}/logout HTTP/1.1
+GET {{host}}/api/v1/logout HTTP/1.1
```
edmofro, line 24: Weirdly the logout endpoint on web-config-server is `GET`, so have updated this to work

packages/data-api/src/EventsFetchQuery.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+    if (this.eventId) {
+      conditions.push('event_id = ?');
+      params.push(this.eventId);
+    }
+
+    if (conditions.length === 0) return '';
+
+    this.paramsArray.push(...params);
+    return conditions.length > 0 ? `WHERE ${conditions.join(' AND ')}` : '';
```
rohan-bes, line 66:
```suggestion
    return `WHERE ${conditions.join(' AND ')}`;
```
We are checking for `conditions.length === 0` just above so can just return here right? 👍

packages/data-api/src/EventsFetchQuery.js
```diff
@@ -0,0 +1,78 @@
...
+    if (this.dataGroupCode) {
+      conditions.push('data_group_code = ?');
+      params.push(this.dataGroupCode);
+    }
+    if (this.eventId) {
+      conditions.push('event_id = ?');
+      params.push(this.eventId);
+    }
+
+    if (conditions.length === 0) return '';
```
rohan-bes, line 63:
```suggestion
    if (conditions.length === 0) {
      return '';
    }
```
Totally optional and correct me if I'm wrong, but I thought we tended to avoid omiting brackets after if clauses?

packages/data-api/src/EventsFetchQuery.js
```diff
@@ -0,0 +1,78 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { DataFetchQuery } from './DataFetchQuery';
+
+export class EventsFetchQuery extends DataFetchQuery {
```
rohan-bes, line 8: Great job breaking out the queries between `fetchAnalytics` and `fetchEvents`! Trying to support one in the same queryBuilder was getting pretty untenable

rohan-bes (commented):
Just leaving review as a comment as I'm still not entirely finished reviewing. That being said, happy for someone else to pick up if you wanna get this one reviewed by tomorrow!

Great work, and thanks so much for implementing this!! I think when all the entity-aggregation stuff comes through then we can have a quick ticket or two to migrate the Laos Schools dashboards/overlays to use this and then hopefully see a significant performance improvement!!

To be completely honest, I do still find the `AnalyticsFetchQuery` logic to be quite hard to follow. I really don't think this is on you though, and I think things have definitely been improved from where they were. I'll take some time tomorrow to brainstorm a bit about how we might find a cleaner way of building the query, but as I've said before I am okay with this area of code being rather niche and convoluted.

My only real significant comment is that I do think that changing to that new entity-aggregation type will break the existing Laos Schools Male Female dashboard in tupaia.org for all levels lower than country.

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -0,0 +1,202 @@
...
+    return [this.getEntityCodeField(), 'data_element_code'];
+  }
+
+  getEntityCommonTableExpression() {
+    // if mapping from one set of entities to another, include the mapped codes as "aggregation_entity_code"
+    const isAggregatingEntities = this.isAggregating && this.aggregation.switches.aggregateEntities;
+    const entityMap = isAggregatingEntities && this.aggregation.config.orgUnitMap;
+    const columns = isAggregatingEntities ? ['code', 'aggregation_entity_code'] : ['code'];
+    const rows = isAggregatingEntities
+      ? Object.entries(entityMap).map(([key, value]) => [key, value.code])
```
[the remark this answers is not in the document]
edmofro: I agree! There are a couple of places that `name` is used, but ideally we'd get that from somewhere else and make this a simpler data structure

packages/database/src/migrations/20210524212900-FixEntityAggregationsLaosSchools-modifies-data.js
```diff
@@ -0,0 +1,60 @@
...
+  Laos_Schools_Male_Female: {
+    dataSourceEntityType: 'school',
+    aggregationEntityType: 'country',
```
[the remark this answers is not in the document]
edmofro: So there's this crazy thing right, where actually you _can_ use `country` as your aggregation entity type, because when you're at the district level, the data source entities being passed in are only those for the district you're in, and they sum up to `LA` rather than `LA_XXX` but who cares, because the sum is the same! It's not ideal, and `requested` is absolutely the future way to go, but I thought this might be good enough for now.

However, you're actually right that I did make a big mistake and this doesn't work any more - at the country or any other level! `MOST_RECENT_PER_ORG_GROUP` is finding the most recent response across all schools, so we only get one school represented in the result, which is obviously not at all what we want. We can do `SUM_PER_ORG_GROUP` instead, and it actually works perfectly fine (with aggregation entity type set to `country`, still works fine at ever level), but it would break if other data were submitted in future, because those schools with multiple data points would have both counted.

What we really want is a `MOST_RECENT_PER_ENTITY` followed by a `SUM_PER_ORG_GROUP`. Unfortunately we're not going to get a heap of efficiency out of the db there, especially while there is only one submission per school, as the first aggregation won't cut down the results at all.

Still, we have to be correct, so I'm backing out this migration altogether 😞 

This makes me think maybe we do have to be less atomic with aggregation types, and allow e.g. a `SUM_PER_ORG_GROUP_MOST_RECENT_PER_ENTITY` 🙄

packages/data-api/src/EventsFetchQuery.js
edmofro: Hmmm, well I think opinion is actually divided there!

I personally prefer using the brackets, but at some point there was pushback and we disabled the eslint rule, so now I have a bias towards brackets but am not consistent and don't really care.

Tupaia's second developer, Marty, was hugely pro the brackets being there, which is what got me in the habit and probably why the eslint rule used to exist.

Either way I'm not fussed!

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -0,0 +1,202 @@
...
+
+    return `
+      CROSS JOIN LATERAL (
+        SELECT ${fields}
+        FROM analytics
+        INNER JOIN entity_codes_and_relations ON entity_codes_and_relations.code = analytics.entity_code
+        WHERE ${conditions.join('\n      AND ')}
+        ORDER BY date DESC
+        LIMIT 1
+      ) as latestPerPeriod
```
rohan-bes, line 183: Small thing, but I really appreciate you giving this join table a human readable name! I just did a loggable query on your branch and found it a lot more understandable than it used to be! 🎉

rohan-bes (approved):
Awesome work, looking forward to building dashboards and overlays that make use of this functionality 👍 

As discussed I do feel it would be acceptable to just us a `SUM_PER_ORG_GROUP` aggregation for the `Laos_Schools_Male_Female` dashboard, and instead we can rely on another process to ensure there's just one submission per year. But as you said, it would be best to approach that in another body of work 👍

I'll message separately, and perhaps create a ticket, about the possibility we have to chain aggregations in the database level, allowing us to do something like `[MOST_RECENT, SUM_PER_ORG_GROUP]`.

PR #2578, opened 2021-05-19
2768 lesmis dashboard deep linking

tcaiger:
### Changes:

- add `useUrlSearchParam` util
- hold the state for `dashboardTab`, `dashboard`, and `year` in the url
- extract sticky bar behaviour into `useStickyBar` hook 
 
Note:
The handling of bad urls could be improved. I will look at that in the next deep linking PR.


packages/lesmis/src/api/queries/useMapOverlayReportData.js
```diff
@@ -21,7 +21,7 @@ import { get } from '../api';
 
 const processMeasureInfo = ({ serieses, measureData, ...rest }) => {
   const processedSerieses = serieses.map(series => {
-    const { values: mapOptionValues, type, scaleType } = series;
```
tcaiger, line 24: scaleType was unused

packages/lesmis/src/components/DashboardReportModal.js
```diff
@@ -91,30 +92,31 @@ export const DashboardReportModal = ({
...
 
   const handleClickOpen = () => {
     setOpen(true);
+    setParams({
+      reportId,
+    });
   };
 
   const handleClose = () => {
     setOpen(false);
```
tcaiger, line 114: clear the reportId, start and end date when closing the modal

packages/lesmis/src/components/DashboardReportModal.js
```diff
@@ -91,30 +92,31 @@ export const DashboardReportModal = ({
...
-      startDate,
-      endDate,
+      startDate: newStartDate,
+      endDate: newEndDate,
     });
   };
 
   const handleClickOpen = () => {
     setOpen(true);
+    setParams({
```
tcaiger, line 108: set the report id to the url when opening the modal

packages/lesmis/src/components/LocationHeader.js
```diff
@@ -123,15 +124,15 @@ export const LocationHeader = () => {
             <ToggleButton
               value="dashboard"
               component={Link}
-              to={makeEntityLink(entityCode, 'dashboard')}
+              to={`${makeEntityLink(entityCode, 'dashboard')}${search}`}
```
tcaiger, line 127: add the url params in the dashboard and map tab links so that it remembers where you are when navigating between the tabs

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -102,15 +97,20 @@ export const DashboardReportTabView = ({ entityCode, TabSelector }) => {
...
+  };
+};
 
-    if (data) {
-      setDefaultDashboard(data, setSelectedDashboard);
-    }
-  }, [data, setSelectedDashboard]);
+export const DashboardReportTabView = ({ entityCode, TabSelector }) => {
+  const [selectedYear, setSelectedYear] = useUrlSearchParam('year', DEFAULT_DATA_YEAR);
+  const [selectedDashboard, setSelectedDashboard] = useUrlSearchParam('dashboard');
```
tcaiger, line 110: @reviewer I thought about adding the url params as constants somewhere but decided that the cons outweighed the pros mostly because each one is only used twice. I am open to a suggestion to change that though.

packages/lesmis/src/views/DashboardView.js
```diff
@@ -483,14 +483,14 @@ const VitalsView = ({ vitals }) => {
...
   const tabOptions = makeTabOptions(entityData?.type);
+  const [params, setParams] = useUrlSearchParams();
 
-  const [selectedTab, setSelectedTab] = useState(tabOptions[0].value);
+  const vitals = useVitalsData(entityCode);
+  const selectedTab = params.dashboardTab || tabOptions[0].value;
 
   const handleChangeTab = event => {
-    setSelectedTab(event.target.value);
+    setParams({ dashboardTab: event.target.value, dashboard: null, year: null });
```
tcaiger, line 493: @reviewer I would be interested to hear your thoughts on `dashboard`and `dashboardTab` as the names of the params.

packages/lesmis/src/views/MapView.js
```diff
@@ -71,7 +71,7 @@ const getDefaultTileSet = () => {
 
 export const MapView = () => {
   const { entityCode } = useUrlParams();
-  const [selectedYear, setSelectedYear] = useState(DEFAULT_DATA_YEAR);
+  const [selectedYear, setSelectedYear] = useUrlSearchParam('year', DEFAULT_DATA_YEAR);
```
tcaiger, line 74: @reviewer I would be interested to get your input on the name `year` for this param. I thought about something more descriptive such as `dashboardYear` but the year is used for the map and the dashboard so I left it as year.

packages/lesmis/src/components/LocationHeader.js
edmofro: And I guess this can't live within `makeEntityLink` because that's used in other cases where we want to clear the search?

packages/lesmis/src/utils/useUrlSearchParams.js
```diff
@@ -8,18 +8,39 @@ import { useHistory, useLocation } from 'react-router-dom';
...
-    history.push({ ...history.location, search: params.toString() });
+
+    if (location.search !== urlParams.toString()) {
+      history.push({ ...location, search: urlParams.toString() });
+    }
   };
 
+  const params = {};
+
+  urlParams.forEach((value, key) => {
```
edmofro, line 29: What kind of data structure is `urlParams`? Is the iterator just a weird custom thing?

packages/lesmis/src/utils/useUrlSearchParams.js
```diff
@@ -8,18 +8,39 @@ import { useHistory, useLocation } from 'react-router-dom';
...
+    params[key] = value;
+  });
+
   return [params, setParams];
 };
+
+export const useUrlSearchParam = (param, defaultValue = null) => {
```
edmofro, line 36: Tidy as hook!!

packages/lesmis/src/views/MapView.js
edmofro: It fully depends on whether the "global year selector" is a shared state across dashboard and map tabs. Personally I would have expected that they were independently controlled, given that they're in quite different places rather than in some common place that doesn't change between the two. For that reason, I would go for `dashboardYear` + `mapYear`. If the state should be shared between dashboard and map overlay, then `year` is the right key.

I'll ask in the channel

packages/lesmis/src/views/DashboardReportTabView.js
edmofro: All good :+1:

packages/lesmis/src/views/DashboardView.js
edmofro: I prefer `dashboard` and `subDashboard`. In fact, when I saw `dashboardTab` I thought it was going to mean these guys
image

rather than these guys
image

Following up on that, though outside the scope of this PR, because the second one is a drop down rather than tabs, can you rename things that refer to them as tabs? E.g. the `makeTabOptions` within this file, which actually makes drop down options.

edmofro (changes requested):
Super clean as you said. Also really happy with a couple of the minor refactors it encouraged.

I was surprised the selected dashboard was in the search rather than url path, but then I realised it's much nicer to keep the selected dashboard when flicking over to the map and back, so great call :+1:

packages/lesmis/src/components/LocationHeader.js
tcaiger: Yes exactly

packages/lesmis/src/utils/useUrlSearchParams.js
tcaiger: I'm not sure I understand the question. [urlSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) is part of the core javascript api it returns an iterable object. An iterator is a javascript thing - it's not custom in the sense that I made it.

edmofro: I guess my question is was why the parameters are `value, key`, which is quite unusual. Thanks for the link to the doc - it seems like the "natural" iterator over the data structure is the more expected `([key, value])` (like Object.entries), but `.forEach` is `(value, key)`. Strange that it would have different formats of iterable depending on which `for` you use, but I guess that's javascript!

packages/lesmis/src/views/MapView.js
edmofro: https://beyondessential.slack.com/archives/G01NEKFK76C/p1621813723006100?thread_ts=1621566541.000300&cid=G01NEKFK76C

Go for `year` :-)

packages/lesmis/src/components/DashboardReportModal.js
```diff
@@ -91,30 +92,31 @@ export const DashboardReportModal = ({
...
-    setParams({
-      startDate,
-      endDate,
+      startDate: newStartDate,
+      endDate: newEndDate,
     });
   };
 
   const handleClickOpen = () => {
     setOpen(true);
```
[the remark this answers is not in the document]
tcaiger: That's a good point. I tried refactoring the open state to be purely based on the presence of the reportId but then realised that it created a lag in the modal opening because the modal waits for the url to be set and then parent components to be re-rendered that use url location params before opening. I then went down a huge rabbit hole trying to optimise the render flow from urls with minimal success. Overall I've left the `open` in state.

edmofro (approved):

PR #2580, opened 2021-05-19
Add script for full rebuild of materialized view

edmofro:
Allows dropping and rebuilding the materialized view logs and analytics table


packages/data-api/scripts/dropAnalyticsMaterializedView.sql
```diff
@@ -0,0 +1,39 @@
...
+
+  RAISE NOTICE 'Dropping Materialized View Logs...';
+  FOREACH source_table IN ARRAY source_tables_array LOOP
+    IF (SELECT EXISTS (
+      SELECT FROM pg_tables
+      WHERE schemaname = 'public'
+      AND tablename   = 'log$_' || source_table
+    ))
+    THEN
+      EXECUTE 'DROP TRIGGER IF EXISTS ' || source_table || '_trigger' || ' ON ' || source_table;
```
rohan-bes, line 30: Hmmm I know this is my fault since I wrote the original code you probably based yours off here, but I suppose we ought to be using disable/re-enable trigger here rather than drop/re-create?

packages/data-api/scripts/dropAnalyticsMaterializedView.sql
```diff
@@ -0,0 +1,39 @@
+do $$
+declare
+  tStartTime TIMESTAMP;
+  source_table TEXT;
+  source_tables_array TEXT[] := array['answer', 'survey_response', 'entity', 'survey', 'question', 'data_source'];
```
rohan-bes, line 5: Not asking for this to happen right now, but it would be good for us in the future to consolidate some of this logic with the createAnalyticsMaterializedView script to keep things DRY and avoid possible bugs 👍

rohan-bes (approved):
Minor comments but don't feel pressed to address them so approving 👍 

Huge thanks for implementing this btw, I really wanted to add a lot more QoL tools for the analytics table, I think it can make our lives a lot easier 👍

packages/data-api/scripts/dropAnalyticsMaterializedView.sql
edmofro: Yep that's true. Not going to address right now

packages/data-api/scripts/dropAnalyticsMaterializedView.sql
edmofro: Aaah that's a great point. I should have thought of this! It would have even saved me the confusion of working out what these new triggers you were creating were...realised after a minute they were actually just the old triggers, and that you'd dropped them a couple of lines up, which made sense but will be clearer for future devs as a disable/re-enable pattern. I will do this now on both scripts!

packages/data-api/scripts/createAnalyticsMaterializedView.sql
```diff
@@ -41,11 +41,11 @@ begin
       AND tablename   = 'log$_' || source_table
     ))
     THEN
-      EXECUTE 'DROP TRIGGER IF EXISTS ' || source_table || '_trigger' || ' ON ' || source_table;
+      EXECUTE 'ALTER TABLE ' || source_table || ' DISABLE TRIGGER ' || source_table || '_trigger';
```
edmofro, line 44: There doesn't seem to be a concise way of checking whether a trigger exists before attempting to disable/enable it. Fortunately, all of the source tables have the trigger on them, but this would be an issue if we added any source tables that didn't in future.

rohan-bes: Ah that's annoying... seems like it throws an error when attempting to disable a non-existent trigger... I guess we can just cross this bridge if and when we come to it 👍

rohan-bes (approved):
Thanks for changing both scripts!

PR #2581, opened 2021-05-20
Standardised ENTITY_API_URL env var across packages

rohan-bes:
- Added `ENTITY_API_URL` to `report-server` and `web-config-server` `.env.examples` files (as they are the servers that directly call `Aggregator` at the moment)

Note: The naming of this `.env` var isn't entirely consistent between our packages. I believe `lesmis-server` still refers to it as `ENTITY_SERVER_API_URL`. 

@tcaiger are you okay with me changing that `.env` var name in `lesmis-sever` as well?


kael89 (approved):
Yay for consistency!

kael89 (approved):

PR #2570, opened 2021-05-17
2806: PSSS - Get syndrome reports

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2806

PR #2582, opened 2021-05-20
2806: PSSS - Get syndrome reports

kael89:
Same as https://github.com/beyondessential/tupaia/pull/2570, but rebased on top of the feature branch, so that it can be merged independently

@rohan-bes the first commit (https://github.com/beyondessential/tupaia/commit/bb1816801db8739ec944e6196a6c64db0f1b93dd) is new, I picked it from the [original base branch](https://github.com/beyondessential/tupaia/pull/2565/files) because it's more relevant to this PR

Everything else should be the same as the original PR


packages/psss-server/src/routes/alerts/FetchAlertsRoute.ts
```diff
@@ -5,38 +5,68 @@
...
+type AlertReportRow = {
+  organisationUnit: string;
+  period: string;
+  syndrome: string;
+};
+
+function validateStatus(category: string): asserts category is AlertCategory {
+  const categories = Object.keys(CATEGORY_TO_REPORT);
+  if (!categories.includes(category)) {
+    throw new Error(`Invalid alert category '${category}', must be one of ${categories}`);
```
rohan-bes, line 24: Nice validation and type assertion!

packages/psss-server/src/routes/alerts/FetchAlertsRoute.ts
```diff
@@ -5,38 +5,68 @@
...
+
+  private fetchAlertData = async () => {
+    const { startWeek, endWeek, orgUnitCodes } = this.req.query;
+    const { category } = this.req.params;
+
+    validateStatus(category);
+    const alertReport = CATEGORY_TO_REPORT[category];
+    const { results } = await this.reportConnection.fetchReport(
+      alertReport,
+      orgUnitCodes?.split(','),
```
rohan-bes, line 55: Wonder if `fetchReport` should be responsible for doing the split... seems like I see this code a lot... but no biggie 👍

packages/psss-server/src/routes/reports/helpers.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { SYNDROMES } from '../../constants';
+
+export const validateSyndrome = (syndrome: unknown) => {
```
rohan-bes, line 8:
```suggestion
export const validateSyndrome = (syndrome: unknown): asserts syndrome is string => {
```
Could even go further and start using string literal types for the syndromes... personally I think it could be really useful but would be a bit of an overhaul at the moment hehe

rohan-bes (approved):
Looks really good! Minor comments but approving anyway 👍

packages/psss-server/src/routes/alerts/FetchAlertsRoute.ts
kael89: It is true that at the moment, we are splitting the string input in the routes, and then joining it again in the connection, which would seem "redundant" at a first glance.

On the other hand, I like the fact that communication between `psss-server` classes is done using TS objects, even if it means that we eventually go from `A => B => A`. It probably makes for cleaner interfaces. Then, the only places were objects are (de)stringified are in the package's "seams" (input and output of the package)

packages/psss-server/src/routes/reports/helpers.ts
kael89: We could do that, but to be honest if feels like a bit of an overkill, at least at this stage... Happy to consider it again in the future!

PR #2584, opened 2021-05-20
2683 map overlay deep linking

tcaiger:
### Changes:

- add map overlay deep linking

---

### DIFF
--- a/packages/lesmis/src/api/queries/useMapOverlayReportData.js
+++ b/packages/lesmis/src/api/queries/useMapOverlayReportData.js
@@ -143,11 +143,6 @@ export const useMapOverlayReportData = ({ entityCode, year }) => {
     setHiddenValues({});
   }, [setHiddenValues, selectedOverlay, entityCode]);
 
-  // reset selected overlay when changing entity
-  useEffect(() => {
-    setSelectedOverlay(null);
-  }, [entityCode]); // setSelectedOverlay = infinite loop
-
   // set default hidden measures when measure data changes
   useEffect(() => {
     const series = measureData ? measureData.serieses : [];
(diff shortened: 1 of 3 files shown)


packages/lesmis/src/api/queries/useMapOverlayReportData.js
```diff
@@ -142,11 +143,6 @@ export const useMapOverlayReportData = ({ entityCode, year }) => {
     setHiddenValues({});
   }, [setHiddenValues, selectedOverlay, entityCode]);
 
-  // reset selected overlay when changing entity
-  useEffect(() => {
-    setSelectedOverlay(null);
-  }, [setSelectedOverlay, entityCode]);
```
edmofro, line 148: Is removing this fine because when changing entity it changes the url to that entity's base url (no search) anyway?

packages/lesmis/src/components/MapOverlaysPanel/MapOverlaysPanel.js
```diff
@@ -32,6 +32,29 @@ const Box = styled(MuiBox)`
   margin-left: 1rem;
 `;
 
+const searchOverlays = (result, child, path, selectedOverlay) => {
```
edmofro, line 35: What was it that we worked on together that was super similar to this?

edmofro (commented):
In the earlier PR you deferred handling bad URLs to this one, but I can't see anything towards that. Did you test and it all falls into place just fine? E.g. going to a url with a dashboard that doesn't exist falls back to the default, etc.?

packages/lesmis/src/api/queries/useMapOverlayReportData.js
tcaiger: yes exactly!

packages/lesmis/src/components/MapOverlaysPanel/MapOverlaysPanel.js
tcaiger: It was processing overlays to only show the visible overlays for each entity level. It's the same nested data structure which is why the recursive functions are quite similar.

edmofro (approved):

PR #2590, opened 2021-05-21
2827: Mark Alert as archived

chris-bes:
### Changes:

- Mark Alert as archived


packages/psss-server/src/routes/alerts/ArchiveAlertRoute.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Route } from '../Route';
+
+export class ArchiveAlertRoute extends Route {
+  async buildResponse() {
+    const { alertId } = this.req.params;
+    const alertSurveyResponse = await this.meditrakConnection?.findSurveyResponseById(alertId);
+
+    if (!alertSurveyResponse) {
+      throw new RespondingError('Alert cannot be found', 500);
+    }
```
chris-bes, line 15: We can also check if the alert is already archived by fetching the Active or Archived Alerts report. But not sure if that's super necessary

packages/psss-server/src/app/addRoutesToApp.ts
```diff
@@ -74,6 +75,7 @@ export function addRoutesToApp(app: Express) {
    * PUT routes
    */
   app.put('/v1/weeklyReport/:countryCode/:siteCode?', handleWith(SaveWeeklyReportRoute));
+  app.put('/v1/alerts/:alertId/archive', handleWith(ArchiveAlertRoute));
```
chris-bes, line 78: Potentially we can change this endpoint to `/v1/alerts/:alertId/:state`, and the Route class can handle changing the state to outbreak/archive/activeAlert

packages/psss-server/src/routes/alerts/FetchAlertsRoute.ts
```diff
@@ -64,7 +65,7 @@ export class FetchAlertsRoute extends Route {
     const { results } = await this.reportConnection.fetchReport(
       `PSSS_${syndrome}_Confirmed_Report`,
       [organisationUnit],
-      [period],
+      [period, getCurrentPeriod('WEEK')],
```
chris-bes, line 68: To make `Cases since alert began` cumulative til today. I think it makes more sense to always sum all the cases til today instead of to the selected week

kael89: Good point!

packages/psss/src/api/queries/useAlerts.js
```diff
@@ -25,3 +26,12 @@ export const useAlerts = (period, orgUnitCodes, alertCategory) => {
     data,
   };
 };
+
+export const useArchiveAlert = alertId =>
```
kael89, line 30: To my knowledge, `get` requests are currently defined under the `queries` folder, and all  `post/put/delete` under `mutations.js`. Unless if there are any "hidden" mutations I'm not aware of, it seems that `useMutation` is not used in `queries`.

I think I like what you did here: grouping queries by resource (eg `alerts`) rather than by operation (eg `get`/`post`). It still feels tempting to make things consistent one way or the other, but I'll leave this decision to you

packages/psss/src/components/DropdownMenu.js
```diff
@@ -96,14 +96,17 @@ const MenuItem = styled(MuiMenuItem)`
...
+export const DropdownMenu = ({ options, onChange, readOnly }) => {
   const [open, setOpen] = useState(false);
   const [value, setValue] = useState((options[0] || {}).value);
 
   const handleMenuItemClick = (event, option) => {
     event.preventDefault();
     event.stopPropagation();
-    setValue(option.value);
+    if (!readOnly) {
+      // sometimes we just want to treat DropdownMenu as Actions Menu
```
kael89, line 107: Good comment!

packages/psss/src/components/DropdownMenu.js
```diff
@@ -153,4 +156,5 @@ DropdownMenu.propTypes = {
     }),
   ).isRequired,
   onChange: PropTypes.func.isRequired,
+  readOnly: PropTypes.bool,
```
kael89, line 159: Let's add a default prop for that (I have a feeling that you linter is disabled 😃 )

packages/psss/src/containers/Panels/AlertsPanel.js
```diff
@@ -88,24 +88,37 @@ AlertsPanelProvider.propTypes = {
...
-  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [isArchiveAlertModalOpen, setIsArchiveAlertModalOpen] = useState(false);
+  const [isCreateOutbreakModalOpen, setIsCreateOutBreakModalOpen] = useState(false);
   const [activeIndex, setActiveIndex] = useState(0);
   const notesState = useFetch(getAlertsMessages);
   const activityState = useFetch(getActivityFeed);
   const countryName = useSelector(state => getCountryName(state, countryCode));
 
   const handleChange = option => {
     // TODO handle changes other than creating an outbreak
```
kael89, line 101: Perhaps we need to update this comment to "TODO allow outbreaks to be created"... Or we can add the case for `Outbreaks` - users shouldn't be able to select this option via the UI anyway

packages/psss/src/containers/Panels/AlertsPanel.js
```diff
@@ -88,24 +88,37 @@ AlertsPanelProvider.propTypes = {
...
+        break;
+      default:
+    }
   };
 
   const handleClose = useCallback(() => {
     setIsOpen(false);
   }, [setIsOpen]);
 
+  const handleCreateOutbreakModalCLose = useCallback(() => {
```
kael89, line 114: I'd probably rename this (and the other handler below) as follows, but up to you

```suggestion
  const handleCloseCreateOutbreakModal = useCallback(() => {
```

packages/psss/src/containers/Modals/ConfirmModal.js
```diff
@@ -0,0 +1,89 @@
...
+  font-size: 2.5rem;
+  margin-bottom: 0.3rem;
+  color: ${props => props.theme.palette.primary.main};
+`;
+
+const PositionedAlert = styled(Alert)`
+  margin-top: 1rem;
+`;
+
+export const ConfirmModal = ({
```
kael89, line 37: This is a great, generic component!! It could potentially be used in place of similar components (such as `AlertCreateModal`), mainly for styling/structural consistency. No need to do so now, of course

packages/psss/src/containers/Modals/ConfirmModal.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+ConfirmModal.propTypes = {
+  isOpen: PropTypes.bool.isRequired,
+  handleClose: PropTypes.func.isRequired,
+  isLoading: PropTypes.bool.isRequired,
+  title: PropTypes.string.isRequired,
+  mainText: PropTypes.string.isRequired,
+  description: PropTypes.string,
+  error: PropTypes.string,
```
kael89, line 85: Let's add defaultProps for these fields

kael89 (approved):
Everything looks great @chris-bes ! I have some minor suggestions but none of them are blockers, so I'm gonna go ahead and approve this 🎉

packages/psss-server/src/routes/reports/ConfirmWeeklyReportRoute.ts
```diff
@@ -128,35 +126,30 @@ export class ConfirmWeeklyReportRoute extends Route {
...
     return response;
   }
 
   async createAlert(countryCode: string, week: string, syndromeCode: string) {
-    return this.meditrakConnection?.createSurveyResponse(ALERT_SURVEY, countryCode, week, {
-      PSSS_Alert_Syndrome: syndromeCode,
-      PSSS_Alert_Archived: 'No',
-    });
+    return this.meditrakConnection?.createSurveyResponse(ALERT_SURVEY, countryCode, week, [
+      { code: 'PSSS_Alert_Syndrome', type: 'Radio', value: syndromeCode },
```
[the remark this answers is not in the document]
edmofro: Good question @kael89! Yes, I think you're probably right - I have a vague memory that in the early days of meditrak we were changing questions all the time before things stabilised a bit, and wanted to know what type of answer it was even if the question type had changed. I can't be sure though as it was years ago and my memory is hazy - it may have been a simple denormalisation for a performance or code simplicity benefit. Given the way we use the data in that table now, it would be bad for lots of reasons if the answer type was different to the question type, so I totally agree that we should be able to remove this column.

PR #2593, opened 2021-05-23
Average aggregation report server

billli0:
- Add average aggregation in report server
- A bit refactor


billli0:
In this PR I move validations from dataElements and dataGroups together to a new file, but it seems like the typescript’s assertions didn’t work in a separate file...

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregate.ts
```diff
@@ -17,31 +17,73 @@ type AggregateParams = {
...
+  });
+  return mergedRow;
+};
+
+const getAggregatedRows = (groupRowFields: GroupRowFields, params: AggregateParams): Row[] => {
+  const aggregatedRowFields: AggregatedRowFields = {};
+  Object.keys(groupRowFields).forEach((groupKey: string) => {
+    Object.keys(groupRowFields[groupKey]).forEach((fieldKey: string) => {
+      const aggregation = params.getFieldAggregation(fieldKey);
+      const fieldValue: FieldValue = aggregations[aggregation](groupRowFields[groupKey][fieldKey]);
```
rohan-bes, line 71:
```suggestion
      const aggregatedValue = aggregations[aggregation](groupRowFields[groupKey][fieldKey]);
```
Might be worth confirming this, but a lot of the time `Typescript` can infer what type a constant will be when you initialize it, so no need to explicitly give it a type 👍 

As for the variable naming, this is a bit more subjective but personally I'm a big fan of descriptive names instead of 'type-like' names

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
-import { Row, FieldValue } from '../../../types';
+import { FieldValue } from '../../../types';
 
 const isUndefined = (value: FieldValue): value is undefined => {
   return value === undefined;
 };
 
-const group = (existingRow: Row, field: string, value: FieldValue) => {
-  existingRow[field] = value;
+const filterUndefinedAndNull = (values: FieldValue[]): NonNullable<FieldValue[]> => {
```
rohan-bes, line 12: Ah I never knew about the `NonNullable` generic type, so handy!

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
+  const filteredValues: NonNullable<FieldValue[]> = [];
+  values.forEach(value => {
+    if (value !== undefined && value !== null) {
+      filteredValues.push(value);
+    }
+  });
+  return filteredValues;
```
rohan-bes, line 19: With a function like this, I'd argue it's better to use the standard `Array.filter` function. Something like this:
```
const isNotNull = <T>(value: T): value is NotNullable<T> => value !== undefined && value !== null;
``` 
^here we are using a generic function to accept any type, and return a boolean to say whether the type is not null

and then later it can be used like:
```
const notNullValues = values.filter(isNotNull);
```

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
+const checkIsNum = (values: FieldValue[]): number[] => {
+  const filteredUndefinedAndNullValues = filterUndefinedAndNull(values);
+  const checkedValues: number[] = [];
+  filteredUndefinedAndNullValues.forEach(value => {
+    if (typeof value !== 'number') {
+      throw new Error(`Expected number, got '${typeof value}'.`);
+    }
+    checkedValues.push(value);
+  });
+  return checkedValues;
```
rohan-bes, line 31: Similarly here, I'd prefer we didn't accept an array of values but rather we used something like:
```
const assertIsNumber = (value: FieldValue): assert value is number => {
  if (typeof value !== 'number') {
    throw new Error(`Expected number, got '${typeof value}'.`);
  }
}
```
and then later we can do:
```
values.forEach(assertIsNumber);
```

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
 };
 
-const count = (existingRow: Row, field: string, value: FieldValue) => {
-  existingRow[field] = ((existingRow[field] as number) || 0) + 1;
+const sum = (values: FieldValue[]): number | undefined => {
+  if (values.length !== 0) {
+    const checkedValues: number[] = checkIsNum(values);
+    return checkedValues.reduce((a, b) => {
+      return a + b;
+    });
```
rohan-bes, line 46: I do love me a map reduce 😍

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
-  const existingValue: FieldValue = existingRow[field];
-  if (!isUndefined(value)) {
-    if (isUndefined(existingValue)) {
-      existingRow[field] = value;
-    } else if (value > existingValue) {
-      existingRow[field] = value;
-    }
+const avg = (values: FieldValue[]): number | undefined => {
+  const numerator: number | undefined = sum(values);
+  const denominator: number = count(values);
```
rohan-bes, line 53: Clever use of other aggregate functions! 👍

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
+  });
+  return checkedValues;
 };
 
-const sum = (existingRow: Row, field: string, value: FieldValue) => {
-  if (typeof value === 'number') {
-    existingRow[field] = ((existingRow[field] as number) || 0) + value;
-  } else {
-    throw new Error(`Expected number, got '${typeof value}'.`);
+const group = (values: FieldValue[]): FieldValue => {
```
rohan-bes, line 34: I think it's actually true that we want to filtering the values for all null/undefined for all aggregations (not just sum, etc.). I hadn't implemented this originally and had just been fixing it on an adhoc basis, but now that someone's taken the time to re-write this it would be nice to prevent future bugs. 👍

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
+  return values.length;
+};
+
+const max = (values: FieldValue[]): FieldValue => {
+  if (values.length !== 0) {
+    let maxValue: FieldValue = values[0];
+    values.forEach(value => {
+      if (!isUndefined(value) && !isUndefined(maxValue) && value > maxValue) maxValue = value;
+    });
+    return maxValue;
```
rohan-bes, line 70:
```suggestion
    return Math.max(...values.filter(isNotNull));
```

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
+  return undefined;
+};
+
+const min = (values: FieldValue[]): FieldValue => {
+  if (values.length !== 0) {
+    let minValue: FieldValue = values[0];
+    values.forEach(value => {
+      if (!isUndefined(value) && !isUndefined(minValue) && value < minValue) minValue = value;
+    });
+    return minValue;
```
rohan-bes, line 81: Same as above but with `Math.min`

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
-  } else {
-    existingRow[field] = value;
+const unique = (values: FieldValue[]): FieldValue => {
+  if (values.length !== 0) {
+    for (let i = 0; i < values.length - 1; i++) {
+      if (values[i] !== values[i + 1]) {
+        return 'NO_UNIQUE_VALUE';
+      }
+    }
+    return values[0];
```
rohan-bes, line 93:
```suggestion
    const distinctValues = [...new Set(values.filter(isNotNull))];
    return distinctValues.length === 1 ? distinctValues[0] : 'NO_UNIQUE_VALUE';
```

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
-const first = (existingRow: Row, field: string, value: FieldValue) => {
-  if (isUndefined(existingRow[field])) {
-    existingRow[field] = value;
+const first = (values: FieldValue[]): FieldValue => {
+  for (let i = 0; i < values.length; i++) {
+    if (!isUndefined(values[i])) {
+      return values[i];
+    }
   }
+  return undefined;
```
rohan-bes, line 108:
```suggestion
  return values.find(isNotNull);
```
You may have noticed that I _hate_ old-fashioned for loops 😝

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,124 @@
...
 
-const last = (existingRow: Row, field: string, value: FieldValue) => {
-  existingRow[field] = value;
+const last = (values: FieldValue[]): FieldValue => {
+  for (let i = values.length; i > 0; i--) {
+    if (!isUndefined(values[i])) {
+      return values[i];
+    }
+  }
+  return undefined;
```
rohan-bes, line 117:
```suggestion
  return values.slice().reverse().find(isNotNull);
```

rohan-bes (changes requested):
Few changes requested, but mainly superficial things, structure is looking really good!! Hopefully in the future we can port these aggregations to expression-parser so that we can make use of them in other parts of the code!!

Ps. One more thing, please add unit tests for `avg` aggregation in `report-server/src/__tests__/reportBuilder/transform/aggregate.test.ts` 🙏

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregate.ts
```diff
@@ -17,31 +17,73 @@ type AggregateParams = {
...
-const aggregate = (rows: Row[], params: AggregateParams): Row[] => {
-  const groupedRows: { [groupKey: string]: Row } = {};
-  const otherRows: Row[] = [];
+type GroupRowFields = {
+  [groupKey: string]: RowFields;
+};
 
+type AggregatedFields = {
+  [fieldKey: string]: FieldValue;
+};
```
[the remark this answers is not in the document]
billli0: Oh yeah you are right, nice catch 👍

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregate.ts
```diff
@@ -17,31 +17,73 @@ type AggregateParams = {
...
+  Object.keys(newRow).forEach((field: string) => {
+    mergedRow[field] = [...(mergedRow[field] || []), newRow[field]];
+  });
+  return mergedRow;
+};
+
+const getAggregatedRows = (groupRowFields: GroupRowFields, params: AggregateParams): Row[] => {
+  const aggregatedRowFields: AggregatedRowFields = {};
+  Object.keys(groupRowFields).forEach((groupKey: string) => {
+    Object.keys(groupRowFields[groupKey]).forEach((fieldKey: string) => {
```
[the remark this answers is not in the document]
billli0: Good point! I did feel a bit unpleasant when I am having a double look at it, after the changes it looks good.

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregate.ts
billli0: Cool stuff, I did it in this PR for a lot of times, I will delete them.

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
billli0: To me filtering the null/undefined value is to prevent the calculation error, I didn't do it on `group` aggregation is because we might want to send `null` as value in some special cases. 
What I think is we can filter the `null` by `filter` transform, but you can show them if you want. With that we can provide more freedom.

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
billli0: I feel like we want to compare string as well? e.g. for period `20210101`, [link to test case](https://github.com/beyondessential/tupaia/blob/3f3f0152152401d6897e4501e1d24f745a624c20/packages/report-server/src/__tests__/reportBuilder/transform/aggregate.test.ts#L107)

but we can do a Array.sort instead of using retro loop 😁

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
billli0: Sorry I just notice I filter undefined for every aggregations... But I still think not to filter them out unless we do calculation. @rohan-bes What do you think?

rohan-bes: >To me filtering the null/undefined value is to prevent the calculation error, I didn't do it on group aggregation is because we might want to send null as value in some special cases.
>
>I think we can filter the null by filter transform, but you can show them if you want. With that we can provide more freedom.

I think you make a really excellent point here! I think you have totally convinced me 🎉 you are correct that it's a potential valid case that we may want to `group` all rows with `null` for a specific field together (and not just `group`, also: `last`, `unique`, etc.)!

How does this sound:
- Always filter out `undefined`. `undefined` is fairly consistently treated as that property not existing on the row, both in javascript as a whole **_AND_**  in BES
- Keep `null`, but filter when doing mathematical calculations as you have done. `null` is treated as 'We know we have no data for this field in this row'. It's a slight distinction, but may be important, and shares consistency with the new 'fill missing values with `null`' logic you implemented in `aggregator`

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
rohan-bes: Ah, great point! Tbh I was a bit lazy and had assumed that Math.max/min supported string compare as well, which is wrong!

You are correct, and I like the idea of using `Array.sort`. Couple things to keep in mind is that we need to handle `null`s. A good question could be: should `min` return `null` if it's present or the lowest non-null value that exists..? I honestly am not sure, keen to hear your thoughts. I think for now either way is acceptable, as using `null` in our system is still fairly uncommon...

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
billli0: Yeah I agree we keep that consistency, and as I remember you already confirmed this logic with the teams, and the logic you commented above looks good to me 👍  so I will apply this to the code.

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
billli0: Thanks Rohan, I am on your side that it is uncommon to return `null` as minimum value, I think we should make the logic consistent as below to return `null` as min value, the trade off is we need to filter them before the calculation transform, which shouldn't be too hard. 

One case that did happen is we want to get min from array [1, 2, 3, null] -> 1, and array [null, null, null, null] -> null. One way to do it is to eliminate all null value in report server, and then use the fill emtpy period in aggregator.

rohan-bes: >One way to do it is to eliminate all null value in report server, and then use the fill emtpy period in aggregator.

I think we will need to expect the report-server to handle `null`s. Going forward, client's wont be able to query the aggregator directly at all, so all data will be coming from the report-server.

>One case that did happen is we want to get min from array [1, 2, 3, null] -> 1, and array [null, null, null, null] -> null. 

That's a good point though, but if it comes down to it, we can always support multiple `min` aggregations, eg. `min` and `minNotNull` so that user can choose which one to use when 👍 

For now, I think I'm slightly leaning towards `min` being `min([1,2,3,null]) => null` as this seems to be the most basic function, but am probably happy either way 😄

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
billli0: Love the Set!

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
billli0: That's a good point, we can do min() and minNotNull() in the future.

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregate.ts
```diff
@@ -17,31 +17,69 @@ type AggregateParams = {
...
+      const aggregatedValue = aggregations[aggregation](fieldValue);
+      if (aggregatedValue !== undefined) {
+        aggregatedRowFields[groupKey] = {
+          ...aggregatedRowFields[groupKey],
+          [fieldKey]: aggregatedValue,
+        };
+      }
+    });
+  });
+  return Object.values(aggregatedRowFields);
```
rohan-bes, line 76: Looks a lot cleaner here, love it!

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,102 @@
...
 
-const isUndefined = (value: FieldValue): value is undefined => {
-  return value === undefined;
+const isNotUndefined = <T>(value: T): value is Exclude<T, undefined> => {
+  return value !== undefined;
 };
 
-const group = (existingRow: Row, field: string, value: FieldValue) => {
-  existingRow[field] = value;
+const isNotNullAndUndefined = <T>(value: T): value is NonNullable<T> => {
```
rohan-bes, line 12:
```suggestion
const isNotNullOrUndefined = <T>(value: T): value is NonNullable<T> => {
```
or
```suggestion
const isNotNullAndNotUndefined = <T>(value: T): value is NonNullable<T> => {
```

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -3,75 +3,102 @@
...
-    existingRow[field] = value;
-  }
+const first = (values: FieldValue[]): FieldValue => {
+  return values.find(isNotUndefined);
 };
 
-const last = (existingRow: Row, field: string, value: FieldValue) => {
-  existingRow[field] = value;
+const last = (values: FieldValue[]): FieldValue => {
+  return values.reverse().find(isNotUndefined);
```
rohan-bes, line 95: You gotta be a bit careful about calling `reverse` on an array, as it actually changes the data in the base array (as opposed to something like `map` that creates a copy of the base array with the values mapped)

Usually people do something like this `array.slice().reverse()` as `slice()` creates a copy of the array 👍

rohan-bes (approved):
Great work @billli0, really excited about these changes as being able to use aggregations in report-server is a big deal!

Also nice refactor of the `aggregate` code, I think you've done a good job setting a precedent for how we can handle `null` vs `undefined` 👍 

Couple of small changes requested, but nothing major so giving an approval anyway.

packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
billli0: Sorry to miss one of your reviews that using `slice()`.. Thank you to point it out again and share the good practice!

PR #2594, opened 2021-05-24
2803 PSSS - Sort alert columns

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2803


packages/ui-components/src/components/Table/TableHeader.js
```diff
@@ -71,10 +72,13 @@ export const TableHeader = React.memo(
       }
 
       if (sortable) {
+        const isActive = key === orderBy;
+        const IconComponent = isActive ? KeyboardArrowDownIcon : UnfoldMoreIcon;
```
tcaiger, line 76: nice clean solution and use of material-ui

packages/psss/src/api/queries/useAlerts.js
```diff
@@ -19,9 +20,14 @@ export const useAlerts = (period, orgUnitCodes, alertCategory) => {
     ...reportRow,
     syndromeName: SYNDROMES[reportRow.syndrome.toUpperCase()],
   }));
+  const sortedData = orderBy(
```
tcaiger, line 23: This is great, it's such a clean way to get multi-sort.

packages/ui-components/src/components/Table/FetchedTable.js
```diff
@@ -0,0 +1,129 @@
...
+        order={order}
+        page={page}
+        rowsPerPage={rowsPerPage}
+        rowIdKey={rowIdKey}
+        isFetching={isFetching}
+        className={className}
+      />
+    );
+  },
+);
```
tcaiger, line 85:
```suggestion

const useSortableTable = (initialData, initialOrderBy, initialOrder) => {
  const [orderBy, setOrderBy] = useState(initialOrderBy);
  const [order, setOrder] = useState(initialOrder);
  const [data, setData] = useState(initialData);

  useEffect(() => {
    const ascending = order !== 'desc';
    const newData = [...initialData].sort(getSortByKey(orderBy, { ascending }));
    setData(newData);
  }, [initialData, orderBy, order]);

  const handleChangeOrderBy = useCallback(
    columnKey => {
      const newOrder = columnKey === orderBy ? toggleOrder(order) : 'asc';
      setOrderBy(columnKey);
      setOrder(newOrder);
    },
    [order, orderBy],
  );

  return {
    handleChangeOrderBy,
    orderBy,
    order,
    data,
  };
};

export const SortableTable = ({ data, orderBy: initialOrderBy, order: initialOrder, ...props }) => {
  const { handleChangeOrderBy, orderBy, order, data: sortedData } = useSortableTable(
    data,
    initialOrderBy,
    initialOrder,
  );
  return (
    <Table
      onChangeOrderBy={handleChangeOrderBy}
      data={sortedData}
      order={order}
      orderBy={orderBy}
      {...props}
    />
  );
};
```

tcaiger: What about extracting the sorting behaviour into a custom hook and then using that to add the sorting functionality to the Table component.

- makes it clearer what is being added
- the hook could be exported and used in apps directly with tables
- don't have to keep updating all the table props. This one is debatable as I know that some people like to list out all the props but I personally think sometimes it's not worth it when we are essentially making a higher order component and using 90% of the original table functionality.

tcaiger: Also what about calling it `SortableTable` rather than `FetchedTable`?

tcaiger (changes requested):
Nice addition to ui-components. In the end I like that the sorting is in ui-components because it makes it easy to simply use the table with sorting. My main suggestion is to rename the component to be `SortingTable` or `SortableTable`. I also left a suggestion about the structure of the component but that is totally optional based on what you prefer.

packages/psss/src/api/queries/useAlerts.js
kael89: Yay for lodash!

packages/ui-components/src/components/Table/FetchedTable.js
kael89: All great points Tom! I love the idea of using a hook - I'll go ahead an implement it!

I also agree with you that in cases like this, it's fine to not list all other props. I did so here to be consistent with other similar components in this package (in specific, with the now deleted `DataFetrchingTable`). Happy to change that!

Finally, I should have added a comment: I called this component `FetchedTable` because I thought that we could also add pagination functionality (I didn't add it now because it is not required). The idea is that `FetchedTable` could have some extra features that are made possible because of the fact that its data are fetched externally, in one go.

I guess that we can implement sorting/pagination as custom hooks, and then we don't need an extra component - clients can use a combination of `Table` + `sorting hook` to achieve the desired functionality.

kael89:
@tcaiger ready for re-review - I followed your advice and used a hook for table sorting!

packages/ui-components/src/hooks/useTableSorting.js
```diff
@@ -0,0 +1,37 @@
...
+  DESC: 'desc',
+};
+
+const toggleOrder = order => (order === DIRECTION.DESC ? DIRECTION.ASC : DIRECTION.DESC);
+
+/**
+ * @param {string | undefined} initialOrderBy (Initial) key of a sorted column
+ * @param {'asc' | 'desc'} initialOrder
+ */
+export const useTableSorting = (data, initialOrderBy, initialOrder = DIRECTION.ASC) => {
```
kael89, line 20: I placed this under a new folder called `hooks`. Not sure if this is better than placing this close to the component it's related to (`Table`). I thought it may be good for hooks and components to be defined in separate locations, to encourage less coupling between them

tcaiger: Nice - good move. I can image there will be more hooks to add over time!

tcaiger (approved):
Looks awesome

PR #2596, opened 2021-05-24
Fix incorrect property key in Alert Panel

chris-bes:
### Changes:
- There's data for total cases per week but it's not showing in the Alert Panel (screenshot here: https://github.com/beyondessential/tupaia-backlog/issues/2851#issuecomment-846803816). The report returns `weeklyCases` instead of `totalCases`. So this is to fix incorrect property key in Alert Panel
- I didn't fix this in another WIP branch and wanted to fix this early because Sepi is testing the Alert Panel

### DIFF
--- a/packages/psss/src/components/AffectedSitesTab.js
+++ b/packages/psss/src/components/AffectedSitesTab.js
@@ -63,7 +63,7 @@ export const AffectedSitesTab = ({ alert }) => {
     }
 
     return weeklyData.map(weekData => {
-      const { status, period, totalCases, percentageChange, sites: siteData } = weekData;
+      const { status, period, weeklyCases, percentageChange, sites: siteData } = weekData;
 
       return (
         <AlertsOutbreaksCard key={period} variant="outlined" mb={5}>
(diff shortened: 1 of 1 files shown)


packages/psss/src/components/AffectedSitesTab.js
```diff
@@ -63,15 +63,15 @@ export const AffectedSitesTab = ({ alert }) => {
     }
 
     return weeklyData.map(weekData => {
-      const { status, period, totalCases, percentageChange, sites: siteData } = weekData;
+      const { status, period, weeklyCases, percentageChange, sites: siteData } = weekData;
```
chris-bes, line 66: We can change the property in the report to be `totalCases` instead. Personally I'm fine with calling it weekly cases so that's why I changed the front end part to make it easier

kael89 (approved):
I feel that the terms have started becoming a bit blurry. On "weekly" vs "total" cases, is the difference the time aspect (_week_) or is about whether data are confirmed or not?

But from a practical standpoint, I'm happy to keep this as is 👍 Thanks for fixing it @chris-bes !

PR #2598, opened 2021-05-24
PSSS: Fix event data placeholder

kael89:
### Issue #:
[No issue] The **Event-based Data** placeholder broke when I refactored the alerts table. After talking with @enunan , I confirmed that the fact that we show alerts in the blurred background is a coincidence. The actual data that will be shown when we implement this feature will not be directly related to alerts.

### Screenshots:
image

### DIFF
--- a/packages/psss/src/containers/Tables/FakeEventBasedDataTable.js
+++ b/packages/psss/src/containers/Tables/FakeEventBasedDataTable.js
@@ -1,0 +1,58 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+import { Table } from '@tupaia/ui-components';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { useAlerts } from '../../api';
+import { AlertMenuCell, SyndromeCell, WeekAndDateCell } from '../../components';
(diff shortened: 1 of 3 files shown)


packages/psss/src/containers/Tables/FakeEventBasedDataTable.js
```diff
@@ -0,0 +1,58 @@
...
+  {
+    title: '',
+    key: 'id',
+    sortable: false,
+    CellComponent: AlertMenuCell,
+    width: '45px',
+  },
+];
+
+export const FakeEventBasedDataTable = ({ countryCode }) => {
```
kael89, line 41: Code here is shamelessly copy-pasted from [AlertsTable.js](https://github.com/beyondessential/tupaia/blob/psss-phase-2/packages/psss/src/containers/Tables/AlertsTable.js), with a few tweaks.

I could instead update `AlertsTable` to accommodate the placeholder's needs, and that wouldn't require many changes. However, I decided it's still not worth doing so just for the needs of a temporary part of the codebase.

rohan-bes: SHAME! 😝

rohan-bes (approved):
Nice, interesting to know what kind of data will eventually be shown here 🤔

PR #2600, opened 2021-05-25
2842: Support 'Public' data access when communicating between micro-services

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2842:

In order to support access to public data between micro-services, we needed to be able to:
- Associate a request with the correct public access policy for the application they are using (ie. tupaia.org has different entities public available than psss.tupaia.org)
- Be able to have micro-services maintain this correct context even if they are calling one another (nested requests)

The solution was to store an `apiClientUserId` in addition to the `userId` in the accessToken of a request after a user authenticates with meditrak-server. When constructing an accessPolicy from that token, we merge the access policies of the `apiClientUser` and the `user`. Since micro-services simply re-use the original request's auth header when requesting from other micro-services, this context will be maintained throughout all nested calls.

- [ ] Add `JWT_SECRET` env var to ci env vars for auth package


rohan-bes:
Unit tests incoming!

packages/auth/src/Authenticator.js
```diff
@@ -17,6 +19,39 @@ export class Authenticator {
...
+    const user = await this.models.user.findById(userId);
+    const userAccessPolicy = await this.getAccessPolicyForUser(user.id);
+
+    if (!apiClientUserId) {
+      return { user, accessPolicy: userAccessPolicy };
+    }
+
+    const apiClientAccessPolicy = await this.getAccessPolicyForUser(apiClientUserId);
+    return { user, accessPolicy: mergeAccessPolicies(userAccessPolicy, apiClientAccessPolicy) };
+  }
```
rohan-bes, line 36: I feel a touch unclear about what a the direct responsibilities of `Authenticator` vs the various authentication middlewares, etc. It seems that decrypting the `authHeader` is up to the middleware, but `Authenticator` handles how we associate users/accessPolicies with the headers... I may not have gotten the separation of concerns quite right here...

packages/auth/src/Authenticator.js
```diff
@@ -17,6 +19,39 @@ export class Authenticator {
...
+
+  /**
+   * Authenticate an api client user
+   * @param {{ username: string, secretKey: string }} apiClientCredentials
+   */
+  async authenticateApiClient({ username, secretKey }) {
```
rohan-bes, line 42: This logic here isn't quite the same as `src/auth/clientAuth.js`. The `clientAuth` version supports the `env.CLIENT_USERNAME, env.CLIENT_SECRET` user credentials (which don't actually exist in the database...).

Personally I'd love to add that api client user to the database as it would allow us to remove such branching logic, but am wary of the regression risk. If the reviewer agrees, I'll create a ticket for it and we can address when there's time 👍 

For now is it acceptable to re-introduce this logic here (basically I've pulled it out from `@tupaia/server-boilerplate`)

packages/auth/src/mergeAccessPolicies.js
```diff
@@ -0,0 +1,31 @@
...
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+/**
+ * Merge policy1 and policy2 together to form a superset access policy
+ * @param {AccessPolicyObject} policy1
+ * @param {AccessPolicyObject} policy2
+ * @returns {AccessPolicyObject} mergedPolicy
+ */
+export const mergeAccessPolicies = (policy1, policy2) => {
```
rohan-bes, line 12: This function _**only**_ works with the modern `AccessPolicyObject` format (rather than the legacy one with permission groups). I don't actually know why and when the legacy access policies are being created... Perhaps some error handling here is required...

packages/auth/src/userAuth.js
```diff
@@ -7,11 +7,38 @@ import jwt from 'jsonwebtoken';
 import { UnauthenticatedError } from '@tupaia/utils';
 import { getJwtToken } from './security';
 
+const ACCESS_TOKEN_EXPIRY_SECONDS = 15 * 60; // User's access expires every 15 mins
+
+export const constructAccessToken = ({ userId, apiClientUserId, refreshToken }) => {
```
rohan-bes, line 12: Moved this out of `meditrak-server` so that creating/decrypting accessTokens is done side by side.

packages/server-boilerplate/src/microService/auth/middleware.ts
```diff
@@ -8,59 +8,54 @@ import { UnauthenticatedError } from '@tupaia/utils';
...
-  if (apiClientUser) {
-    return apiClientUser.id;
+  try {
+    const { accessPolicy } = await authenticator.authenticateApiClient({
+      username,
+      secretKey: password,
+    });
+    return accessPolicy;
+  } catch (error) {
+    // attempting to authenticate as api client failed
```
rohan-bes, line 48: Silently failing here as if the user isn't an `apiClient` then this call will throw and error (that keeps it consistent with the other functions on `Authenticator`)

packages/auth/src/userAuth.js
edmofro: Nice!

packages/meditrak-server/src/tests/routes/authenticate/authenticate.test.js
```diff
@@ -0,0 +1,174 @@
...
+
+    const { entity: demoEntity } = await findOrCreateDummyCountryEntity(models, {
+      code: 'DL',
+    });
+
+    // Create test users
+    userAccount = await findOrCreateDummyRecord(models.user, {
+      first_name: 'Ash',
+      last_name: 'Ketchum',
+      email: 'ash-ketchum@pokemon.org',
```
edmofro, line 49: 😂

packages/meditrak-server/src/tests/routes/authenticate/authenticate.test.js
```diff
@@ -0,0 +1,174 @@
...
+      permission_group_id: publicPermissionGroup.id,
+    });
+
+    await findOrCreateDummyRecord(models.userEntityPermission, {
+      user_id: multiPermissionApiUserAccount.id,
+      entity_id: laosEntity.id,
+      permission_group_id: publicPermissionGroup.id,
+    });
+
+    // Public all other country permissions
```
edmofro, line 116: The comments above are only sometimes present, and this one doesn't seem to relate to the block below, so perhaps they're stale comments?

packages/meditrak-server/src/tests/routes/authenticate/authenticate.test.js
```diff
@@ -0,0 +1,174 @@
...
+    });
+    await findOrCreateDummyRecord(models.userEntityPermission, {
+      user_id: simpleApiUserAccount.id,
+      entity_id: demoEntity.id,
+      permission_group_id: publicPermissionGroup.id,
+    });
+    await findOrCreateDummyRecord(models.userEntityPermission, {
+      user_id: userAccount.id,
+      entity_id: demoEntity.id,
+      permission_group_id: publicPermissionGroup.id,
```
edmofro, line 94: Double up on the first entry?

packages/meditrak-server/src/tests/routes/authenticate/authenticate.test.js
```diff
@@ -0,0 +1,174 @@
...
+    });
+
+    await findOrCreateDummyRecord(models.userEntityPermission, {
+      user_id: multiPermissionApiUserAccount.id,
+      entity_id: kiribatiEntity.id,
+      permission_group_id: publicPermissionGroup.id,
+    });
```
edmofro, line 101: Double up on the last entry?

packages/meditrak-server/src/tests/routes/authenticate/authenticate.test.js
```diff
@@ -0,0 +1,174 @@
...
+
+const userAccountPassword = 'password';
+const simpleApiClientSecret = 'simple';
+const multiPermissionApiClientSecret = 'multiPermission';
+
+let userAccount;
+let simpleApiUserAccount;
+let multiPermissionApiUserAccount;
+
+describe('Authenticate', function () {
```
edmofro, line 27: With all the permissions going on in this, it might be nice to add a summary comment or even assertion that indicates what permissions each account ends up with

packages/auth/src/Authenticator.js
edmofro: Yes totally, after your change to make web-config-server use a db based api client it's just meditrak-app using the one that's in env vars. That's just a bit of laziness because we introduced db based api clients at a high pressure time and didn't bother to add in the existing api client, but please do create a ticket for that (this comment shows we meant to do it pretty quickly, but just never did)
```
  // Still check against environment variables so that these
  // updates can be deployed without breaking the app. This
  // check can be removed once the proper credentials have been
  // added to the production database.
```

packages/auth/src/Authenticator.js
edmofro: Yeah this particular method seems a bit odd, because it doesn't actually check anything. All of the others verify that the authentication requirements are met, and then return the user's details and access policy, whereas this one assumes that the auth requirements have already been met externally. 

More consistent might be:
- Middleware pulls out just the token from the header (i.e. the `xxx` in `Bearer xxx`) 
- Passes that token in here, which calls `getTokenClaims` to validate the token and get the claims out, and then runs the existing content of this method

packages/meditrak-server/src/routes/authenticate.js
```diff
@@ -130,16 +159,26 @@ const checkAuthentication = async req => {
...
   await req.assertPermissions(allowNoPermissions);
 
-  const { refreshToken, user, accessPolicy } = await checkAuthentication(req);
+  const { refreshToken, user, accessPolicy } = await checkUserAuthentication(req);
+  const {
+    user: apiClientUser,
+    accessPolicy: apiClientAccessPolicy,
+  } = await checkApiClientAuthentication(req);
+
+  const mergedAccessPolicy = apiClientUser
```
edmofro, line 168: The idea of merging access policies is a bit odd, so I think it's worth comments in a few places like this
```suggestion
  // Response access policy contains all permissions for the base API client, plus the logged in user
  const mergedAccessPolicy = apiClientUser
```

packages/server-boilerplate/src/microService/auth/middleware.ts
```diff
@@ -79,19 +74,21 @@ export const buildBasicBearerAuthMiddleware = (
...
+        req.models,
+        authenticator,
+        authHeader,
+        apiName,
+      );
     } else {
       throw new UnauthenticatedError('Could not authenticate with the provided access token');
     }
 
-    if (userId) {
```
edmofro, line 91: Seems to be a bit of a change from silently moving on to throwing an error here, which is probably a good thing, but just wanted to acknowledge it

packages/meditrak-server/src/routes/authenticate.js
```diff
@@ -70,6 +80,9 @@ const getAuthorizationObject = async ({ accessPolicy, refreshToken, user, permis
   if (permissionGroups) {
     userDetails.permissionGroups = permissionGroups;
   }
+  if (apiClientUser) {
+    userDetails.apiClient = apiClientUser.email;
```
edmofro, line 84: I'm a little confused about where this ends up getting used

edmofro (changes requested):
Fantastic work, and thanks for adding all those great unit tests! As per slack, let's have a call to talk through the flow of this at a higher level (it all looks great but I'm trying out a new multi-pass review process for more complicated PRs)

packages/auth/src/Authenticator.js
rohan-bes: Created: https://github.com/beyondessential/tupaia-backlog/issues/2884

packages/meditrak-server/src/auth/middleware.js
```diff
@@ -17,7 +17,8 @@ async function authenticateUser(req) {
     '/auth/verifyEmail',
     '/auth/resendEmail',
   ];
-  const tokenUserID = authHeader.startsWith('Bearer') && getUserIDFromToken(authHeader);
+  const { userId: tokenUserID } =
+    authHeader.startsWith('Bearer') && getTokenClaimsFromBearerAuth(authHeader);
```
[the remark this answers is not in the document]
rohan-bes: I was very confused why this error hadn't shown up in my tests... turns out it's not an error!
image

I don't know if I like or hate that javascript supports this 😹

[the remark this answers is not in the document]
rohan-bes: Looks like destructuring a boolean is valid?
image

packages/meditrak-server/src/routes/authenticate.js
rohan-bes: As discussed, nowhere, just a useful thing to see for devs when ensuring the user is logged in under the correct api client (also helps for tests 😉 )

packages/meditrak-server/src/tests/routes/authenticate/authenticate.test.js
rohan-bes: Decided to scrap a lot of the permissions stuff from these tests as we decided it might be best to not merge the accessPolicy upon return from the `/auth` endpoint.

packages/server-boilerplate/src/microService/auth/middleware.ts
rohan-bes: Yeah good point, though I do feel that we ought to be throwing an error here. A non-user request isn't valid right?

edmofro: Yeah I agree, I think it's a good change.

packages/meditrak-server/src/auth/middleware.js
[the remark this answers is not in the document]
edmofro: Oh javascript.... 😅

packages/auth/src/Authenticator.js
```diff
@@ -20,10 +21,16 @@ export class Authenticator {
...
 
   /**
-   * Authenticate by access token claims
-   * @param {{ userId: string, apiClientUserId?: string }} accessTokenClaims
+   * Authenticate by access token string
+   * @param {string} accessToken
    */
-  async authenticateAccessToken({ userId, apiClientUserId }) {
+  async authenticateAccessToken(accessToken) {
+    const { userId, apiClientUserId } = getTokenClaims(accessToken);
```
edmofro, line 28:
```suggestion
    const { userId, apiClientUserId } = getTokenClaims(accessToken); // will throw if access token expired or invalid
```

packages/server-boilerplate/src/microService/auth/middleware.ts
```diff
@@ -78,14 +69,9 @@ export const buildBasicBearerAuthMiddleware = (
...
-      accessPolicyObject = await getBasicAccessPolicy(
-        req.models,
-        authenticator,
-        authHeader,
-        apiName,
-      );
+      accessPolicyObject = await getBasicAccessPolicy(authenticator, authHeader, apiName);
     } else {
-      throw new UnauthenticatedError('Could not authenticate with the provided access token');
+      throw new UnauthenticatedError('Could not authenticate with the Authorization header');
```
edmofro, line 74:
```suggestion
      throw new UnauthenticatedError('Could not authenticate');
```

packages/server-boilerplate/src/microService/auth/middleware.ts
```diff
@@ -20,17 +15,13 @@ const getBearerAccessPolicy = async (
...
-  }
+  const accessToken = getJwtToken(authHeader);
 
-  const { accessPolicy } = await authenticator.authenticateAccessToken({ userId, apiClientUserId });
+  const { accessPolicy } = await authenticator.authenticateAccessToken(accessToken);
   return accessPolicy;
 };
 
 const getBasicAccessPolicy = async (
-  models: ModelRegistry,
```
edmofro, line 40: Good catch

edmofro (approved):
All looking great! Two very minor suggestions but very happy for you to ignore

rohan-bes:
Quick sanity test seemed to go smoothly, merging

PR #2601, opened 2021-05-25
2768 handle bad urls

tcaiger:
#### Refactor login modal to be a view with a route
This is better and clearer for redirecting users too when they encounter a 403. It also sets up things for the upcoming story on registration so there can be a login route and a register route with nice navigation between them.

#### Add handling for bad entity routes
Handle urls such as `/asdf/dashboard` or a valid url where the user doesn't have permission to the entity. 
- if the entityData fetch returns a 403 and the user is logged out, redirect them to the login page
- if the entityData fetch returns a 403 and the user is logged in, redirect them to the 404 page

#### Add handling for deep linking to dashboards that don't exist
- if the dashboard or subDashboard from the url doesn't exist and the user is logged out, redirect them to the login page
- if the dashboard or subDashboard from the url doesn't exist and the user is logged in, show the default dashboard

---


packages/lesmis/src/api/mutations/useLogin.js
```diff
@@ -21,9 +23,13 @@ export const useLogin = () => {
       }),
     {
       onSuccess: () => {
-        queryClient.resetQueries('user');
-        queryClient.resetQueries('entity');
-        queryClient.resetQueries('entities');
+        queryClient.clear();
+
+        if (history.location?.state?.referer) {
+          history.push(history.location.state.referer);
```
edmofro, line 29: Nice! Might be worth a comment above for people walking in without context, though it's pretty clean and clear

packages/lesmis/src/views/DashboardView.js
```diff
@@ -21,7 +21,7 @@ import {
   PartnerLogo,
 } from '../components';
 import { useUrlParams, useUrlSearchParams } from '../utils';
-import { useVitalsData, useEntityData } from '../api/queries';
+import { useVitalsData, useEntityData, useUser } from '../api/queries';
```
edmofro, line 24: Is this new import being used in a later PR maybe?

packages/lesmis/src/views/DashboardView.js
```diff
@@ -480,14 +480,27 @@ const VitalsView = React.memo(({ vitals }) => {
...
+const useDefaultDashboardTab = (selectedDashboard = null, options) => {
+  if (!options) {
+    return null;
+  }
+
+  if (selectedDashboard && options.find(option => option.value === selectedDashboard)) {
+    return selectedDashboard;
+  }
+
+  return options[0].value;
```
edmofro, line 493: There's a chance here that `options` is an empty array

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -49,12 +50,26 @@ const ScrollToTopButton = styled(ArrowUpward)`
 const DEFAULT_DASHBOARD_GROUP = 'Student Enrolment';
 const SCHOOL_DEFAULT_DASHBOARD_GROUP = 'Students';
 
-const getDefaultDashboard = data => {
-  if (!data) {
+// Gets the best default dashboard possible, and check if the selected dashboard is valid
+const useDefaultDashboardTab = (selectedDashboard = null, options) => {
+  const history = useHistory();
+  const { isLoggedIn, isFetching } = useUser();
```
edmofro, line 56:
```suggestion
  const { isLoggedIn, isFetching: isFetchingUser } = useUser();
```
Just because in the context of this hook, once you're past this line `isFetching` could be confused with fetching something about the dashboards

packages/lesmis/src/components/ProfileButton.js
```diff
@@ -26,14 +27,31 @@ const ProfileLinks = () => {
...
       MenuOptions={ProfileLinks}
     />
   ) : (
-    <LoginDialog buttonText="Log in" />
+    <LoginLink
+      variant="outlined"
+      component={RouterLink}
+      to={{
+        pathname: '/login',
+        state: { referer: history.location },
```
edmofro, line 51: It's a tidy little pattern. Out of interest, does it get represented in the URL at all, or is it just maintained internally?

packages/lesmis/src/api/queries/useEntityData.js
```diff
@@ -4,11 +4,30 @@
...
 export const useEntityData = entityCode => {
-  return useQuery(['entity', entityCode], () => get(`entity/${entityCode}`), {
-    staleTime: 1000 * 60 * 60 * 1,
-    refetchOnWindowFocus: false,
-  });
+  return useValidatedQuery(
+    useQuery(['entity', entityCode], () => get(`entity/${entityCode}`), {
+      staleTime: 1000 * 60 * 60 * 1,
+      refetchOnWindowFocus: false,
+      retry: 2,
```
edmofro, line 30: What was the retry addition for?

packages/lesmis/src/api/queries/useEntityData.js
```diff
@@ -4,11 +4,30 @@
  *
  */
 import { useQuery } from 'react-query';
+import { useHistory } from 'react-router-dom';
 import { get } from '../api';
+import { useUser } from './useUser';
+
+const useValidatedQuery = query => {
```
edmofro, line 11: Clever hook! Feels ever so slightly like it should be moved out to a common util place, but if we're only using it here, then it make sense to have it alongside

edmofro (changes requested):
Great changes!

packages/lesmis/src/views/DashboardView.js
tcaiger: nope so I've removed it

packages/lesmis/src/views/DashboardView.js
tcaiger: Yes good point. I've added `options.length === 0` to the early return condition

packages/lesmis/src/components/ProfileButton.js
tcaiger: In react router, the state object is just maintained internally. Obviously, it could be done in the url as well but there didn't seem to be any advantage to going down that route ;)

packages/lesmis/src/api/queries/useEntityData.js
tcaiger: It's just a tweak based on a testing the not found behaviour. The default for the retry param in react-query is actually 3 which means that queries that fail are silently retried 3 times (with an exponential backoff delay). I think this is really nice default behaviour but I wanted it to redirect to the 404 page slightly quicker than that here.

packages/lesmis/src/api/queries/useEntityData.js
tcaiger: Yeah I actually moved it in one of the next PRs

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -49,12 +50,26 @@ const ScrollToTopButton = styled(ArrowUpward)`
...
 const SCHOOL_DEFAULT_DASHBOARD_GROUP = 'Students';
 
-const getDefaultDashboard = data => {
-  if (!data) {
+// Gets the best default dashboard possible, and check if the selected dashboard is valid
+const useDefaultDashboardTab = (selectedDashboard = null, options) => {
+  const history = useHistory();
+  const { isLoggedIn, isFetching: isFetchingUser } = useUser();
+
+  if (!options) {
```
edmofro, line 58:
```suggestion
  if (!options || options.length === 0) {
```
Just to match that other change

edmofro (approved):
One more minor change, but pre-approved :-)

PR #2602, opened 2021-05-25
2836 - 2838: Restore and delete Alerts

chris-bes:
### Issues:
https://github.com/beyondessential/tupaia-backlog/issues/2836
https://github.com/beyondessential/tupaia-backlog/issues/2838

### Changes:
- Restore and delete Alerts

### Screenshots:
image
image
image


packages/psss-server/src/app/addRoutesToApp.ts
```diff
@@ -75,12 +76,13 @@ export function addRoutesToApp(app: Express) {
    * PUT routes
    */
   app.put('/v1/weeklyReport/:countryCode/:siteCode?', handleWith(SaveWeeklyReportRoute));
-  app.put('/v1/alerts/:alertId/archive', handleWith(ArchiveAlertRoute));
+  app.put('/v1/alerts/:alertId/:action', handleWith(ProcessAlertActionRoute));
```
kael89, line 79: Nice abstraction! 💯

packages/psss/src/containers/Tables/ArchiveTable.js
```diff
@@ -69,25 +70,51 @@ const createColumns = isSingleCountry => [
...
   return (
-    <Table
-      data={data}
-      isLoading={isLoading}
-      isFetching={!isLoading && isFetching}
-      errorMessage={error && error.message}
-      noDataMessage="No archived alerts found"
-      columns={columns}
-    />
+    <ArchiveTableContext.Provider
```
kael89, line 89: Nice use of context!

packages/psss-server/src/routes/alerts/DeleteAlertRoute.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { RespondingError } from '@tupaia/utils';
+import { Route } from '../Route';
+
+export class DeleteAlertRoute extends Route {
+  async buildResponse() {
+    const { alertId } = this.req.params;
+
+    // Just to validate if the alert exists
+    const surveyResponse = this.meditrakConnection?.findSurveyResponseById(alertId);
```
kael89, line 14:
```suggestion
    const surveyResponse = await this.meditrakConnection?.findSurveyResponseById(alertId);
```

packages/psss-server/src/routes/alerts/ProcessAlertActionRoute.ts
```diff
@@ -0,0 +1,41 @@
...
+type AlertAction = keyof typeof ACTION_TO_ANSWER;
+
+function validateAction(action: string): asserts action is AlertAction {
+  const actions = Object.keys(ACTION_TO_ANSWER);
+  if (!actions.includes(action)) {
+    throw new Error(`Invalid action '${action}', must be one of ${actions}`);
+  }
+}
+
+export class ProcessAlertActionRoute extends Route {
```
kael89, line 23: Great implementation!!

packages/psss-server/src/routes/alerts/ProcessAlertActionRoute.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { RespondingError } from '@tupaia/utils';
```
kael89, line 6: I get `eslint` errors because of formatting in this file. It's no big deal but I thought I should let you know in case something is misconfigured in your editor 🙌 ( I use "Format on Save" in VSCode)

kael89 (changes requested):
Very clean code @chris-bes , love it! ❤️ 

Apart from the requested changes, there are a lot of similarities between the alert action modals (especially in relation to Status handling), which could potentially justify reusing the same base component. Definitely not required though, plus sometimes being more specific is better.

Finally, I noticed that the `AlertMenuCell` currently does not behave in the same way as `ArchivedAlertMenuCell`. I know that we can alter the state of active alerts via their side panel, but since we've come so far, it would be nice to also implement this in PSSS Phase 2:

active_alert_actions

What do you think? If you feel that we're risking meeting our deadline, we could potentially deliver this next week.

chris-bes:
Thanks for your review @kael89 !

> Apart from the requested changes, there are a lot of similarities between the alert action modals (especially in relation to Status handling), which could potentially justify reusing the same base component. Definitely not required though, plus sometimes being more specific is better.
Agree I have definitely thought of making a base component. But then I felt I was just going to flood the base component by passing a lot of props (because we have a lot of different text/messages between the action modals)... So I changed my mind and rather make it a bit WET this time. But definitely will keep in mind and change it if we run into more similar action modals in the future.

> Finally, I noticed that the AlertMenuCell currently does not behave in the same way as ArchivedAlertMenuCell. I know that we can alter the state of active alerts via their side panel, but since we've come so far, it would be nice to also implement this in PSSS Phase 2:

No I definitely agree, it will be more convenient than having to open the side panel to archive it!

packages/psss-server/src/routes/alerts/DeleteAlertRoute.ts
chris-bes: Oops..

kael89:
> > Apart from the requested changes, there are a lot of similarities between the alert action modals (especially in relation to Status handling), which could potentially justify reusing the same base component. Definitely not required though, plus sometimes being more specific is better.
> 
> Agreed! I have definitely thought of making a base component. But then I felt I was just going to flood the base component by passing a lot of props (because we have a lot of different text/messages between the action modals)... So I changed my mind and rather make it a bit WET this time. But definitely will keep in mind and change it if we run into more similar action modals in the future.

I very recently (today...) [experimented with custom hooks](https://github.com/beyondessential/tupaia/pull/2594/commits/a7e6c84a0febd19bba8e802f47058e9c0bfddd61), and they are a very cool way to reuse functionality between components! Perhaps something like that would be the best middle ground here (thought for the future)

chris-bes:
> Thanks for your review @kael89 !
> 
> > Apart from the requested changes, there are a lot of similarities between the alert action modals (especially in relation to Status handling), which could potentially justify reusing the same base component. Definitely not required though, plus sometimes being more specific is better.
> 
> Agreed! I have definitely thought of making a base component. But then I felt I was just going to flood the base component by passing a lot of props (because we have a lot of different text/messages between the action modals)... So I changed my mind and rather make it a bit WET this time. But definitely will keep in mind and change it if we run into more similar action modals in the future.

Sorry @kael89 , I thought this was going to be easy, but turns out it's a bit tricky. This is because before we only use `ArchiveAlertModal` in the `AlertPanel`. But now we want to move the `ArchiveAlertModal` outside to be used in both `AlertPanel` and `AlertsTable` - for the ActionsMenu (potentially putting it in `AlertsTabView`). To do so, it will require a bit of refactoring because we will need a broader Context than just the current `AlertsPanelContext` (may have to refactor `AlertsPanelContext` to be `AlertsTabContext`).

I think it might require at least 1 day of work to do so and also testing, which I'm not confident now giving the deadline 😓 . I reckon that we should deliver that next week if possible. Do you think that's ok?

kael89:
> Sorry @kael89 , I thought this was going to be easy, but turns out it's a bit tricky. This is because before we only use `ArchiveAlertModal` in the `AlertPanel`. But now we want to move the `ArchiveAlertModal` outside to be used in both `AlertPanel` and `AlertsTable` - for the ActionsMenu (potentially putting it in `AlertsTabView`). To do so, it will require a bit of refactoring because we will need a broader Context than just the current `AlertsPanelContext` (may have to refactor `AlertsPanelContext` to be `AlertsTabContext`).
> 
> I think it might require at least 1 day of work to do so and also testing, which I'm not confident now giving the deadline 😓 . I reckon that we should deliver that next week if possible. Do you think that's ok?

I think that's fine Chris! Could you please confirm with Erin too? Also, I believe it's still worth implementing this some time next week

chris-bes:
Have confirmed with Erin and she said it's fine to leave it!

kael89 (approved):

PR #2603, opened 2021-05-25
Fix alert sentinel site cases

chris-bes:
### Changes:

- Similar to country level, the key is not what the report returns


kael89 (approved):

kael89:
@chris-bes I'll merge this into the feature branch

PR #2604, opened 2021-05-25
PSSS: Minor UI fixes

kael89:
### Issue #:
Addresses a few existing UI issues

### Changes

1. In a past commit I enabled the "Edit Profile" button, thinking that this was all that was left to implement https://github.com/beyondessential/tupaia-backlog/issues/1117. It seems that there is also back-end work that needs to be done. Since this was a nice-to-have for Phase 2, I'm reverting this change until the back-end is built (in a future phase)
2. The current UI was a bit inconsistent when displaying the sentinel site summary in the Weekly Report page:
   * If there was at least one report submitted in the selected week, all sites were shown. Empty data are shown as `-`
   * If there were no reports submitted, then **no** site was shown.

   This PR changes this behaviour by always showing the sentinel site summary:

   no_sentinel_site_data

   I discussed with @tcaiger and it seems that this was the intended behaviour in the first place.

3. The existing UI didn't handle loading/error info correctly in the sentinel site summary. This is fixed now:
   loading_screen

   error_screen
4. There was no provision for an "empty data" message in case no alerts were found. This is now fixed:
   no_active_alerts
   no_archived_alerts

### DIFF
--- a/packages/psss/src/api/queries/useCountrySitesWeeklyReport.js
+++ b/packages/psss/src/api/queries/useCountrySitesWeeklyReport.js
@@ -23,7 +23,6 @@ export const useCountrySitesWeeklyReport = (countryCode, period) => {
 
   return {
     ...query,
-    reportCount: report.length,
     data,
   };
 };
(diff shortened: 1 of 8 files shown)


tcaiger (approved):
Looks good to me

PR #2605, opened 2021-05-25
2853: Changed syndrome threshold crossed indicators

chris-bes:
### Changes:

- As per Erin's request, we don't want to see alert for a single case. So I'm adding total_cases > 1 to all the threshold crossed indicators except for AFR (because AFR threshold crossed is just > 1)


kael89 (approved):
Looks good Chris!

PR #2606, opened 2021-05-26
2491 lesmis user registration

tcaiger:
### Changes:

- Add registration view and registration form
- Add lesmis-server handling for registration
- Extract password strength meter into a ui component and add to the form
- Update create user meditrak endpoint handler to accept extra user permissions

---

### Screenshots:
See issue


packages/lesmis/src/components/Forms/RegisterForm.js
```diff
@@ -0,0 +1,211 @@
...
+`;
+
+const SuccessMessage = () => (
+  <MuiBox>
+    <Heading variant="h4">Register for an account</Heading>
+    <MuiBox mt={4}>
+      <Text>
+        Congratulations, you have successfully signed up to Tupaia. To activate your account please{' '}
+        <b>click the verification link in your email.</b> Once activated, you can use your new
+        account to log in to tupaia.org as well as our app, Tupaia Meditrak on{' '}
```
EMcQ-BES, line 64: Update this message to refer to lesmis

EMcQ-BES: And remove link to meditrak (unnecessary for LESMIS users)

packages/meditrak-server/src/routes/userAccounts/RegisterUserAccounts.js
```diff
@@ -87,6 +89,31 @@ export class RegisterUserAccounts extends CreateUserAccounts {
...
+
+      if (!country) {
+        throw new Error(`No such country: ${countryCode}`);
+      }
+
+      await this.models.userEntityPermission.findOrCreate({
+        user_id: userId,
+        entity_id: country.id,
+        permission_group_id: permissionGroup.id,
+      });
```
EMcQ-BES, line 114: Need some sort of check on the permission group that the user is registering against. Currently someone could send a registration with `BES Admin` and grant themselves access to everything.

EMcQ-BES (changes requested):
A couple of general comments. Security issue definitely needs to be fixed.

packages/meditrak-server/src/routes/userAccounts/RegisterUserAccounts.js
tcaiger: Good point. Any ideas about the best way of doing this? I could add a `assertBESAdminAccess` before adding the extra permissions and then add `BES Admin` permission to the lesmis-server user?

EMcQ-BES: I think we generally want to avoid giving user's `BES Admin` unnecessarily.

One idea would be rather than passing through a country/permission pair we could pass through the service that the user is registering for. So `tupaia` or `lesmis` and then look up a base country/permission pair from that.

tcaiger:
@EMcQ-BES This is ready for re-review. I have updated the RegisterUserAccounts flow to lookup the base permissions from the active apiClientUser

packages/lesmis-server/src/connections/MeditrakConnection.ts
```diff
@@ -18,4 +21,33 @@ export class MeditrakConnection extends SessionHandlingApiConnection {
...
+   *   "error": "Please complete fields."
+   * }
+   */
+  registerUser(userData: RequestBody) {
+    return this.post(
+      'user',
+      {},
+      {
+        countryCode: LESMIS_COUNTRY_CODE,
+        permissionGroupName: LESMIS_PERMISSION_GROUP,
```
EMcQ-BES, line 48: Passing these is obsolete now

EMcQ-BES (approved):
One file needs a little clean up, but the rest looks good so pre-approving

PR #2609, opened 2021-05-26
Fix small bug when deleting alerts

chris-bes:
### Issue:
https://github.com/beyondessential/tupaia-backlog/issues/2838#issuecomment-848457933

### Changes:

- Fixed small bugs when deleting/restoring alerts

### DIFF
--- a/packages/psss/src/containers/Modals/DeleteAlertModal.js
+++ b/packages/psss/src/containers/Modals/DeleteAlertModal.js
@@ -31,20 +31,25 @@ export const DeleteAlertModal = ({ isOpen, handleClose, alertId }) => {
     setStatus(STATUS.SUCCESS);
   }, [deleteAlert]);
 
+  const onClose = useCallback(async () => {
+    setStatus(STATUS.INITIAL);
+    handleClose();
+  }, [setStatus, handleClose]);
+
   if (status === STATUS.SUCCESS) {
     return (
       <SuccessModal
         isOpen={isOpen}
         title="Delete Alert"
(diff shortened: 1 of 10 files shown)


packages/psss/src/containers/Modals/DeleteAlertModal.js
```diff
@@ -31,20 +31,25 @@ export const DeleteAlertModal = ({ isOpen, handleClose, alertId }) => {
...
+  }, [setStatus, handleClose]);
+
   if (status === STATUS.SUCCESS) {
     return (
       <SuccessModal
         isOpen={isOpen}
         title="Delete Alert"
         mainText="Alert successfully deleted"
-        handleClose={handleClose}
+        handleClose={onClose}
```
kael89, line 45: A common convetion in React is to name the prop `on{{Event}}` and the handler `handle{{Event}}`:

```suggestion
        onClose={handleClose}
```

I could have commented this in an earlier PR but I just noticed

kael89 (approved):
Looks good Chris! I left a comment about a convention that shouldn't be too hard to change, but it's totally optional at this stage

packages/psss/src/containers/Modals/DeleteAlertModal.js
chris-bes: I can never say no to following convention 😬 . However I noticed there are other places still doing this, but I've just changed what I've touched 👍

kael89: Thanks for taking the time to do so! 💯

PR #2617, opened 2021-05-28
2342 larval habitat by species bar chart

billli0:
### Issue #:
[#2342](https://github.com/beyondessential/tupaia-backlog/issues/2342)
### Changes:
- one migration to add in report table 
- one migration to add dashboard


packages/database/src/migrations/20210527034457-StriveLarvalHabitatBySpeciesBarChart-modifies-data.js
```diff
@@ -0,0 +1,115 @@
...
+  dataBuilder: 'reportServer',
+  dataServices: [
+    {
+      isDataRegional: true,
+    },
+  ],
+};
+
+const DATA_BUILDER_CONFIG = {
+  reportCode: 'STRIVE_Habitat_By_Species_LHS',
```
rohan-bes, line 66: We still haven't come up with any naming convention for reportCodes and dashboardIds. Ideally they'd follow some format that would make it really easy to look up one based on the other, eg. 
`dashboardId: PG_Strive_Habitat_By_Species_Stacked_Bar_Chart, reportCode: PG_Strive_Habitat_By_Species`

Might be worth something us discussing as a team, but no biggie for now 😄

packages/database/src/migrations/20210527034457-StriveLarvalHabitatBySpeciesBarChart-modifies-data.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+const getChartConfig = () => {
+  const chartConfig = {};
+  SPECIES.forEach((specie, index) => {
```
rohan-bes, line 71:
```suggestion
  SPECIES.forEach((species, index) => {
```
Don't have to bother fixing this 😝

rohan-bes (approved):
Looks great! Nice work!

So far I've been doing the report and the dashboard in the same migration, but I think either way is totally fine, and perhaps splitting them up makes rollback a touch easier 👍

Comments are fluff, feel free to ignore

packages/database/src/migrations/20210527034457-StriveLarvalHabitatBySpeciesBarChart-modifies-data.js
billli0: I was feeling a bit weird to use `specie` but didn't really look into it.. Thanks my naming advisor @rohan-bes !

packages/database/src/migrations/20210527034457-StriveLarvalHabitatBySpeciesBarChart-modifies-data.js
billli0: Yeah that is a good point, I followed the previous report codes naming formate, which has no country code, but it seems like they don't have country code because they are used in multiple countries.

billli0: I still incline to use a shorter `dashboardId` because I prefer a shorter URL, which is clean, clear and more attractive when seen or copied in full, but it is worth to debate in dev chat 😆

PR #2618, opened 2021-05-28
2890: Fixed raw data download bug

chris-bes:
### Changes:

- Fixed a bug that Laos school Raw Data Download does not return values for some fields


edmofro (approved):
lgtm 🙏

PR #2620, opened 2021-05-30
Add utility script for running analytics full refresh

edmofro:
Drops indexes, runs a full refresh of analytics, then recreates indexes

### DIFF
--- a/packages/data-api/package.json
+++ b/packages/data-api/package.json
@@ -18,7 +18,7 @@
     "test:coverage": "yarn test --coverage",
     "install-mv-refresh": "scripts/installMvRefreshModule.sh",
     "build-analytics-table": "scripts/buildAnalyticsMaterializedView.sh",
-    "full-refresh-analytics-table": "scripts/fullRefreshAnalyticsTable.sh",
+    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh",
     "full-rebuild-analytics-table": "scripts/fullRebuildAnalyticsMaterializedView.sh"
   },
   "dependencies": {
--- a/packages/data-api/package.json
+++ b/packages/data-api/package.json
@@ -18,6 +18,7 @@
     "test:coverage": "yarn test --coverage",
     "install-mv-refresh": "scripts/installMvRefreshModule.sh",
     "build-analytics-table": "scripts/buildAnalyticsMaterializedView.sh",
+    "full-refresh-analytics-table": "scripts/fullRefreshAnalyticsTable.sh",
     "full-rebuild-analytics-table": "scripts/fullRebuildAnalyticsMaterializedView.sh"
   },
   "dependencies": {
--- a/packages/data-api/scripts/dropAnalyticsTableIndexes.sql
+++ b/packages/data-api/scripts/dropAnalyticsTableIndexes.sql
@@ -1,0 +1,16 @@
+do $$
+declare
+  tStartTime TIMESTAMP;
+
+begin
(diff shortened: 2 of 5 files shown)


packages/data-api/package.json
```diff
@@ -18,6 +18,7 @@
     "test:coverage": "yarn test --coverage",
     "install-mv-refresh": "scripts/installMvRefreshModule.sh",
     "build-analytics-table": "scripts/buildAnalyticsMaterializedView.sh",
+    "full-refresh-analytics-table": "scripts/fullRefreshAnalyticsTable.sh",
```
rohan-bes, line 21: I had in mind that it might be nice for this script to be `refresh-analytics-table` which takes `-f` / `--full` as an argument (with fast refresh being the default). That way allowing for easily doing either fast or full refresh.

If you agree and don't mind implementing that would be awesome, but of course it's totally optional and the full refresh is the one worth putting any effort into!

packages/data-api/scripts/dropAnalyticsTableIndexes.sql
```diff
@@ -0,0 +1,16 @@
...
+
+begin
+  RAISE NOTICE 'Dropping analytics table indexes...';
+
+  tStartTime := clock_timestamp();
+  DROP INDEX IF EXISTS analytics_data_element_entity_date_idx;
+  RAISE NOTICE 'Dropped [data_element, entity, date] index, took %', clock_timestamp() - tStartTime;
+
+  tStartTime := clock_timestamp();
+  DROP INDEX IF EXISTS analytics_data_element_entity_data_group_event_date_idx;
```
rohan-bes, line 13: The name of this index (and the columns it uses) has been changed in: https://github.com/beyondessential/tupaia-backlog/issues/2714

That issue is currently part of `psss-phase-2` branch, and I believe won't be merged before this one, so I guess it's up to someone working on `psss-phase-2` to ensure that this script gets updated as well

rohan-bes (approved):
Thanks @edmofro, this is awesome and should help avoid others running into the issue you did with the slow full refresh! Just a suggestion about how we could make the yarn task a bit more generic but totally optional.

packages/data-api/scripts/dropAnalyticsTableIndexes.sql
edmofro: Good catch @rohan-bes! Very glad you reviewed this, because that would have been a soft conflict that may have gone unnoticed. I'll remember to let `psss-phase-2` folks know when this has hit dev for their next update

packages/data-api/package.json
edmofro: Good call! I'll work on that now

packages/data-api/scripts/dropAnalyticsTableIndexes.sql
rohan-bes: One possible way we could make this more robust is remove the `IF EXISTS` from the `DROP` statement? I suppose that way it asserts that the index is there? However, that would make this script more annoying to use on a partially built / weird state analytics table so perhaps not... I'm fine with it as is 😄

edmofro: Yeah I see what you mean...It would mean we were more likely to catch changes to the index name, but also make the script a bit more constrained in its use. I'm on the fence!

rohan-bes (approved):
Awesome, thanks for extending the script functionality!!

PR #2621, opened 2021-05-31
2885: Fix no data being returned when requested entity is at the same level as dataSourceEntityType

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2885:

This issue was being caused by the entity-server returning no entities when requested for descendants of type x of entity that is type x, eg. facility descendants of a facility

While this issue could be resolved by improving the logic of the `Aggregator`, it seemed like it might be better to implement a `/relatives` endpoint in entity-server which returns all relatives of the requested entity(ies). Relatives are all ancestors, descendants, and the entity itself. Using a filter on this list of entities can allow us to fetch the correct descendant->ancestor mapping (assuming no cyclical entity type relationships; we don't support this so it's not an issue).

## Changes:
- Implemented `getRelatives()` function on `EntityModel` in `@tupaia/database`
- Introduced new `/relatives` endpoint to fetch relatives of an entity(ies)
- Switch `getDataSourceEntities()` call in Aggregator to use `/relatives` instead of `/descendants`
- Re-worked logic of `/relations` route to return relations among relatives, rather than relations among descendants


packages/entity-server/src/app/createApp.ts
```diff
@@ -33,11 +37,16 @@ export function createApp() {
     .useBasicBearerAuth('entity-server')
     .use<SingleEntityRequest | MultiEntityRequest>('hierarchy/:hierarchyName', attachCommonContext)
     .use<MultiEntityRequest>('hierarchy/:hierarchyName/descendants', attachMultiEntityContext)
+    .use<MultiEntityRequest>('hierarchy/:hierarchyName/relatives', attachMultiEntityContext)
```
rohan-bes, line 40: Now we have `/relatives` and `/relations`. Is this too confusing? 😂 

Initially I definitely felt that way, and was trying to decide which to rename. However, as I worked on this more I did come to feel they are fairly conceptually linked. Relations are the relations among the relatives, and relatives are just the full set of all relations (within a hierarchy).

But there's a good chance I've just gotten too used to the code on this one, and if at first glance it's confusing I am definitely happy to change it (might do it regardless).

I'm a tad reluctant to rename `/relations` as it's really a direct resource for our `entityRelations` table, I think that naming consistency is nice. So it comes to picking a new name for `/relatives`?

packages/entity-server/src/routes/hierarchy/relations/RelationsResponseBuilder.ts
```diff
@@ -48,7 +48,9 @@ export class RelationsResponseBuilder {
       hierarchyId,
       ancestorType,
     );
-    const ancestorCodes = Object.values(descendantAncestorMapping).map(ancestor => ancestor.code);
+    const ancestorCodes = [
+      ...new Set(Object.values(descendantAncestorMapping).map(ancestor => ancestor.code)),
+    ];
```
rohan-bes, line 53: This fixed a small issue (wasn't causing any bugs) where the requested ancestors would be something like : `WHERE code in ('TO','TO','TO','TO',etc..)`

rohan-bes:
I'd like to have a crack at implementing a proper test suite for entity-server in this PR (using supertest as an api testing framework). May not happen if it's taking too long as this fix is somewhat urgent.

packages/entity-server/src/app/createApp.ts
edmofro: It really is confusing, and I'm keen for a rename

I get where you're coming from, but am not totally convinced that `/relations` needs to remain...would you consider `/relationships`, which seems more descriptive to me and makes the distinction between that and `/relatives` pretty clear (at least in my mind)?

packages/entity-server/src/routes/hierarchy/relations/RelationsResponseBuilder.ts
```diff
@@ -128,7 +129,7 @@ export class RelationsResponseBuilder {
     const { entity, hierarchyId, allowedCountries } = this.ctx;
     const { type: descendantType, filter } = this.ctx.descendant;
 
-    const descendants = await entity.getDescendants(hierarchyId, {
+    const descendants = await entity.getRelatives(hierarchyId, {
```
edmofro, line 132: So I guess it would now be possible for someone to request e.g. country descendants and facility ancestors, and it would respond with something pseudo-sensible? That's probably fine to leave on the end user to get right, just noting it as a bit of a logical weirdness

edmofro (changes requested):
Code is all fantastic, really showing how nicely put together the entity-server already is that it's so clean to add these new endpoints.

Let's bounce ideas on the endpoint names, that's the only thing I'd want to change on this (yes I did get confused)

packages/entity-server/src/app/createApp.ts
rohan-bes: As per our discussion, gonna rename `/relations` to `/relationships`! Think this should be a pretty safe change as I think only the Aggregator uses this endpoint for now

packages/entity-server/src/routes/hierarchy/relationships/EntityRelationshipsRoute.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { Route } from '@tupaia/server-boilerplate';
+import { RelationshipsRequest } from './types';
+import { ResponseBuilder } from './ResponseBuilder';
```
rohan-bes, line 8: Decided to just go with `ReponseBuilder` here over `RelationshipsResponseBuilder`. Kind of personal preference a bit, but when some functionality is private to a package, I like to omit the prefixes that are already in the package name. I know this can make code a little less searchable, and perhaps read a little unexpectedly at times, but I think I prefer that over ever increasingly verbose names... Happy to go either way at the end of the day 🤷

rohan-bes:
@edmofro I think I renamed em all, but would love a re-review all the same!

Still plan on doing those unit tests, but may want to merge this before they're ready as there are existing dashboards that are blocked by this card.

Also, unfortunately there may have been a bit of a performance degradation introduced in PSSS due to this change 😿 I don't believe it's introduced by the extra overhead of queries to calculate `relatives` of an entity, but rather due to us now returning data where we previously should have been, and so indicators being forced to crunch through more data... I will continue to investigate and hold off merging until the issue is understood 👍

packages/entity-server/src/routes/hierarchy/relationships/EntityRelationshipsRoute.ts
edmofro: I lean a little towards verbosity because I use `cmd` + `p` in vscode and it's more helpful for that form of navigation (always hate getting to the specific `index.js` file I need to edit...), but I also don't mind it and it doesn't take too much longer to see which folder the file is within :+1:

edmofro (approved):
Looks great! I actually like this change so much haha, especially where it gets used in the aggregator package and so is further removed from the actual `relation` table.

rohan-bes:
Summary of performance investigation:
- While not 100% sure, I think this issue is caused by the fact that now the entity-server is returning data where previously it was returning no entities (ie. when requesting to aggregate from facility level to any other level from a facility). This means that we run deeper into the nested indicators each time we calculate the weekly report, thus taking more time
- Unfortunately, returning early if indicator found among the ancestors didn't improve performance greatly
- A change to entity-server to have the 'shouldFastRespond()' logic work correctly had the most impact, bringing the performance down to a similar level as current PSSS. Current dev takes ~8-9 seconds to load the Tonga unconfirmed data page, this branch now takes ~11 seconds
- Further improvements could definitely be made to the entity-server, for instance the MultiEntity routes could be optimised to fetch all data at once rather than one by one, but it seems a bit out of scope of this ticket and from my testing didn't add a large performance improvement
- @kael89 and I have discussed this one, and somewhat feel that a deeper approach to the performance issues in PSSS, namely in nested indicators, should be undertaken. @edmofro how does this sound to you?

rohan-bes:
@edmofro could I get a re-review on this one? Apologies for the force push, had to rebase off the latest dev, the commits left to review are:

5f29cc0 - Had to fixup psss-server not using new endpoint name after the merge

5ae1e09 - Re-wrote shouldPerformFastResponse() to work correctly, previously it never returned true as 'country_code' is always part of the filter. Should be safe to remove this check for ancestors as we should always have access to them

4a736b9 - Fixed a really nasty bug here that I came across while testing! It had actually become a race condition as to the order of the aggregation list because I was being dumb and pushing to a list during an async loop 🤦 

14e3bb7 - Should have caught this one earlier in testing tbh... it actually wasn't returning the requested entity when it was requesting from and to the entity type

3b49fc5 - Migration here to re-write the PSSS indicators to use more simple logic now that entity-aggregation is working as expected. Offers a small performance boost, but nothing too significant...

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
```diff
@@ -83,26 +79,34 @@ const getAdjustedOrganisationUnitsAndAggregations = async (
...
   );
 
-  return [adjustedOrganisationUnitCodes, adjustedAggregations];
+  const adjustedAggregations = adjustedAggregationsAndDataSourceEntites.map(
+    aggregationAndDataSourceEntities => aggregationAndDataSourceEntities[0],
+  );
+  const dataSourceEntities =
+    adjustedAggregationsAndDataSourceEntites
+      .map(aggregationAndDataSourceEntities => aggregationAndDataSourceEntities[1])
+      .find(dataSourceEntities => dataSourceEntities !== undefined) || organisationUnitCodes;
```
chris-bes, line 107: Just checking, shouldn't this be:
```suggestion
      .filter(dataSourceEntities => dataSourceEntities !== undefined).flat() || organisationUnitCodes;
```

Also `dataSourceEntities` is already declared the scope above

packages/database/src/migrations/20210607063222-RewritePSSSIndicatorsToUseRelativesEntityAggregation-modifies-data.js
```diff
@@ -0,0 +1,278 @@
...
+        code: 'siteWeeklyResponsesCount',
+        config: {
+          formula: 'PSSS_Site_Weekly_Survey_Completed',
+          aggregation: [
+            'FINAL_EACH_WEEK',
+            {
+              type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+              config: {
+                dataSourceEntityType: 'facility',
+                aggregationEntityType: 'requested',
```
chris-bes, line 103: Ahh nice, much simpler!

chris-bes (commented):
LGTM! Just 1 question about populating the `dataSourceEntities`

packages/aggregator/src/analytics/aggregateAnalytics/adjustOptionsToAggregationList.js
rohan-bes: So the logic to determine the dataSourceEntities for an array of aggregations is actually a little bit wonky at the moment... Basically what we do is find the first entity aggregation, get its dataSourceEntities, and use that. In a sense, dataSourceEntities is really just a fetchOption, and its a miscategorisation that we have to put them in an aggregation just to get the fetch working correctly. I've already had to write multiple reports/indicators that have something like:
```
fetch: {
  ...,
  aggregations: [{ type: 'RAW', config: { dataSourceEntityType: 'facility' } }] 
}
```
Just so that I can fetch the data I want. Ideally we'd have something like this instead:
```
fetch: {
  ...,
  organisationUnits: { codes: ['TO', 'PG'], dataSourceEntityType: 'facility' }
}
```

So going back to the logic you raised concerns about, it's actually meant to find the first `not undefined` value (even if that value is an empty array), and ignore all the later ones.

As for the `dataSourceEntities` variable being already declared in scope, I think it's okay as those declarations are in the `Promise.all()` code block?

chris-bes (approved):

edmofro:
> @kael89 and I have discussed this one, and somewhat feel that a deeper approach to the performance issues in PSSS, namely in nested indicators, should be undertaken. @edmofro how does this sound to you?

Happy with that call :+1:

edmofro:
Also thanks for doing the follow up review in my absence @chris-bes!

edmofro:
Closing in favour of https://github.com/beyondessential/tupaia/pull/2643

PR #2624, opened 2021-06-01
1400: Strive Lab Confirmed Positive Results Bar Chart

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/1400:

3rd time's a charm 😂 

Re-implemented this dashboard via the report-server so that we can use entity-aggregation correctly.

Had to add `divide` function to `report-server` as we may be dividing by `undefined` and the `mathjs` native `x / y` operator throws an error for `undefined`. Personally I'd love for us to fork mathjs and start adding support for a few of the things we need... but for now wrapping the functions seems best...


rohan-bes:
Unit tests incoming!

packages/report-server/src/reportBuilder/functions/basic.ts
```diff
@@ -21,22 +21,22 @@ export const last = (values: FieldValue[]): FieldValue => {
...
-export const exists = (value: FieldValue): boolean => {
-  return value !== undefined;
+export const exists = (val: any): boolean => {
+  return val !== undefined;
 };
 
-export const notExists = (value: FieldValue): boolean => {
-  return value === undefined;
+export const notExists = (val: FieldValue): boolean => {
+  return val === undefined;
```
billli0, line 41: Sorry I didn't realise we have basic functions at the top of everything here and I added something similar [here...](https://github.com/beyondessential/tupaia/blob/72ea0a8a8e4d3ba2f033e57c1806e23f19f131fd/packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts#L8). @rohan-bes Do you think we can merge them together?

packages/report-server/src/reportBuilder/functions/math.ts
```diff
@@ -3,12 +3,41 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-import { sum as mathjsSum, Matrix, filter } from 'mathjs';
+import { sum as mathjsSum, divide as mathjsDivide } from 'mathjs';
 
-export const sum = (numbers?: Matrix): number | undefined => {
-  if (numbers === undefined) {
+const isDefined = <T>(value: T): value is Exclude<T, undefined> => value !== undefined;
```
billli0, line 8: Similar here, but the trade off is that, it will be slightly easier to understand the rest of the codes with them here. Either way is fine to me, I don't want to be picky so I leave it to you to decide.

billli0 (approved):
@rohan-bes Code is super clean! This card is hanging for so long and report-sever just saves us 👍

packages/report-server/src/reportBuilder/functions/basic.ts
rohan-bes: So that function there is actually a function we have available when running expressions, eg.
```
{
  "transform": "select",
  "'Has No Data'": "notExists($row.dataValue)"
}
```

It shouldn't really be used in-code the way that your `isNotUndefined` function is being used 👍 

However, you do make a good point that:
a) It is hard to know which functions are available when parsing expressions
b) We often define the same utility functions over and over in typescript, eg. something like `isNotUndefined` exists in a couple of places

Regarding a) it's a bit of a tricky issue, ideally we'd move all the functions into `@tupaia/expression-parser` and then there would be one central place to look them all up! Even better, when we get a UX for creating reports via the admin panel we can auto-suggestion!

Regarding b) I think a `@tupaia/ts-utils` package is probably overdue at this point... We just haven't got around to creating it as of yet...

packages/report-server/src/reportBuilder/functions/math.ts
rohan-bes: See my point 'Regarding b)' in this comment: https://github.com/beyondessential/tupaia/pull/2624#discussion_r644391666

PR #2626, opened 2021-06-02
2844 lesmis users table

tcaiger:
### Changes:

- Add UsersRoute to lesmis-server
- Add lesmis permission group to tupaia admin panel permission assertion
- Return access policy with user from meditrak-server


packages/lesmis-server/src/routes/UsersRoute.ts
```diff
@@ -0,0 +1,62 @@
...
+  'user.position': 'position',
+  'entity.name': 'entityName',
+  'entity.code': 'entityCode',
+  'permission_group.name': 'permissionGroupName',
+  id: 'userEntityPermissionId',
+};
+
+const PERMISSION_GROUPS = ['Laos Schools Admin', 'LESMIS Public', 'Laos Schools Super User'];
+
+// Remove duplicat records and convert user keys to nice camelCase keys for the front end
```
edmofro, line 28:
```suggestion
// Remove duplicate records and convert user keys to nice camelCase keys for the front end
```

packages/lesmis-server/src/routes/UsersRoute.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  constructor(req: Request, res: Response, next: NextFunction) {
+    super(req, res, next);
+
+    this.meditrakConnection = new MeditrakConnection(req.session);
+  }
+
+  async buildResponse() {
+    const userData = await this.meditrakConnection.getUserEntityPermissions({
+      pageSize: '10000',
```
edmofro, line 51:
```suggestion
      pageSize: '10000', // retrieve all entries, as we don't handle pagination
```
10000 might be too small for LESMIS in the end, considering there are more schools than that

@enunan how many users are we expecting?

packages/lesmis-server/src/routes/UsersRoute.ts
```diff
@@ -0,0 +1,62 @@
...
+      columns: JSON.stringify(Object.keys(FIELDS)),
+      sort: JSON.stringify(['user.creation_date DESC']),
+      filter: JSON.stringify({
+        'entity.name': { comparator: 'like', comparisonValue: 'Laos', castAs: 'text' },
```
edmofro, line 55:
```suggestion
        'entity.name': 'Laos',
```
I think that should also work? The `castAs` in particular shouldn't be necessary, I believe that was a fix for enums

Also it's probably safer to filter on the code, in case they change the name to e.g. 'Lao People's Democratic Republic'

```suggestion
        'entity.code': 'LA',
```

packages/lesmis-server/src/utils/accessPolicy.ts
```diff
@@ -7,14 +7,15 @@ import { AccessPolicy } from '@tupaia/access-policy';
 import { PermissionsError } from '@tupaia/utils';
 import { LESMIS_COUNTRY_CODE, LESMIS_PERMISSION_GROUP } from '../constants';
```
edmofro, line 8:
```suggestion
import { LESMIS_COUNTRY_CODE, LESMIS_BASE_PERMISSION_GROUP, LESMIS_ADMIN_PERMISSION_GROUP } from '../constants';
```
Rather than having `LESMIS Public` in a const and `LESMIS Admin` inline, let's pull them both into constants

packages/meditrak-server/src/permissions/constants.js
```diff
@@ -4,4 +4,5 @@
  */
 
 export const BES_ADMIN_PERMISSION_GROUP = 'BES Admin';
+export const LESMIS_ADMIN_PERMISSION_GROUP = 'Laos Schools Admin';
```
edmofro, line 7: Damn, I thought we'd set up the LESMIS permission groups but it doesn't look like that ever happened. This should be `LESMIS Admin`, and we should alter the permission groups in the database to be as we had planned: LESMIS Admin -> LESMIS User -> LESMIS Public

packages/meditrak-server/src/permissions/assertions.js
```diff
@@ -51,22 +55,25 @@ export const assertAnyPermissions = (assertions, errorMessage) => async accessPo
...
     return true;
   }
 
   throw new Error(`Need ${BES_ADMIN_PERMISSION_GROUP} access`);
 };
+
+export const hasTupaiaAdminPanelAccess = accessPolicy =>
+  accessPolicy.allowsSome(null, TUPAIA_ADMIN_PANEL_PERMISSION_GROUP);
+
 export const assertTupaiaAdminPanelAccess = accessPolicy => {
```
edmofro, line 75:
```suggestion
export const assertAdminPanelAccess = accessPolicy => {
```

edmofro (changes requested):
The only significant change is to alter the actual permission structure we're using in the database, and update the hard coded permission group name here to reflect that. Otherwise, looking good!

packages/meditrak-server/src/permissions/assertions.js
tcaiger: Good re-name - nice and simple

tcaiger:
@edmofro - This is ready for re-review. The renaming of the tupaiaAdminAssertion resulted in a lot of changed files so it might make sense to review commit by commit. The rename of tupaiaAdminAssertion is all in the commit named `rename assertTupaiaAdminPanelAccess`.

packages/database/src/migrations/20210602041852-Rename-lesmis-permission-groups-modifies-data.js
```diff
@@ -0,0 +1,36 @@
...
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const updateLESMISReportPermissionGroup = function (db, oldName, newName) {
```
edmofro: `updatePermissionGroupName` is probably more appropriate to what's happening, but in a migration it doesn't matter too much

Personally I also like to mark functions as `async` if they return a promise, just as a flag to anyone using them. Again, in a migration, doesn't matter at all, but just for future reference

packages/database/src/migrations/20210602041852-Rename-lesmis-permission-groups-modifies-data.js
```diff
@@ -0,0 +1,36 @@
...
+  return db.runSql(`
+  UPDATE "permission_group"
+  SET "name" = '${newName}'
+  WHERE "name" = '${oldName}'
+  `);
+};
+
+exports.up = async function (db) {
+  await updateLESMISReportPermissionGroup(db, 'Laos Schools Admin', 'LESMIS Admin');
+  await updateLESMISReportPermissionGroup(db, 'Laos Schools Super User', 'LESMIS User');
```
edmofro: There's actually a bit more to do here, as well as renaming we need to restructure the hierarchy so that each one has its parent set to the one above

edmofro: Ooops, done in a later commit!

edmofro (approved):
🎉 

Just a question remaining for @enunan to let us know how many users we might expect to be on LESMIS. If less than 10,000, this is all good :-)

enunan:
For now there will be less than 10,000 users. Not sure about future but can assume less for foreseeable future

PR #2628, opened 2021-06-02
fetchEvents() to handle multiple data groups in report server

billli0:
### Issue #:
no issue

### Changes:
we want to fetch multiple dataGroups in report-server, ultimately we would want Aggregator to handle that.

### DIFF
--- a/packages/report-server/src/reportBuilder/fetch/functions/dataGroup.ts
+++ b/packages/report-server/src/reportBuilder/fetch/functions/dataGroup.ts
@@ -19,7 +19,7 @@ type DataGroupParams = {
 };
 
 type DataGroupFetchParams = {
-  dataGroupCode: string;
+  dataGroupCodes: string[];
   dataElementCodes?: string[];
   aggregations?: (string | AggregationObject)[];
 };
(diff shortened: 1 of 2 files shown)


packages/report-server/src/reportBuilder/fetch/functions/dataGroup.ts
```diff
@@ -29,26 +29,32 @@ const fetchEvents = async (
...
+    }),
   );
-  const rows = response.map(event => {
-    const { dataValues, ...restOfEvent } = event;
-    return { ...dataValues, ...restOfEvent };
-  });
+
   return {
-    results: rows,
+    results: results.flat(),
```
billli0, line 57: I tried to use .flatMap() but it seems like TypeScript having trouble mapping the types here and throwing a type error.

packages/report-server/src/reportBuilder/fetch/functions/dataGroup.ts
```diff
@@ -29,26 +29,32 @@ const fetchEvents = async (
...
-    aggregations,
-    organisationUnitCodes,
-    hierarchy,
-    {
-      period,
-      startDate,
-      endDate,
-    },
-    dataElementCodes,
+  // TODO: Eventually we want Aggregator to handle fetching multiple dataGroups
```
rohan-bes, line 34: Appreciate this comment so that we know the plan for the future! 🎉

rohan-bes (changes requested):
Looks great @billli0! Can't wait to use this

One request though, I believe we also need to update the validator for `validateDataGroups()` to allow for more than 1 `dataGroup`?

billli0:
> Looks great @billli0! Can't wait to use this
> 
> One request though, I believe we also need to update the validator for `validateDataGroups()` to allow for more than 1 `dataGroup`?

Thanks @rohan-bes , good catching!

rohan-bes (approved):
Awesome!

billli0:
Note that if we fetch with `{dataGroups: ['A', 'B'], dataElements: ['A_1', 'A_2', 'B_1', 'B_2'] }`, it will send all the dataElements grouping with each dataGroup, like `{dataGroupCode: ['A'], dataElements: ['A_1', 'A_2', 'B_1', 'B_2']}`, instead of doing it with filtered dataElements like `{dataGroupCode: ['A'], dataElements: ['A_1', 'A_2']`. 

But Aggregator is smart enough to filter them itself, which is very handy. 👍

PR #2632, opened 2021-06-03
2883 report server output stage

billli0:
### Issue #:
[#2883](https://github.com/beyondessential/tupaia-backlog/issues/2883)
### Changes:
as [comment](https://github.com/beyondessential/tupaia-backlog/issues/2883#issue-904527883)

### DIFF
--- a/packages/report-server/package.json
+++ b/packages/report-server/package.json
@@ -39,6 +39,7 @@
     "body-parser": "^1.18.3",
     "cors": "^2.8.5",
     "express": "^4.16.2",
+    "lodash.pick": "^4.4.0",
     "mathjs": "^7.3.0",
     "winston": "^3.2.1"
   }
--- a/packages/report-server/package.json
+++ b/packages/report-server/package.json
@@ -40,6 +40,7 @@
     "cors": "^2.8.5",
     "express": "^4.16.2",
     "lodash.pick": "^4.4.0",
+    "@types/lodash.pick": "^4.4.0",
     "mathjs": "^7.3.0",
     "winston": "^3.2.1"
   }
--- a/packages/report-server/package.json
+++ b/packages/report-server/package.json
@@ -40,7 +40,6 @@
     "cors": "^2.8.5",
     "express": "^4.16.2",
     "lodash.pick": "^4.4.0",
-    "@types/lodash.pick": "^4.4.0",
     "mathjs": "^7.3.0",
     "winston": "^3.2.1"
   }
--- a/packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
@@ -7,6 +7,7 @@ import { buildOutput } from '../../../reportBuilder/output';
 import {
   MULTIPLE_TRANSFORMED_DATA,
   MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES,
+  MULTIPLE_TRANSFORMED_DATA_FOR_SPECIFIED_COLUMNS,
 } from './output.fixtures';
 
 describe('output', () => {
@@ -79,12 +80,26 @@ describe('output', () => {
       });
 
       it('can return only specified columns', () => {
-        const expectedDataForOnlySpecifiedColumns = {
-          ...expectedData,
+        const expectedDataForThisCase = {
           columns: [
             {
               key: 'Laos',
               title: 'Laos',
+            },
+          ],
+          rows: [
+            {
+              categoryId: 'medical center',
+              dataElement: 'hospital',
+              Laos: 3,
+            },
+            {
+              categoryId: 'medical center',
+              dataElement: 'clinic',
+              Laos: 4,
+            },
+            {
+              category: 'medical center',
             },
           ],
         };
@@ -94,8 +109,8 @@ describe('output', () => {
           rowField: 'FacilityType',
           columns: ['Laos'],
         });
-        expect(output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES)).toEqual(
-          expectedDataForOnlySpecifiedColumns,
+        expect(output(MULTIPLE_TRANSFORMED_DATA_FOR_SPECIFIED_COLUMNS)).toEqual(
+          expectedDataForThisCase,
         );
       });
     });
--- a/packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
@@ -4,88 +4,200 @@
  */
 
 import { buildOutput } from '../../../reportBuilder/output';
-import { MULTIPLE_TRANSFORMED_DATA } from './output.fixtures';
+import {
+  MULTIPLE_TRANSFORMED_DATA,
+  MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES,
+} from './output.fixtures';
 
 describe('output', () => {
   describe('matrix', () => {
-    const expectedData = {
-      columns: [
-        {
-          key: 'Laos',
-          title: 'Laos',
-        },
-        {
-          key: 'Tonga',
-          title: 'Tonga',
-        },
-      ],
-      rows: [
-        {
-          categoryId: 'medical center',
-          dataElement: 'hospital',
-          Laos: 3,
-          Tonga: 0,
-        },
-        {
-          categoryId: 'medical center',
-          dataElement: 'clinic',
-          Laos: 4,
-          Tonga: 9,
-        },
-        {
-          categoryId: 'others',
-          dataElement: 'park',
-          Laos: 2,
-          Tonga: 0,
-        },
-        {
-          categoryId: 'others',
-          dataElement: 'library',
-          Laos: 0,
-          Tonga: 5,
-        },
-        {
-          category: 'medical center',
-        },
-        {
-          category: 'others',
-        },
-      ],
-    };
+    describe('handle columns', () => {
+      const expectedData = {
+        columns: [
+          {
+            key: 'Laos',
+            title: 'Laos',
+          },
+          {
+            key: 'Tonga',
+            title: 'Tonga',
+          },
+        ],
+        rows: [
+          {
+            categoryId: 'medical center',
+            dataElement: 'hospital',
+            Laos: 3,
+            Tonga: 0,
+          },
+          {
+            categoryId: 'medical center',
+            dataElement: 'clinic',
+            Laos: 4,
+            Tonga: 9,
+          },
+          {
+            categoryId: 'others',
+            dataElement: 'park',
+            Laos: 2,
+            Tonga: 0,
+          },
+          {
+            categoryId: 'others',
+            dataElement: 'library',
+            Laos: 0,
+            Tonga: 5,
+          },
+          {
+            category: 'medical center',
+          },
+          {
+            category: 'others',
+          },
+        ],
+      };
 
-    it('can perform without columns', () => {
-      const output = buildOutput({
-        matrix: {
+      it('can perform without columns', () => {
+        const output = buildOutput({
+          type: 'matrix',
           categoryField: 'InfrastructureType',
           rowField: 'FacilityType',
-        },
+        });
+        expect(output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES)).toEqual(expectedData);
       });
-      expect(output(MULTIPLE_TRANSFORMED_DATA)).toEqual(expectedData);
-    });
 
-    it("throw error when columns is not '*' or string array", () => {
-      const output = buildOutput({
-        matrix: {
+      it("throw error when columns is not '*' or string array", () => {
+        const output = buildOutput({
+          type: 'matrix',
           categoryField: 'InfrastructureType',
           rowField: 'FacilityType',
           columns: 1,
-        },
+        });
+        expect(() => {
+          output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES);
+        }).toThrow();
       });
-      expect(() => {
-        output(MULTIPLE_TRANSFORMED_DATA);
-      }).toThrow();
+
+      it('can return only specified columns', () => {
+        const expectedDataForOnlySpecifiedColumns = {
+          ...expectedData,
+          columns: [
+            {
+              key: 'Laos',
+              title: 'Laos',
+            },
+          ],
+        };
+        const output = buildOutput({
+          type: 'matrix',
+          categoryField: 'InfrastructureType',
+          rowField: 'FacilityType',
+          columns: ['Laos'],
+        });
+        expect(output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES)).toEqual(
+          expectedDataForOnlySpecifiedColumns,
+        );
+      });
     });
 
-    it('can perform with categoryField', () => {
-      const output = buildOutput({
-        matrix: {
+    describe('handle categoryField', () => {
+      it('can perform with categoryField', () => {
+        const expectedData = {
+          columns: [
+            {
+              key: 'Laos',
+              title: 'Laos',
+            },
+            {
+              key: 'Tonga',
+              title: 'Tonga',
+            },
+          ],
+          rows: [
+            {
+              categoryId: 'medical center',
+              dataElement: 'hospital',
+              Laos: 3,
+              Tonga: 0,
+            },
+            {
+              categoryId: 'medical center',
+              dataElement: 'clinic',
+              Laos: 4,
+              Tonga: 9,
+            },
+            {
+              categoryId: 'others',
+              dataElement: 'park',
+              Laos: 2,
+              Tonga: 0,
+            },
+            {
+              categoryId: 'others',
+              dataElement: 'library',
+              Laos: 0,
+              Tonga: 5,
+            },
+            {
+              category: 'medical center',
+            },
+            {
+              category: 'others',
+            },
+          ],
+        };
+        const output = buildOutput({
+          type: 'matrix',
           categoryField: 'InfrastructureType',
           rowField: 'FacilityType',
           columns: '*',
-        },
+        });
+        expect(output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES)).toEqual(expectedData);
       });
-      expect(output(MULTIPLE_TRANSFORMED_DATA)).toEqual(expectedData);
+
+      it('perform with no categoryField', () => {
+        const expectedData = {
+          columns: [
+            {
+              key: 'Laos',
+              title: 'Laos',
+            },
+            {
+              key: 'Tonga',
+              title: 'Tonga',
+            },
+          ],
+          rows: [
+            {
+              dataElement: 'hospital',
+              Laos: 3,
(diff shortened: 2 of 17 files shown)


packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    it('can perform with categoryField', () => {
+      const output = buildOutput({
+        matrix: {
+          categoryField: 'InfrastructureType',
+          rowField: 'FacilityType',
+          columns: '*',
+        },
+      });
+      expect(output(MULTIPLE_TRANSFORMED_DATA)).toEqual(expectedData);
+    });
```
rohan-bes, line 88: Might just wanna add a test to check that `it('can return only specified columns')` where something like `columns: ['Laos']` is in the config 👍

packages/report-server/src/reportBuilder/output/functions/default.ts
```diff
@@ -0,0 +1,5 @@
+import { Row } from '../../types';
```
rohan-bes, line 1: Don't wanna be too much of a pain, but always good to the have Copyright at the top of all files 👍 
```suggestion
/**
 * Tupaia
 * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
 */
 
import { Row } from '../../types';
```

packages/report-server/src/reportBuilder/output/functions/matrix/types.ts
```diff
@@ -0,0 +1,9 @@
+export type MatrixColumnParams = string[] | '*';
```
rohan-bes, line 1: Nice use of string literal type here!

packages/report-server/src/reportBuilder/types.ts
```diff
@@ -8,3 +8,11 @@ export type FieldValue = string | number | boolean | undefined | null;
...
   [field: string]: FieldValue;
 }
+
+export interface Matrix {
+  columns: {
+    key: string;
+    title: string;
+  }[];
+  rows: Row[];
+}
```
rohan-bes, line 18: Might be nice to define the output types in-line with their output builder functions rather at the top-level here.

We can then export a union type of all possible output types with something like in `outputBuilders.ts`:
```
export type OutputType = ReturnType<typeof outputBuilders[keyof typeof outputBuilders]>;
```

packages/report-server/src/types.ts
```diff
@@ -33,6 +33,7 @@ export interface ReportConfig {
     dataGroups?: string[];
```
rohan-bes, line 33: Not your code, but might be nice to add `aggregations?: (string | Record<string, unknown)[];` to this type 👍

packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      }).toThrow();
+    });
+
+    it('can perform with categoryField', () => {
+      const output = buildOutput({
+        matrix: {
+          categoryField: 'InfrastructureType',
+          rowField: 'FacilityType',
+          columns: '*',
+        },
```
rohan-bes, line 85: Ah apologies, I think I forgot to leave this in my notes from after the meeting, but I think the plan was to go with a flat object for the output config, one that looks something like:
```suggestion
const output = buildOutput({
          type: 'matrix', 
          categoryField: 'InfrastructureType',
          rowField: 'FacilityType',
          columns: '*',
        },
```

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { Matrix, Row } from '../../../types';
+import { MatrixBuilder } from './matrixBuilder';
+import { MatrixColumnParams, MatrixParams } from './types';
+
+const matrix = (rows: Row[], params: MatrixParams): Matrix => {
+  const matrixBuilder = new MatrixBuilder(rows, params);
+  matrixBuilder.buildBaseColumns().buildBaseRows();
```
rohan-bes, line 12: It might be nicer if the matrix builder itself were responsible for calling the sub-functions, so that the interface could just be:
```suggestion
  return matrixBuilder.build();
```

with build being: 
```
build() {
  this.buildBaseColumns();
  this.buildBaseRows();
  return this.matrixData;
}
```

Also, just a small naming thing, but why `buildBaseColumns()`/`buildBaseRows()` instead of just `buildColumns()`/`buildRows()` ?

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const getPrefixColumns = (columns: unknown): MatrixColumnParams => {
+    if (columns === '*' || columns === undefined) {
+      return '*';
+    }
+    if (!Array.isArray(columns)) {
+      throw new Error(`Expected columns as string array or '*' but got ${columns}`);
+    }
+    if (columns.length === 0) {
+      return '*';
```
rohan-bes, line 29: I have a slight preference towards throwing an error here, but I'm happy either way 👍

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+    columns.forEach(column => {
+      if (typeof column !== 'string') {
+        throw new Error(`Expected columns as string array or '*' but got ${columns}`);
```
rohan-bes, line 33:
```suggestion
        throw new Error(`Expected each column to be a string but got ${column}`);
```

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
```diff
@@ -0,0 +1,58 @@
...
+  const { columns, categoryField, rowField } = params;
+
+  const prefixColumns = getPrefixColumns(columns);
+
+  if (typeof rowField !== 'string') {
+    throw new Error(`Expected rowField as string but got ${rowField}`);
+  }
+
+  if (typeof categoryField !== 'string') {
+    throw new Error(`Expected categoryField as string but got ${categoryField}`);
```
rohan-bes, line 47: Category field should be optional, no?

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
```diff
@@ -0,0 +1,58 @@
...
+    });
+    return columns;
+  };
+  const { columns, categoryField, rowField } = params;
+
+  const prefixColumns = getPrefixColumns(columns);
+
+  if (typeof rowField !== 'string') {
+    throw new Error(`Expected rowField as string but got ${rowField}`);
+  }
```
rohan-bes, line 44: I suppose we ought to be checking that they haven't included the rowField in the columns, or categories, and vice versa? With something like this:
```suggestion
  }
  
  if (Array.isArray(prefixColumns) && prefixColumns.includes(rowField)) {
    throw new Error(`rowField: ${rowField} is already defined as a column`);
  }
  
```

This is fairly optional tho imo, as that's just a silly configuration mistake 😝

packages/report-server/src/reportBuilder/output/output.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const buildParams = (params: unknown): OutputParams => {
+  const defaultType = { type: 'default', config: undefined };
+
+  if (typeof params === 'undefined') {
+    return defaultType;
+  }
+  if (typeof params === 'object') {
+    if ('matrix' in params) {
```
rohan-bes, line 28: I think once the config is re-worked into being a flat object, it'd be good to push all the outputType specific validation logic into their own `buildParams()` function 👍

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  params: MatrixParams;
+
+  constructor(rows: Row[], params: MatrixParams) {
+    this.rows = rows;
+    this.params = params;
+    this.matrixData = { columns: [], rows: [] };
+  }
+
+  assertIsString(values: unknown[]): asserts values is string[] {
```
rohan-bes, line 17: This isn't a major thing, but personally I much prefer when functions are generic and don't rely on any class object state that they get defined outside the class itself (can still be in the same file, that's fine)

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
```diff
@@ -0,0 +1,72 @@
...
+      this.matrixData.columns = columns.map(c => ({ key: c, title: c }));
+    };
+
+    const { prefixColumns, nonColumnKeys } = this.params.columns;
+    if (Array.isArray(prefixColumns)) {
+      assignColumnSetToMatrixData(prefixColumns);
+      return this;
+    }
+    // When prefixColumns === '*', get columns from rows data
+    const columns = new Set();
```
rohan-bes, line 37:
```suggestion
    const columns = new Set<string>();
```
^Might need to double check me on that sytax, but I think this will remove the need for you to do a string assertion on the columns later on 👍

rohan-bes (changes requested):
Few changes requested, but it looking really good!! Nice work! Love the unit tests as well 👍

packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
billli0: @rohan-bes Sounds good! Do we also want to eliminate other unspecified column keys in rows? We have Laos and Tonga in this expectedData, currently we will shows them all. Perhaps the best practice is to eliminate other unspecified rows data for Tonga?

packages/report-server/src/reportBuilder/types.ts
billli0: Love it! the `ReturnType` is handy and makes things clearer.

packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
billli0: That looks right thanks @rohan-bes .

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
billli0: That's a good question @rohan-bes ! I was thinking we will need more functions here to decorate rows and columns, same as those from `web-config-server`, such as adding total summary at the end of each row, but your question inspires me that `transform` can also handle this case.

`transform` is awesome!

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
billli0: Throwing an error makes more sense 👍

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
billli0: Yes I agree more input defenders is better!

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
billli0: Thanks @rohan-bes , I forgot to let it handle without Category field. I will add a test case as well.

packages/report-server/src/reportBuilder/output/output.ts
billli0: Yeah I re-work them and it looks so clean now 👍

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
billli0: Thanks @rohan-bes, I agree taking it out is the best practice, because we don't want the outer functions to access it.

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
billli0: I've tried to but it seems like `Array.from(set)` is not smart enough to send back results with a string type.

packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
rohan-bes: >Do we also want to eliminate other unspecified column keys in rows?
Yeah, I think that would be best. It's a bit unexpected to have a field in a row that isn't specified in the columns 👍

rohan-bes (changes requested):
Couple minor change requested:
- Get the typescript working so that we don't need to do the `assertIsString()` logic (happy to do a quick pair on this one if it's proving difficult)
- Do some validation on the 'type' key in the base `output.ts` file
- Omit the non-row columns from the actuals rows in the return data

Other comments are optional 😄

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
billli0: Thanks @rohan-bes , you are right! We can get rid of `assertIsString` function. 
I guess TypeScript sometimes is lagging to map the type, but totally that's my fault.

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (categoryField !== undefined && typeof categoryField !== 'string') {
+    throw new Error(`Expected categoryField as string but got ${categoryField}`);
+  }
+  const nonColumnKeys = categoryField ? [categoryField, rowField] : [rowField];
+  if (Array.isArray(prefixColumns) && prefixColumns.includes(rowField)) {
+    throw new Error(`rowField: ${rowField} is already defined as a column`);
+  }
```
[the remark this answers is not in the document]
billli0: love it! Will put them in code.

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.rows = rows;
+    this.params = params;
+    this.matrixData = { columns: [], rows: [] };
+  }
+
+  build() {
+    this.buildColumns();
+    this.buildRows();
+    this.modifyRowsByPrefixColumns();
+    this.buildCategories();
```
billli0, line 27: Extract these two logics from `.buildRows()`, hope that is easier to read.

packages/report-server/package.json
```diff
@@ -39,6 +39,8 @@
     "body-parser": "^1.18.3",
     "cors": "^2.8.5",
     "express": "^4.16.2",
+    "lodash.pick": "^4.4.0",
+    "@types/lodash.pick": "^4.4.0",
     "mathjs": "^7.3.0",
```
billli0, line 44: lodash doesn't packages its declaration files with its npm packages. 
If we don't expect `report-server` to be used as a library, we can put `@types/lodash.pick` to `devDependencies`.

Reference: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies

rohan-bes (changes requested):
Just a small request to add some validation on the `type` field in the `ouput` config, and to move the `@types` import lodash to the root directory 👍 

Renaming suggestions are optional, but you mentioned you hate naming things so I am offering my help in sharing the burden ;P

packages/report-server/src/reportBuilder/output/functions/matrix/types.ts
```diff
@@ -0,0 +1,16 @@
+import { Row } from '../../../types';
+
+export type MatrixColumnParams = string[] | '*';
+
+export type MatrixParams = {
+  columns: { prefixColumns: MatrixColumnParams; nonColumnKeys: string[] };
```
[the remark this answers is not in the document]
billli0: Yeah I agree if that is clearer to you as a reader, then they should be unencapsulated. That is one of the benefits to have others to review code 👍

billli0:
Thanks @rohan-bes for the naming! include and exclude make things like twins now 👍

packages/report-server/src/reportBuilder/output/output.ts
```diff
@@ -35,7 +35,7 @@ const buildParams = (params: unknown): OutputParams => {
     }
     return { type, config: restParams };
   }
-  throw new Error(`Expected output config as object but got ${params}`);
+  return { type: 'default', config: 'undefined' };
 };
```
billli0, line 39: return unchanged data if output is not specified.

packages/report-server/src/reportBuilder/output/output.ts
```diff
@@ -14,13 +14,25 @@ type OutputParams = {
...
 const buildParams = (params: unknown): OutputParams => {
+  const isOutPutBuildersType = (value: string): value is keyof typeof outputBuilders => {
+    switch (value) {
+      case 'default':
+      case 'matrix':
+        return true;
+      default:
+        return false;
+    }
+  };
```
billli0, line 30: Find a hacky way to deal with the object key types verification. Not sure this is a good practice?

packages/report-server/src/reportBuilder/output/functions/matrix/types.ts
```diff
@@ -0,0 +1,14 @@
+import { Row } from '../../../types';
+
+export type MatrixParams = {
+  columns: { include: string[] | '*'; exclude: string[] };
```
rohan-bes, line 4:
```suggestion
  columns: { includeFields: string[] | '*'; excludeFields: string[] };
```
After seeing your code in `matrixBuilder.ts` I feel like you got the terms just right there 👍 
Optional ofcourse :)

packages/report-server/src/reportBuilder/output/output.ts
```diff
@@ -0,0 +1,44 @@
...
+const buildParams = (params: unknown): OutputParams => {
+  const isOutPutBuildersType = (value: string): value is keyof typeof outputBuilders => {
+    switch (value) {
+      case 'default':
+      case 'matrix':
+        return true;
+      default:
+        return false;
+    }
+  };
```
rohan-bes, line 30: Yeah I'd say this is a bit too hacky, and a little brittle. It's always good to design the code in such a way that extensibility is easy. Ideally someone can just add a new ouput type to the `outputBuilders` object and then it can automatically be included in the validation logic, something like:
```suggestion
  const isOutPutBuilderType = (value: string): value is keyof typeof outputBuilders => {
    Object.keys(outputBuilders).includes(value);
  };
```

rohan-bes (approved):
Just one final recommendation to improve the validation logic, but pre-approving since aside from that it's good to go! Great work @billli0 this is gonna be super useful 👍

packages/report-server/src/reportBuilder/output/output.ts
billli0: Oh that is very smart 👍

billli0:
Thanks @rohan-bes, suggestions have been applied.

PR #2636, opened 2021-06-04
Editing user's LESMIS permission group

tcaiger:
### Changes:

- add UpdateUserEntityPermissionRoute to lesmis server
- add put support to server boiler plate
- add useUpdateUserEntityPermission mutation to lesmis
- add UserPermissionsCell to users table
---

### Screenshots:
See issue


packages/lesmis-server/src/connections/MeditrakConnection.ts
```diff
@@ -30,4 +30,24 @@ export class MeditrakConnection extends SessionHandlingApiConnection {
...
+
+    // get permission_group byName
+    const permissionGroups = await this.get('permissionGroups', {
+      filter: JSON.stringify({
+        name: permissionGroupName,
+      }),
+    });
+
+    return this.put(
+      `userEntityPermissions/${userEntityPermissionId}`,
```
EMcQ-BES, line 46: I believe `Tupaia Admin Panel` is required for this endpoint, how are we dealing with that for LESMIS users?

EMcQ-BES (approved):
LGTM, just wanna double check we've considered the permissions against that endpoint.

tcaiger:
> LGTM, just wanna double check we've considered the permissions against that endpoint.

Yes that is a good question and yep we have. When we did the update for showing the users in the table we added `hasLesmisAdminAccess` to the `assertAdminPanelAccess` check and the `EditUserEntityPermission` endpoint is handled with that same permission check.

PR #2637, opened 2021-06-04
HOTFIX: delete all COVIDVac_WS survey responses

nick-beyondessential:
### Issue #: No issue. Remove all Covid Samoa Vaccine survey responses for COVIDVac_WS so that they can be reimported.

Will be tested on feature branch then become HOTFIX.

### DIFF
--- a/packages/database/src/migrations/20210604004448-CovidSamoaVaccineDeleteResponses-modifies-data.js
+++ b/packages/database/src/migrations/20210604004448-CovidSamoaVaccineDeleteResponses-modifies-data.js
@@ -1,0 +1,68 @@
+'use strict';
+
+var dbm;
+var type;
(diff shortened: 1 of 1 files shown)


kael89 (approved):
Looks good Nick! Yes let's please test this on a feature branch first.

Also, we should check that @julianam-w has a chance to export all existing responses as a backup, before we deploy this in dev and production! 🙏

julianam-w:
Got a copy of all files used for import but given the nature of the data , it is not necessary as there is only one truth to the data from Tamanu.

michaelnunan:
Yeah just weighing in to agree with Juliana - Tamanu is the source of truth on this, there's no need to backup this data from Tupaia before deploying :-)

PR #2641, opened 2021-06-07
RN-53 Fix Laos EOC periodType

nick-beyondessential:
### Issue #: [2905](https://github.com/beyondessential/tupaia-backlog/issues/2905) Data does not display accurately for this bar chart unless a very narrow date range

Laos EOC's fetchAggregatedAnalyticsByDhisIds wasn't respecting the config periodType, Returning default month aggregation instead of this case of week, throwing out the dates for the chart. Checking first for query dataPeriodType so as to not overwrite indicator specified dataPeriodType.

### Changes:
- parse config.periodType in the two databuilders that have the 'special' indicator fetch
- add to fetchAggregatedAnalyticsByDhisIds post to DHIS so that data is returned in correct period.


packages/web-config-server/src/apiV1/utils/fetchIndicatorValues/fetchAggregatedAnalyticsByDhisIds.js
```diff
@@ -94,15 +94,17 @@ const performEntityAggregation = async (models, analytics, entityAggregation, hi
 };
 
 const getQueryInput = (query, dataElementIds, organisationUnitIds) => {
+  const dataPeriodType = query.dataPeriodType || query.periodType;
```
chris-bes, line 97: We actually had the same problem for the normal Indicators. I solved it by puting the `dataPeriodType` in the `data_source` config. I personally think that's a more ideal solution because then we don't have to keep specifying the same `dataPeriodType` in the `dataBuilderConfig` for every visualisation that use the Indicator data source. 

However I think this is fine as this whole file is a hack and we don't really want to use aggregated analytics endpoint for Program Indicators. We want to use our normal analytics event  (`analytics/events/query/`) endpoint, which should resolve the issue. Once https://github.com/beyondessential/tupaia-backlog/issues/2607 is solved, we can immediately remove this file and use our normal analytics event  (`analytics/events/query/`) endpoint. With that, we won't have to specify the period type anymore.

chris-bes (approved):

packages/web-config-server/src/apiV1/utils/fetchIndicatorValues/fetchAggregatedAnalyticsByDhisIds.js
nick-beyondessential: Thanks @chris-bes

edmofro:
Closing as we no longer support Laos EOC

PR #2642, opened 2021-06-07
1296 add postcode level data

billli0:
### Issue #:
[#1296](https://github.com/beyondessential/tupaia-backlog/issues/1296)
### Changes:
-  Add 'postcode' in 'entity_type'
- Change covideau entity canonical types

### Manual step:
- Import postcode data in AU with [file](https://github.com/beyondessential/tupaia-backlog/issues/1296#issuecomment-857290079) 
- Add a new tracked entity type in DHIS: [DHIS tracked entity type management](https://dev-aggregation.tupaia.org/dhis-web-commons/security/login.action#/list/programSection/trackedEntityType)


packages/database/src/migrations/20210607053528-AddPostcodeLevelDataInAU-modifies-data.js
```diff
@@ -0,0 +1,77 @@
...
+    const point = { type: 'Point', coordinates: [long, lat] };
+    await insertObject(db, 'entity', entity);
+    await db.runSql(`
+        update entity 
+        SET point = ST_Force2D(ST_GeomFromGeoJSON('${JSON.stringify(point)}'))
+        where code = '${entity.code}';
+    `);
+    // Insert relationships into entity_relation table
+    await db.runSql(
+      `UPDATE entity_hierarchy SET canonical_types = '{country, district, sub_district, postcode}' WHERE name = '${HIERARCHY_CODE}';`,
```
rohan-bes, line 66: This works, right? Even though technically the hierarchy is branched such that:
```
     country
        |
      state
   |        |
postcode    LGA
```
Personally I do find it a touch ugly that we can support via such a definition, but it's a lot easier than defining entity-relations for the state<=>postcode and state<=>LGA mapping. Fine by me 👍

packages/database/src/migrations/20210607053528-AddPostcodeLevelDataInAU-modifies-data.js
```diff
@@ -0,0 +1,77 @@
...
+      country_code: 'AU',
+      metadata: {
+        dhis: {
+          isDataRegional: true,
+        },
+        openStreetMaps: {},
+      },
+    };
+    const point = { type: 'Point', coordinates: [long, lat] };
+    await insertObject(db, 'entity', entity);
```
rohan-bes, line 58: We aren't gonna run into any issues with meditrak syncing or anything due to manually inserting entities via a migration like this are we? @nick-beyondessential you had to work on something similar recently with the covid fiji entity import right? If I recall you imported the entities via the admin panel correct?

rohan-bes (commented):
Just leaving this review as a comment for now as I have a recollection that we actually need to import the entities via the admin panel rather than via a migration. All the other logic to create the `postcode` entity type and update the entity hierarchy seems correct tho 👍

packages/database/src/migrations/20210607053528-AddPostcodeLevelDataInAU-modifies-data.js
nick-beyondessential: Yes best practise is to import via admin-panel for meditrak sideeffects, which may not apply in this particular case, mostly to do with adding facilies to tables etc, we can't be sure of catching all of them, so adding via admin is safer.

packages/database/src/migrations/20210607053528-AddPostcodeLevelDataInAU-modifies-data.js
nick-beyondessential: The canonical_types list is not order specific.

packages/database/src/migrations/20210607053528-AddPostcodeLevelDataInAU-modifies-data.js
billli0: Thanks @nick-beyondessential ! Could you share an example file or the right format for the import file with latitude and longitude?

nick-beyondessential: Something like this:

[importPostcodes.xlsx](https://github.com/beyondessential/tupaia/files/6613351/importPostcodes.xlsx)
https://beyond-essential.slab.com/posts/importing-entities-w8ewpm2b

nick-beyondessential: Seeing as you have already written the migration to input postcodes. Maybe one option would be to run locally then export entities and use that file as the input file.

billli0: Thanks @nick-beyondessential !

packages/database/src/migrations/20210607053528-AddPostcodeLevelDataInAU-modifies-data.js
rohan-bes: That's true! I suppose I just find it confusing that it isn't 😝

packages/database/src/migrations/20210607052816-CreatePostCodeEntityType-modifies-schema.js
```diff
@@ -0,0 +1,42 @@
...
+  seed = seedLink;
+};
+
+exports.up = async function () {
+  const db = new TupaiaDatabase();
+  await db.executeSql(`
+    ALTER TYPE public.entity_type ADD VALUE 'postcode';
+  `);
+  return db.closeConnections();
+};
```
rohan-bes, line 25:
```suggestion
exports.up = async function (db) {
  await db.executeSql(`
    ALTER TYPE public.entity_type ADD VALUE 'postcode';
  `);
};
```
Is there a reason we aren't doing this migration the normal way?

rohan-bes (approved):
Just one comment regarding the way we're running the CreatePostCodeEntityType migration, though I suppose it's fine?

Pre-approving anyway cos it's all looking good 👍

packages/database/src/migrations/20210607052816-CreatePostCodeEntityType-modifies-schema.js
billli0: tbh I don't know, but I think that is our traditional way to add an `entity.type` 😁

billli0:
Will merge to epic branch `epic-flu-tracking-visuals` for testing

PR #2643, opened 2021-06-08
2885: Fix no data being returned when requested entity is at the same level as dataSourceEntityType (branch-rename)

rohan-bes:
Exactly the same as: https://github.com/beyondessential/tupaia/pull/2621 + commit: 09027da

09027da: Added `entity`, `entity-api` and `report` to the `SUBDOMAIN_SUFFIXES` in `validateBranchName.sh`
This is because the meditrak-server is at just `<branch-name>-api` so any branch ending in `report` or `entity` will attempt to hit the `report-server`/`entity-server` when attempting to hit the `meditrak-server`

Tbh it seems a bit like the whole branch name suffix restrictions are getting a tad excessive, and we perhaps should come up with a better system to handle this... But that's out of scope for now


chris-bes (approved):
LGTM!

PR #2644, opened 2021-06-08
Added error handling around auth function in web-config-server

rohan-bes:
Changes to the getAccessPolicyForUser function made now have it throw an error when `undefined` userId is passed in (ie. when no basic auth header is provided), now we need to catch that error


nick-beyondessential (approved):
LGTM

PR #2337, opened 2021-04-05
2311: pt4 -Added prefixed `descendant_` and `ancestor_` url parameters for /relations endpoint

rohan-bes:
### Changes:

Allowed for specifying of fields and filtering of ancestors/descendants in `/relations` endpoint, eg:
```
http://{{host}}/{{version}}/hierarchy/laos_schools/LA/relations?ancestor_type=district&descendant_type=school&field=code&descendant_attributes_type=Secondary&groupBy=descendant
=>
All school->district mappings where descendant.attributes.type = 'Secondary'
```

PR #2648, opened 2021-06-09
2337 adult trapping dashboard strive

billli0:
### Issue 
#2337](https://github.com/beyondessential/tupaia-backlog/issues/2337)
### Changes:
two simple migrations


packages/database/src/migrations/20210524111255-StriveAdultTrappingReportConfig-modifies-data.js
```diff
@@ -0,0 +1,75 @@
...
+            dataSourceEntityType: 'facility',
+          },
+        },
+      ],
+      dataElements: ['STRVEC_AE-AT13'],
+    },
+    transform: [
+      {
+        transform: 'filter',
+        where: "exists($row['value'],$row['dataElement'])",
```
rohan-bes, line 42: Does `exists()` support multiple arguments like this? Perhaps it should... could be good...

Regardless, do we need this transform? I think it's safe to assume that every analytic has a `value` and a `dataElement` 👍

packages/database/src/migrations/20210524111255-StriveAdultTrappingReportConfig-modifies-data.js
```diff
@@ -0,0 +1,75 @@
...
+      },
+      {
+        transform: 'select',
+        '...': ['value', 'dataElement'],
+      },
+      {
+        transform: 'aggregate',
+        value: 'group',
+        dataElement: 'count',
+      },
```
rohan-bes, line 52: You can actually remove fields from the rows during 'aggregate' like this: 
```suggestion
      {
        transform: 'aggregate',
        value: 'group',
        dataElement: 'count',
        '...': 'drop',
      },
```

rohan-bes (changes requested):
Looking good! Just a couple of suggestions on how you can optimise the report transforms 👍

packages/database/src/migrations/20210524111255-StriveAdultTrappingReportConfig-modifies-data.js
billli0: That is handy!

packages/database/src/migrations/20210524111255-StriveAdultTrappingReportConfig-modifies-data.js
billli0: happy to drop this.

rohan-bes (approved):
Nice! looks good

PR #2653, opened 2021-06-10
2879 create flutracker phase two indicators

billli0:
### Issue #:
[#2879](https://github.com/beyondessential/tupaia-backlog/issues/2879)

### Changes:
- Add two indicators

### DIFF
--- a/packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
+++ b/packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
@@ -16,75 +16,98 @@ exports.setup = function (options, seedLink) {
   seed = seedLink;
 };
 
-const INDICATORS = [
-  {
-    id: generateId(),
-    code: 'FluTracker_LGA_Percent_First_Nations_ILI',
-    builder: 'analyticArithmetic',
-    config: {
-      formula: 'firstExistingValue(FWV_LGA_ILI_Percentage_prior_2018, FWV_LGA_ILI_Percentage)',
-      parameters: [
-        {
-          builder: 'analyticArithmetic',
-          code: 'FWV_LGA_ILI_Percentage_prior_2018',
-          config: {
-            formula: 'FluTracking_NationalFNILI',
-            aggregation: [
-              {
-                type: 'FINAL_EACH_DAY',
-                config: {
-                  dataSourceEntityType: 'country',
-                },
+const INDICATOR = {
+  id: generateId(),
+  code: 'FluTracker_LGA_Percent_First_Nations_ILI',
+  builder: 'analyticArithmetic',
+  config: {
+    formula:
+      'firstExistingValue(FWV_LGA_ILI_Percentage_prior_March_2021, FWV_LGA_ILI_Percentage_Till_Current)',
+    parameters: [
+      {
+        builder: 'analyticArithmetic',
+        code: 'FWV_LGA_ILI_Percentage_prior_March_2021',
+        config: {
+          formula: 'FluTracking_NationalFNILI',
+          aggregation: ['FINAL_EACH_DAY'],
+        },
+      },
+      {
+        builder: 'analyticArithmetic',
+        code: 'FWV_LGA_ILI_Percentage_Till_Current',
+        config: {
+          formula:
+            '(FWV_LGA_004a_Total_By_Country >= 0 and FWV_LGA_003a_Total_By_Country > 0) ? (FWV_LGA_004a_Total_By_Country / FWV_LGA_003a_Total_By_Country * 100) : 0',
+          parameters: [
+            // Sum up 'FWV_LGA_004a' per period in national level as 'FWV_LGA_004a_Total_By_Country'
+            {
+              builder: 'analyticArithmetic',
+              code: 'FWV_LGA_004a_Total_By_Country',
+              config: {
+                formula: 'FWV_LGA_004a',
+                aggregation: [
+                  'FINAL_EACH_DAY',
+                  {
+                    type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+                    config: {
+                      dataSourceEntityType: 'sub_district',
+                      aggregationEntityType: 'country',
+                    },
+                  },
+                ],
               },
-            ],
+            },
+            // Sum up 'FWV_LGA_003a' per period in national level as 'FWV_LGA_003a_Total_By_Country'
+            {
+              builder: 'analyticArithmetic',
+              code: 'FWV_LGA_003a_Total_By_Country',
+              config: {
+                formula: 'FWV_LGA_003a',
+                aggregation: [
+                  'FINAL_EACH_DAY',
+                  {
+                    type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+                    config: {
+                      dataSourceEntityType: 'sub_district',
+                      aggregationEntityType: 'country',
+                    },
+                  },
+                ],
+              },
+            },
+          ],
+          aggregation: [],
+          defaultValues: {
+            FWV_LGA_004a_Total_By_Country: 'undefined',
+            FWV_LGA_003a_Total_By_Country: 'undefined',
           },
         },
-        {
-          builder: 'analyticArithmetic',
-          code: 'FWV_LGA_ILI_Percentage',
-          config: {
-            formula: 'FWV_LGA_004a / FWV_LGA_003a',
-            aggregation: [
-              {
-                type: 'FINAL_EACH_DAY',
-                config: {
-                  dataSourceEntityType: 'sub_district',
-                  aggregationEntityType: 'requested',
-                },
-              },
-            ],
-          },
-        },
-      ],
-      aggregation: {
-        FWV_LGA_ILI_Percentage_prior_2018: 'FINAL_EACH_DAY',
-        FWV_LGA_ILI_Percentage: 'FINAL_EACH_DAY',
       },
-      defaultValues: {
-        FWV_LGA_ILI_Percentage_prior_2018: 'undefined', // to differentiate between no data and 0
-        FWV_LGA_ILI_Percentage: 'undefined',
-      },
+    ],
+    aggregation: {
+      FWV_LGA_ILI_Percentage_prior_March_2021: 'FINAL_EACH_DAY',
+      FWV_LGA_ILI_Percentage_Till_Current: 'FINAL_EACH_DAY',
+    },
+    defaultValues: {
+      FWV_LGA_ILI_Percentage_prior_March_2021: 'undefined',
+      FWV_LGA_ILI_Percentage_Till_Current: 'undefined',
     },
   },
-];
+};
 
 exports.up = async function (db) {
-  for (const indicator of INDICATORS) {
-    await insertObject(db, 'indicator', indicator);
-    await insertObject(db, 'data_source', {
-      id: generateId(),
-      code: indicator.code,
-      type: 'dataElement',
-      service_type: 'indicator',
-    });
-  }
+  await insertObject(db, 'indicator', INDICATOR);
+  await insertObject(db, 'data_source', {
+    id: generateId(),
+    code: INDICATOR.code,
+    type: 'dataElement',
+    service_type: 'indicator',
+  });
 };
 
 exports.down = async function (db) {
-  for (const indicator of INDICATORS) {
-    await deleteObject(db, 'indicator', { code: indicator.code });
-    await deleteObject(db, 'data_source', { code: indicator.code });
-  }
+  await deleteObject(db, 'indicator', { code: INDICATOR.code });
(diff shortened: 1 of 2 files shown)


packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
```diff
@@ -0,0 +1,115 @@
...
+          formula: 'FluTracking_NationalFNILI',
+          aggregation: ['FINAL_EACH_DAY'],
+        },
+      },
+      {
+        builder: 'analyticArithmetic',
+        code: 'FWV_LGA_ILI_Percentage_Till_Current',
+        config: {
+          formula:
+            '(FWV_LGA_004a_Total_By_Country >= 0 and FWV_LGA_003a_Total_By_Country > 0) ? (FWV_LGA_004a_Total_By_Country / FWV_LGA_003a_Total_By_Country * 100) : 0',
```
rohan-bes, line 40:
```suggestion
            'FWV_LGA_003a_Total_By_Country === 0 ? 0 : FWV_LGA_004a_Total_By_Country / FWV_LGA_003a_Total_By_Country',
```

Couple of things going on here:
1. No need to check if the numerator is greater than 0. We do it for the denominator as div by 0 can cause unexpected behaviour, but if for whatever reason the numerator is tracked as a negative number we should just let it be.
2. We generally treat percentages in our system as decimals, ie. `50%` = `0.5`, so no need to multiple by 100

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+const INDICATOR = {
+  id: generateId(),
+  code: 'FluTracker_LGA_Percent_First_Nations_ILI',
+  builder: 'analyticArithmetic',
+  config: {
+    formula:
+      'firstExistingValue(FWV_LGA_ILI_Percentage_prior_March_2021, FWV_LGA_ILI_Percentage_Till_Current)',
+    parameters: [
```
rohan-bes, line 26: While nested indicator parameters is a really cool design tool, and allows for us to create easily readable indicator formulas as you've done here, right now we haven't really optimised the performance of them so they should be used sparingly. I think given that each parameter here is using a unique dataElementCode means we can get away with just a shallow indicator, such as:
```
  config: {
    formula:
      'firstExistingValue(FluTracking_NationalFNILI, FWV_LGA_003a=== 0 ? 0 : FWV_LGA_004a / FWV_LGA_003a)',
    aggregation: {
      FluTracking_NationalFNILI: 'FINAL_EACH_DAY',
      FWV_LGA_003a: ['FINAL_EACH_DAY', {
        type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
        config: {
          dataSourceEntityType: 'sub_district',
          aggregationEntityType: 'country',
        },
      } ],
      FWV_LGA_004a: ['FINAL_EACH_DAY', {
        type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
        config: {
          dataSourceEntityType: 'sub_district',
          aggregationEntityType: 'country',
        },
      } ],
    },
    defaultValues: {
      FluTracking_NationalFNILI: 'undefined',
      FWV_LGA_003a: 'undefined',
      FWV_LGA_004a: 'undefined'
    }
  }
```

packages/database/src/migrations/20210610060144-AddFluTrackerLGAPercentNonFirstNationsILI-modifies-data.js
```diff
@@ -0,0 +1,89 @@
...
+  seed = seedLink;
+};
+
+const INDICATOR = {
+  id: generateId(),
+  code: 'FluTracker_LGA_Percent_Non_First_Nations_ILI',
+  builder: 'analyticArithmetic',
+  config: {
+    formula:
+      '(FWV_LGA_004b_Total_By_Country >= 0 and FWV_LGA_003b_Total_By_Country > 0) ? (FWV_LGA_004b_Total_By_Country / FWV_LGA_003b_Total_By_Country * 100) : 0',
```
rohan-bes, line 25: Similar comments here about:
- No need to check numerator is >= 0
- Best not to multiply by 100
- Should be able to unpack this indicator to not use indicator parameters 👍

rohan-bes (changes requested):
Nice work putting together these indicators! Logic looks mainly good, and in theory using parameters to make the final formula more readable is a good move, but making them not use nesting should yield better performance 👍

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
billli0: That is a good question @rohan-bes ! But I have noticed the result of `FWV_LGA_ILI_Percentage_prior_March_2021` have a few greater than 1 (100%), and the result of `FWV_LGA_ILI_Percentage_Till_Current` is very small, like 0.0006. But yeah that should be a question to data stewards. I will ask on Tuesday.

billli0: @rohan-bes For point one, checking numerator value is to see if undefined, do you think indicator will not throw error if undefined? We can trust the data input though.

rohan-bes: Hmm that's a good point, I think it's very likely that sometimes we will end up with `undefined / undefined`

While I think the best solution would be for us to migrate the `divide()` function from the report-server to the `expression-parser` to allow for division of undefined (should just return undefined) unfortunately there's no good place in the code for those functions to sit...

I think for now we can just use a more verbose formula, eg:
`FWV_LGA_004a_Total_By_Country === 'undefined' or FWV_LGA_003a_Total_By_Country === 'undefined' ? 'undefined' : FWV_LGA_003a_Total_By_Country === 0 ? 0 :  (FWV_LGA_004a_Total_By_Country / FWV_LGA_003a_Total_By_Country * 100))`

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
billli0: I have tried but seems like `firstExistingValue()` doesn't support expression in the parameters. I agree it would be nice to have this function.

We can keep moving things forward to use parameters. For that new function do you want me to make a card? @rohan-bes

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
```diff
@@ -0,0 +1,74 @@
...
+            dataSourceEntityType: 'sub_district',
+            aggregationEntityType: 'country',
+          },
+        },
+      ],
+    },
+    defaultValues: {
+      FluTracking_NationalFNILI: -1,
+      FWV_LGA_003a: -1,
+      FWV_LGA_004a: -1,
```
billli0, line 52: preset them as -1, which will make formula easier to check if undefined. 
We couldn't use undefined or 'undefined', because I tried `equalText(FluTracking_NationalFNILI, 'undefined')` and `FluTracking_NationalFNILI` is number type, which will throw error.

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
```diff
@@ -0,0 +1,74 @@
...
+  seed = seedLink;
+};
+
+const INDICATOR = {
+  id: generateId(),
+  code: 'FluTracker_LGA_Percent_First_Nations_ILI',
+  builder: 'analyticArithmetic',
+  config: {
+    formula:
+      "firstExistingValue(FluTracking_NationalFNILI == -1 ? 'undefined': FluTracking_NationalFNILI / 100, ((FWV_LGA_003a == 0 or FWV_LGA_004a == -1 or FWV_LGA_003a == -1) ? 'undefined': FWV_LGA_004a / FWV_LGA_003a))",
```
billli0, line 25: Previously we check if value is undefined. We don't need to because if they are, they will be assigned as `defaultValue`.

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
billli0: ^ignore me

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
billli0: `FluTracking_NationalFNILI` / 100 because `FluTracking_NationalFNILI` is in percentage.

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
rohan-bes: Ah clever solution here! Nice one 👍

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
```diff
@@ -0,0 +1,74 @@
...
+      FWV_LGA_003a: [
+        'FINAL_EACH_DAY',
+        {
+          type: 'SUM_PER_PERIOD_PER_ORG_GROUP',
+          config: {
+            dataSourceEntityType: 'sub_district',
+            aggregationEntityType: 'country',
+          },
+        },
+      ],
```
rohan-bes, line 37: This aggregation config worked out in the end? Any idea what was going wrong the other day?

packages/database/src/migrations/20210610060144-AddFluTrackerLGAPercentNonFirstNationsILI-modifies-data.js
```diff
@@ -0,0 +1,72 @@
...
+  seed = seedLink;
+};
+
+const INDICATOR = {
+  id: generateId(),
+  code: 'FluTracker_LGA_Percent_Non_First_Nations_ILI',
+  builder: 'analyticArithmetic',
+  config: {
+    formula:
+      'FWV_LGA_003b == -1 or FWV_LGA_003b == -1 or FWV_LGA_003b == 0 ? undefined : FWV_LGA_004b / FWV_LGA_003b',
```
rohan-bes, line 25: Is `undefined` supported here? I didn't realised mathjs native supported it!

packages/database/src/migrations/20210610060144-AddFluTrackerLGAPercentNonFirstNationsILI-modifies-data.js
```diff
@@ -0,0 +1,72 @@
...
+          config: {
+            dataSourceEntityType: 'sub_district',
+            aggregationEntityType: 'country',
+          },
+        },
+      ],
+    },
+    defaultValues: {
+      FWV_LGA_003b: -1,
+      FWV_LGA_004b: -1,
```
rohan-bes, line 50: I believe you could omit the `defaultValues` for this indicator? The normal indicator logic is to not produce an analytic unless all data elements are present for that period/entity. Since we only want the case where both data elements are present here, not using `defaultValues` is probably the way to go!

rohan-bes (approved):
Great work @billli0 

Left a comment that you may be able to not use `defaultValues` for the `FluTracker_LGA_Percent_Non_First_Nations_ILI` indicator, but pre-approving since things are looking really good 👍

packages/database/src/migrations/20210607020237-AddIndicatorsForAuFlutracking-modifies-data.js
billli0: The fomula was missing a single quatation mark ', but the error said 'aggregation: {error}', which was a bit confusing.

packages/database/src/migrations/20210610060144-AddFluTrackerLGAPercentNonFirstNationsILI-modifies-data.js
billli0: Emm it did support, but Chris suggested me to use 'undefined' and I forgot to change this one. To make them consistent I will add quotes to them.

packages/database/src/migrations/20210610060144-AddFluTrackerLGAPercentNonFirstNationsILI-modifies-data.js
billli0: Cool I will delete `defaultValues` here. So if not using `defaultValues` it will not produce analytic if all are presented?

PR #2654, opened 2021-06-11
Added 'served_build/*' to lesmis .gitignore

rohan-bes:
Build gets generated here on deployment instances so worth ignoring for when deving directly on a feature instance


kael89 (approved):
Awesome work!!! 👏

PR #2655, opened 2021-06-11
2864:  Use known dimensions for dashboard report tests

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2864


packages/web-frontend/cypress/scripts/generateConfig/generateReportConfig.js
```diff
@@ -3,212 +3,62 @@
...
 
-/**
- * We generate the least amount of config required to test each report once.
- * 1. For each report, use the last dashboard group that includes it
- * 2. For each dashboard group, use the first project that includes it
- * 3. For each dashboard group, use an organisation unit from the orgUnitMap config
- * that matches the group's org unit code and level
- */
-export const generateReportConfig = async db => {
+export const generateReportConfig = () => {
```
kael89, line 55: Two notes for this function:

1. I believe in the future we may still want to use the dynamic url building functionality that is deleted here. A use case is: if a specific project is selected
  * find all reports for that project
  * for each report, try crawlling through org units and periods until data is found (in the front end). We could configure a period/org unit range per project, striking a balance between ease of configuration and performance
  
    But this is just food for thought - for now, lets keep it simple and delete whatever is not required

2. Currently, the user must first run `yarn cypress:config` to build the report urls, then run the tests. You may ask: why not build the urls inside the tests themselves? There are two reasons for that:
  i. We may want to enable more elaborate filtering/url building. Eg I adding a `--project strive` option in our e2e test runner could be helpful
  ii. The test runner should have an overview of the url count before tests run, if we are to introduce parallelization

packages/web-frontend/cypress/config/dashboardReports.json
```diff
@@ -1 +1,6 @@
-[]
+{
+  "_description": "A list of urls that will be used for testing dashboard reports. It can either be manually populated, or generated using the contents of params/dashboardReports.json",
```
edmofro, line 2: Nice hybrid model!

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
```diff
@@ -3,12 +3,31 @@
...
+const checkHasMatrixData = body => {
+  const { rows = [], columns = [] } = body;
+
+  const hasColumnCategories = columns[0]?.columns;
+  const getColumnKeys = cols => cols.map(c => c.key);
+  const columnKeys = hasColumnCategories
+    ? columns.map(c => getColumnKeys(c.columns)).flat()
+    : getColumnKeys(columns);
+
+  return rows.some(row => Object.keys(row).some(key => columnKeys.includes(key)));
```
edmofro, line 19: This matrix one feels a little fragile (i.e. if we changed the data structure this would also need to change), but fortunately we will find out as soon as it breaks by virtue of it being part of the e2e test suite!

packages/web-frontend/cypress/scripts/generateConfig/generateReportConfig.js
edmofro: 1. Totally agree, and we can always find this code in the version history
2. Makes sense. I would like to keep an eye on this one, and see if it strikes the right balance in practice. An alternative could be something like building as a postinstall step, so that we're actually committing the generated config at least as the default urls, or adding a `yarn cypress:build-and-run` script. But best to get this into the wild and see how people actually use it and whether there are pain points

edmofro (approved):
All LGTM! Loving the simplifications

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
kael89: Indeed and indeed!

PR #2656, opened 2021-06-11
2931 batch process responses

edmofro:
### Issue #:

### Changes:

- If you upload a file with 100 or less survey responses, there is no behaviour change
- If you upload a file with more than 100 survey responses, it will run through the validation synchronously, but then offload the database updates to happen in batches asynchronously and email the user when done
- The admin panel will show a message in the modal indicating that the user will be emailed (see screenshot)
- The email will contain a list of any survey responses that failed during db updates (though this should be rare as the validation still happens and informs them in the admin panel)

Note that the values for batch size and cooldown time are just guesses at the moment, I'd like to do some playing around on a deployed instance and work out what values work well, but that shouldn't hold up review.

---

### Screenshots: 

### DIFF
--- a/packages/meditrak-server/src/routes/importSurveyResponses/importSurveyResponses.js
+++ b/packages/meditrak-server/src/routes/importSurveyResponses/importSurveyResponses.js
@@ -29,7 +29,7 @@ import {
 import { assertCanImportSurveyResponses } from './assertCanImportSurveyResponses';
 import { assertAnyPermissions, assertBESAdminAccess } from '../../permissions';
 import { SurveyResponseUpdateBatcher } from './SurveyResponseUpdateBatcher';
-import { processUpdatesAndEmail } from './processUpdatesAndEmail';
+import { emailResults } from './emailResults';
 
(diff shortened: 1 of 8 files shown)


packages/meditrak-server/src/routes/importSurveyResponses/importSurveyResponses.js
```diff
@@ -38,183 +39,181 @@ export async function importSurveyResponses(req, res) {
...
+            throw new ImportValidationError(
+              `Invalid survey response id ${columnHeader} causing message: ${
+                error.message
+              } at column ${columnIndex + 1} on tab ${tabName}`,
+            );
           }
         }
+      }
+
+      const infoValidator = new ObjectValidator(constructInfoColumnValidators(models));
```
edmofro, line 110: There really aren't as many changes as this diff would have you believe, particularly in the section below this line - it just seems to have got the line matching a bit wrong. Might be easier to eyeball the two files side by side rather than trust the red and green

PR #2657, opened 2021-06-11
2936: Improve PSSS performance for countries with many sites

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2936:

Had 2 key improvements in mind for improving performance for this issue:
1. 76c37b7 - Caching responses from the entity-server in the Aggregator for the lifetime of a single analytics/events fetch
- Had toyed with this idea a few times and had seen good results from it in my testing. Implementing this brought the performance back to a reasonable response time
- Little curious to hear the reviewer's thoughts on this one. While it seems to work effectively, I could see the approach as being a bit hacky. Do we have concerns about a proliferation of various 'caches' on the session object? Are there memory consumption concerns?

2. afabc1b - Re-wrote the `MultiEntity...Routes` to use a single db fetch to get the their descendants/relatives/etc. rather than looping through each requested entity
- This one seemed like a bit of a no-brainer, but surprisingly didn't actually improve performance that much. I suspect that's because, after the first request, we cache the db response so everything is done in memory from then on
- Regardless, it did add a minor performance improvement, and I think it's a worthwhile change. Gives me more confidence that the system could handle a request for 1000+ entities when we need to support that

I'd still like to implement a test suite for this provided there's time. Would give more confidence in being able to make changes to the internal logic of the routes. However chucking up for review now to get early feedback 👍

### DIFF
--- a/packages/aggregator/src/connections/EntityConnection.js
+++ b/packages/aggregator/src/connections/EntityConnection.js
@@ -13,6 +13,22 @@ export class EntityConnection extends ApiConnection {
   constructor(session) {
     const { getAuthHeader } = session;
     super({ getAuthHeader });
+
+    session.aggregatorEntityConnectionCache = session.aggregatorEntityConnectionCache || {};
+    this.cache = session.aggregatorEntityConnectionCache;
+  }
+
+  // functionArguments should receive the 'arguments' object
+  getCacheKey = (url, query) => `${url}:${JSON.stringify(query)}`;
+
+  runCachedFetch(url, query) {
+    const cacheKey = this.getCacheKey(url, query);
+    if (this.cache[cacheKey]) {
+      return this.cache[cacheKey];
+    }
+
+    this.cache[cacheKey] = this.get(url, query); // may be async, in which case we cache the promise to be awaited
+    return this.cache[cacheKey];
   }
 
   async getDataSourceEntities(
@@ -21,7 +37,7 @@ export class EntityConnection extends ApiConnection {
     dataSourceEntityType,
     dataSourceEntityFilter = {}, // TODO: Add support for dataSourceEntityFilter https://github.com/beyondessential/tupaia-backlog/issues/2660
   ) {
-    return this.get(`hierarchy/${hierarchyName}/relatives`, {
+    return this.runCachedFetch(`hierarchy/${hierarchyName}/relatives`, {
       entities: entityCodes.join(','),
       descendant_filter: `type:${dataSourceEntityType}`,
       field: 'code',
@@ -47,12 +63,12 @@ export class EntityConnection extends ApiConnection {
       query.ancestor_filter = `type:${aggregationEntityType}`;
     }
 
-    const response = await this.get(`hierarchy/${hierarchyName}/relationships`, query);
-
-    const formattedRelationships = {};
-    Object.entries(response).forEach(([descendant, ancestor]) => {
-      formattedRelationships[descendant] = { code: ancestor };
+    return this.runCachedFetch(`hierarchy/${hierarchyName}/relationships`, query).then(response => {
+      const formattedRelationships = {};
+      Object.entries(response).forEach(([descendant, ancestor]) => {
+        formattedRelationships[descendant] = { code: ancestor };
+      });
+      return [Object.keys(formattedRelationships), formattedRelationships];
     });
-    return [Object.keys(formattedRelationships), formattedRelationships];
   }
 }
--- a/packages/entity-server/src/routes/hierarchy/relationships/MultiEntityRelationshipsRoute.ts
+++ b/packages/entity-server/src/routes/hierarchy/relationships/MultiEntityRelationshipsRoute.ts
@@ -4,28 +4,11 @@
  */
 
 import { Route } from '@tupaia/server-boilerplate';
-import { MultiEntityRelationshipsRequest, RelationshipsResponseBody } from './types';
+import { MultiEntityRelationshipsRequest } from './types';
 import { ResponseBuilder } from './ResponseBuilder';
 
 export class MultiEntityRelationshipsRoute extends Route<MultiEntityRelationshipsRequest> {
   async buildResponse() {
-    const { entities } = this.req.ctx;
-
-    const allResponses: RelationshipsResponseBody = {};
-    await Promise.all(
-      entities.map(async entity => {
-        const singleCtx = { ...this.req.ctx, entity };
-        const response = await new ResponseBuilder(
-          this.req.models,
-          singleCtx,
-          this.req.query.groupBy,
-        ).build();
-        Object.entries(response).forEach(([key, value]) => {
-          allResponses[key] = value;
-        });
-      }),
-    );
-
-    return allResponses;
+    return new ResponseBuilder(this.req.models, this.req.ctx, this.req.query.groupBy).build();
   }
 }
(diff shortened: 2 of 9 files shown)


packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -13,6 +13,22 @@ export class EntityConnection extends ApiConnection {
   constructor(session) {
     const { getAuthHeader } = session;
     super({ getAuthHeader });
+
+    session.aggregatorEntityConnectionCache = session.aggregatorEntityConnectionCache || {};
```
rohan-bes, line 17: I rekon this is the part I'm most iffy about on this implementation, somehow just injecting an object into the session and then blindly using between instances feels wrong 😝 

Also, thoughts on the naming convention? I'd actually prefer if this `entityConnectionCache` could be shared between various `EntityConnections` for different parts of the system (eg. `report-server` and `aggregator`) and that's part of why I felt it'd be better just to cache the http request results (so that each client can then manipulate them however they see fit). If that were the case then the `entityConnectionCache` should really be an class which contains all the getCacheKey logic inside it. I shied away from the idea in the end since I didn't quite know where to put that class as it'd be used between packages (probably somewhere in server-boilerplate, `ApiConnection` maybe?)

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -13,6 +13,22 @@ export class EntityConnection extends ApiConnection {
...
+    this.cache = session.aggregatorEntityConnectionCache;
+  }
+
+  // functionArguments should receive the 'arguments' object
+  getCacheKey = (url, query) => `${url}:${JSON.stringify(query)}`;
```
rohan-bes, line 22: I think it can be a bit unreliable using `JSON.stringify()` for building cache strings from objects. I'm it always outputs the keys in the same order.  We use it in our DatabaseModel class though and I think it works alright... Perhaps a 3rd party library like `object-hash` might be better? https://www.npmjs.com/package/object-hash

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -47,12 +63,12 @@ export class EntityConnection extends ApiConnection {
...
-
-    const formattedRelationships = {};
-    Object.entries(response).forEach(([descendant, ancestor]) => {
-      formattedRelationships[descendant] = { code: ancestor };
+    return this.runCachedFetch(`hierarchy/${hierarchyName}/relationships`, query).then(response => {
+      const formattedRelationships = {};
+      Object.entries(response).forEach(([descendant, ancestor]) => {
+        formattedRelationships[descendant] = { code: ancestor };
+      });
+      return [Object.keys(formattedRelationships), formattedRelationships];
```
rohan-bes, line 71: Am I being paranoid here that making this value return from a promise like this could lead to bugs where one caller mutates the object and then all other callers get affected by that? The return objects from this really really should be treated as immutable but there's no way of enforcing that.

Any good ideas on how to ensure that a new object gets returned any time a caller awaits this function?

packages/aggregator/src/connections/EntityConnection.js
edmofro: True! I guess it "works alright" because probably most of the time it comes out the same, and if it doesn't, there's no massive issue other than a bit of extra lag. Still seems worth using `object-hash` though :+1:

packages/aggregator/src/connections/EntityConnection.js
edmofro: It's fair to be a bit paranoid, but this isn't an issue specific to this particular pattern, it's a js wide issue that's more obvious here. 

You can use https://immutable-js.github.io/immutable-js/ for immutable data structures, though I've never actually used it outside of some redux tutorials years ago. I'd say the unexpected interface might make it cause more issues than just keeping the POJO

packages/aggregator/src/connections/EntityConnection.js
edmofro: Yep I had a quick look through this on Friday and this was the main reason I decided not to fully review it and leave it for today. I'd definitely like to find a better place for it than on `session`, especially considering the context is passed in externally and given how static it is, there'd be no reason for an external client not to have that be much more long-lived than it currently is. I reckon `Aggregator` should have the cache stored somewhere that's entirely within its control, and then it can do its own cache invalidation based on whatever strategy ends up making sense. 

Is it too simplistic to have the cache just live on the `EntityConnection` instance, and then lift the instantiation of the entity connection up to happen at the same time the `Aggregator` class is created and subsequently be shared by all calls via that `aggregator`?

edmofro: We could also have a call about this when you're in tomorrow!

edmofro (changes requested):
Thanks for pursuing these thoughts! Let's talk more about where the cache should live, and potentially whether the "per connection" cache invalidation strategy is the most appropriate (it probably is!)

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -47,12 +63,12 @@ export class EntityConnection extends ApiConnection {
       query.ancestor_filter = `type:${aggregationEntityType}`;
     }
 
-    const response = await this.get(`hierarchy/${hierarchyName}/relationships`, query);
-
-    const formattedRelationships = {};
-    Object.entries(response).forEach(([descendant, ancestor]) => {
-      formattedRelationships[descendant] = { code: ancestor };
+    return this.runCachedFetch(`hierarchy/${hierarchyName}/relationships`, query).then(response => {
```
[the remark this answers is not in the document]
rohan-bes: I think we can chalk this up to me still being a bit unfamiliar with the implementation details of `async/await` logic. When I was first playing around with this code I think I accidentally wrote it in such a way that using `await` was resulting in worse performance, and at the time switching to `.then()` improved things. 

At the time I'd been under the impression that calling `await` within an async function was some kind of blocking operation for the caller of that function, but now I'm not sure if that's true... I'm happy to re-write this using `await` and see what results I get 👍

packages/entity-server/src/routes/hierarchy/relationships/MultiEntityRelationshipsRoute.ts
```diff
@@ -4,28 +4,11 @@
...
-          this.req.query.groupBy,
-        ).build();
-        Object.entries(response).forEach(([key, value]) => {
-          allResponses[key] = value;
-        });
-      }),
-    );
-
-    return allResponses;
+    return new ResponseBuilder(this.req.models, this.req.ctx, this.req.query.groupBy).build();
```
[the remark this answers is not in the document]
rohan-bes: I should be fair about the performance, it does improve first time requests by a noticeable amount, however since we cache the results after that it's hard to notice the difference.

Worth noting though that even the worst performing countries in `PSSS` only have around 400 facilities, if we had to do this for all schools in Laos or something then we'd probably see more benefit from this re-write 👍

packages/aggregator/src/connections/EntityConnection.js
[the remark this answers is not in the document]
edmofro: Cool let me know how you go. I'm fairly sure it won't be any more blocking than what's here...but promises are weird!

rohan-bes:
@edmofro Totally hear you on the concerns around how and where the caching is done for the `EntityConnection`, also quite interested to hear about alternative invalidation strategies? Let's organise a time to have a chat about this today 👍

rohan-bes:
Closing out this PR as it's being split off into 2 other PRs:
https://github.com/beyondessential/tupaia/pull/2723 - entityApi cache
https://github.com/beyondessential/tupaia/pull/2727 - Using single db fetch to for multiEntityRoutes

PR #2661, opened 2021-06-13
FQS links open in new tab

edmofro:
This is just a suggestion - @enunan do you think the FQS links should open in a new tab or just replace the LESMIS page and stay on the same tab?

### DIFF
--- a/packages/lesmis/src/views/PageView.js
+++ b/packages/lesmis/src/views/PageView.js
@@ -105,7 +105,7 @@ export const TwoColumnPageView = ({ content }) => {
                   </CardHeader>
                   <CardBody>
                     {links.map(({ name, link }) => (
-                      <Link key={name} href={link}>
+                      <Link key={name} href={link} target="_blank" rel="noopener">
                         <ArrowForwardIcon />
                         {name}
                       </Link>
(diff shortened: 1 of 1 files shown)


tcaiger (approved):

packages/lesmis/src/views/PageView.js
```diff
@@ -105,7 +105,7 @@ export const TwoColumnPageView = ({ content }) => {
                   </CardHeader>
                   <CardBody>
                     {links.map(({ name, link }) => (
-                      <Link key={name} href={link}>
+                      <Link key={name} href={link} target="_blank" rel="noopener">
```
tcaiger, line 108:
```suggestion
                      <Link key={name} href={link} target="_blank">
```

I don't think we need to add rel any more. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener
Setting target="_blank" on <a> elements now implicitly provides the same rel behavior as setting rel="noopener"

edmofro: Ah cool! I guess there's a bit of a risk someone might be using e.g. Chrome from last year, but considering we know the links that are being opened I'm happy to remove it :+1:

PR #2663, opened 2021-06-14
Calculate user rewards dynamically

edmofro:
### Issue #:

https://github.com/beyondessential/tupaia-backlog/issues/2935

### Changes:

- Calculate a user's coconuts and pigs based on the number of survey responses they've submitted at the time the endpoint is hit
- Calculate leaderboard using same tactic
- Delete the table formerly responsible for holding user rewards
- Some other simplifications and cleanup

Motivation:
One less side effect being triggered after survey response submission/import/edit, meaning
1. Less complexity
2. Potential performance improvement with less db interactions on bulk uploads (haven't tested this, but worth it for point 1 anyway)


packages/database/src/TupaiaDatabase.js
```diff
@@ -407,45 +407,6 @@ export class TupaiaDatabase {
     return this.delete('setting', { key });
   }
 
-  /**
```
edmofro, line 410: This wasn't being used anywhere, and from the example appears to have been implemented for the user_reward use case specifically, so I thought it was best to delete it and keep the TupaiaDatabase as free of dead debris as possible

packages/database/src/migrations/20210614003043-RemoveUserRewardsTable-modifies-schema.js
```diff
@@ -0,0 +1,27 @@
...
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`DROP TABLE user_reward`);
+};
+
+exports.down = function (db) {
```
edmofro, line 21: No down because I would have had to regenerate all the existing rewards, and it'll almost certainly never be used, so haven't bothered

packages/meditrak-server/src/routes/getSocialFeed.js
```diff
@@ -94,8 +94,7 @@ const getLeaderboardFeedItem = async models => {
     creation_date: new Date(),
     template_variables: {
       title: 'Tupaia Leaderboard',
-      // Enable this to display pigs in leaderboard once pig ledger is accurate.
-      hasPigs: false,
+      hasPigs: false, // this can be set to true if the meditrak-app table is fixed
```
edmofro, line 97: Pig ledger is now correct, but I tried turning it on and the table in meditrak overflows weirdly. I don't imagine we'll ever fix it, so could probably just delete all the related code, but have already deleted a lot in this PR and that would involve a meditrak release so have left for another day (and perhaps we will fix it up and make it look pretty)

packages/meditrak-server/package.json
```diff
@@ -34,7 +34,6 @@
     "async": "^2.6.2",
     "aws-sdk": "^2.451.0",
     "body-parser": "^1.18.3",
-    "cached": "^4.3.2",
```
edmofro, line 37: The user reward and leaderboard endpoints previously had their sql query results cached, I guess as a performance optimisation, but:
- the cache was only valid for 20 seconds before being marked stale
- the queries themselves (or the new ones at least) are super quick, so could be premature optimisation
- in practise with the cache removed I haven't seen any performance degradation
- it was one of many cache layers/strategies we have around the place, so thought it was better to delete it to remove a point of complexity

packages/meditrak-server/src/routes/getSocialFeed.js
EMcQ-BES: Is there a ticket for that?

edmofro: No, because we'll never do it

EMcQ-BES (approved):
Looks good, pretty minimal actual logic changes, mostly just deletions and moving things around.

A thought that doesn't need to be addressed right now, we have some old tickets for configurable rewards per project that may be affected by this change.

PR #2668, opened 2021-06-14
2931 process responses in background

edmofro:
### Issue #:

After playing with https://github.com/beyondessential/tupaia/pull/2656 I have a suspicion that the batching and cooldown is unnecessary, and the biggest win is just having each survey response added in its own transaction rather than having one massive transaction hit all in one go

This removes the batching from that PR. I still need to test this

### Changes:

- If you upload a file with 100 or less survey responses, there is no behaviour change
- If you upload a file with more than 100 survey responses, it will run through the validation synchronously, but then offload the database updates to happen asynchronously and email the user when done
- The admin panel will show a message in the modal indicating that the user will be emailed (see screenshot)
- The email will contain a list of any survey responses that failed during db updates (though this should be rare as the validation still happens and informs them in the admin panel)

---

### Screenshots: 

### DIFF
--- a/packages/admin-panel/src/importExport/ImportModal.js
+++ b/packages/admin-panel/src/importExport/ImportModal.js
@@ -66,9 +66,10 @@ export const ImportModalComponent = React.memo(
       setFileName(noFileMessage);
     };
 
-    const handleCancel = () => {
+    const handleClose = () => {
       setStatus(STATUS.IDLE);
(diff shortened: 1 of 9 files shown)


edmofro:
This crashed after importing 11 of Juliana's spreadsheets in a row, marking as a failure and closing

PR #2671, opened 2021-06-15
E2e tests: Allow reusing custom dashboard report configuration

kael89:
### Issue #:
[No issue]

This allows users to save url configuration for dashboard reports, and reuse it in the future. For example, if someone wants to test reports in `covidau` and `strive`, they could

1. (if it doesn't exist already) create custom url configuration for each project. Commit the configuration files in the repo for future use
2. Specify the desired configuration in `cypress.json`:

```json
   {
     "tupaia": {
       "dashboardReportParamFiles": [
         "cypress/config/params/dashboardReports/covidau.json",
         "cypress/config/params/dashboardReports/strive.json"
       ]
     }
   }
```

I made the start by adding config for `covidau` reports, which can be used to test https://github.com/beyondessential/tupaia-backlog/issues/797. We could just add all those reports in the base `default.json` config, but we probably don't want to test all reports all the time by default.


packages/web-frontend/cypress/scripts/generateConfig/generateReportConfig.js
```diff
@@ -39,23 +42,33 @@ const buildReportPeriod = (startDate, endDate) => {
...
+
+  return dashboardReportParamFiles.map(readJsonFile).flat();
+};
+
 export const generateReportConfig = () => {
   const logger = getLoggerInstance();
 
-  const urls = allReportParams.map(buildUrl).sort(compareAsc);
+  const urlObjects = getUrlObjects();
+  const urls = uniq(urlObjects.map(buildUrl)).sort(compareAsc);
```
edmofro, line 71: I like the `sort` in here, that's a thoughtful touch

edmofro (approved):
lgtm!

PR #2673, opened 2021-06-15
2814: Integrate new dashboard structure into tupaia.org

chris-bes:
### Changes:

- Integrate new dashboard structure into tupaia.org

### TODO:
- Admin panel tabs for new tables https://github.com/beyondessential/tupaia-backlog/issues/2939


packages/lesmis-server/src/routes/ReportRoute.ts
```diff
@@ -26,20 +27,23 @@ export class ReportRoute extends Route {
     switch (type) {
       case 'dashboard': {
         if (legacy === 'true') {
-          const legacyReport = await this.webConfigConnection.fetchDashboardReport({
+          const legacyReport = await this.webConfigConnection.fetchDashboardReport(reportCode, {
```
chris-bes, line 30: A bit awkward here because web-config-server report endpoint will also check for `legacy`. But I'm not too fussed about it 😬

packages/web-config-server/src/apiV1/RouteHandler.js
```diff
@@ -54,10 +61,17 @@ export class RouteHandler {
...
-  fetchProject = async () =>
-    this.models.project.findOne({ code: this.query.projectCode || 'explore' });
+  fetchAndCacheProject = async () => {
+    if (!this.project) {
+      this.project = await this.models.project.findOne({
+        code: this.query.projectCode || 'explore',
+      });
+    }
+
+    return this.project;
```
chris-bes, line 71: We need project in order to perform permission check for `report` endpoint

packages/web-config-server/src/apiV1/index.js
```diff
@@ -56,14 +55,13 @@ export const getRoutesForApiV1 = () => {
     handleWith(OrganisationUnitHandler),
   );
   api.get('/organisationUnitSearch', handleWith(OrgUnitSearchHandler));
-  api.get('/dashboard', handleWith(DashBoardHandler)); // Legacy style dashboards
-  api.get('/view', handleWith(ViewHandler));
+  api.get('/dashboard', handleWith(DashboardHandler)); // Legacy style dashboards
```
chris-bes, line 58: I will remove this endpoint and also drop `dashboardReport` + `dashboardGroup` in the next PR (and also add Admin Panel tabs for the new tables)

packages/web-config-server/src/apiV1/report.js
```diff
@@ -0,0 +1,123 @@
...
+        : await this.buildReportData(reportCode);
+
+    return {
+      ...reportData,
+      startDate: this.startDate,
+      endDate: this.endDate,
+    };
+  };
+
+  async buildReportData(reportCode) {
```
chris-bes, line 52: I actually haven't tested this route with any new reports yet. Solutions team implemented some new `reportServerDataBuilder` dashboards but they haven't been tested. Once they're tested, I can migrate them to use the new structure to test if this is working

packages/web-frontend/src/components/View/SingleValueWrapper.js
```diff
@@ -26,14 +26,16 @@ import { formatDataValue } from '../../utils';
 
 export class SingleValueWrapper extends PureComponent {
   render() {
-    const { name, valueType, value, total, value_metadata } = this.props.viewContent;
-    const metadata = value_metadata || this.props.viewContent[`${name}_metadata`];
+    const { name, valueType, data, total, value_metadata: valueMetadata } = this.props.viewContent;
```
chris-bes, line 29: Single value reports no longer returns just `value`. I changed this I can remove `viewJson` from the `report` endpoint, and also it can be consistent with how other dashboard item types receiving data

packages/web-frontend/src/containers/DashboardGroup/index.js
```diff
@@ -39,31 +39,24 @@ export default class DashboardGroup extends Component {
...
+
+    const { dashboardCode, entityCode: organisationUnitCode, project } = tab;
 
     // Map views to DashboardItem containers
-    const infoViews = uniqueViews.map(view => {
-      const uniqueViewId = getUniqueViewId(view);
-      if (!view.requiresDataFetch) {
+    const infoViews = tab.items.map(view => {
+      const uniqueViewId = getUniqueViewId(organisationUnitCode, dashboardCode, view.code);
+      if (view.noDataFetch) {
```
chris-bes, line 48: This was determinded in the view.js endpoint before by checking if `dataBuilder` is null. Now I moved it to dashboard_item config

packages/web-frontend/src/screens/mobile/HomeScreen/index.js
```diff
@@ -16,7 +15,7 @@ import { Dashboard } from '../../../components/mobile/Dashboard';
 import { setOrgUnit, toggleDashboardSelectExpand, setDashboardGroup } from '../../../actions';
 import { WHITE } from '../../../styles';
 import {
-  selectCurrentDashboardGroupCode,
+  selectCurrentDashboardName,
```
chris-bes, line 18: So before, in a lot of places, the variable/method names are often `dashboardGroupCode`, even though they are `dashboardName` instead of codes. I found it a bit confusing. So I tried to change the URL to used `dashboardCode` instead of `dashboardName`, but this is tricky because in a lot of places we only have access to `dashboardName` (eg: project default project is reference by `dashboardName`). So I went with changing the variable/method names from `dashboardGroupCode` to `dashboardName` instead to avoid confusion

chris-bes:
One thing I want to ask is, with the new dashboards endpoint. If a dashboard is empty, it won't be returned to the client. However, in LESMIS, there are a few dashboards that are empty. (screenshot below)
image

I don't know if we have to maintain the empty dashboards? If for any reason we have to, then a solution is to add an empty dashboard_item in these until we have real reports?

packages/web-config-server/src/apiV1/dashboards.js
```diff
@@ -3,33 +3,64 @@
...
+    if (!dashboardsWithItems.length && possibleDashboardRelations.length) {
+      return this.getStaticDashboard(entity, NO_ACCESS_DASHBOARD_ITEM_CODE);
+    }
+
+    return dashboardsWithItems;
+  };
+
+  getDashboardsWithItems = async (dashboards, entity, projectCode) => {
+    const permissionGroups = await this.req.getUserGroups(entity.code);
     const sortedDashboards = orderBy(dashboards, ['sort_order', 'name']);
```
chris-bes, line 47: Is it okay to sort all dashboards by name? The order will be different from how it is now in tupaia.org

chris-bes:
One thing I want to ask is, with the new dashboards endpoint. If a dashboard is empty, it won't be returned to the client. However, in LESMIS, there are a few dashboards that are empty. (screenshot below)
image

I don't know if we have to maintain the empty dashboards? If for any reason we have to, then a solution is to add an empty dashboard_item in these until we have real reports?

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
```diff
@@ -93,6 +96,13 @@ const EnlargedDialogComponent = ({
   // Regardless of the drillDown level, we pass the base view content through
   const baseViewContent = contentByLevel?.[0]?.viewContent;
   const drillDownContent = contentByLevel?.[drillDownState.drillDownLevel]?.viewContent;
+  const baseViewConfig = contentByLevel?.[0]?.viewConfig;
+  const drillDownConfig = contentByLevel?.[drillDownState.drillDownLevel]?.viewConfig;
```
EMcQ-BES, line 100: I'm not entirely sure how `contentByLevel` is being calculated? Does that need to be updated for the new way drill downs are laid out?

packages/web-config-server/src/apiV1/report.js
EMcQ-BES: Oh I have a test report I've been using, so I ran your branch to see if it works.

image

EMcQ-BES (commented):
Looks great Chris. A couple of questions but those are mostly for my benefit since I'm not super familiar with the web-frontend code

Looks like you there's only a couple of places with big changes and the rest is just updating variable names and passing data through the whole chain.

edmofro:
> One thing I want to confirm is, with the new dashboards endpoint, if a dashboard is empty, it won't be returned to the client. However, in LESMIS, there are a few dashboards that are empty. (screenshot below)
> image
> 
> I don't know if we have to maintain the empty dashboards? If for any reason we have to, then a solution is to add an empty dashboard_item in these until we have real reports?

Thanks for pointing this out @chris-bes. There was a reason - it's how UNICEF wanted it. Not sure if it's still important, and have asked Erin [here](https://github.com/beyondessential/tupaia/pull/2673#issuecomment-861884177)

packages/web-frontend/src/containers/DashboardItem/index.js
```diff
@@ -54,41 +54,37 @@ export class DashboardItem extends Component {
   componentWillReceiveProps(nextProps) {
     if (
       nextProps.infoViewKey !== this.props.infoViewKey ||
-      this.props.organisationUnit !== nextProps.organisationUnit ||
-      this.props.disasters !== nextProps.disasters
```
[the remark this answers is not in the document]
chris-bes: It is actually redundant because this component never receive any `disasters` props from else where as I've checked.

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialog.js
chris-bes: Yep so it stores the viewConfig/viewContent of each level of drill down.

chris-bes:
@EMcQ-BES Sorry for addressing this PR quite late. I've pushed 2 commits:
https://github.com/beyondessential/tupaia/pull/2673/commits/c74e5188cc8fbd6f44df3c64ddde7d270d6cc5d9: Update new drill down + some fixes for some bugs that I've just found
https://github.com/beyondessential/tupaia/pull/2673/commits/bca092d03898ad78526980f6d15ab9583bec7b29: Addressed your review comments from both this PR and the former PR.

packages/database/src/migrations/20210625012001-FixCustomLegacyReports-modifies-data.js
```diff
@@ -0,0 +1,230 @@
...
+    code: 'Disaster_Response_Comparisons',
+    oldDataBuilderConfig: {
+      dataElementPairs: [['DP_NEW001', 'DP9']],
+    },
+    newDataBuilderConfig: {
+      dataElementPairs: [['DP_NEW001', 'DP9']],
+      leftColumnHeader: 'Normal',
+      rightColumnHeader: 'Current',
+      dataPairNames: ['Inpatient beds'],
+    },
```
chris-bes, line 18: A bit of context, there are actually a few legacy data builders that reference `viewJson`. They have access to it because the previous `view` endpoint queries it. With the new `report` endpoint which should only reference either the modern report or the legacy report tables, it no longer queries for `viewJson` (now called `config` in dashboard_item). 

Most of the `viewJson` references in these legacy data builders are just hardcoded strings which is required to calculate the data. So what I did here is just move whatever they were referencing from `viewJson` before to `data_builder_config`. It is not ideal, but I feel that since these data builders are legacy, I really don't think it's worth to spend time to refactor them to fit the new endpoint structure.

packages/web-config-server/src/apiV1/dataBuilders/generic/comparison/compareDataElementPairs.js
```diff
@@ -18,13 +18,17 @@ import flatten from 'lodash.flatten';
  * }
  */
 export const compareDataElementPairs = async (
-  { dataBuilderConfig, viewJson, query },
+  { dataBuilderConfig, query },
```
chris-bes, line 21: removing `viewJson` reference from legacy data builders like I explained above

edmofro:
@EMcQ-BES according to https://github.com/beyondessential/tupaia/pull/2673#issuecomment-868173347 just https://github.com/beyondessential/tupaia/commit/bca092d03898ad78526980f6d15ab9583bec7b29 left to review - if you can do that we can get this whole load of stuff into regression testing!

EMcQ-BES:
[bca092d](https://github.com/beyondessential/tupaia/commit/bca092d03898ad78526980f6d15ab9583bec7b29) looks perfectly good to me, the fixes it makes are mostly for readability rather than functionality

EMcQ-BES (approved):
I admittedly haven't re-looked at everything, but the changes since my previous review all look good to me

chris-bes:
I have to admit that I'm not very confident with this fix since I'm not familiar with how styling in mobile works. But I have done some testing and it seems fine so far https://github.com/beyondessential/tupaia/pull/2673/commits/f51dc58045b4a7f27d30b6893377a10c1e9e0468

edmofro:
@chris-bes your fix looks good (both @tcaiger and I had a look). That sneaky mv fast refresh change is back, but I'll get that sorted now.

chris-bes:
@tcaiger https://github.com/beyondessential/tupaia/pull/2673/commits/ae3dfa12e4924a6ae46cc0f5818f756274e6c28c LGTM!

PR #2675, opened 2021-06-16
2961: Allow creating/removing indexes on Analytics Table as the Tupaia database user

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2961:

The core issue at play here is that the `owner` of the analytics table must be the `mvrefresh` role, however only a table `owner` or a `superuser` can add/remove indexes on a table in Postgres. 

There could have been a few ways to address this (open a connection as that `mvrefresh` user, temporarily grant `superuser` to the `tupaia` user, etc.) however I felt the most solid going forward would be to add a function to the `mvrefresh` module that allows us to add/remove indexes. Since the `tupaia` user has full access to the functions in that module, it can use these functions to add/remove indexes from the analytics table 👍 

#### Changes:
Updated pg-mv-fast-refresh and analytics table build scripts
- New version pg-mv-fast-refresh version has functions that allow the tupaia user to add/remove indexes from the analytics table
- New scripts to drop analytics table and uninstall the mvrefresh module

#### Deployment steps:
- [x] https://github.com/beyondessential/pg-mv-fast-refresh/pull/2 must already be merged in `@beyondessential/pg-mv-fast-refresh`
- [ ] A full re-installation of the `mvrefresh` module and the analytics table is required, takes ~10 mins of server downtime

### DIFF
--- a/packages/data-api/package.json
+++ b/packages/data-api/package.json
@@ -17,9 +17,11 @@
     "test": "jest",
     "test:coverage": "yarn test --coverage",
     "install-mv-refresh": "scripts/installMvRefreshModule.sh",
+    "uninstall-mv-refresh": "scripts/uninstallMvRefreshModule.sh",
     "build-analytics-table": "scripts/buildAnalyticsMaterializedView.sh",
-    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh",
-    "full-rebuild-analytics-table": "scripts/fullRebuildAnalyticsMaterializedView.sh"
+    "drop-analytics-table": "scripts/dropAnalyticsMaterializedView.sh",
+    "full-rebuild-analytics-table": "scripts/fullRebuildAnalyticsMaterializedView.sh",
+    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh"
   },
   "dependencies": {
     "@tupaia/database": "1.0.0",
--- a/packages/data-api/scripts/buildAnalyticsMaterializedView.sh
+++ b/packages/data-api/scripts/buildAnalyticsMaterializedView.sh
@@ -3,5 +3,4 @@ source .env
 cd scripts
 export PGPASSWORD=$DB_PASSWORD
 psql -h $DB_URL -d $DB_NAME -U $DB_USER -f createAnalyticsMaterializedView.sql
-export PGPASSWORD=$DB_MV_PASSWORD
-psql -h $DB_URL -d $DB_NAME -U $DB_MV_USER -f createAnalyticsTableIndexes.sql
+psql -h $DB_URL -d $DB_NAME -U $DB_USER -f createAnalyticsTableIndexes.sql
--- a/packages/data-api/scripts/createAnalyticsTableIndexes.sql
+++ b/packages/data-api/scripts/createAnalyticsTableIndexes.sql
@@ -13,7 +13,7 @@ begin
     AND c.relkind = 'i')
   THEN
     tStartTime := clock_timestamp();
-    PERFORM mv$addIndexToMv$Table(mv$buildAllConstants(), 'public', 'analytics', 'analytics_data_element_entity_date_idx', 'data_element_code, entity_code, date desc');
+    PERFORM mv$addIndexToMv$Table(mv$buildAllConstants(), 'public', 'analytics', tAnalyticsIndexName, 'data_element_code, entity_code, date desc');
     RAISE NOTICE 'Created analytics index, took %', clock_timestamp() - tStartTime;
   ELSE
     RAISE NOTICE 'Analytics index already exists, skipping';
@@ -25,7 +25,7 @@ begin
     AND c.relkind = 'i')
   THEN
     tStartTime := clock_timestamp();
-    PERFORM mv$addIndexToMv$Table(mv$buildAllConstants(), 'public', 'analytics', 'analytics_data_group_entity_event_date_idx', 'data_group_code, entity_code, event_id, date desc');
+    PERFORM mv$addIndexToMv$Table(mv$buildAllConstants(), 'public', 'analytics', tEventsIndexName, 'data_group_code, entity_code, event_id, date desc');
     RAISE NOTICE 'Created events index, took %', clock_timestamp() - tStartTime;
   ELSE
     RAISE NOTICE 'Events index already exists, skipping';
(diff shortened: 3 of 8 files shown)


packages/data-api/package.json
```diff
@@ -17,9 +17,11 @@
     "test": "jest",
     "test:coverage": "yarn test --coverage",
     "install-mv-refresh": "scripts/installMvRefreshModule.sh",
+    "uninstall-mv-refresh": "scripts/uninstallMvRefreshModule.sh",
     "build-analytics-table": "scripts/buildAnalyticsMaterializedView.sh",
-    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh",
-    "full-rebuild-analytics-table": "scripts/fullRebuildAnalyticsMaterializedView.sh"
+    "drop-analytics-table": "scripts/dropAnalyticsMaterializedView.sh",
+    "full-rebuild-analytics-table": "scripts/fullRebuildAnalyticsMaterializedView.sh",
+    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh"
```
rohan-bes, line 24: Now to fully reinstall the `mvrefresh` module and analytics table you can just do:
```
yarn workspace @tupaia/data-api drop-analytics-table
yarn workspace @tupaia/data-api uninstall-mv-refresh
yarn workspace @tupaia/data-api install-mv-refresh
yarn workspace @tupaia/data-api build-analytics-table
```
Worth capturing in a single script?

packages/data-api/scripts/buildAnalyticsMaterializedView.sh
```diff
@@ -3,5 +3,4 @@ source .env
 cd scripts
 export PGPASSWORD=$DB_PASSWORD
 psql -h $DB_URL -d $DB_NAME -U $DB_USER -f createAnalyticsMaterializedView.sql
-export PGPASSWORD=$DB_MV_PASSWORD
-psql -h $DB_URL -d $DB_NAME -U $DB_MV_USER -f createAnalyticsTableIndexes.sql
+psql -h $DB_URL -d $DB_NAME -U $DB_USER -f createAnalyticsTableIndexes.sql
```
rohan-bes, line 6: Index creation can just be done as `tupaia` user now 🎉

packages/data-api/scripts/createAnalyticsTableIndexes.sql
```diff
@@ -1,16 +1,34 @@
...
-  tStartTime := clock_timestamp();
-  CREATE INDEX IF NOT EXISTS analytics_data_group_entity_event_date_idx ON public.analytics(data_group_code, entity_code, event_id, date desc);
-  RAISE NOTICE 'Created [data_group, entity, event, date] index, took %', clock_timestamp() - tStartTime;
+  IF (SELECT count(*) = 0
+    FROM pg_class c
+    WHERE c.relname = tAnalyticsIndexName 
+    AND c.relkind = 'i')
+  THEN
+    tStartTime := clock_timestamp();
+    PERFORM mv$addIndexToMv$Table(mv$buildAllConstants(), 'public', 'analytics', 'analytics_data_element_entity_date_idx', 'data_element_code, entity_code, date desc');
```
rohan-bes, line 16: I chose not to use `CREATE INDEX IF NOT EXISTS...` in the `mvrefresh` functions I added, perhaps I should have? Refer to that PR here: https://github.com/beyondessential/pg-mv-fast-refresh/pull/2

packages/data-api/package.json
edmofro: Very nice! On the fence about another script, because there are a few here to wade through. I think maybe docs in the readme would be even more useful?

packages/data-api/scripts/createAnalyticsTableIndexes.sql
edmofro: That's what your query for whether the `pg_class` entry exists fits in though right? So I'm totally fine with it. Interested in the reason though?

packages/data-api/scripts/dropAnalyticsTableIndexes.sql
```diff
@@ -1,16 +1,34 @@
...
-  tStartTime := clock_timestamp();
-  DROP INDEX IF EXISTS analytics_data_element_entity_date_idx;
-  RAISE NOTICE 'Dropped [data_element, entity, date] index, took %', clock_timestamp() - tStartTime;
+  IF (SELECT count(*) > 0
+    FROM pg_class c
+    WHERE c.relname = tAnalyticsIndexName 
+    AND c.relkind = 'i')
+  THEN
+    tStartTime := clock_timestamp();
+    PERFORM mv$removeIndexFromMv$Table(mv$buildAllConstants(), 'analytics_data_element_entity_date_idx');
```
edmofro, line 16:
```suggestion
    PERFORM mv$removeIndexFromMv$Table(mv$buildAllConstants(), tAnalyticsIndexName);
```
Same below

packages/data-api/scripts/createAnalyticsTableIndexes.sql
edmofro, line 16:
```suggestion
    PERFORM mv$addIndexToMv$Table(mv$buildAllConstants(), 'public', 'analytics', tAnalyticsIndexName, 'data_element_code, entity_code, date desc');
```
Same below for events

packages/data-api/scripts/dropAnalyticsTableIndexes.sql
edmofro: Or is this not possible because that variable is declared as a `TEXT` value or something...?

edmofro (changes requested):
Only a request to reuse the variables you're declaring, but perhaps it's not possible

packages/data-api/scripts/createAnalyticsTableIndexes.sql
rohan-bes: Yup, the whole code block basically just checks if an index with that name exists:
```
SELECT count(*) = 0
    FROM pg_class c
    WHERE c.relname = tAnalyticsIndexName 
    AND c.relkind = 'i'
```
So this script will handle the if exists/not exists case gracefully. Just was wondering about the `mv$addIndexToMv$Table` function in general, but I felt it might be a bit more consistent to not use `IF EXISTS` in that function

packages/data-api/scripts/dropAnalyticsTableIndexes.sql
rohan-bes: Ah great catch, no I just forgot to do that!

kael89 (approved):

PR #2677, opened 2021-06-16
2931 process survey response import in the background

edmofro:
### Issue #:

### Changes:

- If you upload a file that takes 10 seconds or less to process, there is no change
- If you upload a file that takes more than 10 seconds, it will continue in the background and email the user when done
- The admin panel will show a message in the modal indicating that the user will be emailed (see screenshot in issue)
- The email will contain any validation messages or db update failures

---

### Screenshots: 


edmofro:
Superceded

PR #2683, opened 2021-06-17
Hot fix import in background

edmofro:
Fixes two issues found during investigation of a crash on import:
- Changes in `SurveyResponse.js` relate to https://github.com/beyondessential/tupaia-backlog/issues/2931#issuecomment-863602178
- Changes in `SurveyResponseUpdatePersistor.js` relate to https://github.com/beyondessential/tupaia-backlog/issues/2931#issuecomment-863593789

### DIFF
--- a/packages/meditrak-server/src/database/models/SurveyResponse.js
+++ b/packages/meditrak-server/src/database/models/SurveyResponse.js
@@ -52,11 +52,11 @@ class SurveyResponseType extends DatabaseType {
 }
 
 export class SurveyResponseModel extends MaterializedViewLogDatabaseModel {
+  notifiers = [onChangeMarkAnswersChanged];
+
   get DatabaseTypeClass() {
     return SurveyResponseType;
   }
-
-  notifiers = [onChangeMarkAnswersChanged];
 
   isDeletableViaApi = true;
 
--- a/packages/meditrak-server/src/database/models/SurveyResponse.js
+++ b/packages/meditrak-server/src/database/models/SurveyResponse.js
@@ -102,6 +102,7 @@ const onChangeMarkAnswersChanged = async (
   { new_record: newRecord, old_record: oldRecord, record_id: surveyResponseId },
   models,
 ) => {
+  // No need to mark any answers changed if freshly creating or fully deleting
   if (!newRecord || !oldRecord) return;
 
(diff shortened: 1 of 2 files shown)


packages/meditrak-server/src/database/models/SurveyResponse.js
```diff
@@ -52,23 +52,14 @@ class SurveyResponseType extends DatabaseType {
...
 export class SurveyResponseModel extends MaterializedViewLogDatabaseModel {
+  notifiers = [onChangeMarkAnswersChanged];
+
   get DatabaseTypeClass() {
     return SurveyResponseType;
   }
 
   isDeletableViaApi = true;
 
-  updateById(id, fieldsToUpdate) {
```
edmofro, line 61: Pulling the side effect out of `updateById` has four positive benefits:
1. Means the `markAsChanged` happens in the singleton db rather than the transacting db, so can reuse the existing pub/sub connection
2. Avoids overriding a central method of the base class
3. Makes it more robust to the entity or data time changing via mechanisms other than `updateById`
4. Side effects happen completely asynchronously to the actual update, and afterwards, and now won't be triggered if the update has some db error and doesn't actually process

packages/meditrak-server/src/routes/importSurveyResponses/SurveyResponseUpdatePersistor.js
```diff
@@ -10,17 +10,19 @@ export const UPDATE = 'update';
...
     const answerCount = create.answers.upserts.length;
     const recordCount = answerCount + 1; // +1 for the response itself
     if (currentBatchRecordCount + recordCount > RECORDS_PER_BULK_BATCH) {
-      batches.push([]);
       currentBatchIndex++;
       currentBatchRecordCount = 0;
     }
+    if (!batches[currentBatchIndex]) {
+      batches[currentBatchIndex] = [];
+    }
```
kael89, line 25: Great catch by the testers!

packages/meditrak-server/src/database/models/SurveyResponse.js
kael89: Good points, sounds great!

kael89 (approved):
🙌

PR #2686, opened 2021-06-18
2338 Strive average mortality matrix table 

billli0:
### Issue #:
[#2338](https://github.com/beyondessential/tupaia-backlog/issues/2338) 

### Changes:
- migrations
- `reportServerBuilder` to handle matrix table data


packages/web-config-server/src/apiV1/dataBuilders/generic/reportServer/reportServerDataBuilder.js
```diff
@@ -39,6 +39,10 @@ export class ReportServerBuilder extends DataBuilder {
 
     const response = await this.reportConnection.fetchReport(this.config.reportCode, requestQuery);
 
+    // Matrix table data
+    if (response.results.columns && response.results.rows) {
+      return { ...response.results };
+    }
```
billli0, line 45: We should return {columns: {}, rows: {}, ...} instead of {data: {columns: {}, rows: {}}, ...} . For bar chart we need to return {data: {...}, ...}.

Ideally we can let `report server` output step to handle that. But we can do it in the future if we need a new functionality for bar chart in report-server's output step?

rohan-bes: Ahh good point... man these inconsistencies can be soo annoying hey 😝 

I would say that yeah, this should all actually just be handled by the report-server output step, but obviously that's too large a refactor right now so lets create a ticket for it?

packages/database/src/migrations/20210618033026-PGStriveAverageMortalityMatrixTable-modifies-data.js
```diff
@@ -0,0 +1,164 @@
...
+    },
+    output: {
+      type: 'matrix',
+      rowField: 'insecticide',
+      columns: '*',
+      categoryField: 'mosquitoType',
+    },
+    transform: [
+      {
+        '...': ['STRVEC_AE-IR09', 'STRVEC_AN-IR09', 'orgUnitName'],
```
rohan-bes, line 48: Someone has been copying their transform logic from the database 😝 

Totally trivial, but if ya feel like it re-ordering to have the `transform` setting on top and the `...` at the bottom helps with readability 👍

packages/database/src/migrations/20210618033026-PGStriveAverageMortalityMatrixTable-modifies-data.js
```diff
@@ -0,0 +1,164 @@
...
+      ],
+      dataElements: ['STRVEC_AE-IR09', 'STRVEC_AE-IR02', 'STRVEC_AN-IR09', 'STRVEC_AN-IR02'],
+    },
+    output: {
+      type: 'matrix',
+      rowField: 'insecticide',
+      columns: '*',
+      categoryField: 'mosquitoType',
+    },
+    transform: [
```
rohan-bes, line 46: Do we need to that `filter` step to ensure that both dataElements are present for each row? I think otherwise we may end up doing some incorrect defaulting later on 👍

packages/database/src/migrations/20210618033026-PGStriveAverageMortalityMatrixTable-modifies-data.js
```diff
@@ -0,0 +1,164 @@
...
+      {
+        '...': ['STRVEC_AE-IR09', 'STRVEC_AN-IR09', 'orgUnitName'],
+        transform: 'select',
+        "'insecticide'":
+          "exists($row['STRVEC_AE-IR02']) ? $row['STRVEC_AE-IR02'] : $row['STRVEC_AN-IR02']",
+        "'mosquitoType'": "exists($row['STRVEC_AE-IR09']) ? 'Aedes IR' : 'Anopheles IR'",
```
rohan-bes, line 52:
```suggestion
        '...': ['orgUnitName'],
        transform: 'select',
        "'insecticide'":
          "exists($row['STRVEC_AE-IR02']) ? $row['STRVEC_AE-IR02'] : $row['STRVEC_AN-IR02']",
        "'mosquitoType'": "exists($row['STRVEC_AE-IR09']) ? 'Aedes IR' : 'Anopheles IR'",
        "'mortalityRate'": exists($row['STRVEC_AE-IR09']) ? $row['STRVEC_AE-IR09'] : $row['STRVEC_AN-IR09']
```
Optional, but you can do this if you want to not need to refer to both `STRVEC_AE-IR09` and `STRVEC_AE-IR09` throughout the rest of the transform logic if ya want? Would probably make it a bit more readable too 👍

packages/database/src/migrations/20210618033026-PGStriveAverageMortalityMatrixTable-modifies-data.js
```diff
@@ -0,0 +1,164 @@
...
+      isDataRegional: true,
+    },
+  ],
+  dataBuilderConfig: { reportCode: REPORT.code },
+};
+
+const VIEW_JSON = {
+  name: 'Average Mosquito Mortality',
+  type: 'matrix',
+  periodGranularity: 'one_month_at_a_time',
```
rohan-bes, line 94: Might be worth getting a confirmation that we want data 1 month at a time? I know it's there in the ticket screenshot but just not sure if its consistent with the other dashboards we've been building which I think have just been for all time?

rohan-bes (changes requested):
Awesome work @billli0, so nice to see how elegant the config for this really tricky dashboard has turned out to be 🎉 

Couple of suggestions, only real requested change is to add the filter transform to the start of the report?

packages/web-config-server/src/apiV1/dataBuilders/generic/reportServer/reportServerDataBuilder.js
billli0: I think it is pretty straight forward. I will let report-server handle bar chart output in this PR, if you don't mind reviewing it?

packages/database/src/migrations/20210618033026-PGStriveAverageMortalityMatrixTable-modifies-data.js
billli0: How do you know🤯

packages/database/src/migrations/20210618033026-PGStriveAverageMortalityMatrixTable-modifies-data.js
billli0: Yeah that is a good point 👍 . Will add filter here.

packages/database/src/migrations/20210618033026-PGStriveAverageMortalityMatrixTable-modifies-data.js
billli0: That's brillant 👍 . I definitely want it, thanks.

packages/database/src/migrations/20210618033026-PGStriveAverageMortalityMatrixTable-modifies-data.js
billli0: Sure, I will raise a comment.

billli0: Thanks Rohan, we did need another date picker.

packages/report-server/src/reportBuilder/output/functions/default.ts
```diff
@@ -6,5 +6,5 @@
 import { Row } from '../../types';
 
 export const buildDefault = () => {
-  return (rows: Row[]) => rows;
+  return (rows: Row[]) => ({ data: rows });
```
billli0, line 9: we need to change mapoverlay report server after this one is merged.

rohan-bes: Hmmm won't this change also affect `PSSS`/`lesmis` as well?

rohan-bes (changes requested):
I think I'm happy with the call to make the `bar` output type, but I'd prefer if we did as part of a standalone body of work rather than bundled up with this dashboard. 

There's a few dashboards/overlays in test atm that would need to be updated as well, and especially if we plan on changing the `default` output type we have a much larger regression scope.

If you wanna create a separate ticket for that we can move the code form here over to that branch?

billli0:
As disccussed we are not going to support bar output for now.

rohan-bes (approved):
Nice! LGTM!

PR #2690, opened 2021-06-21
Add report test config example

edmofro:
I was looking for one to get started building reports, so thought I might as well save this for future devs

It's based reasonably closely on the [slab doc](https://beyond-essential.slab.com/posts/reporting-package-154s5c5l) so that there's some consistency when learning how to use the reporting package

### DIFF
--- a/packages/report-server/examples.http
+++ b/packages/report-server/examples.http
@@ -1,5 +1,6 @@
 @baseUrl = http://localhost:8030/v1
-@baseUrl = https://dev-report-api.tupaia.org/v1
+# Uncomment below to fetch against the dev server, rather than localhost
+# @baseUrl = https://dev-report-api.tupaia.org/v1
 @contentType = application/json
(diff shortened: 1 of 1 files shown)


packages/report-server/examples.http
```diff
@@ -1,13 +1,62 @@
-@hostname = localhost
-@port = 8030
-@host = {{hostname}}:{{port}}
-@version = v1
+@baseUrl = http://localhost:8030/v1
+# Uncomment below to fetch against the dev server, rather than localhost
+# @baseUrl = https://dev-report-api.tupaia.org/v1
```
edmofro, line 3: I generally find that in reality, rather than changing the version or port etc. I am more likely to flick between dev and localhost, so this is a more practical variable to save (especially with the change from http to https)

kael89 (approved):
Good initiative!

PR #2691, opened 2021-06-21
2971: E2e tests for map overlays

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2971


packages/web-frontend/cypress/config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "dashboardReports": {
+    "allowEmptyResponse": false,
+    "snapshotTypes": ["responseBody", "html"],
+    "urlFiles": ["cypress/config/dashboardReportUrls/default.json"],
+    "urls": []
+  },
+  "mapOverlays": {
+    "allowEmptyResponse": false,
+    "snapshotTypes": ["responseBody"],
```
kael89, line 10: We should create a `cypress/config/mapOverlayUrls/default.json` file and use it as the standard url specification for map overlay tests

packages/web-frontend/cypress/scripts/testE2e/index.js
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { runScript } from '@tupaia/utils';
+import { testE2e } from './testE2e';
+
+runScript(testE2e);
```
kael89, line 11: I accidentally deleted this file in a past commit, this re-instates it

packages/web-frontend/cypress/scripts/generateConfig/generateReportConfig.js
```diff
@@ -3,34 +3,22 @@
...
-  }
+import config from '../../config.json';
+import { readJsonFile } from './helpers';
+
+const objectUrlSchema = {
+  id: [hasContent, isAString],
+  project: [hasContent, isAString],
+  orgUnit: [hasContent, isAString],
+  dashboardGroup: [hasContent, isAString],
 };
```
kael89, line 19: I quite like the declarativeness of our `ObjectValidator`. I personally find schema-based validation having a lot of benefits:

* easy to define accurate schemas
* easier to catch input errors early
* self-documented configuration for devs
* more helpful messages for end users

It seems to me that `ObjectValidator` was mostly used in the earlier days of Tupaia (`meditrak-sever`) but perhaps not many people are aware of tis existence now. That's  a pity because it's so cool!

---

 Side note:

After working in Tamanu [yup](https://www.npmjs.com/package/yup) came in my attention. It is a cute validation library with an elegant syntax. I found it easy to use, the errors seemed a bit more generic than ours but perhaps this is configurable.

I know that we use `ObjectValidator` for some Tupaia-specific checks (ids, async lookups etc). I was wondering whether we could/should leverage an existing library for simpler, `sync` schema validation, like the one in this file? We could do that, or we could generalise `ObjectValidator` a bit - the following cards are about that:

https://github.com/beyondessential/tupaia-backlog/issues/1201
https://github.com/beyondessential/tupaia-backlog/issues/2096

edmofro (approved):
There's a little bit of code duplication between the dashboard reports and map overlays that could be refactored out, but it could be at the expense of simplicity, so I'm really happy with this as is. Nice work, and great call to get at least the response data e2(almost)e tested

packages/web-frontend/cypress/scripts/generateConfig/generateReportConfig.js
edmofro: Yeah I like `yup`! It could be nice to swap out stuff under the hood of `ObjectValidator` for `yup` and see how that goes, maybe replacing `ObjectValidator` entirely in the end. I definitely agree that schema-based validation is very nice :-)

packages/web-frontend/package.json
```diff
@@ -16,8 +16,8 @@
     "build-desktop": "node scripts/build.js && rm -rf builds/desktop && mv build builds/desktop",
     "build-mobile": "REACT_APP_APP_TYPE=mobile node scripts/build.js && rm -rf builds/mobile && mv build builds/mobile",
     "cypress:config": "yarn run:babel cypress/scripts/generateConfig",
-    "cypress:open": "cross-env NODE_ENV=test cypress open",
-    "cypress:run": "cross-env NODE_ENV=test cypress run",
+    "cypress:open": "cross-env NODE_ENV=test yarn cypress:config && yarn cypress open",
+    "cypress:run": "cross-env NODE_ENV=test yarn cypress:config && cypress run",
```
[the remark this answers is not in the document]
kael89: I can - that's an error 😆 Thanks for catching it!

packages/web-frontend/cypress/README.md
```diff
@@ -43,23 +43,75 @@ Then, run one of the following commands in a new terminal:
...
+  ```json
+  {
+    "id": "COVID_Compose_Daily_Deaths_Vs_Cases",
+    "project": "covidau",
+    "orgUnit": "AU",
+    "dashboardGroup": "COVID-19"
+  }
+  ```
+
+Use any of the fields bellow to specify test urls. You can use multiple fields in the same test run - their results will be combined in the final url list:
```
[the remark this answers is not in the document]
kael89: > bellow: (of a person or animal) emit a deep loud roar, typically in pain or anger

Haha probably not what I meant here! 😆 👀

kael89:
Thanks for the review @edmofro ! I do agree that there are some similarities between the test structure of dashboard reports and map overlays, but as you said, I felt that duplicating would result in more maintainable code than abstracting, in this case. But let's keep an eye on how this progresses in the future!

PR #2692, opened 2021-06-21
Fix dashboard by replacing data element

billli0:
### Issue #:
[#2913](https://github.com/beyondessential/tupaia-backlog/issues/2913)

### Changes:
[Comment](https://github.com/beyondessential/tupaia-backlog/issues/2913#issuecomment-864707540)


packages/database/src/migrations/20210621034818-FixSbAtLeastOneStaffMemberTrained-modifies-data.js
```diff
@@ -0,0 +1,44 @@
...
+  seed = seedLink;
+};
+
+const dashboardIds = [
+  'UNFPA_Country_Facilities_offering_services_At_Least_1_Matrix_SB',
+  'UNFPA_Country_Facilities_offering_services_At_Least_1_Matrix_VU',
+];
+
+exports.up = async function (db) {
+  for (const dashboardId of dashboardIds) {
```
edmofro, line 23: Totally fine to do this the way you've done it, but it would be slightly more efficient to do an `id IN (${arrayToDbString(dashboardIds)})` rather than iterate through (probably negligible difference when you're only iterating through two dashboards)

packages/database/src/migrations/20210621034818-FixSbAtLeastOneStaffMemberTrained-modifies-data.js
```diff
@@ -0,0 +1,44 @@
...
+    await db.runSql(`
+      update "dashboardReport" dr
+      set "dataBuilderConfig" = regexp_replace(dr."dataBuilderConfig"::text, 'RHS1UNFPA03','RHS4UNFPA807','g')::jsonb
```
edmofro, line 26: Nice!

edmofro (approved):
lgtm :-)

packages/database/src/migrations/20210621034818-FixSbAtLeastOneStaffMemberTrained-modifies-data.js
billli0: Thanks for sharing! I will use `id IN` next time.

PR #2695, opened 2021-06-21
Add migration to drop non-operational schools

EMcQ-BES:
### Changes:

- Add migration to delete all data and entities from the list provided by Juliana
- Add the list of school codes under `migrationData`

I've marked this as a draft PR currently because we've been requested not to delete these until after the [entity attributes are corrected](https://github.com/beyondessential/tupaia-backlog/issues/3002)


packages/database/src/migrations/20210621215753-LESMISDropNonOperationalSchools-modifies-data.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`
+    -- Drop triggers so the large change notification doesn't slow us down
```
EMcQ-BES, line 21: I dropped these to speed up the migration but it still takes a few minutes, not sure if there's any cascading into other tables that I'm unaware of or just the fact that we're processing `1850` entries at once.

```
[INFO] Processed migration 20210621215753-LESMISDropNonOperationalSchools-modifies-data
Analytics refresh took: 3064ms
✨  Done in 184.95s.```

File: packages/database/src/migrations/20210621215753-LESMISDropNonOperationalSchools-modifies-data.js [hunk 2]
```diff
@@ -0,0 +1,64 @@
...
+    CREATE TRIGGER entity_trigger
+      AFTER INSERT OR UPDATE OR DELETE
+      ON entity
+      FOR EACH ROW EXECUTE PROCEDURE notification();
+  `);
+};
+
+exports.down = function (db) {
+  return null;
+};
```
EMcQ-BES, line 60: Haven't written a down because if we want to re-add this data we should probably do that with an admin-panel import

packages/database/src/migrations/20210621215753-LESMISDropNonOperationalSchools-modifies-data.js
edmofro: a) should use disable/enable rather than drop/create triggers
b) it may be unsafe to not trigger the side effects of these changes...

edmofro: How long does it take if you leave triggers enabled?

EMcQ-BES: Took about an extra 90s, from discussion in person I've removed the drop triggers from this migration.

edmofro (approved):
lgtm

PR #2698, opened 2021-06-22
2993 update web frontend build

tcaiger:
### Changes:

- Add react scripts package
- Remove custom webpack config files
- Remove old webpack dependencies
- Fix linting errors by setting to warnings
- Fix circular dependency issue in utils

---


.eslintrc
```diff
@@ -152,8 +152,32 @@
     {
       "files": ["packages/web-frontend/**"],
       "rules": {
+        "camelcase": "warn",
         "import/no-named-as-default": "off",
-        "react/destructuring-assignment": "off"
+        "import/no-cycle": "warn",
+        "jsx-a11y/click-events-have-key-events": "warn",
```
tcaiger, line 158: By default react-scripts has these lint properties set to error which crashes the build. It would be good to fix all these lint errors at some point but that is out of scope of this PR.

packages/web-frontend/src/components/View/utils.js
```diff
@@ -40,7 +40,7 @@ const VIEW_TYPES = {
   ...SINGLE_VALUE_COMPONENTS,
 };
 
-export const getViewWrapper = ({ type, viewType }) => {
+export function getViewWrapper({ type, viewType }) {
```
tcaiger, line 43: These functions were causing an error on the hot reload when changing view.js. I believe it is because of a circular dependency issue with hot reload. This fixes it because the functions get hoisted.

packages/web-frontend/package.json
```diff
@@ -11,20 +11,22 @@
   "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
   "main": "dist/index.js",
   "scripts": {
+    "start-new": "PORT=8088 react-scripts start",
+    "build-new": "react-scripts build",
```
edmofro, line 15:
```suggestion
```

.eslintrc
edmofro: Do any of them warrant being changed in our central eslint config?

edmofro (approved):
Thanks for this work, very nice to have some of those things cleaned up and feel a little less behind!

.eslintrc
tcaiger: hmmm I don't think so. The other create-react-app apps seem to have the linting options as warn by default. From my lookup, this is because they are on a different version of react-scripts to the one I just installed. It may be that in the future react will enforce these lint rules more strictly so perhaps at that time we can tackle it.

PR #2699, opened 2021-06-23
2850: Add migration for students by gender stacked bar with GPI

EMcQ-BES:
### Issue #: https://github.com/beyondessential/tupaia-backlog/issues/2850

This is branched from `2963-schools-by-education-level` because I needed to use the translate function added by that PR

### Changes:

- Add migration for students by gender stacked bar with GPI

### Screenshots:
See attached ticket

### DIFF
--- a/packages/database/src/migrations/20210622234149-LESMISStudentCountToStackedBar-modifies-data.js
+++ b/packages/database/src/migrations/20210622234149-LESMISStudentCountToStackedBar-modifies-data.js
@@ -1,0 +1,200 @@
+'use strict';
+
+import { insertObject, generateId, findSingleRecord } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210622234149-LESMISStudentCountToStackedBar-modifies-data.js
```diff
@@ -0,0 +1,200 @@
...
+      'nostu_s5_f',
+      'nostu_s5_m',
+      'nostu_s6_f',
+      'nostu_s6_m',
+      'nostu_s7_f',
+      'nostu_s7_m',
+    ],
+    aggregations: [
+      {
+        type: 'SUM_PER_ORG_GROUP',
```
edmofro, line 51: We need a `MOST_RECENT` first, so that we get the most recent submission per school rather than summing across multiple data points

edmofro (changes requested):
I'm going to assume the utility functions are unchanged from what I had and not review that section.

This looks great! There's a risk that we double count some school populations if they submit their data more than once though - a good habit is to filter for the most recent at the school level before aggregating up to a higher level

EMcQ-BES:
> I'm going to assume the utility functions are unchanged from what I had and not review that section.

The only change to the base of the migration is that I split out the portion that creates the `dashboard_relation` because this needs to be connected to a separate dashboard when under a `school` entity

edmofro (approved):

PR #2705, opened 2021-06-23
2635 fix stock used matrix chart in tonga

billli0:
### Issue #:
[#2635](https://github.com/beyondessential/tupaia-backlog/issues/2635)

### Changes:
Since data is migrated into internal data, values changes from `0, 1` to `No, Yes`. Fix it by changing the viewJson config.


packages/database/src/migrations/20210623032402-FixStockCardMatrixChartTonga-modifies-data.js
```diff
@@ -0,0 +1,53 @@
...
+const dashboardId = 'UNFPA_RH_Stock_Cards_Use_Matrix';
+
+exports.up = async function (db) {
+  await db.runSql(`
+      update "dashboardReport" dr
+      set "viewJson" = regexp_replace(dr."viewJson"::text, '\\"presentationOptions\\"\\: \\{','"presentationOptions" : {"type":"condition",','g')::jsonb
+      where id = '${dashboardId}';
+
+      update "dashboardReport" dr
+      set "viewJson" = regexp_replace(dr."viewJson"::text, '\\"key\\"\\: 0','"condition": "No"','g')::jsonb
```
chris-bes, line 26: I'm not sure if Yes/No and 1/0 is the problem here... We've already sanitized internal data to use 1/0. So the current config should have worked. I'm suspecting it's not working because of something else

chris-bes: As discussed, even though raw values are returned as 1/0, `tableOfDataValues` always translate the value using metadata options, plus Yes/No is always returned as default binary options for internal data. So that's why `tableOfDataValues` will always translate 1/0 to Yes/No. 

I suspect there are more matrices using binary data that are having the same problem, and we just haven't found out yet. So a generic solution would be preferred instead of fixing by changing the config in multiple visualisations. We have a couple of solutions:

- Investigate why we always need to translate values to the metadata options in `tableOfDataValues` (we don't do that in any other data builders). I think this is related how we build matrices using DHIS2 data. We've converted a lot of DHIS2 data to internal. So if it's no longer relevant, we can remove it entirely?

- Check in `tableOfDataValues` if the metadata options is the default Yes/No, if it is then don't perform the options translation.

billli0: Thanks @chris-bes ! I am doing the investigation, and one concern is that if we want to download the matrix it will show 1/0 in spreadsheet, but that is not a major issue for now.

packages/web-config-server/src/apiV1/dataBuilders/generic/table/tableOfDataValues/helpers/getValuesByCell.js
```diff
@@ -49,7 +51,9 @@ export const getValuesByCell = (config, results) => {
   return zipObject(
     Object.keys(groupedResults),
     Object.values(groupedResults).map(([{ value, metadata }]) =>
-      metadata && metadata.options ? metadata.options[value] : value,
+      metadata && metadata.options && !isEqual(metadata.options, DEFAULT_BINARY_OPTIONS_OBJECT) // If the metadata options is the default Yes/No, return default value 1/0 instead of translating value.
```
chris-bes, line 54: Nice one!

Could you add more context on this comment. Just so that someone else can also understand the underlying reason why we are not translating default binary options here.

```suggestion
      metadata && metadata.options && !isEqual(metadata.options, DEFAULT_BINARY_OPTIONS_OBJECT) // metadata options are mainly used for translating data from DHIS2 that has category options. For internal data, it returns default Yes/No metadata options for binary data to support translating from 1/0 to Yes/No in a few other places. For matrices, particularly `tableOfDataValues` data builder, we want to use raw 1/0 values and don't want to apply this translation because we want to keep the values consistent. So this check is for disabling that special case.
```

billli0:
Thanks @chris-bes, changes applied.

chris-bes (approved):
LGTM!

PR #2710, opened 2021-06-23
2973 2974: NER and GER reports for country, province, and district

edmofro:
### Issue #:

- https://github.com/beyondessential/tupaia-backlog/issues/2973
- https://github.com/beyondessential/tupaia-backlog/issues/2974

### DIFF
--- a/packages/database/src/migrations/20210623041504-AddNERGERLESMIS-modifies-data.js
+++ b/packages/database/src/migrations/20210623041504-AddNERGERLESMIS-modifies-data.js
@@ -1,0 +1,231 @@
+'use strict';
+
+import { insertObject, generateId, findSingleRecord } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210623041504-AddNERGERLESMIS-modifies-data.js
```diff
@@ -0,0 +1,231 @@
...
+  }).map(configString => JSON.parse(configString));
+
+  for (const config of configs) {
+    await addNewDashboardItemAndReport(db, config);
+  }
+};
+
+exports.down = async function (db) {
+  for (const lesmisEntityLevel of LESMIS_ENTITY_LEVELS) {
+    await removeDashboardItemAndReport(db, CODE.replace('lesmis_entity_level', lesmisEntityLevel));
```
EMcQ-BES, line 225: I don't think this down would work, `CODE` also has the acronym so you'd need to replace that too

```
const CODE = 'LESMIS_enrolment_rate_acronym_lesmis_entity_level';
```

edmofro: Good catch! Updated

EMcQ-BES (approved):
Looks good :+1:

PR #2711, opened 2021-06-24
Support multi color bar chart

edmofro:
If we ever need it, this is how we could support a multi-color bar chart with a single series, like the screenshot below

image


edmofro:
Closing as unrequired

PR #2713, opened 2021-06-24
Filter dashboards into different sections of dropdown

edmofro:
Allows us to have the ESSDP specific dashboard tabs underneath the right sections. Filters based on what the dashboard code starts with (e.g. if code starts with `LESMIS_ESSDP_LowerSecondary` it goes in 'ESSDP Lower secondary sub-sector')

There are some existing UX issues that I haven't fixed as part of this but have made cards for:
- https://github.com/beyondessential/tupaia-backlog/issues/3016
- https://github.com/beyondessential/tupaia-backlog/issues/3017

### DIFF
--- a/packages/lesmis/src/components/Report.js
+++ b/packages/lesmis/src/components/Report.js
@@ -15,6 +15,7 @@ import { ToggleButton } from './ToggleButton';
 import { DashboardReportModal } from './DashboardReportModal';
 import { ChartTable, TABS } from './ChartTable';
 import * as COLORS from '../constants';
+import { DEFAULT_DATE_RANGE } from '../constants';
 import { useDashboardReportData } from '../api/queries';
 import { yearToApiDates } from '../api/queries/utils';
 
@@ -80,7 +81,7 @@ export const Report = React.memo(
   }) => {
     const { code: itemCode, legacy } = viewConfig;
     const [selectedTab, setSelectedTab] = useState(TABS.CHART);
-    const { startDate, endDate } = yearToApiDates(year);
+    const { startDate, endDate } = year ? yearToApiDates(year) : DEFAULT_DATE_RANGE;
     const { data, isLoading, isError, error } = useDashboardReportData({
       entityCode,
       reportCode,
--- a/packages/lesmis/src/components/Report.js
+++ b/packages/lesmis/src/components/Report.js
@@ -15,7 +15,6 @@ import { ToggleButton } from './ToggleButton';
 import { DashboardReportModal } from './DashboardReportModal';
 import { ChartTable, TABS } from './ChartTable';
 import * as COLORS from '../constants';
-import { DEFAULT_DATE_RANGE } from '../constants';
 import { useDashboardReportData } from '../api/queries';
 import { yearToApiDates } from '../api/queries/utils';
 
@@ -81,7 +80,7 @@ export const Report = React.memo(
   }) => {
     const { code: itemCode, legacy } = viewConfig;
     const [selectedTab, setSelectedTab] = useState(TABS.CHART);
-    const { startDate, endDate } = year ? yearToApiDates(year) : DEFAULT_DATE_RANGE;
+    const { startDate, endDate } = yearToApiDates(year);
     const { data, isLoading, isError, error } = useDashboardReportData({
       entityCode,
       reportCode,
--- a/packages/lesmis/src/constants/constants.js
+++ b/packages/lesmis/src/constants/constants.js
@@ -20,7 +20,6 @@ export const SINGLE_YEAR_GRANULARITY = 'one_year_at_a_time';
 export const MIN_DATA_DATE = '20150101';
 export const MIN_DATA_YEAR = '2015';
 export const DEFAULT_DATA_YEAR = '2020';
-export const DEFAULT_DATE_RANGE = { startDate: MIN_DATA_DATE, endDate: '20211231' };
 
 // Layout Constants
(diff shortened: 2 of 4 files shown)


packages/lesmis/src/components/Report.js
```diff
@@ -80,7 +81,7 @@ export const Report = React.memo(
   }) => {
     const { code: itemCode, legacy } = viewConfig;
     const [selectedTab, setSelectedTab] = useState(TABS.CHART);
-    const { startDate, endDate } = yearToApiDates(year);
+    const { startDate, endDate } = year ? yearToApiDates(year) : DEFAULT_DATE_RANGE;
```
tcaiger, line 84: This shouldn't be needed as there is default date handling in `yearToApiDates`. If it's not working for some reason it would be best to fix it in the yearToApiDates. Note: yearToApiDates is also being used by the map overlays.

edmofro: Oh right cool! I think I tried this fix but my bug was I was sending through `year` when I shouldn't, so I can probably just remove it :-)

packages/lesmis/src/constants/constants.js
```diff
@@ -13,6 +13,7 @@ export const SINGLE_YEAR_GRANULARITY = 'one_year_at_a_time';
 export const MIN_DATA_DATE = '20150101';
 export const MIN_DATA_YEAR = '2015';
 export const DEFAULT_DATA_YEAR = '2020';
+export const DEFAULT_DATE_RANGE = { startDate: MIN_DATA_DATE, endDate: '20211231' };
```
tcaiger, line 16: `20211231 ` is  a bit confusing to me, could you make that a named constant or add a comment to give a bit more background to the developer.

edmofro: Not needed

packages/lesmis/src/views/DashboardView.js
```diff
@@ -439,23 +479,30 @@ export const DashboardView = React.memo(() => {
...
+                  value={selectedOption}
+                  onChange={handleChange}
+                  showPlaceholder={false}
+                  SelectProps={{
+                    MenuProps: { disablePortal: true },
+                  }}
+                />
+                {useYearSelector ? (
+                  <YearSelector value={selectedYear} onChange={setSelectedYear} />
+                ) : null}
```
tcaiger, line 502:
```suggestion
              {useYearSelector && (
                  <YearSelector value={selectedYear} onChange={setSelectedYear} />
                )}
```

packages/lesmis/src/views/DashboardView.js
```diff
@@ -439,23 +479,30 @@ export const DashboardView = React.memo(() => {
           <VitalsView vitals={vitals} />
         </Container>
       </Wrapper>
-      {dropdownOptions.map(({ value, Body, Component }) => (
+      {dropdownOptions.map(({ value, Body, Component, useYearSelector, ComponentProps }) => (
         <TabPanel key={value} isSelected={value === selectedOption} Panel={React.Fragment}>
           <Component
             entityCode={entityCode}
+            year={useYearSelector && selectedYear}
```
tcaiger, line 486:
```suggestion
            year={selectedYear}
```
Un less there is a reason to not just pass through the year and let the component decide whether it is used or not?

packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -120,12 +118,20 @@ const useStickyBar = () => {
...
+  entityCode,
+  TabBarLeftSection,
+  year,
+  filterSubDashboards,
+}) => {
   const [selectedDashboard, setSelectedDashboard] = useUrlSearchParam('subDashboard');
   const { data, isLoading, isError, error } = useDashboardData(entityCode);
   const { scrollToTop, topRef, isScrolledPastTop, onLoadTabBar } = useStickyBar();
-  const activeDashboard = useDefaultDashboardTab(selectedDashboard, data);
+  const subDashboards = useMemo(() => data && data.filter(filterSubDashboards), [
```
tcaiger, line 130:
```suggestion
const subDashboards = useMemo(() => data?.filter(filterSubDashboards), [
```
Minor but I think this should be fine to do...

packages/lesmis/src/views/DashboardView.js
edmofro: Yeah the component doesn't know whether it should use a year selector, because I lifted the year selector out of the component entirely

packages/lesmis/src/views/DashboardReportTabView.js
edmofro: Oh good call! Love this syntax but still getting in the habit

tcaiger (approved):
Thanks for picking this up. Fundamentally I think it's all good so I've pre-approved based on the assumption that you can make those syntax tweaks. I think there is also a bit of component re-factoring that needs to be done but I will pick that up in the other tickets that you raised.

packages/lesmis/src/views/DashboardView.js
edmofro: This is surprising but handy! I didn't realise they had special handling to treat `false` unlike other values during JSX rendering 
image

tcaiger: https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

edmofro: Yeah this bit is what had confused me from that doc
image

PR #2715, opened 2021-06-24
Reset pg-mv-fast-refresh version to latest version

rohan-bes:


edmofro (approved):

edmofro:
Actually @rohan-bes this should probably be against the release branch, unless there's nothing new in dev since freeze

rohan-bes:
>unless there's nothing new in dev since freeze

Looks like there's nothing new on dev, so should be okay.... Happy to merge to release branch but does merging to release branch cause any issues when merging master back to dev?

edmofro:
It doesn't, but it is still easier to go dev then release (avoids master back to dev). Go through dev :+1:

PR #2696, opened 2021-06-22
LESMIS phase one priority visuals

edmofro:
### This is now the base branch for testing LESMIS phase 1 priority visuals 

Old description of this card:

> ### Issue #:
> 
> Visuals from https://github.com/beyondessential/tupaia-backlog/issues/2963
> 1. Schools by level of education at country, province, district level as per https://github.com/beyondessential/tupaia-backlog/issues/2963#issuecomment-865518358 ([screenshot](https://github.com/beyondessential/tupaia-backlog/issues/2963#issuecomment-865578312))
> 2. Schools by district at province level only as per https://github.com/beyondessential/tupaia-backlog/issues/2963#issuecomment-865583217 ([screenshot](https://github.com/beyondessential/tupaia-backlog/issues/2963#issuecomment-866385104))
> 
> Added a couple of very minor features:
> - `translate` function in select in https://github.com/beyondessential/tupaia/pull/2696/commits/2a7c7c47ab034692624112c57feaf203e29b74db
> - ~~custom sort order of stacked bar series' in https://github.com/beyondessential/tupaia/pull/2696/commits/09e63bc44284dba7d1435ed618737f452057e18b~~
> 
> Other than that it's just migrations

PR #2718, opened 2021-06-24
2968: Add children over age for grade

edmofro:
Same as https://github.com/beyondessential/tupaia/pull/2703, but added under international SDGs instead of the district profile

(reverted earlier PR last night so that the feature deployment wouldn't have that migration on it after it cloned at midnight, see https://github.com/beyondessential/tupaia/pull/2696/commits/cfd8fce2d0da02e595a9b9da9b8c44cdab283e7b)

### Issue #:

https://github.com/beyondessential/tupaia-backlog/issues/2968

District level only for now, as per https://github.com/beyondessential/tupaia-backlog/issues/2968#issuecomment-866473967

Screenshot https://github.com/beyondessential/tupaia-backlog/issues/2968#issuecomment-866487801


EMcQ-BES (approved):
lgtm

PR #2720, opened 2021-06-24
Use 'numeral' rather than truncate function with floating point math issues

edmofro:
Fixes https://github.com/beyondessential/tupaia-backlog/issues/2973#issuecomment-867981984


nick-beyondessential (approved):
LGTM

PR #2721, opened 2021-06-25
add support for different legend positions

tcaiger:
### Issue #: [2156](https://github.com/beyondessential/tupaia-backlog/issues/2156) Add support for different legend positions

I have based the position and styling on a new prop called legendPosition. There are other options such as:
- using the theme type (light vs dark)
- adding a project prop (tupaia vs lesmis)
- adding a chartLayout or chartConfig prop

In the end I went with the legendPosition because it is more representative than the first 2 options and simpler than the third. It would be easy to refactor later to add something like chartLayout if needed.

### Changes:
- Add position and style legend based on legendPosition prop
- Added a tooltip to the legend buttons
---

### Screenshots:
see issue

### DIFF
--- a/packages/ui-components/src/components/Chart/Legend.js
+++ b/packages/ui-components/src/components/Chart/Legend.js
@@ -6,20 +6,17 @@
 import React from 'react';
 import styled from 'styled-components';
 import MuiButton from '@material-ui/core/Button';
+import Tooltip from '@material-ui/core/Tooltip';
 
 const LegendContainer = styled.div`
   display: flex;
   flex-wrap: wrap;
-  padding-bottom: 2rem;
+  justify-content: ${props => (props.position === 'bottom' ? 'center' : 'flex-start')};
+  padding: ${props => (props.position === 'bottom' ? '1rem 0 0 3.5rem' : '0 0 2rem 0')};
 `;
 
 const PieLegendContainer = styled(LegendContainer)`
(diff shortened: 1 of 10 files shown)


packages/ui-components/src/components/Chart/Legend.js
```diff
@@ -118,8 +120,12 @@ export const getCartesianLegend = ({ chartConfig, onClick, getIsActiveKey, isExp
           isExporting={isExporting}
           style={{ textDecoration: getIsActiveKey(value) ? '' : 'line-through' }}
         >
-          <Box style={{ background: color }} />
-          <Text>{getDisplayValue(chartConfig, value)}</Text>
+          <Tooltip title="Click to filter data" placement="top" arrow>
```
edmofro, line 123: Nicely snuck in!

edmofro: Should we have the same on the pie legend?

edmofro (approved):
lgtm, nice solution imo!

packages/ui-components/src/components/Chart/Legend.js
tcaiger: No I don't think so. Clicking the pie legend doesn't filter the data (neither on existing and new version of charts). I guess that was never added because it wouldn't really work on pie charts.

PR #2723, opened 2021-06-25
3008: Single Entity Api Connection for Microservices

rohan-bes:

The aim here is to introduce a single instance of a connection class to the entity-api which allows for clients to call the entity-server via a programming interface rather than needing to be aware of the appropriate urls/routes.

I've only added the entity-api to microservices for now, because I wanted to lower the amount of code changes and regression risk as part of this one PR. The orchestration servers should however move over to using their own instances in a similar manner soon enough.

**Note:**
Originally this ticket/branch was created to handle caching of entity-server, however we've decided to try using cloudfront for that instead. Ignore references to caching in the branch name plz 😝 

### Changes:
A few core changes were made as part of this PR:
- Added an `entityApi` interface class that exposes an API for the actual web api of the entity-server itself. This allows clients to call explicit functions for their data rather than passing in urls (ie. `entityApi.getDescendantsOfEntity()` vs `api.get('/hierarchy/${entityCode}/descendants')`)
- Moved report-server over to using server-boilerplate for its construction and route definition.
- Since the server-boilerplate package now needs some types exported from the entity-server package, I've moved all the 'server startup' logic to a file called 'main.ts' so that the type exports can be done in 'index.ts'. As such a the deployment script had to be changed a little bit to suppport this.


full-testable-package.Dockerfile
```diff
@@ -76,6 +76,7 @@ COPY packages/data-broker/. ./packages/data-broker
 COPY packages/database/. ./packages/database
 COPY packages/devops/. ./packages/devops
 COPY packages/dhis-api/. ./packages/dhis-api
+COPY packages/entity-server/. ./packages/entity-server
```
rohan-bes, line 79: Required as currently `server-boilerplate` depends on `entity-server`, so now `entity-server` is an `internalDependency`

packages/aggregator/src/connections/DataSourceEntityProvider.js
```diff
@@ -0,0 +1,52 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ *
+ */
+
+export class DataSourceEntityProvider {
```
rohan-bes, line 7: Wasn't entirely sure what to name this class, but wanted to give it a sort of 'function oriented' name. Rather that it saying 'I'm basically a connection to the entity-server'  I wanted it to say 'here's the data I can provide to you'

packages/aggregator/src/connections/DataSourceEntityProvider.js
```diff
@@ -0,0 +1,52 @@
...
+    dataSourceEntityType,
+    dataSourceEntityFilter,
+  ) {
+    const response = await this.entityApi.getRelationshipsOfEntitiesByDescendant(
+      hierarchyName,
+      entityCodes,
+      { field: 'code' },
+      { filter: { type: this.parseEntityType(aggregationEntityType) } },
+      { filter: { type: this.parseEntityType(dataSourceEntityType), ...dataSourceEntityFilter } },
+    );
```
rohan-bes, line 44: Gotta say, I am so happy with how clean this interface is 😄 Reads so nicely to me!

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -1,58 +0,0 @@
-/*
```
rohan-bes, line 1: Renamed to `DataSourceEntityProvider`

packages/entity-server/src/api/EntityApi.ts
```diff
@@ -0,0 +1,428 @@
...
+type EntityApiQueryOptions = {
+  field?: FlattableEntityFieldName;
+  fields?: ExtendedEntityFieldName[];
+  filter?: EntityFilterObject;
+};
+type RelationshipsSubQueryOptions = Omit<EntityApiQueryOptions, 'fields'>;
+
+type EntityApiResponse<T extends ExtendedEntityFieldName[]> = Required<
+  Pick<EntityResponseObject, T[number]>
+>;
```
rohan-bes, line 39: This is another one of my terrifying nested generic types here, but it's a really neat one. The `Pick<EntityResponseObject, T[number]>` part basically says that, given an array of entity field names, the response will be an object that contains each of those fields, and their type will be correct as well. eg.
```
EntityApiResponse<['code', 'childCodes']> = { code: string, childCodes: string[] }
```

packages/entity-server/src/api/EntityApi.ts
```diff
@@ -0,0 +1,428 @@
...
+  ): Promise<EntityApiResponse<T>>;
+  public async getEntity<T extends ExtendedEntityFieldName[]>(
+    hierarchyName: string,
+    entityCode: string,
+    queryOptions?: {
+      field?: FlattableEntityFieldName;
+      fields?: T;
+      filter?: EntityFilterObject;
+    },
+  ): Promise<FlattenedEntity | Pick<EntityResponseObject, T[number]>> {
```
rohan-bes, line 150: The hideous typescript method overloading syntax strikes again! Kinda torn on this since it's so ugly, yet on the other hand polymorphism is so fantastic... I'm personally willing to bear the cost?

packages/entity-server/src/api/EntityApi.ts
```diff
@@ -0,0 +1,428 @@
...
+      {
+        ...this.stringifyQueryParameters(queryOptions),
+        ...this.stringifyRelationshipsSubQueryParameters(ancestorQueryOptions, 'ancestor'),
+        ...this.stringifyRelationshipsSubQueryParameters(descendantQueryOptions, 'descendant'),
+        groupBy: 'ancestor',
+      },
+    );
+  }
+
+  public async getRelationshipsOfEntityByDescendant(
```
rohan-bes, line 367: Gave different method names to each of the different `groupBy` options, but am considering that perhaps it would be best to use method overloading again...

packages/entity-server/src/index.ts
```diff
@@ -3,32 +3,4 @@
...
-http.createServer(app).listen(port);
-winston.info(`Running on port ${port}`);
-
-/**
- * Notify PM2 that we are ready
- * */
-if (process.send) {
-  process.send('ready');
-}
+export { EntityApi } from './api';
```
rohan-bes, line 6: See the PR description to why all these shenanigans are happening

packages/entity-server/src/routes/hierarchy/MultiEntityRoute.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Route } from '@tupaia/server-boilerplate';
+import { formatEntitiesForResponse } from './format';
+import { MultiEntityRequest } from './types';
+
+export class MultiEntityRoute extends Route<MultiEntityRequest> {
+  async buildResponse() {
+    const { entities, field, fields } = this.req.ctx;
+    return formatEntitiesForResponse(this.req.models, this.req.ctx, entities, field || fields);
+  }
+}
```
rohan-bes, line 15: Surprised we didn't already have this route! It's needed by the report-server to check if the user has access to the entities they've requested data for...

packages/report-server/examples.http
```diff
@@ -12,7 +12,7 @@ content-type: {{contentType}}
 Authorization: {{authorization}}
 
 ### Fetch Report With Test Config
-POST {{baseUrl}}/testFetchReport/xx?organisationUnitCodes=TO&hierarchy=psss&period=2020 HTTP/1.1
+POST {{baseUrl}}/testReport?organisationUnitCodes=TO&hierarchy=psss&period=2020 HTTP/1.1
```
rohan-bes, line 15: Slight url renaming here, and I'm glad to be done away with the need to specify a dummy report code 👍

packages/report-server/src/reportBuilder/reportBuilder.ts
```diff
@@ -4,31 +4,33 @@
...
 }
 
 export class ReportBuilder {
-  config?: ReportConfig;
+  config?: Record<string, unknown>;
 
   testData?: Row[];
 
-  setConfig = (config: ReportConfig) => {
+  setConfig = (config: Record<string, unknown>) => {
```
rohan-bes, line 23: Changing this to `Record<string, unknown>` is valid as each builder runs validation on the config anyway, and the config at this point may truly be unknown

packages/report-server/src/types.ts
```diff
@@ -3,51 +3,21 @@
...
-import { Request } from 'express';
-import { ParamsDictionary, Query } from 'express-serve-static-core';
-import { AccessPolicy } from '@tupaia/access-policy';
-import { TupaiaDatabase, ModelRegistry } from '@tupaia/database';
+import { ModelRegistry } from '@tupaia/database';
+import { ReportModel } from './models';
 
-export interface FetchReportFilter {
+export interface ReportServerModelRegistry extends ModelRegistry {
+  readonly report: ReportModel;
```
rohan-bes, line 10: Super happy that now the report-server only has a dependency on the `Report` model! 👍

packages/server-boilerplate/src/connections/CachedOutboundConnection.ts
```diff
@@ -0,0 +1,81 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { CacheContainer } from 'node-ts-cache';
+import { MemoryStorage } from 'node-ts-cache-storage-memory';
```
rohan-bes, line 7: Decided to go with an external library for the caching mechanism? Overkill perhaps? 

Originally I wanted to use the js decorators that this library provided for a really simple interface, but they proved a tad awkward when being used in a class like this, and it seems we'd have to play with our package.json settings anyway to use decorations since they're still experimental...

packages/server-boilerplate/src/microService/api/ApiBuilder.ts
```diff
@@ -41,13 +44,23 @@ export class ApiBuilder {
...
       req.models = this.models;
 
-      const context = {}; // context is shared between request and response
+      const microServiceAuthHandler = {
+        getAuthHeader: async () => req.headers.authorization || '',
+      };
+      const context: RequestContext = {
+        microServices: {
+          entityApi: new EntityApi(cachedEntityServerConnection, microServiceAuthHandler),
+        },
```
rohan-bes, line 62: We construct a single `cachedEntityServerConnection` for the whole server and then just wrap it in an `EntityApi` per request... Might play around a bit more with the composition of these classes as I think I can do a better job here...

packages/server-boilerplate/src/models/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  {
+    [V in K]: { [field in keyof T[V] & string as `${V}${J}${field}`]: T[V][field] };
+  }[K]
+>;
+
+export type PartialOrArray<T> = {
+  [field in keyof T]?: T[field] extends Record<string, unknown>
+    ? PartialOrArray<T[field]>
+    : T[field] | T[field][];
+};
```
rohan-bes, line 35: Really we ought to have some kind of `ts-utils` package for all the utility types we've been fleshing out...

Inside server-boilerplate is okay for now I guess since all of our ts packages depend on it...

packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -38,7 +38,7 @@ for PACKAGE in ${PACKAGES[@]}; do
         # It's a server, start the pm2 process
         echo "Starting ${PACKAGE}"
         yarn build
-        pm2 start --name $PACKAGE dist --wait-ready --listen-timeout 15000 --time
+        pm2 start --name $PACKAGE 'yarn start' --wait-ready --listen-timeout 15000 --time
```
rohan-bes, line 41: May need to do some testing on this to ensure it performs just as well as it previously has (ie. automatic and manual restart work fine, memory usage consistent, etc.)

Made the change because entity-server no longer has its main file be simply `index.js` under the dist directory, so the previous start script wasn't launching the server. Generally I prefer this method as it allows us to be a bit more custom about how various servers are launched (ie. we can specify whatever args/pre-start steps we want in package.json) but gotta make sure that pm2 plays nicely with that 😄

packages/entity-server/package.json
```diff
@@ -12,14 +12,14 @@
     "url": "git+https://github.com/beyondessential/tupaia"
   },
   "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
-  "main": "dist/index.js",
+  "main": "dist/main.js",
```
rohan-bes, line 15: As far as I can tell specifying the main file here doesn't seem to make a difference... might back this change out...

packages/entity-server/src/index.ts
```diff
@@ -3,32 +3,36 @@
...
-const port = process.env.PORT || 8050;
-http.createServer(app).listen(port);
-winston.info(`Running on port ${port}`);
-
-/**
- * Notify PM2 that we are ready
- * */
-if (process.send) {
-  process.send('ready');
-}
```
rohan-bes, line 34: Moved to `main.ts`

packages/entity-server/src/index.ts
```diff
@@ -3,32 +3,36 @@
...
+  GroupByAncestorRelationshipsResponseBody,
+  GroupByDescendantRelationshipsResponseBody,
+} from './routes/hierarchy/relationships';
+
+export {
+  FlattableEntityFieldName,
+  ExtendedEntityFieldName,
+  EntityResponseObject,
+  FlattenedEntity,
+} from './routes/hierarchy/types';
```
rohan-bes, line 38: Bit sad to have to export so many internal types from the entity-server, but overall it seemed the cleanest approach. Defining the EntityApi within the entity-server was just not cutting it and re-defining these types would be ugly as well...

packages/server-boilerplate/src/connections/ApiConnectionBuilder.ts
```diff
@@ -0,0 +1,29 @@
...
+export class ApiConnectionBuilder {
+  private authHandler?: AuthHandler;
+
+  public handleAuthWith(authHandler: AuthHandler) {
+    this.authHandler = authHandler;
+    return this;
+  }
+
+  public buildAs<Wrapper extends MicroserviceApi>(
+    WrapperClazz: new (apiConnection: ApiConnection) => Wrapper,
```
rohan-bes, line 18: Do we like `Clazz`? Very java of me I know 😝

packages/server-boilerplate/src/connections/ApiConnectionBuilder.ts
```diff
@@ -0,0 +1,29 @@
...
+  ) {
+    if (!this.authHandler) {
+      throw new Error('Must specify an authHandler when building an ApiConnection');
+    }
+
+    const apiConnection = new ApiConnection(this.authHandler);
+    const wrappedApiConnection = new WrapperClazz(apiConnection);
+    apiConnection.baseUrl = wrappedApiConnection.baseUrl;
+    return wrappedApiConnection;
```
rohan-bes, line 27: This little chunk here is super hacky, and as soon as we get a chance to move all ApiConnections over to using this builder pattern I'd love to change it... I feel ideally there'd be a static baseUrl property/function on all wrapper classes that we could use to instantiate the ApiConnection instance with?

packages/server-boilerplate/src/connections/EntityApi.ts
```diff
@@ -0,0 +1,366 @@
...
+    queryOptions?: {
+      field?: FlattableEntityFieldName;
+      fields?: T;
+      filter?: EntityFilterObject;
+    },
+  ) {
+    return this.connection.get(`hierarchy/${hierarchyName}/${entityCode}`, {
+      ...this.stringifyQueryParameters(queryOptions),
+    });
+  }
```
rohan-bes, line 138: Typescript's horrendous method overriding syntax rears its ugly head again 😱 

However, I still feel it's worth it? Polymorphism is a hell of a drug and I don't wanna give it up 😝

packages/aggregator/src/connections/DataSourceEntityProvider.js
```diff
@@ -0,0 +1,52 @@
...
+    hierarchyName,
+    entityCodes,
+    dataSourceEntityType,
+    dataSourceEntityFilter,
+  ) {
+    return this.entityApi.getRelativesOfEntities(hierarchyName, entityCodes, {
+      field: 'code',
+      filter: {
+        type: this.parseEntityType(dataSourceEntityType),
+        ...dataSourceEntityFilter,
```
rohan-bes, line 26: Yay! DataSourceEntityFilter is now working 🎉

packages/server-boilerplate/src/microService/api/ApiBuilder.ts
```diff
@@ -47,7 +49,19 @@ export class ApiBuilder {
...
       req.models = this.models;
 
-      const context = {}; // context is shared between request and response
+      const microServiceAuthHandler = {
+        getAuthHeader: async () => req.headers.authorization || '',
+      };
+
+      const entityApi = new ApiConnectionBuilder()
+        .handleAuthWith(microServiceAuthHandler)
+        .buildAs(EntityApi);
```
rohan-bes, line 58: Thoughts on this ApiConnection construction pattern? Another option might be:
```
      const entityApi = new ApiConnectionBuilder(EntityApi)
        .handleAuthWith(microServiceAuthHandler)
        .build();
```

packages/aggregator/src/connections/DataSourceEntityProvider.js
nick-beyondessential: I love a bit of code personification.

rohan-bes:
Closing out as was completed as part of other PRs

PR #2724, opened 2021-06-25
Update migration utilities to have more expected interface

edmofro:
I realised it was quite unexpected to have `findSingleRecord` query using a raw sql string, so here's an update. Safe enough because we haven't deployed any migrations that rely on it to production.


packages/database/src/utilities/migration.js
```diff
@@ -18,14 +18,18 @@ export const insertObject = async (db, table, data, onError) =>
...
 
 export const deleteObject = async (db, table, condition) => {
-  const [key, value] = Object.entries(condition)[0];
+  const where = Object.entries(condition)[0]
+    .map(([key, value]) => `${key} = '${value}'`)
+    .join(' AND ');
   return db.runSql(`
       DELETE FROM "${table}"
-      WHERE ${key} = '${value}'
+      WHERE ${where}
```
edmofro, line 26: Also thought it was unexpected that you can only pass one condition in to `deleteObject`, so updated that as well. Checked, and all past migrations just use one, so this doesn't have any retrospective effect on historical migrations

packages/database/src/utilities/migration.js
```diff
@@ -18,14 +18,18 @@ export const insertObject = async (db, table, data, onError) =>
   db.insert(table, Object.keys(data), Object.values(data), onError);
 
 export const deleteObject = async (db, table, condition) => {
-  const [key, value] = Object.entries(condition)[0];
+  const where = Object.entries(condition)[0]
```
EMcQ-BES, line 21: It looks like you're running the map on the first `[key, value]` array rather than the parent array of `[key, value]` pairs

```suggestion
  const where = Object.entries(condition)
```

packages/database/src/utilities/migration.js
```diff
@@ -36,6 +40,17 @@ export const findSingleRecord = async (db, query) => {
   return results[0];
 };
 
+export const findSingleRecord = async (db, table, condition) => {
+  const where = Object.entries(condition)[0]
```
EMcQ-BES, line 44:
```suggestion
  const where = Object.entries(condition)
```

EMcQ-BES (changes requested):

packages/database/src/utilities/migration.js
edmofro: You're right! Thanks for the detailed 👀

EMcQ-BES (approved):

edmofro:
I obviously haven't tested the iteration I put up for PR @EMcQ-BES so will do that quickly before merging

edmofro:
Works

PR #2731, opened 2021-06-28
Drop obsolete student gender pie graph

EMcQ-BES:

### DIFF
--- a/packages/database/src/migrations/20210627233338-LESMISDropObsoleteDashboardItem-modifies-data.js
+++ b/packages/database/src/migrations/20210627233338-LESMISDropObsoleteDashboardItem-modifies-data.js
@@ -1,0 +1,34 @@
+'use strict';
+
+import { deleteObject } from '../utilities';
+
+var dbm;
+var type;
+var seed;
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210627233338-LESMISDropObsoleteDashboardItem-modifies-data.js
```diff
@@ -0,0 +1,34 @@
...
+
+exports.up = async function (db) {
+  // Drop the old student gender pie chart
+  await deleteObject(db, 'legacy_report', { code: OBSOLETE_REPORT_CODE });
+  await deleteObject(db, 'dashboard_item', { code: OBSOLETE_REPORT_CODE }); // cascades to dashboard_relation
+};
+
+exports.down = function (db) {
+  return null;
+};
```
EMcQ-BES, line 30: I figure if we really wanna bring this back in the future we should take the opportunity to convert it to a new style report anyway

edmofro (approved):
lgtm

PR #2732, opened 2021-06-28
Change some LESMIS reports to round to 0 dp

edmofro:


edmofro:
@tcaiger you might wonder about changing existing migrations - none have been deployed to dev yet so it's the easiest way, and just involves redeploying the test instance

tcaiger (approved):
lgtm! Thanks for clarifying that they are not deployed yet.

PR #2733, opened 2021-06-28
2974 add support for column value types in the table flippa

tcaiger:
### Changes:

- add support for column value types in the table flippa

---

### Screenshots:
See issue above


edmofro (approved):
Nice, very efficient!

I do feel there's an opportunity to map the chart-wide value type into any chart config entries that don't have their own earlier on, so to save doing this check in each downstream location, but this definitely gets us what we need for now

tcaiger:
> Nice, very efficient!
> 
> I do feel there's an opportunity to map the chart-wide value type into any chart config entries that don't have their own earlier on, so to save doing this check in each downstream location, but this definitely gets us what we need for now

That's a good point but probably does open a bit of a can of worms so sounds good to leave it for now :)

PR #2747, opened 2021-06-29
2928 Consolidate lesmis chart table

tcaiger:
### Changes:

- Rename `Report` component to `DashboardReport``
- Consolidate `ChartTable` component and toggle-able tabs into `Chart` component
- Set a max width on layout components
 
---


packages/lesmis/src/components/DashboardReportModal.js
```diff
@@ -161,27 +150,17 @@ export const DashboardReportModal = ({
                   isLoading={isLoading}
                   startDate={startDate}
                   endDate={endDate}
-                  granularity={viewContent?.granularity}
+                  granularity={periodGranularity}
```
tcaiger, line 153: The period granularity as part of the dashboard response and not in the report response so use `periodGranularity` from the dashboard item list which was fetched upstream. This fixes bug [3012](https://github.com/beyondessential/tupaia-backlog/issues/3012)

edmofro (approved):
All makes sense and lgtm, only question is whether that combined chart/table component might be something we want to lift up into `ui-components`

tcaiger:
> All makes sense and lgtm, only question is whether that combined chart/table component might be something we want to lift up into `ui-components`

It's a good question. I thought about it and it's very hard to make the whole thing work well as a completely re-usable component because the toggle button can be in different places in the page. Next to the title, or above the title or potentially at the top left of the modal. We could solve that using render-props or a state provider but I prefer to just leave the components one grain smaller so that they can be composed as needed by the consumer as it's less complex and more re-factorable.

PR #2749, opened 2021-06-29
Update lesmis year selector

tcaiger:
### Changes:

- add loading state to year selector
- don't clear year state when changing dashboard tabs

---

### DIFF
--- a/packages/lesmis/src/views/DashboardView.js
+++ b/packages/lesmis/src/views/DashboardView.js
@@ -163,6 +163,7 @@ const useDefaultDashboardTab = (selectedDashboard = null, options) => {
 };
 
 export const DashboardView = React.memo(() => {
+  const isFetching = useIsFetching('dashboardReport');
   const { entityCode } = useUrlParams();
   const { data: entityData } = useEntityData(entityCode);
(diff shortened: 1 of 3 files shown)


packages/lesmis/src/views/DashboardView.js
```diff
@@ -197,7 +198,11 @@ export const DashboardView = React.memo(() => {
                   }}
                 />
                 {useYearSelector && (
-                  <YearSelector value={selectedYear} onChange={setSelectedYear} />
+                  <YearSelector
+                    value={selectedYear}
+                    onChange={setSelectedYear}
+                    isLoading={!!isFetching}
```
edmofro, line 204: Is the double bang necessary? Would have thought it was already a bool

edmofro (approved):
Nice and simple. Might be good to get this in before the refactors - could you branch it off 2693 instead?

packages/lesmis/src/views/DashboardView.js
tcaiger: yes because the global isFetching hook returns the number of queries that are loading which is an int.

PR #2750, opened 2021-06-29
Add full refresh after deleting lesmis schools

EMcQ-BES:
### Changes:

- Add full refresh to migration utils
- Run that util after deleting lesmis schools


edmofro:
We released without needing this 🎉

PR #2753, opened 2021-06-30
2994 add visuals for psss Fiji

billli0:
### Issue #:
[#2994](https://github.com/beyondessential/tupaia-backlog/issues/2994)

### Changes:
Copy weekly dashboards and daily overlays from psss Palau  

### Screenshots
[comment](https://github.com/beyondessential/tupaia-backlog/issues/2994#issuecomment-871019188)


packages/database/src/migrations/20210629030806-MapoverlaysForPsssFijiFacilityDemo-modifies-data.js
```diff
@@ -0,0 +1,183 @@
...
+
+const radiusMapOverlayGroup = {
+  id: generateId(),
+  name: 'Syndromic Surveillance Weekly Case Numbers (Radius map)',
+  code: 'FJ_PSSS_Syndromic_Surveillance_Radius_Map_Country',
+};
+const heatMapOverlayGroup = {
+  id: generateId(),
+  name: 'Syndromic Surveillance Weekly Case Numbers (Heat map)',
+  code: 'FJ_PSSS_Syndromic_Surveillance_Heat_Map_Country',
```
rohan-bes, line 121: Hmmm bit of a similar thing here, I think we can rename and re-use the map overlays groups from Palau here as well? ie. `PW_PSSS_Syndromic_Surveillance_Radius_Map_Country` -> `PSSS_Syndromic_Surveillance_Radius_Map_Country`?

The correct overlays should only show up in the appropriate country tho, since the overlays are restricted to the countries in their `countryCodes` field, I believe?

rohan-bes (changes requested):
Few changes requested so that we can better make use of existing dashboards/groups and remove country codes from ids to make the PSSS vizes a bit more generic.

Looking good tho!

packages/database/src/migrations/20210629030241-WeeklyCasesDashboardForFijiDemoFacility-modifies-data.js
```diff
@@ -0,0 +1,115 @@
...
+
+exports.up = async function (db) {
+  const dashboardReportIds = [];
+  for (const [diseaseCode, { dataElement, name, color }] of Object.entries(DISEASE_DATA_ELEMENTS)) {
+    const dashboardReportId = getDashboardReportId(diseaseCode);
+    dashboardReportIds.push(dashboardReportId);
+
+    const dashboardReport = getDashboardReport(dashboardReportId, name, dataElement, color);
+    await insertObject(db, 'dashboardReport', dashboardReport);
+  }
```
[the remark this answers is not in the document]
billli0: @rohan-bes Yes we share the same first thought just to copy the existing Palau visuals to Fiji, but they are using different dataElements (see [commit](https://github.com/beyondessential/tupaia/pull/2753/commits/9f6af95aaf94ab9cabe600013d6ea694c210369b)). One is using indicators, the new one is using new dataElements.

I like your idea to remove `FJ` from the id though, in case we will use them in the future👍

packages/database/src/migrations/20210629030806-MapoverlaysForPsssFijiFacilityDemo-modifies-data.js
billli0: Yeah that's right, if I add `FJ` in countryCodes it will show in Fiji.

packages/database/src/migrations/20210629030001-PsssFijiDemoDashgroup-modifies-data.js
```diff
@@ -0,0 +1,35 @@
...
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await insertObject(db, 'dashboardGroup', {
+    organisationLevel: 'Facility',
+    userGroup: 'PSSS Tupaia',
+    organisationUnitCode: 'FJ',
+    dashboardReports: {},
+    name: 'Syndromic Surveillance National Data',
```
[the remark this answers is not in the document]
billli0: Yeah that's a good point... I will raise a comment in the backlog.

packages/database/src/migrations/20210629030241-WeeklyCasesDashboardForFijiDemoFacility-modifies-data.js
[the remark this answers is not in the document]
rohan-bes: Is there a reason we can't use the same indicators in this dashboard as well? If querying at the site level, then the value should be the same right?

packages/database/src/migrations/20210629030806-MapoverlaysForPsssFijiFacilityDemo-modifies-data.js
billli0: @rohan-bes I think we still need to create new map overlays because they are weekly data, and those for `PW` are daily data.

rohan-bes: @billli0 Yup, we will need to create new overlays 👍

packages/database/src/migrations/20210629030241-WeeklyCasesDashboardForFijiDemoFacility-modifies-data.js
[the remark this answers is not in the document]
billli0: No that's right we can use the same visuals for dashboards. We need to update the previous visuals ids across both legacy and new tables.

packages/database/src/migrations/20210629030806-MapoverlaysForPsssFijiFacilityDemo-modifies-data.js
```diff
@@ -0,0 +1,50 @@
...
+      const newId = getNewId(syndrome, map);
+      const originalId = getOriginalId(syndrome, map);
+      await db.runSql(`
+          UPDATE "mapOverlay" 
+          SET "id" = '${newId}'
+          WHERE "id" = '${originalId}';
+
+          UPDATE "mapOverlay" 
+          SET "countryCodes" = "countryCodes" || '{FJ}'
+          WHERE "id" = '${newId}' and "id" not like '%CON%';
```
billli0, line 36: // We don't need to add `CON` weekly case to FIJI

rohan-bes (approved):
Looks totally awesome!

PR #2754, opened 2021-06-30
3077 lesmis heading order change

EMcQ-BES:
### Changes:

- Add a sort_order to the LESMIS dashboards
- Add a filter in lesmis_server to ignore the `LA_COVID` dashboard (might be unnecessary)


packages/lesmis-server/src/routes/DashboardRoute.ts
```diff
@@ -20,9 +20,10 @@ export class DashboardRoute extends Route {
 
   async buildResponse() {
     const { entityCode } = this.req.params;
-    return this.webConfigConnection.fetchDashboard({
+    const response = await this.webConfigConnection.fetchDashboard({
       organisationUnitCode: entityCode,
       projectCode: LESMIS_PROJECT_NAME,
     });
+    return response.filter((dashboard: any) => dashboard.dashboardCode !== 'LA_COVID');
```
edmofro, line 27: Please add a comment with a link to the discussion why

edmofro (changes requested):
Helpful comments in the migration!

Please update the migration name to use the timestamp @chris-bes has used on the others, plus a `-13` or whatever we're up to after that

chris-bes:
Hey @EMcQ-BES , just to maintain the right order, could you rename your migration to `20210629111111-23-XXX`. The last one right now is `20210629111111-22-FixTypoInTitles-modifies-data`

edmofro (approved):
Thanks!

PR #2765, opened 2021-07-02
3128 remove cachebreaker

IgorNadj:


packages/devops/configs/servers.conf
```diff
@@ -462,6 +462,12 @@ server {
   include /etc/nginx/h5bp/basic.conf;
 
   location / {
+
```
IgorNadj, line 465: Copied from https://stackoverflow.com/a/2068407

edmofro (approved):
lgtm :-)

IgorNadj:
@edmofro added CORS header, was going to put it in a separate PR but easier to put it here

edmofro (approved):
Still lgtm

PR #2766, opened 2021-07-04
1304 show raw data on tooltips

billli0:
### Issue #:
[#1304](https://github.com/beyondessential/tupaia-backlog/issues/1304)
[#2732](https://github.com/beyondessential/tupaia-backlog/issues/2732)

### Changes:
We want to show metadata and original data as a list on popup when hover, This PR will:
- convert those mapOverlays to use `report-server` 
-  let `AreaTips` shows all keys and values from report-server as a list
- not apply changes to `ui-componment` since `lesmis` only shows entity label for now, no including values and other functions.  
- use `measureOptions` to get values from each `measureData` to avoid listing other `measureOptionConfigs`.  
- aggregate postcode level data to district level for state level map overlays   

### Screenshots:
[Screenshot](https://github.com/beyondessential/tupaia-backlog/issues/1304#issuecomment-875478745)


packages/web-frontend/src/containers/Map/AreaTooltip.js
```diff
@@ -35,7 +35,8 @@ export class AreaTooltip extends Component {
   }
 
   render() {
-    const { permanent, onMouseOver, onMouseOut, text, sticky } = this.props;
+    const { permanent, onMouseOver, onMouseOut, texts, sticky } = this.props;
+    const textList = texts.map(text => <span>{text}</span>);
```
rohan-bes, line 39: Hmmm I think we may be missing an opporuntity here to bring the `AreaTooltip` component more inline with the `MeasurePopup` component? These two components behave in similar ways, and I think it would be quite beneficial to align some consistency between them 👍 

Measure Popup works by looping through each series (measureOption) and rendering the formatted display value for that series. I think we could do the same here, with perhaps the one proviso that if there's only a single series, we just display it's value and not title. If we then changed the display type of this overlay to 'fractionAndPercentage' then this would result in something like:
```
[ 'South Australia': 10/50 = 20%]
```
Now admittedly, that's not nearly as pretty as what you've got currently, with there being a nice field for 'Respondents' and 'Flu Vaccinated with Fever and Cough'. However, if we are willing to switch those overlays over to report-server, then we can fairly simply add the 'Respondents' and 'Flu Vaccinated with Fever and Cough' as fields in the report, right?

packages/database/src/migrations/20210702020043-ShowRawDataOnTooltipsInAuFlutracking-modifies-data.js
```diff
@@ -0,0 +1,74 @@
...
+  },
+  {
+    id: 'AU_FLUTRACKING_LGA_Tested_Positive_For_Flu',
+    numerator: 'Participants with symptoms tested positive for flu',
+  },
+  {
+    id: 'AU_FLUTRACKING_LGA_Tested_Positive_For_Covid',
+    numerator: 'Participants with symptoms tested positive for covid',
+  },
+];
```
rohan-bes, line 50: While we're at it, probably worth doing the state level overlays too for consistency, eg. `AU_FLUTRACKING_Fever_And_Cough`

rohan-bes (changes requested):
Bit of a redesign suggestion, happy to talk this one through to hear your thoughts on it?

packages/database/src/migrations/20210706044733-SwitchLGALevelMapOverlayToUseReportServer-modifies-data.js
```diff
@@ -0,0 +1,222 @@
...
+        measureBuilder: 'useReportServer',
+        measureBuilderConfig: { reportCode: id, dataSourceType: 'custom' },
+      },
+      { id },
+    );
+  }
+
+  await db.runSql(`
+    UPDATE "mapOverlay"
+    SET "presentationOptions" = "presentationOptions" || '{"notLegacy": true}'::jsonb
```
billli0, line 211: It will be nice to have a `legacy` column for `mapOverlay` to be consistent to `dashboard`.

billli0:
Description updated.

packages/web-frontend/src/containers/Map/AreaTooltip.js
```diff
@@ -34,8 +41,82 @@ export class AreaTooltip extends Component {
...
+        textList.push(<span key={key}>{`${key}: ${data[key]}`}</span>);
+      });
+      return textList;
+    };
+
+    if (hasMeasureValue) {
+      const legacy = !measureOptions[0].notLegacy;
+      const { formattedMeasureData, onlyOneValue } = legacy
+        ? this.buildLegacyFormattedMeasureData(orgUnitMeasureData, measureOptions)
+        : this.buildFormattedMeasureData(rawOrgUnitMeasureData, measureOptions);
```
rohan-bes, line 70: Hmmm wouldn't it be possible for us to loop through the measureOptions here to work out which of the fields in the `orgUnitMeasureData` should be rendered? Rather than needing to pass in both `orgUnitMeasureData` and `rawOrgUnitMeasureData`? I think if this were the case we also would be able to avoid needing a `notLegacy` key right?

rohan-bes (changes requested):
I think we may want to see if it's possible to avoid introducing the `notLegacy` key and needing to pass in the `rawOrgUnitMeasureData` if possible? My suggestion would be to rely on the `measureOptions` to determine which fields we should render within the tooltip 👍

packages/web-frontend/src/containers/Map/AreaTooltip.js
billli0: Yeah I was thinking the same before. My concern is when we add a new config for `orgUnitMeasureData` we need to add those keys to those fields, and also `displayedValueKey` config will indicate another dynamic `key` of another config. Here `orgUnitMeasureData` and `rawOrgUnitMeasureData` can decouple the thing a bit since they are using sightly different logic. 

Eventually I guess we can refactor `orgUnitMeasureData` to be an extract config for measure data instead of adding configs in the list? @rohan-bes what do you think?

packages/web-frontend/src/utils/measures.js
```diff
@@ -320,58 +320,63 @@ export function getMeasureDisplayInfo(measureData, measureOptions, hiddenMeasure
...
-          displayInfo.color = valueInfo.color;
-          break;
-      }
-      if (valueInfo.isHidden) {
-        displayInfo.isHidden = true;
-      }
-    },
-  );
+  measureOptions
+    .filter(({ hideFromMeasure }) => !hideFromMeasure)
```
billli0, line 324: I just add a filter here. Not sure why Github didn't recognise it.

billli0:
merge PR #2823 to make things work.

packages/web-frontend/src/containers/Map/AreaTooltip.js
```diff
@@ -34,8 +41,76 @@ export class AreaTooltip extends Component {
...
+        : this.buildFormattedMeasureData(orgUnitMeasureData, measureOptions);
+
+      if (onlyOneValue) {
+        return renderTextList({ [orgUnitName]: formattedMeasureData });
+      }
+
+      renderTextList(formattedMeasureData, defaultTextList);
+    }
+
+    return defaultTextList;
```
rohan-bes, line 73:
```suggestion
    const defaultTextList = [
      <Heading key={0} hasMeasureValue={hasMeasureValue}>
        {orgUnitName}
      </Heading>,
    ];
    
    if (!hasMeasureValue) {
      return defaultTextList;
    }
   
    const toTextList = (data) => Object.keys(data).map(key => <span key={key}>{`${key}: ${data[key]}`}</span>);
    
    if (measureOptions.length === 1) {
        return toTextList({ [orgUnitName]: getSingleFormattedValue(orgUnitMeasureData, measureOptions)});
    }

    return defaultTextList.concat(toTextList(this.buildFormattedMeasureData(orgUnitMeasureData, measureOptions)));
```

Feel this simplifies things a bit? Also means we can remove `buildLegacyFormattedMeasureData()` 👍

packages/web-frontend/src/containers/Map/AreaTooltip.js
```diff
@@ -34,8 +41,76 @@ export class AreaTooltip extends Component {
...
+          formattedMeasureData[key] = orgUnitMeasureData[key];
+        }
+      });
+    }
+
+    if (Object.values(formattedMeasureData).length === 1) {
+      return {
+        formattedMeasureData: Object.values(formattedMeasureData)[0],
+        onlyOneValue: true,
+      };
```
rohan-bes, line 105: I'm actually kinda on the fence about this behaviour, leaning towards us always showing it as `<measureName>: <value>` even if there's only a single measure value for this particular org unit.

Assuming that there's more than 1 measureOption (which is the case by the time we call this method) then this is a multi-value map overlay. In which I think the user needs to know which value is which.

eg. For Vic there is just a value for `Respondents`, whereas For NSW there's just a value for `% with flu`
I'd lean toward showing:
```
Vic
Respondents: 7
```
and
```
NSW
% with flu: 6%
```

instead of:
```
Vic: 7
```
and
```
NSW: 6%
```

What do ya think?

packages/web-frontend/src/containers/Map/AreaTooltip.js
```diff
@@ -34,8 +41,76 @@ export class AreaTooltip extends Component {
...
+        orgUnitMeasureData,
+        measureOptions,
+      );
+
+      measureOptions.forEach(({ key }) => {
+        if (key !== measureOptions[0].key) {
+          formattedMeasureData[key] = orgUnitMeasureData[key];
+        }
+      });
+    }
```
rohan-bes, line 99: Is there a reason we're not treating the first measure option and all the others the same? Those other options might also have `valueType` fields or something like that right? Could this just be:
```suggestion
      measureOptions.forEach(({ key, name, valueType }) => {
        const metadata = data[`${dataKey}_metadata`] || data[`${data.name}_metadata`] || {};
        formattedMeasureData[name || key] = formatDataValueByType({ value: orgUnitMeasureData[key], metadata }, valueType);
      });
    }
```

rohan-bes (changes requested):
Halfway thru this but have a couple meetings now so may continue on later today!

Looking good, just some code structure things I've got suggestions about!

And lets pair on getting a fix for that displayInfo issue!

packages/web-frontend/src/containers/Map/AreaTooltip.js
billli0: Current system use either only `orgUnit` or `orgUnit:value`. I think yeah there are too much conditions running here (even though just 3 including the list we are adding). We can reduce to two which are:

```
Vic (only orgUnit)
```

```
Vic (orgUnit with list)
Respondents: 7
...
```

packages/web-frontend/src/containers/Map/AreaTooltip.js
billli0: @rohan-bes Nice refactor !👍

packages/web-frontend/src/containers/Map/AreaTooltip.js
billli0: I was thinking it can leave formating work to `report-server` so that we don't need to increase frontend loading time?

rohan-bes: That's a good concern to have, but I don't think it would affect performance at all, and the consistency between measure options will be very nice 👍

packages/web-frontend/src/containers/Map/AreaTooltip.js
rohan-bes: I think I agree with you 👍 but lets have a quick chat to make sure we're on the same page?

packages/web-frontend/src/containers/Map/AreaTooltip.js
billli0: New changes will let report-server support using metadata only with this formate `${key}_metadata`:
```
key1: 0.02,
key1_metadata_numerator: 2,
key1_metadata_denominator: 100,
key2: 0.01,
key2_metadata_numerator: 1,
key2_metadata_denominator: 100,
```

packages/web-config-server/src/apiV1/measureData.js
```diff
@@ -236,13 +236,15 @@ export default class extends DataAggregatingRouteHandler {
...
-    // Config 'measureConfig' only works for report server builder
+    // Config 'measureConfig' only works for report-server
     const { measureConfig, ...mainMeasureOption } = measureOptions[0];
     if (measureConfig) {
       const { [ADD_TO_ALL_KEY]: configForAllKeys, ...restOfConfig } = measureConfig;
-      getOtherMeasureOptionKeys(mainMeasureOption.key).forEach(key => {
-        measureOptions.push({ ...configForAllKeys, key, name: key, ...restOfConfig[key] });
-      });
+      getOtherMeasureOptionKeys(mainMeasureOption.key)
+        .filter(key => !key.includes('_metadata'))
```
billli0, line 244: hacky way to filter keys with `_metadata` while adding `measureOptions` for other report server data.

packages/web-frontend/src/containers/MeasureLegend/MultiLegend.js
```diff
@@ -55,7 +56,10 @@ const getLegend = measureType => {
 };
 
 const MultiLegend = React.memo(({ measureOptions, isMeasureLoading }) => {
-  const displayedLegends = measureOptions.filter(({ type }) => type !== MEASURE_TYPE_RADIUS);
+  const displayedLegends = measureOptions.filter(
+    ({ type, hideFromLegend }) =>
+      ![MEASURE_TYPE_RADIUS, MEASURE_TYPE_POPUP_ONLY].includes(type) && !hideFromLegend,
```
billli0, line 61: hide from legend if type is `popup-only`.

packages/web-config-server/src/apiV1/measureData.js
rohan-bes: Ahhh I see the issue here.... nice hack! Hopefully we can remove this as more overlays move to the report-server 👍

rohan-bes (approved):
Looks really good @billli0! Nice work!

Really happy with how the change to always display measure name in tooltips looks, great call!

So wanna just wait til https://github.com/beyondessential/tupaia/pull/2823 is merged before merging this? Can be tested first is fine 👍

rohan-bes:
@billli0 https://github.com/beyondessential/tupaia/pull/2823 has just been merged! I think you may wanna merge/rebase this branch off dev again to cleanup the diffs so that it's just your changes 👍

billli0:
@rohan-bes Thanks man 👍

billli0:
aggregate sub-district level data to district level for state level map overlays, which finish the ticket #2732.

PR #2771, opened 2021-07-05
3117 table flippa improvements

tcaiger:
### Changes:

This card is used to collect all the bugs/improvements relating to the new "table flippa" functionality

- Remove the first column of the table is there is not a valid name and it is not timeseries data
- Tweak styles
- Create a blacklist of column names and add dataElementCode to it

---


packages/ui-components/src/components/Chart/Table.js
```diff
@@ -89,22 +93,26 @@ const NoData = styled(SmallAlert)`
...
+const COLUMN_BLACKLIST = ['name', 'timestamp', 'dataElementCode'];
+
 const getColumns = rawData => {
   const columns = [];
   rawData.forEach(row => {
     Object.keys(row)
-      .filter(key => key !== 'name' && key !== 'timestamp')
+      .filter(key => !COLUMN_BLACKLIST.includes(key)) // Don't show columns that don't make sense
+      .filter(key => !columns.includes(key)) // de-dupe
+      .filter(key => typeof row[key] !== 'object') // filter out object values such as metadata
```
edmofro, line 104: Rather than blacklisting, let's follow the same logic as the charts themselves to get the columns, which is (from memory):
- Use the keys in `chartConfig`; and if `chartConfig` doesn't exist
- Use `value` as the only column

edmofro (changes requested):
Good improvements but I think we can match the chart logic more closely

edmofro (approved):
LGTM!

tcaiger:
@edmofro - Can I please get a follow up review on the latest commit. I ended up running viewContent through `parseChartConfig` for maximum consistency between charts and tables.

edmofro (approved):
Great call, this means the order of the table will be according to `legendOrder`, which is what users would expect so definitely an improvement!

PR #2772, opened 2021-07-05
Fix login not working now that cacheBreaker removed

IgorNadj:
The issue was that checkAllowedUnauthRoutes logic was wrong, because it considered query string.

Requests for /api/v1/login?cacheBreaker=xyz would return false, and the control flow would fall to the public access policy.

Removing cacheBreaker made checkAllowedUnauthRoutes return true, which caused a higher up if block to execute, which caused an error where a userId was expected but undefined.

### DIFF
--- a/packages/web-config-server/src/apiV1/getUser.js
+++ b/packages/web-config-server/src/apiV1/getUser.js
@@ -1,4 +1,4 @@
 export const getUser = () => (req, res) => {
-  const { userName, email } = req.session.userJson;
+  const { userName, email } = req.userJson;
   res.send({ name: userName, email });
 };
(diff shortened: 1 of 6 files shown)


edmofro (approved):
Looks good, safe minimal change to fix removing the cachebreaker.

Either in this or a separate PR, can we also try removing the `setSession` for the public user? As you pointed out, if we have a cookie even when logged out, the CDN won't act as a very helpful cache

IgorNadj:
@edmofro added session stuff in this PR for simplicity

edmofro (approved):
Nice, looks great! Glad you're keeping it minimal too, as much as I don't like things like the name `userJson` sticking around

PR #2773, opened 2021-07-05
Remove redundant apicache (probable memory leak)

edmofro:
Through heap snapshot comparison I found a likely culprit for our memory leak: apicache wrapping the `organisationUnit` endpoint.

There were a few things wrong here:
- The `ORGANISATION_UNIT_CACHE_PERIOD` environment variable wasn't set up on production, so it was using the default of 1 hour rather than the more sensible 1 minute configured on dev
- We won't need this server level caching at all once we have the cdn in place, so might as well get rid of it
- Most significantly, it wasn't doing anything helpful, but just storing a whole heap of responses in the cache! Because of the `cacheBreaker`, every request was different, so it was caching per load of tupaia.org


tcaiger (approved):
Thanks for the detailed notes in the description. Makes sense to me to get rid of it and the changes look complete to me.

PR #2774, opened 2021-07-05
3016 lesmis chart loading states

tcaiger:
### Changes:

- Remove loading indicator from year selector
- Add background fetching indicator

---

### Screenshots:
see issue

### DIFF
--- a/packages/lesmis/src/components/Chart.js
+++ b/packages/lesmis/src/components/Chart.js
@@ -74,10 +74,9 @@ const Toggle = ({ value, onChange }) => (
 );
 
 // eslint-disable-next-line react/prop-types
-const ChartTable = ({ viewContent, isLoading, isFetching, isError, error, selectedTab }) => {
-  const showIsLoading = isLoading || (!getIsChartData(viewContent) && isFetching);
+const ChartTable = ({ viewContent, isLoading, isError, error, selectedTab }) => {
   return (
-    <FetchLoader isLoading={showIsLoading} isError={isError} error={error}>
+    <FetchLoader isLoading={isLoading} isError={isError} error={error}>
       {selectedTab === TABS.CHART ? (
         <ChartWrapper>
           <ChartComponent viewContent={viewContent} legendPosition="top" />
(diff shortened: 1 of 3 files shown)


packages/lesmis/src/components/Chart.js
```diff
@@ -72,21 +74,24 @@ const Toggle = ({ value, onChange }) => (
...
-      </ChartWrapper>
-    ) : (
-      <Wrapper>
-        <Table viewContent={viewContent} />
-      </Wrapper>
-    )}
-  </FetchLoader>
-);
+const ChartTable = ({ viewContent, isLoading, isFetching, isError, error, selectedTab }) => {
+  const showIsLoading = isLoading || (!getIsChartData(viewContent) && isFetching);
```
tcaiger, line 78: Also show the full loader if there is no data and we are fetching new data in the backgrond

edmofro: Nice touch, so we keep the animations is there is already data but show a more obvious loader when there is no data 👌

packages/lesmis/src/components/Chart.js
```diff
@@ -111,13 +116,19 @@ export const Chart = ({ name, viewContent, isLoading, isError, error, isEnlarged
   ) : (
     <>
       <Header>
-        <Title>{name}</Title>
+        <FlexStart>
+          <Title>{name}</Title>
+          {isFetching && !isLoading && getIsChartData(viewContent) && (
```
edmofro, line 121: Seems the same slightly complicated decision is being made in two places, so could be better to pull out a variable `isFullLoading`* within this component and then just pass that value through as a prop to the `ChartTable` as `isLoading`, and here use
```suggestion
          {isFetching && !isFullLoading && (
```
*could do with a better name

edmofro (changes requested):
One request, happy for push back. Visual placement looks great :-)

packages/lesmis/src/components/Chart.js
tcaiger: I agree it does seem like they could be combined but I'm not quite sure how to combine the two decisions as the logic for them is actually different. Maybe I'm not seeing a way of doing it?
eg.
`isFetching && !isLoading && getIsChartData(viewContent)` vs `isLoading || (!getIsChartData(viewContent) && isFetching)`

edmofro: Would this work? It doesn't combine them all into one variable, which I also can't see a way to do, but pulls the logic to all be in one place https://github.com/beyondessential/tupaia/pull/2779 (and reuses the tricky decision of "fetching but from no data state, so mark as a full load")

edmofro (approved):

PR #2779, opened 2021-07-06
Move all loading logic to one spot

edmofro:


tcaiger (approved):
Yep that is much better - thanks!

PR #2781, opened 2021-07-06
3138: Update e2e tests new dashboard structure

chris-bes:
### Changes:

- Update e2e tests to fit new dashboard structure

### DIFF
--- a/packages/web-frontend/cypress/config/dashboardReportUrls/covidau.json
+++ b/packages/web-frontend/cypress/config/dashboardReportUrls/covidau.json
@@ -1,6 +1,6 @@
 [
   {
-    "id": "COVID_New_Cases_By_State",
+    "code": "COVID_New_Cases_By_State",
     "project": "covidau",
     "orgUnit": "AU",
     "dashboardGroup": "COVID-19",
@@ -8,55 +8,55 @@
     "endDate": "2020-06-13"
   },
   {
-    "id": "COVID_New_Cases_By_Day",
+    "code": "COVID_New_Cases_By_Day",
     "project": "covidau",
     "orgUnit": "AU",
     "dashboardGroup": "COVID-19"
   },
   {
-    "id": "COVID_Total_Cases_By_State",
+    "code": "COVID_Total_Cases_By_State",
     "project": "covidau",
     "orgUnit": "AU",
     "dashboardGroup": "COVID-19"
   },
   {
-    "id": "COVID_AU_Total_Cases_Each_State_Per_Day",
+    "code": "COVID_AU_Total_Cases_Each_State_Per_Day",
     "project": "covidau",
     "orgUnit": "AU",
     "dashboardGroup": "COVID-19"
   },
   {
-    "id": "COVID_Link_To_Sources",
+    "code": "COVID_Link_To_Sources",
     "project": "covidau",
     "orgUnit": "AU",
     "dashboardGroup": "COVID-19"
   },
   {
-    "id": "COVID_Compose_Cumulative_Deaths_Vs_Cases",
+    "code": "COVID_Compose_Cumulative_Deaths_Vs_Cases",
     "project": "covidau",
     "orgUnit": "AU",
     "dashboardGroup": "COVID-19"
   },
   {
-    "id": "COVID_Total_Cases_By_Type",
+    "code": "COVID_Total_Cases_By_Type",
     "project": "covidau",
     "orgUnit": "AU_Victoria",
     "dashboardGroup": "COVID-19"
   },
   {
-    "id": "COVID_Tests_Per_Capita",
+    "code": "COVID_Tests_Per_Capita",
     "project": "covidau",
     "orgUnit": "AU",
     "dashboardGroup": "COVID-19"
   },
   {
-    "id": "COVID_Total_Tests_Conducted",
+    "code": "COVID_Total_Tests_Conducted",
     "project": "covidau",
     "orgUnit": "AU_Victoria",
     "dashboardGroup": "COVID-19"
   },
   {
-    "id": "COVID_Daily_Cases_By_Type",
+    "code": "COVID_Daily_Cases_By_Type",
     "project": "covidau",
     "orgUnit": "AU_Victoria",
     "dashboardGroup": "COVID-19",
@@ -64,7 +64,7 @@
     "endDate": "2020-07-17"
   },
   {
-    "id": "COVID_Compose_Daily_Deaths_Vs_Cases",
+    "code": "COVID_Compose_Daily_Deaths_Vs_Cases",
     "project": "covidau",
     "orgUnit": "AU",
     "dashboardGroup": "COVID-19"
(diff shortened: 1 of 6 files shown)


packages/web-frontend/cypress/scripts/generateConfig/VisualisationFilter.js
```diff
@@ -70,11 +83,31 @@ class VisualisationFilter {
...
+    return dashboardItems.map(di => {
+      const legacyReport = di.legacy ? legacyReportById[di.report_code] : {};
+
+      if (legacyReport) {
+        return {
+          ...di,
+          data_builder: legacyReport.data_builder,
+          data_builder_config: legacyReport.data_builder_config,
+        };
+      }
```
chris-bes, line 104: Information from the old `dashboardReport` table is now splitted into `dashboard_item` and `legacy_report`. So I feel it's easier for the filter `fetchRecords()` to just return a wrapper object of `dashboard_item` and `legacy_report`. We can do the same for the modern `report` table, but I feel that there's nothing really helpful to filter from `report` table at the moment...

packages/web-frontend/cypress/scripts/generateConfig/VisualisationFilter.js
```diff
@@ -40,20 +48,25 @@ class VisualisationFilter {
       return objectUrls;
     }
 
+    const visualKey = this.getKey();
     const records = await this.fetchRecords(objectUrls);
-    const recordsById = keyBy(records, 'id');
+    // Camel case all the record keys
+    const transformedRecords = records.map(r => mapKeys(r, (_, key) => camel(key)));
```
chris-bes, line 54: This is to ensure any snake cases will be converted to camel cases to match dashboardReportUrls files (eg: default.json). I think at the end, camel case is the standard convention for json keys naming

packages/web-frontend/cypress/config/dashboardReportUrls/covidau.json
```diff
@@ -1,70 +1,70 @@
 [
   {
-    "id": "COVID_New_Cases_By_State",
+    "code": "COVID_New_Cases_By_State",
```
chris-bes, line 3: dashboard_item id is UID, code is what we want to compare

chris-bes:
I just took a stab at this so not very confident. Would be great to have any feedback!

packages/web-frontend/cypress/scripts/generateConfig/VisualisationFilter.js
```diff
@@ -21,6 +22,13 @@ class VisualisationFilter {
     throw new Error('Any subclass of VisualisationFilter must implement the "fetchRecords" method');
   };
 
+  /**
+   * @abstract
+   */
+  getKey = () => {
+    throw new Error('Any subclass of VisualisationFilter must implement the "getKey" method');
+  };
+
```
kael89, line 31: Thanks for adding this method! I would like your opinion on the pattern used - I am still undecided about the best way to handle subclass properties expected by a parent class in JS. I guess that the "safer"/"more proper" OOP approach would be to define abstract getters for them in the parent class (the pattern used above).

However, in the JS world perhaps just defining those properties in the parent class with a comment (eg [DatabaseType.js#L8](https://github.com/beyondessential/tupaia/blob/dev/packages/database/src/DatabaseType.js#L8)) would be enough. Even more so for a simple case like here, with all classes defined in the same file!

If we want to make sure that devs will not forget to override those, we could use [Node assertions](https://nodejs.org/api/assert.html#assert_assert_notequal_actual_expected_message) - we haven't use those much in our codebase but they seem cool!

```js
import assert from 'assert';

class VisualisationFilter {
  // Must override in subclasses
  key = null;

  constructor() {
    assert.notEqual(this.key, null);
  }
}
```

Let me know if you have an opinion on this - they more I look into it, the more keen I am on refactoring it 😆

packages/web-frontend/cypress/scripts/generateConfig/VisualisationFilter.js
kael89: Great call, kudos for paying attention to the details!

packages/web-frontend/cypress/scripts/generateConfig/VisualisationFilter.js
kael89: Good call Chris! Moreover - with the new way of running the tests (against a deployed instance, not localhost) unfortunately most of the time we won't be able to (accurately) use the filters, since they will be using the DB on the host machine, not on the testable instance. For CI/CD we don't even have a DB with actual data, so using filters there would be impossible at the moment.

Which means that all the fancy filtering functionality that was added in the last PR will most probably be unused, at least for a while 😆 

Still, thanks for taking the time to update this!

packages/web-frontend/cypress/scripts/generateConfig/VisualisationFilter.js
```diff
@@ -70,11 +83,31 @@ class VisualisationFilter {
...
-      drillDownLevel: null, // Drill down reports are not currently supported in e2e tests
+  fetchRecords = async objectUrls => {
+    const dashboardItems = await this.db.find('dashboard_item', {
+      code: objectUrls.map(u => u.code),
+    });
+    const legacyDashboardItems = dashboardItems.filter(di => di.legacy);
+    const legacyReports = await this.db.find('legacy_report', {
+      code: legacyDashboardItems.map(di => di.report_code),
+    });
+    const legacyReportById = keyBy(legacyReports, 'code');
```
kael89, line 94:
```suggestion
    const legacyReportByCode = keyBy(legacyReports, 'code');
```

kael89 (approved):
Great work @chris-bes ! Just one minor renaming requested, preapproving.

Also I'd like your opinion on something so I left a question

kael89:
@chris-bes another thing I forgot - could you also update [web-frontend/cypress/Readme.md](https://github.com/beyondessential/tupaia/blob/b14e365350525a05549befcdb17d1225427f88f8/packages/web-frontend/cypress/README.md)? Probably the only update required would be `id` => `code` for report urls

packages/web-frontend/cypress/scripts/generateConfig/VisualisationFilter.js
chris-bes: That's a good point! I also second the latter! It is simpler and clearer to me because I think expressing `this.key` is always better than `this.getKey()` (if we can, and also avoid the need to assign `getKey()` to a variable). Node assertion is handy too!

kael89: In that case, would you be keen to refactor the code to use the simpler pattern? We should also verify that node assertions in the constructor work as intended 😁

chris-bes: Very happy to! I'll apply the same for other properties as well

kael89:
@chris-bes do you mind if I do the remaining changes and merge this? So that Edwin can review the up-to-date code tomorrow

PR #2727, opened 2021-06-25
2936: Switched to doing a single database fetch to get ancestors/descendants/relatives of multiple entities

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2936:

Just one part of an already reviewed PR: https://github.com/beyondessential/tupaia/pull/2657

PR #2792, opened 2021-07-08
3148: Switched to doing a single database fetch to get ancestors/descendants/relatives of multiple entities

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/3148:
- Didn't actually improve performance all that much, but still worth it imo
- Makes the entity-server route logic more readable which is nice

Exact duplicate of https://github.com/beyondessential/tupaia/pull/2727


edmofro (approved):

rohan-bes:
Merging this as only test failures were intermittent timeout tests in indicators

PR #2793, opened 2021-07-08
Add migration for ECE target drill down visuals

EMcQ-BES:
### Changes:

- This is intended to be the base branch for all the tasks associated with the list visual
- Added a migration for the specific ECE drillDowns mentioned in the card


packages/database/src/migrations/20210705232632-AddECEGoalsDrillDowns-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+        },
+      },
+    ],
+  },
+  transform: [
+    'keyValueByDataElementName',
+    {
+      transform: 'select',
+      "'Current'": `$row.${dataElement}/100`,
+      "'Target'": `${target || undefined}`,
```
EMcQ-BES, line 40: This will need to be changed if we swap these targets to being data elements like mentioned in [this comment](https://github.com/beyondessential/tupaia-backlog/issues/2984#issuecomment-875947356)

packages/database/src/migrations/20210705232632-AddECEGoalsDrillDowns-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+  yName: 'Rate',
+  periodGranularity: 'year',
+  valueType: 'percentage',
+  chartConfig: {
+    Current: {
+      color: '#f44336',
+    },
+    Target: {
+      color: '#4caf50',
+    },
```
EMcQ-BES, line 64: Current: red500
Target: green500

packages/database/src/migrations/20210705232632-AddECEGoalsDrillDowns-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+  },
+};
+
+const CONFIG_5_REPORT = {
+  code: 'LESMIS_enrolment_ece_5_target',
+  reportConfig: generateReport('er_district_ece_5_t', '0.86'),
+  frontEndConfig: { ...FRONT_END_CONFIG, name: 'Enrolment rate of 5 year old students in ECE' },
+};
+
+const addNewDashboardItemAndReport = async (
```
EMcQ-BES, line 90: This doesn't add any `dashboard_relation` entries in this migration, because these drill downs aren't directly attached to any dashboards, they're only accessed via drill down

packages/database/src/migrations/20210705232632-AddECEGoalsDrillDowns-modifies-data.js
edmofro: We now have them in indicators, so time to look at this again I think!

packages/database/src/migrations/20210705232632-AddECEGoalsDrillDowns-modifies-data.js
```diff
@@ -0,0 +1,140 @@
...
+  frontEndConfig: { ...FRONT_END_CONFIG, name: 'Enrolment rate of 0-2 year old students in ECE' },
+};
+
+const CONFIG_3_4_REPORT = {
+  code: 'LESMIS_enrolment_ece_3_4_target',
+  reportConfig: generateReport('er_district_ece_3_4_t'),
+  frontEndConfig: {
+    ...FRONT_END_CONFIG,
+    name: 'Enrolment rate of 3-4 year old students in ECE',
+    chartConfig: { Current: { color: '#f44336' } }, // Remove target from legend
```
edmofro, line 76:
```suggestion
    chartConfig: { Current: { color: '#f44336' } }, // Remove target from legend, no 2025 target for 3 - 4
```

edmofro (changes requested):
All lgtm apart from updating to use the indicators for the target values

packages/database/src/migrations/20210705232632-AddECEGoalsDrillDowns-modifies-data.js
EMcQ-BES: This gets swapped to use indicators in the [other PR](https://github.com/beyondessential/tupaia/pull/2878/files#diff-c2b3997ded812179eabbc50d322c6613a97e97e87025ee47806d28d1c399a074R62)

edmofro: Oh of course, I just reviewed that 🤦

edmofro (approved):

tcaiger:
@EMcQ-BES - If possible I was hoping we could tweak the data for the List visuals slightly. It would be ideal if we could sneak the changes in here rather than making another migration. All the changes could actually be done through that admin panel as well so that is another option.

**1. Add drilldown dashboard items to the parent dashboard**
Can we add new relations for the ece dashboards (LESMIS_enrolment_ece_0_2_target, LESMIS_enrolment_ece_3_4_target, LESMIS_enrolment_ece_5_target) to the  LESMIS_ESSDP_EarlyChildhoodSubSector_Schools dashboard? That makes it much easier to display the drill downs on the frontend. This also matches the new pattern that is established in web-frontend for dropdowns (eg. Dashboard Item code 39)

**2. Update the drilldown list config**
Can we update the list config for the circle statistics with the following properties and colors. These are the colors from figma and we ideally need a border color as well to get it 100% accurate.
Screen Shot 2021-08-04 at 9 16 58 AM

edmofro:
The defaulting is a good idea, yep!

How far off do you get by applying a consistent transformation to the primary colour to get a border colour? It'd be nice to avoid having to specify that each time

Do you need the first done sooner to unblock you? I could jump on it now if you need?

tcaiger:
> The defaulting is a good idea, yep!
> 
> How far off do you get by applying a consistent transformation to the primary colour to get a border colour? It'd be nice to avoid having to specify that each time
> 
> Do you need the first done sooner to unblock you? I could jump on it now if you need?

I'm not blocked as I just added the relations on my local. Fair point about the color transformations. I will add the defaults with the exact colors from the design and then calculate a border color if a custom color is set. @EMcQ-BES - Can we just remove the list config altogether then and use the defaults?

tcaiger:
@edmofro and @EMcQ-BES - sorry another question. How much do we care about styling the keywords of the row bold (TARGET, STRATEGY etc)? There are a couple of options for doing this.

1. Add another field like `title` and wrap that in bold at the start of the sentence.
2. Add html such as `Strong` tags to the sentences and then run them through a html parser and render them on the front end.

What do you think?

edmofro:
I would opt for the second strategy, but as you say it depends how much we care. Will ask Erin

EMcQ-BES:
Just confirming, you want me to
- delete the `listConfig` field
- attach the drill down visuals to the `LESMIS_ESSDP_EarlyChildhoodSubSector_Schools` dashboard

EMcQ-BES:
I have made the adjustments in #2955, just letting codeship run before merging that through

PR #2762, opened 2021-07-01
Fix mobile chart legend display

tcaiger:
### Changes:

- Move handling of mobile legend labels to legend component so that the legend value doesn't display on tooltips on mobile

---

PR #2795, opened 2021-07-08
2759 map overlay table flippa

tcaiger:
### Changes:

- Refactor Chart Table Flippa to use react table
- Add Map Table Flippa to ui components
- Add theme support to ui-components modal
- Add MapTable modal to web-frontend with the map table
- Add tooltip to modal button

---

### Screenshots:
see issue


packages/ui-components/src/components/Chart/Table/StyledTable.js
```diff
@@ -0,0 +1,60 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+import styled from 'styled-components';
+import MuiTable from '@material-ui/core/Table';
+
+const DARK_THEME_BORDER = 'rgb(82, 82, 88)';
+
+export const StyledTable = styled(MuiTable)`
```
tcaiger, line 10: @reviewer This StyledTable is duplicate code of the components/Map/Table/StyledTable.js. I kept them seperate because:
1. `components/Chart` and `components/Map` are bundled in to different bundles so the only way to share them would to put it into the main ui-components bundle
2. The styles are quite specific to the table flippa and I was concerned about creating confusion and maintenance issues having another table in ui-components

In saying that, I'm still not sure about it and happy to be convinced of a way to combine them.

packages/ui-components/src/components/Chart/Table/Table.js
```diff
@@ -0,0 +1,101 @@
+/*
```
tcaiger, line 1: The `Chart/Table/Table.js` and `Map/Table/Table.js`are also very similar so the comment above probably applies here as well.

packages/ui-components/src/components/Chart/Table/useChartTable.js
```diff
@@ -0,0 +1,97 @@
...
+  const chartConfig = parseChartConfig(viewContent);
+
+  if (Object.keys(chartConfig).length === 0) {
+    const valueColumn = {
+      Header: 'Value',
+      accessor: row => getFormattedValue(row[DEFAULT_DATA_KEY], viewContent.valueType),
+    };
+
+    return firstColumn ? [firstColumn, valueColumn] : [valueColumn];
+  }
```
edmofro, line 63: Instead of having this separate conditional block, we could use the same pattern as `CartesianChart`
```suggestion
  const chartDataConfig = Object.keys(chartConfig).length > 0 ? chartConfig : { [DEFAULT_DATA_KEY]: {} };
```
And if `parseChartConfig` is only used in a couple of places, we could then move that defaulting right inside `parseChartConfig`

packages/ui-components/src/components/Map/Table/useMapTable.js
```diff
@@ -0,0 +1,58 @@
...
+      accessor: 'name',
+      // eslint-disable-next-line react/prop-types
+      Cell: ({ value }) => <FirstColumnCell>{String(value)}</FirstColumnCell>,
+    },
+    ...configColumns,
+    { Header: 'Submission Date', accessor: 'submissionDate' },
+  ];
+};
+
+const processData = (serieses, measureData) => {
```
edmofro, line 32: In some places we're using `serieses`, and in others sticking with `measureOptions`. I'm fine with either, but could you convert them all to one of the two?

packages/ui-components/src/components/Map/Table/useMapTable.js
```diff
@@ -0,0 +1,58 @@
...
+      name: row.name,
+      ...columns,
+      submissionDate: row.submissionDate ?? 'No data',
+    };
+  });
+};
+
+export const useMapTable = (measureOptions, measureData) => {
+  const columns = useMemo(() => processColumns(measureOptions), []);
+  const data = useMemo(() => processData(measureOptions, measureData), []);
```
edmofro, line 49: Do these memo's not need the `measureOptions` and `measureData` in their dependencies?

packages/ui-components/src/components/Map/Table/useMapTable.js
```diff
@@ -0,0 +1,58 @@
...
+
+  return [
+    {
+      Header: 'Name',
+      accessor: 'name',
+      // eslint-disable-next-line react/prop-types
+      Cell: ({ value }) => <FirstColumnCell>{String(value)}</FirstColumnCell>,
+    },
+    ...configColumns,
+    { Header: 'Submission Date', accessor: 'submissionDate' },
```
edmofro, line 28:
```suggestion
    { Header: 'Most Recent Data Date', accessor: 'submissionDate' },
```

packages/ui-components/src/components/Map/Table/Table.js
```diff
@@ -0,0 +1,78 @@
...
+          ))}
+        </TableHead>
+        <TableBody {...getTableBodyProps()}>
+          {rows.map(row => {
+            prepareRow(row);
+            return (
+              <TableRow {...row.getRowProps()}>
+                {row.cells.map(({ getCellProps, render }) => {
+                  return <TableCell {...getCellProps()}>{render('Cell')}</TableCell>;
+                })}
```
edmofro, line 49:
```suggestion
                {row.cells.map(({ getCellProps, render }) => 
                  <TableCell {...getCellProps()}>{render('Cell')}</TableCell>;
                )}
```
Just to maintain consistency with the `Chart` version where possible

edmofro (changes requested):
Really nice work getting these two table flippas very closely aligned!

packages/ui-components/src/components/Chart/Table/Table.js
edmofro: Yes this one is more of a concern IMO, it'd be nice if we can find a way to share the code. If you could put an hour into it as a timebox, and if you don't get anywhere, then add a comment to each so that people are at least aware that there might be another place to update if they're e.g. fixing a bug

tcaiger: @edmofro - What are your thoughts on me removing the existing Table components from `ui-components` and putting them in psss. And then making this the Table component of ui-components? Alternatively, we could leave in the old table and call this something different such as: DataTable or ReactTable?

packages/ui-components/src/components/Chart/CartesianChart.js
```diff
@@ -162,18 +154,11 @@ export const CartesianChart = ({ viewContent, isEnlarged, isExporting, legendPos
...
-  const config = Object.keys(chartConfig).length > 0 ? chartConfig : { [DEFAULT_DATA_KEY]: {} };
-
-  const sortedChartConfig = Object.entries(config).sort((a, b) => {
-    return CHART_SORT_ORDER[b[1].chartType] - CHART_SORT_ORDER[a[1].chartType];
-  });
-
   const ChartContainer = CHART_TYPE_TO_CONTAINER[defaultChartType];
-
+  const hasDataSeries = chartConfig && Object.keys(chartConfig).length > 1;
+  const chartDataConfig = hasDataSeries ? chartConfig : { [DEFAULT_DATA_KEY]: {} };
```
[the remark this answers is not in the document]
tcaiger: No that is not intentional - good catch 🙏

packages/web-frontend/src/selectors/utils.js
```diff
@@ -10,7 +10,8 @@
  */
 export const safeGet = (cache, args) => (cache.keySelector(...args) ? cache(...args) : undefined);
 
-export const getOrgUnitFromCountry = (country, code) =>
-  country && code ? country[code] : undefined;
+export const getOrgUnitFromCountry = (country, code) => {
+  return country && code ? country[code] : undefined;
+};
```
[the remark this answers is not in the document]
tcaiger: yep

packages/ui-components/src/components/Dialog.js
```diff
@@ -11,7 +11,11 @@ import CloseIcon from '@material-ui/icons/Close';
 import styled from 'styled-components';
 import { IconButton } from './IconButton';
 
+// todo: add colors to theme
```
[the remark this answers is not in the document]
tcaiger: Actually when I think about it a bit more I think those are not really theme colors anyway so actually make sense to be scoped to the dialog.

packages/ui-components/src/components/Chart/Table/Table.js
edmofro: Hmmmm...well I would say that PSSS tables support a couple of extra things that are quite nice, but this table is probably more what you'd expect from a basic table export. 

Another option is to export them both from ui-components, but this one as `Table` and the PSSS one as something indicative of its enhanced functionality like `ComplexTable`/`NestableTable`/`ExpandingRowTable`.

I'm happy to go with your decision on this, you've been closer to all these different tables than anyone

packages/ui-components/src/components/Chart/Legend.js
```diff
@@ -86,52 +85,28 @@ const Text = styled.span`
   line-height: 1.4;
 `;
 
-const getPieLegendDisplayValue = (chartConfig, value, item, isEnlarged, viewContent) => {
```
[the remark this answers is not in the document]
tcaiger: I did some digging and it looks like what happened is that these changes are from a [hotfix](https://github.com/beyondessential/tupaia/pull/2762) I made that somehow got deleted without getting deployed. That was the week when there was supposed to be a hotfix that then became part of the normal release. Are you happy to leave the changes here?

packages/ui-components/src/components/Chart/Table/useChartTable.js
tcaiger: I'm not quite sure I understand. This processing seems too different to the CartesianChart to use the same code. Can you explain a bit more?

packages/ui-components/src/components/Chart/Legend.js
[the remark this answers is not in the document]
edmofro: Aaah right. It'd be nice to separate them out and merge that stuff to dev immediately so it goes into this next release and we don't have to manage the weirdness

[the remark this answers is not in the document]
edmofro: Actually it looks like it did make it into the release after all https://github.com/beyondessential/tupaia/pull/2790/files#diff-1208b28f565ecdc4b707e34f73eca909c4da7b9878fb0aeab5ba6fbcfb00b261

packages/ui-components/src/components/Chart/Table/useChartTable.js
edmofro: I'll have to work out what I was meaning two weeks ago first! 😅

edmofro: Cool just re-read it and can give you a call if you like?

edmofro: Summary though, is that the processing here is actually able to be handled by the block below if we set up a default "value" column if there's no chartConfig defined. That way we can delete this whole section.

It looks like the lack of any config is meaningful in other places though so unfortunately we can't push the defaulting into `parseChartConfig` without making more widespread changes, so leave that bit of my suggestion out.

packages/ui-components/src/components/Chart/Legend.js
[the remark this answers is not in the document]
tcaiger: ah woops. It was the other way around and it the hotfix change hadn't made it into this branch yet. I just copied across and now the diff on the Legend file is gone.

packages/ui-components/src/components/Chart/Table/useChartTable.js
tcaiger: ok I think that makes sense. I might come back to this in the morning.

tcaiger:
@edmofro - This is ready for re-review. See the last commit for the processColumns refactor.

packages/ui-components/src/components/Chart/Table/Table.js
```diff
@@ -0,0 +1,73 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import MuiTableContainer from '@material-ui/core/TableContainer';
+import { DataTable } from '@tupaia/ui-components';
```
edmofro, line 9: From what I can see, there isn't a `DataTable` export in ui-components (also should we just go through the relative directory rather than the named library, or does that cause issues with the code splitting?)

edmofro (approved):
Looks good!

PR #2803, opened 2021-07-09
2729: Add survey period granularity MEL-92

kael89:
### Issue #:
https://github.com/beyondessential/tupaia-backlog/issues/2729

The changes here are made with `tupaia` surveys in mind, but they should also work for `dhis` surveys. However, the following could/should addressed at some point if we are to use this feature for dhis surveys ("technical debt"):
1. Remove `survey.can_repeat` and use `period_granularity` instead (`undefined` would be used for repeating surveys). This would require all non-program DHIS2 surveys to use a period granularity.
2. Do not re-sync survey responses on `outdated` updates (can add a check in [DhisChangeValidator.getValidSurveyResponseUpdates](https://github.com/beyondessential/tupaia/blob/dev/packages/meditrak-server/src/dhis/DhisChangeValidator.js#L104)). Currently re-syncs will be triggered if any response field is updated, even unrelated ones eg `user_id`, `assessor_name`. We would also need to update the `markAsChanged` functionality to identify whether a response was specifically marked for DHIS2 re-sync - see https://github.com/beyondessential/tupaia/pull/2803#discussion_r666786765
3. Ensure that `outdated` is calculated in exactly the same way in our DB as in DHIS2. To my knowledge, the most recent survey response per day is generally the one returned when fetching analytics from DHIS2, but we would need to confirm that
    * this also supports non daily period granularities
    * the latest `end_time` is used per period (and not eg the latest sync time)
    * (super detail) all other things being equal, prefer responses which are already marked as `outdated`, or the ones with the highest id (this is the logic used in this PR as tiebreaker)

From all the above, we should probably skip 1 for now, and 3 may not be an issue in most cases


packages/meditrak-server/src/database/SurveyResponseChangeHandler/SurveyResponseChangeHandler.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { OutdatedStatusUpdater } from './OutdatedStatusUpdater';
+
+const UPDATE_DEBOUNCE_TIME = 250; // wait 250ms after changes before updating, to avoid double-up
+
+export class SurveyResponseChangeHandler {
```
kael89, line 10: I'd like to get some feedback on this one: while the name of the class is generic, it currently only handles changes specifically related to keeping the `outdated` status up to date. This could change in the future though, by adding more handlers.

There is one more change handler that is currently defined in the [SurveyResponse.js](https://github.com/beyondessential/tupaia/blob/dev/packages/meditrak-server/src/database/models/SurveyResponse.js#L101) model but could (should?) be moved here. I've found the fact that notification handlers may be defined in multiple places confusing in the past.

Do you reckon this is fine as is or should we:
* a) move the notification from the model here, or
* b) leave the notification as is and rename this class to `SurveyResponseOutdatedStatusUpdater` or similar. I started with this approach but then changed it because I didn't like the class name, and I've found that ugly names may be an indication of wrong abstractions

packages/meditrak-server/src/tests/routes/importSurveyResponses/testOutdatedStatusUpdate.js
```diff
@@ -0,0 +1,114 @@
...
+    );
+
+    expect(response.statusCode).to.equal(200);
+
+    await models.database.waitForAllChangeHandlers();
+    // Statuses listed in the same order as the corresponding survey responses in the import spreadsheet
+    await assertOutdatedStatuses('Test_Yearly', [true, true, false, false, false, false]);
+    await assertOutdatedStatuses('Test_Monthly', [true, true, false, false, false, false]);
+    // Responses for non periodic surveys are never "outdated"
+    await assertOutdatedStatuses('Test_No_Granularity', [false, false, false, false, false, false]);
```
kael89, line 112: Just record creation tests here, since updates/deletes are tested in the corresponding `surveyResponse` routes (get/put/delete). Although given the significance of the import process in our ecosystem, it may be worth adding a few more cases here

packages/meditrak-server/src/tests/routes/surveyResponse/testOutdatedStatusUpdate.js
```diff
@@ -0,0 +1,495 @@
...
+const SURVEYS = {
+  [yearlySurveyId]: buildSurvey(yearlySurveyId, 'yearly'),
+  [quarterlySurveyId]: buildSurvey(quarterlySurveyId, 'quarterly'),
+  [monthlySurveyId]: buildSurvey(monthlySurveyId, 'monthly'),
+  [weeklySurveyId]: buildSurvey(weeklySurveyId, 'weekly'),
+  [dailySurveyId]: buildSurvey(dailySurveyId, 'daily'),
+  [nonPeriodicSurveyId]: buildSurvey(nonPeriodicSurveyId),
+};
+
+export const testOutdatedStatusUpdate = app => {
```
kael89, line 42: I chose to add the majority of test cases for this feature in the CUD routes for surveyResponse, instead of doing so in the model level. The latter would cover a bigger spectrum of use cases (eg changes in migrations), but I still felt that testing in the api level is more appropriate.

We could have additional tests in the model level, but I thought that would be over-testing. Plus, one could consider the fact that change handling logic is also available in the model level an implementation detail.

packages/meditrak-server/src/database/SurveyResponseChangeHandler/SurveyResponseChangeHandler.js
```diff
@@ -0,0 +1,76 @@
...
+  constructor(models, updateDebounceTime = UPDATE_DEBOUNCE_TIME) {
+    this.models = models;
+    this.updateDebounceTime = updateDebounceTime;
+    this.outdatedStatusUpdateCandidates = [];
+    this.scheduledUpdateTimeout = null;
+    this.scheduledUpdatePromise = null;
+    this.scheduledUpdatePromiseResolve = null;
+    this.activeUpdatePromise = null;
+    this.changeHandlerCancellers = [];
```
kael89, line 19: I was heavily inspired (read: shamelessly copied) the batching logic currently defined in [AnalyticsRefresher.js](https://github.com/beyondessential/tupaia/blob/dev/packages/database/src/analytics/AnalyticsRefresher.js). This seems to be a useful pattern so perhaps worth abstracting under a shared class at some point!

I didn't feel it was worth the refactor in this PR, but if you feel strongly about it happy to give it a shot

packages/meditrak-server/src/database/models/Survey.js
```diff
@@ -66,7 +82,15 @@ const onChangeUpdateDataGroup = async (
...
     case 'update': {
       const { code, data_source_id: dataSourceId } = newRecord;
-      return models.dataSource.updateById(dataSourceId, { code });
+      await models.dataSource.updateById(dataSourceId, { code });
+
+      if (newRecord.period_granularity !== oldRecord.period_granularity) {
+        // the `outdated` status of survey responses must be re-calculated, since it depends on
+        // the period granularity of the survey they belong to
+        // no need to await for the operation, just get it going
+        models.surveyResponse.markAsChanged({ survey_id: newRecord.id });
```
kael89, line 91: Using `markAsChanged` here is not optimal, because a record could be changed for any reason (see eg [AggregateDataPusher.js#L398](https://github.com/beyondessential/tupaia/blob/dev/packages/meditrak-server/src/dhis/pushers/data/aggregate/AggregateDataPusher.js#L398), where we just want to update the dhis sync queue). A potential improvement would be being able to add custom metadata in the marked change, which could be used by the change handler to decide whether the change is relevant.

edmofro:
Thanks for the notes about tech debt to improve the use of `outdated` with DHIS sync. Some notes below

> 1. Remove `survey.can_repeat` and use `period_granularity` instead (`undefined` would be used for repeating surveys). This would require all non-program DHIS2 surveys to use a period granularity.

Actually I'd like to keep `can_repeat`, but break the implication that `true` means syncing via the events api. The reason I'd like to keep it is that it acts as a front end flag to the `meditrak-app`, indicating that on the submission page there should be a "Submit & Repeat" button for surveys that are often performed in bulk. It's a shame this is tied to event based syncing, because actually it might be nice on some other surveys where e.g. at the end of a week the nurse enters each day's data.

> 2. Do not re-sync survey responses on `outdated` updates (can add a check in [DhisChangeValidator.getValidSurveyResponseUpdates](https://github.com/beyondessential/tupaia/blob/dev/packages/meditrak-server/src/dhis/DhisChangeValidator.js#L104)). Currently re-syncs will be triggered if any response field is updated, even unrelated ones eg `user_id`, `accessor_name`. We would also need to update the `markAsChanged` functionality to identify whether a response was specifically marked for DHIS2 re-sync - see [2729: Add survey period granularity #2803 (comment)](https://github.com/beyondessential/tupaia/pull/2803#discussion_r666786765)

Seems like a good change to make across several fields, not just the `outdated` field. Could you please add a card for this?

> 3. Ensure that `outdated` is calculated in exactly the same way in our DB as in DHIS2. To my knowledge, the most recent survey response per day is generally the one returned when fetching analytics from DHIS2, but we would need to confirm that
>    
>    * this also supports non daily period granularities

This would be good to confirm

>    * the latest `end_time` is used per period (and not eg the latest sync time)

Yep I can confirm (from the code) this is true
```
const laterSamePeriodSurveyResponses = await this.models.surveyResponse.find(
      {
        id: {
          comparator: '!=',
          comparisonValue: surveyResponse.id,
        },
        survey_id: surveyResponse.survey_id,
        entity_id: surveyResponse.entity_id,
        data_time: {
          comparisonType: 'whereBetween',
          args: [periodBounds],
        },
        end_time: {
          comparator: '>',
          comparisonValue: surveyResponse.end_time,
        },
      },
      {
        sort: ['data_time DESC'],
      },
    );
```

>    * (super detail) all other things being equal, prefer responses which are already marked as `outdated`, or the ones with the highest id (this is the logic used in this PR as tiebreaker)

I'm not stressed about the super detail tiebreaking logic - it would be extremely unusual that two survey responses were submitted with the same period and exact same `end_time`, unless through a bulk import, which we can be a bit less paternal about.

packages/database/src/utilities/isMarkedChange.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+export const isMarkedChange = changeDetails => {
```
edmofro, line 6: Clever! And really helpful comment, maybe it should even go above the function

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
```diff
@@ -193,21 +193,20 @@ export async function importSurveys(req, res) {
...
         if (req.query.countryIds) {
-          // Set the countries this survey is available in
           fieldsToForceUpdate.country_ids = getArrayQueryParameter(req.query.countryIds);
         }
         if (surveyGroup) {
-          // Set the survey group this survey is attached to
           fieldsToForceUpdate.survey_group_id = surveyGroup.id;
         }
         if (req.query.permissionGroup) {
-          // A non-default permission group was provided
```
edmofro, line 204: Comments like this seem a bit self-explanatory, but if someone has put them there I don't see any harm in keeping them (the `non-default` term here could give someone easy mental access to what's going on)

packages/meditrak-server/src/database/SurveyResponseChangeHandler/SurveyResponseChangeHandler.js
rohan-bes: I'd lean towards b) however I'd probably also wanna move that change hander from `SurveyResponse` to another class beside here called something like `AnswerChangePublisher`?

As for the class name being ugly, how does `SurveyResponseOutdater` sound? It kinda fits with our existing naming convention of `<noun (often model)><verb>`, ie. `AnalyticsRefresher`, `EntityHierarchyCacher`?

I'm with you that having change notification handlers scattered around the place can be confusing. And I think over time they may grow more similar functionally (sharing debounce logic code and such) for performance reasons.

rohan-bes: While it might seem a bit odd, I think the convention for these 'change handler routine' classes is for them to be defined in the `@tupaia/database` package (see. `AnalyticsRefresher`, `EntityHierarchyCacher`). Even if they are instantiated by the meditrak-server, I think they are meant to be agnostic towards that, and are solely a 'database' concern? 

Happy to get pushback on that though since I do think the pattern is a bit un-intuitive

rohan-bes (commented):
Partway through @kael89 but will submit now since have a few meetings and figured I might take a few passes to get through this one. Also, walk-n-talk on this would be great at some point!

packages/database/src/utilities/isMarkedChange.js
kael89: I'd leave it there since its an implementation detail and not part of the function's contract - in the future, we could calculate this in a different way

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
kael89: I beg to differ but I'm super happy to leave this as is!

packages/database/src/utilities/isMarkedChange.js
edmofro: True! Perhaps a comment above that says 
```suggestion
// indicates whether a change notification was manually triggered, rather than a true database change
export const isMarkedChange = changeDetails => {
```
I feel the name may not be initially obvious so this sort of comment could be useful

packages/data-api/scripts/createAnalyticsMaterializedView.sql
```diff
@@ -29,7 +29,7 @@ declare
       question ON question.id = answer.question_id
     INNER JOIN
       data_source ON data_source.id = question.data_source_id
-    WHERE data_source.service_type = ''tupaia''';
+    WHERE data_source.service_type = ''tupaia'' AND survey_response.outdated IS FALSE';
```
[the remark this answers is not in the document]
kael89: Certainly! Happy to do some performance benchmarks - your analytics experience could be useful there

packages/database/src/utilities/isMarkedChange.js
kael89: Good idea, will do!

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
edmofro: Keen to hear your reasoning - I guess you find they clutter things and you get distracted by them? For me it depends on whether there's anything conveyed by the comment other than "set x" that a junior dev/open source dev/someone new to the codebase could get something out of that they might take longer to understand by purely reading the code. For e.g. I would definitely agree the next comment down is useless (`Set or update the code for this survey`), but find there might be value in the one about `countries this survey is available in` because it says a bit more about what `country_ids` actually does. Agree it wouldn't justify a comment if you were writing the code fresh, but just that we maybe lose a little something by removing them. Extremely minor point that I've already spent too many words on, so very happy either way!

packages/meditrak-server/src/database/SurveyResponseChangeHandler/OutdatedStatusUpdater.js
```diff
@@ -0,0 +1,261 @@
...
+    startDate: moment(datetime).startOf(momentUnit).format(DATETIME_FORMAT),
+    endDate: moment(datetime).endOf(momentUnit).format(DATETIME_FORMAT),
+  };
+  periodRangeCache[periodGranularity][date] = range;
+
+  return range;
+};
+
+const getIdOfMostRecentResponse = surveyResponses => {
+  assert.notStrictEqual(surveyResponses.length, 0);
```
edmofro, line 87: Nice, I haven't seen anyone use assertions using this library before but it certainly makes things tidy

packages/meditrak-server/src/routes/importSurveys/importSurveys.js
kael89: I agree with you that a couple of those comments are useful! I guess maintainers thought that they should follow the existing pattern and added comments that are not particularly useful - that's what I wanted to avoid. This may lead someone to ignore comments that are actually useful - happened to me when I deleted all those comments! 😛 

Let's keep the ones that seem to convey the non-obvious then! 🙌

edmofro: Perfect! Yes haha I agree, we shouldn't require consistency if that introduces noise

packages/meditrak-server/src/database/SurveyResponseChangeHandler/OutdatedStatusUpdater.js
kael89: Yeap I wanted to comment on that! Happy that you like it. The idea is:

If it's a user (end user or dev) related error, throw an error. If it's just a precondition checking, use an assertion. So what this conveys is "the length of `surveyResponses` should really be <> 0 and supposedly the existing code doesn't allow that. But if for whatever reason this is not the case, let the devs know asap"

I find function preconditions and postconditions [Eiffel - Design By Contract](https://www.eiffel.org/doc/eiffel/ET-_Design_by_Contract_%28tm%29%2C_Assertions_and_Exceptions) an interesting mental pattern. Many mode modern languages haven't seem to picked up on that (perhaps it's too verbose?), but checking preconditions with one liner assertions seems like a good balance to me!

kael89:
> Thanks for the notes about tech debt to improve the use of `outdated` with DHIS sync. Some notes below

Thanks for checking those points @edmofro !

1. That makes sense - let's keep that field then!
2. Card: https://github.com/beyondessential/tupaia-backlog/issues/3167
3. Agree that the tiebreaker logic is a detail so we shouldn't bother

I have created https://github.com/beyondessential/tupaia-backlog/issues/3167 to capture 1 and 3. I am not sure whether we'll justify the time to implement it though - keeping things as is (by not using `period_granularity` for DHIS2 surveys) may work just as fine for the foreseeable future.

packages/meditrak-server/src/database/SurveyResponseChangeHandler/SurveyResponseChangeHandler.js
edmofro: Also keen to break out the tricky part, and also happy for that to be a separate PR (though I'm not sure if a raised ticket would stand a good chance of making it into a sprint)

packages/meditrak-server/src/database/SurveyResponseChangeHandler/SurveyResponseChangeHandler.js
edmofro: A few points here:
1. Love the name `SurveyResponseOutdater`, great suggestion @rohan-bes 
2. Agree that it's getting confusing having notification responders in lots of different places. This is starting to become more significant because a) they're a source of performance bottleneck if used naively, and b) when we scale horizontally, we'll need to make sure only one of the instances is performing the side effects, so having many places to turn that switch on isn't great. We can review where change handlers sit more holistically when looking at horizontal scaling.
3. I'm keen to keep the change handler routines in `@tupaia/database` as you suggest @rohan-bes, mainly because we may decide to have a completely separate "change handler" service that responds to notifications and makes side effects (because of the horizontal scaling point above), plus I don't think this is necessarily a `meditrak-server` responsibility anyway (it probably _is_ the most natural fit, but mainly because it's a catch-all)

edmofro (commented):
Since @rohan-bes picked up the review proper I've just given this a conceptual review in spare moments today - will leave the line-by-line stuff to him.

What I see is fantastic! For a reasonably complicated asynchronous, batched operation with several edge cases you've produced something that is super clear, with really helpful comments, well structured code, and thorough testing 🎉

packages/meditrak-server/src/database/SurveyResponseChangeHandler/OutdatedStatusUpdater.js
edmofro: Yeah that's a great rule, love it :-)

packages/meditrak-server/src/database/SurveyResponseChangeHandler/SurveyResponseChangeHandler.js
rohan-bes: I'm thinking along the same lines with you on 3. @edmofro and another concern of mine is that we may need to evolve to a slightly more advanced messaging protocol soon enough, and if so having all the subscribers/publishers in the same place alongside the event definitions might be helpful (think redux actions/reducers).

I say this because I'm a bit worried that we may start seeing some nasty feedback loops between this handler and the `AnalyticsRefresher`, and I think such issues may just become more common over time. 

Ofcourse let's see how it goes in testing with large data submissions before we get too worried, just set Juliana on it for a day 😝

packages/meditrak-server/src/database/SurveyResponseChangeHandler/OutdatedStatusUpdater.js
```diff
@@ -0,0 +1,260 @@
...
+  constructor(models) {
+    this.models = models;
+  }
+
+  /**
+   * Not all changed survey responses need to trigger an outdated status update
+   * This method can be used to filter out changed records that are not worth
+   * considering, without having to look into the DB for more information
+   */
+  static getCandidatesFromChangeDetails = changeDetails => {
```
rohan-bes, line 129: Absolutely love how clever this function is to filter changes and only process the necessary ones! I think we should do something similar for the `AnalyticsRefresher` too!

packages/meditrak-server/src/tests/routes/importSurveyResponses/testOutdatedStatusUpdate.js
rohan-bes: Would also love to see those tests added, but won't push on it 👍

rohan-bes (approved):
Nice work man, really clean, efficient, and elegant code! I think you're aware of the few refactors that've been suggested, so I'll be approving in advance (will take a quick lookover again when you add those changes).

Just a note on testing strategy, I rekon it'll be really important for us to get some confidence about how this runs with large uploads (1000s of responses all updating existing responses). I am a bit concerned about performance...

However, dunno how you feel about this, but I'd probably also be okay with this PR getting merged and integrated with the codebase prior to any tricky performance issues getting resolved, and we can just not enable this feature on any surveys? (Perhaps even hiding the UI option for now)

Regardless, let's see how testing goes and discuss from there!

packages/meditrak-server/src/database/SurveyResponseChangeHandler/OutdatedStatusUpdater.js
kael89: Yeap good call!

packages/meditrak-server/src/database/SurveyResponseChangeHandler/OutdatedStatusUpdater.js
```diff
@@ -0,0 +1,261 @@
...
+  return {
+    surveyId: survey.id,
+    entityId: surveyResponse.entity_id,
+    startDate,
+    endDate,
+  };
+};
+
+const groupResponsesForSurveyByDimensionCombo = (surveyResponses, survey) =>
+  Object.values(groupBy(surveyResponses, sr => JSON.stringify(getDimensionCombo(sr, survey))));
```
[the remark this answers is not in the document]
kael89: In this PR, the object is instantiated in the same way every time: 

https://github.com/beyondessential/tupaia/blob/76fdebf396088af5bbc7af45a7f51429ec6ddc31/packages/meditrak-server/src/database/SurveyResponseChangeHandler/OutdatedStatusUpdater.js#L108

and is stringified in the same way every time (with `JSON.stringify`. So, even if the object key order does not match the insertion order, I would assume the resulting string is deterministic?

By the way, ES6 **does** guarantee the insertion order for non-number keys: https://2ality.com/2015/10/property-traversal-order-es6.html. (Using number keys in a hash is a probably a bad practice)

We don't set a [target](https://babeljs.io/docs/en/babel-preset-env#no-targets) in Babel so we currently transpile to ES5 (default), so I guess that the above doesn't apply in our case. Having said that, it's probably a good idea to change this since our server runs node `v12.18.3`!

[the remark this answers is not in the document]
kael89: (I'm happy to use a 3rd party library btw, that's often the way to go, but I don't think it's required in this case)

kael89:
@rohan-bes add 3 new + 1 merge commit (starting with https://github.com/beyondessential/tupaia/pull/2803/commits/6fdfec47fdf659eab560570b255e400d3a28577a)

I also did some local performance tests yesterday and the results seem satisfactory 😃 I used a 10-question survey with randomised answers, for count starting at 100 responses and doubling up to 6400. I first took measurements using `dev`, then the new functionality with none, daily and monthly granularities.

* The duration for survey response import/deletion + analytics refresh deletion is similar to before (no more than 10-20% up , eg from ~=40 to 45 secs for importing a few thousand responses)
* The duration for the outdated response handling itself seems pretty good - maximum of 2 secs for importing 6400 survey responses with a 10% outdated rate)

I will attempt to refactor the change handler classes in a separate PR to extract the common change batching logic

packages/meditrak-server/src/dhis/DhisChangeSideEffectHandler.js
```diff
@@ -21,9 +21,12 @@ export class DhisChangeSideEffectHandler extends ChangeSideEffectHandler {
...
           FROM survey_response
           JOIN survey ON survey_response.survey_id = survey.id
+          JOIN data_source ds ON survey.data_source_id = ds.id
           JOIN entity ON survey_response.entity_id = entity.id
-          WHERE survey_response.id IN (${batchOfSurveyResponseIds.map(() => '?').join(',')})
-          AND ${this.models.surveyResponse.getOnlyEventsQueryClause()};
+          WHERE
+            survey_response.id IN (${batchOfSurveyResponseIds.map(() => '?').join(',')})
+            AND ${this.models.surveyResponse.getOnlyEventsQueryClause()}
+            AND ds.service_type = 'dhis';
```
kael89, line 29: @edmofro  We have missed this in the past - we testing this locally, I was wondering why changed responses were always processed twice (they were marked as changed by this function)

edmofro: Right! Good catch, thanks for tidying up 🙏

kael89:
@rohan-bes fyi - I merged a couple of PRs reviewed by @edmofro (https://github.com/beyondessential/tupaia/pull/2891, https://github.com/beyondessential/tupaia/pull/2921) and rebased from `dev`

This should be ready for testing now 😃

kael89:
@rohan-bes added one more commit (https://github.com/beyondessential/tupaia/pull/2803/commits/b7376f6ecb42eb2e2ea7216029cbc5d47b6cb30b) to show non-outdated responses by default in the admin panel

kael89:
@edmofro after some investigation, I think I have found the source of the [notifications not working issue](https://linear.app/tupaia/issue/MEL-92#comment-fb01644c)! Our deployment process ([deployPackages.sh](https://github.com/beyondessential/tupaia/blob/dev/packages/devops/scripts/deployment/deployPackages.sh)) does the following in order:
1. Builds and start `meditrak-server`. This includes dynamically fetching the fields for the `Survey` and `SurveyResponse` models. At this stage migrations have not run, so the new fields (`outdated` in Survey, `period_granularity` in `SurveyResponse`) are not included in the models
2. Runs migrations, which add the new fields in the db

My theory is that after 2 happens, the new fields are still not added in the models. Console logs in [SurveyResponseOudater](https://github.com/beyondessential/tupaia/blob/2729-add-survey-period-granularity/packages/database/src/changeHandlers/SurveyResponseOutdater.js) indicate this. However, this raises some questions:
i. How come we haven't come across this issue in the past? Were we just lucky?
ii. Is there a way to invalidate the meditrak-server model fields in a migration?

My theory is also consistent with the fact that this feature works if `meditrak-server` is restarted: this time, the fields will be already present in the db when the server starts, so they will be added in the models successfully.

It also indicates an easy workaround: after the new fields end up in the production db, this won't be an issue anymore. However, if the above theory is correct, it would be nice to find a more long-term solution to the general issue.

edmofro:
Wow, nice find! That's hugely satisfying to have an answer that cleanly explains the behaviour above

i. Haven't come across it because in the past, we lazily loaded the schema the first time it was needed. In [this PR](https://github.com/beyondessential/tupaia/commit/6d09b7ac8bd4f1ba6400a49febee1df8d3208a2e#diff-cc52ec21ed7d955ee22da2eb18bf330290a5c79cf4f09d30f27ca17ae88c5151) I changed it to load on model construction (unnecessary byproduct of the actual change required). If we change that back though, we're still at risk of this same bug, it will just be more rare, so I suggest we keep that change and fix it in some other way
ii. I don't think so, at least not currently. Happy to have a call about this and come up with some solution, but I don't have anything off the top of my head!

PR #2806, opened 2021-07-11
3100: viz-builder-server scaffolding

chris-bes:
### Changes:

- Add viz-builder-server scaffolding.


packages/server-boilerplate/src/orchestrator/auth/middleware.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    if (!authHeader || !authHeader.startsWith('Bearer')) {
+      throw new UnauthenticatedError(
+        'No authorization header provided - must be Bearer Auth Header',
+      );
+    }
+    
+    const accessToken = getJwtToken(authHeader);
+    const authConnection = new AuthConnection();
+    const { accessPolicy: accessPolicyObject } = await authConnection.authenticateAccessToken(accessToken);
```
chris-bes, line 30: So from our tech design discussion, admin panel will assume that all requests coming in will contain an access token, and its acceptable that we dont have to care about session management. However, in order to authenticate an access token using the `Authenticator` class in the `auth` package, we need to instantiate a Model registry that at least contains the User model (this is how the middleware in microservice servers do it). As I understand, for orchestration servers, we want to limit the model access so that if we decide to migrate away from using the tupaia db then it should be easier to accomplish (although I'm not sure if we would ever want to do that for viz-builder?). 

So just to follow the pattern mentioned above, I decide to authenticate the access token by sending a request to `auth` endpoint instead of directly calling Authenticator. I've done a bit of research and found that although there's no standard way of how resource server validate access token, using auth server (meditrak-server) to authenticate the access token is a common pattern.

The only downside of this is we will need to have 1 extra call for every request coming in viz-builder-server. So slower performance is expected...

PR #2809, opened 2021-07-12
3133 unfpa marshall islands missing

billli0:
### Issue #:
[#3133](https://github.com/beyondessential/tupaia-backlog/issues/3133)
### Changes:
~Add relations in `ancestor_descendant_relation` for Marshall Island `facilties->district`, `facilties->country`, and `facilties->project` levels.~

Add relations in `entity_relation` for Marshall Island `facilties->district` level.


packages/database/src/migrations/20210709031533-UnfpaMarshallIslandsMissing-modifies-data.js
```diff
@@ -0,0 +1,74 @@
...
+    await db.runSql(`
+    SELECT * from entity 
+    WHERE country_code = 'MH' AND "type" = 'facility'; 
+ `)
+  ).rows;
+
+  for (const facility of facilities) {
+    const { parent_id: ancestorId, id: descendantId } = facility;
+    // console.log(facility);
+    await insertObject(db, 'ancestor_descendant_relation', {
```
rohan-bes, line 42: Hmmm I think this might not be the right approach here. I think we actually automatically build up the `ancestor_descendant_relation` table via the `EntityHierarchyCacher`. It gets built from the `entity_relation` table, which is where we want to be adding the entities to. A good example to draw from would be: `tupaia/packages/database/src/migrations/20210615072057-UpdateUnfpaNonCanonicalRelations-modifies-data.js`

rohan-bes (changes requested):
I think the correct way to ensure that the relations are built correctly is to add them to the `entity_relation` table and let the meditrak-server automatically build up the `ancestor_descendant_relation` table.

Looking at this issue though, I think it's possible it may have been introduced here: 20210615072057-UpdateUnfpaNonCanonicalRelations-modifies-data.js
Might be worth checking with @nick-beyondessential on it? I think the issue is caused because there are no explicit relations defined between the districts and facilities for Marshall Islands. However it's possible there may be other countries that have a similar issue as well?

Btw the meditrak-server is responsible for handling the buiding of the `ancestor_descendant_relation` table so you may want it running in the background while you run this migration?

packages/database/src/migrations/20210709031533-UnfpaMarshallIslandsMissing-modifies-data.js
nick-beyondessential: eek - I think you are correct @rohan-bes looks like Marshall Islands got missed. And yes should be updating entity_relation not ancestor_descendant_relation.

rohan-bes: Not your fault @nick-beyondessential I think it's a very easy mistake to make. The canonical hierarchy for UNFPA goes country->district->facility, which is the same hierarchy that Marshall Islands uses. However, since the relations at generation distance 2 have some entity_relations defined for other countries, it stop using the canonical hierarchy for that generation. I think we should be pretty wary of how we talk about relations within hierarchies, since often we talk about type->type relations (eg. country->district) but the caching logic just relations on distance relations...

Personally I'd love if we just made the call to do away with canonical hierarchies all together, and had to have explicit entity_relations defined for everything... I know it would blow up the table size, but the hybrid model confuses me quite often!

billli0:
@rohan-bes Do I need to do anything to build up the `ancestor_descendant_relation` after adding to `entity_relation`?

rohan-bes (approved):
Nice, LGTM!

PR #2810, opened 2021-07-12
3000: E2E lesmis config 

sepidehdd:
### Changes:

-  adding Lesmis visuals to E2E config

### DIFF
--- a/packages/web-frontend/cypress/config/dashboardReportUrls/default.json
+++ b/packages/web-frontend/cypress/config/dashboardReportUrls/default.json
@@ -1,5 +1,231 @@
 [
   {
+    "code": "LESMIS_repetition_rate_primary_district",
+    "project": "laos_schools",
+    "orgUnit": "LA_Phouvong District",
+    "dashboardGroup": "Student Outcomes",
+    "startDate": "2021-01-01",
+    "endDate": "2021-12-31"
+  },
+  {
+    "code": "LESMIS_repetition_rate_lowerSecondary_district",
+    "project": "laos_schools",
+    "orgUnit": "LA_Phouvong District",
+    "dashboardGroup": "Student Outcomes",
+    "startDate": "2021-01-01",
+    "endDate": "2021-12-31"
+  },
+  {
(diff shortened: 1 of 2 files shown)


packages/web-frontend/cypress/config/dashboardReportUrls/default.json
```diff
@@ -1,4 +1,230 @@
 [
+  {
+    "code": "LESMIS_repetition_rate_primary_district",
```
kael89, line 3: The urls are currently sorted by `code`. I believe it would be beneficial if we maintain this sort order, could you move them further down the file?

kael89 (changes requested):
Hi @sepidehdd , great work adding the new config! 👏 

One small changed requested (see my comment).

Also, a question for you: those reports were added in the `default.json` config (which will be the base config we'll use when running our usual regression tests. Do you think that all these new Lesmis reports should be included? Or should we just test a part of them? I don't have an answer to that, happy either way - but it's good to make that decision.

Also, I suggest creating a new `lesmis.json` config file and copy/paste the new reports there, in case we want to test only Lesmis reports for whatever reason.

sepidehdd:
Thanks Kostas for your review, to answer your question I think lets keep them all for E2E since they are the subset of Lesmis visualizations (we only included the dashboard reports with dashboardgroupID required as "No")

also I will create a lesmis.json file shortly.

kael89:
> Thanks Kostas for your review, to answer your question I think lets keep them all for E2E since they are the subset of Lesmis visualizations (we only included the dashboard reports with dashboardgroupID required as "No")
> 
> also I will create a lesmis.json file shortly.

Sounds great Sepi, thanks!!

kael89 (approved):
Looks great!

kael89:
Feel free to merge this when the latest [merge in dev](https://app.codeship.com/projects/379708/builds/54767b76-0574-4dda-8daf-c25aba35612e?step=pull_latest_into_ec2_reinstate_SSH_Private_Key_File) finishes @sepidehdd - please choose "Squash and merge" for this PR

PR #2811, opened 2021-07-12
2839 fix unfpa dashboards missing fiji

billli0:
### Issue #:
[#2839](https://github.com/beyondessential/tupaia-backlog/issues/2839)

### Changes:
apply logic from [this comment](https://github.com/beyondessential/tupaia-backlog/issues/2839#issuecomment-851087096)


packages/web-config-server/src/apiV1/dataBuilders/generic/percentage/percentagesOfValueCountsPerPeriod.js
```diff
@@ -16,7 +16,7 @@ const filterFacility = async (models, filterCriteria, analytics) => {
   const facilities = await models.facility.find({
     type: {
       comparator: filterCriteria.comparator,
-      comparisonValue: '1',
+      comparisonValue: '1%',
```
nick-beyondessential, line 19: Should work as long as we don't add more than 9 facility types. :thinking:

billli0: oh yeah @nick-beyondessential that's a good point. Not sure if that works with Tupaia but I will try `"type" similar to '(1|1.%)'`, which should handle more than 9 facility types.

nick-beyondessential: I wouldn't stress too much about it - not sure how likely it is.

billli0: Will keep that in mind.

nick-beyondessential (approved):
LGTM

PR #2812, opened 2021-07-12
Handle missing user sessions

tcaiger:
### Changes:

- clear session cookie from browser if the server can't find it in the db (in lesmis-server and server-boilerplate)
- add some error boundaries to lesmis to stop whole app crashing if there is a js error
- handle no data better in lesmis dashboard tab component

I also investigated psss and it is already handling the case of no session record in the db well by logging the user out completely.

---

### DIFF
--- a/packages/lesmis/src/components/MapOverlaysPanel/MapOverlaysPanel.js
+++ b/packages/lesmis/src/components/MapOverlaysPanel/MapOverlaysPanel.js
@@ -13,6 +13,7 @@ import { MapOverlayGroup } from './MapOverlayGroup';
 import { MapOverlaysLoader } from './MapOverlaysLoader';
 import { FlexStart } from '../Layout';
 import { MapOverlaysPanelContainer as Container } from './MapOverlaysPanelContainer';
+import { ErrorBoundary } from '../ErrorBoundary';
 
 const Header = styled.div`
   padding: 1.25rem 1.875rem 2rem;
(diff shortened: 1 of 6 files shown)


packages/lesmis/src/views/DashboardReportTabView.js
```diff
@@ -163,37 +163,36 @@ export const DashboardReportTabView = ({
       </StickyTabBarContainer>
       <DashboardSection ref={topRef}>
         <FetchLoader isLoading={isLoading} isError={isError} error={error}>
```
tcaiger, line 165: The only actual change here is chaning: `subDashboards &&
            subDashboards.map` to `subDashboards?.map`

packages/lesmis/src/components/MapOverlaysPanel/MapOverlaysPanel.js
```diff
@@ -69,35 +70,37 @@ export const MapOverlaysPanel = ({
   ]);
 
   return (
```
tcaiger, line 72: the only change here is wrapping everything with `ErrorBoundary`

tcaiger: The error boundary will show an error message if there are any unhandled errors further down the component tree rather than the whole app crashing.

EMcQ-BES (approved):
lgtm

PR #2813, opened 2021-07-12
3164: Admin Panel returns all dashboard relations when expanding a dashboard in Dashboard tab

chris-bes:
### Changes:

- Do not return all dashboard relations when expanding a dashboard if user is BES Admin


EMcQ-BES (approved):
Looks good, simple fix

PR #2816, opened 2021-07-12
Release v2021-27-1 (hotfix: e2e dumps)

kael89:
This is not a bug but I thought it would be beneficial to release as hotfix, because it blocks proper usage of baseline e2e data in e2e.tupaia.org.

We will need to update the cron jobs in production after this is released.


edmofro (approved):
lgtm, and happy for it to go out as a hotfix

PR #2817, opened 2021-07-12
1648: Use session cookie in Admin Panel

chris-bes:
### Changes:

- Instead of changing the existing /auth endpoint, I've implemented 2 new routes strictly used only for Admin Panel which are /login and /logout which will set up/clear the session cookie. I think the existing /auth endpoint can be treated as our auth server module.
- Cleaned up access token and refresh token storage in the front end.
- Added `admin_panel_session` table


packages/meditrak-server/src/auth/middleware.js
```diff
@@ -25,6 +26,17 @@ async function authenticateUser(req) {
     return { userId: tokenUserID };
   }
 
+  if (req.sessionCookie?.id) {
+    const sessionId = req.sessionCookie?.id;
+    const sessionModel = new AdminPanelSession(req.database);
+    const session = await sessionModel.findById(sessionId);
```
edmofro, line 32: Any reason not to do:
```suggestion
    const session = await req.models.session.findById(sessionId);
```

packages/admin-panel/src/authentication/actions.js
```diff
@@ -61,9 +59,14 @@ export const loginError = errorMessage => ({
...
-  type: LOGOUT,
-});
+export const logout = () => async (dispatch, getState, { api }) => {
+  await api.logout();
+
+  dispatch({
+    // Set state to logging out
+    type: LOGOUT,
+  });
+};
```
edmofro, line 69: Perhaps dispatch the logout before waiting for the actual logout to complete, so that the user experiences it quickly. The ideal thing would be to add error handling, but I'm happy to just go with showing the logged out situation immediately

packages/admin-panel/src/api/TupaiaApi.js
```diff
@@ -6,10 +6,7 @@
 import { saveAs } from 'file-saver';
 
 import { verifyResponseStatus, stringifyQuery } from '@tupaia/utils';
-import { AccessPolicy } from '@tupaia/access-policy';
```
edmofro, line 9: Hooraaaaay for the simplification of this file 🎉

packages/database/src/migrations/20210712082624-AddAdminPanelSessionTable-modifies-schema.js
```diff
@@ -0,0 +1,37 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.createTable('admin_panel_session', {
```
edmofro, line 18: Could you please create a card for storing the session outside of the postgres database? We could use redis or some other session cache, rather than having all these extra dependencies on the central tupaia db

edmofro (changes requested):
All really good code @chris-bes, though I am not so hot on muddying the role of meditrak-server further by adding login and logout that are admin panel specific. I am probably entering this a bit late and should have joined tech design sessions, but what do you think of just making an admin-panel orchestration server (seeing as you're putting in the scaffolding for a viz builder server anyway)? 

Happy to have a call about this :-)

packages/database/src/migrations/20210712082624-AddAdminPanelSessionTable-modifies-schema.js
chris-bes: Created: https://github.com/beyondessential/tupaia-backlog/issues/3188

packages/meditrak-server/src/auth/middleware.js
chris-bes: This part of code is no longer relevant!

chris-bes:
I've reverted the code in meditrak-server, only keeping the changes in admin-panel front end so that it can work along with the admin-panel-server in the next PR

edmofro (approved):
lgtm!

PR #2820, opened 2021-07-13
Allow non-5432 db ports

Alastair-L:
### Issue #:
No issue
### Changes:
Uses a configurable database port rather than assuming 5432.

This is so that I can have postgresql running on both windows side and wsl at the same time. If we don't want it to be configurable, that's fine, I can just keep it locally.

Also, I didn't look too hard for anywhere else where the port would need to be changed, hence why I'm creating as a draft.


rohan-bes (approved):
LGTM! Nice one 🎉

rohan-bes:
Ah may just need to add the key to the CI env vars to get the tests running? To fix this error when installing the analytics table:
`psql: error: invalid integer value "-X" for connection option "port"`

edmofro:
Thanks for reviewing this @rohan-bes, and good pickup on the codeship env vars. Could be even nicer though to make it default to `5432` everywhere rather than having the `DB_PORT` env var as a strict requirement?

rohan-bes:
> Thanks for reviewing this @rohan-bes, and good pickup on the codeship env vars. Could be even nicer though to make it default to `5432` everywhere rather than having the `DB_PORT` env var as a strict requirement?

Ah that's a good point @edmofro ! Arguably also the case for a couple of the other DB connection env vars too!

Hopefully it's not too tricky to update the analytics table shell scripts to have a default port?

Alastair-L:
@rohan-bes would you mind rereviewing? Quite a few changes since your passing review 🙏

packages/utils/src/requireEnv.js
```diff
@@ -17,3 +17,11 @@ export const requireEnv = variable => {
...
   return value;
 };
+
+export const getEnvVarOrDefault = (variable, defaultValue) => {
+  const value = process.env[variable];
+  if (value === undefined) {
+    return defaultValue;
+  }
+  return value;
+};
```
rohan-bes, line 27: Nice, this will come in handy! We've been doing a lot of `process.env[x] || 'default'` about the place which is a tad ugly imo. Will be nice to switch to this 👍

rohan-bes (approved):
LGTM!

Alastair-L:
@edmofro are you happy for me to merge this?

packages/data-api/scripts/buildAnalyticsMaterializedView.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 source .env
+
+# Set default port in case it wasn't in .env
+: "${DB_PORT:=5432}"
```
edmofro, line 5: Wow nice to learn about this defaulting syntax!

edmofro (approved):
Yep!

PR #2823, opened 2021-07-13
3032: Use POST for MultiEntity routes in Entity-Server 

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/3032:

Pretty straight-forward issue here. We were requesting with a URL param that was too large previously when there were lots of entities being request (AU had requests for ~500 LGAS).

Switching to POST and moving the entities to the body resolves this issue. Luckily no refactor was needed in lesmis or psss 🎉


packages/server-boilerplate/src/microService/api/ApiBuilder.ts
```diff
@@ -81,6 +81,14 @@ export class ApiBuilder {
     return this;
   }
 
+  post<T extends ExpressRequest<T> = Request>(
```
edmofro, line 84: Woah it's our first `post`? Crazy

packages/aggregator/src/connections/EntityConnection.js
```diff
@@ -21,11 +21,14 @@ export class EntityConnection extends ApiConnection {
...
   ) {
-    return this.get(`hierarchy/${hierarchyName}/relatives`, {
-      entities: entityCodes.join(','),
-      descendant_filter: `type:${dataSourceEntityType}`,
-      field: 'code',
-    });
+    return this.post(
+      `hierarchy/${hierarchyName}/relatives`,
+      {
+        filter: `type:${dataSourceEntityType}`,
```
edmofro, line 27: Why did `descendant_filter` become `filter`?

edmofro (commented):
All looks good with just one question that I'll look into now, but you might answer sooner than I figure it out

packages/aggregator/src/connections/EntityConnection.js
edmofro: It looks like it's correct now, and was maybe a bug before?

edmofro (approved):
Looks correct to me, approved :-)

packages/aggregator/src/connections/EntityConnection.js
rohan-bes: Yup, was actually a bug before, and a pretty nasty one at that! We were always returning all relatives instead of the ones of the correct level!

packages/server-boilerplate/src/microService/api/ApiBuilder.ts
rohan-bes: I know!

PR #2824, opened 2021-07-13
3143: Keep ssh connection alive during front end builds

edmofro:
### Issue #:

https://github.com/beyondessential/tupaia-backlog/issues/3143

I'm not at all confident this will help, but it's possible that it will so worth a try.

### DIFF
--- a/packages/devops/scripts/ci/deployFrontend.sh
+++ b/packages/devops/scripts/ci/deployFrontend.sh
@@ -4,7 +4,7 @@ DIR=$(dirname "$0")
 DEPLOYMENT_URL=$(${DIR}/determineDeploymentUrl.sh)
 if curl --output /dev/null --silent --head --fail $DEPLOYMENT_URL; then
     echo "Deployment for ${CI_BRANCH} exists, updating with latest changes"
-    ssh ubuntu@$DEPLOYMENT_URL "cd tupaia; yarn workspace @tupaia/${PACKAGE} build"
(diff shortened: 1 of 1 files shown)


packages/devops/scripts/ci/deployFrontend.sh
```diff
@@ -4,7 +4,7 @@ DIR=$(dirname "$0")
 DEPLOYMENT_URL=$(${DIR}/determineDeploymentUrl.sh)
 if curl --output /dev/null --silent --head --fail $DEPLOYMENT_URL; then
     echo "Deployment for ${CI_BRANCH} exists, updating with latest changes"
-    ssh ubuntu@$DEPLOYMENT_URL "cd tupaia; yarn workspace @tupaia/${PACKAGE} build"
+    ssh -o ServerAliveInterval=15 ubuntu@$DEPLOYMENT_URL "cd tupaia; yarn workspace @tupaia/${PACKAGE} build"
```
edmofro, line 7: Sends a null packet to the server every 15 seconds if no data is received, which should help to keep the ssh connection alive and receiving logs

EMcQ-BES: Note that the default value of `ServerAliveCountMax` is 3 so the timeout on an actual connection problem will now be 45 seconds instead of whatever the default is.

edmofro: Yep good point, but happy with that

EMcQ-BES (approved):

PR #2825, opened 2021-07-13
Avoid creating spurious countries, throw error instead

edmofro:
@avaek is looking into why project entity imports are not getting a match on the country ISO code, but in the meantime we might as well get rid of the behaviour that's causing this and leave it as an error instead.

### DIFF
--- a/packages/utils/src/getCountryCode.js
+++ b/packages/utils/src/getCountryCode.js
@@ -10,13 +10,7 @@ export const getCountryCode = (countryName, entityObjects = []) => {
   }
 
   const { getCode: getCountryIsoCode } = require('countrynames');
-  // Use the country ISO code if there's a direct match,
-  //otherwise base on the two letter facility code prefix
-  const countryCode =
-    getCountryIsoCode(countryName) || entityObjects.length
-      ? entityObjects[0].code.substring(0, 2)
-      : null;
-
+  const countryCode = getCountryIsoCode(countryName);
   if (!countryCode) throw new ImportValidationError(`${countryName} is not a recognised country`);
   return countryCode;
 };


packages/utils/src/getCountryCode.js
```diff
@@ -10,13 +10,7 @@ export const getCountryCode = (countryName, entityObjects = []) => {
   }
 
   const { getCode: getCountryIsoCode } = require('countrynames');
-  // Use the country ISO code if there's a direct match,
-  //otherwise base on the two letter facility code prefix
-  const countryCode =
-    getCountryIsoCode(countryName) || entityObjects.length
-      ? entityObjects[0].code.substring(0, 2)
-      : null;
-
```
edmofro, line 19: Original reason for the line that interpolates the country code from the entity code prefix was to deal with demo land. Now that we have an explicit check for Demo Land (see a few lines up), we shouldn't have any other countries coded using non-ISO codes, so can remove this dangerous behaviour.

EMcQ-BES (approved):
lgtm

EMcQ-BES:
Actually, just saw some unit tests are failing when this is run against Laos? Can you check that out

edmofro:
Right, that shows it wasn't just being used for Demo Land! Glad we've got unit tests

I guess Laos is Laos People's Democratic Republic officially. It's a shame `countrynames` doesn't have a more fuzzy matching search.

Which also shows why it failed for Fiji (Republic of Fiji) and Samoa (Independent State of Samoa)!

@avaek you can call off your investigation now, but please write up a ticket for fixing this, and another for deleting the existing country duplicates 🙏 

Closing this as it would stop us importing entities into the countries for which we use a non ISO name

PR #2826, opened 2021-07-13
3001 Add postcodes in explore

nick-beyondessential:
### Changes:
- Add postcode type to explore canonical types in explore hierarchy
- Redefine covid_samoa districts so that explore remains canonical and 'undefined entities' show only in covid_samoa.
- hide postcodes from explore search bar

### DIFF
--- a/packages/database/src/migrations/20210713012631-AddPostcodeTypeToExplore-modifies-data.js
+++ b/packages/database/src/migrations/20210713012631-AddPostcodeTypeToExplore-modifies-data.js
@@ -1,0 +1,68 @@
+'use strict';
+
+import { codeToId, findSingleRecord, insertObject, generateId } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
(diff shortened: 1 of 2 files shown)


packages/database/src/migrations/20210713012631-AddPostcodeTypeToExplore-modifies-data.js
```diff
@@ -0,0 +1,68 @@
...
+        id: generateId(),
+        parent_id: samoaId,
+        child_id: district.id,
+        entity_hierarchy_id: covidSamoaId,
+      }),
+    ),
+  );
+
+  // set Unknown District parent_id to null so doesn't show up in other samoa projects
+  await db.runSql(`update entity set "parent_id" = null where "code" = 'WS_Unknown_District'`);
```
billli0, line 59: It works! A bit curious how it will show up under Samoa when it has a `null` parent.

billli0 (approved):
LGTM!

packages/database/src/migrations/20210713012631-AddPostcodeTypeToExplore-modifies-data.js
nick-beyondessential: The idea is by setting parent_id = null, the entity only exists in an alternate hierarchy, which we define in the entity_relation table for covid_samoa only. The two other Samoa districts retain their parent_id so will show up in other projects, but the 'unknown entities' district->facility->vilage will not.

billli0: That makes sense 👍

PR #2827, opened 2021-07-13
Enable e2e tests in CI/CD

kael89:
This allows us to run e2e tests in CI/CD! I tested this yesterday (a previous version of the code, will need to check again) and it works.

* https://app.codeship.com/projects/379708/builds/f3160c9d-139c-4bfd-98ee-ed36e36d9467?step=pull_latest_into_ec2_reinstate_SSH_Private_Key_File - run for just one dashboard (tests pass)
* https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/a53ff7d8-4d3e-4c24-a1f6-351bcea975ed?step=e2e-test_e2e-test-web-frontend: a lot of Laos Schools reports failed because the e2e user did not have the required permissions - this is fixed now

We may still need to tweak a few things (timeouts etc), but the foundations seem ok

I also updated the documentation: https://beyond-essential.slab.com/posts/e-2-e-test-guide-cnj4rgdb


packages/web-frontend/cypress/support/utils.js
```diff
@@ -32,10 +32,7 @@ const stripHtmlAttributes = (html, attrs) => {
  */
 export const serializeReactToHTML = jQueryEl => {
   const html = jQueryEl[0].outerHTML;
-  const sanitizedHtml = stripHtmlAttributes(html, HTML_ATTRS_TO_STRIP_FROM_SNAPSHOTS).replace(
-    /(&amp;|)cacheBreaker=[0-9a-z]+/g,
-    '',
-  );
+  const sanitizedHtml = stripHtmlAttributes(html, HTML_ATTRS_TO_STRIP_FROM_SNAPSHOTS);
```
edmofro, line 35: Good catch removing this, really glad we got rid of it (thanks @IgorNadj!)

packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -32,6 +29,13 @@ for PACKAGE in ${PACKAGES[@]}; do
     # name (e.g. [branch-name]-api.tupaia.org -> specific-branch-api.tupaia.org)
     sed -i -e "s/\[branch-name\]/${BRANCH}/g" .env
 
+    if [[ "$BRANCH" == *-e2e ]]; then
+        # Update e2e environment variables
+        if [[ $PACKAGE == "meditrak-server" || $PACKAGE == "web-config-server" ]]; then
+            sed -i -e "s/^AGGREGATION_URL_PREFIX=dev-$/AGGREGATION_URL_PREFIX=e2e-/g" .env
+        fi
+    fi
```
kael89, line 37: This means that all branches with an `-e2e` suffix will be connected to `e2e` aggregation by default.

Doing so in this way is probably not ideal, since the custom env logic is a bit hidden in this file. However, I felt that we don't need a more elaborate mechanism in this case.

I added a note in the e2e doc:

> By default, all instances whose subdomain ends in -e2e will be connected to the e2e aggregation servers

but I should probably add a note in a more generic doc. Deployment logic is scattered in many places (which is definitely justified), perhaps we could mitigate this with a single doc describing it.

codeship-services.yml
```diff
@@ -26,6 +28,17 @@ testing:
   encrypted_dockercfg_path: dockercfg.encrypted
   default_cache_branch: 'dev'
 
+e2e:
+  build:
+    dockerfile: e2e.Dockerfile
+  encrypted_env_file: ./ci-env-vars.encrypted
+  cached: true
+  encrypted_dockercfg_path: dockercfg.encrypted
+  default_cache_branch: 'dev'
```
edmofro, line 37:
```suggestion
  default_cache_branch: 'e2e'
```
The cache probably won't work from `dev` as dev won't run e2e and build from that Dockerfile at all, so instead we could keep an `e2e` branch semi-regularly updated from dev? As a performance optimisation only, so wouldn't be too important if we forgot to keep it updated. Either that or we could just make `dev` build the e2e docker container, even if it doesn't use it.

packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -32,6 +29,13 @@ for PACKAGE in ${PACKAGES[@]}; do
     # name (e.g. [branch-name]-api.tupaia.org -> specific-branch-api.tupaia.org)
     sed -i -e "s/\[branch-name\]/${BRANCH}/g" .env
 
+    if [[ "$BRANCH" == *-e2e ]]; then
+        # Update e2e environment variables
+        if [[ $PACKAGE == "meditrak-server" || $PACKAGE == "web-config-server" ]]; then
+            sed -i -E 's/^AGGREGATION_URL_PREFIX="?dev-"?$/AGGREGATION_URL_PREFIX=e2e-/g' .env
```
edmofro, line 35: Oh nice, this is where you do the manual switch! I thought it would be in the js code, but this makes way more sense as it is a concern of our infrastructure rather than the code 🎉

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
```diff
@@ -36,7 +36,7 @@ const urlToRouteRegex = url => {
     throw new Error(`'${url}' is not a valid report url: it must contain a 'report' query param`);
   }
 
-  return new RegExp(`report.*\\WisExpanded=true&.*itemCode=${itemCode}[&$]`);
+  return new RegExp(`report/.*\\?(.*&|)isExpanded=true&(.*&|)itemCode=${itemCode}(&|$)`);
```
edmofro, line 39: I've seen you change this a few times so I'm not going to bother trying to parse it, will trust you've got it sorted now :-)

packages/web-frontend/cypress/scripts/testE2e/testE2e.js
```diff
@@ -5,41 +5,109 @@
...
-const runPackageScript = script =>
-  execSync(`yarn workspace @tupaia/web-frontend ${script}`, { stdio: 'inherit' });
+const printResults = ({ baseUrl, compareUrl }, { baseError, compareError }) => {
+  const logger = getLoggerInstance();
+
+  logger.info();
+
+  const printResult = (description, error) => {
+    if (error) {
+      logger.error(`❌ ${description}`);
```
edmofro, line 31: Do we want to log the error as well?

edmofro (approved):
Very nice! This is so great to be at a point where we can have e2e tests in CI/CD, thank you for your really quite heroic persistence over many many months to get there 🎖️ 🎉 👏

packages/web-frontend/cypress/e2e/dashboardReports.spec.js
kael89: Haha yeah this time I tested it more thoroughly locally! It would help if I would have added unit tests to test variations for this in the first place, but I thought no point testing the test code 😁

packages/devops/scripts/deployment/deployPackages.sh
kael89: I thought the same initially, then I was lead here during development!

packages/web-frontend/cypress/scripts/testE2e/testE2e.js
kael89: We do! It will have been logged already (in https://github.com/beyondessential/tupaia/blob/e337ee627a52360701cc8a0ee2b0e6a203d5fe80/packages/web-frontend/cypress/scripts/testE2e/testE2e.js#L98)

This is just to show a summary of the results per url in the end:

image

edmofro: Gotcha, great!

codeship-services.yml
kael89: I see, that makes a lot of sense!

By the way , currently the https://e2e.tupaia.org ("default" baseline branch) is a production clone. I know that testers usually use `dev` to test stuff midweek, but before a release we use production as a baseline.

What do you think? In that case, we can regularly update `e2e` from `master` (instead of `dev`). We can see how this goes and maintain two permanent baseline branches, if need be (https://e2e.tupaia.org, https://dev-e2e.tupaia.org).

kael89:
@edmofro I fixed a few more stuff, this should be ready to merge now! Keen to hear your thoughts on https://github.com/beyondessential/tupaia/pull/2827#discussion_r670028296 when you have time

codeship-services.yml
edmofro: Happy for `e2e` to be regularly updated from `master` :-)

codeship-services.yml
```diff
@@ -34,7 +34,7 @@ e2e:
   encrypted_env_file: ./ci-env-vars.encrypted
   cached: true
   encrypted_dockercfg_path: dockercfg.encrypted
-  default_cache_branch: 'dev'
+  default_cache_branch: 'master'
```
edmofro, line 37: Will we run these steps on`master`? I don't think so right?

kael89: Oh yeap you are right... I assumed that CI/CD will pick up the commit identity between branches, but I guess this is not the case

kael89:
Merging this straight into `dev` as most changes are in cypress tests. This does include some CI/CD related changes: I have already tested those, and the deployment process for feature instances should also be tested while deploying branches for the release testing

PR #2828, opened 2021-07-13
2761: % of non-First Nation participants with influenza like illness (ILI) Map Overlay

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/2761:

Requires https://github.com/beyondessential/tupaia/pull/2823 to already be merged

### Changes:
- Measure Options are now no longer displayed in the legend if they have 'hideFromLegend' = true
- For ShadedSpectrum map overalys, we now use the color of the overlay if it already exists

---

### Screenshots:
In ticket https://github.com/beyondessential/tupaia-backlog/issues/2761#issuecomment-879481493


packages/web-frontend/src/containers/MeasureLegend/MultiLegend.js
```diff
@@ -55,7 +55,9 @@ const getLegend = measureType => {
 };
 
 const MultiLegend = React.memo(({ measureOptions, isMeasureLoading }) => {
-  const displayedLegends = measureOptions.filter(({ type }) => type !== MEASURE_TYPE_RADIUS);
+  const displayedLegends = measureOptions.filter(
+    ({ type, hideFromLegend }) => type !== MEASURE_TYPE_RADIUS && hideFromLegend !== true,
+  );
```
rohan-bes, line 60: Quite surprised we didn't have this previously, as we attach this key to the measureOption by default...

packages/web-frontend/src/utils/measures.js
```diff
@@ -349,14 +349,16 @@ export function getMeasureDisplayInfo(measureData, measureOptions, hiddenMeasure
...
-          displayInfo.color = resolveSpectrumColour(
-            scaleType,
-            scaleColorScheme,
-            valueInfo.value || (valueInfo.value === 0 ? 0 : null),
-            min,
-            max,
-            noDataColour,
-          );
+          displayInfo.color =
+            displayInfo.color ||
```
rohan-bes, line 353: This kind of behaviour can be seen in this function already, however in general it seems fairly inconsistent.

It really is a bit of a hack to be honest and could definitely be talked into doing something more comprehensive...

billli0: I got the same feeling as you did, but not sure if this change will effect some visuals? We previously always want to replace displayInfo.color with the last `measureOption`.
  
But I just don't have any guts to change it.

rohan-bes: I think we may be in the clear on this. I used this SQL query to check for the following: 
- Any shaded-spectrum mapOverlays with linked measures
- Any shaded-spectrum mapOverlays that were a linked measure of another overlay
```
select * from "mapOverlay" mo
where ("presentationOptions" ->>'displayType' = 'shaded-spectrum' and "linkedMeasures" != null)
or (id in (select unnest("linkedMeasures") as linked_measures
  from "mapOverlay" mo) and "presentationOptions" ->>'displayType' = 'shaded-spectrum')
```
I got 0 results, which I think means that 'for all map overlays with shaded-spectrum type, there is just a single measureOption for each'? So this change should be safe?

I feel we can still come up with a different approach to fixing this anyway, lets chat at some point today about it 👍

billli0 (approved):
Code looks clean 👍

packages/database/src/migrations/20210713071004-AddFluTrackerNonFirstNationWithIliMapOverlay-modifies-data.js
```diff
@@ -0,0 +1,142 @@
...
+    },
+    measureConfig: {
+      Participants: {
+        type: 'popup-only',
+        hideFromLegend: true,
+      },
+      'Participants with influenza like illness (ILI)': {
+        type: 'popup-only',
+        hideFromLegend: true,
+      },
```
billli0, line 92:
```suggestion
      $all {
        type: 'popup-only',
        hideFromLegend: true,
      }

```
We can use $all config

packages/database/src/migrations/20210713071004-AddFluTrackerNonFirstNationWithIliMapOverlay-modifies-data.js
```diff
@@ -0,0 +1,142 @@
...
+    ],
+  },
+};
+
+const MAP_OVERLAY = {
+  id: REPORT_CODE,
+  name: '% of non-First Nation participants with influenza like illness (ILI)',
+  userGroup: 'Public',
+  dataElementCode: '% of participants with influenza like illness (ILI)',
```
billli0, line 62: It looks fine for me to use a customised value key, but previously all reports use `value`. Perhaps we can make it consistent to other reports?

packages/database/src/migrations/20210713071004-AddFluTrackerNonFirstNationWithIliMapOverlay-modifies-data.js
rohan-bes: Ahh good call!

PR #2829, opened 2021-07-14
3179: Use explicit list of unofficial country names

edmofro:
### Issue #:

https://github.com/beyondessential/tupaia-backlog/issues/3179

### DIFF
--- a/packages/meditrak-server/src/routes/importEntities/assertCanImportEntities.js
+++ b/packages/meditrak-server/src/routes/importEntities/assertCanImportEntities.js
@@ -6,9 +6,9 @@ import { getCountryCode } from '@tupaia/utils';
 import { TUPAIA_ADMIN_PANEL_PERMISSION_GROUP } from '../../permissions';
 
 export const assertCanImportEntities = async (accessPolicy, entitiesByCountryName) => {
-  const countryCodes = Object.entries(entitiesByCountryName).map(([countryName, entities]) => {
-    return getCountryCode(countryName, entities);
-  });
+  const countryCodes = Object.keys(entitiesByCountryName).map(countryName =>
+    getCountryCode(countryName),
+  );
 
   for (let i = 0; i < countryCodes.length; i++) {
     const countryCode = countryCodes[i];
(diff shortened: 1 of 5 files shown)


packages/utils/src/getCountryCode.js
```diff
@@ -4,19 +4,20 @@
...
+};
 
+export const getCountryCode = countryName => {
   const { getCode: getCountryIsoCode } = require('countrynames');
-  // Use the country ISO code if there's a direct match,
-  //otherwise base on the two letter facility code prefix
   const countryCode =
-    getCountryIsoCode(countryName) || entityObjects.length
-      ? entityObjects[0].code.substring(0, 2)
-      : null;
```
edmofro, line 18: The problem here was that I'm guessing this code intended to act as a lazily evaluated conditional with an early return (i.e., if getCountryIsoCode returned something, use that), but instead the whole || was just used as the condition for the ternary operator. That results in "if there's a valid ISO code for this country name, use the first two letters of the first entry in the spreadsheet", which is not what anyone would have wanted, I'm sure.

packages/meditrak-server/src/routes/importEntities/assertCanImportEntities.js
```diff
@@ -6,9 +6,9 @@ import { getCountryCode } from '@tupaia/utils';
 import { TUPAIA_ADMIN_PANEL_PERMISSION_GROUP } from '../../permissions';
 
 export const assertCanImportEntities = async (accessPolicy, entitiesByCountryName) => {
```
kael89, line 8: It seems this function requires just country names now:

```suggestion
export const assertCanImportEntities = async (accessPolicy, countryNames) => {
```

packages/utils/src/getCountryCode.js
kael89: Great explanation!

kael89 (approved):
Looks good! I left one suggestion but it's minor, approving

packages/meditrak-server/src/routes/importEntities/assertCanImportEntities.js
edmofro: Thorough review for a quick pR! I've made that change, good suggestion

PR #2836, opened 2021-07-14
3100: Add admin-panel-server

chris-bes:
### Changes:

- Add admin-panel-server


packages/devops/configs/servers.conf
```diff
@@ -234,6 +234,50 @@ server {
         internal;
   }
 
+}
+server {
+
+  listen       443 ssl http2;
+  listen       [::]:443 ssl http2;
+  server_name  ~^.*admin-panel-api\.tupaia\.org$;
+  root         /usr/share/nginx/html;
```
edmofro, line 243: Will be a conflict here with @IgorNadj's upcoming changes to move ssl termination to the ELB. Please both watch this and take care to update the other branch after the first merge to dev

edmofro (approved):
Wow that's pretty fancy stuff with the proxy! Honestly I was expecting to just pass all requests that weren't otherwise handled to `MeditrakConnection`, but you've gone to the next level of generic handling 🎉 

All looks good to me as the base layer of admin-panel orchestration server, really nice to get consistency here with other front ends!

packages/devops/configs/servers.conf
```diff
@@ -258,7 +258,7 @@ server {
   include /etc/nginx/h5bp/basic.conf;
 
   location / {
-              proxy_pass http://localhost:8040;
+              proxy_pass http://localhost:8070;
```
chris-bes, line 261: I spent the entire day scratching my head on why admin-panel of the deployment doesn't work, and any code changes just didn't seem to have any effects... turned out it was because of this wrong port 🤕

PR #2837, opened 2021-07-14
3118 + 3119: /dataSources and /permissionGroups

chris-bes:
### Changes:

- Add /dataSources and /permissionGroups


chris-bes:
I've also thought of this that might be worth to discuss. Since we're now having `admin-panel-server` that can forward unhandled requests to meditrak-server by default, is it worth implement these proxy endpoints anymore (endpoints that just call meditrak-server and return whatever without any data manipulation)? It seems a bit unnecessary now because meditrak-server endpoints support everything we need right? (search, limit, pagination,...). So we don't even have to write proxy endpoints like this. Just let viz-builder front end call the meditrak-server similar /dataSources endpoint format, just different base url (admin-panel-server)

The only concern is if in the future we need to manipulate any data returned from meditrak-server for viz-builder, then admin-panel-server seems the be the most appropriate place. But even if that happens, I'd imagine we would just need to add the endpoint into this admin-panel-server

PR #2838, opened 2021-07-15
3143: Added keep alive during running migrations

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/3143:

Identical to the change here, just added for migrations as well: https://github.com/beyondessential/tupaia/pull/2824


edmofro (approved):
lgtm!

I do feel we might as well add it for backend deployments too, as now it's becoming the rule rather than the exception to have the keep alive interval. I'll do that now (on this branch)

edmofro:
@rohan-bes ready to merge if you're happy with that extra couple of spots

PR #2840, opened 2021-07-15
Increase request timeout

kael89:
Most of the reports in [default.json](https://github.com/beyondessential/tupaia/blob/dev/packages/web-frontend/cypress/config/dashboardReportUrls/default.json) load in 15secs. There are a couple that do not, which prevents us from testing them, but I am a bit hesitant to further increasing the timeout:

* It can act as a performance test (performance degradation can be caught earlier)
* Reports can timeout because they are slow, or because the url doesn't exist. The second scenario may happen because of invalid url configuration or bugs preventing the reports from loading. No reason to wait too long in those cases

Would like to get another opinion on that though. Another option would be using a very generous timeout for data correctness tests, and regularly run performance tests with a stricter timeout.

### DIFF
--- a/packages/web-frontend/cypress.json
+++ b/packages/web-frontend/cypress.json
@@ -3,7 +3,7 @@
   "testFiles": "*.spec.js",
   "integrationFolder": "cypress/e2e",
   "projectId": "pty5cc",
-  "requestTimeout": 10000,
+  "requestTimeout": 15000,
   "useRelativeSnapshots": true,
   "viewportHeight": 900,
   "viewportWidth": 1400


IgorNadj (approved):
Happy for you to decide on how we want to catch those slow reports, but code change approved if you decide to do it

kael89:
> Happy for you to decide on how we want to catch those slow reports, but code change approved if you decide to do it

Yeah I'm not sure either, but I thought let's set the standards where they really are at the moment (15 secs should be enough for the vast majority of reports). We can still trigger runs with custom timeouts if we want to test for performance

PR #2841, opened 2021-07-15
3191: Hot fix bug in analytics table processing

edmofro:
Goes with https://github.com/beyondessential/pg-mv-fast-refresh/pull/3 

Issue:

https://github.com/beyondessential/tupaia-backlog/issues/3191

Have tested locally, could do with a light regression test tomorrow morning but it is just one line of code changed...

The other thing is it's really quite safe to merge because it would only apply if we then uninstalled and reinstalled mv-fast-refresh, or manually ran that part of the sql (the `CREATE OR REPLACE FUNCTION`)


rohan-bes (approved):
Nice work @edmofro! One thing to note is that just to be careful when merging the pg-mv-refresh PR to ensure that the commit id stays the same (I'm worried that 'squash-n-merge' might alter the commit id?)

edmofro:
Really good note there @rohan-bes, the commit id did indeed change so I've updated this PR to use the new one

PR #2738, opened 2021-06-29
2970: Add repetition rate visuals at district

edmofro:
### Issue #:

https://github.com/beyondessential/tupaia-backlog/issues/2970#issuecomment-870128100

Screenshot [here](https://github.com/beyondessential/tupaia-backlog/issues/2970#issuecomment-870128100)

PR #2843, opened 2021-07-16
2738 first nation with ili graph

billli0:
### Issue #:
[#2738](https://github.com/beyondessential/tupaia-backlog/issues/2738)

### Change
Just one miragion, other files are merged from PR #2823.


packages/database/src/migrations/20210714053110-FirstNationWithILIGraph-modifies-data.js
```diff
@@ -0,0 +1,106 @@
...
+          '($row.period).substring(0, 4)': '$row.value',
+        },
+        {
+          transform: 'aggregate',
+          name: 'group',
+          '...': 'last',
+        },
+        {
+          transform: 'select',
+          "'timestamp'": "periodToTimestamp(('2020').concat($row.name))",
```
billli0, line 51: Hackly adding a timestamp, so that we can order the xAxis.

packages/database/src/migrations/20210714053110-FirstNationWithILIGraph-modifies-data.js
```diff
@@ -0,0 +1,106 @@
...
+          '...': '*',
+        },
+      ],
+    },
+    permission_group_id: await permissionGroupNameToId(db, 'Public'),
+  });
+
+  await insertObject(db, 'dashboard_item', {
+    id: dashboardItemId,
+    code: 'AU_FLUTRACKING_First_Nation_With_ILI',
```
billli0, line 61: Every time it comes to my mind that if we can set a report_code as null in `dashboard_item` and use its own code as `report_code`, because they are mostly the same.

packages/database/src/migrations/20210714053110-FirstNationWithILIGraph-modifies-data.js
nick-beyondessential: Defaulting to name order if no timestamp sounds like a good idea.
So 2020 is just an arbitrary date to create a weekly order?

nick-beyondessential (approved):
LGTM @billli0 Good work!.
Couldn't see any explicit conflicts but might be a good idea to keep and eye on @rohan-bes 's
https://github.com/beyondessential/tupaia/pull/2723

packages/database/src/migrations/20210714053110-FirstNationWithILIGraph-modifies-data.js
billli0: @nick-beyondessential That's right, yeah I incline that we don't always need to put a timestamp here just for the sorting, I would say that if they need a customised sorting they should do that like `stackOrder` in `chartConfig`.

PR #2844, opened 2021-07-16
Allow custom db name in refresh script

kael89:
Helps if you have multiple tupaia dbs locally

Usage example: `yarn refresh-database dump.sql -d tupaia2`


.gitignore
```diff
@@ -10,6 +10,7 @@
 *.log*
 .nyc_output
 *.aes
+*.code-workspace
```
rohan-bes, line 13: What's this doing?

scripts/node/refreshDatabase.js
```diff
@@ -31,11 +36,12 @@ class RefreshDatabaseScript extends Script {
...
-    this.execDbCommand('DROP DATABASE IF EXISTS tupaia');
-    this.execDbCommand('CREATE DATABASE tupaia WITH OWNER tupaia');
-    this.execDbCommand('CREATE EXTENSION postgis', { db: 'tupaia' });
+    const dbName = this.args.database || 'tupaia';
+    this.execDbCommand(`DROP DATABASE IF EXISTS ${dbName}`);
+    this.execDbCommand(`CREATE DATABASE ${dbName} WITH OWNER tupaia`);
+    this.execDbCommand('CREATE EXTENSION postgis', { db: dbName });
     this.execDbCommand('ALTER USER tupaia WITH SUPERUSER');
-    this.exec(`psql -U tupaia -f "${this.args.dumpPath}"`); // Relative to the repo root
+    this.exec(`psql -U tupaia -d ${dbName} -f "${this.args.dumpPath}"`); // Relative to the repo root
```
rohan-bes, line 44: In some ways I feel it may be better for us to just add a `.env` file to the root directory with our standard db env vars (`DB_NAME`, `DB_USER`, etc.), thoughts on that? But also happy with this as well 👍

rohan-bes (approved):
Couple of comments but nothing major

.gitignore
kael89: Oh sorry I should have added a comment, this allows people to use custom local workspace files, without seeing them in the repo diff.

I maintain two clones of the tupaia repo locally (`tupaia` and `tupaia2`). I use the default workspace file for the first (`tupaia-packages.code-workspace`), and I have added a `tupaia2.code-workspace` file in `tupaia2` so I can quickly toggle between them in VSCode 😄

scripts/node/refreshDatabase.js
kael89: I also thought that we could use the .env file from the database package, since most probably the same value will be used here and there. But then I wasn't sure if this script should depend in any information from that package - although now that you mentioned, perhaps this script should live in the `database` package in the first place!

kael89: How do you feel about the following?
* Move the `refreshDatabase.js`/`dumpDb.sh` scripts under the `database` package
* Make the first read the `.env` files from there to determine the default db name (instead of hard coding them to `tupaia`)
* Maintain support for the existing `yarn` scripts in the root level - they can just call the `database` scripts

rohan-bes: That sounds great!

packages/utils/src/Script.js
```diff
@@ -1,18 +1,47 @@
...
+ * @property {string} [usage]
+ * @property {string} [version]
+ *
+ * @see https://github.com/yargs/yargs/blob/master/docs/api.md
+ */
+
+/**
+ * @param {ScriptConfig} scriptConfig
+ */
+export const getArgs = scriptConfig => {
```
kael89, line 28: This function (and the JsDoc above) are just moved here from the existing `script.js` file (now renamed to runScripts.js`)

kael89 (commented):
@rohan-bes I made the changes we discussed - now the refresh database script will read the db name and user from the `.env` file under `packages/database`

PR #2848, opened 2021-07-16
Ensure volumes are deleted on instance termination

edmofro:
I noticed there were a heap of elastic block store volumes just hanging around. Turns out the instances we start have the behaviour for volumes "DeleteOnTermination" set to false by default. This changes that.


IgorNadj (approved):

edmofro:
Already tested, will merge in and manually deploy to lambda

PR #2850, opened 2021-07-19
Fix tests

kael89:
Addresses the following issues:

* Fix flaky indicator tests by running them serially (was able to reproduce this locally)
* Extend from centralised jest config, defined in the project root
* Fix jest tests timing out locally
* Always build TS types for JS packages that are configured for type building
* .other small fixes


jest.config-js.json
```diff
@@ -0,0 +1,13 @@
...
+  "collectCoverageFrom": ["<rootDir>/src/**/*.js)"],
+  "coveragePathIgnorePatterns": ["__tests__"],
+  "testMatch": ["<rootDir>/src/__tests__/**/**.test.js"],
+  "setupFilesAfterEnv": ["../../jest.setup.js"],
+  "transformIgnorePatterns": ["/node_modules/"],
+  "transform": {
+    "^.+\\.jsx?$": "../../jestTransformer.js"
+  },
+  "testEnvironment": "node",
+  "testTimeout": 100000
```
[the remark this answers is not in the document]
kael89: Increasing the `testTimeout` fixes category **1** issues (see above comment). It's not an ideal solution, but:
* We do something similar for [mocha tests](https://github.com/beyondessential/tupaia/blob/dev/.mocharc.js#L9), and it also happens in Tamanu (`30 secs` used there)
* A more "proper" solution which would involve refactoring DB instantiation logic does not seem straightforward. Furthermore, the issue may be caused by `jest` logic and may not be something that is under our control. I feel it's not worth me spending more time on this, but happy for someone else to have a look if they want

packages/auth/src/__tests__/buildLegacyAccessPolicy.test.js
```diff
@@ -133,89 +133,89 @@ describe('buildLegacyAccessPolicy', () => {
       const user = await upsertDummyRecord(models.user);
 
       // Create a facility nested deep within a new country
-      const canada = await findOrCreateDummyRecord(
+      const greece = await findOrCreateDummyRecord(
```
kael89, line 136: We also insert Canada entities in [testAccessPolicyHandler.js](https://github.com/beyondessential/tupaia/blob/dev/packages/auth/src/__tests__/buildAccessPolicy/testAccessPolicyHandler.js#L18). The entity existence check there may happen concurrently with the one that is triggered here. In that case both checks will fail, and both tests will attempt to create the same entity in the DB.

packages/psss/src/__tests__/period.test.js
```diff
@@ -36,25 +38,23 @@ describe('period', () => {
   });
 
   it('Gets Days Till Due Day', async () => {
-    const sunday = new Date(1606561200000); // Sun 29th November 2020
-    const monday = new Date(1606647600000); // Mon 30th November 2020
-    const tuesday = new Date(1606734000000); // Tuesday 1st December 2020
-    const wednesday = new Date(1606820400000); // Wednesday 2nd December 2020
+    const sunday = '2020-11-29';
```
kael89, line 41: Tests would pass in NZ but not in Melbourne

scripts/bash/buildInternalDependencies.sh
```diff
@@ -17,10 +17,6 @@ while [ "$1" != "" ]; do
         shift
         watch=true
         ;;
-    --withTypes)
```
kael89, line 20: Currently we don't build TS types by default for JS packages that support them (eg `aggregator`). This is for historical reasons: when we first added this feature, TS was new to us and we didn't want to risk breaking existing functionality just to support it.

I feel that now is a good time to support TS types by default, but happy to re-consider

jest.config-ts.json
```diff
@@ -2,6 +2,9 @@
   "clearMocks": true,
   "collectCoverageFrom": ["<rootDir>/src/**/*.ts"],
   "coveragePathIgnorePatterns": ["__tests__"],
+  "testMatch": ["<rootDir>/src/__tests__/**/**.test.ts"],
   "setupFilesAfterEnv": ["../../jest.setup.js"],
-  "testMatch": ["<rootDir>/src/__tests__/**/*.test.ts"]
+  "transformIgnorePatterns": ["/node_modules/"],
+  "testEnvironment": "node",
+  "testTimeout": 30000
```
kael89, line 9: This comment has been updated to take into account the latest information collected by test runs. Originally I had the suspicion that tests may not run in parallel in our CI/CD, but it seems that they do

One of jest's selling points is it's ability to run tests in parallel. However, this may cause concurrency issues when using the same database for all tests.

Ideally we should

use a different database per test run thread
do not use the local dev database for the tests, but rather use an empty db (perhaps pre-populated with test-specific data)
instead of executing expensive delete statements to clean up the db, truncate tables between tests when possible
Until the above are implemented, we need to update the test timeout. I had done a bit of investigation in the past; my conclusion was that timeouts are probably related to acquiring DB connections, but I didn't figure out the exact mechanism. We also use a (more generous - 100 secs) test timeout in mocha tests. It seems that this is an old config but it did not make its way to jest when we migrated the tests. This config is also used in Tamanu (30 secs used there)

rohan-bes: Good point @kael89 about a more correct mechanism for using a db while testing! My concerns might be that it may become tricky to keep the testDB and actualDB in sync (in our local).

I suppose another option is for us to 'unit'-y in our unit tests. I don't really like this suggestion of mine, as I personally feel we get a lot more bang for our buck with these 'integration test style unit tests'. However at the same time they don't exactly feel like the best fit for jest? Keen to hear your thoughts on this!

packages/indicators/package.json
```diff
@@ -14,9 +14,8 @@
     "build": "babel src --out-dir dist --ignore \"src/__tests__/**\" --extensions .ts --config-file \"../../.babelrc-ts.js\"",
     "lint": "tsc; eslint . --ext .ts",
     "lint:fix": "yarn lint --fix",
-    "test": "jest",
-    "test:coverage": "jest --coverage",
-    "test:watch": "jest --watch"
+    "test": "jest --runInBand",
```
rohan-bes, line 17: This `runInBand` config basically runs the tests in serial rather than parallel right? You'd recommend we use this wherever the tests use the database?

scripts/bash/buildInternalDependencies.sh
rohan-bes: I love this change! Having to manually rebuild the types each time after running `yarn` was starting to get tedious 😝

packages/psss/src/__tests__/period.test.js
rohan-bes: Great catch!

rohan-bes (approved):
Nice work! Can't wait to start writing more tests and running them in local! Lately I've been trying to get into TDD but test flakiness was making it rather tricky!

Might wanna touch base with you about what the plan is long term? It seems like our workaround for now is to run jest tests in serial if the package's tests use the database? Perhaps we wanna start breaking our tests into integration (serial) and unit (parallel) and running them independently? Happy to discuss other options too!

packages/indicators/package.json
kael89: Exactly! This has been giving us some pain since we migrate from mocha to jest. I made the change in packages that cause us trouble for now, but perhaps we should proactively update the others... What do you think?

Another idea:
* Explicitly separate integration (using an actual DB) and unit tests in our code base. Either use different folders, or different file extensions
* Run unit tests in parallel and integration tests serially (for now)
* In the future, enabling parallel integration testing by using a new DB for each testing thread. I'm not familiar with this pattern but I briefly read about it yesterday

I think it's high time we don't use a local DB for running the tests:
1. The DB in our CI/CD starts with a clean state, so the local and CI/CD test dbs do not match
2. We have some elaborate data cleaning mechanism to reset the DB to the same state it was before the tests started running. It would make things simpler to start with a "clean" (or populated with test setup data) DB. It would also make all DB operations in the tests faster

Tagging you too here @edmofro so that you are aware of these findings and thoughts

kael89: (Relevant comment above: https://github.com/beyondessential/tupaia/pull/2850#discussion_r672276857)

jest.config-ts.json
kael89: I also thought the same, but was a bit sceptical that such a popular framework like `jest` does not properly support integration tests by default. But then it dawned to me - we are probably doing DB testing wrong!

I believe that Tamanu does use a different DB per thread pattern - if my understanding is correct, when the tests run there are 3 local SQLite test dbs

packages/indicators/package.json
edmofro: This is all so true! ​We should absolutely have a separate test db. I believe McLean even tried to set this up at one point, but ran into enough obstacles that he backed out. But yes, for all of those reasons you say, it'd be ideal to have a separate db spun up for tests, and even better a separate one for each thread!

Note that in order to make tests independent, we'd still need either the same sort of data cleaning protocols, a drop and restore, or tests that only make db changes that other tests can live with (difficult to be certain). So long as you're using the same database across more than one test (e.g. per thread), you need to make sure earlier tests don't interfere with it. 

Final thing - in Tamanu they use SQLite for testing which makes things fast and lightweight, but does lead to a) some additional burden maintaining SQLite support and b) some discrepancies where a test may show a false negative or false positive because there is a slight behaviour difference between SQLite and Postgresql. For those reasons I'd like to keep our test db using postgres.

Could you make a card for this @kael89?

kael89: I agree with using PostgreSQL in out test db (I also came across this discrepancies), also we should certainly clean up the test DB between test runs! But using a clean DB as a starting point will make this easier (eg you can truncate some tables, which is much faster than deleting selected records)

Card created: https://github.com/beyondessential/tupaia-backlog/issues/3210

kael89:
@rohan-bes I will spin up a new feature instance, to make sure that everything works well after the build/test changes here. I will do a smoke test in our apps, if everything looks fine I'll merge straight into `dev`

PR #2852, opened 2021-07-19
3032: correct number of teachers report config

edmofro:
### Issue #:

Addresses this comment: https://github.com/beyondessential/tupaia-backlog/issues/3032#issuecomment-878915524

I had a look around and wasn't able to find any other situations where the data wasn't being summed up the levels, so hopefully this is the only instance.

Thanks @rohan-bes for identifying this existing bug as part of regression testing your changes!


packages/database/src/migrations/20210719023634-CorrectNumberTeachersByEducationLevel-modifies-data.js
```diff
@@ -0,0 +1,177 @@
...
+      "'name'":
+        "translate($row.dataElement, { noteach_pe_public_f: 'Primary Education (Public)', noteach_pe_public_m: 'Primary Education (Public)', noteach_pe_private_f: 'Primary Education (Private)', noteach_pe_private_m: 'Primary Education (Private)', noteach_se_public_f: 'Secondary Education (Public)', noteach_se_public_m: 'Secondary Education (Public)', noteach_se_private_f: 'Secondary Education (Private)', noteach_se_private_m: 'Secondary Education (Private)' })",
+      '...': ['value', 'dataElement'],
+    },
+    'keyValueByDataElementName',
+    {
+      name: 'group',
+      transform: 'aggregate',
+      '...': 'sum',
+    },
```
edmofro, line 54: You really don't need to review this all - only the first and third transform steps have changed!

Before it was doing a `"last"` when aggregating and grouping, which is obviously wrong when we want to sum up the levels. We can't sum the org unit and period, but don't actually need them, so the first transform step now drops them from the start.

EMcQ-BES (approved):
Looks good to me

tcaiger:
@edmofro - do you mind if I merge this into dev?

PR #2854, opened 2021-07-19
3102 viz builder fe scaffolding

tcaiger:
### Issue #: [3102](https://github.com/beyondessential/tupaia-backlog/issues/3102) Viz Builder Frontend Scaffolding

The main design considerations are: 
- vizBuilder code should be encapsulated so that it can be exported and imported into Lesmis later this year
- vizBuilder should be seperate to admin-panel so that we can use modern react techniques such as react-query

### Changes:

- Added top level route for viz builder app
- Created viz builder app with placeholder components and seperate providers for theme etc
- Added user query to admin-panel server to get user details and determine if user is logged in and has permission

---


packages/admin-panel/src/api/TupaiaApi.js
```diff
@@ -10,18 +10,6 @@ import { verifyResponseStatus, stringifyQuery } from '@tupaia/utils';
 const FETCH_TIMEOUT = 45 * 1000; // 45 seconds in milliseconds
 
 export class TupaiaApi {
-  constructor() {
```
tcaiger, line 13: The store is no longer being used in this class.

packages/admin-panel/src/vizBuilder/components/Toolbar.js
```diff
@@ -0,0 +1,55 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
```
tcaiger, line 5: This is just a placeholder file until the component is fully developed

packages/admin-panel/src/vizBuilder/components/Tabs.js
```diff
@@ -0,0 +1,16 @@
+/*
+ * Tupaia
```
tcaiger, line 2: This is just a placeholder file until the component is fully developed

packages/admin-panel/src/vizBuilder/components/JsonEditor.js
```diff
@@ -0,0 +1,16 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
```
tcaiger, line 4: This is just a placeholder file until the component is fully developed

packages/admin-panel/src/vizBuilder/api/mutations/useSaveViz.js
```diff
@@ -0,0 +1,19 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 20211Beyond Essential Systems Pty Ltd
+ *
+ */
```
tcaiger, line 5: This is just a placeholder file until the component is fully developed

packages/admin-panel/src/authentication/ProfileButton.js
```diff
@@ -3,45 +3,30 @@
...
+  isBESAdmin: PropTypes.bool,
 };
 
-export const ProfileButton = connect(
-  state => ({
-    user: getUser(state),
-  }),
-  null,
-)(ProfileButtonComponent);
+ProfileButton.defaultProps = {
```
tcaiger, line 30: I converted the profile button component to be unaware of redux so that it can be re-used in the VizBuilder

packages/admin-panel/src/index.js
```diff
@@ -1,33 +1,34 @@
...
-import { PersistGate } from 'redux-persist/lib/integration/react';
 import 'react-table/react-table.css';
+import AdminPanel from './App';
+import { AdminPanelProviders, VizBuilderProviders } from './utilities';
+import { Footer, Navbar } from './widgets';
 
-import { theme } from './theme';
-import { store, persistor } from './store';
-import { App } from './App';
+const VizBuilder = lazy(() => import('./vizBuilder'));
```
tcaiger, line 13: Lazy loading the viz builder means that webpack will split it into its own bundle and only serve it to the user when they hit the `viz-builder` route

packages/admin-panel/src/index.js
```diff
@@ -1,33 +1,34 @@
...
-          </ThemeProvider>
-        </MuiThemeProvider>
-      </StylesProvider>
-    </PersistGate>
-  </Provider>,
+  <Router>
+    <Suspense fallback={<div>loading...</div>}>
+      <Switch>
+        <Route path="/viz-builder" exact>
+          <VizBuilderProviders>
```
tcaiger, line 20: Viz Builder has its own providers so that there is no cross pollination between it and admin panel so that:
- it can be exported as a standalone app
- hot-reloading works
- can have a different theme if need be
- can easily add other providers like react-query provider etc
- better code splitting

packages/admin-panel/src/pages/LogoutPage.js
```diff
@@ -0,0 +1,40 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import React, { useEffect } from 'react';
```
tcaiger, line 6: I added a logout page so that the VizBuilder could have a logout button

packages/admin-panel/src/rootReducer.js
```diff
@@ -2,23 +2,22 @@
...
-const appReducer = combineReducers({
-  authentication,
-  tables,
-  autocomplete,
-  editor,
-  dataChangeListener,
-});
-
 export const rootReducer = (state, action) => {
+  const appReducer = combineReducers({
```
tcaiger, line 13: I had to move the combineReducers inside the root reducer as I noticed a race condition where the sub-reducers hadn't finished runnning when the rootReducer ran

packages/admin-panel/src/utilities/createNestedReducer.js
```diff
@@ -29,5 +29,5 @@ export const createNestedReducer = (defaultNestedState, nestedStateChanges) => {
   Object.entries(nestedStateChanges).forEach(([actionType, actionHandler]) => {
     stateChanges[actionType] = createNestedStateChanger(defaultNestedState, actionHandler);
   });
-  return createReducer({}, stateChanges);
+  return createReducer(defaultNestedState, stateChanges);
```
tcaiger, line 32: This fixes an existing bug where the default state of the reducers was net setup for createdNestedReducer types.

packages/admin-panel/src/VizBuilderApp/api/mutations/useSaveViz.js
```diff
@@ -0,0 +1,19 @@
+/*
+ * Tupaia
```
tcaiger, line 2: This is just a placeholder file until the component is fully developed

packages/admin-panel/src/VizBuilderApp/components/JsonEditor.js
```diff
@@ -0,0 +1,16 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
```
tcaiger, line 3: This is just a placeholder file until the component is fully developed

packages/admin-panel/src/VizBuilderApp/components/Tabs.js
```diff
@@ -0,0 +1,16 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
```
tcaiger, line 4: This is just a placeholder file until the component is fully developed

packages/admin-panel/src/VizBuilderApp/components/Toolbar.js
```diff
@@ -0,0 +1,55 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
```
tcaiger, line 5: This is just a placeholder file until the component is fully developed

packages/admin-panel/src/VizBuilderApp/api/fetchError.js
```diff
@@ -0,0 +1,20 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ *
+ */
+export default class FetchError extends Error {
+  constructor(message, code, ...params) {
+    // Pass remaining arguments (including vendor specific ones) to parent constructor
+    super([message, ...params]);
```
kael89, line 9: I see that this is also used in `lesmis`, does this work? I would expect the following syntax:

```suggestion
    super(message, ...params);
```

packages/admin-panel/src/VizBuilderApp/api/queries/useUser.js
```diff
@@ -0,0 +1,22 @@
...
+    staleTime: 1000 * 60 * 60 * 1,
+    refetchOnWindowFocus: false,
+    ...options,
+  });
+
+  const user = query.data;
+  const isLoggedIn = user && Object.keys(user).length > 0;
+  const isBESAdmin = user && user.isBESAdmin;
+
+  return { ...query, isLoggedIn, isBESAdmin };
```
kael89, line 21: One suggested refactor is to keep the `isBESAdmin` prop in the `user` object, since we generally pass both of them in the same components. That's not just for convenience but it probably makes sense conceptually

kael89 (approved):
Awesome work @tcaiger ! A left a few comments but they are all minor - pre-approving 🎉

packages/admin-panel/src/VizBuilderApp/components/Flexbox.js
```diff
@@ -0,0 +1,36 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import styled from 'styled-components';
+import MuiBox from '@material-ui/core/Box';
+
+export const FlexStart = styled(MuiBox)`
```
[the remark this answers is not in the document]
tcaiger: Yes good suggestion, I have moved them to ui components (most of them already existed)

packages/admin-panel/src/VizBuilderApp/components/FullPageLoader.js
```diff
@@ -0,0 +1,19 @@
...
+import styled from 'styled-components';
+import CircularProgress from '@material-ui/core/CircularProgress';
+import { FlexCenter } from './Flexbox';
+
+const LoadingContainer = styled(FlexCenter)`
+  height: 100vh;
+  padding-bottom: 10%;
+`;
+
+export const FullPageLoader = () => (
```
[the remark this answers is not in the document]
tcaiger: Yes agreed this could be useful elsewhere!

packages/admin-panel/src/VizBuilderApp/api/fetchError.js
tcaiger: Yep this is correct the api is that the constructor takes an array!

packages/admin-panel/src/VizBuilderApp/api/queries/useUser.js
tcaiger: I'm don't feel too strongly about it but it is a bit more convenient to have isBESAdmin outside the user object to check for access before showing user details. It's kind of user meta data so I think it's ok to be seperate.

packages/admin-panel/src/VizBuilderApp/api/fetchError.js
kael89: This is the JS Error class right? I checked the docs and also tried this locally but couldn't get it working with an array...

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error - "ES6 Custom Error Class"

packages/admin-panel/src/VizBuilderApp/api/queries/useUser.js
kael89: I see what you mean - `user` may be `undefined`... Fair enough 🙂

PR #2855, opened 2021-07-19
3208 exporting hotfix

IgorNadj:


packages/web-config-server/src/export/buildExportUrl.js
```diff
@@ -6,4 +6,4 @@
 import { stringifyQuery } from '@tupaia/utils';
 
 export const buildExportUrl = (req, resource, queryParams) =>
-  stringifyQuery(`//${req.headers.host}${req.baseUrl}`, `export/${resource}`, queryParams);
+  stringifyQuery('', `export/${resource}`, queryParams);
```
rohan-bes, line 9: So URL is now just `export/${resource}`? I suppose this makes the `req` argument redundant but I guess we don't wanna remove it now and increase the scope of the refactor?

As long as it's noted that we need to clean this up 👍

rohan-bes (approved):
Nice, LGTM!

packages/web-config-server/src/export/buildExportUrl.js
IgorNadj: Cleaned up, good spotting

rohan-bes (approved):

PR #2856, opened 2021-07-19
3195 report server insert row function

EMcQ-BES:
### Changes:

- Add an `insert` transform following a similar pattern to `select`
- Add unit tests for insert

### DIFF
--- a/packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
@@ -1,0 +1,124 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { SINGLE_ANALYTIC, MULTIPLE_ANALYTICS } from './transform.fixtures';
+import { buildTransform } from '../../../reportBuilder/transform';
+
+describe('insert', () => {
+  // SAME AS SELECT FUNCTIONALITY
+  it('can insert row with basic values', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        "'number'": '1',
+        "'string'": "'Hi'",
+        "'boolean'": 'false',
+      },
+    ]);
+    expect(transform(SINGLE_ANALYTIC)).toEqual([...SINGLE_ANALYTIC, { number: 1, string: 'Hi', boolean: false }]);
+  });
+
+  it('can insert row with values from previous row', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        "'dataElementValue'": '$row.value',
+      },
+    ]);
+    expect(transform(SINGLE_ANALYTIC)).toEqual([...SINGLE_ANALYTIC, { dataElementValue: 4 }]);
+  });
+
+  it('can select a value from the row as a field name', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        '$row.dataElement': '$row.value',
+      },
+    ]);
+    expect(transform(SINGLE_ANALYTIC)).toEqual([...SINGLE_ANALYTIC, { BCD1: 4 }]);
+  });
+
+  it('can execute functions', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        "'period'": "periodToDisplayString($row.period, 'DAY')",
+      },
+    ]);
+    expect(transform(SINGLE_ANALYTIC)).toEqual([...SINGLE_ANALYTIC, { period: '1st Jan 2020' }]);
+  });
+
+  // SPECIFIC TO INSERT
+  it('can insert multiple rows', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        "'dataElementValue'": '$row.value',
+      },
+    ]);
+    expect(transform(MULTIPLE_ANALYTICS)).toEqual([
+      { period: '20200101', organisationUnit: 'TO', dataElement: 'BCD1', value: 4 },
+      { dataElementValue: 4 },
+      { period: '20200102', organisationUnit: 'TO', dataElement: 'BCD1', value: 2 },
+      { dataElementValue: 2 },
+      { period: '20200103', organisationUnit: 'TO', dataElement: 'BCD1', value: 5 },
+      { dataElementValue: 5 },
+    ]);
+  });
+
+  it('can insert new rows before the relative row', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        "'dataElementValue'": '$row.value',
+        position: 'before',
+      },
+    ]);
+    expect(transform(MULTIPLE_ANALYTICS)).toEqual([
+      { dataElementValue: 4 },
+      { period: '20200101', organisationUnit: 'TO', dataElement: 'BCD1', value: 4 },
+      { dataElementValue: 2 },
+      { period: '20200102', organisationUnit: 'TO', dataElement: 'BCD1', value: 2 },
+      { dataElementValue: 5 },
+      { period: '20200103', organisationUnit: 'TO', dataElement: 'BCD1', value: 5 },
+    ]);
+  });
+
+  it('can insert new rows at the beginning of the list', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        "'dataElementValue'": '$row.value',
+        position: 'start',
+      },
+    ]);
+    expect(transform(MULTIPLE_ANALYTICS)).toEqual([
+      { dataElementValue: 4 },
+      { dataElementValue: 2 },
+      { dataElementValue: 5 },
+      { period: '20200101', organisationUnit: 'TO', dataElement: 'BCD1', value: 4 },
+      { period: '20200102', organisationUnit: 'TO', dataElement: 'BCD1', value: 2 },
+      { period: '20200103', organisationUnit: 'TO', dataElement: 'BCD1', value: 5 },
+    ]);
+  });
+
+  it('can insert specific new rows using a where clause', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        "'dataElementValue'": '$row.value',
+        position: 'after',
+        where: "not(eq($row.period, '20200101'))",
+      },
+    ]);
+    expect(transform(MULTIPLE_ANALYTICS)).toEqual([
+      { period: '20200101', organisationUnit: 'TO', dataElement: 'BCD1', value: 4 },
+      { period: '20200102', organisationUnit: 'TO', dataElement: 'BCD1', value: 2 },
+      { dataElementValue: 2 },
+      { period: '20200103', organisationUnit: 'TO', dataElement: 'BCD1', value: 5 },
+      { dataElementValue: 5 },
+    ]);
+  });
+});
--- a/packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
@@ -180,16 +180,10 @@ describe('insert', () => {
     const transform = buildTransform([
       {
         transform: 'insert',
-        where: 'and(exists($next.organisationUnit), not(eq($row.organisationUnit, $next.organisationUnit)))',
-        "'Total_BCD1'": "sum($allPrevious.BCD1)",
-        "'Total_BCD2'": "sum($allPrevious.BCD2)",
+        where: 'not(eq($row.organisationUnit, $next.organisationUnit))',
+        "'Total_BCD1'": "sum($where(f($otherRow) = equalText($otherRow.organisationUnit, $row.organisationUnit)).BCD1)",
+        "'Total_BCD2'": "sum($where(f($otherRow) = equalText($otherRow.organisationUnit, $row.organisationUnit)).BCD2)",
       },
-      {
-        transform: 'insert',
-        where: 'eq($index, length($table))',
-        "'Total_BCD1'": "sum($all.BCD1) - sum($allPrevious.Total_BCD1)",
-        "'Total_BCD2'": "sum($all.BCD2) - sum($allPrevious.Total_BCD2)",
-      }
     ]);
     expect(transform(AGGREGATEABLE_ANALYTICS)).toEqual([
       { period: '20200101', organisationUnit: 'TO', BCD1: 4 },
--- a/packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
@@ -3,7 +3,7 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { SINGLE_ANALYTIC, MULTIPLE_ANALYTICS } from './transform.fixtures';
+import { SINGLE_ANALYTIC, MULTIPLE_ANALYTICS, AGGREGATEABLE_ANALYTICS } from './transform.fixtures';
 import { buildTransform } from '../../../reportBuilder/transform';
 
 describe('insert', () => {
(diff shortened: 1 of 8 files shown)


packages/report-server/src/reportBuilder/transform/functions/insert.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+
+    parser.next();
+    return newRow;
+  });
+  let insertCount = 0;
+  rowsToInsert.forEach((newRow, index) => {
+    if (newRow !== undefined) {
+      returnArray.splice(params.position(index, insertCount), 0, newRow);
+      insertCount++;
```
EMcQ-BES, line 40: Very open to other options for how to do this interlacing of arrays

packages/report-server/src/reportBuilder/transform/functions/position.ts
```diff
@@ -0,0 +1,33 @@
...
+
+type PositionParams = {
+  position?: string;
+};
+
+// Calculate which position to splice a new row into an array
+const positionAliases = {
+  before: (index: number, insertCount: number) => index + insertCount,
+  after: (index: number, insertCount: number) => index + insertCount + 1,
+  start: (index: number, insertCount: number) => insertCount,
```
EMcQ-BES, line 14: I don't know if `start` is necessarily useful, but it's basically free with the current implementation

edmofro (changes requested):
Thanks for fixing that stuff we discussed!

All looks great, I'd just like to suggest three more test cases:
- can sum all values in an inserted totals row at the end of other rows
- can insert a row using a condition that compares the relative row with the one above (to support a row at the end of each category)
- can sum values for a subset of rows using where (to support category totals)

I believe all should be supported, but if any aren't just add the test case and comment it out with a note that it needs more work to support

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        position: 'before',
+      },
+    ]);
+    expect(transform(MULTIPLE_ANALYTICS)).toEqual([
+      { dataElementValue: 4 },
+      { period: '20200101', organisationUnit: 'TO', dataElement: 'BCD1', value: 4 },
+      { dataElementValue: 2 },
+      { period: '20200102', organisationUnit: 'TO', dataElement: 'BCD1', value: 2 },
+      { dataElementValue: 5 },
+      { period: '20200103', organisationUnit: 'TO', dataElement: 'BCD1', value: 5 },
```
rohan-bes, line 85: Apologies for jumping in on this review, I was super curious! 🙏 

It's looking really good, and I very much appreciate the tests!!

Wanted to note that I'm a bit surprised that the default behaviour is to insert a row for each existing row if no `where` is specified. 

I do see the logic, and it's consistent with our other existing transforms that basically treat having no `where` clause as saying 'match every row'. 

However I feel it's a bit unexpected, I guess I had assumed that by default we'd just insert a single row (at the end by default)? And we'd only do the 'insert per matching row' if a `where` clause is specified?

Curious to hear your thoughts on this?

EMcQ-BES: @rohan-bes I looked at doing that to start with but I didn't like how inconsistent it felt.

Things like what does `position: 'before'` mean if there's no `where` clause? What happens if you reference `$row` when there's no `where` clause?

It winds up sorta being two separate functions depending on whether you have a `where`, and that started feeling a little messy in code.

I agree that the default behaviour is perhaps a bit unexpected, I can certainly change it to work like you've mentioned, or perhaps I could split the functionality into single/multi insert.

What would you prefer?

rohan-bes: No I do see your point... I think I could be won over pretty easily to the current implementation if there were an easy way to insert a single row. Perhaps something like `where: $row.index = 1` (single row at start) and `where: $row.index = $table.length` (where table is some lookup I'd suggest that is just an array of all rows, single row at end).

I'm not suggesting we do this now btw, if the transform you've got here meets all your requirements then that's all we need 😄 Just wanna make sure we have some kind of plan if we feel we need a single row insert behaviour...

EMcQ-BES: Cool, I've added a couple of extra lookup, `$index`, `$previous` and `$table` just because they're all useful for random things we might wanna do with insertions

I also added `length(Array)` as function because there was some weirdness with the parsing of `Array.length` format in reports.

And I've added some extra unit tests showing the single row at start, single row at end.

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    expect(transform(SINGLE_ANALYTIC)).toEqual([...SINGLE_ANALYTIC, { period: '1st Jan 2020' }]);
+  });
+
+  // SPECIFIC TO INSERT
+  it('can insert a single row at start', () => {
+    const transform = buildTransform([
+      {
+        transform: 'insert',
+        position: 'before',
+        where: 'eq($index, 0)',
```
rohan-bes, line 59: How would you feel about us using a 1-based index rather than a 0-based index? I know it goes against our creed, but I'm partial towards it since we plan on displaying the row# in the data table preview in the upcoming Admin Panel Viz Builder: 
image

Probably not a decision we need to make right away tho 👍

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
rohan-bes: Ah awesome! Thanks for adding those lookups, I feel they'll definitely come in handy!

packages/report-server/src/reportBuilder/functions/basic.ts
```diff
@@ -40,3 +40,7 @@ export const exists = (val: any): boolean => {
 export const notExists = (val: FieldValue): boolean => {
   return val === undefined;
 };
+
+export const length = (val: FieldValue[]): number => {
```
rohan-bes, line 44:
```suggestion
export const length = (val: any[]): number => {
```
Being nitpicky, but we're passing in a `Row[]` when we call `length($table)` 👍

rohan-bes (approved):
Love it ❤️ You extended the architecture perfectly and the tests are nice and thorough!

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        "'Total_BCD1'": "sum($where(f($otherRow) = eq($otherRow.organisationUnit, $previous.organisationUnit)).BCD1)",
+        "'Total_BCD2'": "sum($where(f($otherRow) = eq($otherRow.organisationUnit, $previous.organisationUnit)).BCD2)",
+      },
+      {
+        transform: 'insert',
+        where: 'eq($index, length($table) - 1)',
+        "'Total_BCD1'": 'sum($all.BCD1) - sum($all.Total_BCD1)',
+        "'Total_BCD2'": 'sum($all.BCD2) - sum($all.Total_BCD2)',
+      },
+    ]);
```
edmofro, line 194: This is crying out for a `$next`...but you've already added a heap of awesome stuff, and it is possible just using `$previous` as you've demonstrated, so happy keeping things restricted for now

edmofro (approved):
Even more powerful than last time I looked - thanks for the suggestions @rohan-bes and the implementation @EMcQ-BES! 💪 

Approved from my perspective, but if you guys want to chat about the 1-indexing option I'd be happy to jump on a quick call :-)

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
EMcQ-BES: Personally I think it would make more sense to make the front end zero indexed. 🤔

If it's visualising data it makes sense to keep it the way we use it in data. Plus it means there's less places to have an off by one error.

edmofro: I would definitely agree with you @EMcQ-BES if our audience was going to stay internal devs forever...but the idea with the viz builder is that we could open it up to non devs. Probably after the solutions team the next lot of users would be people who are savvy in Excel, and they're likely to be much more familiar with 1-indexing than 0-indexing. So from the UX perspective, I'm keen to keep the front end starting at 1

The question then is where we want to translate that, and because we're exposing the raw transform step code to the user in the same interface as the preview table, I'd probably also be in favour of making it 1-indexed there too.

If the UI was fully taking care of holding the user's hand when writing "where" statements etc., and giving them options through a drop down etc., maybe we could keep the transform step 0-indexed and make the UI do the translation, but we're a long way off having an interface that is that tight.

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
EMcQ-BES: I wrote the first insert here before I added the extra lookups. I suppose you could do a next lookup as `$table[$index + 1]` if you really wanted to

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
EMcQ-BES: Actually I might be on your side... attempting to use `$table[0]` causes an error because it's apparently one-based... I'm not sure which point of the process makes that happen but currently it means `$table[$index]` is not equal to `$row` and that definitely feels incorrect.

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
```diff
@@ -0,0 +1,214 @@
...
+        where: 'and(exists($next.organisationUnit), not(eq($row.organisationUnit, $next.organisationUnit)))',
+        "'Total_BCD1'": "sum($allPrevious.BCD1)",
+        "'Total_BCD2'": "sum($allPrevious.BCD2)",
+      },
+      {
+        transform: 'insert',
+        where: 'eq($index, length($table))',
+        "'Total_BCD1'": "sum($all.BCD1) - sum($allPrevious.Total_BCD1)",
+        "'Total_BCD2'": "sum($all.BCD2) - sum($allPrevious.Total_BCD2)",
+      }
```
edmofro, line 192: Nice work on the `$next`! I think that will come in handy, and be a thing that "just works" when someone expects that if there's a previous, there's a next.

As a more general transform step that _should_ work across any number of categories, could we try out the following?
```suggestion
      {
        transform: 'insert',
        where: 'not(eq($row.organisationUnit, $next.organisationUnit))',
        "'Total_BCD1'": "sum($where(f($otherRow) = equalText($otherRow.organisationUnitCode, $row.organisationUnitCode)).BCD1)",
        "'Total_BCD2'": "sum($where(f($otherRow) = equalText($otherRow.organisationUnitCode, $row.organisationUnitCode)).BCD2)"
      }
```

EMcQ-BES: Yep that works perfect!

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
EMcQ-BES: I've adjusted `$index` to be one-based

edmofro (approved):
Thanks for going way beyond the scope of what you needed to - excited about these improvements!

packages/report-server/src/__tests__/reportBuilder/transform/insert.test.ts
rohan-bes: Few things to note:
> Probably after the solutions team the next lot of users would be people who are savvy in Excel, and they're likely to be much more familiar with 1-indexing than 0-indexing.

This is how I see it too! And tbh even as a dev, I still enjoy 'tables' as being 1-indexed. The fact that a table is an array of rows seems a bit like an implementation detail 👍 On that note I'm somewhat keen for us to steer away from having arrays as a construct in the report-server expressions; ie. use `sum(a, b, c)` instead of `sum([a,b,c])`. This is for a couple reasons:
- Arrays are quite a programmer-y concept and probably not so familiar to our hopefully non-dev audience
- mathjs matrices and arrays aren't quite the same, and I've been caught out by bugs where I thought I was dealing with an array when in fact it was a mathjs matrix. At the same time, I don't wanna lean too heavily into using mathjs matrices wherever we might want to use arrays as they seem inappropriate much of the time, so best to avoid all together?

Regarding that point, I absolutely love your `length()` function @EMcQ-BES! It reminds me of python's `len()` function which worked for arrays, strings, dictionaries (objects), you name it. Python was the first language I learned and was very beginner friendly! I think if we extended this function (and others too) to be polymorphic over time, it would save our users from needing to think about data structures 🎉

packages/report-server/src/reportBuilder/functions/basic.ts
```diff
@@ -40,3 +40,7 @@ export const exists = (val: any): boolean => {
 export const notExists = (val: FieldValue): boolean => {
   return val === undefined;
 };
+
+export const length = (val: any[]): number => {
+  return val.length;
+};
```
rohan-bes, line 46: You're written so many tests already, so I hate to press for even more, but could we get some unit tests for this in the function unit tests: `src/__tests__/reportBuilder/functions.test.ts`

I know it gets tests by the other higher level insert tests, but I think the idea is that all these functions in report-server may migrate into the `expression-parser` package soon enough, and if so having tests in place for all of them will make that migration a lot easier 👍

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -22,8 +22,12 @@ type RowLookup = {
  */
 type Lookups = {
   row: Row;
+  previous: Row;
+  next: Row;
   all: RowLookup;
   allPrevious: RowLookup;
+  index: number; // one-based index, this.currentRow + 1
```
rohan-bes, line 29: Curious how you guys feel about maybe moving the `index` to be a field on a row, rather than an individual lookup here. That way it would be referenced like so: `$row.index` instead of `$index`

To me it seems to have:
Pros:
- Allows for some funky lookup behaviour, eg. 
`"'Increase on past 3 weeks'": 'row.value - sum(where(otherRow =  row.index - 3 <= otherRow.index and otherRow.index < row.index)).value'` 
- Makes a bit of sense to me, as we see the 'row #' in the table in the UI preview in the viz builder, so the user is already thinking of it as a 'column' in a way (subject I know)

Cons:
- Could introduce column naming clashes... I've tried really hard to avoid these so far, since we even support dynamic column names based on data from survey results, so ensuring that a clash doesn't occur seems _**very**_ hard...

Happy with current implementation as it stands though, it meets our needs and I don't know if I've come down off the fence on this issue yet... However if either of you are swayed to go with the `$row.index` implementation perhaps it's best if we take that route now?

rohan-bes (approved):
A request for more test, and a comment that I wouldn't mind getting some opinions on, but overall super happy with what's been done here @EMcQ-BES. I love the new lookups, and with this step I feel we're getting closer to what hopefully might be a complete set of transforms required for 95% of our report-building needs 🎉

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
edmofro: I'm pretty hesitant to add anything to the row that might plausibly be used as a column name.

I see the value of having the index on the row for those past three weeks type references...but I also think that a) you could already solve it by using a transform that applies $index to each row, giving you all the powers you have above, b) we could make an alias to do point a if it was a common case, and c) we could support the same behaviour in other ways that might be even more friendly than using a `where` (like a future $useRow($index - 3) or something...)

This has made me think though...is `$index` itself quite a dev-centric term? Should we instead be using a more excel-user friendly term like `$rowNumber`?

EMcQ-BES: I think for consistency I prefer `$row.something` to always be a literal column reference. And like Edwin says, it's very easy to add index to the columns.

```
{
  transform: 'select',
  "'index'": '$index',
},
```

Also if you wanna get into the weeds `INDEX` is a function in excel to perform a lookup in a table, so they _should_ be familiar with the concept, but also this is definitely me not wanting to keep making changes 😅

edmofro: Haha fair enough, we've lumped enough changes on you already...happy to keep `$index`!

packages/report-server/src/reportBuilder/functions/basic.ts
EMcQ-BES: I've added a unit test to that file. I don't think it's really a complicated enough function to require plural tests

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
rohan-bes: Great points @edmofro and @EMcQ-BES I'm happy with keeping it as `$index` too! 

> Should we instead be using a more excel-user friendly term like $rowNumber?

I actually had this thought too when we were first coming up with this, but when I did a bit of a google it seemed like `index` was actually the most commonly used term as @EMcQ-BES mentions!

PR #2860, opened 2021-07-20
3120: admin-panel-server entities route

chris-bes:
### Changes:

- Added entities route to admin-panel-server
- Allow filter in entity-server to have operators (eg: ==, !=,...)


packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -64,42 +98,37 @@ const toFilterClause = (queryClause: string) => {
...
 
-    filter[field] = parsedValue;
+    filter[field] = convertValueToAdvancedCriteria(operator, parsedValue);
   });
 
-  filter.country_code = buildCountryCodeFilter(allowedCountries, filter.country_code);
+  // To always force returning only entities in allowed countries, even if there is country_code filter in the query params.
+  filter[QueryConjunctions.AND] = {
+    country_code: allowedCountries,
+  };
```
chris-bes, line 131: This is a bit of a hack to merge the allowedCountries filter and the optional country_code filter (especially when we want to do `country_code` `!=`) without having to do any custom logic

packages/admin-panel-server/src/app/createApp.ts
```diff
@@ -47,6 +48,7 @@ export function createApp() {
   const app = new OrchestratorApiBuilder(new TupaiaDatabase())
     .useSessionModel(AdminPanelSessionModel)
     .verifyLogin(hasTupaiaAdminPanelAccess)
+    .get('/v1/hierarchy/:hierarchyName/:entityCode', handleWith(EntitiesRoute))
```
chris-bes, line 51: We also need the root project entity code here to get all the entities of a hierarchy. I think the `/projects` endpoint in meditrak-server can supply that.

packages/admin-panel-server/src/routes/EntitiesRoute.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  constructor(req: Request, res: Response, next: NextFunction) {
+    super(req, res, next);
+
+    this.entityConnection = new EntityConnection(req.session);
+  }
+
+  async buildResponse() {
+    const { hierarchyName, entityCode } = this.req.params;
+    const { fields, search } = this.req.query;
```
chris-bes, line 22: I'm thinking we should leave the `fields` query parameter for front end to decide. The reason is in case we want to implement ancestor chain, we will need to request for extra `child_codes` field?

packages/entity-server/src/models/types.ts
```diff
@@ -6,12 +6,49 @@
...
+ * }
+ */
+type ConjunctionCriteria<T> = {
+  [field in QueryConjunctions]?: NormalCriteria<T>;
+}
+
+type FilterCriteria<T> = NormalCriteria<T> & ConjunctionCriteria<T>;
+
+export type DbConditional<T> = FilterCriteria<Omit<T, ObjectLikeKeys<T>>> &
+  FilterCriteria<Flatten<Pick<T, ObjectLikeKeys<T>>, '->>'>>;
```
chris-bes, line 51: Have to admit I'm not very confident with these types. I'm not also sure if they should sit in this file. I'm putting them in this file because they are closely related to the BaseModel

chris-bes:
Will also update examples.http in entity-server once review is done

packages/entity-server/src/models/types.ts
rohan-bes: You've actually done such an amazing job with these types! I know they're super hard to wrap your head around but I think you've got it almost perfect!

packages/entity-server/src/models/types.ts
```diff
@@ -6,12 +6,49 @@
...
+/**
+ * Example:
+ * {
+ *    _and_: {
+ *        country_code: ['TO', 'KI']
+ *    }
+ * }
+ */
+type ConjunctionCriteria<T> = {
+  [field in QueryConjunctions]?: NormalCriteria<T>;
```
rohan-bes, line 45:
```suggestion
  [field in QueryConjunctions]?: FilterCriteria<T>;
```
I _believe_ that we support recursion in this config, eg:
```
{
  name: 'TO'
  _or_: {
    type: 'facility',
    _or_: {
      parent_id: '1234',
      location_type: 'area'
    }  
  }
}
=>
WHERE code = 'TO' OR (type = 'facility' OR (parent_id = '1234' and location_type = 'area'))
```

Lucky for us, type also supports all kind of recursive insanity 🎉

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -52,10 +51,45 @@ const toJsonBKey = <T extends keyof NestedFilterQueryFields>(nestedField: T): Js
   return [field, value].join(JSONB_FIELD_DELIMITER) as JsonBKey<T>;
 };
 
+// Inspired by Google Analytics filter: https://developers.google.com/analytics/devguides/reporting/core/v3/reference?hl=en#filters
+const operatorToSqlComparator = {
+  '==': null, // Exact match
+  '!=': '!=', // Does not match
+  '=@': 'ilike', // Contains sub string
+};
```
rohan-bes, line 59: Couple things about the types here:
1. `'==': '=', // Exact match` would also work right? I have a slight preference towards that because it reads a bit more clearly to me, and because typescript doesn't let you mess around with 'falsey' values as much as JS. ie. `null` != `undefined` 
2. Currently the type of this object is:
```
{
  '==': null;
  '!=': string;
  '=@': string;
};
```
The types of the keys are string literals (https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html) whereas the values are just string (or null for the null one). This is how typescript works by default, but it does have it's drawbacks. 

For instance, later on you are trying to find the comparator for the operator by looking up in this table. Unfortunately the type will always be `string | null` at that point, which isn't really what you want. Instead you want the type to be `'!=' | 'ilike' | null` right?

Luckily there's a way to get this working in typescript as well!
```suggestion
const operatorToSqlComparator = {
  '==': '=' as const, // Exact match (switched to '=' here as it's my preference but if you stick with null theres no need for the 'as const')
  '!=': '!=' as const, // Does not match
  '=@': 'ilike' as const, // Contains sub string
};
```
This changes the type of the object to:
```
{
  '==': '=';
  '!=': '!=';
  '=@': 'ilike';
};
```
Which allows you to do typesafe lookups against this object!

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -52,10 +51,45 @@ const toJsonBKey = <T extends keyof NestedFilterQueryFields>(nestedField: T): Js
...
+const filterOperators = Object.keys(operatorToSqlComparator);
```
rohan-bes, line 60:
```suggestion
type Operator = keyof typeof operatorToSqlComparator;

const filterOperators = Object.keys(operatorToSqlComparator) as Operator[];
```
`Object.keys()` is an area that typescript kinda drops the ball imo... It always returns type of `string[]` even tho it could theoretically return a string literal array like `('==' | '!=' | '=@')[]`. Oh well, best to cast it since we know what it'll be and then our types start playing nicely again 👍

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -52,10 +51,45 @@ const toJsonBKey = <T extends keyof NestedFilterQueryFields>(nestedField: T): Js
...
+
+const formatComparisonValue = (value: string | string[], operator: string) => {
```
rohan-bes, line 62:
```suggestion
const formatComparisonValue = (value: string | string[], operator: Operator) => {
```

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -52,10 +51,45 @@ const toJsonBKey = <T extends keyof NestedFilterQueryFields>(nestedField: T): Js
...
+  if (operator === '=@' && typeof value === 'string') {
+    return `%${value}%`;
+  }
+
+  return value;
+};
+
+const convertValueToAdvancedCriteria = (operator: string, value: string | string[]) => {
```
rohan-bes, line 70:
```suggestion
const convertValueToAdvancedCriteria = (operator: Operator, value: string | string[]) => {
```
This one is important, as otherwise the lookup in the `operatorToSqlComparator` won't give us a typesafe response

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -52,10 +51,45 @@ const toJsonBKey = <T extends keyof NestedFilterQueryFields>(nestedField: T): Js
...
+
+  return {
+    comparator,
+    comparisonValue: formatComparisonValue(value, operator),
+  };
+};
+
 const toFilterClause = (queryClause: string) => {
-  const clauseParts = queryClause.split(FIELD_VALUE_DELIMITER);
+  const operator = filterOperators.find(o => queryClause.includes(o));
```
rohan-bes, line 80: Once we cast `filterOperators` to `Operator[]` then the type of `operator` here becomes `Operator | undefined`. Doing the `if(!operator)` check later on then ensures that the type is `Operator` for the rest of the codeflow 👍

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -64,42 +98,37 @@ const toFilterClause = (queryClause: string) => {
...
   const filter: Writable<NotNullValues<EntityFilter>> = {};
 
-  filterClauses.forEach(([field, value]) => {
+  filterClauses.forEach(([field, operator, value]) => {
     const parsedValue = value.includes(MULTIPLE_VALUES_DELIMITER)
       ? value.split(MULTIPLE_VALUES_DELIMITER)
       : value;
 
-    filter[field] = parsedValue;
+    filter[field] = convertValueToAdvancedCriteria(operator, parsedValue);
```
rohan-bes, line 125: When I checked out this branch, it was originally showing a pretty complex typescript error here. This is the downside of dealing with complex nested generic types as we are doing here, the compiler isn't great at explaining what went wrong... However I think the real issue is that we weren't being typesafe with the `comparator` field, which the `EntityFilter` type expected to have the explicit `comparator: '!=' | 'ilike'`, whereas we were giving it `comparator: string | null`. I think if you follow all the suggestions I made here, then it should clear up the error (atleast it has in my local). Let me know if it doesn't and we can work on it together 👍

packages/entity-server/src/models/types.ts
```diff
@@ -6,12 +6,49 @@
 import { DatabaseModel, DatabaseType } from '@tupaia/database';
 import { ObjectLikeKeys, Flatten } from '../types';
 
-type PartialOrArray<T> = {
-  [field in keyof T]?: T[field] | T[field][];
+type FilterComparators = '!=' | 'ilike';
```
rohan-bes, line 9:
```suggestion
type FilterComparators = '!=' | 'ilike' | '=';
```
See comments about this in `filter.ts`

rohan-bes (changes requested):
Looks amazing @chris-bes, nice work!

Totally happy with the functionality, just suggestions on how to resolve the ts errors and ensure the types match the model.

So nice to see someone else getting stuck into writing some pretty complex types, I think you did a great job!

rohan-bes:
Oh sorry @chris-bes one last request! Could you possibly break this PR into?
- Changes to entity-server/aggregator
- Changes to admin-panel-server

I'm totally cool with the changes to admin-panel-server staying on the epic branch, but I'd prefer if the entity-server/aggreator changes go straight to dev? There's a few refactors going on around entity-server right now and I'd prefer to incorporate your changes sooner rather than later!

chris-bes:
@rohan-bes yep good idea! Thank you for your throughout review as well

packages/entity-server/src/models/types.ts
```diff
@@ -6,12 +6,49 @@
...
+  [field in keyof T]?: T[field] | T[field][] | AdvancedFilterValue<T[field]>;
 };
 
-export type DbConditional<T> = PartialOrArray<Omit<T, ObjectLikeKeys<T>>> &
-  PartialOrArray<Flatten<Pick<T, ObjectLikeKeys<T>>, '->>'>>;
+export enum QueryConjunctions {
+  AND = '_and_',
+  OR = '_or_',
+  RAW = '_raw_',
+}
```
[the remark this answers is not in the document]
chris-bes: I've actually tried that before. However typescript couldn't resolve the keys coming from @tupaia/database. So that's why I sorta gave up and just created an enum here :(:

image

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
chris-bes: The reason I had `null` was to avoid having the advanced criteria if the operator is `==`. So I still prefer the filter object to be:
```
{
     type: 'district'
}
```
instead of always
```
{
     type: {
           comparator: '=',
           comparisonValue: 'district'
     }
}
```

However, I've just realized I didn't add the logic to follow the 1st format mentioned above (I was working on this quite late so that might explain 😄 ). I've followed your `=` suggestion, but still have an if check for `=` operator to avoid using comparator/comparisonValue if possible

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
chris-bes: I think that was the problem! Thanks heaps!

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
chris-bes: I've checked and I think it still works, but in a coincident way(?), so I tweaked the code in ResponseBuilder.ts to fit with this. Hopefully that's alright

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -52,54 +51,96 @@ const toJsonBKey = <T extends keyof NestedFilterQueryFields>(nestedField: T): Js
...
+  allowedCountries: string[],
+  queryFilter?: string,
+): EntityFilter => {
   if (!queryFilter) {
-    return { country_code: allowedCountries };
+    // default filter to only get entities in allowed countries.
+    return {
+      [QueryConjunctions.AND]: {
+        country_code: allowedCountries,
+      },
```
chris-bes, line 125: What do you think about return this as default if there's no filter instead of the previous:
```
{ country_code: allowedCountries }
```

When it was returning the previous format, I found it hard to debug the country_code filter because I kinda got confused a lot regarding whether country_code is something from the query filter or is it the default allowed countries. So I just change it to return this format (consistent with how we have the default filter when merging at the end of this function). So when I see this, then I know it's the default allowed countries filter.

chris-bes:
Have broken up this PR into 2 commits:
https://github.com/beyondessential/tupaia/pull/2860/commits/986f137ccc0a99229d8f6331209f9ddd23c7e16c: entity-server change
https://github.com/beyondessential/tupaia/pull/2860/commits/986f137ccc0a99229d8f6331209f9ddd23c7e16c: admin-panel-server change.

I accidently did a force push instead of creating a new PR so sorry if it makes re-review harder. We can cherry-pick the first commit to a separate PR. Then when it passes testing and gets merged to dev, I can merge dev back and resolve the merge conflicts if any.

packages/entity-server/src/models/types.ts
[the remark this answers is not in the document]
rohan-bes: Ah! Perhaps try `[field in keyof typeof QUERY_CONJUNCTIONS]`? I think I mistyped

rohan-bes (approved):
All looks good @chris-bes, excellent work!

My only standing request is that the changes to entity-server get merged straight to dev? (happy to give the PR a review when you break it up). Unless viz-builder-epic-phase-1 is gonna get merged to dev real soon, then I guess I don't mind as much 👍

chris-bes:
> My only standing request is that the changes to entity-server get merged straight to dev? (happy to give the PR a review when you break it up). Unless viz-builder-epic-phase-1 is gonna get merged to dev real soon, then I guess I don't mind as much 👍

Yep I was waiting for approval to cherry-pick into another PR. I think it will also require some regression testing, so happy to break it into another task

PR #2861, opened 2021-07-20
2262 Add weekly date selector in au fluctracker

billli0:
### Issue #:
[#2262](https://github.com/beyondessential/tupaia-backlog/issues/2262) 
### Changes:
- Add weekly date selector for all map overlays in australia fluctracker


edmofro (approved):
LGTM! Before merging could you make sure none of the map overlays are connected to other map overlay groups, and if they are, that it's ok they also get the date picker within those groups too? Approving the code though :-)

billli0:
Thanks @edmofro, that is a good point! Have checked they are not connected to other map overlay groups.

PR #2863, opened 2021-07-20
3008: Added MultiEntityRoute to get entity data for multiple entities

rohan-bes:

Can be merged independently.

Introduction of a new route to the entity-server to get multiple entities:
```
POST http://{{host}}/{{version}}/hierarchy/explore
{
    "entities" : ["TO","PG"]
}
```
Returns:
```
[<TO entity object>, <PG entity object>]
```


packages/entity-server/src/app/createApp.ts
```diff
@@ -36,9 +37,11 @@ export function createApp() {
   return new MicroServiceApiBuilder(new TupaiaDatabase())
     .useBasicBearerAuth('entity-server')
     .use<SingleEntityRequest | MultiEntityRequest>('hierarchy/:hierarchyName', attachCommonContext)
+    .use<MultiEntityRequest>('hierarchy/:hierarchyName$', attachMultiEntityContext)
     .use<MultiEntityRequest>('hierarchy/:hierarchyName/descendants', attachMultiEntityContext)
     .use<MultiEntityRequest>('hierarchy/:hierarchyName/relatives', attachMultiEntityContext)
     .use<MultiEntityRequest>('hierarchy/:hierarchyName/relationships', attachMultiEntityContext)
+    .post<MultiEntityRequest>('hierarchy/:hierarchyName$', handleWith(MultiEntityRoute))
```
edmofro, line 44: There are enough of these building up that it may get a little confusing which middlewares are attached, and someone adding a new one might miss the fact they should attach the middleware. There's a different pattern that I think could make everything more obvious, which is that in express you can generally call `.post` with middlewares inserted as params between the endpoint and the handler. It wouldn't be too hard to update the `ApiBuider` classes to allow that pattern too, and might make things like this a little more straightforward.

Out of official scope for this PR, so feel free to ignore, but let's both keep in mind for next time we do this sort of thing

packages/entity-server/src/routes/hierarchy/MultiEntityRoute.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { Route } from '@tupaia/server-boilerplate';
+import { formatEntitiesForResponse } from './format';
+import { MultiEntityRequest } from './types';
+
+export class MultiEntityRoute extends Route<MultiEntityRequest> {
+  async buildResponse() {
+    const { entities, field, fields } = this.req.ctx;
+    return formatEntitiesForResponse(this.req.models, this.req.ctx, entities, field || fields);
+  }
```
edmofro, line 14: Actual handler is so simple - this package is super DRY!

edmofro (approved):
lgtm :-)

packages/entity-server/src/app/createApp.ts
rohan-bes: Love this suggestion! Will do 👍

packages/entity-server/src/app/createApp.ts
```diff
@@ -33,45 +34,63 @@ import { attachRelationshipsContext } from '../routes/hierarchy/relationships/mi
...
+    new MicroServiceApiBuilder(new TupaiaDatabase())
+      .useBasicBearerAuth('entity-server')
+
+      // MultiEntity routes
+      .post<MultiEntityRequest>(
+        'hierarchy/:hierarchyName$',
+        attachCommonContext,
+        attachMultiEntityContext,
+        handleWith(MultiEntityRoute),
+      )
```
rohan-bes, line 47: @edmofro Thanks for this suggestion, I really love it ❤️ 

I decided to go with being explicit and defining all middlewares required for each route per route. While this is a bit more verbose it has a few advantages over using any url matching logic:
- It makes it more obvious which route uses which middleware, which as you mentioned was getting quite confusing
- It allows us to ensure the type of the Request matches the middleware for each route. Previously we were using the `SingleEntityRequest` and `MultiEntityRequest` to add the middlewares, where now we use the request of the route itself, which may differ from the base types. This added strictness is really nice imo!

packages/entity-server/src/routes/hierarchy/middleware/attachCommonContext.ts
```diff
@@ -2,17 +2,19 @@
...
 
 const throwNoAccessError = (hierarchyName: string) => {
   throw new PermissionsError(`No access to requested hierarchy: ${hierarchyName}`);
 };
 
 export const attachCommonContext = async (
-  req: SingleEntityRequest | MultiEntityRequest,
+  req: Request<{ hierarchyName: string }, any, any, { field?: string; fields?: string }> & {
+    ctx: CommonContext;
+  },
```
rohan-bes, line 17: Had to make these middleware types a bit less strict since now we are using them for a whole gambit of different `...Request` types. However I'm actually glad to do this. I think the general rule for function signatures should be 'be as relaxed as is safe with the types you accept'. That leads to the highest level of flexibility and polymorphism 👍

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -123,35 +123,35 @@ export class ApiBuilder {
   addRoute<T extends ExpressRequest<T> = Request>(
     method: 'get' | 'post' | 'put',
     path: string,
-    handler: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>,
+    ...handlers: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>[]
```
rohan-bes, line 126: Did this for the `OrchestrationApiBuilder` as well, for consistencies sake 👍

packages/entity-server/src/routes/hierarchy/types.ts
```diff
@@ -17,10 +17,10 @@ export interface MultiEntityRequestParams {
   hierarchyName: string;
 }
 
-export type RequestBody = Record<string, unknown>;
+export type RequestBody = Record<string, never>;
```
edmofro, line 20: So this base type is now an empty request body?

packages/entity-server/src/routes/hierarchy/types.ts
```diff
@@ -83,7 +83,7 @@ export interface SingleEntityRequest<
 
 export interface MultiEntityRequest<
   P = MultiEntityRequestParams,
-  ResBody = EntityResponse | EntityResponse[],
+  ResBody = EntityResponse[],
```
edmofro, line 86: Yeah having this tighter is nice!

edmofro (approved):
LGTM! Thanks for making that change :-)

packages/entity-server/src/routes/hierarchy/types.ts
rohan-bes: Yup! In some ways you could argue `unknown` is more appropriate, but I like the strictness that `never` add. Now if you try to attach the `attachMultipleEntities` middleware to a `SingleEntityRoute` you get a typescript error 👍

rohan-bes:
Performed sanity testing on a feature instance and it all looks good, merging to dev

PR #2864, opened 2021-07-20
3008: Moved entity-server start script to `main.ts`

rohan-bes:

Can be merged independently.

Moved the logic to start the entity-server from `index.ts` over to `main.ts`. This allows us to export things from the `entity-server` (importantly types which are required by https://github.com/beyondessential/tupaia/pull/2723)


packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -42,7 +42,7 @@ for PACKAGE in ${PACKAGES[@]}; do
         # It's a server, start the pm2 process
         echo "Starting ${PACKAGE}"
         yarn build
-        pm2 start --name $PACKAGE dist --wait-ready --listen-timeout 15000 --time
+        pm2 start --name $PACKAGE 'yarn start' --wait-ready --listen-timeout 15000 --time
```
rohan-bes, line 45: I believe this should be fine, but will deploy and test in a feature instance to ensure behaviour remains consistent with the way we initialized the daemon previously

edmofro: Could you look in particular at the prestart behaviour on restart? My only worry would be that it runs the prestart script when e.g. web-config-server crashes, meaning noticeable downtime before it comes back online.

If that's not the case, this is totally cool, and we can remove the `yarn build` above as the prestart hook should take care of that (plz check all packages though 🙏 )

rohan-bes: Excellent call about the `prestart` hook @edmofro ! I tested it and indeed it was rebuilding over every restart!

Got any clever ideas about how we might avoid this? My simplistic solution was to add a `start-post-build` script to each server that identical to the `start` one but due to it having a different name it avoids the `prestart` hook. I'm not the biggest fan of my implementation, nor the name, but was a little stumped to come up with anything better. Doesn't seem like node lets you skip the hook...

edmofro: Bummer...

One trick I've used in another spot is an environment variable
```
"postinstall": "test -n \"$SKIP_BUILD_INTERNAL_DEPENDENCIES\" || yarn build-internal-dependencies"
```

Which then allows
```
## run yarn without building internal dependencies, so we can cache that layer without code changes
## within internal dependencies invalidating it
RUN SKIP_BUILD_INTERNAL_DEPENDENCIES=true yarn install
```

We could use that same pattern here, with the result being something like
```
pm2 start --name $PACKAGE 'SKIP_PREBUILD=true yarn start' --wait-ready --listen-timeout 15000 --time
```

rohan-bes: I think I like that a bit more than my fix! Let's go with it 👍

rohan-bes: @edmofro apologies for the slow turnaround on this one. I tried testing the `test` command in my local bash shell and wasn't seeing anything so I thought maybe it wasn't working. Turns out it was, it just wasn't logging anything to the console 🤦 

Seems to work fine in my local now 🎉 will test on a feature instance before merging 👍

packages/server-boilerplate/package.json
```diff
@@ -17,11 +17,7 @@
     "build": "babel src --out-dir dist --source-maps --ignore \"src/__tests__/**\" --extensions .ts --config-file \"../../.babelrc-ts.js\"",
     "lint": "tsc; eslint . --ext .ts",
     "lint:all": "yarn run lint \"src/**/*.{ts,jsx}\"",
-    "lint:fix": "yarn lint --fix",
-    "prestart": "npm run -s build",
-    "start": "node dist",
-    "start-dev": "LOG_LEVEL=debug ../../scripts/bash/backendStartDev.sh 9998 -ts",
-    "start-verbose": "LOG_LEVEL=debug npm run start-dev"
```
edmofro, line 24: Good call removing these

edmofro (approved):
Looks good!

rohan-bes:
So I did find one issue introduced by running the orchestration servers via `yarn start`. It seems that the `process.send('ready')` doesn't bubble up from the yarn process to pm2, which means that pm2 always waits the full 15 second timeout that we configured before declaring the process as 'started'.

I've spent a bit of time trying to look into ways to fix this within pm2; either by not starting the process in fork mode, or by listening to something other than the `process.send('ready')` signal (the docs suggest this is possible by listening to the webserver being available somehow), but haven't had much luck. The docs seem pretty sparse tbh...

Not exactly sure how I feel about this issue... On the one hand I don't think it affects server downtime at all; the server comes online whenever it comes online afterall? This basically just means we always wait the max configured time (which we _could_ change if we wish) before moving on in the deployment process. On the other hand, it does feel like a loss in functionality for arguably not much reason... @edmofro any thoughts and feelings on this one?

edmofro:
Ooosh yeah that's a bit awkward huh. So it's actually only meditrak-server that we really need to have fully up before moving on in the deployment process - the reason for having pm2 wait was originally to make sure change notification handling was listening on meditrak-server before running any migrations. If you can think of a solution to only that problem, I'd be happy to get rid of the pm2 wait behaviour everywhere else.

rohan-bes:
>Ooosh yeah that's a bit awkward huh. So it's actually only meditrak-server that we really need to have fully up before moving on in the deployment process - the reason for having pm2 wait was originally to make sure change notification handling was listening on meditrak-server before running any migrations. If you can think of a solution to only that problem, I'd be happy to get rid of the pm2 wait behaviour everywhere else.

Oh yeah! Good point about it really just being meditrak-server that cares about this logic. So I guess it's worth noting that with the new 'broken' behaviour we actually retain the same level of assuredness that the server is up before running migrations.
The behaviour used to be:
`wait until 'ready' is sent, or until 15 seconds have elapsed, before proceeding`
Now it is:
`wait until 15 seconds have elapsed, before proceeding`
So we won't be introducing any risk of running migrations before the server is up.

That being said it does really annoy me that we're losing this functionality. I'm gonna give it one more shot to try and this alternative config working this morning:
image

rohan-bes:
Okay so after bashing my head against this for longer then I care to admit, I have given up 😿 I don't think there's a way (outside some crazy process piping shenanigans) to get pm2 to hear messages from the node process if started via `yarn start`. What's more, this gives me concerns that it may complicate matters when introducing pm2 cluster mode which I know Igor is working on.

Given that, I think maybe there's two approaches left here:
1. Move all servers over to having `main.ts`/`main.js` be their main files. Then they can all be started via `node dist/main.js`.
2. Scrap this whole `main.ts` thing all together, and instead just import types from server packages by referring to an explicit file, ie. `import type { EntityApiQueryOptions } from '@tupaia/entity-server/src/type-exports';`

Of these two options, I think I'm leaning towards 2. I do see the merits of 1, but it doesn't really seem to be worth the effort. Which unfortunately means this whole PR was a bit of a waste of both our times and for that I am sorry 😞 

Am gonna continue on with the rest of the PR's related to this work as if I'm going with 2, but let me know if you'd prefer to go with 1, or have another option I missed?

edmofro:
Aaah bummer. I guess there's one other option, which is just to accept the 15 seconds...but yeah, I'd probably go for option 2 above. Thanks for knowing when to admit defeat!

PR #2868, opened 2021-07-20
3008: EntityApi class attached to microservices SOL-32

rohan-bes:

Must be merged after other changes.

Added `EntityApi` class and attached it to the context for all microservices. 

### Changes:
- Exported a number of types from entity-server


packages/server-boilerplate/src/connections/ApiConnectionBuilder.ts
```diff
@@ -0,0 +1,29 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
```
nick-beyondessential, line 3:
```suggestion
 * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
```

packages/server-boilerplate/src/connections/ApiConnectionBuilder.ts
```diff
@@ -0,0 +1,29 @@
...
+export class ApiConnectionBuilder {
+  private authHandler?: AuthHandler;
+
+  public handleAuthWith(authHandler: AuthHandler) {
+    this.authHandler = authHandler;
+    return this;
+  }
+
+  public buildAs<Wrapper extends MicroserviceApi>(
+    WrapperClazz: new (apiConnection: ApiConnection) => Wrapper,
```
nick-beyondessential, line 18: So Java!

packages/server-boilerplate/src/microService/api/ApiBuilder.ts
```diff
@@ -27,7 +29,7 @@ export class ApiBuilder {
     this.models = new ModelRegistry(transactingConnection);
     this.app = express();
 
-    this.version = 'v1'; // Default version
+    this.version = 'v[0-9]'; // Default version
```
nick-beyondessential, line 32: Out of curiosity, what is going on here?

rohan-bes: Ahh good question! So basically our api endpoints always go `http://<host>/<version>/<endpoint>` and so the ApiBuilder allows you to set a version so that it can staple it to each route, thus saving a us from having to remember to put it in when defining every url. 

The problem is that for some reason some of the `.env` vars preset in our deployment instances used `v2` for report-server... I considered changing them but kinda figured that unless we are explicitly using versioning (which I don't think any of our services actually do) then why not just make it a wildcard to make life easier?

nick-beyondessential (approved):
Looks good to me

rohan-bes:
Test comment

rohan-bes:
e2e test failures are known flaky tests, see: https://linear.app/tupaia/issue/MEL-383/fix-flaky-reports

Merging as smoke testing of PSSS and Lesmis has been successful and tupaia e2e test failures were either timeouts or flaky tests

PR #2870, opened 2021-07-20
3102 add user route to admin panel server

tcaiger:
### Issue #: [3102](https://github.com/beyondessential/tupaia-backlog/issues/3102) Viz Builder Frontend Scaffolding

This is just a supporting PR for the upcoming [3102 viz builder fe scaffolding](https://github.com/beyondessential/tupaia/pull/2854).

### Changes:

- add user route to admin panel server
- add camelcase-keys package

---


packages/admin-panel-server/src/connections/MeditrakConnection.ts
```diff
@@ -2,14 +2,30 @@
...
+import { QueryParameters } from '@tupaia/server-boilerplate';
+import { SessionHandlingApiConnection } from './SessionHandlingApiConnection';
+import { BES_ADMIN_PERMISSION_GROUP } from '../constants';
 
 const { MEDITRAK_API_URL = 'http://localhost:8090/v2' } = process.env;
 
-export class MeditrakConnection extends ApiConnection {
+const isBESAdmin = (policy: Record<string, string[]>) => {
+  return new AccessPolicy(policy).allowsSome(null, BES_ADMIN_PERMISSION_GROUP);
+};
```
chris-bes, line 15: I was thinking maybe we can have a middleware that gets run for all the viz-builder-route that can check for permission. However I haven't got that done yet, so I think for now this should be ok!

packages/admin-panel-server/src/connections/SessionHandlingApiConnection.ts
```diff
@@ -0,0 +1,47 @@
...
+  async getAuthHeader() {
+    if (this.session) {
+      return this.session.getAuthHeader();
+    }
+
+    return DEFAULT_AUTH_HEADER;
+  }
+}
+
+export abstract class SessionHandlingApiConnection extends ApiConnection {
```
chris-bes, line 35: I actually don't think you need this whole class/file right? I think for lesmis, because we do allow some public functionalities (users don't need to log in). And my understanding is this SessionHandlingApiConnection class in lesmis-server is to allow that. But admin-panel might not be the same case?

If you agree then MeditrakConnection can just extends ApiConnection from `@tupaia/server-boilerplate` instead of this class!

packages/admin-panel-server/src/routes/UserRoute.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Request, Response, NextFunction } from 'express';
+import { Route } from '@tupaia/server-boilerplate';
+import { MeditrakConnection } from '../connections';
+
+export class UserRoute extends Route {
+  private readonly meditrakConnection: MeditrakConnection;
+
+  constructor(req: Request, res: Response, next: NextFunction) {
+    super(req, res, next);
+    console.log('user route');
```
chris-bes, line 15: console.log() here!

packages/admin-panel-server/src/connections/MeditrakConnection.ts
```diff
@@ -2,14 +2,30 @@
...
+
+export class MeditrakConnection extends SessionHandlingApiConnection {
   baseUrl = MEDITRAK_API_URL;
 
+  async getUser() {
+    // if user is not logged in, return null rather than fetching the api client user details
+    if (!this.hasSession) {
+      return {};
+    }
```
chris-bes, line 24: Similar comment in `SessionHandlingApiConnection` class, we probably don't need to do this as users are always required to log in with admin-panel?
```suggestion
```

packages/admin-panel-server/src/app/createApp.ts
```diff
@@ -47,6 +47,7 @@ export function createApp() {
   const app = new OrchestratorApiBuilder(new TupaiaDatabase())
     .useSessionModel(AdminPanelSessionModel)
     .verifyLogin(hasTupaiaAdminPanelAccess)
+    .get('/v1/user', handleWith(UserRoute))
```
chris-bes, line 50: I'm thinking that since this is admin-panel-server now instead of viz-builder-server, we might have to be more specific with our endpoints path. Maybe we can have `/v1/vizBuilder/user` (and similar pattern for other viz-builder endpoints as well). But I think you don't need to change this now, I can bulk change that when it is decided

chris-bes (changes requested):
Looks good! I just have some minor suggestions

packages/admin-panel-server/src/connections/MeditrakConnection.ts
tcaiger: Yes that's a good idea. Yes sounds like it will just be easiest for you to refactor it when you get to it so I'll just leave it as is.

packages/admin-panel-server/src/connections/SessionHandlingApiConnection.ts
tcaiger: Yes that's a very good point. I'll remove the new class!

packages/admin-panel-server/src/app/createApp.ts
tcaiger: Yep for sure. If you end up adding that middleware for the permission checking then we might be able to remove this route from admin-panel-server altogether and use the '/me' endpoint from meditrak-server. I really just needed it to add the isBESAdmin boolean. Again feel free to refactor or delete this down the track - I was mostly just getting something working so I could unblock my front-end work.

chris-bes:
Looks good to me! I've also just realized that we may not have to create /user route in admin-panel-server, as it will just forward the request to meditrak-server /me endpoint (sorry for not realizing this sooner). However, I did see that you have some camelCase manipulation so maybe we still need the endpoint.

chris-bes (approved):

PR #2873, opened 2021-07-21
Fix permissions on dump db script

edmofro:
On my local I was unable to dump the database from dev. Found that the permissions no longer included `x`, presumably changed unintentionally during https://github.com/beyondessential/tupaia/commit/93caa6830e4dc088df37730b068a7d3487f8a1b6

image

This reinstates it to `755`


kael89 (approved):
Oh classic... Good catch, thanks for fixing this!

PR #2876, opened 2021-07-22
Fix bad quotes on db command

EMcQ-BES:
### Changes:

- Noticed this while trying to refresh my local


kael89 (approved):
Thanks for fixing this @EMcQ-BES

edmofro:
I don't think this needs testing so feel free to merge @EMcQ-BES

PR #2878, opened 2021-07-22
2984 Add report for list visual

EMcQ-BES:
### Changes:

- Add a report to generate a list for list visual

---

### Screenshots:

Here's the results from postman as a `.txt` file

[results.txt](https://github.com/beyondessential/tupaia/files/6860032/results.txt)


edmofro (changes requested):
Very nice, and the product of some other hard work so it's good to see it all come together.

Seeing an actual example and thinking through the potential future uses, I'd like to change the word `statistic` to `value`, or something else that doesn't imply it being numeric. One other rename suggestion inline, and otherwise just an optional thought that you can choose to investigate or not :-)

packages/database/src/migrations/20210708023241-LESMISListVisualSummaryReport-modifies-data.js
```diff
@@ -0,0 +1,343 @@
...
+      "'statistic'":
+        'eq(max($where(f($otherRow) = equalText($otherRow.parent, "HLO1")).statistic), min($where(f($otherRow) = equalText($otherRow.parent, "HLO1")).statistic)) ? last($where(f($otherRow) = equalText($otherRow.parent, "HLO1")).statistic) : 0',
+    },
+  ],
+};
+
+const FRONT_END_CONFIG = {
+  name: 'ESSDP List Summary',
+  type: 'list',
+  valueType: 'trafficLight',
```
edmofro, line 254: Rather than `trafficLight`, it might be better to go with `color`, and then provide a mapping of value to color hex code. Keeps the colour list open to uses outside of red/orange/green

EMcQ-BES: Updated. Added mappings to a `listConfig` field

packages/database/src/migrations/20210708023241-LESMISListVisualSummaryReport-modifies-data.js
```diff
@@ -0,0 +1,343 @@
...
+  transform: [
+    'keyValueByDataElementName',
+    // Build list from the bottom up
+    // Because we need the lower level values to calculate the higher ones
+    {
+      transform: 'select',
+      where: 'exists($row.er_summary_ece_0_2_t)',
+      "'code'": "'ECETarget0_2'",
+      "'label'":
+        "'TARGET: Enrolment rate of 0-2 years old children: gender, GPI, total – 2025 target of 7%'",
```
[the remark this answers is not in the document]
EMcQ-BES: Experimented with this a little bit and couldn't find a nice way to make it happen.

packages/database/src/migrations/20210708023241-LESMISListVisualSummaryReport-modifies-data.js
```diff
@@ -0,0 +1,357 @@
...
+      value: 1,
+    },
+    Yellow: {
+      color: '#ffeb3b',
+      value: 0,
+    },
+    Red: {
+      color: '#f44336',
+      value: -1,
+    },
```
edmofro, line 267: Hmm, I think this might work better with the value as the key. Generally the front end will be interested in asking "for this value, what colour should I use?" (and maybe "what hover text", which would involve the `Green` etc. moving in under a `name` or `hover` or `label` key)

edmofro (approved):
lgtm, but may still be changes when the front end comes through next sprint. Approved for now!

PR #2880, opened 2021-07-22
hotfix: Fixup dataSourceEntityType for a few lesmis reports

rohan-bes:
### Issue https://beyondessential.slack.com/archives/C01MAA3NKM1/p1626925063242100:

Previous `dataSourceEntityType` was incorrect but it was still working due to bug in `EntityConnection` in `Aggregator`. Now that bug has been fixed, so this one must be too.

### DIFF
--- a/packages/database/src/migrations/20210722044503-FixDataSourceEntityTypeLesmisProvinceStudentDashboards-modifies-data.js
+++ b/packages/database/src/migrations/20210722044503-FixDataSourceEntityTypeLesmisProvinceStudentDashboards-modifies-data.js
@@ -16,25 +16,41 @@ exports.setup = function (options, seedLink) {
   seed = seedLink;
 };
 
-const dashboards = [
+const PROVINCE_REPORTS = [
   'LESMIS_gross_intake_ratio_primary_province',
   'LESMIS_gross_intake_ratio_lower_secondary_province',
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210722044503-FixDataSourceEntityTypeLesmisProvinceStudentDashboards-modifies-data.js
```diff
@@ -0,0 +1,43 @@
...
+const dashboards = [
+  'LESMIS_gross_intake_ratio_primary_province',
+  'LESMIS_gross_intake_ratio_lower_secondary_province',
+  'LESMIS_gross_intake_ratio_upper_secondary_province',
+];
+
+exports.up = function (db) {
+  return db.runSql(`
+    UPDATE report
+    SET config = jsonb_set(config, '{fetch,aggregations,0,config,dataSourceEntityType}','"district"', false)
```
EMcQ-BES, line 28: Is this gonna have the same issue on the country level reports?

edmofro: Yep I believe you're right, good catch! @rohan-bes I'll update the migration now

edmofro (approved):

PR #2882, opened 2021-07-22
add suffix -ssh for updating dump database script

billli0:
### Change #:
Add suffix -ssh for updating dump database.

It has been tested locally.


kael89 (changes requested):
Great initiative fixing this @billli0! I suggest adding the `-ssh` suffix to all `server` inputs, to retain the existing functionality and keep things simpler for the script's users

kael89:
So an example usage would be: `yarn dump-database <identityFile> -s branch` => would ssh into `branch-ssh.tupaia.org`

kael89 (approved):
Nice one!

PR #2866, opened 2021-07-20
3008: Switched report-server to using server-boilerplate

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/3008:

Part of: https://github.com/beyondessential/tupaia/pull/2723

Can be merged independently.

Migrate report-server over to using server-boilerplate package.

### Changes:
- Moved some types from entity-server to server-boilerplate

PR #2884, opened 2021-07-22
3121 + 31222: /saveDashboardVisualisation + /fetchReportPreviewData

chris-bes:
### Changes:

- /saveDashboardVisualisation
- /fetchReportPreviewData

### Some architecture decisions:
May have already been discussed in tech design, but still mentioning for context:

1. _The representation of visualisation (report + dashboard_item) in the front end should be human readable (ie: shouldn't be aware of report/dashboard_item)._ 

The reason is we want to make this closer to the users as it will also be used in other operation such as import/export. Hence users can even make small changes here and there in the import/export files. As stated in this comment (https://github.com/beyondessential/tupaia-backlog/issues/3121#issuecomment-873780717), the payload should look something like this (not the final version):
```
{
  "name": "New Dashboard Item",
  "data": {
    "dataElements": ["BCD1"],
    "aggregation": {
      "BCD1": "FINAL_EACH_DAY"
    }
  },
  "presentation": {
    "chartType": "bar",
    "config": {
      "colorPalette": "default"
    }
  }
}
```

_2. All the endpoints in the backend that perform operations related to visualisations should receive the full payload (the whole representation mentioned above) regardless, even if some information may be redundant._ 

The reason is most of the time, operations will need the whole payload. In some special cases like when fetching preview data, we don't need the whole payload, but still need a bit of information here and there between the data and presentation to build the report config (because report's output step actually lives in `presentation` instead of `data`). I also think sending the whole viz representation as payload will also help remove the need of the front-end to be aware of the existence of dashboard_item/report, instead, that responsibility can be delegated to the orchestration server. The front end doesn't have to do the tricky part of extracting the right information and just have to be aware of only one thing which is the visual representation.


packages/admin-panel-server/src/viz-builder/extractors/DashboardVisualisationExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { yup } from '@tupaia/utils';
+
+import { DashboardVisualisationObject } from '../types';
+import { omitBy, isNil } from 'lodash';
+
+export abstract class DashboardVisualisationExtractor {
```
chris-bes, line 11: This abstract class is for extracting the generic visualisation presentation to what we need (ie: to dashboard_item + report). I was thinking that for we can extend the class to fit our need. Some examples are:
- When we save a draft visualisation version, we don't need to validate everything in dashboard_item and report
- When we publish a visualisation, we will need to fully validate everything in dashboard_item and report
- When we fetch preview data, we only need to extract the report/data part.
- When we implement step by step play, we also need to extract only up to a certain steps,...

packages/admin-panel-server/src/viz-builder/extractors/DraftDashboardItemExtractor.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { DashboardVisualisationExtractor } from './DashboardVisualisationExtractor';
+
+export class DraftDashboardItemExtractor extends DashboardVisualisationExtractor {
```
chris-bes, line 8: Initially, I was thinking of just having 1 extractor `DraftVisualisationExtractor` class that will be responsible for extracting both report/dashboard_item from the visualisation object. However, there is 1 case that we only need to extract the report part. So I split it up into 2 extractor classes, 1 for report and 1 for dashboard_item. However, I do have a slight feeling that I may have over engineered it...

packages/admin-panel-server/src/viz-builder/extractors/DashboardVisualisationExtractor.ts
```diff
@@ -0,0 +1,145 @@
...
+            otherwise: yup.array().of(yup.string()),
+          }),
+        dataGroups: yup
+          .array()
+          .of(yup.string())
+          .when('dataElements', { // dataElements is required if there are no dataGroups 
+            is: (dataElements: string[]) => !dataElements || dataElements.length === 0,
+            then: yup.array().of(yup.string()).required().min(1),
+            otherwise: yup.array().of(yup.string()),
+          }),
```
chris-bes, line 55: When saving a draft version, I think we still need at to have some minimum requirements so that we just don't save whatever and avoid some column constraints in the database

packages/meditrak-server/src/routes/CreateDashboardVisualisation.js
```diff
@@ -0,0 +1,65 @@
...
+    return transactingModels.dashboardItem.create(dashboardItemObject);
+  }
+
+  async createRecord() {
+    await this.assertPermissions(assertBESAdminAccess);
+
+    return this.models.wrapInTransaction(async transactingModels => {
+      const { report: reportObject, dashboardItem: dashboardItemObject } = this.newRecordData;
+      await this.createReport(transactingModels, reportObject);
+      await this.createDashboardItem(transactingModels, dashboardItemObject);
```
chris-bes, line 45: Another option is have separate endpoint for saving report and dashboard item. The reason I choose to have 1 combined endpoint is to have 1 transaction so that if it fails saving 1 of them, it shouldn't save the other.

packages/report-server/src/routes/fetchReport.ts
```diff
@@ -19,20 +19,30 @@ const getFilterFromReq = (req: ReportsRequest): FetchReportQuery => {
...
     if (body.testConfig) {
+      const { permissionGroup: permissionGroupName } = query;
+      if (permissionGroupName) {
+        await checkUserHasAccessToReport(
+          models,
+          accessPolicy,
+          permissionGroupName,
+          filter.organisationUnitCodes.split(','),
+        );
+      }
```
chris-bes, line 37: This is to allow permission check when fetch preview data

packages/admin-panel-server/src/app/createApp.ts
```diff
@@ -49,6 +53,8 @@ export function createApp() {
     .useSessionModel(AdminPanelSessionModel)
     .verifyLogin(hasTupaiaAdminPanelAccess)
     .get('/v1/hierarchy/:hierarchyName/:entityCode', handleWith(FetchHierarchyEntitiesRoute))
+    .get('/v1/fetchReportPreviewData', handleWith(FetchReportPreviewData))
+    .post('/v1/saveDashboardVisualisation', handleWith(SaveDashboardVisualisation))
```
chris-bes, line 57: Since viz-builder endpoints live in admin-panel-server now, I think we have to name our endpoints a bit more specific (eg: cannot name /save anymore). I have a thought that we should have all the endpoints for viz-builder to start with `vizBuilder`, ie:
- `/v1/vizBuilder/fetchPreviewData`
- `/v1/vizBuilder/save`

That way we don't have to be too specific with our endpoints naming anymore. Also, if we want to check for permission for all viz-builder endpoints, we can easily add a middleware for endpoints that match `vizBuilder`?

packages/report-server/src/app/addRoutesToApp.ts
```diff
@@ -27,7 +27,7 @@ export function addRoutesToApp(app: Express) {
   /**
    * POST routes
    */
-  app.post('(/v[0-9]+)?/testFetchReport/:reportCode', wrappedFetchReportWithAsyncCatch); // POST endpoint for testing fetching a report (able to provide test data and config in request body)
```
chris-bes, line 30: For this endpoint I don't think we need to provide a report code?

packages/admin-panel-server/examples.http
```diff
@@ -0,0 +1,113 @@
...
+                    "organisationUnit": "group",
+                    "period": "group",
+                    "Doctors": "sum",
+                    "Midwives": "sum",
+                    "Nurses": "sum",
+                    "Aides": "sum",
+                    "Others": "sum"
+                }
+            ]
+        }
```
rohan-bes, line 55: I've actually noticed a bit of a tricky bit of logic we have here...
The report-server has a final 'output' step where it is able to reformat that base table that comes from the transform step into something that the frontend can consume. A few things to note about the output step:
1. We plan on having it's configuration be specified in the 'Presentation' section of the config
2. Depending on whether we are generating a 'Data' or 'Chart' preview, we conditionally want to include it in the config we send the to the report-server. If we are viewing the 'Data' table, then we don't want the report-server to process the output step as otherwise the data will come back in weird shape, but for Chart preview we want it to be processed.

I guess this brings up a few things we need to consider:
1. Who makes the call to decide if we send through to the report-server. The orchestration server here? Or does the front-end edit what's sent through to this endpoint based on which tab the user is on?
2. How do we avoid constantly refetching data whenever the user jumps between tabs? I assume by front-end caching? Should be fine 👍 
3. Who works out how to combine the 'Output' section of the Presentation config with the Report? Frontend or backend?

I'd probably lean towards backend for both 1, and 3 👍 I imagine the frontend will send through the following:
```
POST {{baseUrl}}/fetchReportPreviewData?entityCode=TO&previewMode=data
{
    "previewConfig": <all viz config>
}     
```
And then the backend works out what to send to the report-server based on which preview mode it is (data/chart). Would you agree?

packages/admin-panel-server/examples.http
```diff
@@ -0,0 +1,113 @@
+@baseUrl = http://localhost:8070/v1
+# Uncomment below to fetch against the dev server, rather than localhost
+# @baseUrl = https://dev-admin-api.tupaia.org/v1
+@contentType = application/json
+# In order to setup authorization, please set 'email' and 'password' in your restClient environement variables
+# see: https://marketplace.visualstudio.com/items?itemName=humao.rest-client#environment-variables
+@authorization  = Basic {{email}}:{{password}}
+
+### Fetch Dashboard Preview Data
+POST {{baseUrl}}/fetchReportPreviewData?entityCode=TO
```
rohan-bes, line 10: We do need to send in the hierarchy here as well, however I think we agree to scrap the hierarchy selector and use the 'Project' selector that sits near the header, right @tcaiger ? If so I guess the hierarchy will be part of the 'main' config of the viz, so we will need that to come in this payload somehow.

There's actually another sub issue here, tho we may be able to avoid it for now, but 'Projects' and hierarchies aren't actually the same thing... We are _**meant**_ to be able to share one hierarchy among many projects, and thus not have a one-to-one mapping between projects and hierarchies, however we've never properly implemented this as yet. For now, I believe that there's always a hierarchy for a project, and they always share the same code as well (may need to double check me on that one)? If so, then project code should suffice, but we should be aware that this may change in the future...

packages/meditrak-server/src/routes/CreateDashboardVisualisation.js
rohan-bes: I like your reasoning @chris-bes ! I also quite like that we are closely coupling the 'report' and 'dashboardItem' throughout much of logic relating to 'dashboardVisualisations'. Afterall, a dashboardItem requires a report, so it's often nice to treat them as an atomic unit 👍

packages/report-server/src/app/addRoutesToApp.ts
```diff
@@ -27,7 +27,7 @@ export function addRoutesToApp(app: Express) {
...
+  app.post('(/v[0-9]+)?/testFetchReport', wrappedFetchReportWithAsyncCatch); // POST endpoint for testing fetching a report (able to provide test data and config in request body)
```
rohan-bes, line 30: Just a heads up that this endpoint will be changing soon!

packages/report-server/src/routes/fetchReport.ts
rohan-bes: Hehe I'm glad someone finally added a permission check to here! The security hole was giving me anxiety 🙀 

We do however have to take a further look at the permissions model for this endpoint (and the preview functionality in the viz builder in general) to ensure that everything is locked down!

Even this isn't quite secure, since we aren't able to do any checks on the data elements the user is requesting, so they can query for whatever data they like, and as long as they have some permissions to that country, they can get the data (regardless if users with those permissions should have access to that data...)

packages/admin-panel-server/src/viz-builder/types.ts
```diff
@@ -0,0 +1,34 @@
...
+type Aggregation = string | AggregationObject;
+
+type Transform = Record<string, unknown>;
+
+type DataObject = {
+  dataElements: string[];
+  dataGroups: string[];
+  aggregations: Aggregation[];
+  transform: Transform[];
+};
```
rohan-bes, line 20: We really do need a common way of defining types across packages... I think we have aggregation objects defined in like 3 packages now 😝 

In a refactor I'm working on I've moved the 'DbModel' types from entity-server into server-boilerplate. I may do the same for the ReportModel types from report-server and then use them here too!

rohan-bes (changes requested):
Awesome work @chris-bes! It's looking really good!

Some notes and changes requested:
- A few scenarios we aren't quite yet covering with the 'fetchReportPreview' route
- A refactor suggestion around the 'extractors'
- Notes about some upcoming merge conflicts that will occur when I merge https://github.com/beyondessential/tupaia/pull/2866

Happy to talk through any of these comments since they're a bit extensive I know 👍 

Also, apologies again for my tardiness on this review!!! Next time feel free to smack me over the head, or publicly shame me in the reviews channel 😝

packages/admin-panel-server/examples.http
chris-bes: Ah yes, definitely agree! In fact I'd point the same idea out in the description of this PR (point 2):

> 2. All the endpoints in the backend that perform operations related to visualisations should receive the full payload (the whole representation mentioned above) regardless, even if some information may be redundant.
>
> The reason is most of the time, operations will need the whole payload. In some special cases like when fetching preview data, we don't need the whole payload, but still need a bit of information here and there between the data and presentation to build the report config (because report's output step actually lives in presentation instead of data). I also think sending the whole viz representation as payload will also help remove the need of the front-end to be aware of the existence of dashboard_item/report, instead, that responsibility can be delegated to the orchestration server. The front end doesn't have to do the tricky part of extracting the right information and just have to be aware of only one thing which is the visual representation.

packages/admin-panel-server/src/viz-builder/extractors/DraftDashboardItemExtractor.ts
chris-bes: Yep I think I agree with that. I still think we need inheritance here just because we will need different validation between  saving a draft version and saving a published version (draft version can still miss some fields, whereas published version we will have to validate everything). So I'd imagine the inhritance model would look like below:

```
                       AbstractDashboardVisualisation
                       //                         \\
                    //                               \\
                 //                                     \\
               V                                           V
DraftDashboardVisualisation                 PublishedDashboardVisualisation
```

What do you reckon?

rohan-bes: I think that should work 👍 Sometimes I get a bit worried about employing inheritance since it can often lead to additional complexity, but what you describe does seem pretty clean. Basically the only difference would be the validation logic right?

packages/admin-panel-server/examples.http
rohan-bes: Oh sorry that I missed your point! You worded it perfectly, totally agree with you on that 👍

packages/admin-panel-server/src/viz-builder/extractors/DraftDashboardItemExtractor.ts
chris-bes: Definitely understand your concern about inheritance adding additional complexity 👍 . Yep I think validation is the main difference.

packages/admin-panel-server/src/viz-builder/extractors/DashboardVisualisationExtractor.ts
rohan-bes: Hmmm you do make a good point... I think step-by-step play was something I wasn't considering when I made my suggestion particularly... I can see how pushing all the functionality into a single class could have issues down the line.

Generally in these kinds of situations I try to favor composition over inheritance. For instance, perhaps if we were to separate the validation logic from the extraction logic, we may be able to have an easier time mixing and matching them as is needed...

Regardless I'm really happy to let you make the call on this one for the following reasons:
- You seem to have a pretty solid grasp on the various different edge cases
- I definitely don't have a solid idea of how this should come together
- We are just start out on this project, so refactors will inevitably come regardless as requirements make themselves clear
- I've already held this PR up long enough

packages/admin-panel-server/src/routes/FetchReportPreviewData.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import { Request, Response, NextFunction } from 'express';
+
+import { Route } from '@tupaia/server-boilerplate';
+
+import { ReportConnection } from '../connections';
+import { DraftReportExtractor } from '../viz-builder';
+
+export class FetchReportPreviewData extends Route {
```
rohan-bes, line 14:
```suggestion
export class FetchReportPreviewDataRoute extends Route {
```
Small thing, but it's a bit of a convetion for us to suffix all Route classes with `Route` 👍

rohan-bes:
I checked out the code for this branch, and one thing I noticed was quite a few typescript errors about the place. A lot of them weren't even related to this code, rather more general issues in the package itself. I don't wanna hold up bringing these changes in, as I think getting fast feedback on the functionality is more important, but wondering if you'd like to do a pair programming session some time soon where we can work through the package and clean the errors up?

I appreciate that typescript, and even more the conventions around the server-boilerplate, are really new and not well circulated throughout the wider team as yet. Hoping that pair programming can facilitate some knowledge transfer? 😄

packages/admin-panel-server/examples.http
chris-bes: Yeah I was aware of this, and I thought we can actually temporarily solve this by letting the /projects endpoint in meditrak-server to also return the entity hierarchy code (we just need to do a custom join). Might not be the best way but I think it's the safer approach than assuming project code is also the same as the entity hierarchy code/name

packages/admin-panel-server/src/viz-builder/extractors/DashboardVisualisationExtractor.ts
chris-bes: Sorry for holding up this PR for a long time. I've spent a lot of time trying to get it right in the first place but honestly don't have a solid idea at this early stage as well (I've tried some different ways but thought they were overengineered/clunky). So I went with the simpler approach which is to have one `DashboardVisualisation` object and let it accept different validators to mix and match (as I think validators are the main difference between different endpoints). Plus, I think it's the structure that is easiest to be refactored later when we implement more extracting cases

packages/admin-panel-server/src/app/createApp.ts
```diff
@@ -47,8 +58,22 @@ export function createApp() {
   const app = new OrchestratorApiBuilder(new TupaiaDatabase())
     .useSessionModel(AdminPanelSessionModel)
     .verifyLogin(hasTupaiaAdminPanelAccess)
-    .get('/v1/user', handleWith(UserRoute))
-    .get('/v1/hierarchy/:hierarchyName/:entityCode', handleWith(FetchHierarchyEntitiesRoute))
+    .get('/v1/user', verifyBESAdminAccess, handleWith(UserRoute))
```
rohan-bes, line 61: Should this `user` route be using `BESAdminAccess` privileges now?

packages/admin-panel-server/src/app/createApp.ts
```diff
@@ -47,8 +58,22 @@ export function createApp() {
...
+    )
+    .post(
+      '/v1/fetchReportPreviewData',
+      verifyBESAdminAccess,
+      handleWith(FetchReportPreviewDataRoute),
+    )
+    .post(
+      '/v1/saveDashboardVisualisation',
+      verifyBESAdminAccess,
+      handleWith(SaveDashboardVisualisationRoute),
```
rohan-bes, line 75: Nice work adapting to use the new server-boilerplate middleware interface 👍

packages/admin-panel-server/src/routes/FetchReportPreviewDataRoute.ts
```diff
@@ -0,0 +1,49 @@
...
+    const { previewConfig } = this.req.body;
+    if (!previewConfig) {
+      throw new Error('Preview config is empty');
+    }
+
+    const { config: reportConfig } = new DashboardVisualisationExtractor(
+      previewConfig,
+      new DraftDashboardItemValidator(),
+      new DraftReportValidator(),
+    ).getReport(previewMode as PreviewMode);
```
rohan-bes, line 39: We can handle typecasting the `previewMode` by specifying a `Request` type for this route, however let's go through that together at another time?

packages/admin-panel-server/src/viz-builder/extractors/validators.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { yup } from '@tupaia/utils';
+
+import { VisualisationValidator, DashboardVisualisationObject } from '../types';
+
+export class DraftReportValidator implements VisualisationValidator {
```
rohan-bes, line 10: Love the switch to implementing an interface here rather than extending an abstract class 👍

packages/admin-panel-server/src/viz-builder/extractors/validators.ts
```diff
@@ -0,0 +1,57 @@
...
+    }, ['dataElements', 'dataGroups']);
+  }
+
+  validate(visualisation: DashboardVisualisationObject) {
+    const { data: dataObject } = visualisation;
+    this.validationSchema.validateSync(dataObject);
+  }
+}
+
+export class DraftDashboardItemValidator implements VisualisationValidator {
```
rohan-bes, line 42: I think some folks may take issue with exporting multiple classes from the same file, however personally I think it's fine in this case 👍

rohan-bes (approved):
>I've spent a lot of time trying to get it right in the first place but honestly still don't have a solid idea at this early stage as well (I've tried some different ways but thought they were overengineered/clunky). So I went with the simpler approach which is to have one DashboardVisualisation object and let it accept different validators to mix and match (as I think validators are the main difference between different endpoints). Plus, I think it's the structure that is easiest to be refactored later when we implement more extracting cases

No need to apologise @chris-bes definitely time well worth spent!! I both like the changes you've made, and totally agree with that keeping it simple and anticipating future refactors is the wisest approach. @kael89 has a lot to say on the cost of committing to a bad abstraction, and I think he's very correct about that. Much better for us to keep our option open, rather than spend forever trying to define a structure that isn't entirely apparent to us as of yet 👍 

Super happy with how things are looking right now. I like how you've achieved a nicely decomposed way of validating visualisations for the various cases we may need. Also the general use of typescript and server-boilerplate is looking a lot nicer (tho i still feel there may be benefit in us combing through it at some point?).

Nice work on this one, and I really appreciate you taking the time to think things through and rework this 🎉

packages/admin-panel-server/src/app/createApp.ts
chris-bes: This one is kinda only used for viz-builder at the moment. But I think you're right, there's no reason why it couldn't be use by the general admin-panel app in the future

PR #2885, opened 2021-07-22
Fix lambda scripts for aws api limit

IgorNadj:
### Issue #: No issue

Fixes error:
image


rohan-bes (approved):
Okay, I'll be honest it's pretty hard to parse exactly what's going on here hahaha

The basic gist seems to be that previously instead of looping through elbs we were finding all tags for all elbs using `elbv2.describe_tags` at the start, and this was hitting some limit with the new setup?

From what I can tell the behaviour is still the same as before, and it may be faster to return now cos we exit once we find the first matching arn rather than finding all matches and returning the first in the list?

I guess the proof is in the pudding, I can now delete instances 🎉 Thanks @IgorNadj !!

PR #2887, opened 2021-07-22
3202: Perform Chained Aggregations in Database for Tupaia Data Api

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/3202:

I'd say this is quite close to completion but still flagging as draft since there's tests to fix/write and the functionality ought to be extended somewhat.

**Overview:**
The goal of this feature is to allow multiple aggregation operations all in the database when fetching data via the Tupaia Data Api. The reasons for this are:
a) It's generally a good practice to perform as much computation in the DB as possible (especially things like aggregation)
b) We experience a particular performance hit whenever the db returns a large number of rows of analytics
c) For LESMIS, many reports require first a `FINAL_EACH_YEAR` aggregation before doing the `SUM_PER_PERIOD_PER_ORG_GROUP` aggregation that dramatically reduces the row count

**Implementation:**
We build up an sql query in `AnalyticsFetchQuery.js` that recursively wraps the base analytics fetch with an aggregation on that fetch. This way we can perform as many aggregations as requested (up til the first 'unsupported' aggregation). An example of the sql query looks like this:
```
FETCH
entity: PG, 
period: 2020, 
aggregations: ["FINAL_EACH_WEEK", { "type": "SUM_PER_PERIOD_PER_ORG_GROUP", "config": { "dataSourceEntityType": "facility", "aggregationEntityType": "country" }}]

QUERY
WITH entity_relations_a2 (code, aggregation_entity_code)
AS (VALUES ('PG_MADANG01','PG'), ('PG_MILNEBAY01','PG'), ('PG_MOROBE01','PG'), ('PG_Mugil','PG'), ('PG_NATIONALCAPITALDISTRICT01','PG'), ('PG_NEWIRELAND01','PG'), ('PG_SANDUAN01','PG'), ('PG_SIMBU01','PG'), ('PG_WESTERN02','PG'))

SELECT entity_code AS "entityCode", data_element_code AS "dataElementCode", period, value, type
FROM (

    SELECT aggregation_entity_code as entity_code, data_element_code, sum(value::numeric)::text as value, MAX(type) as type, MAX(day_period) as day_period, MAX(week_period) as week_period, MAX(month_period) as month_period, MAX(year_period) as year_period, MAX(date) as date, MAX(period) as period
    FROM (

        SELECT entity_code as entity_code, data_element_code, most_recent(value, date) as value, MAX(type) as type, MAX(day_period) as day_period, MAX(week_period) as week_period, MAX(month_period) as month_period, MAX(year_period) as year_period, MAX(date) as date, MAX(week_period) as period
        FROM (

            SELECT *, day_period as period from analytics
            INNER JOIN (
                VALUES ('PG_MADANG01'), ('PG_MILNEBAY01'), ('PG_MOROBE01'), ('PG_Mugil'), ('PG_NATIONALCAPITALDISTRICT01'), ('PG_NEWIRELAND01'), ('PG_SANDUAN01'), ('PG_SIMBU01'), ('PG_WESTERN02')
            INNER JOIN (
                VALUES ('BCD1')
            WHERE date >= '2020-01-01T00:00:00.000' AND date <= '2020-12-31T23:59:59.999') as base_analytics

        GROUP BY entity_code, data_element_code, week_period) as a1

    INNER JOIN entity_relations_a2 on entity_relations_a2.code = a1.entity_code
    GROUP BY aggregation_entity_code, data_element_code) as a2

ORDER BY date;

```

So far results are looking promising. This query is able to handle fetching the most recent per school in Laos and aggregating to the country level in ~3 seconds, with reports using such a query loading in 5-10 seconds. However the following tests are yet to be conducted:
- How does it perform when many users are querying at once, does the database go inactive due to so much heavy computation
- How does it perform when many aggregations are used, so far I've tested 3 levels deep (which seemed fine) but in theory it could be many more
- Has performance degraded in the low/no aggregation requests? (if so it shouldn't be hard to have branching logic that runs the old query if 0/1 aggregations are being used)

### Changes:
A few changed were implemented as part of this PR that aren't directly related to the feature but are handy nonetheless:
- Added a `most_recent` aggregation function to the database itself. This makes the sql query _**significantly**_ more simple to read and write. Without it I'd be sceptical that we'd be able to implement chained aggregations are the previous `CROSS LATERAL JOIN` method was rather unruly
- Switched the Analytics table over to using period strings in the `day_period`, `week_period`, etc. columns. This isn't needed at all but makes the analytics table more readable, and opens up options in the future for us to easily perform fetch via query string instead of just startDate/endDate (allowing for dis-continuous period fetching)
- Switched the db codeship service over to postgres 10 (same as what we use in production/feature instances)

Performance benchmarks:
- TBD


codeship-services.yml
```diff
@@ -40,8 +40,8 @@ e2e:
     - /dev/shm:/dev/shm
 
 db:
-  image: mdillon/postgis:9.6-alpine
-  cached: true
+  image: mdillon/postgis:10-alpine
+  cached: false
```
rohan-bes, line 44: Will switch this back to cached as soon as I'm sure it's the correct distro to use 👍

packages/data-api/scripts/createAnalyticsMaterializedView.sql
```diff
@@ -12,10 +12,10 @@ declare
...
       answer.text as value,
       question.type as type,
-      date_trunc(''day'', survey_response.data_time) as "day_period",
-      date_trunc(''week'', survey_response.data_time) as "week_period",
-      date_trunc(''month'', survey_response.data_time) as "month_period",
-      date_trunc(''year'', survey_response.data_time) as "year_period",
+      to_char(survey_response.data_time, ''YYYYMMDD'') as "day_period",
+      concat(extract (year from survey_response.data_time), ''W'', extract (week from survey_response.data_time)) as "week_period",
+      to_char(survey_response.data_time, ''YYYYMM'') as "month_period",
+      to_char(survey_response.data_time, ''YYYY'') as "year_period",
```
rohan-bes, line 18: I've tested this with a full refresh, and the time taken was the same as before (suggesting this string formatting doesn't take any longer than the previous `date_trunc` function). Still need to test with a large survey response upload tho 👍

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -3,119 +3,130 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { DataFetchQuery, ANSWER_SPECIFIC_FIELDS } from './DataFetchQuery';
+import { DataFetchQuery } from './DataFetchQuery';
+
+const VALUE_AGGREGATION_FUNCTIONS = {
+  SUM: 'sum(value::numeric)::text',
+  MOST_RECENT: 'most_recent(value, date)',
```
rohan-bes, line 10: This can be extended for sure! `COUNT`, `AVG` etc.

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -3,119 +3,130 @@
...
-    this.aggregation = { type: firstAggregation?.type };
-    this.aggregation.switches = AGGREGATION_SWITCHES[firstAggregation?.type]; // add internal switches
-    this.aggregation.config = firstAggregation?.config; // add external config, supplied by client
-    this.isAggregating = !!this.aggregation.switches;
+    this.aggregations = [];
+    const originalAggregations = [...options.aggregations];
+    for (let i = 0; i < originalAggregations.length; i++) {
+      const aggregation = originalAggregations[i];
+      if (!AGGREGATION_SWITCHES[aggregation?.type]) {
+        break; // We only support chaining aggregations if all aggregations are supported type
```
rohan-bes, line 60: This comment needs an update, should be `Only chain aggregations up until the first unsupported type`

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -3,119 +3,130 @@
...
+      }
+      const dbAggregation = { type: aggregation?.type };
+      dbAggregation.switches = AGGREGATION_SWITCHES[aggregation?.type]; // add internal switches
+      dbAggregation.config = aggregation?.config; // add external config, supplied by client
+      dbAggregation.stackId = i + 1;
+      this.aggregations.push(dbAggregation);
+      options.aggregations.shift();
```
rohan-bes, line 67: This is new, we are now `unshifting` the aggregation from the options, basically meaning that `Aggregator` now only runs the remaining aggregations. Couple things of note here:
- We now must _**fully**_ support the aggregation in the database level! Aggregator will no longer pick up the slack for us if we do a half assed job of it 😝 
- Mutating the aggregations list here is very 'un functional' of me, and has a huge risk for bugs being introduced around it... I think if we moved to using a proper `Queue` implementation for the aggregation list, and we could `enQueue` and `deQueue` cleanly then I wouldn't have any issues... how does the reviewer feel about me giving that a shot?

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -144,58 +155,43 @@ export class AnalyticsFetchQuery extends DataFetchQuery {
...
       SELECT ${this.getAliasedColumns()}
-      FROM (
-        ${this.getA1Select()}
-        FROM analytics
-          ${this.getA1InnerJoins()}
-          ${this.getA1WhereClause()}
-          ${this.getA1GroupByClause()}
-      ) as a1
-      ${this.getLatestPerPeriodClause()}
+      FROM ${this.aggregations.reduce(wrapAnalyticsInAggregation, baseAnalytics)}
```
rohan-bes, line 194: While this query builder reads to me a lot more simply than the previous iteration, it still has it's tricks about it 😢 Food for a walk-n-talk I think...

packages/database/src/migrations/20210715093133-AddMostRecentAggregationFunction-modifies-schema.js
```diff
@@ -0,0 +1,69 @@
...
+  create FUNCTION most_recent_final_fn(value_and_date) RETURNS text AS $$
+      SELECT $1.value;
+  $$ LANGUAGE SQL; 
+  
+  create AGGREGATE most_recent(text, timestamp) (
+      sfunc = most_recent_fn,
+      stype = value_and_date, 
+      finalfunc = most_recent_final_fn,
+      initcond = '(NULL,-infinity)'
+  );
```
rohan-bes, line 50: This aggregation was so simple to implement in the end that I feel totally embarrassed that I missed it the first time round 🤦 At the time I didn't even stop to think that you can use more than one column in an aggregate function (despite the fact that is obviously happens when doing weighted averages and such)

packages/data-api/src/AnalyticsFetchQuery.js
edmofro: Heh, yes it's a bit of a worry that we have to do things properly! ;-) 

I think the interface could be more encapsulated than that, so that a) we're not nastily mutating this array, but also b) we could plausibly transition to an http interface rather than a functional one in future (i.e. there could be a split somewhere between aggregator and Tupaia's data-api). This would allow having our data-api be a completely different system, and also mean we could potentially make use of other data services' inbuilt aggregation, and have them flag it back to the aggregator in a similar way.

The response from data-broker back to the aggregator is already an object, so there's room to put such information... what do you think?

image

edmofro (commented):
Clever and beautiful!! I love it. It's crazy that it's actually simpler and easier to follow, but so much more powerful.

One thing I found especially delightful is that all the CTEs can sit on their own up the top, so they don't clutter the nested queries, which keeps them semi-readable.

Keen for a walk and talk on this, but from my perspective I have no bones to pick!

packages/data-api/src/AnalyticsFetchQuery.js
rohan-bes: > This would allow having our data-api be a completely different system, and also mean we could potentially make use of other data services' inbuilt aggregation, and have them flag it back to the aggregator in a similar way.

I'm really glad to hear your say this! I've actually been thinking along the lines that it might be nice if our internal datasources (TupaiaDataApi and Indicators) could be run as separate services!

I like that suggestion 👍

rohan-bes:
Closing this one out in favour of https://github.com/beyondessential/tupaia/pull/2979

PR #2888, opened 2021-07-22
3187: Bypass permissions check assertion if responding early

edmofro:
### Issue #:

https://github.com/beyondessential/tupaia-backlog/issues/3187 (but there is also chat about a legitimate typo inside that issue, which is not really related)

If the process prior to the permissions check in import takes longer than the "email instead" timeout, the response telling the user they will be emailed instead isn't able to be sent because we have an assertion that makes sure permissions checking has been handled. This bypasses that assertion, as if we're responding with a "you'll be emailed" message, any permissions issues will go out with the email.

FWIW, the step that takes a long time is 
```
const workbook = xlsx.readFile(req.file.path);
```

So we can't do much to speed that up.


kael89 (approved):
Good solution!

PR #2891, opened 2021-07-23
2729: Extract change batching

kael89:
Implementing the refactoring discussed in https://github.com/beyondessential/tupaia/pull/2803#discussion_r666780953.

I am submitting this PR to get feedback on the architecture. If we agree on that then the following steps would be:
1. Refactor the tests: move generic change handling assertions to a `ChangeHandler.test.js` test file, move the surveyResponse `outdated` calculation tests to the database package
2. Do the refactoring for [EntityHierarchyCacher](https://github.com/beyondessential/tupaia/blob/dev/packages/database/src/cachers/EntityHierarchyCacher.js) too? I am not sure if this would be worth the extra dev time + testing time + regression risk, but I'm slightly leaning towards doing it. One concern is that the batching logic is a bit different than the one used in AnalyticsRefresher/SurveyResponseOutdater (earlier design), but I can give this a try


packages/database/src/changeHandlers/SurveyResponseOutdater.js
```diff
@@ -3,74 +3,264 @@
...
+    case 'delete':
+      // If the deleted record was the most recent response in its dimension combo,
+      // a new one must be selected
+      return oldRecord.outdated ? [] : [oldRecord];
+    default:
+      throw new Error(`${type} is not a supported change type`);
+  }
+};
+
+export class SurveyResponseOutdater extends ChangeHandler {
```
kael89, line 163: The diff shown in this file is not very helpful, but most of the code here comes from https://github.com/beyondessential/tupaia/blob/3c47f0929be04e51ff2296403977d425f4b07a77/packages/database/src/changeHandlers/OutdatedStatusUpdater.js

packages/database/src/changeHandlers/AnalyticsRefresher.js
```diff
@@ -4,71 +4,21 @@
...
-      clearTimeout(this.scheduledRefreshTimeout);
-    }
-
-    if (!this.scheduledRefreshPromise) {
-      this.scheduledRefreshPromise = new Promise(resolve => {
-        this.scheduledRefreshPromiseResolve = resolve;
-      });
-    }
+  constructor(models) {
+    super(models);
```
edmofro, line 13: This line makes me laugh every time

packages/database/src/changeHandlers/ChangeHandler.js
```diff
@@ -0,0 +1,115 @@
...
+  handleChangeQueue = async recordType => {
+    // remove timeout so any changes added now get scheduled anew
+    this.scheduledTimeout = null;
+    this.scheduledPromise = null;
+
+    const changeQueueHandler = this.changeHandlers[recordType];
+    assert.ok(changeQueueHandler);
+
+    await changeQueueHandler(this.changeQueue[recordType]);
+    this.changeQueue[recordType] = [];
```
edmofro, line 112: I believe this could be a race condition, where the listeners could add new records to the change queue while the existing batch of changes is being processed. Those new ones would then be cleared off the change queue when the existing batch finishes processing, despite not having been processed themselves.

Fix could look something like:
```suggestion
    const queuedChanges = this.changeQueue[recordType];
    this.changeQueue[recordType] = [];
    try {
      await changeQueueHandler(queuedChanges);
    } catch (e) {
      this.changeQueue.push(...queuedChanges);
      throw e;
    }
```

packages/database/src/changeHandlers/ChangeHandler.js
```diff
@@ -0,0 +1,115 @@
...
+   * A map of change pushers by record type. Each pusher can return a list of changes that will be
+   * queued and scheduled for handling at a later stage, when changes stop coming through.
+   *
+   * If no pusher is defined for the record type, then the `changeDetails` generated by the db
+   * will be simply added to the change queue
+   *
+   * @protected
+   * @type {Record<string, Function>}
+   */
+  changePushers = {};
```
edmofro, line 19: Would suggest here that the functions aren't actually doing the pushing, they're doing more of a mapping from one format of change details to another

```suggestion
  /**
   * A map of change translators by record type. Each translator can alter the change details that will be
   * queued and scheduled for handling at a later stage, when changes stop coming through.
   *
   * If no translator is defined for the record type, then the `changeDetails` generated by the db
   * will be simply added to the change queue
   *
   * @protected
   * @type {Record<string, Function>}
   */
  changeTranslators = {};
```

packages/database/src/changeHandlers/SurveyResponseOutdater.js
edmofro: Thanks for pointing that out @kael89 - I diffed those locally and the changes are super minimal! Nice to collapse two files into one now the complexity lives in the parent class too :-)

edmofro (changes requested):
This seems well worth doing to me, so thanks @rohan-bes for the suggestion and @kael89 for quickly following up!

The one suggestion I'm not 100% on is what we should call the things that take the change notification details and turn it into the thing that should be processed, so am open to further discussion there.

packages/database/src/changeHandlers/AnalyticsRefresher.js
kael89: Hahaha

packages/database/src/changeHandlers/ChangeHandler.js
kael89: I really like that suggestion, since the actual DB changes are not necessarily the ones pushed into the change queue. For example, if we'd use this base class in `EntityHierarchyCacher`, the items pushed would be hierarchy subtree rebuild jobs.

I tried to emphasise this by contrasting `changeDetails` with `changeQueue`, but "translators" makes it even clearer!

kael89:
How do you feel about not doing point 2 mentioned in the [PR's description](https://github.com/beyondessential/tupaia/pull/2891#issue-695597323) at this stage @edmofro ? I am a bit nervous of the extra testing that would be required, and feel that we probably shouldn't delay further delivering this feature to our users

The code changes wouldn't take long I believe, also you've put together some pretty thorough tests for entity hierarchy caching, but we could still miss something and it increase the testing scope

packages/database/src/changeHandlers/ChangeHandler.js
kael89: Awesome observation @edmofro !! You are really the async queue processing master ⏳🐱‍👤💯

edmofro:
Yeah I'm happy with that @kael89. After all, this is already bonus refactoring that you're doing 😛

packages/database/src/changeHandlers/ChangeHandler.js
kael89: One potential issue here @edmofro is that if a batch of changes fails for a fundamental reason (eg error in code), we'll keep on pushing the same changes in the queue and the handling will fail again and again

I guess a solution would be implementing a more elaborate queue management system (similar to `dhis_sync_queue`)... It does sounds a bit overengineered for what we want to achieve now, but ignoring those changes completely is not ideal either.

kael89: I'm thinking that this is an existing issue of handling change notifications: if the information is not persisted in the DB, we'll have no way of knowing that a change handling failed (and needs to run again) apart from checking the logs.

One solution would be adding a `change_handler_success` column in `survey_response`, which is set to `false` if something goes wrong in a change handler. We can then ignore the records whose change handling failed, and try to re-handle them in the future after the issue is fixed. This is a simplistic approach which assumes that only one change handler is defined for `survey_response`. It requires refactoring the change translators to return a map of db records to change queue items.

I'm not a huge fan of the above, but may be a simple compromise between simplicity and accuracy. Of course, the other option is ignoring this potential danger for now, and deal with bugs if and when they arise 🔥

edmofro: I'd go for ignoring the potential danger, and letting the change in the queue fail again and again. The mitigation for that would be something we need to do anyway: logging to an external service that will do things like group logs and allow us to explore them easily, and potentially notify us in unexpected situations. I believe @IgorNadj has started some work looking at external logging services?

IgorNadj: Yep we have https://github.com/beyondessential/tupaia-backlog/issues/2003. I never got to it this sprint. Still something we should do but will take a good few days to set up properly.

kael89: We were discussing about this with @IgorNadj and he suggested the following: for each failed batch, re-try a max number of times (say 3), if the batch fails then log some info and stop retrying

With that in mind, I suggest using a `logChangeInfo` to log debugging info when we keep re-trying for a failed batch. By default it would just log the count of changes in a batch. It can be overridden in child classes to extract more useful debugging info. For example, `SurveyResponseOutdater` could extract the record ids of survey responses in a failed batch.

Thoughts @edmofro ?

edmofro: It sounds like a good pattern to me. Obviously it would also be great to get you onto viz builder as soon as we can, as it's just Chris once Tom moves back to NZ after tomorrow, so please don't spend too long on it, but what you suggest sounds like it shouldn't take too long :-)

kael89: It is what it is - I agree with the priorities we have set and ideally we wouldn't want to spend too much time on this. On the other hand, we've been burned by invalid DB state creeping without us being aware (eg DHIS responses that were not added in the sync queue), so I think we should be careful given this will affect the core of our system

PR #2896, opened 2021-07-25
3232: Switch LESMIS Vitals reports to use Entity Aggregation

rohan-bes:
### Changes:
- Added entity-aggregation to lesmis vitals reports
- Stopped sending through descendants to ReportRoute when fetching vitals data


packages/lesmis-server/src/routes/ReportRoute.ts
```diff
@@ -58,10 +59,10 @@ export class ReportRoute extends Route {
           {
             // Report server can accept arrays so the parameters are plural
             organisationUnitCodes: entityCode,
-            projectCodes: LESMIS_PROJECT_NAME,
+            hierarchy: LESMIS_HIERARCHY_NAME,
             ...this.req.query,
           },
-          this.req.body,
+          {},
```
rohan-bes, line 65: Body is no longer required, but it will be removed as part of a different PR: https://github.com/beyondessential/tupaia/pull/2866

packages/lesmis/src/api/queries/useVitalsData.js
```diff
@@ -87,32 +87,30 @@ const useDistrictReport = entity =>
...
     'LESMIS_multi_school_vitals',
     {
-      data: {
+      params: {
         endDate: utcMoment().format(endDateFormat),
-        organisationUnitCodes: decendants.join(),
       },
     },
-    decendants.length > 0,
+    descendants.length > 0,
```
rohan-bes, line 99: While entity aggregation should be able to handle this (ie. it will just return no data if requested for an entity that has no school descendants), I figured it's still a good practice to keep these checks on the front-end to avoid redundant calls to report-server?

EMcQ-BES: I think this is fine for now. We probably want to remove it long term because processing these hierarchies on the front end causes a fair bit of slowdown (plus I think it's generally against best practices to do this much processing in the front end).

But I think this work is already above and beyond the scope required by you so I'm happy to leave that for team nz to deal with down the line.

EMcQ-BES (approved):
Looks great, thanks for doing this @rohan-bes

PR #2898, opened 2021-07-26
3177 delete duplicate countries

avaek:
Addresses: https://github.com/beyondessential/tupaia-backlog/issues/3177

This basically deletes any duplicate permissions (where the same permission exists for both the duplicate & real country) and updates all the remaining permissions for the duplicate countries to be for their equivalent real country instead
- I add a "note" to the `note` column regarding changes this migration makes to the `access_request` table, I'm not sure if I should do this but I think this `note` column is for internal BES use only so I added the note just in case it is ever useful in troubleshooting in future

Once the permissions are cleaned up the duplicate countries in the `entity` table are able to be deleted


packages/database/src/migrations/20210714013524-DeleteDuplicateCountriesFromEntityTable-modifies-data.js
```diff
@@ -0,0 +1,147 @@
...
+  "'60949c6061f76a622c020231'", // duplicate Samoa
+];
+
+const realCountryEntityIds = [
+  "'5d51f501f013d6320f3cf633'", // real Fiji
+  "'5d3f8844f327a531bfd8ad77'", // real Papua New Guinea
+  "'5df1b88c61f76a485cd1ca09'", // real Samoa
+];
+
+const findMatchingCountries = (duplicateId, realId) => {
```
edmofro, line 31:
```suggestion
const checkForCountryMatch = (duplicateId, realId) => {
```

packages/database/src/migrations/20210714013524-DeleteDuplicateCountriesFromEntityTable-modifies-data.js
```diff
@@ -0,0 +1,147 @@
...
+  "'60da867061f76a20cf01e23f'", // duplicate Fiji
+  "'60eb91df61f76a7f770790a3'", // duplicate Papua New Guinea
+  "'60949c6061f76a622c020231'", // duplicate Samoa
+];
+
+const realCountryEntityIds = [
+  "'5d51f501f013d6320f3cf633'", // real Fiji
+  "'5d3f8844f327a531bfd8ad77'", // real Papua New Guinea
+  "'5df1b88c61f76a485cd1ca09'", // real Samoa
+];
```
edmofro, line 29: If you store an object mapping the country code to the id, you can then refer to that in all the places you currently have an id directly in the code (which is a no-no, because it's so much easier to get an id wrong or not know what country an id relates to compared with a named variable or using the code).

```suggestion
const duplicateEntityIdByCountry = {
  FJ: "'60da867061f76a20cf01e23f'", // duplicate Fiji
  PG: "'60eb91df61f76a7f770790a3'", // duplicate Papua New Guinea
  WS: "'60949c6061f76a622c020231'", // duplicate Samoa
};

const realEntityIdByCountry = {
  FJ: "'5d51f501f013d6320f3cf633'", // real Fiji
  PG: "'5d3f8844f327a531bfd8ad77'", // real Papua New Guinea
  WS: "'5df1b88c61f76a485cd1ca09'", // real Samoa
};
```

You can still get the arrays you have currently whereever you need them, using `Object.values`

```
const duplicateCountryEntityIds = Object.values(duplicateEntityIdByCountry);
const realCountryEntityIds = Object.values(realEntityIdByCountry);
```

packages/database/src/migrations/20210714013524-DeleteDuplicateCountriesFromEntityTable-modifies-data.js
```diff
@@ -0,0 +1,147 @@
...
+
+const findMatchingCountries = (duplicateId, realId) => {
+  // checks if both are Fiji entities
+  if (duplicateId === "'60da867061f76a20cf01e23f'" && realId === "'5d51f501f013d6320f3cf633'") {
```
edmofro, line 33: Following the earlier suggestion, this would become

```suggestion
  if (duplicateId === duplicateEntityIdByCountry.FJ && realId === realEntityIdByCountry.FJ) {
```

packages/database/src/migrations/20210714013524-DeleteDuplicateCountriesFromEntityTable-modifies-data.js
```diff
@@ -0,0 +1,147 @@
...
+  DELETE FROM user_entity_permission
+    WHERE id IN (${arrayToDbString(permissionIdsToDelete)});
+  `);
+
+  // updates remaining Fiji permissions
+  await db.runSql(`
+  UPDATE user_entity_permission
+    SET entity_id = '5d51f501f013d6320f3cf633'
+      WHERE entity_id = '60da867061f76a20cf01e23f';
+  `);
```
edmofro, line 94: Level one upgrade:

```suggestion
  // updates remaining Fiji permissions
  await db.runSql(`
  UPDATE user_entity_permission
    SET entity_id = ${realEntityIdByCountry.FJ}
      WHERE entity_id = ${duplicateEntityIdByCountry.FJ};
  `);
```

Level two upgrade - do all three countries in one code block using a loop:
```suggestion
  // updates remaining entity permissions
  for (const countryCode of Object.keys(realEntityIdByCountry)) {
    await db.runSql(`
      UPDATE user_entity_permission
      SET entity_id = ${realEntityIdByCountry[countryCode]}
        WHERE entity_id = ${duplicateEntityIdByCountry[countryCode]};
    `);
  }
```

edmofro (changes requested):
Nice! It looks like it should all work logically, but we just need to clean up the ids that are in the code. Ideally we wouldn't have ids showing up in code at all, and instead would just fetch them from the db using the code, but as I said when working on it together it's fine to have a few ids so long as they're well commented and will be the same ids used across all tupaia deployments (which is the case here)

avaek:
Thanks for that detailed & educational review @edmofro - I had a feeling something like that was possible but I wasn't sure of the exact syntax

I've implemented those suggestions and also put my changes to the `access_request` table inside the loop that you outlined for the `user_entity_permission` table above

edmofro (approved):
Looks great!

PR #2899, opened 2021-07-26
Fix dump database script - dump location

kael89:
After I recently move those scripts under the `packages/database` folder, the `dumpDatabase` script will download the dump file in that location.

The previous behaviour was downloading the dump in the monorepo root by default. This is also assumed by the `yarn refresh-database` script defined in the root `package.json`, which expects the dump to be in the root. This script re-instates the previous behaviour regarding dump default locations

### DIFF
--- a/packages/devops/scripts/ci/validateBranchName.sh
+++ b/packages/devops/scripts/ci/validateBranchName.sh
@@ -60,7 +60,6 @@ function validate_name_chars() {
 }
 
 branch_name=$(get_branch_name)
-check_name_is_not_reserved $branch_name
 validate_name_ending $branch_name
 validate_name_length $branch_name
 validate_name_chars $branch_name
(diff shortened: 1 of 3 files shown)


packages/devops/scripts/ci/validateBranchName.sh
```diff
@@ -60,7 +60,6 @@ function validate_name_chars() {
 }
 
 branch_name=$(get_branch_name)
-check_name_is_not_reserved $branch_name
```
edmofro, line 63: Why this change?

edmofro (commented):
This happened to me and I didn't realise why, just thought I must have called the script from the wrong directory! Thanks for fixing. Just the question about the branch name validation, otherwise it all looks good.

packages/devops/scripts/ci/validateBranchName.sh
kael89: Fixes an existing (non breaking issue) - the function `check_name_is_not_reserved ` was removed, but its call was not.

This function used to ensure that the branch was not `e2e`, now we maintain that branch (clone of `master`)

edmofro: Gotcha

edmofro (approved):

PR #2901, opened 2021-07-26
E2e config updates

kael89:
Good news - it seems that a lot of Laos Schools e2e test timeouts were caused by broken urls! This PR fixes those


packages/web-frontend/cypress/config/dashboardReportUrls/default.json
```diff
@@ -324,39 +324,39 @@
     "dashboard": "WASH"
   },
   {
-    "code": "LESMIS_repetition_rate_primary_district",
+    "code": "LESMIS_age_of_grade1_entrance",
```
kael89, line 327: Apart from fixing wrong dashboard names, I sorted some items that were out of order.

The implied order in config files is by ascending `code` - this generally makes it easier to compare with other configs, find what is/is not included, which org unit variations are used for the same code etc

packages/web-frontend/cypress/scripts/generateConfig/generateOverlayConfig.js
```diff
@@ -49,7 +49,10 @@ const parseUrl = url => {
 
   const [path, queryParams] = url.split('?');
   const searchParams = new URLSearchParams(queryParams);
-  const [project, orgUnit] = path.split('/').filter(x => x !== '');
+  const [project, orgUnit] = path
+    .split('/')
+    .filter(x => x !== '')
+    .map(decodeURIComponent);
```
kael89, line 55: @BinduMadhaviTammireddy this is a code (and not config) related change. It fixes a minor bug that occurs if you use urls that include encode HTML characters, eg `/laos_schools/LA_Phouvong%20District/Schools?report=LESMIS_num_lower_secondary_classrooms&reportPeriod=1st_Jan_2021-31st_Dec_202` (`%20` is an encoded space)

It's a small change but let me know if you think another dev should have a look at this one!

packages/web-frontend/cypress/config/dashboardReportUrls/lesmis.json
```diff
@@ -1,25 +1,25 @@
...
     "endDate": "2021-12-31"
   },
   {
-    "code": "LESMIS_repetition_rate_upperSecondary_district",
+    "code": "LESMIS_age_of_grade1_entrance",
     "project": "laos_schools",
-    "orgUnit": "LA_Phouvong District",
-    "dashboard": "Student Outcomes",
+    "orgUnit": "LA_Attapue",
+    "dashboard": "Student%20Enrolment",
```
BinduMadhaviTammireddy, line 14: In the above two json configs one of the `"dashboard": "Student%20Enrolment",` one of the dashboard's have `20%` while the other does not any particular reason?

packages/web-frontend/cypress/scripts/generateConfig/generateOverlayConfig.js
BinduMadhaviTammireddy: thank you, it was informative.

BinduMadhaviTammireddy (approved):

packages/web-frontend/cypress/config/dashboardReportUrls/lesmis.json
kael89: Oops good catch Bindu - `20%` is an encoded html space. Both should work but using a space is more human-friendly, so I'm gonna change that

PR #2904, opened 2021-07-27
3202: Added 'most_recent' aggregation function to the database

rohan-bes:
### Issue https://github.com/beyondessential/tupaia-backlog/issues/3202:

Part of: https://github.com/beyondessential/tupaia/pull/2887
Can be merged independently

Here we're adding a new aggregation function to the database. In order to ensure the tests pass, I've also updated the docker image we use for the database during tests to be version 10 of postgres (the same as we use in deployment instances).

- [x] Test that when merging this branch to a branch that has the older version of postgres, the image updates and gets cached correctly

### DIFF
--- a/codeship-services.yml
+++ b/codeship-services.yml
@@ -40,7 +40,7 @@ e2e:
     - /dev/shm:/dev/shm
 
 db:
-  image: mdillon/postgis:9.6-alpine
+  image: mdillon/postgis:10-alpine
   cached: true
   encrypted_dockercfg_path: dockercfg.encrypted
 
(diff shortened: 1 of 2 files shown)


codeship-services.yml
```diff
@@ -40,7 +40,7 @@ e2e:
     - /dev/shm:/dev/shm
 
 db:
-  image: mdillon/postgis:9.6-alpine
+  image: mdillon/postgis:10-alpine
```
rohan-bes, line 43: The migration below was getting a syntax error with the older version of postgres we were previously using. Version 10 is the one we use in prod/feature instances so I went with that 👍

edmofro (approved):
lgtm, had already reviewed in a draft PR so only skimmed this time around :-)

PR #2906, opened 2021-07-27
3233: Handle session expired in Admin Panel

chris-bes:
### Issue: 
https://github.com/beyondessential/tupaia-backlog/issues/3233
https://linear.app/tupaia/issue/MEL-373/

### Changes:

- Handle session expired in Admin Panel


packages/server-boilerplate/src/orchestrator/session/attachSession.ts
```diff
@@ -11,7 +11,7 @@ export const attachSession = async (req: Request, res: Response, next: NextFunct
   try {
     const sessionId = req.sessionCookie?.id;
     if (!sessionId) {
-      throw new UnauthenticatedError('User not authenticated');
+      throw new UnauthenticatedError('Session not found or has expired. Please log in again.');
```
chris-bes, line 14: Tom suggested a nice idea to which is to use a new http response code to indicate that the session has expired (https://github.com/beyondessential/tupaia-backlog/issues/3233#issuecomment-887094229). However, I think with the case of session cookie not found, it could be because of some other reasons rather than just the session has expired (eg: someone unauthenticated trying to access the server). So I'm still keeping 401 as the response code here.

chris-bes, line 14: Tom suggested a nice idea to which is to use a new http response code to indicate that the session has expired (beyondessential/tupaia-backlog#3233 (comment)). However, I think with the case of session cookie not found, it could be because of some other reasons rather than just the session has expired (eg: someone unauthenticated trying to access the server). So I'm still keeping 401 as the response code here.

packages/admin-panel/src/api/TupaiaApi.js
```diff
@@ -77,9 +79,21 @@ export class TupaiaApi {
     };
   }
 
+  async checkIfAuthorized(response) {
+    // Unauthorized
+    if (response.status === 401) {
```
tcaiger, line 84: Initially I was wondering whether it is too aggressive to always log out users when there is a 401 but after thinking about it for a while I agree that it makes good sense.

tcaiger (approved):
Looks good. Thanks for handling this. I'm wondering if there is a similar change that needs to happen in psss? If so, would you mind creating a card for it? I don't want to hold up this change and the psss fix is quite a low priority.

chris-bes:
> Looks good. Thanks for handling this. I'm wondering if there is a similar change that needs to happen in psss? If so, would you mind creating a card for it? I don't want to hold up this change and the psss fix is quite a low priority.

Yep, PSSS already log people out when response code is 401. However, now that you mentioned, I've realized the error message if that happens is still `User not authenticated`. Will change to the same one here!

PR #2908, opened 2021-07-27
1550 web config server replication

IgorNadj:


packages/devops/scripts/ci/deployBackend.sh
```diff
@@ -7,7 +7,15 @@ DEPLOYMENT_SSH_URL=$(${DIR}/determineDeploymentSshUrl.sh)
 DEPLOYMENT_URL=$DEPLOYMENT_SSH_URL # ssh url will resolve to tupaia web frontend desktop over HTTP
 if curl --output /dev/null --silent --head --fail $DEPLOYMENT_URL; then
     echo "Deployment for ${CI_BRANCH} exists, updating with latest changes"
-    ssh -o ServerAliveInterval=15 ubuntu@$DEPLOYMENT_SSH_URL "cd tupaia/packages/${PACKAGE} && yarn build && pm2 delete ${PACKAGE} || true && pm2 start --name ${PACKAGE} dist --wait-ready --listen-timeout 15000 --time;"
+
+    REPLICATION_PM2_CONFIG=''
+    if [[ $PACKAGE == "web-config-server" ]]; then
+        # as many replicas as cpu cores - 1
+        REPLICATION_PM2_CONFIG='-i -1'
```
IgorNadj, line 14: N-1 leaves 1 core for kernel/other processes. If web-config-server gets overloaded other processes can still run.

rohan-bes (approved):
Looks pretty good to me! Excited to see how things go using cluster mode 👍 

One general point of concern for web-config-server would be that it's still reliant directly on the entityModel to fetch entity-relations (for org unit hierarchy stuff and legacy dashboards that do entity aggregation). This model does do some caching logic for relations, and when things are uncached it can run a bit slow. Each process will have its own separate cache, so there's a chance that things will be a bit bumpy once in a while? I believe the cache is persistent atleast, so over time it should all smooth out 👍 

I appreciate that this is just an investigation, but a bit curious as to why we picked the web-config-server to be the server to scale up?

IgorNadj:
Thanks for that @rohan-bes 

This is yeah, the first step. Good catch on the caching issue, sounds like it should work itself out. We chose web-config-server because it is the bottleneck in the flu tracking load testing we simulated in #2995. I also left a comment here https://github.com/beyondessential/tupaia-backlog/issues/1550#issuecomment-887287603 about future replication tuning

PR #2914, opened 2021-07-28
VizBuilder Create New Page and Metadata Form

tcaiger:
### Issue
- Includes Permission Group Dropdown https://github.com/beyondessential/tupaia-backlog/issues/3049
And other work not covered in tickets for editing the metadata for visualisations

### Changes:

- Add CreateNew and Main screens
- Add Metadata form for editing viz meta data
- Use meta data form in CreateNew screen and EditModal
- Tweak page layout by making containers wider

---

### Screenshots:
Added [here](https://app.zenhub.com/workspaces/active-sprints-5eea9d3de8519e0019186490/issues/beyondessential/tupaia-backlog/3102)


packages/admin-panel/src/VizBuilderApp/views/Main.js
```diff
@@ -0,0 +1,40 @@
...
+`;
+
+// Todo: add https://github.com/jacobbuck/react-beforeunload#readme
+export const Main = () => {
+  const [enabled, setEnabled] = useState(false);
+  return (
+    <>
+      <Toolbar />
+      <Container maxWidth="xl">
+        {/*<Panel setEnabled={setEnabled} />*/}
```
tcaiger, line 31: These components are coming soon

packages/ui-components/src/components/Dialog.js
```diff
@@ -30,7 +30,7 @@ export const Dialog = ({ children, ...props }) => (
 );
 
 Dialog.propTypes = {
-  children: PropTypes.array.isRequired,
+  children: PropTypes.any.isRequired,
```
tcaiger, line 33: Children can be a lot of things. Array, Object, Function, Node, Null, Undefined etc

packages/admin-panel/src/VizBuilderApp/components/MetadataForm.js
```diff
@@ -0,0 +1,94 @@
...
+          placeholder="Select Project"
+          defaultValue={project}
+          options={projects.map(p => p.code)}
+          disabled={isLoadingProjects}
+          error={!!errors.project}
+          helperText={errors.project && errors.project.message}
+          inputRef={register({
+            required: 'Required',
+          })}
+        />
```
chris-bes, line 50: I think we're actually be able to save multi projects for a visualisation now. Hopefully this can be easily changed to use `allowMultipleValues` like in Admin-Panel?

chris-bes (approved):
LGTM overall! Just a minor request to allow selecting multiple projects. It's already end of sprint so if you don't have time it's fine.

packages/admin-panel/src/VizBuilderApp/components/MetadataForm.js
tcaiger: hmm that's a good point. Setting multiple projects would make it more difficult to display a list of locations though. I had a similar question to Rohan about whether or not we should allow multiple locations to be set on a viz and he thought it wasn't a priority for now. @rohan-bes - do you think this is a similar feature? In other words, can we ignore setting multiple projects on a viz for now?

packages/admin-panel/src/VizBuilderApp/components/MetadataForm.js
```diff
@@ -0,0 +1,94 @@
...
+        <TextField
+          name="name"
+          label="Name"
+          defaultValue={name}
+          error={!!errors.name}
+          helperText={errors.name && errors.name.message}
+          inputRef={register({
+            required: 'Required',
+          })}
+        />
```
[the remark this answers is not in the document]
tcaiger: That is easy enough to add. @rohan-bes - do we want to add a text field for code to the form? Or should the code be generated somehow?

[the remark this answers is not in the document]
rohan-bes: Let's add a form input I think. Auto generated sounds like it could be a cool future feature but for now I think manually entering it will be easiest and most intuitive

packages/admin-panel/src/VizBuilderApp/components/MetadataForm.js
rohan-bes: Might need to touch base with you guys on this on monday to confirm we all have a common understanding. As far as I'm aware, the 'Select Project' input is sort of a stub thing for now, as we don't actually associate a dashboard item / report with a project in any way. Whether a dashboard item can be seen in a project is entirely dependant on what dashboard relations the dashboard item has, and we aren't configuring those in the Viz Builder (as of yet).
What this selector (and the one in the main Viz Builder UI) really achieve is a way for determining which hierarchy to use for the preview data.
Regardless I think having this selector is good, and forcing the user to enter something is nice too, as it is a real business rule and even it isn't truly enforced on a technical level. @chris-bes makes a good point that multiple project dashboard items are supported, and aren't even uncommon, but given that this selector here is sort of a stub I'm okay with it being a single project selector for now 👍 

I think implementing a way to save dashboard relations as part of the viz builder is a pretty high priority next feature, and adding multi project support at that point would be nice 👍 

Happy to chat with yall on this one though to get your thoughts on it?

PR #2916, opened 2021-07-28
NZ-801: Data Service Sync Groups

EMcQ-BES:
### Issue #:
NZ-801
NZ-799
resolves NZ-66

### Changes:

- Remove previous migration for FQS mapping
- Add `data_service_sync_group` table
- Add `FQS1` and `FQS2` to new table
- Add sync group support to data-broker
- Update KoBo translator to pull from new table
- Update KoBo service to pull from multiple surveys at once


packages/database/src/migrations/20210728020128-AddFQSSurveySyncGroups-modifies-data.js
```diff
@@ -0,0 +1,59 @@
...
+];
+
+exports.up = async function (db) {
+  for (const { code, entityQuestionCode, questionCodeMapping } of KOBO_SURVEYS) {
+    await insertObject(db, 'data_service_sync_group', {
+      id: generateId(),
+      code,
+      service_type: 'kobo',
+      config: {
+        // koboSurveyCode: 'xyz',
```
EMcQ-BES, line 43: Figured I'd keep these codes as a manual step. Feels like something that shouldn't be committed to the repo

packages/database/src/modelClasses/DataSource.js
```diff
@@ -9,9 +9,11 @@ import { TYPES } from '../types';
 
 const DATA_ELEMENT = 'dataElement';
 const DATA_GROUP = 'dataGroup';
+const SYNC_GROUP = 'syncGroup';
 const DATA_SOURCE_TYPES = {
   DATA_ELEMENT,
   DATA_GROUP,
+  SYNC_GROUP,
```
EMcQ-BES, line 16: I'm not 100% on adding this to the data source types. This means that technically the model doesn't match the db, since I haven't added `syncGroup` as a valid type in the `data_source` table. But we are conceptually using it as another data source type.

packages/data-broker/src/services/indicator/IndicatorService.js
```diff
@@ -21,10 +21,16 @@ export class IndicatorService extends Service {
   }
 
   async pull(dataSources, type, options) {
-    if (type === this.dataSourceTypes.DATA_GROUP) {
-      throw new Error('Event pulling is not supported in IndicatorService');
+    switch (type) {
```
EMcQ-BES, line 24: I think this is the only service that needs to be updated to handle `syncGroup`

`DhisService` and `TupaiaService` both map the data source type to functions so should just error on a newly added type. (There's already precedence for this with `DhisService` only having one metadata puller);

`WeatherService` checks directly if type is `DATA_ELEMENT` or `DATA_GROUP` so will error if the type doesn't match.

packages/database/src/migrations/20210728020128-AddFQSSurveySyncGroups-modifies-data.js
```diff
@@ -0,0 +1,67 @@
...
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const KOBO_SURVEYS = [
+  {
+    code: 'FQS1',
+    entityQuestionCode: 'FQS1Primary_location/FQS1PrimarySchool',
+    questionCodeMapping: FQS1_CODE_MAPPING,
+    internalSurveyCode: 'LESMIS_FQS1',
```
edmofro, line 28: I'd like to keep some conceptual consistency with the other data source records, where the base `code` is internal. Is your worry here that we might sync multiple external surveys into a single internal survey?

```suggestion
    code: 'LESMIS_FQS1',
    entityQuestionCode: 'FQS1Primary_location/FQS1PrimarySchool',
    questionCodeMapping: FQS1_CODE_MAPPING,
    externalSurveyCode: 'FQS1',
```

edmofro: Oh ok, I see the comment further down now - looks like the base `code` is just for the specific record, and both the internal and external codes are inside the config?

packages/database/src/migrations/20210728020128-AddFQSSurveySyncGroups-modifies-data.js
edmofro: Hmm I would be pretty comfortable committing it to the repo, it seems to be in the same class of info as the question code mapping. Makes release process easier!

edmofro (changes requested):
My comments are all pretty superficial, this is looking good!

packages/database/src/modelClasses/DataSource.js
EMcQ-BES: The way I see it there are two options for what we want here, and both of them currently involve some single use special handling.

Either we're passing the `DataBroker` the data source type, i.e. `dataElement, dataGroup, syncGroup`, and there's special handling that says `syncGroup` is pulled from a separate database table. (Which is the current implementation)

I'm actually coming around on this being here, because no matter where we put the definition of a sync group type, we're still passing data broker two data sources and a pseudo data source. I think anywhere else we put it would make the code messier for no real gain.

Alternatively we could update `DataBroker` to accept the database type, i.e. `data_source, data_service_sync_group` and there's special handling that says `data_source` can be either a `dataElement` or a `dataGroup`

This would probably be better long term if we want to split `dataElement` and `dataGroup` to separate tables, but would require a lot of refactoring which I think is outside the scope of this task.

edmofro: If Ethan says something is outside the scope of the current task, I'd better listen ;-)

Happy to go with what you've got already. Long term we do want to split `dataElement` and `dataGroup`, but who knows how far away that will be.

PR #2921, opened 2021-07-28
2729: Fix issues and update tests MEL-92

kael89:
Continuing from https://github.com/beyondessential/tupaia/pull/2891 and hopefully last part of https://github.com/beyondessential/tupaia-backlog/issues/2729

I could commit directly in the base branch, but I thought that reviewing separately may be easier. Let me know if this is not the case, and I can merge this in the base branch


packages/meditrak-server/src/database/models/Survey.js
```diff
@@ -68,7 +68,7 @@ const onChangeUpdateDataGroup = async (
       const { code, data_source_id: dataSourceId } = newRecord;
       await models.dataSource.updateById(dataSourceId, { code });
 
-      if (newRecord.period_granularity !== oldRecord.period_granularity) {
+      if (oldRecord && newRecord.period_granularity !== oldRecord.period_granularity) {
```
edmofro, line 71: Nice catch

packages/database/src/changeHandlers/ChangeHandler.js
```diff
@@ -35,17 +25,16 @@ export class ChangeHandler {
    */
   debounceTime = 1000; // ms
 
-  /**
-   * @type {Record<string, Array>} Changes by record type
-   */
-  changeQueue = {};
+  changeQueue = [];
```
kael89, line 28: Thanks to the tests, I realised that while I was using a single promise for changes of all types, I was handling only one type of change before fulfilling that promise. I decided to allow
* separate translators per record type
* a single handler for changes of all types

The idea is that records in multiple tables can trigger the same type of change. This is pattern is already used in `AnalyticsRefresher` and `EntityHierarchyCacher`. Future classes could still look on the translated details to define a custom behaviour based on each change's record type, if need be.

packages/database/src/tests/changeHandlers/AnalyticsRefresher.test.js
```diff
@@ -192,98 +190,4 @@ describe('AnalyticsRefresher', () => {
     });
     await assertAnalyticsMatch(updatedAnalytics);
   });
-
-  it('debounces analytics table refreshes that are started at the same time', async () => {
```
kael89, line 196: These very handy test cases were moved in `ChangeHandler.test.js`

packages/meditrak-server/src/database/models/Survey.js
kael89, line 71: Another bug caught by the tests

packages/database/src/tests/changeHandlers/SurveyResponseOutdater.test.js
```diff
@@ -3,66 +3,504 @@
...
+const yearlySurveyId = generateTestId();
+const quarterlySurveyId = generateTestId();
+const monthlySurveyId = generateTestId();
+const weeklySurveyId = generateTestId();
+const dailySurveyId = generateTestId();
+const nonPeriodicSurveyId = generateTestId();
+
+const userId = generateTestId();
+
+const SURVEYS = {
```
kael89, line 40: I moved the test cases from `testOutdatedStatusUpdate.js` in this file, and moved the existing test cases (which asserted generic change handling behaviour) in `ChangeHandler.test.js`

packages/database/src/testUtilities/upsertDummyRecord.js
```diff
@@ -79,11 +81,17 @@ export const findOrCreateDummyRecord = async (model, findCriteria, data) => {
 };
 
 export const findOrCreateRecords = async (models, recordsByType) => {
+  const data = [];
```
edmofro, line 84:
```suggestion
  const data = {};
```
Same below - I think!

packages/database/src/changeHandlers/ChangeHandler.js
```diff
@@ -94,22 +96,48 @@ export class ChangeHandler {
...
+
+  logFailedChanges = failedChanges => {
+    winston.error(
+      [
+        `Failed to handle change batch after trying ${MAX_RETRY_ATTEMPTS} times`,
+        'Debugging info:',
+        this.getChangeDebuggingInfo(failedChanges),
+      ].join('\n'),
+    );
+  };
```
edmofro, line 142: This came out nice 🎉

packages/database/src/tests/changeHandlers/ChangeHandler.test.js
```diff
@@ -0,0 +1,250 @@
...
+
+      this.changeTranslators = {
+        project: sinon.stub().callsFake(changeDetails => [changeDetails.new_record?.id]),
+        user: sinon.stub().callsFake(changeDetails => [changeDetails.new_record?.id]),
+      };
+    }
+
+    getChangeDebuggingInfo = changes => `Failed ids: ${changes}`;
+
+    handleChanges = sinon.stub().resolves();
```
edmofro, line 31:
```suggestion
      this.resetMocks();
    }

    getChangeDebuggingInfo = changes => `Failed ids: ${changes}`;

```

packages/database/src/tests/changeHandlers/ChangeHandler.test.js
```diff
@@ -0,0 +1,250 @@
...
+  });
+
+  it('handles multiple changes in batches', async () => {
+    // make a bunch of different changes, with small delays between each to model real life
+    const sleepTime = DEBOUNCE_TIME / 2;
+
+    // create
+    const projectIds = [];
+    for (let i = 0; i < 3; i++) {
+      const projects = await upsertDummyRecord(models.project);
```
edmofro, line 87: Good call making these tests go deeper and integrate the db notifications

edmofro (approved):
I really like the new structure not needing to define both `pushers` and `handlers` - `translators` alone is enough when the handler method can be overridden 🎉 

Nice tests too! I didn't review SurveyResponseOutdater.test.js line by line as it was over 500 long, but the way the tests are structured is really nice (matches previous structure in the old file) and the test case descriptions all look good :-)

Only one required change, unless I'm missing something: `[]` -> `{}`. With that minor thing in mind, I'm approving this PR 👍

packages/database/src/testUtilities/upsertDummyRecord.js
kael89: Oops yeah - great catch! 👏

packages/database/src/tests/changeHandlers/ChangeHandler.test.js
kael89: Good suggestion!

PR #2931, opened 2021-07-29
Rename viz builder store

tcaiger:
### Changes:
This is just a refactor to rename the store in the vizBuilder app to be `vizBuilderConfigStore` and change its api to expose `VizBuilderConfigProvider ` and `useVizBuilderConfig ` to signpost to developers the purpose of the store.

### DIFF
--- a/packages/admin-panel/src/rootReducer.js
+++ b/packages/admin-panel/src/rootReducer.js
@@ -5,19 +5,19 @@
 import { combineReducers } from 'redux';
 import { reducer as authentication, LOGOUT } from './authentication';
 import { reducer as tables } from './table';
-import { reducer as autocomplete } from './autocomplete';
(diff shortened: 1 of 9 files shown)


packages/admin-panel/src/rootReducer.js
```diff
@@ -5,19 +5,19 @@
 import { combineReducers } from 'redux';
 import { reducer as authentication, LOGOUT } from './authentication';
 import { reducer as tables } from './table';
-import { reducer as autocomplete } from './autocomplete';
+import { reducer as autocomplete } from './autocomplete/reducer'; // Needs to be imported from reducer file or console shows autocomplete not found error
```
tcaiger, line 8: This is slightly un-related but I noticed an error in the console and for some reason adding the import path fixes it.

kael89 (approved):
Looks good!

PR #2934, opened 2021-07-30
NZ-812: Use more conventional route handling in meditrak-server

edmofro:
### Issue #:

resolves NZ-812
foundation towards NZ-803

This touches a lot of files, but is mostly moving things around. Still, I've split the commits out so they can be reviewed one by one, which might make it easier to spot the significant changes. You might want to look at the diff as a whole first to get your bearings for the final result.


packages/meditrak-server/src/apiV2/CRUDHandler/helpers.js
```diff
@@ -3,8 +3,8 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-// if the endpoint is /surveys/5a5d1c66ae07fb3fb025c3a3/answers, the resource is 'surveys'
-export const extractResourceFromEndpoint = endpoint => endpoint.split('/')[1];
+// if the endpoint is surveys/5a5d1c66ae07fb3fb025c3a3/answers, the resource is 'surveys'
+export const extractResourceFromEndpoint = endpoint => endpoint.split('/')[0];
```
edmofro, line 7: After this change, `req.endpoint` changes from e.g. `/surveys` to `surveys`, which was an unintentional byproduct of using `.split` in the middleware, but one I quite like.

packages/meditrak-server/src/app/constructApiV2.js
```diff
@@ -0,0 +1,320 @@
+/**
```
edmofro: If you locally diff this file against the deleted `addRoutesToApp`, you'll notice very simple changes between the two. I was hoping github would detect it as a move and edit rather than a delete and create.

packages/meditrak-server/src/apiV2/CRUDHandler/CRUDHandler.js
```diff
@@ -6,18 +6,18 @@
...
 import { resourceToRecordType } from '../../utilities';
-import { extractResourceFromEndpoint, extractChildResourceFromEndpoint } from './helpers';
+import { extractResourceFromPath, extractChildResourceFromPath } from './helpers';
 
 export class CRUDHandler extends RouteHandler {
   constructor(req, res) {
     super(req, res);
     const { recordId, parentRecordId } = this.params;
-    this.parentResource = parentRecordId ? extractResourceFromEndpoint(this.endpoint) : undefined;
+    this.parentResource = parentRecordId ? extractResourceFromPath(this.path) : undefined;
```
edmofro, line 15: Previously we were using a middleware to get the "endpoint" out of the path, essentially just stripping the `/v2` off the front of e.g. `/v2/surveys`. Now that we're using a router under `/v2`, express does this for us, so we can just use `req.path`

packages/meditrak-server/src/apiV2/index.js
```diff
@@ -3,7 +3,18 @@
  * Copyright (c) 2017 Beyond Essential Systems Pty Ltd
  */
 
-import { catchAsyncErrors } from './middleware';
+import express from 'express';
```
edmofro, line 6: If you locally diff this file against the deleted addRoutesToApp, you'll notice very simple changes between the two. I was hoping github would detect it as a move and edit rather than a delete and create.

packages/meditrak-server/src/apiV2/middleware/catchAsyncErrors.js
```diff
@@ -0,0 +1,17 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+/**
+ * All routes will be wrapped with an error catcher that simply passes the error to the next()
+ * function, causing error handling middleware to be fired. Otherwise, async errors will be
+ * swallowed.
+ */
+export const catchAsyncErrors = routeHandler => (res, req, next) => {
```
kael89, line 11: I see this is attached to all handlers, it would be nice to somehow generalise its usage... (No specific suggestion at this stage)

kael89 (approved):
Looks great! Awesome work making this change. Obviously I didn't review this line by line, but it looks correct overall 🎉

A couple of additional suggestions:
* We can rename the `routes` folder in tests to `apiV2`
* (some point in the future) I think it'd be nice for the api folder structure to match the endpoints exactly (eg moving files like `GETClinics.js` under a `clinics` folder, even if such folders only export one handler)

packages/meditrak-server/src/apiV2/middleware/catchAsyncErrors.js
edmofro: Good point, but will keep that out of scope for now 🙏

edmofro:
The directory cleanup suggestions are great! I'll make them in a later PR to avoid merging those downstream. Thanks for the review :-)

edmofro:
@kael89 to avoid getting stuck on this task for too much longer, I've just made the `tests/routes` -> `tests/apiV2` change, and left adding folder wrappers around all of those endpoints as that would require more thinking (e.g. some might be grouped, as in the current case of `optionSets`)

That commit is [here](https://github.com/beyondessential/tupaia/pull/2964/commits/6d841dd646d4ff8f233145440f1d05daebf38847) and on another PR, because there were changes in the meantime that made it easier to apply downstream

Merging this into the base branch :-)

PR #2939, opened 2021-08-02
NZ-804: Convert email on timeout handler to generic middleware

edmofro:
### Issue #:
NZ-804
Builds towards NZ-803


packages/admin-panel/src/importExport/ImportModal.js
```diff
@@ -161,8 +157,11 @@ export const ImportModalComponent = React.memo(
               errorMessage={fileErrorMessage}
               isLoading={status === STATUS.LOADING}
             >
-              {status === STATUS.SUCCESS ? (
-                <p>{successMessage}</p>
+              {status === STATUS.TIMEOUT ? (
+                <p>
+                  Import is taking a while, and will continue in the background. You will be emailed
+                  when the import process completes.
```
edmofro, line 163: Decided the message being displayed here was more of a front end concern, the back end should really just return the status (makes generic handling slightly easier on the back end too)

packages/meditrak-server/src/apiV2/importSurveyResponses/constructEmailFromResponse.js
```diff
@@ -0,0 +1,36 @@
+/**
```
edmofro, line 1: This is just the leftover from the deleted file `setupEmailResponseTimeout.js` that didn't go into the new generic middleware handler. Could be diff'd against that locally if you want.

packages/meditrak-server/src/apiV2/importSurveyResponses/importSurveyResponses.js
```diff
@@ -29,15 +29,13 @@ import {
...
 import { SurveyResponseUpdatePersistor } from './SurveyResponseUpdatePersistor';
-import { setupEmailResponseTimeout } from './setupEmailResponseTimeout';
 import { getFailureMessage } from './getFailureMessage';
 
 /**
  * Creates or updates survey responses by importing the new answers from an Excel file, and either
  * updating or creating each answer as appropriate
  */
 export async function importSurveyResponses(req, res) {
-  setupEmailResponseTimeout(req, res); // if the import takes too long, the results will be emailed
```
edmofro, line 40: Now attached via middleware (more expected within express)

packages/meditrak-server/src/apiV2/middleware/emailAfterTimeout.js
```diff
@@ -0,0 +1,52 @@
+/**
```
edmofro, line 1: Lifted from `setupEmailResponseTimeout.js`, with tweaks to make it more generic

packages/meditrak-server/src/apiV2/middleware/emailAfterTimeout.js
```diff
@@ -0,0 +1,53 @@
...
+  // override the respond function so that when the endpoint handler finishes (or throws an error),
+  // the response is sent via email
+  res.overrideRespond = responseBody => {
+    const { subject, message } = constructEmailFromResponse(responseBody);
+    sendResponseAsEmail(user, subject, message);
+  };
+};
+
+// if the import takes too long, the results will be emailed
+export const emailAfterTimeout = constructEmailFromResponse => (req, res, next) => {
```
rohan-bes, line 41: Is there a reason that we are passing in the `constructEmailFromResponse` function as an argument here? Rather than just importing that function into this file and calling it directly?  ie.
```suggestion
export const emailAfterTimeout = (req, res, next) => {
```
I suppose it's because there may be other routes that want to use a different function for constructing the email? I mainly ask cos I got a bit confused trying to follow the code flow in this PR as I didn't realise that `constructEmailFromResponse` was a parameter instead of just the function itself.

Totally cool if we want it as a parameter if we anticipate different routes having different email building logic. I feel like at that point we'd want the original `constructEmailFromResponse` function to be given a more specific name, eg. `constructImportEmailFromResponse`, but keep this parameter the same?

Not really making any suggestion for change here, just flagging that I did get a little lost for what it's worth 👍

rohan-bes (approved):
Looks really good! Nice and clean!

One comment about a bit of confusion I had following the code, but it's nothing major 👍

packages/meditrak-server/src/apiV2/middleware/emailAfterTimeout.js
edmofro: Yes that's exactly right! In a coming change this will be applied across both import and export routes, which each have their own requirements for interpreting the response that would normally be sent to the front end, and converting it into an email. Good suggestion for naming the original function something more specific - I've done that for the export one but will change the import one to do the same in the PR I'm working on atm. Will leave it named as is here to avoid the upstream merge chain :-)

I've tried to make this series of PRs quite granular, but sometimes that makes it harder for the reviewer to see the bigger picture. 

Interested in your perspective there.... 

It's working well on my end, as it encourages me to focus things into discrete tasks (and Linear's subtasks being super lightweight help with that too), which I don't traditionally do. That's pretty helpful because I spend so much of my day distracted by whatever meetings or other things are going on that having a contained thing to come back to feels like I can be more productive in the non-distracted hours :-)

PR #2942, opened 2021-08-02
NZ-793 kobo sync service

EMcQ-BES:
### Issue #:
Resolves NZ-793 NZ-796 NZ-795 NZ-800

### Changes:

- Add sync with kobo to `meditrak-server`
- Slightly adjust kobo translation to work better for creating survey responses
- Add manual kobo sync endpoint
- Add answer translations to the translation layer


packages/data-broker/src/services/kobo/KoBoTranslator.js
```diff
@@ -12,10 +12,13 @@ export class KoBoTranslator {
 
   async fetchEntityInfoFromKoBoAnswer(koboEntityCode) {
     const entity = await this.models.entity.findOne({ code: `LA_sch_${koboEntityCode}` });
+    if (!entity) {
+      return { orgUnit: koboEntityCode, orgUnitName: koboEntityCode };
+    }
```
EMcQ-BES, line 17: If the translation layer can't find an entity, it passes through the raw value from kobo. We could possibly error or skip responses at this point, but I felt like that decision isn't really the duty of the translation layer.

packages/database/src/migrations/migrationData/20210728020128-AddFQSSurveySyncGroups.js
```diff
@@ -1,67 +1,283 @@
...
-  lesmis_fqs1_latrine: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q5',
-  lesmis_fqs1_board: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q6',
-  lesmis_fqs1_map: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q7',
-  lesmis_fqs1_table: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q8',
-  lesmis_fqs1_chair: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q9',
-  lesmis_fqs1_t_tablechair: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q10',
+  lesmis_fqs1_stdbuilding: {
+    koboQuestionCode: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q1',
+    answerMap: FQS1_ANSWER_MAP,
+  },
```
EMcQ-BES, line 19: The answer maps are against individual questions so we can adapt this if we decide to translate FQS2 questions individually

packages/database/src/migrations/migrationData/20210728020128-AddFQSSurveySyncGroups.js
```diff
@@ -1,67 +1,283 @@
...
+  lesmis_fqs1_stdbuilding_c: { koboQuestionCode: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q1Comment' },
+  lesmis_fqs1_office: {
+    koboQuestionCode: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q2',
+    answerMap: FQS1_ANSWER_MAP,
+  },
+  lesmis_fqs1_office_c: { koboQuestionCode: 'FQS1PrimaryPart1/FQS1PrimaryPart1Q2Comment' },
```
EMcQ-BES, line 25: These comment questions were missing from the earlier version of the tupaia survey I was looking at, so I've also added them in now

packages/meditrak-server/src/kobo/manualKoBoSync.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import { DataBroker } from '@tupaia/data-broker';
+import { respond } from '@tupaia/utils';
+import { syncWithKoBo } from './startSyncWithKoBo';
+
+export async function manualKoBoSync(req, res, next) {
+  const { models } = req;
+
+  const dataBroker = new DataBroker();
```
EMcQ-BES, line 13: Could probably have these two processes share a data broker, but with the model registry being a singleton there's not a lot of overhead to it 🤔

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+    for (const [surveyCode, responses] of Object.entries(koboData)) {
+      const survey = await transactingModels.survey.findOne({ code: surveyCode });
+
+      for (const responseData of responses) {
+        if (responseData.eventDate > newSyncTime) {
+          newSyncTime = responseData.eventDate;
+        }
+        const entity = await transactingModels.entity.findOne({ code: responseData.orgUnit });
+        if (!entity) {
+          winston.warn(`Skipping KoBo sync for unknown entity ${responseData.orgUnit}`);
```
EMcQ-BES, line 60: Currently just skipping responses if we can't map them to an entity. I'm not sure if this is the behaviour we want, open to alternatives

packages/data-broker/src/services/kobo/KoBoTranslator.js
edmofro: Yeah I think that's fair.

packages/meditrak-server/src/app/addRoutesToApp.js
```diff
@@ -242,6 +243,7 @@ export function addRoutesToApp(app) {
   app.post('(/v[0-9]+)/indicators', createIndicators);
   app.post('(/v[0-9]+)/permissionGroups', createPermissionGroups);
   app.post('(/v[0-9]+)?/dashboardRelations', createDashboardRelations);
+  app.post('(/v[0-9]+)?/koboSync', requestManualKoBoSync);
```
edmofro, line 246:
```suggestion
  app.post('(/v[0-9]+)?/syncFromService', requestManualKoBoSync);
```
And then have the query parameters take the `sync_service.code`

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+    const dataBroker = new DataBroker();
+    setInterval(() => syncWithKoBo(models, dataBroker), PERIOD_BETWEEN_SYNCS);
+  }
+}
+
+export async function syncWithKoBo(models, dataBroker) {
+  await pullLatest(models, dataBroker);
+}
+
+async function pullLatest(models, dataBroker) {
```
edmofro, line 29:
```suggestion
```

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+  if (process.env.KOBO_SYNC_DISABLE) {
+    // eslint-disable-next-line no-console
+    console.log('KoBo sync is disabled');
+  } else {
+    const dataBroker = new DataBroker();
+    setInterval(() => syncWithKoBo(models, dataBroker), PERIOD_BETWEEN_SYNCS);
+  }
+}
+
+export async function syncWithKoBo(models, dataBroker) {
```
edmofro, line 25:
```suggestion
export async function syncWithKoBo(models, dataBroker, serviceCode) {
```

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+import keyBy from 'lodash.keyby';
+import winston from 'winston';
+import { DataBroker } from '@tupaia/data-broker';
+import { generateId } from '@tupaia/database';
+import { reformatDateStringWithoutTz } from '@tupaia/utils';
+
+const PERIOD_BETWEEN_SYNCS = 10 * 60 * 1000; // 10 minutes between syncs
+const SERVICE_NAME = 'KoBo';
```
edmofro, line 13:
```suggestion
const SERVICE_TYPE = 'kobo';
```

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+
+export async function startSyncWithKoBo(models) {
+  if (process.env.KOBO_SYNC_DISABLE) {
+    // eslint-disable-next-line no-console
+    console.log('KoBo sync is disabled');
+  } else {
+    const dataBroker = new DataBroker();
+    setInterval(() => syncWithKoBo(models, dataBroker), PERIOD_BETWEEN_SYNCS);
```
edmofro, line 21:
```suggestion
    const koboSyncServices = await models.syncService.find({ type: SERVICE_TYPE });
    koboSyncServices.forEach(service => setInterval(() => syncWithKoBo(models, dataBroker, service.code), PERIOD_BETWEEN_SYNCS));
```
This also requires something I didn't think of in the other PR - a `sync_service.type` column

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+}
+
+export async function syncWithKoBo(models, dataBroker) {
+  await pullLatest(models, dataBroker);
+}
+
+async function pullLatest(models, dataBroker) {
+  const syncCursor = await models.syncCursor.findOne({ service_name: SERVICE_NAME });
```
edmofro, line 30:
```suggestion
  const syncService = await models.syncService.findOne({ type: SERVICE_TYPE, service_code: serviceCode });
  
  if (!syncCursor) {
    throw new Error(`No KoBo sync service with the code ${serviceCode} exists`);
  }
```

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+
+  // Pull data from KoBo
+  const koboData = await dataBroker.pull(
+    {
+      code: syncCursor.config.koboSurveys,
+      type: dataBroker.getDataSourceTypes().SYNC_GROUP,
+    },
+    {
+      startSubmissionTime: reformatDateStringWithoutTz(syncCursor.sync_time),
```
edmofro, line 39:
```suggestion
      startSubmissionTime: syncService.sync_cursor,
```
Strangely I think using a worse column type (`TEXT`) for the sync cursor will make our jobs easier here - we just save what comes back, and know it won't have been [mis]interpreted and changed at any steps along the way

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
edmofro: I guess the other option is to block sync until it gets deleted or the entity gets mapped correctly. I think this is probably better :+1:

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+    },
+  );
+
+  const apiUser = await models.apiClient.findOne({ username: process.env.KOBO_API_USERNAME });
+  if (!apiUser) {
+    throw new Error('Cannot find API client user for KoBo');
+  }
```
edmofro, line 46: This is probably the best solution to the problem, but I might mention this behaviour to Erin because she might not be expecting it.

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -0,0 +1,95 @@
...
+        if (!entity) {
+          winston.warn(`Skipping KoBo sync for unknown entity ${responseData.orgUnit}`);
+          continue;
+        }
+
+        const surveyResponse = await transactingModels.surveyResponse.create({
+          id: generateId(),
+          survey_id: survey.id,
+          user_id: apiUser.user_account_id,
+          assessor_name: 'KoBo Integration',
```
edmofro, line 68: Although we are pretty limited to using a generic user id, we can probably do a better job here - the KoBo data payload includes "username", which is not found in the ones I've done on the test account but might end up defined. Can we use that instead? I realise it means breaking the strict event format, but that's ok with me because it's in an additive way so future consumers of the data as an actual event would still get everything they needed (and more)

edmofro (changes requested):
After suggesting that we could just use the `setting` table, I've gone all in on over engineering the sync service pattern! Let me know what you think, happy to discuss in person.

Also congrats on pulling this together, the flow is all super nice and I think you've put things in all the right places 🎉

packages/data-broker/src/services/kobo/KoBoTranslator.js
```diff
@@ -26,9 +29,13 @@ export class KoBoTranslator {
...
-    for (const [tupaia, kobo] of Object.entries(questionCodeMapping)) {
-      if (restOfFields[kobo] !== undefined) {
-        dataValues[tupaia] = restOfFields[kobo];
+    for (const [tupaia, { koboQuestionCode, answerMap }] of Object.entries(questionMapping)) {
+      if (restOfFields[koboQuestionCode] !== undefined) {
+        let koboValue = restOfFields[koboQuestionCode];
+        if (answerMap) {
+          koboValue = answerMap[koboValue] || koboValue;
+        }
+        dataValues[tupaia] = koboValue;
```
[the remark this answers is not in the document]
EMcQ-BES: I didn't realise you could combine dot and bracket notation like that... Neat!

Looking back at this, one possible situation we're not supporting is explicitly mapping an answer to `undefined`. That's probably an incredibly minor edge case but something I thought I'd note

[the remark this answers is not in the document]
edmofro: True, I'm happy to leave that unsupported :-)

packages/data-broker/src/services/kobo/KoBoTranslator.js
```diff
@@ -31,17 +38,16 @@ export class KoBoTranslator {
...
+        const koboValue = restOfFields[koboQuestionCode];
+        dataValues[tupaia] =
+          answerMap?.[koboValue] !== undefined ? answerMap[koboValue] : koboValue;
       }
     }
 
     return {
       event,
       eventDate,
+      koboUser,
```
edmofro, line 50:
```suggestion
      assessor: koboUser,
```
Might as well use a generic term that we might reuse across sync services

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
edmofro: I don't think this needs mentioning to Erin now that we've got the assessor's name coming through 🎉

packages/database/src/migrations/20210804234015-AddKoBoDataSourceEntities-modifies-data.js
```diff
@@ -0,0 +1,48 @@
...
+    `)
+  ).rows;
+
+  return schools.map(row =>
+    insertObject(db, 'data_service_entity', {
+      id: generateId(),
+      entity_code: row.code,
+      config: {
+        kobo_id: row.kobo_id,
+      },
```
edmofro, line 34: Wow fully expected `data_service_entity` to specify which service the mapping was for, but I guess with different config entries we could have one row that mapped multiple services. Might need to use more specific service codes than `dhis_id` and `kobo_id` in future, but we can leave that for another day

edmofro (approved):
LGTM! One final suggestion for a rename, but pre-approving :-)

edmofro (approved):

PR #2943, opened 2021-08-02
NZ-813: Import and export routes sit within their own routers

edmofro:
### Issue #:
NZ-813
Towards NZ-803

Purely moving import and export related files under their own routers, should be no logic changes


EMcQ-BES (approved):
What an enormous change set for very little actual change.

Looks good, will admit I started to zoned out on all the file moves by the end, but it looks like the unit tests are still passing so I'm happy.

Codeship failing on branch name too long. Not a big deal. Tends to happen when using the linear generated branch name

edmofro:
Replaced with shorter branch name at https://github.com/beyondessential/tupaia/pull/2964

PR #2944, opened 2021-08-02
MEL-283: Admin panel map overlay groups and map overlay group relations

chris-bes:
### Changes:

- Admin panel: map overlay groups and map overlay group relations

### DIFF
--- a/packages/meditrak-server/src/routes/mapOverlayGroupRelations/GETMapOverlayGroupRelations.js
+++ b/packages/meditrak-server/src/routes/mapOverlayGroupRelations/GETMapOverlayGroupRelations.js
@@ -1,0 +1,65 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { GETHandler } from '../GETHandler';
+import { assertAnyPermissions, assertBESAdminAccess } from '../../permissions';
+import {
+  assertMapOverlayGroupRelationsGetPermissions,
+  createMapOverlayGroupRelationDBFilter,
+  createRelationsViaParentOverlayGroupDBFilter,
+} from './assertMapOverlayGroupRelationsPermissions';
+import { assertMapOverlayGroupsGetPermissions } from '../mapOverlayGroups';
+
+/**
+ * Handles endpoints:
+ * - /mapOverlayGroupRelations
+ * - /mapOverlayGroupRelations/:mapOverlayGroupRelationId
+ */
+export class GETMapOverlayGroupRelations extends GETHandler {
+  permissionsFilteredInternally = true;
+
+  customJoinConditions = {
+    map_overlay_group: ['map_overlay_group.id', 'map_overlay_group_relation.map_overlay_group_id'],
+  };
+
+  async findSingleRecord(mapOverlayId, options) {
+    const mapOverlay = await super.findSingleRecord(mapOverlayId, options);
+
+    const mapOverlayChecker = accessPolicy =>
+      assertMapOverlayGroupRelationsGetPermissions(accessPolicy, this.models, mapOverlayId);
+
+    await this.assertPermissions(assertAnyPermissions([assertBESAdminAccess, mapOverlayChecker]));
+
+    return mapOverlay;
+  }
+
+  async getPermissionsFilter(criteria, options) {
+    const dbConditions = await createMapOverlayGroupRelationDBFilter(
+      this.accessPolicy,
+      this.models,
+      criteria,
+    );
+    return { dbConditions, dbOptions: options };
+  }
+
+  async getPermissionsViaParentFilter(criteria, options) {
+    // Check parent permissions
+    const mapOverlayGroupPermissionChecker = accessPolicy =>
+      assertMapOverlayGroupsGetPermissions(accessPolicy, this.models, this.parentRecordId);
+
+    await this.assertPermissions(
+      assertAnyPermissions([assertBESAdminAccess, mapOverlayGroupPermissionChecker]),
+    );
+
+    const dbConditions = await createRelationsViaParentOverlayGroupDBFilter(
+      this.accessPolicy,
+      this.models,
+      criteria,
+      this.parentRecordId,
+    );
+
+    return { dbConditions, dbOptions: options };
+  }
+}
--- a/packages/meditrak-server/src/routes/mapOverlayGroups/GETMapOverlayGroups.js
+++ b/packages/meditrak-server/src/routes/mapOverlayGroups/GETMapOverlayGroups.js
@@ -1,0 +1,41 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { GETHandler } from '../GETHandler';
+import { assertAnyPermissions, assertBESAdminAccess } from '../../permissions';
+import {
+  assertMapOverlayGroupsGetPermissions,
+  createMapOverlayGroupDBFilter,
+} from './assertMapOverlayGroupsPermissions';
+/**
+ * Handles endpoints:
+ * - /mapOverlayGroups
+ * - /mapOverlayGroups/:mapOverlayGroupId
+ */
+export class GETMapOverlayGroups extends GETHandler {
+  permissionsFilteredInternally = true;
+
+  async findSingleRecord(mapOverlayId, options) {
+    const mapOverlay = await super.findSingleRecord(mapOverlayId, options);
+
+    const mapOverlayGroupChecker = accessPolicy =>
+      assertMapOverlayGroupsGetPermissions(accessPolicy, this.models, mapOverlayId);
+
+    await this.assertPermissions(
+      assertAnyPermissions([assertBESAdminAccess, mapOverlayGroupChecker]),
+    );
+
+    return mapOverlay;
+  }
+
+  async getPermissionsFilter(criteria, options) {
+    const dbConditions = await createMapOverlayGroupDBFilter(
+      this.accessPolicy,
+      this.models,
+      criteria,
+    );
+    return { dbConditions, dbOptions: options };
+  }
+}
--- a/packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
+++ b/packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
@@ -1,0 +1,213 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import { keyBy } from 'lodash';
+
+import { hasBESAdminAccess } from '../../permissions';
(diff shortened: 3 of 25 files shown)


packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
```diff
@@ -0,0 +1,213 @@
...
+  const permittedMapOverlayGroupRelations = await models.mapOverlayGroupRelation.find({
+    child_id: permittedMapOverlays.map(mo => mo.id),
+    child_type: 'mapOverlay',
+  });
+  const rootMapOverlayGroup = await models.mapOverlayGroup.findRootMapOverlayGroup();
+  const permittedMapOverlayGroupIds = findPermittedMapOverlayGroupIds(
+    relationByChildId,
+    permittedMapOverlayGroupRelations,
+    rootMapOverlayGroup.id,
+  );
```
kael89, line 205: I believe we don't need the `findPermittedMapOverlayGroupIds` method:

```suggestion
  const permittedMapOverlayGroupIds = getUniqueEntries(
    permittedMapOverlayGroupRelations.map(r => r.map_overlay_group_id),
  );
```

packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
```diff
@@ -0,0 +1,213 @@
...
+  const result = await hasMapOverlayGroupEditPermissions(accessPolicy, models, mapOverlayId);
+
+  if (result.result) {
+    return true;
+  }
+
+  throw new Error(result.errorMessage);
+};
+
+const findPermittedMapOverlayGroupIds = (
```
kael89, line 151: Nice recursive function!! 👏

Recursion is notoriously complicated, so a comment that explains the general recursion logic would be beneficial. What would be especially helpful is describing how we traverse the relation tree (if I got this right, bottom-up until we reach the root nodes)

packages/meditrak-server/src/routes/mapOverlayGroups/GETMapOverlayGroups.js
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Handles endpoints:
+ * - /mapOverlayGroups
+ * - /mapOverlayGroups/:mapOverlayGroupId
+ */
+export class GETMapOverlayGroups extends GETHandler {
+  permissionsFilteredInternally = true;
+
+  async findSingleRecord(mapOverlayId, options) {
+    const mapOverlay = await super.findSingleRecord(mapOverlayId, options);
```
kael89, line 21:
```suggestion
  async findSingleRecord(mapOverlayGroupId, options) {
    const mapOverlayGroup = await super.findSingleRecord(mapOverlayGroupId, options);
```

packages/meditrak-server/src/routes/mapOverlayGroupRelations/GETMapOverlayGroupRelations.js
```diff
@@ -0,0 +1,65 @@
...
+  assertMapOverlayGroupRelationsGetPermissions,
+  createMapOverlayGroupRelationDBFilter,
+  createRelationsViaParentOverlayGroupDBFilter,
+} from './assertMapOverlayGroupRelationsPermissions';
+import { assertMapOverlayGroupsGetPermissions } from '../mapOverlayGroups';
+
+/**
+ * Handles endpoints:
+ * - /mapOverlayGroupRelations
+ * - /mapOverlayGroupRelations/:mapOverlayGroupRelationId
```
kael89, line 18: Should we also add `/mapOverlayGroups/:parentRecordId/mapOverlayGroupRelations`?

packages/meditrak-server/src/routes/mapOverlayGroupRelations/GETMapOverlayGroupRelations.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+export class GETMapOverlayGroupRelations extends GETHandler {
+  permissionsFilteredInternally = true;
+
+  customJoinConditions = {
+    map_overlay_group: ['map_overlay_group.id', 'map_overlay_group_relation.map_overlay_group_id'],
+  };
+
+  async findSingleRecord(mapOverlayId, options) {
+    const mapOverlay = await super.findSingleRecord(mapOverlayId, options);
```
kael89, line 28:
```suggestion
  async findSingleRecord(mapOverlayGroupRelationId, options) {
    const mapOverlayGroupRelation = await super.findSingleRecord(mapOverlayGroupRelationId, options);
```

packages/meditrak-server/src/routes/mapOverlayGroupRelations/GETMapOverlayGroupRelations.js
```diff
@@ -0,0 +1,65 @@
...
+
+    const mapOverlayChecker = accessPolicy =>
+      assertMapOverlayGroupRelationsGetPermissions(accessPolicy, this.models, mapOverlayId);
```
kael89, line 31: I'm not sure if this will work: `assertMapOverlayGroupRelationsGetPermissions` expects a map overlay group id, but if I'm not mistaken here we have access to a `mapOverlayGroupRelationId` (named `mapOverlayId` in the code above)

kael89 (changes requested):
Awesome work @chris-bes ! Looks great overall - I left just a few minor comments/questions 🚀 

It would be also great to add some integration tests for this functionality - we could prioritise the non-trivial cases. Permission testing coverage seems pretty high so it would be good to keep it up.

(Not specifically related to your PR) Seeing that a lot of code is written to allow permission checking, I was wondering whether we could define permissions in a more declarative way. The CRUD operation abstraction looks great, perhaps we could also abstract away some trivial permission behaviour and relationships. One example: delete permissions seems to always be using the edit permissions, so this could be the default behaviour. Food for thought 😃

packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
chris-bes: Ah I had to do it this way because we can have multi map overlay groups, just getting the group ids from `permittedMapOverlayGroupRelations` is not enough (`permittedMapOverlayGroupRelations` will only contain the relations that have map overlay children). So I had to traverse up the list until it hits the root map overlay group

packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
chris-bes: I should have done that 😿

packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
kael89: You are right - I realised that later, after reading `findPermittedMapOverlayGroupIds`... Perhaps a one-line comment would help clarify

packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
```diff
@@ -0,0 +1,213 @@
...
+    throw new Error(`No map overlay group exists with id "${mapOverlayGroupId}"`);
+  }
+
+  const childMapOverlayRelations = await models.mapOverlayGroupRelation.find({
+    map_overlay_group_id: mapOverlayGroupId,
+  });
+
+  if (!childMapOverlayRelations.length) {
+    return true;
+  }
```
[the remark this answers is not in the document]
chris-bes: That's a good point! My thought is, if it is an empty map overlay group, it doesn't really affect anything (won't be shown in tupaia.org), so you should be freely do whatever you want. That way it's also flexible when you're building a new map overlay group and adding map overlays into it later on.

I guess in this case I would prefer to follow whatever is more useful in terms of business model. What do you reckon?

packages/meditrak-server/src/tests/routes/GETMapOverlayGroupRelations.test.js
```diff
@@ -0,0 +1,151 @@
+/**
```
chris-bes, line 1: I'm aware that `GETMapOverlay.test.js`, `GETMapOverlayGroup.test.js` and `GETMapOverlayGroupRelation.test.js` have a lot of duplicated code. But I'm thinking that this is because they just really have similar permission model. Would totally understand if you think it needs to be made less DRY and happy to do some more work to refactor them if that's the case

packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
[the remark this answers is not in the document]
kael89: I agree! Just wanted to point out the fact, to make sure that it's a conscious decision 🙂 Let's proceed as is

packages/meditrak-server/src/tests/routes/GETMapOverlayGroupRelations.test.js
kael89: No need to do so in this PR - we could decide to do this on a larger scale at some point, if it makes sense

packages/meditrak-server/src/routes/mapOverlayGroups/assertMapOverlayGroupsPermissions.js
```diff
@@ -148,6 +142,11 @@ export const assertMapOverlayGroupsEditPermissions = async (accessPolicy, models
   throw new Error(result.errorMessage);
 };
 
+/**
+ * We can have multi level map overlay groups. So to find all the permitted map overlay groups,
+ * we have to traverse from the bottom map overlay group relations (that have map overlays as childrens) until
+ * we hit the root map overlay group (top) to also include any nested map overlay groups.
+ */
```
kael89, line 149: Excellent!!

kael89 (approved):
Great work @chris-bes ! Thank you for adding those tests too 🎉

kael89:
@chris-bes by the way the tests failed in CI/CD...

chris-bes:
My apology for the delay, I've been quite unorganized the last few days. The tests passed locally but didn't in CI/CD, turns out it was because of database inconsistency...

packages/database/src/modelClasses/MapOverlayGroup.js
```diff
@@ -21,8 +21,12 @@ export class MapOverlayGroupModel extends DatabaseModel {
     return MapOverlayGroupType;
   }
 
+  get RootMapOverlayGroupCode() {
+    return ROOT_MAP_OVERLAY_GROUP_CODE;
+  }
+
```
kael89, line 27: I'd use UPPER_SNAKE_CASE for this, since it's a class constant, but we are highly inconsistent on how we expose those so not a big deal

PR #2947, opened 2021-08-03
sol-277 COVID-19 Samoa: Adding sub_district level between district and village level

nick-beyondessential:
### Changes:
- import sub_divisions into Samoa for Covid-Samoa project (Only exist in this project) This was done via migration rather than admin panel because:
  a. One of the districts (Unknown District has no geolocation, which hides from map but breaks admin import)
  b. avoid timing hassles with mixed import/migration of entity_relations
- migrate Samoa villages in Covid-Samoa to District(sub-district) parents.
- Remove facilities from Covid-Samoa
- Add household entity type
- import household entities and add hierarchy relations
- remove facility overlays


packages/database/src/migrations/20210803035205-CovidSamoaAddVillageEntityRelations-modifies-data.js
```diff
@@ -0,0 +1,62 @@
...
+  // console.log('villageRelations: ', villageRelations);
+  const villageCodes = Object.keys(villageRelations);
+  const subDistrictCodes = Object.values(villageRelations);
+
+  const samoaVillageRows = (
+    await db.runSql(`
+      select code, id from entity where "code" in (${arrayToDbString(villageCodes)});
+    `)
+  ).rows;
+  const samoaVillageIds = Object.assign({}, ...samoaVillageRows.map(x => ({ [x.code]: x.id })));
```
rohan-bes, line 30: Oh this is a nifty trick!!! I usually use a reduce in this scenario but yours seems quicker!
Gotta be careful tho with spread operator in function arguments, can lead to maximum call stack limit exceeded when dealing with large arrays (10K +)

rohan-bes (approved):
Nice work, LGTM

rohan-bes:
@nick-beyondessential have we tested these migrations on a feature instance to see if there's any issues? (it seems safe tho)

packages/database/src/migrations/20210803035205-CovidSamoaAddVillageEntityRelations-modifies-data.js
nick-beyondessential: Ah good to know!

nick-beyondessential:
Will push through to testing. Which I assume when passed we'll need a new hotfix PR

nick-beyondessential:
@rohan-bes added household entity type.

nick-beyondessential:
Closed in favour of https://github.com/beyondessential/tupaia/pull/2957

PR #2949, opened 2021-08-03
2729: Disallow period granularity for dhis surveys (MEL-92)

kael89:
### Issue #:
Addresses https://linear.app/tupaia/issue/MEL-92#comment-26ffa86f

image

### DIFF
--- a/packages/meditrak-server/src/dataAccessors/editSurvey.js
+++ b/packages/meditrak-server/src/dataAccessors/editSurvey.js
@@ -3,6 +3,7 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 import { assertCanAddDataElementInGroup } from '../database';
+import { validateSurveyFields } from './validateSurveyFields';
 
 const RESOURCE_TYPES = {
   DATA_GROUP: 'dataGroup',
@@ -98,6 +99,14 @@ class SurveyEditor {
       // eslint-disable-next-line no-param-reassign
       model[fieldName] = fieldValue;
     });
+
+    if (model.databaseType === 'survey') {
+      await validateSurveyFields(this.models, {
+        code: model.code,
+        periodGranularity: model.period_granularity,
+      });
+    }
+
     if (isDataSource) {
       this.updateDataSourceConfig(model, updatedFields);
     }
--- a/packages/meditrak-server/src/dataAccessors/editSurvey.js
+++ b/packages/meditrak-server/src/dataAccessors/editSurvey.js
@@ -89,21 +89,22 @@ class SurveyEditor {
     const updatedFields = this.updatedFieldsByResource[resourceType];
     const isDataSource = model.databaseType === this.models.dataSource.databaseType;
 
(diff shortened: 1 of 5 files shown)


edmofro (commented):
Nice validation, but does this mean we can no longer edit the data service of a survey directly? I know we validate against that on import because people could do it accidentally, but we left the route of the direct edit if they want to intentionally change that right?

kael89:
> Nice validation, but does this mean we can no longer edit the data service of a survey directly? I know we validate against that on import because people could do it accidentally, but we left the route of the direct edit if they want to intentionally change that right?

I didn't know that, I discussed with @IgorNadj and we assumed that we didn't want people to be able to edit the data service either (and leave this be handled by migrations)...

Probably I conscious decision was taken then following the above line of thought, but I didn't know that. Happy to change this so that service type can still be changed during editing!

edmofro:
Or perhaps it wasn't a conscious decision at all! I think that's a nice outcome though, that people have the ability to change the data service type through the admin panel if they know what they're doing (presumably they do if they're doing that through edit - it's a very active step to take to go in there and change it)

kael89:
> Or perhaps it wasn't a conscious decision at all! I think that's a nice outcome though, that people have the ability to change the data service type through the admin panel if they know what they're doing (presumably they do if they're doing that through edit - it's a very active step to take to go in there and change it)

Good point. Personally, I'm fine either way: being able to edit it is the best UX, not being able to may simplify some things in code. I'd go for the former if it is not a _rare_ case, and with the latter if it is a rare case indeed. That, keeping in mind that we are still open to migrations: if this is not the case in the future, then the former would be a must-have.

Happy to re-instate this functionality 🙂

edmofro:
It probably is a rare case at the moment, and even in future, but maintaining the ability for things to be done through the admin panel and not migrations aligns with the long term strategy of needing less dev intervention for configuration and maintenance of Tupaia projects. So although it means the code is a bit more cluttered, I'm going to push for that option please 🙏

kael89:
@edmofro ready for re-review

packages/meditrak-server/src/dataAccessors/editSurvey.js
```diff
@@ -88,16 +89,25 @@ class SurveyEditor {
...
-        // Handle config last since its value depends on other fields
-        // For the same reason, do not set model fields to `undefined`
-        // so that the actual record values can be used for config calculation
-        return;
-      }
-      // eslint-disable-next-line no-param-reassign
-      model[fieldName] = fieldValue;
-    });
+    Object.entries(updatedFields)
+      .filter(([, fieldValue]) => fieldValue !== undefined)
```
kael89, line 93: In the past we'd let `undefined` fields in the model, since they won't actually be updated in the db. We'd still skip those for the data source model, because setting its fields to `undefined` caused issues with using the model state to draw conclusions (see lines 92-96 in existing code)

This causes an issue for the `survey` model too now, so I decided it's high time to just skip those fields!

edmofro (approved):
Thanks for the update - lgtm!

PR #2952, opened 2021-08-03
NZ-62 list visual UI

tcaiger:
### Changes:

- Add ListVisual to ui-components
- Add ListVisual story to storybook
- Add a mock module to storybook to handle `child_process` module in `utils/scripts.js`

---

### Screenshots:
See issue


packages/lesmis/src/api/queries/useDashboardData.js
```diff
@@ -6,8 +6,37 @@
...
+      }
+      return codes;
+    }, []);
+
+export const useDashboardData = entityCode => {
+  const query = useQuery(['dashboard', entityCode], () => get(`dashboard/${entityCode}`), {
     staleTime: 60 * 60 * 1000,
     refetchOnWindowFocus: false,
   });
+
```
tcaiger, line 27: I've added the data processing in the api data layer to keep it out of the components. The processing is:
 - parse drilldown config
 - remove un-supported visual types

packages/ui-components/src/components/ListVisual/ListVisual.js
```diff
@@ -0,0 +1,157 @@
...
+    if (item.statistic !== undefined) {
+      displayConfig = getDisplayConfig(config, item.statistic);
+    }
+
+    return { ...item, rowType, valueType: config.valueType, displayConfig, drillDownCode };
+  });
+};
+
+export const ListVisual = React.memo(
+  ({ viewContent, isLoading, dashboard, isError, error, drillDowns, DrillDownComponent }) => {
```
tcaiger, line 90: The `DrillDownComponent` is passed in as a prop so that the ListVisual can be de-coupled from the rest of the UI and moved into storybook. In future we may want to use this component in web-frontend or other apps and this is setup to handle that. It does make the component slightly more complex though...

packages/ui-components/src/components/ListVisual/ListVisual.js
```diff
@@ -0,0 +1,157 @@
...
+
+    const list = processData(config, data, drillDowns);
+
+    return (
+      <Container isLoading={isLoading}>
+        <FetchLoader isLoading={isLoading} isError={isError} error={error}>
+          {list?.map(({ rowType, valueType, label, displayConfig, drillDownCode }, index) => {
+            const ValueComponent = VALUE_TYPE_COMPONENTS[valueType];
+
+            if (rowType === 'drilldown') {
```
edmofro, line 101: Instead of making it the row type, could we make it a wrapper that can go around any row? So a row could be a `subheader`, but be wrapped in a drilldown?

That "wrapping in a drilldown" also feels more easily applied across other components, such as the segment of a pie chart etc.

packages/ui-components/src/components/ListVisual/ListVisual.js
```diff
@@ -0,0 +1,157 @@
...
+
+            return (
+              <RowComponent
+                // eslint-disable-next-line react/no-array-index-key
+                key={index}
+                label={label}
+              >
+                <ValueComponent displayConfig={displayConfig} />
+              </RowComponent>
+            );
```
edmofro, line 127: Can we please add indentation based on the levels of parent deep? The header and subheader styles are nice, but the back end allows for infinite layers of nested rows, so it'd be good to have at least something that indicated that below level three here on the front end

edmofro (changes requested):
All the code looks really nice! I just have some points about making this better able to handle cases other than the specific one we're currently dealing with.

I also have a proposition for how we might handle infinite drill downs:
Instead of creating a ReportModal for every report available, we share a single ReportModal component that shows up based on a url param. This means not needing to pass through all of the possible drill downs and have them pre-construct a ReportModal, in case they need to be displayed - instead the ReportModal just grabs the report it should show from url params. I wonder if a shared modal might also be a little lighter performance wise? It may cause some visual glitch issues if the report changed over at the wrong stage in the slide animation, but hopefully that's manageable.

Reigning myself back in though, if you didn't think deeper drilldowns was in scope, we can leave that for another day.

packages/lesmis/src/api/queries/useDashboardReportData.js
```diff
@@ -5,31 +5,69 @@
...
+    dashboards: useDashboardData({ entityCode }),
+  });
+
+  const dashboard = query.data?.dashboards?.find(dash =>
+    dash.items.find(item => item.reportCode === reportCode),
   );
+
+  const dashboardItem = dashboard?.items.find(item => item.reportCode === reportCode);
+
+  const reportCodes = getReportCodesByCode(query.data?.dashboards);
```
tcaiger, line 63: The report code is almost always the same as the dashboardItem code but my understanding is that it is not safe to assume they are the same, right?

edmofro: That's right

tcaiger:
@edmofro - Ready for another review. I took your suggestion and simplified the modals to be just one modal. The See more button is now just a router link to the modal and the drilldowns of the ListVisual are the same. This means that infinite drilling down is supported using links. At the moment it works well for the ListVisual modal where you can expand the ListVisual and then drill down to the underlying reports (you can still drilldown the un-expanded ListVisual too). I tweaked the url history so you can click back through the drilldowns to wherever you came from.

packages/lesmis/src/api/queries/useDashboardData.js
```diff
@@ -6,8 +6,37 @@
...
-  useQuery(['dashboard', entityCode], () => get(`dashboard/${entityCode}`), {
+const getDrillDownCodes = dashboardItems =>
+  dashboardItems
+    .filter(d => !!d.drillDown)
+    .reduce((codes, d) => {
+      const entry = d.drillDown?.itemCodeByEntry;
+
+      if (entry) {
+        const values = Object.values(d.drillDown.itemCodeByEntry);
+        return [...codes, ...values];
```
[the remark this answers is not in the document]
edmofro: You didn't like these rename suggestions?

packages/lesmis/src/components/DashboardReport.js
```diff
@@ -29,83 +31,66 @@ const Footer = styled(FlexEnd)`
...
-            isError={isError}
-            error={error}
-            drillDowns={drillDowns}
-            dashboard={dashboard}
-            DrillDownComponent={DashboardReportModal}
-          />
-        </Container>
-      );
-    }
+    const { reportData, dashboardItemConfig: config, reportCodes } = data;
```
edmofro, line 45: `reportCodes` feels like quite a different concern to me, and maybe should be a different query to the one that fetches the data for a specific dashboard item?

packages/ui-components/src/components/ListVisual/Rows.js
```diff
@@ -0,0 +1,115 @@
...
+`;
+
+export const DrillDownRow = ({ label, entityCode, reportCode, children }) => {
+  const { search } = useLocation();
+  return (
+    <LinkRow
+      component={RouterLink}
+      to={{
+        pathname: `/${entityCode}/dashboard`,
+        search: `${search}&reportCode=${reportCode}`,
```
edmofro, line 99: Are we guaranteed to have an existing search, so we're sure we use `&` not `?` here?

edmofro (changes requested):
Really nice work here, I'm a big fan of the new single modal and link structure, love that the back button works through drill downs, and there are some great tweaks along the way too (like collapsing the large switch between list and chart visuals into a tiny ternary)

There are a couple of suggestions from last time that I'd like your comment on, even if your opinion is they should be left for now because they're not an immediate requirement

I also reckon there's a slightly lighter data flow where we wouldn't need to pass down the report codes, but I'm not confident so might have a play with that shortly and come back to you

packages/ui-components/src/components/ListVisual/Rows.js
tcaiger: Well spotted but actually react-router is clever enough to take care of this for us

packages/ui-components/src/components/ListVisual/ListVisual.js
```diff
@@ -70,88 +68,71 @@ const processData = (config, data, drillDowns) => {
...
-                >
-                  <ValueComponent displayConfig={displayConfig} />
-                </DrillDownRow>
-              );
-            }
-
             const RowComponent = ROW_TYPE_COMPONENTS[rowType];
 
             return (
               <RowComponent
```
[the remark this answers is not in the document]
tcaiger: This feels a bit fiddly for something that we don't yet need and might never need. If it's ok, I'd prefer to look at this when it's required.

packages/ui-components/src/components/ListVisual/ListVisual.js
```diff
@@ -70,88 +68,71 @@ const processData = (config, data, drillDowns) => {
...
 
     if (config?.drillDown?.itemCodeByEntry[item.code] !== undefined) {
-      drillDownCode = config?.drillDown?.itemCodeByEntry[item.code];
+      const drillDownCode = config?.drillDown?.itemCodeByEntry[item.code];
 
       // check that the drill down config exists before trying to render a drilldown row
-      if (drillDowns.find(d => d.code === drillDownCode)) {
+      if (reportCodes[drillDownCode] !== undefined) {
         rowType = 'drilldown';
+        drillDownReportCode = reportCodes[drillDownCode];
```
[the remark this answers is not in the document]
tcaiger: Yes good idea. I will take a look at this now.

packages/lesmis/src/components/DashboardReport.js
tcaiger: I know what you mean, it seems like there should be an easier way. All I'm trying to do is to map the dashboardItemCode to the reportCode. I could add another query like useReportCodes or useReportCodeByItemCode that would make another request (probably hit the cache) to the `dashboard` end point. Is that what you mean? Other options I can think of are:
- adding react context somewhere to store the report codes or dashboard data
- iterating on the drilldown configs after they are fetched and adding the reportcodes next to the dashboardItemCodes
- adding the reportCodes to the backend config
- Changing the backend config to be report codes instead of item codes
- Adding the dashboardItemCode to the url

But none of them seem that good to me 🤷‍♂️

packages/lesmis/src/api/queries/useDashboardData.js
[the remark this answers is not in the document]
tcaiger: Ah I thought it said they were no longer relevant after the refactoring but now that I look closer they are. Yep they look good.

tcaiger:
@edmofro - I have updated the link to be a decorator on the ListVisual rows. I also added a filter on lesmis-server to remove all the legacy dashboards as they are not supported anymore. The only remaining question is on the [reportCodes data flow](https://github.com/beyondessential/tupaia/pull/2952#discussion_r686479917).

packages/ui-components/src/components/ListVisual/Rows.js
edmofro: Nice!

packages/lesmis/src/components/DashboardReport.js
edmofro: I was going for using the dashboard item code as the base unit of reference that got passed around, and then within the dashboard report we have another hook that gets the report code at that point. This would ideally hit the cached react query from the `/dashboard` endpoint.

I started this on Tuesday end of day but ended up in circles so gave up. Might have another little crack today if I get a chance

packages/ui-components/src/components/ListVisual/ListVisual.js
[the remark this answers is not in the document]
edmofro: Yep no worries

edmofro (approved):
All the changes look good! Happy to approve without making that report codes change, I shouldn't be getting us stuck on something as academic as that when we're trying to fast track to the finish of LESMIS without using too much more dev time :-)

PR #2954, opened 2021-08-04
Use e2e aggregation for the e2e instance

kael89:
I realised that e2e instances correctly point to `e2e` aggregation, but the main baseline instance (e2e.tupaia.org) does not


edmofro (approved):

kael89:
@edmofro I tested the condition check locally, but I'm going to also test it with deploying instances having various names before merging this to dev

PR #2956, opened 2021-08-04
NZ-806: Apply email on timeout to exports

edmofro:
### Issue #:
Resolves NZ-806, NZ-805
Towards NZ-803

### Changes:
- Exports now go via download if they take too long
- If a spreadsheet is too large, it will be compressed

Note that the compression is only semi-successful, as depending on how you uncompress it, it may be assumed to be a directory rather than a file. I'm not bothering to fix this as I am currently working on NZ-815


edmofro:
Closing in favour of https://github.com/beyondessential/tupaia/pull/2963, which adds the download link strategy

PR #2957, opened 2021-08-04
Sol 277 Covid-Samoa Districts [hotfix]

nick-beyondessential:
### Changes:
- import sub_divisions into Samoa for Covid-Samoa project (Only exist in this project) This was done via migration rather than admin panel because:
  a. One of the districts (Unknown District has no geolocation, which hides from map but breaks admin import)
  b. avoid timing hassles with mixed import/migration of entity_relations
- migrate Samoa villages in Covid-Samoa to District(sub-district) parents.
- Remove facilities from Covid-Samoa
- Add household entity type
- import household entities and add hierarchy relations
- remove facility overlays
- includes sub_district overlays [sol-292]


rohan-bes (approved):

packages/database/src/migrations/20210803033646-CovidSamoaSubDistrictRelations-modifies-data.js
```diff
@@ -0,0 +1,68 @@
...
+  const covidSamoaId = (await findSingleRecord(db, 'entity_hierarchy', { name: 'covid_samoa' })).id;
+
+  // Fetch all sub_districts in Samoa
+  const samoaSubDistricts = (
+    await db.runSql(`
+      select id, parent_id, code from entity where type = 'sub_district' and country_code = 'WS';
+    `)
+  ).rows;
+
+  console.log('samoaSubDistricts: ', samoaSubDistricts);
```
PShoe, line 31: @nick-beyondessential  might want to remove this console.log at some point?

PR #2961, opened 2021-08-05
Sol 255 create default list of map overlay urls

BinduMadhaviTammireddy:
### Issue #:
https://linear.app/tupaia/issue/SOL-255/create-default-list-of-map-overlay-urls-for-e2e-tests

### Changes:
added mapOverlayUrls folder with default.json for regression testing and fanafana.json for snapshot testing

- Example

---


kael89 (approved):
Nicely done @BinduMadhaviTammireddy ! We should eventually sort the map overlay url config by id (can be done with a script) and also add more "default" urls, but this is a great start!

PR #2963, opened 2021-08-05
NZ-806 Email download link on export timeout

edmofro:
### Issue #:
Resolves NZ-806, NZ-805, NZ-813
Towards NZ-803

### Manual release steps:
- [ ] Add .env variable ADMIN_PANEL_SERVER_URL to meditrak-server

### Changes:
- Exports now go via email if they take too long
- Email contains a link to download the file, which will only work once, for the same user who requested it


packages/meditrak-server/src/s3/getUniqueFileName.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+const getRandomInteger = () => Math.floor(Math.random() * 1000000 + 1);
+
+export function getUniqueFileName(originalName = '') {
```
kael89, line 8: Good call extracting this function

packages/meditrak-server/src/apiV2/export/download/createDownloadLink.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import path from 'path';
+
+export const createDownloadLink = filePath => {
+  const fileName = path.basename(filePath);
+  return `http://localhost:8070/v1/export/download/${fileName}`;
```
kael89, line 10: The port may be configured to something different than `8070`

Also I see that this is used in the export email body, I guess this won't work for users that don't run tupaia locally

packages/meditrak-server/src/utilities/sendEmail.js
```diff
@@ -5,7 +5,7 @@
 import nodemailer from 'nodemailer';
 import { getIsProductionEnvironment } from '../devops';
 
-export const sendEmail = (to, subject, text) => {
+export const sendEmail = (to, subject, text, attachments) => {
```
kael89, line 8: (Not required, but helpful for devs)

```suggestion
export const sendEmail = async (to, subject, text, attachments) => {
```

packages/meditrak-server/src/apiV2/middleware/emailAfterTimeout.js
```diff
@@ -31,9 +31,9 @@ const setupEmailResponse = async (req, res, constructEmailFromResponse) => {
 
   // override the respond function so that when the endpoint handler finishes (or throws an error),
   // the response is sent via email
-  res.overrideRespond = responseBody => {
-    const { subject, message } = constructEmailFromResponse(responseBody);
-    sendResponseAsEmail(user, subject, message);
+  res.overrideRespond = async responseBody => {
```
kael89, line 34: Since we don't `await` for `sendResponseAsEmail()` in this function:

```suggestion
  res.overrideRespond = responseBody => {
```

kael89 (changes requested):
Great work @edmofro ! I love how the whole functionality is leveraging small utility functions with clearly defined responsibilities, makes this a pleasure to read!

Main requested change is related to the download url in the export email

packages/meditrak-server/src/apiV2/export/download/createDownloadLink.js
edmofro: Oh this is just a massive oversight, I should have come back to this! Thanks for spotting 🙏

edmofro:
I thought this one had been lost in the depths of the reviews channel - thanks for picking it up and spotting those details @kael89!

edmofro:
@kael89 ready for re-review. Now requires a manual step of adding an environment variable

kael89 (approved):
Looks good!

PR #2966, opened 2021-08-05
NZ-823 kobo additional improvements

EMcQ-BES:
### Issue #:
Resolves NZ-823 NZ-825

### Changes:

- Add `sync_service_log` db table
- Push KoBo logs to new table

### Screenshots:

https://linear.app/tupaia/issue/NZ-825#comment-1974723c


packages/database/src/migrations/20210805212932-AddSyncServiceLogTable-modifies-schema.js
```diff
@@ -0,0 +1,37 @@
...
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`
+    CREATE TABLE sync_service_log (
+      id TEXT PRIMARY KEY,
+      service_code TEXT NOT NULL,
+      service_type service_type NOT NULL,
```
EMcQ-BES, line 22: This is maybe unnecessary since `sync_service.code` are unique. I figured we might want to filter by `service_type` but we could also easily do that with a join.

edmofro: I like it :-)

Log tables don't need to be fully normalised - we'll probably push them into an external logging service at some point anyway

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -62,7 +61,7 @@ export async function syncWithKoBo(models, dataBroker, serviceCode) {
         }
         const entity = await transactingModels.entity.findOne({ code: responseData.orgUnit });
         if (!entity) {
-          winston.warn(`Skipping KoBo sync for unknown entity ${responseData.orgUnit}`);
+          await syncService.log(`Skipping KoBo sync for unknown entity ${responseData.orgUnit}`);
```
edmofro, line 64: Can we add more info here that might allow us to retry the record, e.g. the id?

edmofro (changes requested):
I realise this is a draft but early review

packages/meditrak-server/src/kobo/startSyncWithKoBo.js
EMcQ-BES: Done

edmofro (approved):
lgtm!

PR #2967, opened 2021-08-05
MEL-39: Use raw data table for viz-builder preview

chris-bes:
### Changes:

- Use raw data table for preview data in viz-builder


packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
```diff
@@ -86,9 +86,19 @@ const EditorContainer = styled.div`
   }
 `;
 
+const getColumns = data => {
+  const columnKeys = [...new Set(data.map(d => Object.keys(d)).flat())];
```
tcaiger, line 90: Nice use of Set, this is super concise.

tcaiger (approved):
The refactor is looking great. Thanks for doing this!

PR #2969, opened 2021-08-06
NZ-824: Update KoBoApi to fetch data in paginated chunks

EMcQ-BES:
### Changes:

- Add a max number of kobo entries to fetch
- Fetch from kobo in batches

### DIFF
--- a/packages/kobo-api/src/KoBoApi.js
+++ b/packages/kobo-api/src/KoBoApi.js
@@ -5,6 +5,7 @@
 import { fetchWithTimeout, stringifyQuery, takesDateForm } from '@tupaia/utils';
 
 const MAX_FETCH_WAIT_TIME = 15 * 1000; // 15 seconds
+const MAX_FETCH_ENTRIES = 50;
 
 export class KoBoApi {
   constructor() {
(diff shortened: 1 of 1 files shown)


packages/kobo-api/src/KoBoApi.js
```diff
@@ -5,6 +5,7 @@
 import { fetchWithTimeout, stringifyQuery, takesDateForm } from '@tupaia/utils';
 
 const MAX_FETCH_WAIT_TIME = 15 * 1000; // 15 seconds
+const MAX_FETCH_ENTRIES = 50;
```
EMcQ-BES, line 8: I picked `50` pretty much randomly. I was testing with `10`

edmofro (approved):
This looks great, and if KoBo do things right on their end should work perfectly. There is a tiny risk that their internal queries don't have a consistent results order, which would mean pagination might be inconsistent as a result. We can almost certainly rely on them to have got it right, but if we wanted to add an extra layer of confidence, we could send through a sort parameter with our fetch, e.g. sort by `_submission_time`. 

That would be extremely cautious though, and I'm definitely happy as is - so over to you!

EMcQ-BES:
> This looks great, and if KoBo do things right on their end should work perfectly. There is a tiny risk that their internal queries don't have a consistent results order, which would mean pagination might be inconsistent as a result. We can almost certainly rely on them to have got it right, but if we wanted to add an extra layer of confidence, we could send through a sort parameter with our fetch, e.g. sort by `_submission_time`.
> 
> That would be extremely cautious though, and I'm definitely happy as is - so over to you!

@edmofro I did consider that possibility, but this generates the same final url as KoBo generates for their `next` field so I presume they are consistent. More specifically from playing around I think by default the responses are sorted by `_id` from high to low.

edmofro:
Cool, seems good to me!

PR #2972, opened 2021-08-08
Viz builder epic

chris-bes:
### Changes:

- Example

---


packages/ui-components/src/components/Chart/Table/getChartTableData.js
```diff
@@ -78,14 +78,11 @@ const processData = ({ data, chartType }) => {
   return data;
 };
 
-export const useChartTable = viewContent => {
+export const getChartTableData = viewContent => {
```
tcaiger, line 81: @chris-bes - sorry one thing I missed in my review is that when you change the name of this function to `getChartTableData` it will throw an error about using a `hook` inside a regular function. [Hooks need to be called inside React functional components or custom hooks ](https://reactjs.org/docs/hooks-rules.html). I think we need to keep the `useMemos` for preparing the data for `react-table` so probably best to just rename this back to `useChartTable `

packages/ui-components/src/components/Map/Table/getMapTableData.js
```diff
@@ -44,15 +43,11 @@ const processData = (serieses, measureData) => {
   });
 };
 
-export const useMapTable = (serieses, measureData) => {
+export const getMapTableData = (serieses, measureData) => {
```
tcaiger, line 46: @chris-bes - same issue applies here about the use of a hook inside a regular function.

tcaiger:
Sorry, dis-regard my previous comments (which I just deleted). It seems to be fine to use useMemo in a normal function.

PR #2979, opened 2021-08-10
SOL-8: Aggregation Chaining in Tupaia Data Api

rohan-bes:
### Issue SOL-8:

**Overview:**
The goal of this feature is to allow multiple aggregation operations all in the database when fetching data via the Tupaia Data Api. The reasons for this are:
a) It's generally a good practice to perform as much computation in the DB as possible (especially things like aggregation)
b) We experience a particular performance hit whenever the db returns a large number of rows of analytics
c) For LESMIS, many reports require first a `FINAL_EACH_YEAR` aggregation before doing the `SUM_PER_PERIOD_PER_ORG_GROUP` aggregation that dramatically reduces the row count

**Implementation:**
We build up an sql query in `AnalyticsFetchQuery.js` that recursively wraps the base analytics fetch with an aggregation on that fetch. This way we can perform as many aggregations as requested (up til the first 'unsupported' aggregation). An example of the sql query looks like this:
```
FETCH
entity: PG, 
period: 2020, 
aggregations: ["FINAL_EACH_WEEK", { "type": "SUM_PER_PERIOD_PER_ORG_GROUP", "config": { "dataSourceEntityType": "facility", "aggregationEntityType": "country" }}]

QUERY
WITH entity_relations_a2 (code, aggregation_entity_code)
AS (VALUES ('PG_MADANG01','PG'), ('PG_MILNEBAY01','PG'), ('PG_MOROBE01','PG'), ('PG_Mugil','PG'), ('PG_NATIONALCAPITALDISTRICT01','PG'), ('PG_NEWIRELAND01','PG'), ('PG_SANDUAN01','PG'), ('PG_SIMBU01','PG'), ('PG_WESTERN02','PG'))

SELECT entity_code AS "entityCode", data_element_code AS "dataElementCode", period, value, type
FROM (

    SELECT aggregation_entity_code as entity_code, data_element_code, sum(value::numeric)::text as value, MAX(type) as type, MAX(day_period) as day_period, MAX(week_period) as week_period, MAX(month_period) as month_period, MAX(year_period) as year_period, MAX(date) as date, MAX(period) as period
    FROM (

        SELECT entity_code as entity_code, data_element_code, most_recent(value, date) as value, MAX(type) as type, MAX(day_period) as day_period, MAX(week_period) as week_period, MAX(month_period) as month_period, MAX(year_period) as year_period, MAX(date) as date, MAX(week_period) as period
        FROM (

            SELECT *, day_period as period from analytics
            INNER JOIN (
                VALUES ('PG_MADANG01'), ('PG_MILNEBAY01'), ('PG_MOROBE01'), ('PG_Mugil'), ('PG_NATIONALCAPITALDISTRICT01'), ('PG_NEWIRELAND01'), ('PG_SANDUAN01'), ('PG_SIMBU01'), ('PG_WESTERN02')
            INNER JOIN (
                VALUES ('BCD1')
            WHERE date >= '2020-01-01T00:00:00.000' AND date <= '2020-12-31T23:59:59.999') as base_analytics

        GROUP BY entity_code, data_element_code, week_period) as a1

    INNER JOIN entity_relations_a2 on entity_relations_a2.code = a1.entity_code
    GROUP BY aggregation_entity_code, data_element_code) as a2

ORDER BY date;

```

**Changes:**
- We now fully process the aggregation in the data-source and flag that the aggregation has been 'processed' in the response to data-broker. Aggregator will only run over the 'unprocessed' aggregations. As such, if an aggregation is to be performed in a data-source, it must be functionally identical to aggregator.
- Switched the Analytics table over to using period strings in the `day_period`, `week_period`, etc. columns. This is helpful for performing chained aggregations, makes the analytics table more readable, and opens up options in the future for us to easily perform fetch via query string instead of just startDate/endDate (allowing for dis-continuous period fetching)


packages/aggregator/src/Aggregator.js
```diff
@@ -32,9 +32,14 @@ export class Aggregator {
     return this.dataBroker.getDataSourceTypes();
   }
 
-  processAnalytics = (analytics, aggregationOptions, requestedPeriod) => {
+  processAnalytics = (analytics, aggregationOptions, aggregationsProcessed, requestedPeriod) => {
```
rohan-bes, line 35: Thoughts on the term `aggregationsProcessed`? Feels a bit wordy to me, but couldn't really think of anything nicer...

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -3,124 +3,172 @@
...
+    return true;
+  }
+
+  const switchSupportedKeys = aggregationSwitch.supportedConfigKeys || [];
+  return Object.keys(config).every(
+    key => COMMONLY_SUPPORTED_CONFIG_KEYS.includes(key) || switchSupportedKeys.includes(key),
+  );
+};
+
+export class AnalyticsFetchQuery {
```
rohan-bes, line 72: Decided to _yet again_ refactor the structure of this class. Now that the response type has deviated from the `EventsFetchQuery` I felt it would be best to decouple the two by removing the inheritance model and either shifting logic to `SqlQuery` or just duplicating between the two.

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -3,124 +3,172 @@
...
   }
 
-  getCommonFields() {
-    return [this.getEntityCodeField(), 'data_element_code'];
+  getEntityCodeField(aggregation) {
+    return aggregation.switch.aggregateEntities ? 'aggregation_entity_code' : 'entity_code';
   }
 
-  getEntityCommonTableExpression() {
+  getCtesAndParams(aggregation) {
```
rohan-bes, line 128: Rather than keeping the params in some internal state, I moved these functions to be more of 'build a query part and returns the associated params' kind of thing. I did that because the order of the params is very important, and so when building up query parts, the caller must ensure that the parts and the params are inserted in the same order.

packages/data-broker/src/DataBroker.js
```diff
@@ -104,15 +106,38 @@ export class DataBroker {
...
+          { analytics: source.results, aggregationsProcessed: sourceAggregationsProcessed },
+        ]);
+
+    return {
+      results: newResults,
+      metadata: {
+        dataElementCodeToName: {
+          ...target.metadata.dataElementCodeToName,
+          ...source.metadata.dataElementCodeToName,
+        },
```
rohan-bes, line 137: Changed the return signature a bit here! 
Previous:
```
{
  results: analytics,
  metadata: { ... },
}
```
Now:
```
{
  results: [{ analytics, aggregationsProcessed (number)}, ... ],
  metadata: { ... }
}
```
Basically, now instead of returning one set of analytics that is the merged result of all dataSources, we return an array of analytics batches that are merged by the number of aggregations they have processed. This allows aggregator to then run the remaining aggregations for each batch based on the number they've already processed.

Initially I didn't support this, and instead I just flagged to the datasources that they cannot process aggregations if the original fetch had multiple datasources in it. However, having multiple datasources is actually quite a common thing to do, especially when combining data elements and indicators in a fetch! So I felt it was worth it to support this, even if it makes the code a bit tricker to follow?

packages/aggregator/src/Aggregator.js
```diff
@@ -32,9 +32,15 @@ export class Aggregator {
     return this.dataBroker.getDataSourceTypes();
   }
 
-  processAnalytics = (analytics, aggregationOptions, requestedPeriod) => {
+  processAnalytics = (results, aggregationOptions, requestedPeriod) => {
     const { aggregations = [], filter } = aggregationOptions;
-    const aggregatedAnalytics = this.aggregateAnalytics(analytics, aggregations, requestedPeriod);
+    const aggregatedAnalytics = results.reduce((array, { analytics, aggregationsProcessed }) => {
+      const remainingAggregations = aggregations.slice(aggregationsProcessed);
```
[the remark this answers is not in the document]
edmofro: Oh ok I see now I've reviewed more of the PR: `aggregationsProcessed` is a number. I had pictured a subset of the original array being sliced off - good to know that's _not_ how `slice` works haha

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
```diff
@@ -190,8 +197,8 @@ describe('DataBroker', () => {
     });
 
     describe('events', () => {
-      const assertServicePulledEventsOnce = (service, dataElements) =>
-        expect(service.pull).toHaveBeenCalledOnceWith(dataElements, 'dataGroup', options);
+      const assertServicePulledEventsOnce = (service, dataElements, optionsOverride = options) =>
```
edmofro, line 200: I can't see how this and the similar case above are actually using `optionsOverride`. In every instance it seems to just pass in the same default options?

packages/data-broker/src/DataBroker.js
edmofro: Yep agreed. Having a comment above the function that made it obvious what was happening means it wouldn't even need to be fully followed 🙏

edmofro: I wonder if rather than the pure number of aggregations processed, data-api should be passing back the sub-array of aggregations processed, as it was passed in? Now that the response format has changed here, it could be nice to avoid baking in the same set of aggregations being run across all data sources. Just thinking of our discussion today, in case we ever wanted to push poly-aggregation support into the data services.

edmofro (approved):
Mostly pretty familiar, so no major requested changes here at all!

General comments:
1. Very happy with the removal of the weird building up of parameters - that model worked a lot better with only one or two things being pushed in, but is hopeless now we've got so many
2. I generally prefer a double spread than array.concat, but given this is all about performance optimisation, the concat is definitely justified! I wasn't sure in the past whether it was faster, but confirmed it is: https://stackoverflow.com/a/57191593/1988532
3. The new response format from data-api is a bit unexpected, so I'm even more glad we have `aggregator` in the pipeline to be a common place to interpret it!

Very cool stuff @rohan-bes 😁

edmofro:
@rohan-bes regarding 

>  Rebuild analytics table to use new period formatting

We've had to do the rebuild a few times now, and if it keeps happening I wonder if it's worth trying to make a zero downtime process that builds to a different table and then does the drop-then-rename process? On the other hand I keep feeling like "well it'll be stable after this" haha

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -3,124 +3,166 @@
...
+    this.aggregations = [];
+    if (options.aggregations) {
+      for (let i = 0; i < options.aggregations.length; i++) {
+        const aggregation = options.aggregations[i];
+        const aggregationSwitch = AGGREGATION_SWITCHES[aggregation?.type];
+        if (!aggregationSwitch || !supportsConfig(aggregationSwitch, aggregation?.config)) {
+          break; // We only support chaining aggregations up to the last supported type
+        }
+        const dbAggregation = { type: aggregation?.type };
+        dbAggregation.switch = AGGREGATION_SWITCHES[aggregation?.type]; // add internal switch
```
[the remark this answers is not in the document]
rohan-bes: Ahhh I see ya! Hmmm I think I'm okay with it either way? Both make sense 👍

packages/data-api/src/AnalyticsFetchQuery.js
```diff
@@ -3,124 +3,166 @@
...
+    fields.push(`${this.getEntityCodeField(aggregation)} as entity_code`);
+    fields.push(`data_element_code`);
+    fields.push(`${aggregationFunction} as value`);
+    fields.push('MAX(type) as type');
+    fields.push('MAX(day_period) as day_period');
+    fields.push('MAX(week_period) as week_period');
+    fields.push('MAX(month_period) as month_period');
+    fields.push('MAX(year_period) as year_period');
+    fields.push('MAX(date) as date');
+    fields.push(`MAX(${groupByPeriodField}) as period`);
```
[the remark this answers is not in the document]
rohan-bes: Great suggestion! Will do!

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
rohan-bes: Good catch, I had changed these tests cos previously I was overriding the options at points, but then removed that code but forgot to update here! Will fix

rohan-bes:
>I wonder if rather than the pure number of aggregations processed, data-api should be passing back the sub-array of aggregations processed, as it was passed in? Now that the response format has changed here, it could be nice to avoid baking in the same set of aggregations being run across all data sources. Just thinking of our discussion today, in case we ever wanted to push poly-aggregation support into the data services.

I did actually do this initially, and in some ways I like it better so I'm cool with going back to doing it that way, but here's why I changed my mind:
1. Given that aggregations must be processed in order, holding a number for how got processed _should_ be sufficient to know which ones were processed
2. I was thinking a bit about trying to minimize the amount of data passed through the various service boundaries. It's not really an issue for now, since it all happens within the one node process, but it's pretty realistic to assume that we may one day split out the internal datasource (tupaia data-api and indicators) into separate servers/processes in which case the amount of data that gets serialized while communicating does matter (a bit). Usually this would be a fairly minor concern, but sometimes we have orgUnitMaps that can be rather large...
3. It would be less performant in data-broker to do a deep equality check between an array of processedAggregations than to just compare a number. On the other hand, just comparing array length felt a bit odd/lazy (very weak point I do appreciate)
4. I found myself calling `aggregationsProcessed.length` a **lot** which felt a bit silly

I think the strongest point here is 1, I just can't see a way that we could process aggregations out of order, so it seems to me that a number would be sufficient even if we moved to more complex aggregation processing system?

I do however feel that returning an array of aggregations processed could make debugging a bit easier in some scenarios, as it's more descriptive?

Happy to go back to doing that if you feel it's best, but wanted to let you know why I went against it in the first place 👍

rohan-bes:
> We've had to do the rebuild a few times now, and if it keeps happening I wonder if it's worth trying to make a zero downtime process that builds to a different table and then does the drop-then-rename process? On the other hand I keep feeling like "well it'll be stable after this" haha

Oooh that's a really good idea! I actually had another thought that maybe we could move the logic of the scripts that build/drop the analytics table into database functions themselves so that we can call them in migrations?

We've had to change it a few times, and I think realistically we will need to change it a few more. I might create a card to set up this and we could pull it in as DX work?

packages/data-api/src/SqlQuery.js
```diff
@@ -4,11 +4,17 @@
...
 
 export class SqlQuery {
-  static parameteriseArray = arr => `(${arr.map(() => '?').join(',')})`;
+  static array = arr => `(${arr.map(() => '?').join(',')})`;
 
-  static parameteriseValues = rows =>
+  static values = rows =>
     `VALUES (${rows.map(values => values.map(() => `?`).join(',')).join('), (')})`;
 
+  static innerJoin = (codes, columnName) => `
```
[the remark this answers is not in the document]
rohan-bes: Good point about keeping this class generic for possible re-use purposes. I've reworked the method signature to take in a base table name as well 👍

packages/data-api/src/AnalyticsFetchQuery.js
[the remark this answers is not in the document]
edmofro: Yeah same here :+1:

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
edmofro: One down, one to go ;-)

edmofro:
On the aggregations processed format: I agree that under the current system, a number is just fine, so we should go with that. 

My worry was more about future-proofing that avoids baking in "one set of aggregations gets processed for the whole fetch", where we might want to change it to use fairly complex query planning and handling, e.g.
- we pass through a map of aggregations by data element right to the service, allowing the data service to optimally fetch the data for one report however it sees fit
- similar but less likely: each data service (mainly talking internal data-api here) might decide to apply more of the aggregations to some data elements, but less to others, for some crazy reason about it working out the relative tradeoff of doing stuff in db vs. node based on number of rows in the analytics table etc. 

Having thought about it more though, I think both require a significant amount of refactoring around this area, so at that point changing from the number to the list of aggregations wouldn't be a dramatic component of the work, and there's therefore no point in doing it ahead of time. Let's stick with the simple option!

In that case though, I wonder if there's a better name that indicates it's a count, so others like me don't get confused?

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
rohan-bes: Ahh woops 🤦‍♂️ Thank u!

rohan-bes:
> On the aggregations processed format: I agree that under the current system, a number is just fine, so we should go with that.
> 
> My worry was more about future-proofing that avoids baking in "one set of aggregations gets processed for the whole fetch", where we might want to change it to use fairly complex query planning and handling, e.g.
> 
> * we pass through a map of aggregations by data element right to the service, allowing the data service to optimally fetch the data for one report however it sees fit
> * similar but less likely: each data service (mainly talking internal data-api here) might decide to apply more of the aggregations to some data elements, but less to others, for some crazy reason about it working out the relative tradeoff of doing stuff in db vs. node based on number of rows in the analytics table etc.
> 
> Having thought about it more though, I think both require a significant amount of refactoring around this area, so at that point changing from the number to the list of aggregations wouldn't be a dramatic component of the work, and there's therefore no point in doing it ahead of time. Let's stick with the simple option!
> 
> In that case though, I wonder if there's a better name that indicates it's a count, so others like me don't get confused?

Oooh I get you now, great point. Honestly I hadn't considered how we might attempt to defer coordination of aggregations being processed to a level lower than `Aggregator`, but you're right we may do just that! Will keep this in mind for sure 👍 

Good call on the rename, I think I may have even got confused myself at times too! Personally I find `aggregationsProcessed` to be a tad wordy as well as being a bit unclear, but not too sure how best to shorten it... Could always go with the classic `numAggregationsProcesssed`? `aggregationsIndex` feels a tad too unclear for me but I do like how short it is... I'll mull on this one a bit this morning but will fall back to `numAggregationsProcesssed` if I can't think of anything better 👍

edmofro:
Have spent a couple mins thinking and don't have anything better than `numAggregationsProcessed` 🤷

edmofro (approved):

PR #2980, opened 2021-08-10
Nz 27 export table flippas to excel

tcaiger:
### Changes:

- Example

---

### DIFF
--- a/packages/ui-components/src/components/DataTable/useDataTableExport.js
+++ b/packages/ui-components/src/components/DataTable/useDataTableExport.js
@@ -1,0 +1,48 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { useImperativeHandle } from 'react';
+import xlsx from 'xlsx';
+import { stringToFilename } from '@tupaia/utils';
+import moment from 'moment';
+
+export const useDataTableExport = (ref, rows, columns) => {
+  useImperativeHandle(ref, () => ({
+    exportData(title) {
+      // Get data from react table properties
+      const body = rows.map(row => row.values);
+      const header = columns.map(col => col.id);
+
+      // Make xlsx worksheet
+      // @see https://github.com/SheetJS/sheetjs
+      const sheet = xlsx.utils.aoa_to_sheet([[title]]);
+      xlsx.utils.sheet_add_json(sheet, body, {
+        header,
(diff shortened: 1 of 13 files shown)


packages/ui-components/src/components/DataTable/useDataTableExport.js
```diff
@@ -0,0 +1,56 @@
...
+    const sheetName = `Export on ${date}`;
+    const workbook = { SheetNames: [sheetName], Sheets: { [sheetName]: sheet } };
+
+    // Make filename
+    const fileName = title
+      ? stringToFilename(`export-${title}-${date}.xlsx`)
+      : `export-${date}.xlsx`;
+
+    // Write file. This will trigger the file download in the browser
+    return xlsx.writeFile(workbook, fileName);
```
chris-bes, line 52: I think `xlsx` also has `writeFileAsync()` which may be handy if it's trying to write a large file as I assume it won't block the process. However I've never tried it and not sure if it triggers the file download automatically so happy to go with this one

packages/utils/src/string.js
```diff
@@ -32,3 +33,10 @@ export const getArticle = word => {
   }
   return word.match(/^[aeiou]/i) ? 'an' : 'a';
 };
+
+const truncate = (string, length) => (string.length <= length ? string : string.slice(0, length));
+
+export const stringToFilename = string => {
+  const sanitized = sanitize(string).replace(/\s+/g, '-').toLowerCase().toLowerCase();
```
chris-bes, line 40: Only 1 `toLowerCases()` should be enough?

```suggestion
  const sanitized = sanitize(string).replace(/\s+/g, '-').toLowerCase();
```

chris-bes (approved):
LGTM! I only have 2 minor suggestions so pre approving. 1 is we don't need useSortBy when exporting excel and the other one is just a typo (double `useLowerCase()`)

packages/ui-components/src/components/DataTable/useDataTableExport.js
```diff
@@ -0,0 +1,56 @@
...
+import { useSortBy, useTable } from 'react-table';
+import moment from 'moment';
+
+export const useDataTableExport = (columnsData, data, title) => {
+  const { rows, columns } = useTable(
+    {
+      columns: columnsData,
+      data,
+    },
+    useSortBy,
```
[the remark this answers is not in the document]
tcaiger: yes great point

PR #2981, opened 2021-08-11
Add comment to avoid calculation when checking max length

Alastair-L:
### Issue #: No issue

Just adding a comment, as I have put 64 - 17 - 1 into my calculator about 5 times 😂. (Could also be useful for any new devs that join in the future)

### DIFF
--- a/packages/devops/scripts/ci/validateBranchName.sh
+++ b/packages/devops/scripts/ci/validateBranchName.sh
@@ -11,6 +11,8 @@ SUBDOMAIN_SUFFIXES=(admin admin-api aggregation api config export lesmis lesmis-
 MAX_SUBDOMAIN_LENGTH=64
 MAX_SUBDOMAIN_SUFFIX_LENGTH=$(get_max_length "${SUBDOMAIN_SUFFIXES[@]}")
 MAX_BRANCH_NAME_LENGTH=$((MAX_SUBDOMAIN_LENGTH - ${MAX_SUBDOMAIN_SUFFIX_LENGTH} - 1)) # Subtract 1 for the connecting `-`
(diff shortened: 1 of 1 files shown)


kael89 (approved):
Looks good! (But you know what they say about comments being easy to become outdated...)

Another solution would be to `echo $MAX_BRANCH_NAME_LENGTH` in your local, for debugging purposes. Or if you think this information is always useful, perhaps update the script to print the max length even when the check succeeds. Currently it does so on failure, see line 41

Alastair-L:
Yeah that's a good point about echoing it (I never think to run ci/cd scripts locally)!

I definitely see what you mean about the comment becoming outdated, but:
a) The comment is datestamped, which may assist in determining if it is outdated or not
b) `tonga-aggregation` is pretty long, so I see it as entirely possible that we'll have this limit for a long time
c) The comment is next to all logic which could possibly change the length, so any future changer is likely to see it and update it.

I don't think that this information is useful enough to merit always logging it unfortunately (I think forgetting the max length might be a bit of a me thing)

Alastair-L:
Also, may I merge this without testing as it is just a comment?

kael89:
> Yeah that's a good point about echoing it (I never think to run ci/cd scripts locally)!
> 
> I definitely see what you mean about the comment becoming outdated, but:
> a) The comment is datestamped, which may assist in determining if it is outdated or not
> b) `tonga-aggregation` is pretty long, so I see it as entirely possible that we'll have this limit for a long time
> c) The comment is next to all logic which could possibly change the length, so any future changer is likely to see it and update it.
> 
> I don't think that this information is useful enough to merit always logging it unfortunately (I think forgetting the max length might be a bit of a me thing)

I also agree with your points, they should mitigate the risk of this comment becoming outdated 🙂 Feel free to merge directly!

PR #2982, opened 2021-08-11
MEL-391: [e2e] support date range in map overlay urls

kael89:
### Issue #:
MEL-391

Also added a script that can be used to format e2e config files. It sorts url objects by id/code.

Note that date range in map overlay urls is currently broken on navigation (see SOL-101). We'd need to fix this issue before the new functionality can be actually used in e2e tests


packages/web-frontend/cypress/scripts/generateConfig/generateOverlayConfig.js
```diff
@@ -19,6 +19,8 @@ const urlSchema = buildUrlSchema({
     id: oneOrArrayOf(yup.string().required()),
     project: yup.string().required(),
     orgUnit: yup.string().required(),
+    startDate: yup.string(),
+    endDate: yup.string(),
```
rohan-bes, line 23: Does this need `overlayPeriod` as well?

rohan-bes (approved):
LGTM! Only couple of minor comments 👍

packages/utils/src/object.js
```diff
@@ -37,6 +37,29 @@ export function getSortByKey(key, options) {
   return getSortByExtractedValue(o => o[key], options);
 }
 
+export const orderBy = (array, valueMappers, orders = []) => {
```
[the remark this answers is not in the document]
kael89: That's a fair point! The import statement will be different of course, but people could be confused.

I don't expect this to be an issue in practice... I think it's more probable that people will be surprised by the default `lodash` behaviour of sorting `'100'` before `'20'` (at least I did). So in a sense it's like "protecting" them from this unpleasant surprise too 😄 

By the way we do something similar with `mapKeys` and `mapValues`, which are also lodash functions. I think I introduced those, but at the time I didn't know there were part of `lodash`.

I'm open to other names and in general I agree with your reasoning, but in this specific case I'd vote to keep it as is

packages/web-frontend/cypress/scripts/generateConfig/generateOverlayConfig.js
kael89: Good observation! Made me realise I should also extract `overalyPeriod` when parsing string urls to objects (see `parseUrl()`) 👀 

The reason I don't include it in the schema is to "discourage" people from directly specifying `overlayPeriod` in their configs (they should use `startDate`/`endDate` instead)... We do the same in report config parsing. But truth is people can still pass this in, if they want (no error is thrown) 😀 

So in essence, `overlayPeriod` should just be used by local objects created by the script, to avoid doing an unnecessary conversion from `url?overlayPeriod=period` => `{ startDate, endDate }` => `url?overlayPeriod`

kael89:
Merging straight into `dev` as this affects e2e tests only

PR #2983, opened 2021-08-11
Sol 311 build indicators for pmos

billli0:
### Issue #:
https://linear.app/tupaia/issue/SOL-311/build-indicators-for-pacmossi


rohan-bes (approved):
Awesome @billli0 this should make things heaps simpler 👍

PR #2984, opened 2021-08-11
Sol 295 dashboard number of tests conducted

Alastair-L:
### Issue #: SOL-295

New dashboard (my first using the report server!). Thanks to @billli0 for showing me the ropes.


Alastair-L:
The base can be considered `dev` for review purposes, all changes in `allow-non-5432-db-ports` are just for my local convenience (I hope they'll be merged soon)

packages/database/src/migrations/20210804063133-AddFijiCOVIDDashboardTestsByDay-modifies-data.js
```diff
@@ -0,0 +1,145 @@
...
+      ],
+      dataElements: ['COVIDTest_FJNumTest'],
+    },
+    transform: [
+      {
+        '...': 'drop',
+        value: 'sum',
+        period: 'group',
+        transform: 'aggregate',
+      },
```
rohan-bes, line 45: Nothing major, but I think I have a slight preference to doing aggregations in the `aggregations` part of `fetch`. eg. `SUM_EACH_DAY`.

We've been introducing performance optimisations to that process, where we actually run the aggregation itself at the database level that can substantially speed up fetch speed in some cases. 

I suspect in this case it probably won't make much of a difference (unless there were many responses per day, then it would actually be rather useful) so consider it optional, but just wanted to mention my general preference towards aggregating during the fetch if possible 👍

packages/database/src/migrations/20210804063133-AddFijiCOVIDDashboardTestsByDay-modifies-data.js
```diff
@@ -0,0 +1,145 @@
...
+      {
+        "'name'": "periodToDisplayString($row.period, 'DAY')",
+        transform: 'select',
+        "'timestamp'": 'periodToTimestamp($row.period)',
+        "'Tests'": '$row.value',
+        '...': [],
```
rohan-bes, line 51:
```suggestion
```
Think I mentioned this in another PR, but this field is optional 👍

rohan-bes (approved):
Looks good to me! A suggestion to specify the aggregation in the fetch rather than using an `aggregate` transform but consider it optional 👍 

Also, might wanna merge dev back into this PR, as I believe that will resolve all the diffs which have the `DB_PORT` stuff?

packages/database/src/migrations/20210804063133-AddFijiCOVIDDashboardTestsByDay-modifies-data.js
Alastair-L: Thanks Rohan, this is fantastic :))

packages/database/src/migrations/20210804063133-AddFijiCOVIDDashboardTestsByDay-modifies-data.js
Alastair-L: Amazing

edmofro:
Darn, this one actually passed testing but will need to be revised using the new format and then tested again @Alastair-L

Alastair-L:
Thanks Edwin for the heads up. 
Closing in favor of new PR (probably not needed but just to keep things clean): https://github.com/beyondessential/tupaia/pull/3206

PR #2985, opened 2021-08-11
2729: Fixes (MEL-92)

kael89:
@edmofro Eventually I decided to follow your advice and disallow updating the reporting period for surveys that have responses against them.

Some of our surveys already have ~ 90,000 responses. As we have discussed, updating them all at once breaks `meditrak-server` and/or the database. I tried batching the changes and adding delays between each batch: this is successful but not robust, since

* Only relatively small batches are guaranteed to not stress the system (eg 1000 responses/minute). Bigger batches often work, but in combination with analytic updates that are triggered at the same time, we cannot guarantee the stability of the system. I guess a more robust solution would be using a db-based queue
* We can assume that people will not change period granularities often
* We can still use migrations to mark survey response as changed and thus outdate them in a controlled way

### DIFF
--- a/packages/meditrak-server/src/database/models/Survey.js
+++ b/packages/meditrak-server/src/database/models/Survey.js
@@ -19,12 +19,12 @@ class SurveyType extends DatabaseType {
   async questions() {
     const questions = await this.database.executeSql(
       `
-      SELECT q.* FROM question q
-      JOIN survey_screen_component ssc ON ssc.question_id  = q.id
-      JOIN survey_screen ss ON ss.id = ssc.screen_id
-      JOIN survey s ON s.id = ss.survey_id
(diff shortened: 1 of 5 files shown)


packages/meditrak-server/src/database/models/Survey.js
```diff
@@ -43,6 +43,9 @@ class SurveyType extends DatabaseType {
     const countries = await this.getCountries();
     return countries.map(c => c.code);
   }
+
+  hasResponses = async () =>
+    !!(await this.otherModels.surveyResponse.findOne({ survey_id: this.id }));
```
edmofro, line 48: Nice thought limiting it to 1 to reduce db and memory overhead

edmofro (approved):
lgtm!

PR #2988, opened 2021-08-12
NZ-2 LESMIS Target Districts

EMcQ-BES:
### Issue #:
Resolves NZ-2

### Changes:

- Add `{ type: LESMIS_Target_District }` to 40 target district attributes


rohan-bes (approved):
Nice, LGTM!

PR #2990, opened 2021-08-12
Standardise vitals titles

EMcQ-BES:
### Issue #:
Resolves NZ-36

### Changes:

- Change headings in the vitals sections of lesmis to standardised "details"

---

### Screenshots:
See attached ticket


tcaiger (changes requested):
The changes look good. It wasn't explicitly mentioned in the ticket description but the mock up has a new field for the "Name of province". It would be good to add this since we are removing the parentProvince name and the code is not as human readable. The same thing applies to parentDiscrict name as well.

37fdeef6-1432-48b0-bfb7-90efdc4047e8

tcaiger (approved):
Looks good

PR #2993, opened 2021-08-12
Mel 372 tupaia client 

IgorNadj:
MEL-372

Manual release steps:
[ ] Tell devs to add required .env vars
[ ] Add required env vars to production/dev/etc


packages/api-client/src/connections/ApiConnection.ts
```diff
@@ -0,0 +1,108 @@
...
+      fetchConfig.body = JSON.stringify(body);
+    }
+
+    const response = await this.fetchWithTimeout(queryUrl, fetchConfig);
+    await this.verifyResponse(response);
+    return response.json();
+  }
+
+  private async fetchWithTimeout(url: string, config: RequestInit, timeout: number = DEFAULT_MAX_WAIT_TIME): Promise<Response> {
+    return nodeFetch(url, {...config, timeout } );
```
IgorNadj, line 80: Reimplemented this method to avoid importing from utils. Because we are publishing this package, we can't have it depend on @tupaia/* otherwise the whole of tupaia will be shipped with the package.

packages/api-client/src/connections/ApiConnection.ts
```diff
@@ -0,0 +1,108 @@
...
+      if (responseJson.error) {
+        throw new Error(`API error ${response.status}: ${responseJson.error}`)
+      }
+    }
+  };
+
+  private stringifyQuery(baseUrl: string = '', endpoint: string, queryParams: QueryParameters): string {
+    const urlAndEndpoint = baseUrl ? `${baseUrl}/${endpoint}` : endpoint;
+
+    const queryString = stringify(queryParams);
```
IgorNadj, line 102: Using qs library. See comment above on why not just use the one in utils

packages/api-client/src/connections/EntityApi.ts
```diff
@@ -101,14 +99,10 @@ type Prefix<T, P extends string> = {
   [field in keyof T & string as `${P}_${field}`]: T[field];
 };
 
-export class EntityApi implements MicroserviceApi {
-  public baseUrl = ENTITY_API_URL;
```
IgorNadj, line 105: Externalised this config variable. 

Previously this meant that EntityApi needs to know to look on process.env, however, the api client may be configured in a different way. E.g. in tamanu they don't use env vars. 

New implementation has the TupaiaApiClient responsible for picking good default base urls, and these can be overridden by passing in custom ones, e.g. for feature branches.

packages/api-client/src/connections/MeditrakApi.ts
```diff
@@ -0,0 +1,25 @@
...
+  answers: Answers;
+}
+
+export type Answers = {
+  [key: string]: string; // question_code -> value
+}
+
+export class MeditrakApi extends BaseApi {
+
+  async createSurveyResponses(responses: SurveyResponse[]): Promise<void> {
```
IgorNadj, line 21: The existing MeditrakConnections we could have used use the postChanges endpoint, which requires generateId() etc. to be done before the request is made to Meditrak. And then we would have had to copy those functions into this package for publishing.

I decided on using POST /surveyResponse because it didn't have these extra requirements.

It seems both API endpoints are in active development so it seems there is a choice to choose between the two.

This is why this class looks very empty.

packages/api-client/src/constants.ts
```diff
@@ -0,0 +1,19 @@
+export const DATA_TIME_FORMAT = 'YYYY-MM-DD HH:mm:ss';
+
+export type EndpointBaseUrlSet = {
+  entity: string;
+  meditrak: string;
+  report: string;
+}
+
+export const LOCALHOST_ENDPOINT_BASE_URLS: EndpointBaseUrlSet = {
```
IgorNadj, line 9: For convenience, allows you to use the default set of localhost urls.

However most people will use this package via server-boilerplate, which reads and passes in the .env vars, and feature branches and production will do the same.

packages/lesmis-server/src/connections/EntityConnection.ts
```diff
@@ -4,12 +4,15 @@
  *
  */
 import camelcaseKeys from 'camelcase-keys';
-import { QueryParameters } from '@tupaia/server-boilerplate';
+import { QueryParameters } from '@beyondessential/tupaia-api-client';
```
IgorNadj, line 7: I decided to move all of the api-client logic out of server-boilerplate so that devs wouldn't add more code there. So we have some quite unsightly imports used in deprecated classes.

packages/lesmis-server/src/connections/SessionHandlingApiConnection.ts
```diff
@@ -3,7 +3,8 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { ApiConnection, AuthHandler, SessionType } from '@tupaia/server-boilerplate';
+import { ApiConnection, AuthHandler } from '@beyondessential/tupaia-api-client';
```
IgorNadj, line 6: As above

>I decided to move all of the api-client logic out of server-boilerplate so that devs wouldn't add more code there. So we have some quite unsightly imports used in deprecated classes.

packages/admin-panel-server/src/connections/ReportConnection.ts
```diff
@@ -1,16 +0,0 @@
-/**
```
IgorNadj, line 1: Nothing used these. Deleted.

packages/api-client/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "types": "./dist/**/**.d.ts",
+  "license": "SEE LICENSE IN LICENSE",
+  "scripts": {
+    "build": "babel src --out-dir dist --ignore \"src/__tests__/**\" --extensions .ts --config-file \"../../.babelrc-ts.js\"",
+    "lint": "tsc; eslint . --ext .ts",
+    "lint:fix": "yarn lint --fix",
+    "build:typedefs": "tsc",
+    "test": "jest --runInBand",
+    "test:coverage": "yarn test --coverage",
+    "publish:run": "yarn publish --no-git-tag-version",
```
IgorNadj, line 15: `yarn publish` is a very handly tool that increments the version, commits the change and tags the commit. However, it doesn't know we are using a monorepo, and we don't want tags for this package to be pushed into the monorepo. So I wrote publish:run to be used instead.

packages/api-client/src/connections/EntityApi.ts
```diff
@@ -14,12 +14,10 @@ import type {
   EntityResponseObject,
 } from '@tupaia/entity-server/src/type-exports';
 
-import { FilterCriteria, AdvancedFilterValue } from '../models';
+import type { FilterCriteria, AdvancedFilterValue } from '@tupaia/server-boilerplate/src/type-exports';
```
IgorNadj, line 17: Importing types from other packages is ok. They don't count towards being an npm dependency. 

However, these imported types are not present in the published package. So currently other users of this package do not get tupaia internal package typings. I looked at bundling them all up into a single .d.ts file, and some people do this, but it's not official or foolproof. So I went with this MVP solution.

rohan-bes (commented):
Few comments, but mainly wanted to discuss the package structure with you at some point. I was a little surprised to see that:
a) We had a non-internal package within the mono-repo, usually when we have packages that are intended for consumption by 3rd parties, we put them in a separate repo. I can totally see why we'd want to keep them in the mono-repo, but it's rather atypical
b) That the package isn't being included in our usual yarn workspace. And that when we do use this package internally, we are referencing the package bundle we'd get from npm. I'm not sure what support we'd have a for a hybrid model here, where the package can both be used internally via the workspace and exposed/published externally in npm, but I have a feel we'd really quite like to use it as a workspace package internally if possible

I may also need a hand building the type definitions locally? I got a bunch of ts errors coming from other internal packages when I tried `yarn build:typedefs`. I'm guessing this is because we are importing types from `server-boilerplate` and `entity-server`? Makes sense, happy to chat about what we should do about this 👍

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
```diff
@@ -18,6 +18,7 @@ jest.mock('@tupaia/database', () => ({
   ModelRegistry: jest.fn().mockImplementation(() => mockModels),
 }));
 
+// TODO: fix this mock
```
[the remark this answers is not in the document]
IgorNadj: Ah this was a todo for myself when I had a different implementation in mind. Reverted.

packages/api-client/src/constants.ts
```diff
@@ -0,0 +1,19 @@
+export const DATA_TIME_FORMAT = 'YYYY-MM-DD HH:mm:ss';
```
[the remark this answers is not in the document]
IgorNadj: Added

packages/server-boilerplate/src/microService/api/ApiBuilder.ts
```diff
@@ -53,15 +54,17 @@ export class ApiBuilder {
...
+        meditrak: process.env.MEDITRAK_API_URL,
+        report: process.env.REPORT_API_URL,
+      }
 
       const context: RequestContext = {
-        microServices: {
-          entityApi,
-        },
-      }; // context is shared between request and response
+        services: new TupaiaApiClient(microServiceAuthHandler, baseUrls),
```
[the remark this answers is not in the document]
IgorNadj: Yeah I couldn't think of a good name for this either. AWS calls these clients, e.g. s3client, but I'm not sure I like that name. Services is a good generic name methinks. I agree I don't think it's too important as it's easy to change.

PR #2994, opened 2021-08-12
Nz 811 lesmis dashboard png exports

tcaiger:
### Changes:

- Example

---

### DIFF
--- a/packages/lesmis/src/components/DashboardReportModal.js
+++ b/packages/lesmis/src/components/DashboardReportModal.js
@@ -5,6 +5,7 @@
  */
 import React, { useRef, useState } from 'react';
 import styled from 'styled-components';
+import { sleep, stringToFilename } from '@tupaia/utils';
 import { useTheme } from '@material-ui/core/styles';
 import {
   Box,
@@ -13,9 +14,7 @@ import {
   Typography,
   Dialog as MuiDialog,
   Container as MuiContainer,
-  IconButton,
 } from '@material-ui/core';
-import GetApp from '@material-ui/icons/GetApp';
 import downloadJs from 'downloadjs';
 import domtoimage from 'dom-to-image';
 import CircularProgress from '@material-ui/core/CircularProgress';
@@ -26,6 +25,7 @@ import { DialogHeader } from './FullScreenDialog';
 import { useDashboardReportDataWithConfig } from '../api/queries';
 import { useUrlParams, useUrlSearchParams } from '../utils';
 import { DashboardReport } from './DashboardReport';
+import { SplitButton } from './SplitButton';
 
 const Transition = React.forwardRef(function Transition(props, ref) {
   return <Slide direction="left" ref={ref} {...props} />;
@@ -58,7 +58,7 @@ const Header = styled(FlexSpaceBetween)`
 `;
 
 const Toolbar = styled(FlexStart)`
-  display: ${props => (props.isExporting ? 'none' : 'flex')};
+  display: ${props => (props.exporting ? 'none' : 'flex')};
 `;
 
 const Heading = styled(Typography)`
@@ -75,7 +75,7 @@ const Description = styled(Typography)`
 `;
 
 const ExportLoader = styled(FlexColumn)`
-  display: ${props => (props.isExporting ? 'flex' : 'none')};
+  display: ${props => (props.exporting ? 'flex' : 'none')};
   align-items: center;
   position: absolute;
   top: 0;
@@ -85,6 +85,11 @@ const ExportLoader = styled(FlexColumn)`
   background: white;
   z-index: 1;
 `;
+
+const EXPORT_OPTIONS = [
+  { id: 'xlsx', label: 'Export to XLSX' },
+  { id: 'png', label: 'Export to PNG' },
+];
 
 const exportToPng = (node, filename) => {
   return new Promise(resolve => {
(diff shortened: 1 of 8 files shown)


packages/utils/src/string.js
```diff
@@ -32,3 +33,10 @@ export const getArticle = word => {
   }
   return word.match(/^[aeiou]/i) ? 'an' : 'a';
 };
+
+const truncate = (string, length) => (string.length <= length ? string : string.slice(0, length));
+
+export const stringToFilename = string => {
```
rohan-bes, line 39: Interesting piece of PR cross pollination here... @tcaiger you may be interested to know that @kael89 just added a `utils/src/filesystem.js` file which has a few functions for dealing with I/O (https://github.com/beyondessential/tupaia/pull/2982/files#diff-c309acd37e4d0905cc5faa87d21a46e5751fbbb6c8e115696c91197709bcda99). I feel this function _may_ be slightly more at home there (and rename to just `toFilename()`) but it's much of a muchness 👍

kael89: Great work picking up cross-PR similarities @rohan-bes ! 👏

packages/ui-components/src/components/SplitButton.js
```diff
@@ -0,0 +1,118 @@
...
+    setSelectedId(id);
+    setOpen(false);
+  };
+
+  const handleToggle = () => {
+    setOpen(prevOpen => !prevOpen);
+  };
+
+  const handleClose = event => {
+    if (anchorRef.current && anchorRef.current.contains(event.target)) {
```
edmofro, line 52: What's happening here? Could you add a comment for front end noobs like me?

edmofro: Oh worked it out
```suggestion
    // if the user clicks away, but their click still lands on some part of the component, keep open
    if (anchorRef.current && anchorRef.current.contains(event.target)) {
```

packages/lesmis/src/utils/useExportToPNG.js
```diff
@@ -0,0 +1,42 @@
...
+    const node = exportRef.current;
+
+    setIsExporting(true);
+    setIsExportLoading(true);
+
+    await exportToPng(node, sanitisedFileName);
+    setIsExporting(false);
+
+    // Allow some time for the chart to resize
+    await sleep(1000);
```
edmofro, line 37: Would it be worth creating a new component just for the export process to look at, so then we could immediately remove it after export, rather than reusing the main component that then has to reanimate back to its non-exporting state?

edmofro (changes requested):
Great (and fast) work @tcaiger!

I like this new split button component, seems like we could use that a few places so thanks for putting it in `ui-components` and adding the storybook

Just a couple of minor changes suggested, take them or leave them, but also a question to consider about creating a sandboxed export component - let me know what you think

packages/utils/src/string.js
tcaiger: Great suggestion!

packages/lesmis/src/components/DashboardReportModal.js
```diff
@@ -111,26 +149,40 @@ export const DashboardReportModal = () => {
           handleClose={handleClose}
           title={config?.dashboardName ? config?.dashboardName : 'Loading...'}
         />
-        <Wrapper>
+        <ExportLoader $isExporting={isExportLoading}>
```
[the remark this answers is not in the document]
tcaiger: It's new syntax in [styled-components](https://styled-components.com/docs/api#transient-props) for passing props to components when you don't want them to propagate to the dom. Maybe I'll add a 2 minute dev chat item to show this?!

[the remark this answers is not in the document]
edmofro: Sounds good!

packages/ui-components/src/components/SplitButton.js
```diff
@@ -0,0 +1,118 @@
...
+  &.MuiButtonBase-root {
+    margin: 0;
+  }
+`;
+
+const ButtonGroup = styled(MuiButtonGroup)`
+  margin-right: 1rem;
+`;
+
+export const SplitButton = ({ options, selectedId, setSelectedId, onClick, ButtonComponent }) => {
```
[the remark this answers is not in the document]
tcaiger: hmm I'm not sure about DropOptionButton. It makes me think that we will be removing options. Although not super widely used, I think SplitButton a somewhat universally used name for this kind of thing. eg. [material-ui ](https://material-ui.com/components/button-group/#split-button) or [w3schools](https://www.w3schools.com/howto/howto_css_button_split.asp)

packages/ui-components/src/components/SplitButton.js
tcaiger: Copied comment to my commit

packages/ui-components/src/components/SplitButton.js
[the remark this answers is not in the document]
edmofro: True, very convinced by those links!

packages/lesmis/src/utils/useExportToPNG.js
tcaiger: I just had a bit of a play refactoring it and it wasn't working out too well so I've abandoned it. It creates a bit of duplicated and/or messy code and the user experience is no better off. I also think the 1 second delay with the loader is quite a nice interaction so I'm pretty keen to keep it as is.

edmofro (approved):

PR #2996, opened 2021-08-12
MEL-390: Admin panel e2e tests scaffolding

kael89:
### Issue #:
MEL-390

Added a simple test case (`example.spec.js`) as a proof of concept - we should delete this file as soon as we start adding the actual test cases


BinduMadhaviTammireddy (approved):
Thank you

PR #2997, opened 2021-08-12
MEL-387: Allow editting existing viz

chris-bes:
### Changes:

- Edit existing viz.
- A few more clean up

### Work Flow:
- Choose an existing dashboard item, if it is not legacy, a button `Edit using Viz-Builder` should appear.
image

- If clicked, it will take you to the viz-builder page with all of the config already populated.


packages/admin-panel-server/src/routes/SaveDashboardVisualisationRoute.ts
```diff
@@ -50,12 +52,12 @@ export class SaveDashboardVisualisationRoute extends Route {
     let result;
 
     // Update visualisation if id exists
-    if (visualisation.id) {
-      result = await this.meditrakConnection.updateResource(`dashboardVisualisations/${visualisation.id}`, {}, body);
+    if (dashboardVisualisationId) {
```
chris-bes, line 55: I've changed the save route (save an existing dashboard viz) to `/PUT dashboardVisualisation/:dashboardVisualisationId` to make it a bit more resty. However, Create/Save dashboard viz still share the same RouteHandler

packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationCombiner.ts
```diff
@@ -0,0 +1,65 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { DashboardItem, Report } from '../types';
+
+export class DashboardVisualisationCombiner {
```
chris-bes, line 8: I've separated the logic that combines dashboardItem + report => visualisation to this separate class so that we don't have to cramp all the logic in the `DashboardVisualisationExtractor` class

packages/admin-panel/src/VizBuilderApp/api/mutations/useSaveVisualisation.js
```diff
@@ -4,20 +4,22 @@
...
-// Todo: Finish Save Button @see https://linear.app/tupaia/issue/MEL-20/save-button
 export const useSaveVisualisation = config =>
   useMutation(
-    ['saveDashboardVisualisation', config],
-    () =>
-      post('saveDashboardVisualisation', {
+    ['dashboardVisualisation', config],
+    () => {
+      if (config.id) {
+        return put(`dashboardVisualisation/${config.id}`, {
```
chris-bes, line 14: I've changed the save route (save an existing dashboard viz) to /PUT dashboardVisualisation/:dashboardVisualisationId to make it a bit more resty

packages/admin-panel/src/VizBuilderApp/views/Main.js
```diff
@@ -21,9 +26,50 @@ const RightCol = styled(FlexColumn)`
...
+      <StyledAlert severity="error" variant="standard">
+        {error.message}
+      </StyledAlert>
+    );
+  }
+
+  // Wait until visualisation is loaded to populated the field correctly if we are viewing an existing viz
+  if (visualisationId && !visualisationLoaded) {
+    return <Progress size={100} />;
+  }
```
chris-bes, line 71: There's a bit of a problem with JSON Edit box that it doesn't automatically re-render the json value inside the box if the value is changed. So I thought it'd be better to wait until the viz is done fetched and then render the whole component. Also if there's any error, don't render anything

packages/admin-panel/src/editor/Editor.js
```diff
@@ -33,7 +33,22 @@ export const Editor = ({ fields, recordData, onEditField }) => {
...
+          if (!show) {
+            return false;
+          }
+
+          // show or hide a field based on another field's value
+          if (visibilityCriteria) {
+            return Object.entries(visibilityCriteria).every(
+              ([key, value]) => recordData[key] === value,
+            );
+          }
```
chris-bes, line 48: This is to allow showing or hiding a field based on another field's value.

packages/admin-panel/src/widgets/InputField.js
```diff
@@ -196,6 +206,24 @@ export const InputField = ({
...
       );
       break;
+    case 'link': {
+      const { path, parameters = {} } = linkOptions;
+      let link = path;
+      Object.entries(parameters).forEach(([paramKey, recordProperty]) => {
+        const linkParam = recordData[recordProperty];
+        if (linkParam) {
+          link = link.replace(`:${paramKey}`, linkParam);
+        }
```
chris-bes, line 216: This is to construct the Route path `viz-builder/:visualisationId`, so that we can pass the id of exisiting viz to the viz-builder page

packages/admin-panel-server/src/routes/FetchReportPreviewDataRoute.ts
```diff
@@ -28,16 +28,25 @@ export class FetchReportPreviewDataRoute extends Route {
...
+      throw new Error('Requires preview config to fetch preview data');
+    }
+
+    if (!hierarchy) {
+      throw new Error('Requires hierarchy to fetch preview data');
+    }
+
+    if (!entityCode) {
+      throw new Error('Requires entity to fetch preview data');
     }
```
tcaiger, line 42: Great helpful error messages!

packages/admin-panel-server/src/routes/SaveDashboardVisualisationRoute.ts
tcaiger: Nice one. Resty is good.

packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationCombiner.ts
tcaiger: Nice and clean

packages/admin-panel/src/VizBuilderApp/App.js
```diff
@@ -37,12 +37,8 @@ export const App = ({ Navbar, Footer }) => {
...
         {Navbar && <Navbar user={user} isBESAdmin={isBESAdmin} />}
         <Switch>
-          <Route path="/viz-builder/new" exact>
-            <CreateNew />
-          </Route>
-          <Route path="/">
-            <Main />
-          </Route>
+          <Route path="/viz-builder/new" exact component={CreateNew} />
+          <Route path="/viz-builder/:visualisationId?" component={Main} />
```
tcaiger, line 41: Any reason to add the components using the component prop? It's only a small thing but [the convention in react-router ](https://reactrouter.com/web/api/Route/route-render-methods) is to pass them as children.

packages/admin-panel/src/VizBuilderApp/views/Main.js
```diff
@@ -21,9 +26,50 @@ const RightCol = styled(FlexColumn)`
   flex: 1;
 `;
 
+const StyledAlert = styled(SmallAlert)`
+  margin-top: 30px;
+  margin: auto;
```
tcaiger, line 31: This will result in the margin-top: 30px getting ignored. Is that intentional?

packages/admin-panel/src/VizBuilderApp/views/Main.js
```diff
@@ -21,9 +26,50 @@ const RightCol = styled(FlexColumn)`
...
+`;
+
+const Progress = styled(CircularProgress)`
+  margin-top: 30px;
+  margin: auto;
```
tcaiger, line 36: Same here? Do you mean: `margin: 30px auto auto;`?

packages/admin-panel/src/VizBuilderApp/views/Main.js
tcaiger: That seems a bit strange since it is a controlled component and the state is kept externally the whole time. Hopefully it doesn't create other issues not updating. Anyway this looks like a good workaround and probably removes a whole lot of edge cases rendering things before the data is fetched 👍

packages/meditrak-server/src/routes/index.js
```diff
@@ -167,6 +168,7 @@ export default {
   getDashboards: useRouteHandler(GETDashboards),
   getDashboardItems: useRouteHandler(GETDashboardItems),
   getDashboardRelations: useRouteHandler(GETDashboardRelations),
+  getDashboardVisualisations: useRouteHandler(GETDashboardVisualisation),
```
tcaiger, line 171: It's minor but any reason for the difference in plurality? Shouldn't GETDashboardVisualisation have an "s"?

tcaiger: And the filename would also be plural right?

tcaiger (changes requested):
Looks great. Just a couple of tiny questions about the route syntax and pluralisation of the new meditrak endpoint.

packages/admin-panel/src/VizBuilderApp/App.js
chris-bes: Ah the main reason is only by this way, the `Main` component can access the `:visualisationId` parameter from the Route path (I tried some other ways but was not successful)

tcaiger: You should be able to just use [the react-router `useParams`](https://reactrouter.com/web/api/Hooks/useparams) hook to get the param in the `Main` component. 
eg. ` const { visualisationId } = useParams();` 

Then you wont need to add the components as named props any more on the `Route` above. Let me know if that doesn't work.

packages/meditrak-server/src/routes/index.js
chris-bes: Ah yes, this was actually intentional because I don't think we will ever get multiple dashboard visualisations. But I understand the concern of inconsistency. Have changed it to plural

packages/admin-panel/src/VizBuilderApp/views/Main.js
chris-bes: Sorry, it was a mistake, the 30px is redundant

packages/admin-panel/src/VizBuilderApp/App.js
chris-bes: Thanks for that suggestion, it works! That's much better

chris-bes:
Have addressed your reviews @tcaiger !

tcaiger (approved):
Looking good!

PR #2998, opened 2021-08-12
Add report using strange report server method

Alastair-L:
### Issue #: SOL-297

This uses a new method of filling empty values that isn't `FILL_EMPTY`. The reason for this is 50% because I wanted to explore the report server functionality, and 50% because it avoids having to make a new indicator for what conceptually makes more sense as being done by the report server.

I argue that it conceptually makes more sense to fill AFTER fetching data as the filled data is not "real" data, it is only to assist with generating this one report.

If the reviewed agrees with this method for the general case, we should make some kind of alias or simpler way of doing this in the future.


packages/database/src/migrations/20210811034325-AddCOVIDFijiDashboardTestsPerCapita-modifies-data.js
```diff
@@ -0,0 +1,201 @@
...
+        '...': ['period'],
+      },
+      {
+        transform: 'sort',
+        by: '$row.period',
+      },
+      {
+        transform: 'select',
+        "'Population'":
+          'exists($allPrevious.Population) ? subset($allPrevious.Population, index(size($allPrevious.Population))) : -1',
```
rohan-bes, line 78: Hahaha somehow I knew that you'd end up finding some crazy things we can do with the report-server 😝

I do like your thinking and you've found a surprisingly clean way of achieving it here! 

I wouldn't mind having a chat with you about the relative pros and cons of using an indicator/aggregation to achieve this. As I see it:
- The data itself isn't actually 'fake', rather we have no easy way of correlating it correctly. It's true that the population on each of these periods is the value of the most recent population survey result for that period. This seems to be a combined issue of a) not being able to specify a date range that a survey response is applicable for, b) not being able to disaggregate data in our system (ie convert data from a higher level to a lower level)
- This is quite a general case issue in our system, we run into this problem fairly often. While in this case it the calculation can be performed in the report-server, there may be cases where it must be done in an indicator (ie. nested indicators). While that doesn't necessarily mean we shouldn't use the report-server here, we should keep in mind that we can't always use it to solve this issue.
- For what it's worth, it actually sort of makes sense to me for there to be a `Tests per 100k` indicator, as that is a real data point. I feel this is where indicators excel, both conceptually and technically, since they are all about producing a single data point from multiple other data points. The report-server can also do this, but that's more cos the report-server is nebulous enough that it can do most things (tho none of them in a particularly optimal fashion).

I listed a few points there as counter points to doing this via report-server, but I'm more on the fence about it than I made it seem just then. Would be keen to chat on it?

packages/database/src/migrations/20210811034325-AddCOVIDFijiDashboardTestsPerCapita-modifies-data.js
```diff
@@ -0,0 +1,201 @@
...
+        '...': ['period', 'Tests'],
+      },
+      {
+        transform: 'filter',
+        where: 'exists($row.Tests) and $row.Population != -1',
+      },
+      {
+        transform: 'select',
+        "'Tests per 100k'": '$row.Tests / ($row.Population * 100000)',
```
rohan-bes, line 87:
```suggestion
        "'Tests per 100k'": '$row.Tests / ($row.Population / 100000)',
```
I _think_, right?

packages/database/src/migrations/20210811034325-AddCOVIDFijiDashboardTestsPerCapita-modifies-data.js
```diff
@@ -0,0 +1,201 @@
...
+        "'name'": "periodToDisplayString($row.period, 'DAY')",
+        "'timestamp'": 'periodToTimestamp($row.period)',
+        '...': [],
```
rohan-bes, line 90:
```suggestion
```
Shouldn't be necessary I believe 👍

rohan-bes (changes requested):
Couple of tiny changes requested, but more keen to talk through the options for how we may support this?

packages/database/src/migrations/20210811034325-AddCOVIDFijiDashboardTestsPerCapita-modifies-data.js
rohan-bes: ^Unignore me??? I truly cannot fathom this maths problem... I need an adult 😢

packages/database/src/migrations/20210811034325-AddCOVIDFijiDashboardTestsPerCapita-modifies-data.js
Alastair-L: >somehow I knew that you'd end up finding some crazy things we can do with the report-server

I tried my very best 😂

> The data itself isn't actually 'fake', rather we have no easy way of correlating it correctly. It's true that the population on each of these periods is the value of the most recent population survey result for that period.

(assume we have a population survey response for 01/01/21)
Although that is true, I guess we have 'true' data for 01/01/21, but the population for 02/01/21 is 'fake' as it is just an estimate. The population on 02/01/21 isn't really the result for that response, rather we use the **best estimate** for the population on 02/01/21, which **just happens** to just be the most recent response. We could equally use some kind of interpolation (and possibly will in the future). 

In my mental model (which might be mistaken!), I think of `fill_empty` similarly to `interpolate_empty`, which means either they both should be a aggregations or both should be calculations

> b) not being able to disaggregate data in our system (ie convert data from a higher level to a lower level)

Not too sure how this one comes into it? Do you mean like submitting data for a year which gets fetched for individual days? In which can the population for a place doesn't remain the same for a year, so it still (in my view) makes more sense to put the population in for the date of the census and allow each report to decide what to do with that data?

> While in this case it the calculation can be performed in the report-server, there may be cases where it must be done in an indicator (ie. nested indicators)

This is an excellent point!!! I hadn't thought of that... (and it might be reason enough given that it's generally best practice to do the same thing the same way every time)

> It actually sort of makes sense to me for there to be a `Tests per 100k` indicator

Oooo, yeah that makes a lot of sense... I think you might be right... (but probably not worth the effort to change at this stage)

>  I'm more on the fence about it than I made it seem just then

Likewise, I'm not sure at all. If I'm honest, unlike what I mentioned in the description it's more like 80% playing with the reports server (which I LOVE and ADORE), and only 20% because I thought this method would be better 😂.

> Would be keen to chat on it?

I'd love to! I'll slack you :))

packages/database/src/migrations/20210811034325-AddCOVIDFijiDashboardTestsPerCapita-modifies-data.js
Alastair-L: eeeek! I'm not sure either!! I'll give you a proof one way or the other 😂

packages/database/src/migrations/20210811034325-AddCOVIDFijiDashboardTestsPerCapita-modifies-data.js
Alastair-L: Amazing. Could it be possible to love the reports server more

rohan-bes (approved):
As per our discussion I just want to summarise that it seems:
- We both agree there isn't a great way of handling this issue generally in Tupaia
- Arguably a `Tests per 100k` indicator might be a bit cleaner than this, but only marginally and doesn't seem to justify the effort
- There are multiple avenues we could take to deal with this in the future, but generally we'd need to introduce some ability to correlated data across period ranges
- As this stands, it's perfectly fine

Pre-approving, though I do think that calc should be divide by 100k!

Alastair-L:
Closing as out of date

PR #3003, opened 2021-08-13
Bugfix: Historical data in list visual

EMcQ-BES:
### Issue #:
NZ-62

### Changes:

- Change reports and indicators to use `FINAL_EACH_YEAR` instead of `MOST_RECENT` so they actually show historical data
- Compare target and actual values at the same scale


tcaiger (approved):
LGTM

PR #3004, opened 2021-08-13
MEL-397: Address viz-builder prototype feedback

chris-bes:
### Changes:

- Address viz-builder prototype feedback

### DIFF
--- a/packages/admin-panel/src/VizBuilderApp/api/queries/useReportPreview.js
+++ b/packages/admin-panel/src/VizBuilderApp/api/queries/useReportPreview.js
@@ -4,7 +4,6 @@
  */
 import { useQuery } from 'react-query';
 import { post } from '../api';
-import { DEFAULT_REACT_QUERY_OPTIONS } from '../constants';
 
 export const useReportPreview = (visualisation, project, location, enabled, setEnabled) =>
   useQuery(
@@ -21,7 +20,8 @@ export const useReportPreview = (visualisation, project, location, enabled, setE
       return response.results;
     },
     {
-      ...DEFAULT_REACT_QUERY_OPTIONS,
+      // do not use query cache here because report data 
+      // should be refetched frequently according to config changes
       enabled,
       keepPreviousData: true,
       onSettled: () => {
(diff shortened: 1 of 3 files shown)


packages/admin-panel/src/VizBuilderApp/api/queries/useReportPreview.js
```diff
@@ -21,7 +20,8 @@ export const useReportPreview = (visualisation, project, location, enabled, setE
       return response.results;
     },
     {
-      ...DEFAULT_REACT_QUERY_OPTIONS,
+      // do not use query cache here because report data 
+      // should be refetched frequently according to config changes
```
chris-bes, line 24: Disable query caching here because I think data should be up to date

packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
```diff
@@ -114,7 +126,11 @@ export const PreviewSection = ({ enabled, setEnabled }) => {
   const columns = useMemo(() => getColumns(reportData), [reportData]);
   const data = useMemo(() => reportData, [reportData]);
 
-  const viewContent = { data, ...visualisation.presentation };
+  // only update Chart Preview when play button is clicked
+  useEffect(() => {
+    const newViewContent = { data, ...visualisation.presentation };
+    setViewContent(newViewContent);
```
tcaiger, line 132: Like Rohan mentioned, there are pros and cons to this but it's well implemented either way.

tcaiger (approved):
Looks good

packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
rohan-bes: I am keen on this change overall, but it does really seem to suggest that the 'play' button should somewhere other than inside the report config section... perhaps in the preview header area?

rohan-bes: (Not suggesting we change that now btw, let's get Dave to take a look at it at some point 👍)

tcaiger: Yep let the design master figure it out :)

PR #3006, opened 2021-08-13
Update deprecated @babel/polyfill in web-frontend

billli0:
### Issue #:
https://linear.app/tupaia/issue/SOL-291/js-error-object-doesnt-support-property-or-method-fromentries
### Changes:
Update deprecated @babel/polyfill. [Reference link](https://babeljs.io/docs/en/babel-polyfill). Updating will support more es syntax using babel core-js

#### Thought
Currently no need to update @babel/polyfill for other packages; mocha test still require babel/polyfill, maybe it can work without it. 
Very keen to update all babel/polyfill since it will support more es syntax, but it will be safe to pair with a senior dev, or I will do it bit by bit.


rohan-bes (approved):
Okay, I'll admit that this isn't exactly my area of expertise, so this may not be my most thorough review. Looks clean and safe tho, so it should be all good 👍 

I think we wanna make sure the testers test this one on IE is all 👍

billli0:
@rohan-bes Good point👍 I will ask tester to do an e2e test and see if we can do it on IE.

PR #3007, opened 2021-08-15
NZ-828: Split large exports into multiple files

edmofro:
### Issue #:
NZ-828
NZ-24
NZ-859

Towards NZ-803


packages/admin-panel/src/importExport/ExportButton.js
```diff
@@ -19,18 +19,13 @@ const buildExportQueryParameters = (rowIdQueryParameter, rowData, filterQueryPar
...
-};
-
 export const ExportButton = ({ actionConfig, row }) => (
   <IconButton
     onClick={async () => {
       const { exportEndpoint, rowIdQueryParameter, extraQueryParameters, fileName } = actionConfig;
       const queryParameters = buildExportQueryParameters(rowIdQueryParameter, row);
       const endpoint = `export/${exportEndpoint}${!queryParameters && row.id ? `/${row.id}` : ''}`;
-      const processedFileName = processFileName(fileName, row);
+      const processedFileName = makeSubstitutionsInString(fileName, row);
```
edmofro, line 28: Removed the file type suffix, the browser will apply this automatically on saving, and we now have multiple file types supported (zip, xlsx)

packages/meditrak-server/package.json
```diff
@@ -59,7 +60,6 @@
     "s3urls": "^1.5.2",
     "semver-compare": "^1.0.0",
     "string-similarity": "^1.2.2",
-    "sussol-utilities": "^0.4.6",
```
edmofro, line 62: Removed reliance on this funky library I put together back in my mSupply days

packages/meditrak-server/src/apiV2/export/zipMultipleFiles.js
```diff
@@ -3,10 +3,14 @@
...
 import AdmZip from 'adm-zip';
 
 export function zipMultipleFiles(exportPath, files) {
+  const filePath = `${exportPath}/tupaia_export_${Date.now()}.zip`;
   const zip = new AdmZip();
-  files.forEach(zip.addLocalFile);
-  zip.writeZip(`${exportPath}/tupaia_export_${Date.now()}.zip`);
+  files.forEach(file => zip.addLocalFile(file));
+  zip.writeZip(filePath);
+  files.forEach(file => fs.unlinkSync(file));
```
edmofro, line 14: Deletes temp files once they're zipped

packages/utils/src/respond.js
```diff
@@ -23,14 +23,14 @@ export function respond(res, responseBody, statusCode) {
...
-export function respondWithDownload(res, filePath, deleteAfterDownload = false) {
+export function respondWithDownload(res, filePath, deleteAfterDownload = true) {
   const { overrideRespond } = res;
   if (overrideRespond) {
     return overrideRespond({ filePath });
   }
   return res.download(filePath, () => {
     if (deleteAfterDownload) {
-      fs.unlink(filePath); // delete file from disk after download
+      fs.unlinkSync(filePath); // delete file from disk after download
```
edmofro, line 33: This wasn't actually working before, as `unlink` requires a callback

packages/utils/src/respond.js
```diff
@@ -23,14 +23,14 @@ export function respond(res, responseBody, statusCode) {
     .send(JSON.stringify(responseBody));
 }
 
-export function respondWithDownload(res, filePath, deleteAfterDownload = false) {
+export function respondWithDownload(res, filePath, deleteAfterDownload = true) {
```
edmofro, line 26: Swapped the default so that temp files get automatically deleted after download

packages/utils/src/string.js
```diff
@@ -32,3 +32,15 @@ export const getArticle = word => {
...
 };
+
+/**
+ * Returns the string chopped off at maxLength, with ... replacing the last three characters if it
+ * overflowed
+ * @param  {string}   string     The string to be truncated
+ * @param  {integer}  maxLength  The maximum number of characters in the returned string
+ * @return {string}              The truncated string, with ellipses if it overflowed
+ */
+export const truncateString = (string, maxLength) => {
```
edmofro, line 43: Added this straight out of `sussol-utilities` so that I could remove that dependency

packages/web-config-server/src/export/ExportSurveyDataHandler.js
```diff
@@ -84,7 +83,7 @@ export class ExportSurveyDataHandler extends RouteHandler {
         origin: -1, // Append to bottom of worksheet starting on first column
       });
 
-      const sheetName = truncateString(surveyName, 31);
+      const sheetName = surveyName.substring(0, 31); // stay within excel sheet name limit
```
edmofro, line 86: To fix NZ-24

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
```diff
@@ -0,0 +1,303 @@
+/**
```
edmofro, line 1: Most of this comes from `exportSurveyResponses.js`, but with things moved around to be 
- more efficient (large exports are significantly faster, anecdotally)
- broken up into more discrete chunks
- able to be called from a place that doesn't extract the parameters from a url query (not actually required, but this is where I found the theoretical split in responsibilities between this file and the original)
- able to save multiple xlsx files and then zip them at the end, removing overhead from memory as it goes along (the main motivation!)

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
```diff
@@ -0,0 +1,303 @@
...
+export const INFO_ROW_HEADERS = ['Entity Code', 'Entity Name', 'Date'];
+
+const getBlankWorkbook = () => ({ SheetNames: [], Sheets: {} });
+
+/**
+ * Exports excel documents containing the relevant survey responses, splitting into multiple files
+ * and zipping if too large.
+ *
+ * Note that this is a large function that is broken up into many internal helper functions, a bit
+ * like a class. The main body of the function is right at the bottom.
```
edmofro, line 47: I originally went for a class, but given that it is fundamentally just a loop with a short lifespan, and that it didn't really make sense to have some aspects passed as arguments to the constructor and others as parameters to the main entrypoint, I changed it to a function and found the mental model sat better. I'm happy for this to be vetoed though, if the layout is too unexpected.

packages/utils/src/respond.js
kael89: Good call!

kael89 (changes requested):
Looks great overall @edmofro ! Thanks for doing this cleanup, it makes this section much more readable + maintainable. Just a few minor changes requested (most are optional but should be easy to change)

About using a function including other functions: I generally prefer the syntactic sugar of a class
* boundaries between method are defined a bit more clearly
* `this` can help attach mental scope to functions/fields etc...

But it may be a personal preference and/or a matter of habit. I am certainly fine with this staying as is, also I haven't seen this as a class, if I did I may have drawn the same conclusion as you did 😃

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
```diff
@@ -0,0 +1,303 @@
...
+  for (const survey of surveysWithAccess) {
+    // Get the current set of questions, in the order they appear in the survey
+    const questions = await findQuestionsInSurvey(models, survey.id);
+
+    if (surveyResponse) {
+      const entity = await models.entity.findById(surveyResponse.entity_id);
+      await exportResponses([surveyResponse], { [entity.id]: entity }, survey, questions);
+    } else {
+      const surveyResponses = await findResponsesForSurvey(survey);
+      const entitiesById = keyBy(entities, 'id');
```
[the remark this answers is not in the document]
edmofro: Excellent call!

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
```diff
@@ -0,0 +1,303 @@
...
+      if (surveyResponses.length > MAX_RESPONSES_PER_FILE) {
+        for (const batchOfResponses of chunk(surveyResponses, MAX_RESPONSES_PER_FILE)) {
+          await exportResponses(batchOfResponses, entitiesById, survey, questions);
+          saveCurrentWorkbook();
+        }
+      } else {
+        await exportResponses(surveyResponses, entitiesById, survey, questions);
+      }
```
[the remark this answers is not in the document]
edmofro: I prefer to keep it here, because otherwise you have the saving of the workbook happening inside that function as well as in this outer one (in the case where no batching is done), which feels a little like mixing responsibilities/confusing to me

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
```diff
@@ -0,0 +1,303 @@
...
+      { columns: [{ [`${TYPES.SURVEY_RESPONSE}.id`]: 'survey_response.id' }], sort: [] },
+    );
+
+    // Add any questions that are in survey responses but no longer in the survey
+    const allQuestionIds = getUniqueEntries(answers.map(a => a['question.id']).flat());
+    const validQuestionIds = questions.map(q => q.id);
+    const outdatedQuestionIds = allQuestionIds.filter(id => !validQuestionIds.includes(id));
+    const outdatedQuestions = await Promise.all(
+      outdatedQuestionIds.map(async questionId => models.question.findById(questionId)),
+    );
```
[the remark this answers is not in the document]
edmofro: Oh man, good catch!

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
```diff
@@ -0,0 +1,303 @@
...
+      const questionInfo = infoColumnKeys.map(columnKey => question[columnKey]);
+      const exportRow = preQuestionRowCount + questionIndex;
+      exportData[exportRow] = questionInfo;
+    });
+
+    // Add the answers on the right columns to the exportData
+    const questionIdToIndex = questionsForExport.reduce(
+      (dict, question, index) => ({ ...dict, [question.id]: index }),
+      {},
+    );
```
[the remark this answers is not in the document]
edmofro: I actually think this is easier to read too, will go with it. I've seen a fair bit of the reduce around lately and generally haven't pushed back, so I think we can expect both to become familiar over time :-)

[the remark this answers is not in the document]
kael89: Cool, just bear in mind the performance aspect too (not here, in general...) We've been burned by `reduce` (spreading `dict` in each pass) in the past 😉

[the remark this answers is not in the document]
edmofro: Yes true!

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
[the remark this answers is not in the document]
kael89: Good point! In that case, could we rename `exportResponses` to something else (eg `addResponsesToSheet`)?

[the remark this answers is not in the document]
edmofro: Oh yeah that's better!

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
```diff
@@ -0,0 +1,306 @@
...
+    addDataToSheet(survey.name, exportData);
+  };
+
+  /** Main body of the function below this point, everything above is helper functions */
+
+  const surveysWithAccessStatus = await Promise.all(
+    surveys.map(async s => ({ ...s, accessible: await checkAccessToSurvey(s) })),
+  );
+  const surveysByAccessStatus = groupBy(surveysWithAccessStatus, 'accessible');
+  surveysByAccessStatus[false]?.forEach(survey => {
```
kael89, line 277:
```suggestion
  (surveysByAccessStatus[false] || []).forEach(survey => {
```

Although it would be event better to extract those:
```js
  const { true: surveysWithAccess = [], false: surveysWithoutAccess = [] } = surveysByAccessStatus;
  // etc
```

kael89 (commented):
Just one more change (required, since it may break things, but pre-approving to make things easier) 🙌

packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
edmofro: The `?` deals with it: "only run the thing after the `.` if there's something defined before the `?`"

I do think your syntax is a good suggestion though, will make that change

kael89: Oh I see - so it's not like we'll try to run an `undefined` function. That's convenient!

PR #3009, opened 2021-08-15
Move where clause to beginning of select

EMcQ-BES:
### Issue:
https://beyondessential.slack.com/archives/C02649511JM/p1628804173010000

### Changes:

- Exit select early if we're not returning a new row


rohan-bes (approved):
Nice one!

rohan-bes:
Actually, it might be worth adding a unit test for this corner case?

rohan-bes (approved):
Awesome, thanks @EMcQ-BES !

PR #3013, opened 2021-08-16
Sol-37 connect gaps in data lines

billli0:
### Issue #:
https://linear.app/tupaia/issue/SOL-37/connecting-gaps-in-data-lines-and-change-x-axis-to-be-yearly

### Changes:
- add connectNulls in line chart to connect gaps

- add insertOrUpdateRootJsonEntry() in migration, because JSONB_SET can not modify root json entry.

- migration for config changes

### Screenshot 
See comment in card


packages/database/src/utilities/migration.js
```diff
@@ -102,18 +102,39 @@ export async function updateValues(db, table, newValues, condition) {
   return db.runSql(`UPDATE "${table}" SET ${setQuery} WHERE ${whereQuery}`, params);
 }
 
+const getValueString = values => {
```
edmofro, line 105:
```suggestion
const getJsonColumnValueString = values => {
```

packages/database/src/utilities/migration.js
```diff
@@ -102,18 +102,39 @@ export async function updateValues(db, table, newValues, condition) {
...
+      .map(item => (typeof item === 'string' ? `"${item}"` : item))
+      .join(',');
+    return `[${arrayString}]`;
+  }
+
+  if (typeof values === 'object') {
+    return JSON.stringify(values);
+  }
+
+  throw new Error('Function insertJsonEntry: we should either insert object or array');
```
edmofro, line 117:
```suggestion
  throw new Error('Function getJsonColumnValueString: we should either insert object or array');
```

edmofro (approved):
Nice one @billli0! Just one rename suggestion, which is up to you :-)

packages/database/src/utilities/migration.js
billli0: Love it!

edmofro (approved):

PR #3015, opened 2021-08-16
SOL-377: Switched 'sum' in report-server to use multi arguments instead of array

rohan-bes:
### Issue SOL-377:

Switching `sum` over to taking multiple arguments like `divide` does. I think this format in general should be preferred over taking an array as an argument.

### DIFF
--- a/packages/report-server/src/__tests__/reportBuilder/functions.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/functions.test.ts
@@ -65,8 +65,7 @@ describe('functions', () => {
     });
 
     describe('length', () => {
-      it('returns the length of an input array', () =>
-         expect(functions.length([1, 2, 3])).toBe(3));
+      it('returns the length of an input array', () => expect(functions.length([1, 2, 3])).toBe(3));
     });
   });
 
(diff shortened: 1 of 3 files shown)


packages/report-server/src/__tests__/reportBuilder/functions.test.ts
```diff
@@ -103,25 +102,24 @@ describe('functions', () => {
...
-    //     expect(functions.sum([1, undefined, 3], undefined, 2)).toBe(6));
+      it('can combine numbers and arrays', () =>
+        expect(functions.sum([1, undefined, 3], undefined, 2)).toBe(6));
 
-    //   it('returns undefined if all values are undefined', () =>
-    //     expect(functions.sum([undefined, undefined, undefined], undefined, undefined)).toBe(
-    //       undefined,
-    //     ));
-    // });
+      it('returns undefined if all values are undefined', () =>
```
rohan-bes, line 118: Should this be `0` maybe? I wasn't too sure... I feel either could be seen as appropriate...

packages/report-server/src/reportBuilder/functions/math.ts
```diff
@@ -7,12 +7,22 @@ import { sum as mathjsSum, divide as mathjsDivide, Matrix, filter } from 'mathjs
 
 const isDefined = <T>(value: T): value is Exclude<T, undefined> => value !== undefined;
 
-export const sum = (numbers?: Matrix): number | undefined => {
-  if (numbers === undefined) {
+export const sum = (
+  ...args: (number | undefined | (number | undefined)[])[]
+): number | undefined => {
+  const numbers = args.flat();
```
EMcQ-BES, line 13: I was thinking about this recently and part of the issue this causes in older reports is that if a report uses `sum([1, 2, 3])` format, that `[1, 2, 3]` is being passed as a `DenseMatrix` _not_ an array. So there's an additional case to deal with here.

rohan-bes: That's true! And I think this was the case when I first tried to make this refactor and broke a bunch of lesmis vizes 😢 

I added a migration to this work to re-write those reports to use the new syntax! They seemed to work correctly when testing in local 😄 

However, you raise a good point that we may want to support dense matrix regardless? I've gone back and forth on this... At first I tried to support it, but then began to feel that either:
a) A person is directly writing `sum([a, b, c, ...])` in which case, they shouldn't and should write `sum(a, b, c, ...)` instead
b) We are calling `sum` on a result of a function that returns a dense matrix, eg. `sum(dotProduct([a,b,c], [d,e,f]))`. This case is a bit trickier to handle, but I feel like we'd prefer to ensure that the inner function returns an array rather than a dense matrix. The whole dense matrix thing is a carry over from `mathjs` which we kinda wanna avoid using if possible.

How does this sound to you? I can understand if we wanna support the dense matrix however it does make the code a bit uglier 😢

EMcQ-BES: I think that's a fair assessment. Can we add an error message at least? I recall part of the issue when we first tried this was that the sum running against a dense matrix returned undefined sometimes rather than just erroring out.

If we aren't supporting it probably better to have it error out early and descriptively rather than attempting to process it and hoping it will error out at some point right?

packages/report-server/src/__tests__/reportBuilder/functions.test.ts
EMcQ-BES: I think undefined is the correct answer

packages/report-server/src/reportBuilder/functions/math.ts
rohan-bes: Mmm yeah good point... I'll add some kind of validator 👍

packages/report-server/src/reportBuilder/functions/math.ts
```diff
@@ -3,21 +3,45 @@
...
+const isNumericOrArray = (
+  value: unknown,
+): value is number | undefined | Array<number | undefined> =>
+  isNumeric(value) || (Array.isArray(value) && value.every(isNumeric));
 const isDefined = <T>(value: T): value is Exclude<T, undefined> => value !== undefined;
 
-export const sum = (numbers?: Matrix): number | undefined => {
-  if (numbers === undefined) {
+export const sum = (
+  ...args: (number | undefined | (number | undefined)[])[]
```
rohan-bes, line 17: Arguably we should make the input type `unknown` at this point, since we're now doing validation on the input so there's no need to specify its type. However, I still find it rather useful to have the type definition here, both for readability and to ensure that whenever we call this function directly in code we still get compile time type checking 👍

EMcQ-BES (approved):
Looks good! 👍

PR #3018, opened 2021-08-17
SOL-9 Exported FETP survey file is missing numeric only options

billli0:
### Issue #:
https://linear.app/tupaia/issue/SOL-9/exported-fetp-survey-file-is-missing-numeric-only-options

### Investigation:
Originally we expect JSON.parse() will throw error when parsing an non-object string, but in fact it can successfully parse number string, boolean string, and null.    

### Changes:
- Add an util function for a stricter JSON.parse.


nick-beyondessential (approved):
LGTM!

PR #3019, opened 2021-08-17
Sol 92 pacmossi project card

PShoe:
### Changes:
- add pacmossi project card
- no dashboard items were added, only Vector Surveillance dashboard 

- Example

---


packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,105 @@
...
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const countryCodes = ['FJ', 'KI', 'NR', 'NU', 'PG', 'SB', 'TV', 'VU'];
```
nick-beyondessential, line 19: Should 'TO' be included here?

PShoe: @nick-beyondessential thank you, have added!

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+const projectDescription = 'Pacific Mosquito Surveillance Strengthening for Impact';
+
+const userGroups = ['PacMOSSI', 'PacMOSSI Senior'];
+
+const dashboardName = 'Vector Surveillance';
+const projectDashboardItemCode = 'project_details';
+const projectDashboardCode = `${projectCode}_${projectName}`;
+
+const getCountryDashboardCode = (countryCode, userGroup) =>
+  `${countryCode}_PacMOSSI_Country_${userGroup.split(' ').join('_')}`;
```
nick-beyondessential, line 32: We can now define user permissions `dashboard_relation` so would suggest removing user suffix. And `Country` was also relic of old system.  ``${countryCode}_${project}_${name}` seems a common pattern.
```suggestion
  `${countryCode}_PacMOSSI_${dashboardName.split(' ').join('_')}`;
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+
+const hierarchyNameToId = async (db, name) => {
+  const record = await db.runSql(`SELECT id FROM entity_hierarchy WHERE name = '${name}'`);
+  return record.rows[0] && record.rows[0].id;
+};
+
+const addCountryToProject = async (db, countryCode) => {
```
nick-beyondessential, line 39:
```suggestion
const addCountryToProject = async (db, countryCode, entityHierarchyId) => {
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+  await insertObject(db, 'entity_relation', {
+    id: generateId(),
+    parent_id: await codeToId(db, 'entity', projectCode),
+    child_id: await codeToId(db, 'entity', countryCode),
+    entity_hierarchy_id: await hierarchyNameToId(db, projectCode),
```
nick-beyondessential, line 44:
```suggestion
    entity_hierarchy_id: entityHierarchyId,
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+    userGroups.map(userGroup =>
+      insertObject(db, 'dashboard', {
+        id: generateId(),
+        code: getCountryDashboardCode(countryCode, userGroup),
+        name: dashboardName,
+        root_entity_code: countryCode,
+        sort_order: 1,
+      }),
+    ),
+  );
```
nick-beyondessential, line 56:
```suggestion
  await insertObject(db, 'dashboard', {
        id: generateId(),
        code: getCountryDashboardCode(countryCode),
        name: dashboardName,
        root_entity_code: countryCode,
        sort_order: 1,
      });
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+
+exports.up = async function (db) {
+  await insertObject(db, 'entity', {
+    id: generateId(),
+    code: projectCode,
+    parent_id: await codeToId(db, 'entity', 'World'),
+    name: projectName,
+    type: 'project',
+  });
+
```
nick-beyondessential, line 84: either 
```suggestion
const entityHierarchyId = generateId();
 
```
or you could use the following fetch after insert:
```suggestion
const entityHierarchyId =await hierarchyNameToId(db, projectCode);
```
I'd probably go the former.

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+  await insertObject(db, 'entity_hierarchy', {
+    id: generateId(),
```
nick-beyondessential, line 86:
```suggestion
    id: entityHierarchyId,
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+    code: projectCode,
+    description: projectDescription,
+    sort_order: 16,
+    image_url: 'https://tupaia.s3-ap-southeast-2.amazonaws.com/uploads/PacMOSSI_Background.JPG',
+    default_measure: '126,171',
+    user_groups: `{${userGroups[0]}, ${userGroups[1]}}`,
+    logo_url: 'https://tupaia.s3-ap-southeast-2.amazonaws.com/uploads/PacMossi_Logo.jpg',
+    entity_id: await codeToId(db, 'entity', projectCode),
+    entity_hierarchy_id: await hierarchyNameToId(db, projectCode),
```
nick-beyondessential, line 118:
```suggestion
    entity_hierarchy_id: entityHierarchyId,
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+    name: projectName,
+    type: 'project',
+  });
+
+  await insertObject(db, 'entity_hierarchy', {
+    id: generateId(),
+    name: projectCode,
+  });
+
+  await Promise.all(countryCodes.map(countryCode => addCountryToProject(db, countryCode)));
```
nick-beyondessential, line 90:
```suggestion
  await Promise.all(countryCodes.map(countryCode => addCountryToProject(db, countryCode, entityHierarchyId)));
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+  });
+};
+
+exports.down = async function (db) {
+  const hierarchyId = await hierarchyNameToId(db, projectCode);
+
+  for (const countryCode of countryCodes) {
+    for (const userGroup of userGroups) {
+      await db.runSql(
+        `DELETE FROM "dashboard" WHERE code = '${getCountryDashboardCode(countryCode, userGroup)}'`,
```
nick-beyondessential, line 128:
```suggestion
        `DELETE FROM "dashboard" WHERE code = '${getCountryDashboardCode(countryCode)}'`,
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+    entity_id: await codeToId(db, 'entity', projectCode),
+    entity_hierarchy_id: await hierarchyNameToId(db, projectCode),
+  });
+};
+
+exports.down = async function (db) {
+  const hierarchyId = await hierarchyNameToId(db, projectCode);
+
+  for (const countryCode of countryCodes) {
+    for (const userGroup of userGroups) {
```
nick-beyondessential, line 126:
```suggestion
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,144 @@
...
+const projectName = 'PacMOSSI';
+const projectDescription = 'Pacific Mosquito Surveillance Strengthening for Impact';
+
+const userGroups = ['PacMOSSI', 'PacMOSSI Senior'];
+
+const dashboardName = 'Vector Surveillance';
+const projectDashboardItemCode = 'project_details';
+const projectDashboardCode = `${projectCode}_${projectName}`;
+
+const getCountryDashboardCode = (countryCode, userGroup) =>
```
nick-beyondessential, line 31:
```suggestion
const getCountryDashboardCode = (countryCode) =>
```

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
```diff
@@ -0,0 +1,141 @@
...
+    permissionGroup: `${permissionGroups[0]}`,
+    entityTypes: ['project'],
+    projectCodes: [`${projectCode}`],
+  });
+
+  await insertObject(db, 'project', {
+    id: generateId(),
+    code: projectCode,
+    description: projectDescription,
+    sort_order: 16,
```
billli0, line 114:
```suggestion
    sort_order: 15,
```

Got ya!

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
PShoe: Have changed.

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
PShoe: Have added entityHierarchyId

billli0 (approved):
Looks good overall! I will give a pre-approve. It would be nice to change on a minor thing.

billli0:
I run the code and it works. But it will be safe to have a double review from @nick-beyondessential .

nick-beyondessential (approved):
LGTM!

packages/database/src/changeHandlers/EntityHierarchyCacher.js
```diff
@@ -52,8 +52,10 @@ export class EntityHierarchyCacher {
     old_record: oldRecord,
     new_record: newRecord,
   }) => {
-    if (oldRecord.canonical_types === newRecord.canonical_types) {
-      return null; // if the canonical types are the same, the change won't invalidate the cache
+    if (oldRecord && newRecord) {
```
PShoe, line 55: This is @edmofro code, not sure if okay to touch, someone should definitely have a look:

was getting this bug:
```error: TypeError: Cannot read property 'canonical_types' of null```

File: packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js (hunk 13 above)
Comment (PShoe) [reply packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js]: Ah it must start at 0, thank you @billli0 !

File: packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js [hunk 15]
```diff
@@ -0,0 +1,146 @@
...
+    parent_id: await codeToId(db, 'entity', 'World'),
+    name: projectName,
+    type: 'project',
+  });
+
+  const entityHierarchyId = generateId();
+  await insertObject(db, 'entity_hierarchy', {
+    id: entityHierarchyId,
+    name: projectCode,
+    canonical_types: '{country,district,field_station,larval_habitat}',
```
PShoe, line 86: @nick-beyondessential - I just saw this in another project migration (20210324041508-AddSupplyChainFijiProject-modifies-data.js). Are we meant to put canoical_types in here?

nick-beyondessential: Yes I think this makes sense as field_station & larval_habitat are not default canonical types so wouldn't be picked up otherwise. We may need to assess the hierarchy structure of each country i.e do any have sub_districts?

billli0: I've checked it today from mock data and entities that DL has country -> district -> sub_district -> field_station hierarchy. Others are country -> district -> field_station.

packages/database/src/changeHandlers/EntityHierarchyCacher.js
nick-beyondessential:
```suggestion
    if (oldRecord?.canonical_types === newRecord?.canonical_types) {
```

edmofro: Good catch team, that's a bug from me! @nick-beyondessential nice syntax, recommend you use that @PShoe :-)

edmofro: Actually there is a slight logic difference in the syntax, and while the difference in behaviour is extremely edge casey, it makes me lean towards your original syntax for safety.

Logic difference explanation:
While the default for the column `canonical_types` is `{}`, there isn't an explicit `NOT NULL` constraint. That means you could get a situation where we're e.g. deleting a hierarchy, and for some reason its `canonical_types` had been set to `NULL`, in which which case you'd get:

`newRecord`: `null` because it is a delete, so `newRecord?.canonical_types` evaluates to `null`
`oldRecord`: `{ ..., canonical_types: null }`, so `oldRecord?.canonical_types` also evaluates to `null`

In which case the condition thinks that there is no change, and we don't invalidate the cache, which is incorrect. 

In fact, in this case we'd also be deleting the relation records etc. so the invalidation should happen as a byproduct of other actions, but just to avoid ever having to think about this again, @PShoe let's go with the way you have it!

PShoe: Thanks all! @nick-beyondessential / @edmofro

packages/database/src/migrations/20210816012839-AddProjectPacmossi-modifies-data.js
PShoe: @billli0 when you're up, want to run that one past me? I.e. do I need larval_habitat, and where you checked this? :)

billli0: @PShoe Sure, any time!

billli0: From Edwin's [comment ](https://beyondessential.slack.com/archives/CDGS6P56K/p1627953164025600) I think we are not going to use canonical type in the future projects, as we prefer to use entity relations.

edmofro: That's right @billli0! Are `field_station` and `larval_habitat` entities all imported through the admin panel though, or are some created dynamically by filling out a survey on meditrak app? If the latter is true, we would still need canonical types for them to get automatically included in the hierarchy

PShoe: Confirmed by @gerrybes, all field_station and larval_habitat entities will be created dynamically by filling out a survey on meditrak app, so will add to canonical_types as is current code.

PR #3020, opened 2021-08-17
NZ-904: Stop feature instances out of hours

edmofro:
### Issue #:
NZ-904

- Adds 2 lambda scripts, to stop and start all instances according to their "StopAtUTC" and "StartAtUTC" tags
- Modifies the createOrRestore lambda script to
  - Attach "StopAtUTC" and "StartAtUTC" tags at 6am and 7pm Melbourne time to any new feature instances
  - Allow running the restore process on a stopped feature instance, keeping it stopped the whole time
  - Break out some utilities that are utilised by the new scripts

#### Manual release steps
- [ ] Add tags to existing dev and feature instances so they get stopped overnight
- [ ] Update the `createOrRestoreDeployment` lambda function, including utilities.py (other two already created)


packages/devops/scripts/lambda/createOrRestoreDeployment.py
```diff
@@ -228,37 +210,32 @@ def lambda_handler(event, context):
 
     filters = [
         { 'Name': 'tag-key', 'Values': ['RestoreFrom'] },
-        { 'Name': 'instance-state-name', 'Values': ['running'] }
+        { 'Name': 'instance-state-name', 'Values': ['stopped'] }
```
IgorNadj, line 213: Does this mean you can only restore from stopped instances?

packages/devops/scripts/lambda/startTaggedInstances.py
```diff
@@ -0,0 +1,32 @@
...
+import time
+from utilities import *
+
+ec2 = boto3.resource('ec2')
+ec = boto3.client('ec2')
+iam = boto3.client('iam')
+route53 = boto3.client('route53')
+loop = asyncio.get_event_loop()
+
+async def wait_for_instance(instance_id, to_be):
```
IgorNadj, line 13: Import from utilities instead?

packages/devops/scripts/lambda/utilities.py
```diff
@@ -32,7 +36,6 @@ def tags_contains(tags, key, value):
     tag_matching_key = tags_matching_key[0]
     return tag_matching_key['Value'] == value
 
-
```
IgorNadj, line 35: Minor: do you have python lint? The two blank lines between top level functions comes from there https://www.python.org/dev/peps/pep-0008/#blank-lines
Pretty handy to have anyway, caught a bunch of issues for me while working on this last

IgorNadj (approved):
Looks great! Some minor comments

kael89:
Thanks for implementing this @edmofro !! 🎉 I haven't reviewed the code but just one feature-related comment: perhaps it would be worth to also stop instances on weekends by default?

edmofro:
Oh good point @kael89... I'll think about how this could be tweaked to do that, I didn't have it in mind so it's not immediately obvious but I'll come back to you

packages/devops/scripts/lambda/createOrRestoreDeployment.py
edmofro: That's right - if you want to restore an existing instance, you have to stop it first. I don't think people ever really do this in practice, so just kept the code clean and easy by restricting it to our expected overnight case - they'll be stopped.

packages/devops/scripts/lambda/startTaggedInstances.py
edmofro: Oops, good catch - that's the whole reason I moved that function to utilities 🤦

packages/devops/scripts/lambda/utilities.py
edmofro: Oh I don't! Will add it

edmofro: Actually it looks like I do, but it's not picking up the two lines thing. Do you know what config I might need to tweak?
image

IgorNadj: Hmm not sure, this is what I see using PyCharm CE

Screen Shot 2021-08-17 at 2 36 43 pm

edmofro: Huh...yeah I've just added the double line break manually.

IgorNadj (approved):
🏌️

edmofro:
@kael89 I worked out a nifty way of stopping the instances over the weekend - there's a cron job firing events at lambda to get it to check what needs stopping or starting every hour, but I can just make that not fire those check events on the weekend, so it won't start again until Monday. No code changes needed 🎉

PR #3021, opened 2021-08-17
SOL-8: Added support for background rebuilding of the analytics table

rohan-bes:
- Moved analytics table utility functions to the database
- Added ability to force rebuild the analytics table via build_analytics_table(force (default false))


packages/data-api/scripts/buildAnalyticsMaterializedView.sh
```diff
@@ -6,5 +6,9 @@ source .env
 
 cd scripts
 export PGPASSWORD=$DB_PASSWORD
-psql -p $DB_PORT -h $DB_URL -d $DB_NAME -U $DB_USER -f createAnalyticsMaterializedView.sql
-psql -p $DB_PORT -h $DB_URL -d $DB_NAME -U $DB_USER -f createAnalyticsTableIndexes.sql
+if [[ "$1" == "--force" || "$1" == "-f" ]]; then
```
rohan-bes, line 9: We can now force the analytics table to rebuild using `yarn workspace @tupaia/data-api build-analytics-table --force`
This will rebuild the table in the background, and then swap in the table once it's built, so tupaia should stay live the whole time.

packages/data-api/scripts/fastRefreshAnalyticsTable.sh
```diff
@@ -8,4 +8,4 @@ source .env
 
 cd scripts
 export PGPASSWORD=$DB_PASSWORD
-psql -p $DB_PORT -h $DB_URL -d $DB_NAME -U $DB_USER -c "SELECT mv\$refreshMaterializedView('analytics', 'public', true);"
+psql -p $DB_PORT -h $DB_URL -d $DB_NAME -U $DB_USER -tc "SELECT mv\$refreshMaterializedView('analytics', 'public', true);"
```
rohan-bes, line 11: Adding `-t` to all these places just makes the output a little neater (doesn't mention anything about the number of rows returned)

packages/database/src/migrations/20210816230203-MoveAnalyticsTableUtilityFunctionsToDatabaseFunctions-modifies-schema.js
```diff
@@ -0,0 +1,250 @@
...
+          tStartTime := clock_timestamp();
+          PERFORM mv$createMaterializedView(
+              pViewName           => 'analytics_tmp',
+              pSelectStatement    =>  pSqlStatement,
+              pOwner              => 'public',
+              pFastRefresh        =>  TRUE
+          );
+          RAISE NOTICE 'Created analytics table, took %', clock_timestamp() - tStartTime;
+          PERFORM mv$removeMaterializedView('analytics', 'public');
+          PERFORM mv$renameMaterializedView('analytics_tmp', 'analytics', 'public');
```
rohan-bes, line 160: This is using a new function in `pg-mv-fast-refresh` that has been added to rename a MV. Seen here: https://github.com/beyondessential/pg-mv-fast-refresh/pull/4

edmofro (approved):
Looks great! Was actually an easy review using `partial-diff` locally, the migrations are almost all exactly the same as the deleted files :-)

PR #3023, opened 2021-08-18
MEL-387: Fix edit existing viz small bugs

chris-bes:
### Changes:

- Fix edit existing viz small bugs

### DIFF
--- a/packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationCombiner.ts
+++ b/packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationCombiner.ts
@@ -34,7 +34,7 @@ export class DashboardVisualisationCombiner {
     const { config: reportConfig } = this.report;
     const { output } = reportConfig;
     const { config: dashboardItemConfig } = this.dashboardItem;
-    const { type, ...config } = dashboardItemConfig;
+    const { type, name, ...config } = dashboardItemConfig;
 
     return {
       type,
--- a/packages/admin-panel/src/VizBuilderApp/api/constants.js
+++ b/packages/admin-panel/src/VizBuilderApp/api/constants.js
@@ -7,6 +7,5 @@
 export const DEFAULT_REACT_QUERY_OPTIONS = {
   retry: 0,
   // should be refetched in the background every hour
-  staleTime: 1000 * 60 * 60,
   refetchOnWindowFocus: false,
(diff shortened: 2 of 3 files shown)


packages/admin-panel/src/VizBuilderApp/api/constants.js
```diff
@@ -7,6 +7,5 @@
 export const DEFAULT_REACT_QUERY_OPTIONS = {
   retry: 0,
   // should be refetched in the background every hour
-  staleTime: 1000 * 60 * 60,
```
chris-bes, line 10: The query cache has caused some annoying scenarios such as when you update the viz from somewhere else (Admin Panel Dashboard Item modal) and it doesn't reflect in the viz-builder page. So I think it's better to disable the cache and always fetch the latest data

packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationCombiner.ts
```diff
@@ -34,7 +34,7 @@ export class DashboardVisualisationCombiner {
     const { config: reportConfig } = this.report;
     const { output } = reportConfig;
     const { config: dashboardItemConfig } = this.dashboardItem;
-    const { type, ...config } = dashboardItemConfig;
+    const { type, name, ...config } = dashboardItemConfig;
```
chris-bes, line 37: The `name` of the viz should be updated by the Main Edit modal of the viz-builder, not through chart config. So this is to excluding it from the chart config

packages/admin-panel/src/VizBuilderApp/api/constants.js
kael89: Good call - viz builder data should be fairly quick to fetch!

packages/admin-panel/src/VizBuilderApp/api/queries/useReportPreview.js
```diff
@@ -20,6 +21,7 @@ export const useReportPreview = (visualisation, project, location, enabled, setE
       return response.results;
     },
     {
+      ...DEFAULT_REACT_QUERY_OPTIONS,
```
kael89, line 24: A nice improvement would be defining our own `query` helper, which uses `DEFAULT_REACT_QUERY_OPTIONS`

kael89 (approved):
Looks good @chris-bes !

PR #3025, opened 2021-08-18
MEL-13: Viz builder export

kael89:
### Issue #:
MEL-13

New viz export: [My_New_Viz.zip](https://github.com/beyondessential/tupaia/files/7005241/My_New_Viz.zip)
Existing viz export: [LESMIS_num_schools_by_education_level.zip](https://github.com/beyondessential/tupaia/files/7005242/LESMIS_num_schools_by_education_level.zip)

### DIFF
--- a/packages/admin-panel/src/VizBuilderApp/api/mutations/useExportVisualisation.js
+++ b/packages/admin-panel/src/VizBuilderApp/api/mutations/useExportVisualisation.js
@@ -1,0 +1,18 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { useMutation } from 'react-query';
+import { download } from '../api';
+
+export const useExportVisualisation = config =>
+  useMutation(['export/dashboardVisualisation', config], () =>
+    download(
+      'export/dashboardVisualisation',
+      {
+        data: { visualisation: config },
+      },
+      `${config.code || 'New Dashboard Visualisation'}.json`,
+    ),
+  );
--- a/packages/server-boilerplate/src/routes/Route.ts
+++ b/packages/server-boilerplate/src/routes/Route.ts
@@ -24,6 +24,8 @@ type DownloadResBody = {
   type?: string;
 };
 
+type RouteType = 'default' | 'download';
+
 export class Route<
   Req extends ExpressRequest<Req> = Request,
   Res extends ExpressResponse<Req> = Response<ResBody<Req>>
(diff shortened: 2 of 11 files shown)


packages/admin-panel/src/VizBuilderApp/api/mutations/useExportVisualisation.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { useMutation } from 'react-query';
+import { download } from '../api';
+
+export const useExportVisualisation = config =>
+  useMutation(['export/dashboardVisualisation', config], () =>
```
kael89, line 10: This method is called `useExportVisualisation`, but specifically calls `export/dashboardVisualisation`. We can refactor this when we add map overlay support, either by providing the type of viz to export (if the process is almost identical), or by renaming it to `useExportDashboardVisualisation`.

The decision was made based on an existing mutation (`useSaveVisualisation`)

chris-bes: Yep I agree, for front end, I'm leaning towards the generic term `visualisation` if possible (because I think mostly the front end process between building map overlay and building dashboard are the same, and they will most likely use the same visualisation schema, ie: (having `data`, `presentation`,...). The only difference is just calling different endpoints based on the type. It's still very early to tell but that's my gut feeling!

Only when it comes to the back end, then I'm leaning towards calling them separately (ie: DashboardVisualisation, MapOverlayVisualisation,...). This is because I think backend process for map overlays and dashboards are likely going to be different.

packages/server-boilerplate/src/routes/Route.ts
```diff
@@ -41,12 +57,30 @@ export class Route<
     // swallowed.
     try {
       const response = await this.buildResponse();
-      this.respond(response, 200);
+      if (this.type === 'download') {
+        this.download(response);
+      } else {
+        this.respond(response, 200);
+      }
```
chris-bes, line 64: Ah nice!

chris-bes (approved):
LGTM! 💟

packages/server-boilerplate/src/routes/Route.ts
```diff
@@ -41,12 +57,30 @@ export class Route<
...
+    if (type === '.json') {
+      writeJsonFile(filePath, contents);
+    } else {
+      fs.writeFileSync(filePath, contents);
+    }
+
+    this.res.download(filePath, () => {
+      fs.unlinkSync(filePath);
+    });
+  };
```
kael89, line 82: @edmofro tagging this as candidate for refactoring after https://github.com/beyondessential/tupaia/pull/2965 is merged (can use `respondWithDownload` from that PR)

edmofro: Good call!

PR #3026, opened 2021-08-18
MEL-412 allow entity import with no location

avaek:
### Changes:
- We use entities with no location data like "Unknown Island" in Samoa to record some data
- Importing/updating children of these entities is impossible via the admin panel because of the error that was being thrown because these entities don't have location data
- This change allows entities like "Unknown Island" to continue to have no location data and for us to be able to import/update children of this entity in the admin panel

### DIFF
--- a/packages/admin-panel/src/pages/resources/EntitiesPage.js
+++ b/packages/admin-panel/src/pages/resources/EntitiesPage.js
@@ -64,6 +64,14 @@ const IMPORT_CONFIG = {
         type: 'boolean',
       },
     },
+    {
+      label: 'Automatically fetch GeoJSON (defaults to "Yes")',
+      parameterKey: 'automaticallyFetchGeojson',
+      type: 'boolean',
+      editConfig: {
+        type: 'boolean',
+      },
+    },
   ],
 };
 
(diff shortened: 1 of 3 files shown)


packages/meditrak-server/src/routes/importEntities/importEntities.js
```diff
@@ -15,9 +15,13 @@ import { assertCanImportEntities } from './assertCanImportEntities';
...
   try {
-    const { models } = req;
+    const { models, query } = req;
+
     let entitiesByCountryName;
-    const pushToDhis = req?.query?.pushToDhis ? req?.query?.pushToDhis === 'true' : true;
+    const pushToDhis = query?.pushToDhis ? query?.pushToDhis === 'true' : true;
+    const automaticallyFetchGeojson = query?.automaticallyFetchGeojson
+      ? query?.automaticallyFetchGeojson === 'true'
+      : true;
```
edmofro, line 24: This is a) a bit funky, and b) could be DRYer given they both do the same thing. Might be good to have generic handling of boolean options coming from the front end?

packages/admin-panel/src/pages/resources/EntitiesPage.js
```diff
@@ -64,6 +64,14 @@ const IMPORT_CONFIG = {
         type: 'boolean',
       },
     },
+    {
+      label: 'Automatically fetch GeoJSON (defaults to "Yes")',
```
edmofro, line 68: Would be good if the default showed up as the checked box, rather than having to state it here in the label.

edmofro (approved):
A couple of suggestions, but out of scope and out of time for now!

Nice one @avaek, this seems a great change that should stop a lot of wasted time

PR #3035, opened 2021-08-18
Bugfix: Check value of kobo disable var, not just existence

EMcQ-BES:
### Changes:

- Example

---


kael89 (approved):
I think this had also gotten me in the past 🙂 Fix looks good!

PR #3042, opened 2021-08-19
SOL-275: Added General and Clinical Dashboards to Nauru eHealth project

rohan-bes:
### Issue SOL-275:
I've only approached half of the ticket in this dev work here, basically I've cleaned up the permissions model and added the dashboards themselves. The second part of the ticket will be to manually add dashboard relations to turn on dashboards items via the admin panel, which can be a data task. I've discussed with Erin and she's both happy with this approach and is interested in how to do that herself.

### Changes:
- Cleaned up permissions model for Nauru eHealth


nick-beyondessential (approved):
LGTM!

PR #3047, opened 2021-08-19
Disable generate source maps to reduce build size

chris-bes:
### Changes:

- Disable generate source maps to reduce build size.

### Notes:
- Sandbox environment here: http://chris-dev-clone.tupaia.org
- Can confirm that source maps are not generated in sandbox environment:
chrisdevclone
tupaiaorg

- I haven't checked how much size were reduced in front end builds without generating sourcemaps, but the build was deployed fine without failing here:
https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/9477e7a1-cba7-4024-afbb-5490d03c82f2?component=step_deploy-all_deploy-frontends_deploy-psss
- Tested on sandbox deployment, and the front ends (tupaia, admin-panel, psss, lesmis,...) were working fine

### DIFF
--- a/packages/admin-panel/package.json
+++ b/packages/admin-panel/package.json
@@ -11,7 +11,7 @@
   "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
   "main": "dist/index.js",
   "scripts": {
-    "build": "react-scripts build",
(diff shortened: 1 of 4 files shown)


edmofro (approved):
lgtm, thanks for testing this @chris-bes!

PR #3049, opened 2021-08-19
MEL-28: Viz builder import

kael89:
### Issue #:
MEL-28


packages/meditrak-server/src/routes/dashboardVisualisations/GETDashboardVisualisations.js
```diff
@@ -44,12 +75,38 @@ export class GETDashboardVisualisations extends GETHandler {
...
+        permissionGroup: permissionGroup.name,
+      },
     };
   }
+
+  async findRecords(inputCriteria) {
+    const criteria = parseCriteria(inputCriteria);
+
+    const dashboardItems = await this.models.dashboardItem.find(criteria);
+    const reports = await this.models.report.find({ code: dashboardItems.map(di => di.code) });
```
edmofro, line 95:
```suggestion
    const reports = await this.models.report.find({ code: dashboardItems.map(di => di.report_code) });
```

packages/meditrak-server/src/routes/dashboardVisualisations/GETDashboardVisualisations.js
```diff
@@ -44,12 +75,38 @@ export class GETDashboardVisualisations extends GETHandler {
...
+
+    return dashboardItems.map(dashboardItem =>
+      buildVisualisationObject(dashboardItem, reportsByCode, permissionGroupsById),
+    );
+  }
+
+  async countRecords(inputCriteria) {
+    return this.database.count(TYPES.DASHBOARD_ITEM, parseCriteria(inputCriteria), {
+      report: ['dashboard_item.code', 'report.code'],
+    });
```
edmofro, line 110: It looks like you're wanting to join to `report` here, but I think the syntax is something like:
```suggestion
    return this.database.count(TYPES.DASHBOARD_ITEM, parseCriteria(inputCriteria), {
      joinWith: 'report',
      joinCondition: ['dashboard_item.report_code', 'report.code'],
    });
```
However, maybe it's just not needed at all for a count?

packages/meditrak-server/src/routes/dashboardVisualisations/GETDashboardVisualisations.js
```diff
@@ -3,14 +3,45 @@
...
+  return {
+    'dashboard_item.id': id,
+    'dashboard_item.code': code,
+    legacy: false,
+  };
+};
+
+const buildVisualisationObject = (dashboardItemObject, reportsByCode, permissionGroupsById) => {
+  const { model, ...dashboardItem } = dashboardItemObject;
+  const report = reportsByCode[dashboardItem.code];
```
edmofro, line 29:
```suggestion
  const report = reportsByCode[dashboardItem.report_code];
```

edmofro (changes requested):
Quick turnaround @kael89, nice one! 

There's a little confusion with using the same code in the construct (dashboardVisualisation) bleeding into an assumption that the `dashboard_item` and `report` have the same code - the join is actually on `dashboard_item.report_code`. I'm not sure I've spotted them all so please do a sweep.

Otherwise the code is all great! I did have a bit of a UX surprise though - I assumed you would be able to import _within_ the viz builder, and have it not immediately save to the db. This would allow importing a draft without it persisting anywhere, and continuing to work on it. I do agree that when importing through the original dashboard report tab, it makes total sense to save to the db immediately, so this is definitely a good addition, but perhaps we can note that additional "draft import" in another card?

packages/admin-panel-server/.gitignore
```diff
@@ -1,83 +1 @@
-dist
```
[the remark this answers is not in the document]
kael89: That's exactly right! 🙂

packages/meditrak-server/src/routes/dashboardVisualisations/GETDashboardVisualisations.js
kael89: Oops good catch!

kael89: About whether it's needed: I used the implicit definition of a "dashboard visualisation" being a combination of a dashboard_item + a report, where both resources must be present.

In my implementation, all requested items should have a report at this point: if not, an error would have been thrown in [L31](https://github.com/beyondessential/tupaia/blob/mel-28-viz-builder-import-endpoint/packages/meditrak-server/src/routes/dashboardVisualisations/GETDashboardVisualisations.js#L31). However, in the interest of clarity and robustness in the face of potential future refactoring, I'm thinking of keeping the join here - let me know if you disagree.

kael89:
> There's a little confusion with using the same code in the construct (dashboardVisualisation) bleeding into an assumption that the `dashboard_item` and `report` have the same code - the join is actually on `dashboard_item.report_code`

Great catch @edmofro !!



> I did have a bit of a UX surprise though - I assumed you would be able to import _within_ the viz builder, and have it not immediately save to the db

About the above: I assumed the same, but after a discussion with @rohan-bes I decided to stick to the given UX (see https://linear.app/tupaia/issue/MEL-11#comment-5d3917f0). I have a feeling that we should re-examine some UX aspects of import/export, and I have left some [comments](https://linear.app/tupaia/issue/MEL-62#comment-622118e3) in the Copy/Clone card. However, none of them are blockers, so I will wait for feedback from users and the product owner before implementing any further changes.

packages/meditrak-server/src/routes/dashboardVisualisations/GETDashboardVisualisations.js
edmofro: Happy to keep it :-)

edmofro (approved):
Cool, I'm glad you guys had already covered that chat. My sense is that we should update the UX as you suggested in the card, and perhaps the flows would be something like:
- From dashboard report page
  - Click "new" button
  - Can choose to enter details or just choose a file to import
  - If importing, import can be "import draft" or "import and save", defaults to "import and save"
- From viz builder initial screen
  - Can choose to enter details or just choose a file to import
    - "Import draft" or "import and save", defaulting to "import draft"
- From viz builder proper
  - An import button next to the export button would allow
    - "Import draft" or "import and save", defaulting to "import draft"

But as you say, happy for these UX tweaks to be made in the next round, once we have some use.

PR #3050, opened 2021-08-19
nz-913 LESMIS Session expiry causes entity fetch to fail

tcaiger:
### Changes:

The issue appears to be that the session is expired and although we are handling the case where there is no session in the database there is no handling for when the session is expired or the refresh token is invalid. The fix is to make a logout request to the server to clear the sessionCookie. This will cover other possible potential edge cases as well. It's the same approach that we took to handle the session issues on admin-panel.

---


packages/lesmis/src/api/api.js
```diff
@@ -58,6 +58,11 @@ const request = async (endpoint, options) => {
     if (error.response) {
       const { data } = error.response;
 
+      // logout the user out if we get a 401 response
+      if (error.response.status === 401 || data.code === 401) {
+        request('logout', { method: 'post' });
```
edmofro, line 63: Should we `return` after this, to avoid the error handling below?

edmofro (commented):
Good solve - just a question and I'm not actually sure what the answer is

packages/lesmis/src/api/api.js
tcaiger: hmm good question. I'm not entirely sure either. I did a quick test and in all the cases I can think of it makes no difference. I'm inclined to not return as it is more predictable to keep the consistent flow of error handling?!

edmofro: Sure thing!

edmofro (approved):

PR #3055, opened 2021-08-20
Update Covid Samoa SubDistricts Geodata (hotfix - test on dev)

nick-beyondessential:
Update for geodata boundaries for Covid Samoa to be done as a hotfix.
Done as migration because of 'Unknown' type enties without geodata or entity 'parent_id's that break normal entity import.
Will test on dev and recreate pull request against master once checked by @gerrybes.


edmofro:
@nick-beyondessential did you talk to @avaek about a workaround to get this into prod without a release? It'd be good to avoid a whole release process just for the data update if possible 🙏

nick-beyondessential:
@edmofro yes but the other issue is the null parent_id is also confounding import - could do some direct sqling but that seems riskier?

nick-beyondessential:
Done via admin panel and sql update so closing this PR

PR #3058, opened 2021-08-20
Remove need for parent_code to be set on entity import

edmofro:
### Issue #:
SOL-392


packages/meditrak-server/src/routes/importEntities/getOrCreateParentEntity.js
```diff
@@ -140,5 +140,5 @@ export async function getOrCreateParentEntity(
   // no explicit parent code provided, use either subdistrict or district as parent entity
   if (subDistrict) return { parentGeographicalArea: subDistrict, parentEntity: subDistrictEntity };
   if (district) return { parentGeographicalArea: district, parentEntity: districtEntity };
-  throw new Error('No parent entity defined');
```
nick-beyondessential, line 143: Probably part of a wider conversation but thinking it'd be good if we were able to throw a warning so that if a parent is missed by accident rather than by design the importer is notified.

edmofro: Yeah it's a good point, but I am thinking we're probably moving towards a two phase import where you import the relations separately anyway

nick-beyondessential: Removing parent_code from geojson import file results in
image

edmofro: You're taking this over as a starting point right @nick-beyondessential?

nick-beyondessential: Yes sounds good.

packages/meditrak-server/src/apiV2/import/importEntities/updateCountryEntities.js
```diff
@@ -147,7 +143,7 @@ export async function updateCountryEntities(
       { code },
       {
         name,
-        parent_id: parentEntity.id,
+        parent_id: parentEntity ? parentEntity.id : null,
```
nick-beyondessential, line 146: A bit weird but using parentEntity?.id didn't update parent_id to null whereas using a trinary statement does. Couuldn't work out why.

edmofro: `undefined` get stripped before making it to the db update query, whereas `null` doesn't :+1:

nick-beyondessential: Ah ofcourse - was thinking it'd default to null, undefined makes sense.

packages/meditrak-server/src/apiV2/import/importEntities/updateCountryEntities.js
```diff
@@ -93,12 +93,8 @@ export async function updateCountryEntities(
...
     codes.push(code);
-    const { parentGeographicalArea, parentEntity } = await getOrCreateParentEntity(
-      transactingModels,
-      entityObject,
-      country,
-      pushToDhis,
-    );
+    const { parentGeographicalArea, parentEntity } =
+      (await getOrCreateParentEntity(transactingModels, entityObject, country, pushToDhis)) || {};
     if (entityType === transactingModels.entity.types.FACILITY) {
```
nick-beyondessential, line 98: Will leave parent_id compulsory for facilities.

packages/admin-panel/src/pages/resources/EntitiesPage.js
```diff
@@ -65,7 +65,7 @@ const IMPORT_CONFIG = {
       },
     },
     {
-      label: 'Automatically fetch GeoJSON (defaults to "Yes")',
+      label: 'Automatically fetch GeoJSON (defaults to "Yes", requires parent_code)',
```
nick-beyondessential, line 68: Happy to remove this if too cluttered but auto geoJson fetch will fail with no parent_codes
image

edmofro: Yeah I'd leave off and have it throw an error that can show up here instead

edmofro (commented):
Looking good but I can't actually approve it as I opened the PR @nick-beyondessential

kael89 (approved):
Had a look across both @edmofro 's and your commits, looks good @nick-beyondessential ! A lot of code paths in this feature so easy to miss a detail or too... 🤞

PR #3059, opened 2021-08-20
Treat includeRootEntity param in URL as string in entity server

billli0:
### Issue #:
entity-server endpoint: entities getDescendants() always return root entities. 

### Changes:
treat `includeRootEntity` param in URL as string in entity server


edmofro (approved):
lgtm!

PR #3062, opened 2021-08-20
MEL-449: Added BES Data Admin permission restrictions on /testReport endpoint

rohan-bes:
### Issue MEL-449:

Had a couple options of how we could enforce permissions here:
1. Give access to all org units for users with 'BES Admin' access
2. Give access to org units if user has 'BES Data Admin' access to that org unit

It seems that generally in Admin Panel we allow users to do operations if they have 'BES Admin' access to any country. However, for viewing data in our applications, we restrict it per org unit, and the highest org unit level permission you can be given is 'BES Data Admin'. For this reason, I felt it was more appropriate to go with the second option.

### DIFF
--- a/packages/report-server/src/routes/TestReportRoute.ts
+++ b/packages/report-server/src/routes/TestReportRoute.ts
@@ -22,11 +22,43 @@ export type TestReportRequest = Request<
   ReportRouteQuery
 >;
 
+const BES_DATA_ADMIN_PERMISSION_GROUP = 'BES Data Admin';
+
 export class TestReportRoute extends Route<TestReportRequest> {
+  async checkUserHasAccessToReport(hierarchy = 'explore', requestedOrgUnitCodes: string[]) {
+    const { accessPolicy, ctx } = this.req;
+
+    const foundOrgUnits = await ctx.microServices.entityApi.getEntities(
+      hierarchy,
+      requestedOrgUnitCodes,
+      {
+        fields: ['code', 'country_code'],
+      },
+    );
+    const foundOrgUnitCodes = foundOrgUnits.map(orgUnit => orgUnit.code);
+
+    const missingOrgUnitCodes = requestedOrgUnitCodes.filter(
+      orgUnitCode => !foundOrgUnitCodes.includes(orgUnitCode),
+    );
+    if (missingOrgUnitCodes.length > 0) {
+      throw new Error(`No entities found with codes ${missingOrgUnitCodes}`);
+    }
+
+    const countryCodes = new Set(foundOrgUnits.map(orgUnit => orgUnit.country_code));
+    countryCodes.forEach(countryCode => {
+      if (
+        countryCode === null ||
+        !accessPolicy.allows(countryCode, BES_DATA_ADMIN_PERMISSION_GROUP)
+      ) {
+        throw new Error(`No ${BES_DATA_ADMIN_PERMISSION_GROUP} access for user to ${countryCode}`);
+      }
+    });
+  }
+
   async buildResponse() {
     const { query, body } = this.req;
     const { testData, testConfig, ...restOfBody } = body;
-    const { organisationUnitCodes, ...restOfParams } = { ...query, ...restOfBody };
+    const { organisationUnitCodes, hierarchy, ...restOfParams } = { ...query, ...restOfBody };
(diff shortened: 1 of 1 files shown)


packages/report-server/src/routes/TestReportRoute.ts
```diff
@@ -22,11 +22,43 @@ export type TestReportRequest = Request<
...
+    const countryCodes = new Set(foundOrgUnits.map(orgUnit => orgUnit.country_code));
+    countryCodes.forEach(countryCode => {
+      if (
+        countryCode === null ||
+        !accessPolicy.allows(countryCode, BES_DATA_ADMIN_PERMISSION_GROUP)
+      ) {
+        throw new Error(`No ${BES_DATA_ADMIN_PERMISSION_GROUP} access for user to ${countryCode}`);
+      }
+    });
+  }
```
rohan-bes, line 56: @billli0 depending on in this PR of your PR: https://github.com/beyondessential/tupaia/pull/3061 goes in first, one of us should refactor this route and the `FetchReportRoute` to share this common code, as they have a lot of similarity now.

billli0: Thanks for letting me know @rohan-bes 🙌

packages/report-server/src/routes/TestReportRoute.ts
```diff
@@ -22,11 +22,43 @@ export type TestReportRequest = Request<
...
+      orgUnitCode => !foundOrgUnitCodes.includes(orgUnitCode),
+    );
+    if (missingOrgUnitCodes.length > 0) {
+      throw new Error(`No entities found with codes ${missingOrgUnitCodes}`);
+    }
+
+    const countryCodes = new Set(foundOrgUnits.map(orgUnit => orgUnit.country_code));
+    countryCodes.forEach(countryCode => {
+      if (
+        countryCode === null ||
```
edmofro, line 50: I guess this is currently the case, but confirming we're baking in that no data can be stored directly against the project entity (or at least, that it can't then be visualised via the report-server)

edmofro (approved):
Thanks for highlighting and plugging this important gap @rohan-bes!!

packages/report-server/src/routes/TestReportRoute.ts
rohan-bes: I'm actually pretty keen to get rid the `country !== null` expectation, as I think it should be fine for us to work with project entities when fetching data, and I feel the report-server can handle this pretty graciously! However, there's a tricky thing that we haven't considered yet, which is how do we handle permissions being at a finer level of granularity than the entities we're fetching for? eg. 
- User has access to UNFPA project, and access to Tonga and Papua New Guinea.
- Report is 'Regional cases of Malaria' with dataSourceEntityType = `facility`, aggregationEntityType = `project`
- We want the user to see a value that is the total cases in Tonga and Papua New Guinea
- Unfortunately, we have no way of knowing the user permissions at the data calculation time, so we sum the values of facilities in UNFPA

Until we solve this general issue, I don't think we can allow data to be fetched against the project entity 😢 (or at the very least not allow entity aggregation if fetching against the project entity)

edmofro: Yeah I'm very keen to make things work for the project level too, and ideally it would be nice to have permissions able to be applied at a lower level in the hierarchy also (so you could have permission to see data just for a specific village in Tonga, for e.g.). This actually once worked in Tupaia, but we didn't maintain it because no one used it, so it stopped working in a lot of cases, and then I simplified it out for the code cleanliness, performance gains and reduction of maintenance that just assuming permissions are at the country level gave us. I feel we're coming up to the time where we need to bite back into that complexity though!

rohan-bes: I think having the entity-server be responsible for providing the data source entities available to the user puts us in a pretty good position to support this! In theory, we could just pass in a permission group(s) to the entity-server and it will return to us the data source entities that the user has that level of permissions for 🎉

I feel though it's worth us taking the time to work out how we want to determine the set of permissions that a data fetch ought to be associated with though...

edmofro: True! And yes definitely worth going over the options - I'll set up a meeting for that

PR #3064, opened 2021-08-21
Sol 296 7 day positivity dashboard

Alastair-L:
### Issue #: SOL-296


packages/database/src/migrations/20210819150000-AddCovidFiji7DayPositivityRateReport.js
```diff
@@ -0,0 +1,160 @@
...
+        transform: 'aggregate',
+        period: 'group',
+        '...': 'drop',
+        COVID_FJ_7_Day_Rolling_Num_Tests: 'sum',
+        COVID_FJ_7_Day_Rolling_Pos_Tests: 'sum',
+      },
+      {
+        transform: 'filter',
+        where:
+          'exists($row.COVID_FJ_7_Day_Rolling_Num_Tests) and $row.COVID_FJ_7_Day_Rolling_Num_Tests > 0',
```
rohan-bes, line 53: Is this filter on `$row.COVID_FJ_7_Day_Rolling_Num_Tests > 0` to avoid division by 0? If so, yeah this can be rather annoying 😞 

Often times I achieve this by doing a ternary in the select instead, eg:
```
        transform: 'select',
        "'value'": ' $row.COVID_FJ_7_Day_Rolling_Num_Tests == 0 ? 0 : $row.COVID_FJ_7_Day_Rolling_Pos_Tests / $row.COVID_FJ_7_Day_Rolling_Num_Tests',
```

The reason for this is to still display data when the value is actually meant to be 0 (eg. a known 0 tests performed that day should have 0 value, not 'no data').

Still happy to keep the filter for `exists($row.COVID_FJ_7_Day_Rolling_Num_Tests)` though, however I'm guessing that should never actually happen?

rohan-bes (approved):
All looking good to me! Just had one comment about the `filter` in the transforms, but aside from that it's all looking good, so pre-approving! 👍 

One question though, how is the performance looking with the indicators? I'd love for it to work fast but I have a worry it might be rather slow..?

packages/database/src/migrations/20210819150000-AddCovidFiji7DayPositivityRateReport.js
Alastair-L: Hmmm, yeah that's a good point! Although I'm not sure if 0% positivity makes sense if there are no tests from a project point of view, I'll ask on the card :)).

Alastair-L: I think the filter can stay based on this comment: https://linear.app/bes/issue/SOL-296#comment-4a7b6553?
Let me know if you're happy with that 👍

Alastair-L:
Speaking to performance, I couldn't see any problems with the ~ 480 data points over a year I entered (~4 seconds), however that data was only against 4 facilities (and the indicator acts before the entity aggregation, so it could be potentially ~60x slower if there's data for all facilities)?

packages/database/src/migrations/20210819150000-AddCovidFiji7DayPositivityRateReport.js
rohan-bes: Yup, makes sense to me! Good call 👍

rohan-bes:
> Speaking to performance, I couldn't see any problems with the ~ 480 data points over a year I entered (~4 seconds), however that data was only against 4 facilities (and the indicator acts before the entity aggregation, so it could be potentially ~60x slower if there's data for all facilities)?

In theory the performance should really be most affected by the number of fetches, rather than the amount of records each fetch returns. Therefore increasing the number of facilities fetched for oughtn't matter... Might have to wait and see in practice though 👍 

If it is poor, is performing the entity aggregation in the indicator an option?

edmofro:
Where did this one get to guys? Either way, the report will now need to be updated. Whew @Alastair-L building up a list of a few things to get through, hope you have a uni break sometime soon 😅

Alastair-L:
Closing in favor of: https://github.com/beyondessential/tupaia/pull/3209
@rohan-bes would you please be able to review?

PR #3076, opened 2021-08-23
MEL-441: Update react scripts

edmofro:
### Issue #:
Towards MEL-441

### Changes:
Update react scripts to 4.0.3, which is the latest. This probably has a bunch of improvements in it, but specifically interested in:
- Uses a more recent version of `terser`, which has a fix in it for using too many cpus and running out of memory
- Forces us to fix linter errors, most significantly the import cycles, which may have been a significant cause of the build being resource intensive


packages/psss/src/components/Table/CellComponents/ArchivedAlertMenuCell.js
```diff
@@ -1,9 +1,9 @@
 import React, { useCallback, useContext } from 'react';
+import PropTypes from 'prop-types';
 import { ActionsMenu } from '@tupaia/ui-components';
-import { ArchiveTableContext } from '../../../containers/Tables';
 import * as COLORS from '../../../constants/colors';
 
-export const ArchivedAlertMenuCell = React.memo(({ id }) => {
+export const ArchivedAlertMenuCell = React.memo(({ id, ArchiveTableContext }) => {
```
kael89, line 6: I haven't seen this pattern again (passing the context as prop), instinctively it seems a bit unexpected but it may be fine - would like to get another opinion from @tcaiger 

I assume that the motivation for this change was to avoid an import dependency cycle, and I know that this is sometimes tricky to achieve, so I'm personally not fussed either way!

packages/lesmis/src/constants/page.js
```diff
@@ -112,7 +112,7 @@ export const FQS_PAGE = {
             practices” that could be shared with other cluster schools.
           </ListItem>
           <ListItem component="li">
-            those standards that need improvement -> priorities for improvement to be included in
+            those standards that need improvement -&gt priorities for improvement to be included in
```
kael89, line 115: I'd be happy for us to disable this ESLint rule ([react/no-unescaped-entities](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md))

.eslintrc
```diff
@@ -24,7 +24,8 @@
         "packageDir": ["../../", "."]
       }
     ],
-    "jsx-a11y/href-no-hash": "off"
+    "jsx-a11y/href-no-hash": "off",
+    "react/destructuring-assignment": "off"
```
kael89, line 28: I generally like this rule , since it makes it very clear to see whether something comes from `state` or from `props`. However, I understand that there are cases where it may be more elegant to not destructure and just use eg. `this.props.aFunction()`, so disabling it in the top level seems like a good call.

packages/admin-panel/src/widgets/InputField.js
```diff
@@ -1,257 +0,0 @@
...
-  parentRecord,
-  variant,
-}) => {
-  const inputType = getInputType({ options, optionsEndpoint, type });
-  let inputComponent;
-
-  switch (inputType) {
-    case 'autocomplete':
-      inputComponent = (
-        <Autocomplete
```
kael89, line 50: I like this refactoring!

packages/admin-panel/src/widgets/InputField/registerInputFields.js
```diff
@@ -0,0 +1,165 @@
...
+import React from 'react';
+import moment from 'moment';
+import { TextField, DatePicker, DateTimePicker, RadioGroup, Select } from '@tupaia/ui-components';
+import { stripTimezoneFromDate } from '@tupaia/utils';
+import { registerInputField } from './InputField';
+import { Autocomplete } from '../../autocomplete';
+import { JsonInputField } from './JsonInputField';
+import { JsonEditor } from './JsonEditor';
+
+// "InputField" is treated as a dynamic factory, where different input types can be supported
```
kael89, line 15: Clever pattern!

packages/admin-panel/src/widgets/InputField/Autocomplete.js
```diff
@@ -0,0 +1,42 @@
...
+    disabled={disabled}
+    allowMultipleValues={allowMultipleValues}
+    parentRecord={parentRecord}
+  />
+);
+
+AutocompleteField.propTypes = inputFieldPropTypes;
+AutocompleteField.defaultProps = inputFieldDefaultProps;
+
+registerInputField('autocomplete', AutocompleteField);
```
kael89, line 42: This is already registered in `registerInputFields`:

```suggestion
```

kael89 (approved):
Great refactor! So the new version of `react-scripts` enforces us to fix linting errors? That's interesting! I'd assume that they would allow users to skip this, but I definitely love the cleanup in this PR ♥

It makes me think whether it would pay off to be a bit stricter with following the linting rules. I think that as soon as you get used to them, it's very easy to follow them in practice. Plus, if a rule is too strict we can turn it into a warning, or disable it.

A few minor suggestions but pre-approving

packages/psss/src/components/Table/CellComponents/ArchivedAlertMenuCell.js
kael89: Or could we extract `ArchiveTableContext` in a new file and thus avoid the dependency loop?

kael89:
@edmofro one last comment I forgot: there is a file in this PR that doesn't seem to be used (`packages/admin-panel/src/widgets/InputField/Enum.js`)

packages/admin-panel/src/widgets/InputField/Autocomplete.js
edmofro: Ooops, left over from a first attempt, thanks for catching!

packages/psss/src/components/Table/CellComponents/ArchivedAlertMenuCell.js
edmofro: @chris-bes could you follow up on this one?

tcaiger: Yes, I agree with Kostas. thanks

edmofro:
Thanks for the review @kael89! I'm sure there is a way to ask react-scripts to be less strict, but it was a great opportunity to keep in place something that will give us a bit more discipline, especially around circular dependencies which I strongly suspect have been contributing to our heavy resource use during builds

All my bits fixed up, @chris-bes will respond after lunch then this will be ready for re-review

kael89:
> I'm sure there is a way to ask react-scripts to be less strict, but it was a great opportunity to keep in place something that will give us a bit more discipline, especially around circular dependencies which I strongly suspect have been contributing to our heavy resource use during builds

Definitely a good call, kudos for that!

packages/psss/src/components/Table/CellComponents/ArchivedAlertMenuCell.js
chris-bes: Definitely was a poor effort to fix the dependency cycle issue 🤦‍♂️ . Thanks for pointing that out @kael89 !

kael89 (approved):
Nice cleanup!!

chris-bes:
Have also addressed my bits

packages/web-frontend/src/components/Marker/markerIcons.js
```diff
@@ -17,11 +23,12 @@ import Help from '@material-ui/icons/Help';
 import CheckBox from '@material-ui/icons/CheckBox';
 
 // from https://thenounproject.com/ochavisual/collection/ocha-humanitarian-icons/
-import Cyclone from '../../images/cyclone.svg';
-import Earthquake from '../../images/earthquake.svg';
-import Tsunami from '../../images/tsunami.svg';
-import Volcano from '../../images/volcano.svg';
-import Flood from '../../images/flood.svg';
+
+import Cyclone from '!!react-svg-loader!../../images/cyclone.svg';
```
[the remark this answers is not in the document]
tcaiger: Hey @chris-bes, I just saw this conversation. I have another PR open to move all the marker code to ui-components along with all the map-components from web-frontend. I ran into a similar issue there and resolved it in the rollup build config. So let's either wait until those changes are merged in to dev or come up with a quick work around so that you can move on. I'll send you a PM.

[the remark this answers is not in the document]
tcaiger: @edmofro What do we think about hardcoding the disaster icons to tupaia orange as a quick fix for this temporary code? (As far as I can tell they are always displayed orange). Then when my map refactoring is complete this code will be deleted.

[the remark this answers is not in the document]
edmofro: Eeek yes sorry @tcaiger could you take this over now @chris-bes is off on Tamanu? Hardcoding to tupaia orange is totally fine with me.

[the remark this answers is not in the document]
tcaiger: @edmofro - yes can do. My understanding is that this file is the last outstanding issue. Is that correct?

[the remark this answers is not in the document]
edmofro: Yes plus a merge from dev with some conflicts to resolve, and my accidental commit before lunch (was meant for a different branch). Thanks heaps for taking care of the svg stuff, I'll rebase the branch now to get rid of that commit and resolve the conflicts!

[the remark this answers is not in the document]
edmofro: Rebased and resolved conflicts. Hopefully I got it all right. Ready for your input on the svgs @tcaiger!

tcaiger:
@edmofro and @kujo4pm - Ready for review. I just added 1 new commit.

packages/web-frontend/package.json
```diff
@@ -95,16 +94,7 @@
...
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
-  }
+  "browserslist": [
+    "defaults"
```
tcaiger, line 98: F some reason those settings cause an error with svg imports. Using the defaults works fine though and the list of browsers it includes is very similar so I am keen to go for the defaults. See here for a list of the browsers included in each. https://beyondessential.slack.com/archives/C01UPMP1LPN/p1630627300000900

edmofro:
I can't review my own PR unfortunately, but your commit lgtm! Thanks heaps for that @tcaiger 

(@kujo4pm I believe you can ignore that tag ;-) )

kael89 (approved):
Lgtm too, thanks @tcaiger !

packages/web-frontend/package.json
IgorNadj: Thanks for that Tom! BTW I ran into the same issue in admin-panel, could you make the same change there as well? Thanks

PR #3084, opened 2021-08-24
Analytic table week_period has incorrect year when 0 week

billli0:
### Issue #:
https://linear.app/bes/issue/SOL-394/analytic-table-week-period-has-incorrect-year-when-0-week

### Changes:
copy and paste migration `20210816230203-MoveAnalyticsTableUtilityFunctionsToDatabaseFunctions-modifies-schema` then change `extract(year)` to `extract(isoyear)`


packages/database/src/migrations/20210824054133-FixAnalyticTableWeekPeriodIncorrectYear-modifies-schema.js
```diff
@@ -0,0 +1,116 @@
...
+      SELECT
+        entity.code as entity_code,
+        entity.name as entity_name,
+        question.code as data_element_code,
+        survey.code as data_group_code,
+        survey_response.id as event_id,
+        answer.text as value,
+        question.type as type,
+        to_char(survey_response.data_time, ''YYYYMMDD'') as "day_period",
+        concat(extract (isoyear from survey_response.data_time), ''W'', to_char(extract (week from survey_response.data_time), ''FM09'')) as "week_period",
```
rohan-bes, line 34: Such an easy fix! Nice one 🎉

packages/database/src/migrations/20210824054133-FixAnalyticTableWeekPeriodIncorrectYear-modifies-schema.js
```diff
@@ -0,0 +1,116 @@
...
+        );
+        RAISE NOTICE 'Created analytics table, took %', clock_timestamp() - tStartTime;
+        PERFORM mv$removeMaterializedView('analytics', 'public');
+        PERFORM mv$renameMaterializedView('analytics_tmp', 'analytics', 'public');
+      ELSE
+        RAISE NOTICE 'Analytics Materialized View already exists, skipping';
+      END IF;
+    END IF;
+  end $$ LANGUAGE plpgsql;
+`);
```
rohan-bes, line 107: So just a heads up, there's one more thing you gotta do here. You'll need to actually rebuild the analytics table now that we've changed the schema for it. This can be done by calling:
```
await db.runSql('SELECT build_analytics_table(true);');
await db.runSql('SELECT create_analytics_table_indexes();');
```

However, this migration will then take several minutes to run... There is a slightly hacky option we could use if we wish to avoid running the rebuild of the analytics table here... we could update the existing migration? That would work, since the first migration hasn't made it to prod as yet so it's still being run each night... quite hacky but might be a good idea..?

rohan-bes (changes requested):
Nice one @billli0 and very clean fix.

I mention that there's one more step that needs to happen, and suggest we might wanna just update the existing migration?

packages/database/src/migrations/20210824054133-FixAnalyticTableWeekPeriodIncorrectYear-modifies-schema.js
billli0: Oh that is a good point! If that is not in prod, we can just change the existing migration👏

billli0: Also noted the steps you shared to rebuild the analytics table. Thanks @rohan-bes .

rohan-bes (approved):
Nice 😄

PR #3088, opened 2021-08-25
Sol-90: Added PacMossi raw data downloads

PShoe:
### Changes:

- Add raw data downloads for pacmossi countries at country level

---


packages/database/src/migrations/20210824045814-AddPacMossiRawDataDownload-modifies-data.js
```diff
@@ -0,0 +1,171 @@
...
+    child_id: dashboardItemId,
+    entity_types: '{country}',
+    project_codes: '{pacmossi}',
+    permission_groups: '{PacMOSSI Senior}',
+    sort_order: 2,
+  });
+};
+
+exports.up = async function (db) {
+  await Promise.all(COUNTRY_CODES.map(countryCode => addDashboardItemToCountry(db, countryCode)));
```
billli0, line 149:
```suggestion
  await Promise.all(COUNTRY_CODES.map(async countryCode => addDashboardItemToCountry(db, countryCode)));
```

Without async, sometimes it works but sometimes it won't, it would be safe to always have `async` for asynchronous functions.

packages/database/src/migrations/20210824045814-AddPacMossiRawDataDownload-modifies-data.js
```diff
@@ -0,0 +1,171 @@
...
+  seed = seedLink;
+};
+
+const PROJECT_NAME = 'PacMOSSI';
+const COUNTRY_CODES = ['FJ', 'KI', 'NR', 'NU', 'PG', 'SB', 'TV', 'TO', 'VU'];
+const DASHBOARD_NAME = 'PacMOSSI Raw Data Downloads';
+const DASHBOARD_ITEM_NAME = 'Download Vector Surveillance Survey Data';
+
+const getCountryDashboardCode = countryCode =>
+  `${countryCode}_${PROJECT_NAME}_${DASHBOARD_NAME.split(' ').join('_')}`;
```
billli0, line 32:
```suggestion
  `${countryCode}_${DASHBOARD_NAME.split(' ').join('_')}`;
```

Seems like unnecessary to have two `pacmossi` in code `FJ_PacMOSSI_PacMOSSI_Raw_Data_Downloads`,

packages/database/src/migrations/20210824045814-AddPacMossiRawDataDownload-modifies-data.js
```diff
@@ -0,0 +1,171 @@
...
+
+const getCountryDashboardItemCode = countryCode =>
+  `${countryCode}_${PROJECT_NAME}_${DASHBOARD_ITEM_NAME.split(' ').join('_')}`;
```
billli0, line 35: ^ same here.

billli0 (changes requested):
LGTM! Code is clean and works fine👍 Just a few minor suggestions.

packages/database/src/migrations/20210824045814-AddPacMossiRawDataDownload-modifies-data.js
PShoe: Fixed!

packages/database/src/migrations/20210824045814-AddPacMossiRawDataDownload-modifies-data.js
PShoe: These only have one pacmossi in them ex: `KI_PacMOSSI_Raw_Data_Downloads`

packages/database/src/migrations/20210824045814-AddPacMossiRawDataDownload-modifies-data.js
PShoe: Added

packages/database/src/migrations/20210824045814-AddPacMossiRawDataDownload-modifies-data.js
billli0: I see, that's on me. 🙌

billli0 (approved):

PR #3094, opened 2021-08-25
Distribution of species overlays by districts and countries

billli0:
### Issue #:
https://linear.app/bes/issue/SOL-81/regional-map-overlay-country
https://linear.app/bes/issue/SOL-80/map-overlay-district


packages/database/src/migrations/20210824231816-DistributionOfSpeciesPacmoss-modifies-data.js
```diff
@@ -0,0 +1,179 @@
...
+  ['Aedes', 'Ae_Polynesiensis'],
+  ['Aedes', 'Ae_Rotumae'],
+  ['Aedes', 'Ae_Scutellaris'],
+  ['Aedes', 'Ae_Vigilax'],
+  ['Culex', 'Cx_Annulirostris'],
+  ['Culex', 'Cx_Quinquefasciatus'],
+  ['Culex', 'Cx_Sitiens'],
+  ['Mansonia', 'Mn_Uniformis'],
+];
+const HIERARCHIES = ['Country', 'District'];
```
nick-beyondessential, line 37:
```suggestion
const HIERARCHYLEVELS = ['Country', 'District'];
```
Optional naming suggestion just to make it a bit clearer.

packages/report-server/src/routes/helpers.ts
```diff
@@ -39,7 +39,6 @@ export const getRequestedOrgUnitObjects = async (
     throw new Error(`No entities found with codes ${orgUnitCodesInArray}`);
   }
 
-  // If entity is a project
```
nick-beyondessential, line 42: This seems like a weird change was it from the conflict?

nick-beyondessential (approved):
LGTM

packages/report-server/src/routes/helpers.ts
billli0: I delete it for a purpose since the code can explain itself.

nick-beyondessential: Good point :)

PR #3102, opened 2021-08-26
NZ-22 Update web-frontend to use ui components tile picker

tcaiger:
### Changes:

- Use the TilePicker from ui-components in web-frontend
- Add ReferenceTooltip component to ui-components and use it to update the TilePicker
- Use the ReferenceTooltip from ui-components in web-frontend

---


packages/ui-components/src/components/Map/TilePicker/index.js
```diff
@@ -5,3 +5,4 @@
  */
 
 export * from './TilePicker';
+export * from './constants';
```
edmofro, line 8: Was this needed in the end? I can't see it being used, but that might come in a later PR

edmofro (approved):
Love a good clear out! Too many deletions to check in detail, but all of the changes look great

PR #3113, opened 2021-08-27
Only run change side effects in meditrak-server database

edmofro:
I noticed a secondary (non release-blocking) issue while debugging the meditrak-app sync issue this morning: migrations adding indicators were throwing uncaught errors that showed they were trying to create two data sources for each indicator added.

The reason for this is that `IndicatorModel` has a change handler that creates a data source whenever a new one is added. This model is defined centrally, and imported wherever a `ModelRegistry` is created. When Tupaia is running, model registries are created all over the place - even if you're just running meditrak-server, two will be created (one by data-broker).

This PR stops all those extra model registries from calling the models' notifiers.


edmofro:
Merging this without testing with the expectation that 1 week of passive testing on dev + the regression test should be enough

PR #3117, opened 2021-08-27
Fix incorrect assertion that ancestor_filter:type must be defined

rohan-bes:
image

### DIFF
--- a/packages/entity-server/src/routes/hierarchy/relationships/middleware.ts
+++ b/packages/entity-server/src/routes/hierarchy/relationships/middleware.ts
@@ -48,8 +48,10 @@ const getSubContext = (
   const field = (queryField ? extractFieldFromQuery(queryField) : baseField) || 'code';
   const filter = extractFilterFromQuery(allowedCountries, queryFilter);
   const { type, ...restOfFilter } = filter;
-  if (type === null || typeof type !== 'string') {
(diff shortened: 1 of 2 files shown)


packages/entity-server/src/routes/hierarchy/relationships/middleware.ts
```diff
@@ -48,8 +48,10 @@ const getSubContext = (
   const field = (queryField ? extractFieldFromQuery(queryField) : baseField) || 'code';
   const filter = extractFilterFromQuery(allowedCountries, queryFilter);
   const { type, ...restOfFilter } = filter;
-  if (type === null || typeof type !== 'string') {
-    throw new Error(`${from}_filter:type must be a single, not null type constraint`);
+  if (type === null || (type !== undefined && typeof type !== 'string')) {
```
kael89, line 51: Makes it a bit easier to see the allowed types:

```suggestion
  if (typeof type !== 'undefined' && typeof type !== 'string') {
```

kael89 (approved):
Looks good! I left a minor suggestion, but approving

packages/entity-server/src/routes/hierarchy/relationships/middleware.ts
rohan-bes: Oh yeah that's much better!

PR #3121, opened 2021-08-29
NZ-917: Build packages on codeship, then push

edmofro:
#### Manual deployment steps:
- [ ] Update nginx servers.conf
- [ ] Add latest environment variable files to LastPass

### Issue #:

NZ-917

### Changes:

- Build all packages on codeship, within Docker container, pulling environment variables from new LastPass folder
- Then push entire tupaia directory up to replace the one on the instance
- Finally, restart all backends


packages/devops/configs/servers.conf
```diff
@@ -3,7 +3,7 @@ server {
   listen       80;
   listen       [::]:80;
   server_name  ~^.*admin\.tupaia\.org$;
-  root         /home/ubuntu/tupaia/packages/admin-panel/served_build/;
+  root         /home/ubuntu/tupaia/packages/admin-panel/build;
```
edmofro, line 6: Now that we don't need to swap out the build on the server, we can use the more standard `build` directory so that there are less surprises

packages/devops/scripts/ci/pushBuilds.sh
```diff
@@ -0,0 +1,16 @@
...
+PACKAGE=$1
+DIR=$(dirname "$0")
+${DIR}/copyFromCommonVolume.sh # pull in .env files and builds from common volume
+if [ -f "/common/deployment_exists" ]; then
+    echo "Deployment for ${CI_BRANCH} exists, updating with latest builds"
+    DEPLOYMENT_SSH_URL=$(${DIR}/determineDeploymentSshUrl.sh)
+    cd /
+    tar -zcf tupaia.tar.gz /tupaia
+    scp -o StrictHostKeyChecking=no -p /tupaia.tar.gz ubuntu@$DEPLOYMENT_SSH_URL:./tupaia.tar.gz
+    ssh -o ServerAliveInterval=15 ubuntu@$DEPLOYMENT_SSH_URL "tar -zxf tupaia.tar.gz && rm tupaia.tar.gz"
```
edmofro, line 13: Zip, push via scp, then unzip on the server. Tried a few ways and this (although slow at around 8 minutes) was the fastest. I have some thoughts around ways we can improve this in the future, but this keeps it at around the same total build time (~30 mins when the docker image is fully cached) as we have currently, but with the builds happening off the machine so they don't consume resources and should be more stable

package.json
```diff
@@ -21,6 +21,7 @@
     "build-admin-panel": "yarn workspace @tupaia/admin-panel build",
     "build-internal-dependencies": "./scripts/bash/buildInternalDependencies.sh",
     "build-web-frontend": "yarn workspace @tupaia/web-frontend build",
+    "download-env-vars": "./scripts/bash/downloadEnvironmentVariables.sh",
```
edmofro, line 24: These now come from LastPass, which will make things _so_ much easier to manage!

codeship-steps.yml
```diff
@@ -1,123 +1,142 @@
...
 - type: serial
-  name: deploy-all
-  service: deployment
+  name: Build front ends
+  service: builder
   exclude: -e2e$
   steps:
-    - type: serial
-      name: deploy-backends
+    # front end builds are split into batches as they consume a lot of memory
```
edmofro, line 69: Despite having four cores on the machine now, and 7.5GB memory, the front ends still seemed slightly flakey when run in parallel. When https://github.com/beyondessential/tupaia/pull/3076 is merged I will try again with all four in parallel 🤞

packages/devops/scripts/ci/buildPackage.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -x
+
+DIR=$(dirname "$0")
+PACKAGE=$1
+${DIR}/copyFromCommonVolume.sh # pull in the .env files from common volume
```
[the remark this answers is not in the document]
IgorNadj: Very funky 🎷 

I haven't tried on my machine, but why copy at all? Generally you mount the volume and each container works on the files in place.

packages/devops/scripts/ci/checkDeploymentExists.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -x
+
+DIR=$(dirname "$0")
+DEPLOYMENT_SSH_URL=$(${DIR}/determineDeploymentSshUrl.sh)
+if curl --output /dev/null --silent --head --fail $DEPLOYMENT_SSH_URL; then
+  touch /common/deployment_exists
```
IgorNadj, line 7: Good idea!

packages/devops/scripts/ci/setupEnvironmentVariables.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -x
+DIR=$(dirname "$0")
+
+# packages needing .env during CI/CD are all deployable, plus auth and data-api
```
IgorNadj, line 5: Minor: move to a getPackagesWithEnvVars.sh?

e2e.env.encrypted
```diff
@@ -0,0 +1,2 @@
+codeship:v2
```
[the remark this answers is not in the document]
IgorNadj: Same with x.Dockerfiles. No reason to have them in root really. The only one that really belongs there is `Dockerfile`, representing the tupaia image if we went with a single fat image.

packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -6,36 +6,18 @@ pm2 list
 echo "Deleting..."
 pm2 delete all
 
-# Set the path of environment variables from parameter store
-if [[ $BRANCH == "master" ]]; then
-    ENVIRONMENT="production"
-else
-    # Any other branch uses the default dev environment variables
-    ENVIRONMENT="dev"
```
IgorNadj, line 14: We still need this, no? Feature branches should pull down the lastpass note for [Dev] right?

packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -67,7 +49,7 @@ for PACKAGE in ${PACKAGES[@]}; do
         yarn migrate
 
         # After running migrations it's good to ensure that the analytics table is fully built
-        yarn download-parameter-store-env-vars --package-name data-api --environment $ENVIRONMENT
+        yarn download-env-vars $BRANCH data-api
```
IgorNadj, line 52: See other note about extracting the list of packages that have .env vars out to a script. If we do that, we don't need a special case here for data-api, all packages would have .env vars added at the top of this script.

packages/web-frontend/package.json
```diff
@@ -12,9 +12,9 @@
   "main": "dist/index.js",
   "scripts": {
     "analyze": "source-map-explorer build/static/js/main.*",
-    "build": "mkdir -p builds && npm run build-desktop && npm run build-mobile",
-    "build-desktop": "GENERATE_SOURCEMAP=false react-scripts build && rm -rf builds/desktop && mv build builds/desktop",
-    "build-mobile": "GENERATE_SOURCEMAP=false REACT_APP_APP_TYPE=mobile react-scripts build && rm -rf builds/mobile && mv build builds/mobile",
+    "build": "mkdir -p builds && npm run build-desktop && mv build builds/desktop && npm run build-mobile && mv build builds/mobile && mv builds build",
```
IgorNadj, line 15: I think I get the idea behind this change. But I have a feeling this might cause issues if someone goes to rebuild and runs `yarn build-desktop && yarn build-mobile` instead of `build`, it would not actually work like they expect it to.

scripts/bash/downloadEnvironmentVariables.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+echo ${LASTPASS_PASSWORD} | LPASS_DISABLE_PINENTRY=1 lpass login ${LASTPASS_EMAIL}
+
+for PACKAGE in $PACKAGES; do
+    ENV_FILE_PATH=${DIR}/../../packages/${PACKAGE}/.env
+
+    # checkout branch specific env vars, or dev as fallback, temporarily redirecting stderr
+    exec 3> /dev/stderr 2> /dev/null
+    lpass show --notes ${PACKAGE}.${BRANCH}.env > ${ENV_FILE_PATH} || lpass show --notes ${PACKAGE}.dev.env > ${ENV_FILE_PATH}
```
IgorNadj, line 24: Minor: I suggest a new group in LastPass to keep these separate from normal tupaia credentials.

tupaia-ci-cd.Dockerfile
```diff
@@ -15,69 +20,47 @@ COPY package.json ./
...
 COPY babel.config.json ./
 COPY .babelrc-ts.js ./
+COPY tsconfig-js.json ./
 RUN mkdir ./scripts
 COPY scripts/. ./scripts
 
+RUN mkdir -p ./packages/devops/scripts/ci
+COPY packages/devops/scripts/ci/makePlaceholderPackageFolders.sh ./packages/devops/scripts/ci
+
+RUN ./packages/devops/scripts/ci/makePlaceholderPackageFolders.sh
```
IgorNadj, line 30: Minor: I like that this makes the Dockerfile shorter but it comes at the cost of no longer colocating the `mkdir` and `COPY`. My personal preference is colocation to make changes easier and a long Dockerfile, because it's not hiding the root cause, which is that we have a fat image. Fat images are bad (generally speaking in container land), and the solution is to split it out. Or if we decide to have a fat image we should embrace it :)

IgorNadj (commented):
Looking great! I like the idea of just zipping it up and shipping that way, it avoids the issues of baking env vars in docker images completely, and does what we need it to.

I have a few minor comments and a question about needing to copy from/to a common dir.

packages/devops/configs/servers.conf
```diff
@@ -351,37 +351,7 @@ server {
...
-        internal;
-  }
-
-}
-server {
-
-  listen       80;
-  listen       [::]:80;
-  server_name  ~^.*export\.tupaia\.org$;
-  root         /home/ubuntu/tupaia/packages/web-frontend/builds/exporter;
```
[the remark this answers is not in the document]
edmofro: Yeah, I should've added a note here but we haven't used the export specific site for a long time, so this is just clutter

packages/lesmis/package.json
```diff
@@ -31,8 +30,6 @@
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.57",
     "@material-ui/styles": "^4.9.10",
-    "@tupaia/access-policy": "3.0.0",
```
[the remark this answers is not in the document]
edmofro: Yeah I removed this from a couple of places because it wasn't being used inside the package, so I'm assuming it was a copy paste of another package.json when these got spun up

packages/psss/package.json
```diff
@@ -69,6 +66,8 @@
     "whatwg-fetch": "^3.0.0"
   },
   "devDependencies": {
+    "@tupaia/auth": "1.0.0",
```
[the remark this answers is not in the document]
edmofro: Yep this is only being used during cypress testing so no need for it to be a dependency proper

packages/web-config-server/src/tests/TestableApp.js
```diff
@@ -1,4 +1,3 @@
-import {} from 'dotenv/config'; // Load the environment variables into process.env
```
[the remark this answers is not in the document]
edmofro: Yeah, another one where I noticed a copy paste leaving remnants of the original - the import _is_ needed in meditrak-server, which is where this file was copied from, but isn't needed for any of the web-config-server tests

lastpass-env-vars.encrypted
```diff
@@ -0,0 +1,2 @@
+codeship:v2
```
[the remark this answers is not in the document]
edmofro: Oops, this one needs deleting since I changed the naming convention, thanks!

README.md
```diff
@@ -52,6 +52,10 @@ if you prefer to pre-build internal dependencies, add `--skip-internal` to the a
 
 Most packages will require a .env file. `.env.example` files indicate the required variables per package.
 
+🔑 **BES internal:** If you have access to the LastPass folder Tupaia Environment Variables, you can get most environment variables
```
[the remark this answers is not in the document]
edmofro: Yeah true, that's a good point. I'll remove it and add it to Slab

codeship-services.yml
```diff
@@ -6,51 +6,77 @@
 #   service (see below)
 # - 'cypress' service includes a base cypress container, plus code and node_module dependencies
 #   required by e2e tests in web-frontend
-# - the db service provides a postgres db with postgis (and takes a little while to run, thus the
+# - the 'db' service provides a postgres db with postgis (and takes a little while to run, thus the
```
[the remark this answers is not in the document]
edmofro: Ah yeah, good call!

e2e.env.encrypted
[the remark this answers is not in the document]
edmofro: Good point, I'll move these somewhere nicer

packages/devops/scripts/ci/setupEnvironmentVariables.sh
edmofro: Maaybe...there are other packages that can possibly take env vars I believe, but are only needed for some local things like cypress testing. I'll think on this.

packages/devops/scripts/deployment/deployPackages.sh
edmofro: Yeah but that's now built into `downloadEnvironmentVariables.sh` 🎉

packages/devops/scripts/deployment/deployPackages.sh
edmofro: True. Ok getting closer to sold on this!

tupaia-ci-cd.Dockerfile
edmofro: Yeah no worries, I'm happy to change this back. There was another use for the util at one point during development but that disappeared, so there's less justification for it now

scripts/bash/downloadEnvironmentVariables.sh
edmofro: Already how it's done ;-)

rego@beyondessential.com.au _only_ has access to "Tupaia Environment Variables", which is separate from our main shared folder

README.md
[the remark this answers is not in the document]
edmofro: https://beyond-essential.slab.com/posts/tupaia-monorepo-setup-v5egpdpq#step-3-add-env-files and https://beyond-essential.slab.com/posts/ci-cd-using-codeship-uzxspw8z

packages/web-frontend/package.json
edmofro: Yeah that's true that that case wouldn't work, but I guess I'm trying to make _most_ cases more expected (at the sacrifice of this one). So I can imagine that locally, people would generally build one or the other depending on what they were working on, and the result would be in the conventional `build` folder now. Gonna leave this one as is!

edmofro:
For some reason I can't respond to your main question inline, so quote replying this one:

> I haven't tried on my machine, but why copy at all? Generally you mount the volume and each container works on the files in place.
> 
> E.g.
> 
> /home/tupaia/packages/...
> docker run lastpass-add-env-vars
> /home/tupaia/packages/*/.env
> docker run builder
> /home/tupaia/packages/*/.env,dist
> docker run pusher
> /home/tupaia/packages,/home/built.zip

This would be so much simpler! Unfortunately I don't think it's possible to get this _and_ have `yarn install` and `yarn build-internal-dependencies` cached in the docker image layers, because any mounted volume _isn't_ available at build time

packages/devops/scripts/ci/buildPackage.sh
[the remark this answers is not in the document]
IgorNadj: Convo continues below

IgorNadj:
> This would be so much simpler! Unfortunately I don't think it's possible to get this _and_ have `yarn install` and `yarn build-internal-dependencies` cached in the docker image layers, because any mounted volume _isn't_ available at build time

Ah yep. I've thought this through and the issue goes back to our env vars being needed at build time. 

There may be other workarounds (one I considered is moving the lastpass image to working at build time, and using Codeship's depends_on to give us an order of operations kind of thing) but they are all workarounds, so none seem better than any other. But ultimately the issue is of baking env vars. In the future we should clean this up. Happy to approve as is, this solution looks fit for purpose!

IgorNadj (approved):
Looks great! :shipit:

IgorNadj:
Heya just saw a few more commits come through, not sure if they were intended to be part of this PR but this caught my eye anyway: https://github.com/beyondessential/tupaia/pull/3121/commits/13f3670ddc4074f759db3ce997d928e3d937d554

Minor note on env vars: My thinking on env vars is that in the future if/when each service is a container they will need to be defined for each one anyway. 

And I saw a change to @tupaia/api-client, just a couple of notes:
1. Be careful, @tuapaia/api-client is following SemVer and that commit has a breaking change, bump it to 2.0.0 instead. We might need to unpublish that patch update so Tamanu doesn't fall over. It's not currently deployed so it's not urgent but let me know what we end up doing with this before we merge that feature into Tamanu.
2. Defaulting endpoint urls to production is a little bit dangerous

edmofro:
Yep @IgorNadj you're already onto it haha

On the env vars - there's still the same ability to override using environment variables that we can make use of in any future containerised world, this just adds sensible defaulting which allows us to simplify our .env files in the current context, and avoid going overboard with defining them when we shouldn't really need to. For example, the entity-server currently requires the URL of the report-server to be defined even though it never makes use of the report-server, so it's nice if we don't have to think about how it should be defined.

On your point about versioning: great pickup. Is Tamanu relying on those constants being exported? I didn't realise that was part of the expected interface of the package, thanks for pointing it out.

IgorNadj:
Sounds good. Could you add back in a way of getting localhost|dev|prod urls explicitly somehow, tamanu needs those to configure different environments. It was using LOCALHOST_ENDPOINT_BASE_URLS etc. before, it needs something similar with the new way.

edmofro:
Ah right, sure thing :+1:

edmofro:
(also fyi, I hadn't published that patch so Tamanu shouldn't fall over from that)

edmofro:
@IgorNadj I was battling with this yesterday, then today it worked fine! I realised it was because:
- the old `@tupaia/api-client` module was sitting on the machine in `node_modules` from the start of the day, when that was the one being used
- the new version of the code just uses the mono-repo link to use `@tupaia/api-client` in place, rather than pulling one from npm to stick in `node_modules`
- counter to my intuition, `tar` doesn't overwrite the whole directly when it extracts the new one
- thus the old `@tupaia/api-client` was still in `node_modules` and being picked up when running, rather than using the `2.0.1` version in place

So the overnight wipe and reclone meant that `@tupaia/api-client` wasn't in `node_modules` today, which is why it worked fine, but I've also made it replace the entire directory on push now so unexpected things like this don't happen in future. Good lesson, glad it happened before hitting production!

Now ready for re-review

IgorNadj (approved):
LGTM!

PR #3124, opened 2021-08-30
SOL-395: Filter out answers for non-exported questions

edmofro:
### Issue #:
SOL-395

Hot fix as it is breaking downloads in prod

### Changes:
My changes for long running exports had not properly filtered out non-exported answers, which was causing a bug as it didn't know where on the spreadsheet it should go.


kael89 (approved):

edmofro:
Merging directly to prod, I feel it's safer than what's there at the moment - testing can take place post-release.

PR #3129, opened 2021-08-31
NZ-43 Add header to no data tooltip

EMcQ-BES:
### Changes:

- Add `NoDataTooltip` to display the series name even when there's no data

---

### Screenshots:
See ticket


packages/ui-components/src/components/Chart/Tooltip.js
```diff
@@ -132,6 +132,23 @@ const SingleValueTooltip = ({ valueType, payload, periodGranularity, labelType }
   );
 };
 
+const NoDataTooltip = ({ payload, periodGranularity }) => {
+  const data = payload[0]?.payload;
+  const { name = undefined, timestamp = undefined } = data || {};
```
EMcQ-BES, line 137: This is also where we wind up as data is loading, so need to handle the possibility of literally everything being undefined

tcaiger: I think another option would be to return null if data is undefined but overall it's much of a muchness.

tcaiger (approved):
Looks good to me. It's good to have this fixed. I left a comment but it's not particularly important.

PR #3131, opened 2021-08-31
Add entities relations for penfaa_samoa

billli0:
### Issue #:
https://linear.app/bes/issue/SOL-402/add-entities-for-samoa-ncd-screening-project
### Changes:
A migration to add entity relation for penfaa_samoa


packages/database/src/migrations/20210831002757-AddEntitiesRelationToSamoaForPenfaaSamoa-modifies-data.js
```diff
@@ -0,0 +1,57 @@
...
+  const record = await db.runSql(`SELECT id FROM entity_hierarchy WHERE name = '${name}'`);
+  return record?.rows[0].id;
+};
+
+exports.up = async function (db) {
+  const entityHierarchyId = generateId();
+  await insertObject(db, 'entity_hierarchy', {
+    id: entityHierarchyId,
+    name: projectCode,
+    canonical_types: '{country,district,sub_district,school}',
```
nick-beyondessential, line 32: Not sure `sub_district` should be a canonical type here. They currently only exist in covid_samoa hierarchy (no parent_id) as defined in `entity_relation` table. So perhaps better to remove `sub_district` and define entity_relation rows for each sub_district with newly created `entityHierarchyId`.
I'm assuming we don't want the 'Unknown' entities in this hierarchy (something to check with @julianam-w ) so they would be not be added. If `Unknown Island` is not needed then `district` can remain canonical (else define similar to `covid_samoa` heirarchy.

nick-beyondessential: as a suggestion something like:
```
select * from entity_relation er 
join entity e on er.child_id = e.id and e."type" = 'sub_district'
where er.entity_hierarchy_id in (select eh.id from entity_hierarchy eh where eh."name" = 'covid_samoa');
```

julianam-w: Please exclude all "Unknown" (district, sub_district, facility, village, household)

nick-beyondessential: So leaving `district` as canonical will exclude Unknown Island (as it has no parent_id on entity. The above query will exclude 'Unknown District' and facility, village, and household are not used in this hierarchy.

nick-beyondessential (changes requested):
Sub_districts need entity_relations added.

packages/database/src/migrations/20210831002757-AddEntitiesRelationToSamoaForPenfaaSamoa-modifies-data.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  Promise.all(
+    samoaHierarchyData.map(async ({ parent_code: parentCode, code }) => {
+      const parentId = await codeToId(db, 'entity', parentCode);
+      const childId = await codeToId(db, 'entity', code);
+      await insertObject(db, 'entity_relation', {
```
nick-beyondessential, line 39: So to record our conversation - schools have already been imported with sub_district parents - probably best to leave school as canonical (not create entity_relations) so that can be updated by admin panel.

nick-beyondessential: and maybe double check that schools don't show up in covid_samoa :)

billli0: Checked that schools are not showing in my local, but will let testers to check that as well.

billli0:
Thanks @nick-beyondessential for clarifying the hierarchy logic again!

nick-beyondessential (approved):
LGTM!

billli0:
Closed because changes apply in different PR: https://github.com/beyondessential/tupaia/pull/3136

PR #3137, opened 2021-08-31
SOL-400: Add ability to delete dashboardItems from admin panel

rohan-bes:
### Issue SOL-400:

Will also delete the associated report

[ ] - Release task: @rohan-bes Delete the Palau PSSS dashboard items that were deprecated by: SOL-398

### DIFF
--- a/packages/meditrak-server/src/apiV2/dashboardItems/DeleteDashboardItem.js
+++ b/packages/meditrak-server/src/apiV2/dashboardItems/DeleteDashboardItem.js
@@ -15,4 +15,13 @@ export class DeleteDashboardItem extends DeleteHandler {
       assertAnyPermissions([assertBESAdminAccess, dashboardItemChecker]),
     );
   }
+
+  async deleteRecord() {
+    const dashboardItem = await this.resourceModel.findById(this.recordId);
+    if (dashboardItem.report_code) {
+      const reportModel = dashboardItem.legacy ? this.models.legacyReport : this.models.report;
+      await reportModel.delete({ code: dashboardItem.report_code });
+    }
+    return super.deleteRecord();
+  }
 }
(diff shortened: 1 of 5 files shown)


packages/meditrak-server/src/apiV2/dashboardItems/DeleteDashboardItem.js
```diff
@@ -15,4 +15,13 @@ export class DeleteDashboardItem extends DeleteHandler {
       assertAnyPermissions([assertBESAdminAccess, dashboardItemChecker]),
     );
   }
+
+  async deleteRecord() {
+    const dashboardItem = await this.resourceModel.findById(this.recordId);
+    if (dashboardItem.report_code) {
+      const reportModel = dashboardItem.legacy ? this.models.legacyReport : this.models.report;
+      await reportModel.delete({ code: dashboardItem.report_code });
+    }
```
billli0, line 24: Love that you want to take care of report and legacy_report table👍 Something could be out of scope but is there chance one report is using by more than one dashboard items in the future?

rohan-bes: > is there chance one report is using by more than one dashboard items in the future?

I'm inclined to say no. Both because it seems we've fallen pretty heavily in the camp of having a 1-to-1 mapping between dashboard items and reports, and because we actually give them the same `code` currently, and since `code` is a unique field for both tables, we'd don't really support multiple dashboard items sharing the same report.

It's a good question though... Arguably we should combine the report config and the dashboard item config in the one object rather than referencing across tables, but that's a discussion for another time...

billli0: Yeah dashboard item now needs to handle the legacy report as well, so that's quite hard to combine.

PR #3138, opened 2021-08-31
Sol 402 add sub district parents ws

billli0:
### Changes:
Add parent id to samoa sub_district entities


packages/database/src/migrations/20210831074805-AddSubDistrictParentsSamoa-modifies-data.js
```diff
@@ -0,0 +1,48 @@
...
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const getEntityRelations = async db => {
+  const record = await db.runSql(`
+    SELECT er.parent_id, er.child_id FROM entity_relation er 
+    JOIN entity e ON er.child_id = e.id AND e."type" = 'sub_district' AND e.name not like 'Unknown%'
+    WHERE er.entity_hierarchy_id in (SELECT eh.id from entity_hierarchy eh where eh."name" = 'covid_samoa');
```
nick-beyondessential, line 23:
```suggestion
    WHERE er.entity_hierarchy_id in (SELECT eh.id from entity_hierarchy eh where eh."name" = 'penfaa_samoa'); 
```

packages/database/src/migrations/20210831074805-AddSubDistrictParentsSamoa-modifies-data.js
```diff
@@ -0,0 +1,48 @@
...
+    WHERE er.entity_hierarchy_id in (SELECT eh.id from entity_hierarchy eh where eh."name" = 'covid_samoa'); 
+  `);
+  return record?.rows;
+};
+
+exports.up = async function (db) {
+  const samoaHierarchyRelations = await getEntityRelations(db);
+  Promise.all(
+    samoaHierarchyRelations.map(async ({ parent_id: parentId, child_id: childId }) => {
+      await updateValues(db, 'entity', { parent_id: parentId }, { id: childId });
```
nick-beyondessential, line 32: So this will add parent_ids to sub_districts (which will mess up explore.) Weren't we going to set the school type entities parent_id to district (parent of sub_districts) and leave sub_district parent_id null?
And should also then recreate the sub_district -> school entity_relations.

rohan-bes, line 32: Sorry to jump in on this one, but shouldn't this be:
```suggestion
      await updateValues(db, 'entity', { parent_id: parentId }, { parent_id: childId });
```

The goal is to set the schools (currently children of sub_districts) to be the children of their sub_district's parent.
The first fetch gets us all sub_district->parent relations, now we want to set all children of each sub_district, to have that parent?

nick-beyondessential (approved):
LGTM

packages/database/src/migrations/20210831074805-AddSubDistrictParentsSamoa-modifies-data.js
```diff
@@ -0,0 +1,76 @@
...
+const hierarchyNameToId = async (db, name) => {
+  const record = await db.runSql(`SELECT id FROM entity_hierarchy WHERE name = '${name}'`);
+  return record?.rows[0].id;
+};
+
+exports.up = async function (db) {
+  const entityHierarchyId = await hierarchyNameToId(db, projectCode);
+  const subDistrictEntityRelations = await getSubDistrictEntityRelations(db);
+
+  const subDistrictToDistrictMapping = Object.fromEntries(
```
rohan-bes, line 47: Wow I never knew `Object.fromEntries` exists. I always do an array reduce for this, but this reads a lot nicer 👍

billli0: Learn this from @kael89 😬

PR #3143, opened 2021-09-01
Fix boolean label within JsonInputField

tcaiger:
### Changes:

- Fix label z-index issue. I do not fully understand why but the legend is getting positioned behind other layers and position relative brings it into the right pane. I tested on chroms, firefox and safari and works as expected.

---

### Screenshots:
See issue


edmofro (approved):
Huh...nicely fixed, I had a brief struggle with this one before giving up and making the card, and certainly did not try that!

PR #3149, opened 2021-09-02
Hotfix: Update UNFPA Raw Data Download Survey for SB

kael89:
### Issue #:
Addresses https://beyondessential.slack.com/archives/GAK5V0HHQ/p1630556369013300

Config is generally copied from the existing `UNFPA_Raw_Data_Reproductive_Health_Facility` report, then changed the survey code and name to use the SB-specific version of the `RHFSC` survey: `RHFSCSI`.

Edit: I also updated the existing `UNFPA_Raw_Data_Reproductive_Health_Facility` report, since we temporarily enabled both `RHFSC` and `RHFSCSI` there, but we only really need the first (which is the original config)

### DIFF
--- a/packages/database/src/migrations/20210902055844-UpdateUNFPASbRawDataDownloadSurvey-modifies-data.js
+++ b/packages/database/src/migrations/20210902055844-UpdateUNFPASbRawDataDownloadSurvey-modifies-data.js
@@ -1,14 +1,10 @@
 'use strict';
 
-import { deleteObject, findSingleRecordBySql, insertObject, generateId } from '../utilities';
+import { deleteObject, insertObject, generateId, updateValues } from '../utilities';
 
 var dbm;
 var type;
 var seed;
-
-const OLD_REPORT_CODE = 'UNFPA_Raw_Data_Reproductive_Health_Facility';
-const NEW_REPORT_CODE = 'UNFPA_Raw_Data_Reproductive_Health_Facility_SB';
-const DASHBOARD_CODE = 'SB_UNFPA_Raw Data Downloads';
 
 /**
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20210902055844-UpdateUNFPASbRawDataDownloadSurvey-modifies-data.js
```diff
@@ -0,0 +1,103 @@
...
+      transformations: [{ type: 'transposeMatrix' }],
+    },
+  },
+});
+
+const swapDashboardItemInRelation = async (db, dashboardCode, oldItemCode, newItemCode) =>
+  db.runSql(`
+    UPDATE dashboard_relation
+    SET child_id = new_di.id
+    FROM dashboard d, dashboard_item old_di, dashboard_item new_di
```
rohan-bes, line 41: Huh... I never knew you could update records from the result set of a join! Nice one!!

rohan-bes (approved):
LGTM! Bit concerned about what @edmofro said about this download hanging meditrak-server on prod when he tested it. I feel like it ought to be a coincidence, but it is probs worth testing?

kael89:
> LGTM! Bit concerned about what @edmofro said about this download hanging meditrak-server on prod when he tested it. I feel like it ought to be a coincidence, but it is probs worth testing?

Thanks for double checking @rohan-bes - in the meantime I managed to use those raw data downloads fine in prod, so I guess it was a coincidence it the end...

(The change here is to enable one survey or the other, depending on the UNFPA country you have selected - currently, our workaround is to display both of them in all UNFPA countries 🙂 )

kael89:
Rebased off the most recent master, for cleaner commit history

PR #3150, opened 2021-09-02
Added 'single' aggregation function to report-server

rohan-bes:
Throws an error is there's more than 1 value per group


packages/report-server/src/reportBuilder/transform/functions/aggregate/aggregations.ts
```diff
@@ -95,6 +95,21 @@ const last = (values: FieldValue[]): FieldValue => {
   return values.slice().reverse().find(isNotUndefined);
 };
 
+const single = (values: FieldValue[]): FieldValue => {
```
rohan-bes, line 98: I was pretty temped to call this `only` instead. I think it really depends on the order of the config. Since we go `column: aggregation` I preferred `BCD1: single` over `BCD1: only`

billli0 (approved):
LGTM! It makes sense to me from user prospective.

PR #3158, opened 2021-09-03
[HotFix] Fix covid-samoa household overlay

nick-beyondessential:
Hotfix

- Unvaccinated households down to 3k which is acceptable to reinstate the household overlay while a better performance enhancement goes through usual test cycle.
- Re-add point data correcting original import mistake

Also includes https://github.com/beyondessential/tupaia/pull/3149

### DIFF
--- a/packages/database/src/migrations/20210804054649-CovidSamoaReinstateHouseholdOverlay-modifies-data.js
+++ b/packages/database/src/migrations/20210804054649-CovidSamoaReinstateHouseholdOverlay-modifies-data.js
@@ -1,0 +1,58 @@
+'use strict';
+
+import { codeToId, insertObject, generateId } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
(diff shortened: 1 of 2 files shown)


packages/database/src/migrations/20210804054649-CovidSamoaReinstateHouseholdOverlay-modifies-data.js
```diff
@@ -0,0 +1,58 @@
...
+  await insertObject(db, 'map_overlay_group_relation', mapOverlayGroupRelation(mapOverlayGroupId));
+  await db.runSql(`
+    update report
+        set config = jsonb_set(config, '{transform,1,where}', '"notEq($row.HVS_covidstatus, ''Fully vaccinated'')"')
+        where code = '${reportCode}';
+  `);
+  // add fetchOnLevel config
+  await db.runSql(`
+    update "mapOverlay" 
+      set "presentationOptions" = "presentationOptions" || '{"fetchOnlyDisplayLevel": true}'
```
rohan-bes, line 41: Should we maybe not do the `fetchOnLevel` stuff here and save that for when we actually implement the performance stuff?

rohan-bes (approved):
LGTM! Just recommend we backout the `fetchOnLevel` stuff for now 👍 Other than that good to go!

packages/database/src/migrations/20210804054649-CovidSamoaReinstateHouseholdOverlay-modifies-data.js
nick-beyondessential: Good call! I just missed taking it out.

PR #3164, opened 2021-09-06
MEL-454: Viz builder export improvements

kael89:
### Issue #:

MEL-454

### DIFF
--- a/packages/admin-panel-server/src/routes/ExportDashboardVisualisationRoute.ts
+++ b/packages/admin-panel-server/src/routes/ExportDashboardVisualisationRoute.ts
@@ -4,30 +4,151 @@
  *
  */
 
-import { Request } from 'express';
+import { Request, Response, NextFunction } from 'express';
+import { keyBy } from 'lodash';
 
+import { camelKeys } from '@tupaia/utils';
 import { Route } from '@tupaia/server-boilerplate';
-
-import { DashboardVisualisationObject } from '../viz-builder';
+import { MeditrakConnection } from '../connections';
+import { DashboardVisualisationCombiner, DashboardVisualisationObject } from '../viz-builder';
+import {
+  Dashboard,
+  DashboardRecord,
+  DashboardItemRecord,
+  DashboardRelationObject,
+  DashboardRelationRecord,
+} from '../viz-builder/types';
 
 export type ExportDashboardVisualisationRequest = Request<
-  Record<string, never>,
+  { dashboardVisualisationId?: string },
   { contents: DashboardVisualisationObject; filePath: string; type: string },
   { visualisation: DashboardVisualisationObject },
-  Record<string, never>
+  Record<string, any>
 >;
 
 export class ExportDashboardVisualisationRoute extends Route<ExportDashboardVisualisationRequest> {
   protected readonly type = 'download';
 
+  private readonly meditrakConnection: MeditrakConnection;
+
+  constructor(req: ExportDashboardVisualisationRequest, res: Response, next: NextFunction) {
+    super(req, res, next);
+
+    this.meditrakConnection = new MeditrakConnection(req.session);
+  }
+
   async buildResponse() {
-    const { visualisation } = this.req.body;
+    this.validate();
+
+    const { visualisation: builtVisualisation } = this.req.body;
+
+    const dashboardItem = await this.findExistingDashboardItem();
+    const visualisation =
+      builtVisualisation || (await this.buildDashboardItemVisualisation(dashboardItem));
     const fileBaseName = visualisation.code || 'new_dashboard_visualisation';
 
+    const { id, ...visualisationWithoutId } = visualisation;
+    const { dashboards, dashboardRelations } = await this.buildDashboardsAndRelations(
+      dashboardItem,
+    );
+
     return {
-      contents: visualisation,
+      contents: {
+        ...visualisationWithoutId,
+        dashboards,
+        dashboardRelations,
+      },
       filePath: `${fileBaseName}.json`,
       type: '.json',
     };
   }
+
+  private validate = () => {
+    const { dashboardVisualisationId } = this.req.params;
+    const { visualisation } = this.req.body;
+
+    if (!dashboardVisualisationId && !visualisation) {
+      throw new Error('You must provide either a visualisation id or a visualisation object');
+    }
+    if (dashboardVisualisationId && visualisation) {
+      throw new Error(
+        'Both visualisation id and visualisation object provided, must provide one or the other',
+      );
+    }
+    if (visualisation && !visualisation.code) {
+      throw new Error('Visualisation object must have a code');
+    }
+  };
+
+  private findExistingDashboardItem = async (): Promise<DashboardItemRecord> => {
+    const { dashboardVisualisationId } = this.req.params;
+
+    if (dashboardVisualisationId) {
+      const dashboardItem = await this.meditrakConnection.fetchResources(
+        `dashboardItems/${dashboardVisualisationId}`,
+      );
+      if (!dashboardItem) {
+        // We assert that the record exists, since a specific id was provided
+        throw new Error(`Could not find visualisation with id ${dashboardVisualisationId}`);
+      }
+      return dashboardItem;
+    }
+
+    const { code } = this.req.body.visualisation;
+    const [dashboardItem] = await this.meditrakConnection.fetchResources('dashboardItems', {
+      filter: {
+        code,
+      },
+    });
+    return dashboardItem;
+  };
+
+  private buildDashboardItemVisualisation = async (dashboardItem: DashboardItemRecord) => {
+    if (dashboardItem.legacy) {
+      return dashboardItem;
+    }
+
+    const visualisation = await this.meditrakConnection.fetchResources(
+      `dashboardVisualisations/${dashboardItem.id}`,
+    );
+    const combiner = new DashboardVisualisationCombiner(
+      visualisation.dashboardItem,
+      visualisation.report,
+    );
+    return combiner.getVisualisation();
+  };
+
+  private buildDashboardsAndRelations = async (dashboardItem: DashboardItemRecord) => {
+    if (!dashboardItem) {
+      return { dashboards: [], dashboardRelations: [] };
+    }
+
+    const relationRecords: DashboardRelationRecord[] = await this.meditrakConnection.fetchResources(
+      `dashboardItems/${dashboardItem.id}/dashboardRelations`,
+    );
+    const dashboardIds = relationRecords.map(dr => dr.dashboard_id);
+    const dashboardRecords: DashboardRecord[] = await this.meditrakConnection.fetchResources(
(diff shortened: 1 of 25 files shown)


packages/admin-panel/src/api/TupaiaApi.js
```diff
@@ -68,10 +65,12 @@ export class TupaiaApi {
...
-    const contentType = response.headers.get('content-type');
-    if (contentType.startsWith('application/json')) {
-      return extractHeadersAndBody(response);
+    // Check if this is an early response indicating it will be emailed
+    if (isJsonResponse(response)) {
+      const body = await response.clone().json();
+      if (body.emailTimeoutHit) {
+        return { headers: response.headers, body };
+      }
     }
```
kael89, line 74: @edmofro I updated this to allow `json` file downloads

packages/database/src/TupaiaDatabase.js
```diff
@@ -465,12 +465,22 @@ function buildQuery(connection, queryConfig, where = {}, options = {}) {
...
       if (typeof columnSpec === 'string') return columnSpec;
       const [alias, selector] = Object.entries(columnSpec)[0];
+
+      if (selector?.castAs) {
+        const { castAs } = selector;
+        if (!VALID_CAST_TYPES.includes(castAs)) {
+          throw new Error(`Cannot cast as ${castAs}`);
+        }
+        return { [alias]: connection.raw(`??::${castAs}`, [alias]) };
+      }
```
kael89, line 475: @edmofro you may want to have a look at the changes in this file since you are the one with the best knowledge of the core DB classes

packages/meditrak-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -44,15 +43,15 @@ export const generateLinkHeader = (resource, pageString, lastPage, originalQuery
...
-  return `${baseRecordType}.${unprocessedColumnSelector}`;
+  const [resource, column] = unprocessedColumnSelector.includes('.')
+    ? unprocessedColumnSelector.split('.')
+    : [baseRecordType, unprocessedColumnSelector];
+
+  const recordType = resourceToRecordType(resource);
+  const model = models.getModelForDatabaseType(recordType);
+  const customSelector = model?.customColumnSelectors?.[column];
+  const selector = `${recordType}.${column}`;
+  return customSelector ? customSelector(selector) : selector;
```
kael89, line 54: This is a core change - it allows us to use `customColumnSelectors` regardless of whether the provided column specified a model name or not

packages/meditrak-server/src/apiV2/dashboardItems/assertDashboardItemsPermissions.js
```diff
@@ -80,11 +82,11 @@ export const createDashboardItemsDBFilter = async (accessPolicy, models, criteri
...
-  // then pull the corresponding dashboard items
-  const dashboardRelationsFilter = createDashboardRelationsDBFilter(accessPolicy);
-  const permittedDashboardRelations = await models.dashboardRelation.find(dashboardRelationsFilter);
-  const permittedDashboardItems = await models.dashboardItem.find({
-    id: permittedDashboardRelations.map(dr => dr.child_id),
+  // then pull the corresponding dashboards
+  const permittedRelationConditions = createDashboardRelationsDBFilter(accessPolicy, criteria);
+  const permittedDashboardItems = await models.dashboardItem.find(permittedRelationConditions, {
+    joinWith: TYPES.DASHBOARD_RELATION,
+    joinCondition: ['dashboard_relation.child_id', 'dashboard_item.id'],
```
kael89, line 89: Small refactoring here, to use one join query instead of two

rohan-bes (commented):
Awesome work @kael89! Some big challenges with this PR and some solid solutions!

Mainly a bit concerned about unexpected side-effects when importing viz with insufficient permissions atm, however was also hoping we could walk-n-talk with this one so I could get a better idea of what's going on with the database changes?

packages/admin-panel-server/src/routes/ExportDashboardVisualisationRoute.ts
```diff
@@ -4,30 +4,144 @@
...
+        ...visualisationWithoutId,
+        dashboards,
+        dashboardRelations,
+      },
       filePath: `${fileBaseName}.json`,
       type: '.json',
     };
   }
+
+  private validate = () => {
```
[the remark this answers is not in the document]
kael89: Yeap I agree - sounds like a proper OOP design!

packages/database/src/TupaiaDatabase.js
edmofro: Looks good!

packages/meditrak-server/src/apiV2/GETHandler/helpers.js
edmofro: Nice cleanup too! The ternary with destructuring into array works really nicely

packages/admin-panel-server/src/routes/ImportDashboardVisualisationRoute.ts
```diff
@@ -38,25 +73,36 @@ export class ImportDashboardVisualisationRoute extends Route<ImportDashboardVisu
...
-    fs.unlinkSync(this.req.file.path);
+    const { dashboards, dashboardRelations, ...visualisation } = this.readFileContents();
 
     const extractor = new DashboardVisualisationExtractor(
       visualisation,
       draftDashboardItemValidator,
       draftReportValidator,
     );
-    const body = extractor.getDashboardVisualisationResource();
+    const extractedViz = extractor.getDashboardVisualisationResource();
```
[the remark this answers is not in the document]
kael89: I am totally with you here... I generally sticked to `visualisation` for consistency, since it was already used in most cases when I jumped into the project, but I would prefer if we used `viz` exclusively and introduce it as a term in our codebase.

Not sure if it's worth refactoring now... But I would like us to promote the idea that using well thought after and established abbreviations makes code more readable! How would our world be if we had to write `findByIdentifier`, `dashboardIdentifier` instead of `findById`, `dashboardId`... 😄

packages/admin-panel-server/src/routes/ImportDashboardVisualisationRoute.ts
```diff
@@ -86,4 +132,60 @@ export class ImportDashboardVisualisationRoute extends Route<ImportDashboardVisu
...
+      assert.ok(dashboardId, `Could not find id for dashboard with code ${dashboardCode}`);
+      return snakeKeys({
+        ...dashboardRelation,
+        dashboard_id: dashboardId,
+        child_id: vizId,
+      });
+    });
+
+    await this.deleteDashboardItemRelations(vizId);
+    await this.createDashboardRelations(relationsToCreate);
```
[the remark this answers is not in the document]
kael89: Oh man that is a really good point... Not something that would happen often, but something that can lead in a really unpleasant UX (and an incomplete DB state) when it does! Yeap, let's brainstorm approaches to tackle this

packages/admin-panel-server/src/viz-builder/dashboardVisualisation/validators.ts
```diff
@@ -44,3 +44,18 @@ export const draftDashboardItemValidator = yup.object().shape({
...
+  sortOrder: yup.number(),
+});
+
+export const dashboardRelationObjectSchema = yup.object().shape({
+  dashboardCode: yup.string().required(),
+  entityTypes: yup.array().of(yup.string()).required(),
+  projectCodes: yup.array().of(yup.string()).required(),
+  permissionGroups: yup.array().of(yup.string()).required(),
+  sortOrder: yup.number(),
+});
```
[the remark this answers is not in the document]
kael89: Need to get this on a T-shirt!

packages/admin-panel-server/src/routes/ImportDashboardVisualisationRoute.ts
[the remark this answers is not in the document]
edmofro: Definitely agree with this too, and the extra justification is that `viz` is also what the business (i.e. Erin) uses, so it's not just some software dev abbreviation but still fits with domain driven design :-)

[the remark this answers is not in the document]
kael89: Domain driven design ftw!!

packages/admin-panel-server/src/routes/ImportDashboardVisualisationRoute.ts
```diff
@@ -4,18 +4,36 @@
...
+  Dashboard,
+  DashboardRecord,
+  DashboardRelationObject,
+  DashboardRelationRecord,
+} from '../viz-builder/types';
+
+const importFileSchema = yup.object().shape(
+  {
+    dashboards: yup.array().of(dashboardSchema),
+    dashboardRelations: yup.array().of(dashboardRelationObjectSchema),
```
kael89, line 32: @rohan-bes I decided to simplify things here, and allow dashboards and relations to be specified on their own. I suspect that most users will just export/import using the admin panel, but "power" users will be able to do some sophisticated operations by updating the import file, eg

1. A user wants to update a viz and its dashboard, and migrate the updates across instances
2. The user has created some new temporary relations in instance A (don't ask me why 😄 ), and doesn't want to copy them across instance B
3. The solution: delete `dashboardRelations` from the import file

Another example:
1. The user exports a viz, including a new dashboard
2. After the fact, they realize they also want to also attach the viz to an existing dashboard
3. They can simply add a new relation entry in the config file

Users will be warned by error messages in case any required key is missing

packages/admin-panel-server/src/routes/ImportDashboardVisualisationRoute.ts
```diff
@@ -70,12 +97,17 @@ export class ImportDashboardVisualisationRoute extends Route<ImportDashboardVisu
   };
 
   private createVisualisation = async (visualisation: Record<string, unknown>) => {
-    const { id } = await this.meditrakConnection.createResource(
-      'dashboardVisualisations',
-      {},
-      visualisation,
-    );
-    return id;
+    await this.meditrakConnection.createResource('dashboardVisualisations', {}, visualisation);
```
kael89, line 100: It's a bit inconvenient that our meditrak-server mutation endpoints do not return the affected resources...

packages/admin-panel-server/src/routes/ImportDashboardVisualisationRoute.ts
rohan-bes: @kael89 I like this! Usually I just create one or two relations when testing in local, even if the dashboard item requires 10+ in deployment. So being able to add/remove relations from the file is very handy!

rohan-bes (approved):
Looks good to me! Super excited for this! Thanks for making the logic a bit safer 🎉

kael89:
@rohan-bes I squashed all commits in one and branched off the latest `dev`, because the original base branch for this PR is now squashed in `dev`

PR #3166, opened 2021-09-06
Fix refresh table script

IgorNadj:
Looks like this was accidentally broken in afbe089e445a98ee47081d018675086c2776e68f

### DIFF
--- a/packages/data-api/package.json
+++ b/packages/data-api/package.json
@@ -20,8 +20,7 @@
     "uninstall-mv-refresh": "scripts/uninstallMvRefreshModule.sh",
     "build-analytics-table": "scripts/buildAnalyticsMaterializedView.sh",
     "drop-analytics-table": "scripts/dropAnalyticsMaterializedView.sh",
-    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh",
-    "refresh-analytics-table:full": "scripts/refreshAnalyticsTable.sh --full"
+    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh"
   },
   "dependencies": {
     "@tupaia/database": "1.0.0",
--- a/packages/data-api/package.json
+++ b/packages/data-api/package.json
@@ -20,7 +20,8 @@
     "uninstall-mv-refresh": "scripts/uninstallMvRefreshModule.sh",
     "build-analytics-table": "scripts/buildAnalyticsMaterializedView.sh",
     "drop-analytics-table": "scripts/dropAnalyticsMaterializedView.sh",
-    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh"
+    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh",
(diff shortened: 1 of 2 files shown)


packages/data-api/package.json
```diff
@@ -20,7 +20,8 @@
     "uninstall-mv-refresh": "scripts/uninstallMvRefreshModule.sh",
     "build-analytics-table": "scripts/buildAnalyticsMaterializedView.sh",
     "drop-analytics-table": "scripts/dropAnalyticsMaterializedView.sh",
-    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh"
+    "refresh-analytics-table": "scripts/refreshAnalyticsTable.sh",
+    "refresh-analytics-table:full": "scripts/refreshAnalyticsTable.sh --full"
```
rohan-bes, line 24: Is there a specific reason to add this rather than just use `yarn workspace @tupaia/data-api refresh-analytics-table --full`?

packages/data-api/scripts/fullRefreshAnalyticsTable.sh
```diff
@@ -8,6 +8,6 @@ source .env
 
 cd scripts
 export PGPASSWORD=$DB_PASSWORD
-psql -p $DB_PORT -h $DB_URL -d $ $DB_USER -tc "SELECT drop_analytics_table_indexes();"
-psql -p $DB_PORT -h $DB_URL -d $ $DB_USER -tc "SELECT mv\$refreshMaterializedView('analytics', 'public');"
-psql -p $DB_PORT -h $DB_URL -d $ $DB_USER -tc "SELECT create_analytics_table_indexes();"
+psql -p $DB_PORT -h $DB_URL -d $DB_NAME -U $DB_USER -tc "SELECT drop_analytics_table_indexes();"
+psql -p $DB_PORT -h $DB_URL -d $DB_NAME -U $DB_USER -tc "SELECT mv\$refreshMaterializedView('analytics', 'public');"
+psql -p $DB_PORT -h $DB_URL -d $DB_NAME -U $DB_USER -tc "SELECT create_analytics_table_indexes();"
```
rohan-bes, line 13: Awesome, nice catch!

rohan-bes (approved):
I'd slightly push back against adding the node script, but am curious to know if there's a reason I'm not aware of for adding it? Approving regardless 👍

packages/data-api/package.json
IgorNadj: Just for discoverability, you would have to look inside the script to know that is an option.

IgorNadj: @rohan-bes thoughts?

rohan-bes: Oh sorry, ummm I suppose I'm not too fussed 👍 There's documentation in the README of the package explaining the different options, but that's a bit out of the way and doesn't outline them in detail. I'm cool with this 😄

IgorNadj: Ah it would help if I read it! 😆

PR #3170, opened 2021-09-06
Startup script should exit if any command fails

edmofro:
In https://linear.app/bes/issue/WAI-930#comment-09e03aa1 we experienced an issue where all appeared fine with the feature instance, but that was because it was serving the previously built production code. This means the startup script would exit early rather than getting to the part where nginx fires up and makes it look like it's finished successfully


EMcQ-BES (approved):
lgtm

PR #3173, opened 2021-09-07
SOL-410: Migrate report-server to use new transform syntax

rohan-bes:
### Issue SOL-410:

I tried to keep the commits in this PR pretty atomic, so it might be easier for the reviewer to go through them 1 by 1 instead of the whole thing!

Syntax changes proposal: https://beyond-essential.slab.com/posts/report-server-transform-config-design-change-proposals-k8l3tch0#thread-12c3fjiw


packages/report-server/src/reportBuilder/transform/functions/insertRows.ts
```diff
@@ -0,0 +1,79 @@
...
+};
+
+const paramsValidator = yup.object().shape({
+  columns: mapStringToStringValidator,
+  where: yup.string(),
+  position: yup
+    .mixed<'before' | 'after' | 'start'>()
+    .oneOf(['before', 'after', 'start'])
+    .default('after'),
+});
```
rohan-bes, line 34: Made the switch to `yup` for all input validation. Makes code a LOT more readable, and typescript works really nice with it. However, have a slight concern about the performance implications...

packages/report-server/examples.http
```diff
@@ -39,25 +39,22 @@ Authorization: {{authorization}}
...
-          "'Midwives'": "$row.BCD47",
-          "'Nurses'": "sum($row.BCD48, $row.BCD48a, $row.BCD48b)",
-          "'Aides'": "$row.BCD49",
-          "'Others'": "sum($row.BCD50, $row.BCD51, $row.BCD52, $row.BCD53, $row.BCD54, $row.BCD55)",
-          "...": ["period", "organisationUnit"]
+          "title": "Calculate occuptions from data elements",
+          "description": "Sums the raw data elements values to their appropriate occuption",
+          "transform": "updateColumns",
+          "insert": {
+            "Doctors": "$row.BCD46",
```
rohan-bes, line 46: Note: Switching the `$row.x` to `$x` change is a bit trickier than it first seems. Since mathjs throws an error for any undefined symbol, we actually need to do some pre-processing on symbols with `$` at the start to ensure we can handle the `undefined` case. Mathjs fully supports this, and even have documented how to do this: https://mathjs.org/examples/advanced/custom_scope_objects.js

However, I feel it's a bit out of the scope of these refactors, as the correct place to do this would be in the `expression-parser` package itself. So for now we will still live with `$row.x` for a bit longer?

packages/report-server/src/reportBuilder/transform/functions/groupRows/groupRows.ts
```diff
@@ -0,0 +1,120 @@
...
+      }
+    });
+    return mergedRow;
+  });
+};
+
+const groupRows = (rows: Row[], params: GroupRowsParams): Row[] => {
+  const { groups, ungroupedRows } = buildGroups(rows, params);
+  const mergedRows = mergeGroups(groups, params);
+  return mergedRows.concat(ungroupedRows);
```
rohan-bes, line 102: Re-wrote the functions in this class to be a bit cleaner. In doing so it's really solidified what's going on in this operation which is we are first `group`ing rows, and then `merging` those groups.

This ties into a bit of what @edmofro was saying that perhaps a better term for this operation would be `mergeRows`. It also relates to this comment here: https://beyond-essential.slab.com/posts/report-server-transform-config-design-change-proposals-k8l3tch0#thread-12c3fjiw

Would appreciate the reviewer's thoughts on this. Some possible options:
1. Keep as is, `groupRows` `by` `mergeUsing`
2. Use single term, either `groupRows` `by` `groupUsing` or `mergeRows` `by` `mergeUsing`
3. Use 2 terms, either `groupAndMergeRows` `groupBy` `mergeUsing` or just `mergeRows` `groupBy` `using` ?

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -14,11 +15,11 @@ type RowLookup = {
...
- * '$row.BCD1' => 8
- * '$all.BCD1' => [5, 8, 4]
- * '$allPrevious.BCD1' => [5, 8]
- * '$where(f($otherRow) = $otherRow.BCD1 < $row.BCD1).BCD1' => [5, 4]
- * '$row.BCD1 + sum($allPrevious.BCD1)' => 21
+ * '_row.BCD1' => 8
+ * '_all.BCD1' => [5, 8, 4]
+ * '_allPrevious.BCD1' => [5, 8]
+ * '_where(f(_otherRow) = _otherRow.BCD1 < _row.BCD1).BCD1' => [5, 4]
+ * '_row.BCD1 + sum(_allPrevious.BCD1)' => 21
```
rohan-bes, line 22: `#` is a comment in mathjs, so we can't use that to prefix the lookups unfortunately 😢  I tried `@` as well, but that gives a syntax error, so I went with `_` in the end... Would prefer something a bit more visually obvious any suggestions?

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -71,14 +87,29 @@ export class TransformParser extends ExpressionParser {
...
+    this.set('_row', this.lookups.row);
+    this.set('_index', this.lookups.index);
     addRowToLookup(this.lookups.row, this.lookups.allPrevious);
+    this.addRowToScope(this.lookups.row);
   }
 
-  whereFunction = (check: (row: Row) => boolean) => {
+  public addRowToScope = (row: Row) => {
+    Object.entries(row).forEach(([field, value]) => {
+      if (value !== undefined && value !== null && !field.includes(' ')) {
```
rohan-bes, line 100: Hadn't really thought about this, but the shallow variable lookup will only work with columns that have no whitespace in them. There's always the `_row` lookup to get whitespace columns (eg. `_row['Total Cases']`)

packages/report-server/examples.http
rohan-bes: Ignore this, it turned out to not be too hard to support this so I've added support!

edmofro (changes requested):
Super fast and nice work here @rohan-bes!! Very very happy with these changes, just some specific comments but some of them were addressed along the way :-)

packages/report-server/src/reportBuilder/transform/functions/groupRows/groupRows.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+    const groupKey = params.createGroupKey(row);
+    groupsByKey[groupKey] = merge(groupsByKey[groupKey] || {}, row);
+    parser.next();
+  });
+
+  return { groups: Object.values(groupsByKey), ungroupedRows };
+};
+
+const merge = (previousRow: Group, newRow: Row): Group => {
```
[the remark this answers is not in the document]
edmofro: Covered in a later commit :-)

packages/report-server/src/reportBuilder/transform/functions/insertRows.ts
```diff
@@ -0,0 +1,87 @@
...
+  const parser = new TransformParser(rows, functions);
+  const rowsToInsert = returnArray.map(() => {
+    const shouldInsertNewRow = params.where(parser);
+    if (!shouldInsertNewRow) {
+      parser.next();
+      return undefined;
+    }
+    const newRow: Row = {};
+    Object.entries(params.columns).forEach(([key, expression]) => {
+      const newRowKey = key.startsWith('=') ? `${parser.evaluate(key.substring(1))}` : key;
```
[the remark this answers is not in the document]
edmofro: Covered in a later commit :-)

packages/report-server/src/__tests__/reportBuilder/transform/parser.test.ts
```diff
@@ -62,11 +67,11 @@ describe('parser', () => {
...
   });
 
-  it('sort supports row parser lookups', () => {
+  it('sortRows supports row parser lookups', () => {
     const transform = buildTransform([
       {
-        transform: 'sort',
-        by: '$row.BCD1',
+        transform: 'sortRows',
+        by: '=$BCD1',
```
edmofro, line 74: Oh I forgot to mention this point - I am a little nervous of people getting confused about when to use the `=` (i.e. `Total` and `=$BCD1` vs. `'Total'` and `$BCD1`). I had thought it was pretty clean because it's just `=` required on the LHS, not required on RHS, but then I noticed this example where we (very sensibly!) default to using literals the RHS rather than automatically evaluating. One option is to use `=` everywhere, like excel does, but it's a big amount of clutter to add to the very common use case. Any thoughts?

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
rohan-bes: Ignore this, got `@` working in the end! mathjs is so configurable it's amazing! 🎉

packages/report-server/src/__tests__/reportBuilder/transform/groupRows.test.ts
```diff
@@ -4,35 +4,59 @@
...
 
-describe('aggregate', () => {
-  it('can not group by a any field', () => {
+describe('groupRows', () => {
+  it('throws error when mergeUsing is no specified', () => {
+    expect(() =>
+      buildTransform([
+        {
+          transform: 'groupRows',
+          by: 'period',
```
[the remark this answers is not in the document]
rohan-bes: Yeah I've been thinking this for a while too! `single` seems appropriate! We actually have a strategy called `default` that is just `last`, however I think `single` is a better fit 👍

packages/database/src/migrations/20210907022500-ConvertExistingReportsToNewTransformSyntax-modifies-data.js
```diff
@@ -0,0 +1,213 @@
...
+      } else {
+        newConfig.by = column;
+      }
+    } else {
+      const columnKey = column === '...' ? '*' : column;
+      const mergeStrategy = aggregateFunc === 'drop' ? 'exclude' : aggregateFunc;
+      if (typeof newConfig.mergeUsing === 'object' && newConfig.mergeUsing !== null) {
+        newConfig.mergeUsing = { ...newConfig.mergeUsing, [columnKey]: mergeStrategy };
+      } else {
+        newConfig.mergeUsing = { [columnKey]: mergeStrategy };
```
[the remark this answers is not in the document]
rohan-bes: It's probably worth it, just requires keeping a list of the columns so far and handling the first new strategy case a bit and I was feeling lazy 😝 I'll try and do this 👍

packages/report-server/src/__tests__/reportBuilder/transform/insertRows.test.ts
```diff
@@ -163,29 +186,35 @@ describe('insert', () => {
...
       {
-        transform: 'insert',
-        where: 'not(eq($row.organisationUnit, $next.organisationUnit))',
-        "'Total_BCD1'": "sum($where(f($otherRow) = equalText($otherRow.organisationUnit, $row.organisationUnit)).BCD1)",
-        "'Total_BCD2'": "sum($where(f($otherRow) = equalText($otherRow.organisationUnit, $row.organisationUnit)).BCD2)",
+        transform: 'insertRows',
+        where: 'not(eq($organisationUnit, @next.organisationUnit))',
+        columns: {
+          Total_BCD1:
+            'sum(@where(f(@otherRow) = equalText(@otherRow.organisationUnit, $organisationUnit)).BCD1)',
```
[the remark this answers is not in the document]
rohan-bes: I have a few problems with the `where`...  Main one is that we need to use the function syntax with it! However I think I found a way to avoid that if we create it as a `rawFunction` (https://mathjs.org/docs/expressions/customization.html#custom-argument-parsing)

I agree that `@where` doesn't fit with the others here... Possible suggestion is that `where` is the base function, and the other lookups are properties of that function, eg.
```
cost whereFunction = () => { ... do the where logic };
const lookup = whereFunction;
lookup.current = row;
lookup.previous = lastRow;
...etc

parser.set('@lookup', lookup);

So then in a report it would go:

{
  transform: 'updateColumns',
  insert: {
    rowValue: '@lookup.current.value',
    lastRowValue: '@lookup.previous.value',
    sumMatchingRowsValue: 'sum(@lookup(f(@otherRow) = equalText(@otherRow.organisationUnit, $organisationUnit)).value'
  }
}
```

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -14,11 +16,11 @@ type RowLookup = {
  * Lookups object for rows within the transform data
  *
  * eg. if rows = [{BCD1: 5}, {BCD1: 8}, {BCD1: 4}], and currentRow = 1
- * '$row.BCD1' => 8
- * '$all.BCD1' => [5, 8, 4]
- * '$allPrevious.BCD1' => [5, 8]
- * '$where(f($otherRow) = $otherRow.BCD1 < $row.BCD1).BCD1' => [5, 4]
- * '$row.BCD1 + sum($allPrevious.BCD1)' => 21
+ * '@row.BCD1' => 8
```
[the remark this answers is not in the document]
rohan-bes: Current is much better! love it 👍

packages/report-server/src/__tests__/reportBuilder/transform/parser.test.ts
rohan-bes: Yeah I'm a bit unsure too...

Given that `=` isn't too hard to type out, part of me in inclined to go with it everywhere...

Another thought is that if we get a UI around this, we can have the input field itself be toggleable between expression and non-expression, and then maybe we can add sensible defaults there?

packages/report-server/src/__tests__/reportBuilder/transform/insertRows.test.ts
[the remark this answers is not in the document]
edmofro: Interesting idea...I think I prefer the existing, shorter syntax for those simple lookups.

I'm actually fine with `where` using the function syntax, _if_ we don't also use the `@`. Can we just make `where` a standard function?

packages/report-server/src/reportBuilder/functions/math.ts
```diff
@@ -1,62 +1,20 @@
...
+  'undefined, undefined': (undef: undefined, undef2: undefined) => undefined,
+});
 
-  return validNumbers.reduce((numerator, denominator) => mathjsDivide(numerator, denominator));
-};
+export const add = typed('add', {
+  'number, undefined': (num: number, undef: undefined) => num,
+  'undefined, number': (undef: undefined, num: number) => num,
+  'undefined, undefined': (undef: undefined, undef2: undefined) => undefined,
+});
```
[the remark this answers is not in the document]
rohan-bes: Okay bad news bears, seems that using this can't force the `mathjs.sum` function to return `undefined` if passed values that are all `undefined` (it returns `0` which isn't our expected result) 😢 So may still need to manually override `sum` the old fashioned way... But still will be very happy to be able to use `$a + $b` without worry if they're defined or not 🎉

packages/report-server/src/__tests__/reportBuilder/transform/insertRows.test.ts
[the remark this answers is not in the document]
rohan-bes: (btw not sold on the term `lookup`, could be anything, or even just `@`?)

packages/report-server/src/__tests__/reportBuilder/transform/parser.test.ts
edmofro: Good call on the UI (side note, more and more I'm thinking that even when the UI exists, power users will want to write these freehand, because it will just be more fluid if you're creating a bunch)

I'm somewhat inclined to use `=` everywhere too...

packages/report-server/src/reportBuilder/functions/math.ts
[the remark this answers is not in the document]
edmofro: Bugger!

packages/report-server/src/reportBuilder/transform/functions/excludeColumns.ts
```diff
@@ -0,0 +1,59 @@
...
+});
+
+const excludeColumns = (rows: Row[], params: ExcludeColumnsParams): Row[] => {
+  const parser = new TransformParser(rows, functions);
+  return rows.map(row => {
+    const matchesWhere = params.where(parser);
+    if (!matchesWhere) {
+      parser.next();
+      return row;
+    }
```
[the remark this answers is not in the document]
rohan-bes: Tbh I'm not actually sure how much we need `where` in a few of these places. It's a nice thing to have in a way, but I don't think I've ever done a `groupRows` `where` before...

Might not get to it as part of this PR if that's okay, but I do agree with you that if we want it to be standard we should standardise it!

packages/report-server/src/__tests__/reportBuilder/transform/insertRows.test.ts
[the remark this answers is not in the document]
edmofro: Aaaah now that's interesting... `@` _is_ `where`, and there are some shortcuts into `@previous` etc. but also `@(...)`

I'm still wondering if there's anything wrong with just `where(...)` with no prefix though?

packages/report-server/src/reportBuilder/transform/functions/groupRows/groupRows.ts
```diff
@@ -0,0 +1,130 @@
...
+  });
+  return mergedRow;
+};
+
+const mergeGroups = (groups: Group[], params: GroupRowsParams): Row[] => {
+  return groups.map(group => {
+    const mergedRow: Row = {};
+    Object.entries(group).forEach(([fieldKey, fieldValue]) => {
+      const mergeStrategy = params.getMergeStrategy(fieldKey);
+      const mergedValue = mergeStrategies[mergeStrategy](fieldValue);
```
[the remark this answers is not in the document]
rohan-bes: Good call! Btw I've been considering doing a bulk-rename of `FieldValue` to `CellValue` and instead of `field` using the term `column`, eg.

```
type Row = {
  [column: string]: CellValue;
}
```
?

Probably not as part of this PR, but curious to know if you like the abstraction?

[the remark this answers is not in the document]
edmofro: Yep I prefer that :+1:

packages/report-server/src/reportBuilder/transform/functions/groupRows/groupRows.ts
rohan-bes: Ah thanks for looking up the R style! I'm helping Gerry edit a viz via viz-builder today, so I might get his opinion on merge vs group!

Just so ya know, @Alastair-L is also in the `mergeRows` > `groupRows` camp, so you guys are picking up steam!

packages/report-server/src/reportBuilder/transform/functions/excludeColumns.ts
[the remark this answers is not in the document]
edmofro: Yep that's fine

packages/report-server/src/reportBuilder/transform/functions/sortRows.ts
```diff
@@ -0,0 +1,100 @@
...
+const sortRows = (rows: Row[], params: SortParams): Row[] => {
+  const { by, direction } = params;
+  if (typeof by === 'string' && by.startsWith('=')) {
+    const firstDirection = Array.isArray(direction) ? direction[0] : direction;
+    return rows.sort(getCustomRowSortFunction(by.substring(1), firstDirection));
+  }
+
+  const arrayBy = typeof by === 'string' ? [by] : by;
+  const arrayDirection = typeof direction === 'string' ? [direction] : direction;
+  return orderBy(rows, arrayBy, arrayDirection);
```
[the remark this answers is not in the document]
rohan-bes: Yeah I think @kael89 is to thank for that one! It's actually quite relevant here as I believe it handles alpha-numeric sorting correctly for periods, which the base `lodash` one doesn't do as we'd expect!

packages/report-server/src/reportBuilder/transform/functions/groupRows/groupRows.ts
edmofro: Haha yay @Alastair-L 🙌 

I know the `withMatching` was a sticking point, and I had a think about some alternatives if you are still workshopping this, combinations of which are below:

```
[
  {
    "transform": "mergeRows",
    "withSame": "period",
    "using": "sum"
  },
  {
    "transform": "combineRows",
    "using": "sum",
    "splitOn": "period"
  },
  {
    "transform": "squashRows",
    "onePer": "period",
    "using": "sum"
  },
  {
    "transform": "mergeRows",
    "groupBy": "period",
    "using": "sum"
  }
]
```
I think the first and last (your suggestion) are my faves, and I lean slightly towards the last

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
rohan-bes: Yeah, I think this might be a tricky one to support. mathjs isn't liking the `'` in a variable name, and I'm worried if I change how it parses that char it'll break strings...

edmofro: Tricky, we can navigate that one later!

packages/report-server/src/reportBuilder/transform/transform.ts
```diff
@@ -50,7 +50,7 @@ const buildParams = (params: unknown): TransformParams => {
   } = params;
   if (typeof transformStep !== 'string' || !(transformStep in transformBuilders)) {
     throw new Error(
-      `Expected transform to be one of ${Object.keys(transformBuilders)} but got ${transform}`,
+      `Expected transform to be one of ${Object.keys(transformBuilders)} but got ${transformStep}`,
```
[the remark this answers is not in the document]
rohan-bes: Ah thanks for catching that!

packages/database/src/migrations/20210907022500-ConvertExistingReportsToNewTransformSyntax-modifies-data.js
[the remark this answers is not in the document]
rohan-bes: Good point! I'll do it 👍

packages/report-server/src/__tests__/reportBuilder/transform/insertRows.test.ts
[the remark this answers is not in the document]
rohan-bes: I think I might just have a hangup on _lookups_ being this special thing to reference other data in the table, and thus needing to have a special syntax. I think that's probably not a good justification for keeping where being special haha! 

Would having both bother you? I can see the argument against it since it's usually not wise having two ways to do the one thing?

packages/report-server/src/__tests__/reportBuilder/transform/parser.test.ts
rohan-bes: Should we trial `=` everywhere and see how it goes? We can always go back if it's annoying?

edmofro: See what Gerry thinks in your session today ;-)

packages/report-server/src/__tests__/reportBuilder/transform/insertRows.test.ts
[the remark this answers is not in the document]
edmofro: I'd prefer to go with the special syntax than both I think, agree that we're better to be opinionated than allow multiple options here

packages/report-server/src/reportBuilder/transform/functions/groupRows/groupRows.ts
rohan-bes: So when I ran through this with @gerrybes he definitely described the operation as 'grouping by'. However, once I showed him the merge config he also felt it made total sense 😂

Are we all okay with `mergeRows` `groupBy` `using` ? It seems people really like group by as a term, but merge rows is a better description of the actual operation of this transform?

packages/report-server/src/__tests__/reportBuilder/transform/parser.test.ts
rohan-bes: Okay, interestingly @gerrybes did actually seem to find the `=` in one place but not the other to be a bit confusing! So I'm pretty keen to try `=` is always required for expressions 👍

One other thing he noted was:
`=$a + $b` is a tad confusing, at least in comparison to `=($a + $b)`. I can definitely see his point here, however I feel forcing people to wrap all their expressions in quotes is a bit too verbose. Since expression-parser will allow you to wrap or not wrap, we can leave this is an optional stylistic thing?

edmofro:
Yep I like that option a lot.

edmofro:
Agreed, leaving that up to the individual preference works. Personally I’ll
probably use a space after the = but no brackets, assuming that also works?

packages/report-server/src/__tests__/reportBuilder/transform/insertRows.test.ts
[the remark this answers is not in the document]
rohan-bes: Definitely happy to chat about this one with others and change it later on! Kinda just keen to get these changes out at this point as I feel we can discuss semantics til the cows come home! 🐮 

Also worth noting that where is a pretty complex operation and I hope most people aren't using it! If it's in heavy use I feel it's symptomatic of report-server needing more specific built in functions!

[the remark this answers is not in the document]
rohan-bes: Okay took one look at the 'special syntax' for where and in it's current form it's pretty unreadable (**LOTS** of symbols 😝 ) am going with `where` !

packages/report-server/src/__tests__/reportBuilder/transform/groupRows.test.ts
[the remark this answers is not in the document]
rohan-bes: Just noting, this default will apply:
- To all rows if no `using` is given
- To rows that have no explicit `using`, and if there's no `*` in the `using`

packages/report-server/src/reportBuilder/transform/functions/groupRows/groupRows.ts
```diff
@@ -0,0 +1,130 @@
...
+type GroupRowsParams = {
+  createGroupKey: (row: Row) => string;
+  getMergeStrategy: (field: string) => keyof typeof mergeStrategies;
+  where: (parser: TransformParser) => boolean;
+};
+
+const paramsValidator = yup.object().shape({
+  by: yup.lazy((value: unknown) => {
+    if (typeof value === 'string' || value === undefined) {
+      return yup.string();
```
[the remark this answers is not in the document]
rohan-bes: Yeah this is required by `yup.lazy`. `lazy` is not only doing lazy validation, but lazy validation validation 😂 in the sense that it doesn't decide what validator to use until it's seen the object. So you still need to tell it to validate a string as a string 😖

rohan-bes:
> Personally I’ll
probably use a space after the = but no brackets, assuming that also works?

Yup, that works 👍

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -50,31 +49,39 @@ export class TransformParser extends ExpressionParser {
...
       this.rows.forEach(row => addRowToLookup(row, this.lookups.all));
-      addRowToLookup(this.lookups.row, this.lookups.allPrevious);
-      this.addRowToScope(this.lookups.row);
+      addRowToLookup(this.lookups.current, this.lookups.allPrevious);
+      this.addRowToScope(this.lookups.current);
 
       Object.entries(this.lookups).forEach(([lookupName, lookup]) => {
         this.set(`@${lookupName}`, lookup);
       });
+      this.set('where', this.whereFunction); // '@' prefix for where
```
edmofro, line 71: Is that comment missing a word?

packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
```diff
@@ -13,19 +13,23 @@ import { buildCreateGroupKey } from './createGroupKey';
...
   where: (parser: TransformParser) => boolean;
 };
 
 const paramsValidator = yup.object().shape({
-  by: starSingleOrMultipleColumnsValidator,
-  mergeUsing: yup.lazy((value: unknown) => {
-    const mergeStrategyNameValidator = yup
+  groupBy: starSingleOrMultipleColumnsValidator,
+  using: yup.lazy((value: unknown) => {
+    const undefinedMergeStrategyNameValidator = yup
```
edmofro, line 25: Would slightly prefer `optionalMerge...` because of the `.required()` use below (clever, btw)

packages/report-server/src/reportBuilder/transform/functions/insertRows.ts
```diff
@@ -43,8 +42,7 @@ const insertRows = (rows: Row[], params: InsertParams): Row[] => {
     }
     const newRow: Row = {};
     Object.entries(params.columns).forEach(([key, expression]) => {
-      const [newRowKey, newRowValue] = getParsedColumnKeyAndValue(key, expression, parser);
-      newRow[newRowKey] = newRowValue;
+      newRow[parser.evaluate(key)] = parser.evaluate(expression);
```
edmofro, line 45: Love that it not only simplifies the rules for users, it also simplifies our code

packages/report-server/src/__tests__/reportBuilder/transform/updateColumns.test.ts
```diff
@@ -12,9 +12,9 @@ describe('updateColumns', () => {
       {
         transform: 'updateColumns',
         insert: {
-          number: '1',
-          string: "'Hi'",
-          boolean: 'false',
+          number: '=1',
```
edmofro, line 15: This one's kinda a shame huh, I think it might be unexpected that it doesn't automatically interpret numeric only characters as numbers but let's leave it for now and see what users get tripped up on

packages/report-server/src/__tests__/reportBuilder/transform/parser.test.ts
```diff
@@ -48,22 +48,23 @@ describe('parser', () => {
       {
         transform: 'updateColumns',
         insert: {
-          row: '$BCD1',
-          lastAll: 'last(@all.BCD1)',
-          sumAllPrevious: 'sum(@allPrevious.BCD1)',
+          row: '= $BCD1 + $BCD2',
```
edmofro, line 51: Maybe this should be called 'arithmetic' rather than 'row'?

packages/admin-panel-server/examples.http
```diff
@@ -93,20 +95,22 @@ Authorization: {{authorization}}
             "transform": [
                 "keyValueByDataElementName",
                 {
+                    "title": "Calculate occuptions from data elements",
```
edmofro, line 98: `occupations`

packages/admin-panel-server/examples.http
```diff
@@ -93,20 +95,22 @@ Authorization: {{authorization}}
...
+                    "description": "Sums the raw data elements values to their appropriate occuption",
```
edmofro, line 99: `occupation`

packages/admin-panel-server/examples.http
```diff
@@ -146,20 +150,22 @@ Authorization: {{authorization}}
...
```
edmofro, line 154: As above

packages/database/src/migrations/20210907022500-ConvertExistingReportsToNewTransformSyntax-modifies-data.js
```diff
@@ -57,29 +66,51 @@ const convertAggregateToGroupRows = oldConfig => {
...
+              commonStrategy,
+            ]),
+          );
+          usingObject[columnKey] = mergeStrategy;
+          newConfig.using = usingObject;
+        }
       } else {
-        newConfig.mergeUsing = { [columnKey]: mergeStrategy };
+        columnsWithBaseStrategy.push(columnKey);
+        newConfig.using = mergeStrategy;
```
edmofro, line 113: Very nice work condensing these down where possible!

packages/report-server/src/__tests__/reportBuilder/transform/mergeRows.test.ts
```diff
@@ -7,38 +7,39 @@ import {
...
+        groupBy: 'period',
+      },
+    ]);
+    expect(transform(SINGLE_MERGEABLE_ANALYTICS)).toEqual([
+      { period: '20200101', BCD1: 4, BCD2: 4, BCD3: 4 },
+    ]);
   });
 
-  it('when no by is specified, group to a single row', () => {
+  it('when no groupBy is specified, group to a single row', () => {
```
edmofro, line 38: I really like the term `mergeRows` now that neither of the keys are strictly required - it means that you could have

| Doctors | Nurses | Midwives |
|---------|--------|----------|
| 1       |        |          |
|         | 5      |          |
|         |        | 2        |

Then run a simple `mergeRows` and get

| Doctors | Nurses | Midwives |
|---------|--------|----------|
| 1       | 5      | 2        |

edmofro (approved):
Amazing! I think this is ready to go! I have a few super minor things inline, but up to you whether you action any, and either way this is ready for test juuust in case we could get it through today ;-)

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
rohan-bes: Ahaha yup 🤦 I that a lot!

packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
rohan-bes: I was kinda annoyed by this tbh, as far as I can tell `yup` has no `undefined` validator, so instead I returned a non-required string validator. Seems a bit funky to me but it gets the job done 🤷 

Yup, `optional` is a better term, will rename 👍

packages/report-server/src/__tests__/reportBuilder/transform/updateColumns.test.ts
rohan-bes: Yeah tis a bit. I suspect we will change this soon enough since sometimes we do things like `insertColumns 'Total': '1'` followed by a `mergeRows` to sum all the totals...

In theory now we could actually move to supporting values other than strings in here as a way to get around that? eg. `{ number: 1 }` instead of `{ number: '1' }`

I chose not to do that right now because:
a) It would probably break a bunch of code in places, and would definitely break the typescript
b) If we're gonna just have a single text input box for this in a UI anyway, we don't want users to need to specify if they're using strings vs raw numbers/booleans, so it might be best to only use strings internally as well

packages/report-server/src/__tests__/reportBuilder/transform/parser.test.ts
rohan-bes: Good call! I'll also add checks for all the other lookups (`@next`,`@table`,etc.)

packages/report-server/src/__tests__/reportBuilder/transform/updateColumns.test.ts
edmofro: Good points, and in general I think we're at a good point in terms of scope of change with this PR, and can work out how we want to handle numbers in the next iteration rather than dragging this out further

packages/database/src/migrations/20210907022500-ConvertExistingReportsToNewTransformSyntax-modifies-data.js
rohan-bes: I was just being lazy before, it wasn't too bad in the end. Thanks for pushing for it 🙌

packages/report-server/src/__tests__/reportBuilder/transform/mergeRows.test.ts
rohan-bes: Yeah me too! I think I had been a bit worried in the past about having a default for `mergeRows` since someone might accidentally omit the `using` and then it would work fine but the data would get messed up (lets say if `last` were the default). 

But now that we have `single` it's actually quite a strict case! And it's kind of natural to just say 'merge those rows' without needing to specify the merge strategy since it's sort of obvious!

packages/report-server/src/reportBuilder/transform/transform.ts
```diff
@@ -3,70 +3,78 @@
...
-  transformSteps.forEach((transformStep: TransformParams) => {
-    transformedRows = transformStep.apply(transformedRows);
+  transformSteps.forEach((transformStep: TransformParams, index: number) => {
+    try {
+      transformedRows = transformStep.apply(transformedRows);
+    } catch (e) {
+      const titlePart = transformStep.title ? ` (${transformStep.title})` : '';
+      const errorMessagePrefix = `Error in transform[${index + 1}]${titlePart}: `;
+      throw new Error(`${errorMessagePrefix}${(e as Error).message}`);
+    }
```
rohan-bes, line 47: @edmofro hope you're okay with me sneaking this in! It's just a bit of logic to improve the error messaging when the report fails. Now you get something like:
`Error in transform[2] (Calculate occuptions from data elements): Unexpected operator ) (char 52)`

rohan-bes: I went with 1 based index again to be consistent... but might be a tad surprising to devs...

edmofro: Great addition!

packages/admin-panel-server/examples.http
```diff
@@ -43,23 +43,22 @@ Authorization: {{authorization}}
...
                 "keyValueByDataElementName",
                 {
-                    "transform": "select",
-                    "'Doctors'": "$row.BCD46",
-                    "'Midwives'": "$row.BCD47",
-                    "'Nurses'": "sum($row.BCD48, $row.BCD48a, $row.BCD48b)",
-                    "'Aides'": "$row.BCD49",
-                    "'Others'": "sum($row.BCD50, $row.BCD51, $row.BCD52, $row.BCD53, $row.BCD54, $row.BCD55)",
-                    "...": ["period", "organisationUnit"]
+                    "title": "Calculate occuptions from data elements",
```
edmofro, line 46: Heh, now I'm wondering if I've always just spelt occupation incorrectly!

edmofro (approved):
All looks good, your choice on the spelling ;-)

packages/admin-panel-server/examples.http
rohan-bes: Woops!! I honestly can't spell!!

PR #3175, opened 2021-09-07
Add api-client to packages using ts eslint config

edmofro:
### Issue #:
See https://beyondessential.slack.com/archives/C02DKC9U1FU/p1630986969002500


kael89 (approved):
Thanks for sorting those entries - scanning lists when they are not sorted can be a struggle!

PR #3176, opened 2021-09-07
MEL-472 Remove default to zero for empty string

IgorNadj:
Currently if a data value is `''` (empty string), it will be converted to 0 before being passed in to the parser. This is not an appropriate default since empty string is a valid data value representing no data. This also means you can not differentiate between empty string and zero in the expression formula.

The change I went with is to remove the defaulting. 

Any indicators/reports that rely on `'' -> 0` should fail outright due to errors being thrown inside internal mathjs functions which only work on number type arguments not string, and the fix for these broken vizes (if any) is to convert empty string to a number explicitly in the formula, e.g. `(equalText('', a) ? 0 : a) + 5`. We may also want to add a helper function for this if this use case becomes common.

Any indicators/reports that rely on `null|undefined -> 0` should fail similarly, with mathjs throwing errors due to the type, and the fix for these broken vizes (if any) is to define defaultValues e.g. `indicator.config: { "formula": "a + 5", "defaultValues": { "a": 0 }}`.

### DIFF
--- a/packages/expression-parser/src/expression-parser/ExpressionParser.js
+++ b/packages/expression-parser/src/expression-parser/ExpressionParser.js
@@ -114,8 +114,7 @@ export class ExpressionParser {
    */
   setScope(scope = {}) {
     Object.entries(scope).forEach(([name, value]) => {
-      const expressionValue = value || 0;
-      this.set(name, expressionValue);
+      this.set(name, value);
     });
   }
 


packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -114,8 +114,7 @@ export class ExpressionParser {
    */
   setScope(scope = {}) {
     Object.entries(scope).forEach(([name, value]) => {
-      const expressionValue = value || 0;
-      this.set(name, expressionValue);
+      this.set(name, value);
```
rohan-bes, line 117: Great change, very much approve! The defaulting behaviour was quite unexpected, and could break lots of behaviour.

Just a note that I renamed this function in https://github.com/beyondessential/tupaia/pull/3173/files#diff-6d258c85fc6c829d3bcd8670f8d94362dba252a275bbe68374799674c494c03e but I don't think changes should conflict...

rohan-bes (approved):
Looks good to me 👍 Just a warning that the expression parser is also used by meditrak-app for Arithmetic questions, so that's another scope of regression risk!

IgorNadj:
Branch name too long, replaced with https://github.com/beyondessential/tupaia/pull/3183

PR #3177, opened 2021-09-07
Sol 265 import new survey

BinduMadhaviTammireddy:
### Issue #:

Import a survey and check it through search .
### Changes:

- Example

---

### DIFF
--- a/packages/admin-panel/cypress/e2e/survey/surveyImport.spec.js
+++ b/packages/admin-panel/cypress/e2e/survey/surveyImport.spec.js
@@ -26,7 +26,7 @@ beforeEach(() => {
 describe('import new survey file', () => {
   // Check error message.
   it('Name does not match the sheet Name ', () => {
-    enterSurveyName('Test');
+    enterSurveyName('Test Survey_12');
     cy.uploadFile('surveys/Test Survey_1.xlsx');
     importSurvey();
     // assertion for error message.
--- a/packages/admin-panel/cypress/e2e/survey/surveyImport.spec.js
+++ b/packages/admin-panel/cypress/e2e/survey/surveyImport.spec.js
@@ -15,6 +15,7 @@ import {
   importSurvey,
   searchBySurveyName,
   closeImportSurveyForm,
+  checkSurveyByName,
 } from '../../support';
 
 beforeEach(() => {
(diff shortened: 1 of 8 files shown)


packages/admin-panel/cypress/e2e/survey/surveyImport.spec.js
```diff
@@ -12,15 +13,40 @@ import {
...
+});
+
+describe('import new survey file', () => {
+  // Check error message.
+  it('Name does not match the sheet Name ', () => {
+    enterSurveyName('Test');
+    cy.uploadFile('surveys/Test Survey_1.xlsx');
+    importSurvey();
+    // assertion for error message.
+    cy.get('form').should('includes.text', 'Import failed');
```
kael89, line 32: While checking front end messages is a great start, in order to have more confidence in our tests we should also check the state of the data after the import. Of course, this should be done using a UI accessible to the user (eg the survey search table). I guess that this is what a tester would do too: instead of just trusting the dialog message, you would verify that things work as expected by searching the data.

In this particular case, what we could do is search the survey table and confirm that the survey was not imported.

Happy for this to be added at a later stage though 🙂

packages/admin-panel/cypress/e2e/survey/surveyImport.spec.js
```diff
@@ -12,15 +13,40 @@ import {
...
+    searchBySurveyName('Test Survey_1');
+    // assertion for survey search.
+    cy.FirstRowElementOfTable().eq(1).should('have.text', 'Test Survey_1');
   });
 
-  it('import a single survey to a single country', () => {
+  it('import a new survey by filling all the fields', () => {
     enterSurveyName('Test Survey_1');
     enterCountryName('Demo Land');
     enterPermissionGroup('Admin');
```
kael89, line 52: I believe the tests would benefit from a more generic util, eg

```js
fillImportSurveyForm({
  name: 'Test Survey_1',
  countryName: 'Demo Land'
});
```

And since we only display one form dialog at a time, this could actually be generalised so that it can be used in any admin panel test:

```js
fillDialogForm({
  Survey Name: 'Test Survey_1',
  Countries: ['Demo Land']
});
```

In which case we may decide its worth making this a command, since it will be present in plenty of admin panel tests (`cy.fillDialogForm`)

kael89 (approved):
Nice work @BinduMadhaviTammireddy, I think you've gotten the hang of it! 🎉 My suggestions are mainly about introducing more generic utils so that we can easily add more tests in the future. Moreover, the right level of abstraction (in this case,  probably a bit higher up) enhances maintainability and readability, but the last point is a bit subjective in this case.

Although I believe you've got this, please let me know if you are keen on introducing more generic helpers and want a second pair of eyes - I'd be happy to do a pair programming session on that!

Approving the PR since the code looks good as is!

kael89 (approved):

PR #3178, opened 2021-09-08
Support hide from popup in tooltips

billli0:
### Issue #:
SOL-711 Support specifying `hideFromPopup` for different values in area tooltips (hover). 

config:

>  "values": [
        {
          "value": null,
          "hideFromPopup": true
        }
      ]

### Change:
Use the same logic in `PopupMarker` for rendering the list and hideFromPopup.

### DIFF
--- a/packages/web-frontend/src/containers/Map/AreaTooltip.js
+++ b/packages/web-frontend/src/containers/Map/AreaTooltip.js
@@ -74,7 +74,12 @@ export class AreaTooltip extends Component {
             {orgUnitName}
           </Heading>
           {hasMeasureValue && (
-            <PopupDataItemList key={1} measureOptions={measureOptions} data={orgUnitMeasureData} />
+            <PopupDataItemList
+              key={1}
+              measureOptions={measureOptions}
+              data={orgUnitMeasureData}
+              showNoDataLabel="true"
+            />
           )}
         </div>
       </Tooltip>
--- a/packages/web-frontend/src/containers/Map/AreaTooltip.js
+++ b/packages/web-frontend/src/containers/Map/AreaTooltip.js
@@ -13,47 +13,16 @@
  */
 
 import React, { Component } from 'react';
+import { PopupDataItemList } from '@tupaia/ui-components/lib/map';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
 import { Tooltip } from 'react-leaflet';
-import { getFormattedInfo } from '../../utils/measures';
 
 const Heading = styled.span`
   text-align: center;
   font-weight: ${props => (props.hasMeasureValue ? 'bold' : 'normal')};
 `;
-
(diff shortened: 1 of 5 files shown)


packages/web-frontend/src/containers/Map/AreaTooltip.js
```diff
@@ -71,22 +66,28 @@ export class AreaTooltip extends Component {
...
+              key,
+              metadata,
+              ...otherConfigs,
+            });
+            const formattedKey = name || key;
+            return valueInfo?.hideFromPopup ? null : (
+              <span key={key}>{`${formattedKey}: ${formattedValue}`}</span>
+            );
+          })
+          .filter(popupItem => popupItem !== null)
```
billli0, line 84: A bit hard to do a filter before map for hideFromPopup in the `value` config.

packages/web-frontend/src/containers/Map/AreaTooltip.js
```diff
@@ -71,22 +66,28 @@ export class AreaTooltip extends Component {
...
+            const { formattedValue, valueInfo } = getFormattedInfo(orgUnitMeasureData, {
+              key,
+              metadata,
+              ...otherConfigs,
+            });
+            const formattedKey = name || key;
+            return valueInfo?.hideFromPopup ? null : (
+              <span key={key}>{`${formattedKey}: ${formattedValue}`}</span>
+            );
+          })
```
rohan-bes, line 83: I see a lot of commonality between this function and the one in `MeasurePopup.js`, think it's worth refactoring so that they share the same code? I suspect that the MeasurePopup would benefit from having this `getMetadata()` logic?

rohan-bes (changes requested):
Just a request to refactor out the common code with MeasurePopup.js? Otherwise is looking great!

packages/web-frontend/src/containers/Map/AreaTooltip.js
billli0: @rohan-bes That's a good point, will refactor the code to share the function.👍

rohan-bes (approved):
Love it when the `-` is bigger than the `+`!! Nice work @billli0!

Good job at keeping ui-components up to date as well! Would be really nice to make the switch to just ui-components soon 🙏

billli0:
Hi @tcaiger, we want to add a small thing in tooltips and do a bit refactoring for the popup list in this PR. Since your PR `nz-22-update-map-ui-components` is also across the same code, do you think I can merge this one into dev, then raise a PR to apply the changes to yours?

tcaiger:
> Hi @tcaiger, we want to add a small thing in tooltips and do a bit refactoring for the popup list in this PR. Since your PR `nz-22-update-map-ui-components` is also across the same code, do you think I can merge this one into dev, then raise a PR to apply the changes to yours?

Hey @billli0, yes that sounds good to me. Go for it.

PR #3179, opened 2021-09-08
Sol 299 positivity measure

Alastair-L:
### Issue #: SOL-299

Also includes a small admin panel change


packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
```diff
@@ -100,7 +112,7 @@ const getColumns = data => {
   const columns = columnKeys.map(columnKey => {
     return {
       Header: columnKey,
-      accessor: row => row[columnKey],
+      accessor: row => convertValueToPrimitive(row[columnKey]),
```
Alastair-L, line 115: Otherwise it will error, I took a look at the react table docs (which is where this is eventually rendered) and they say that accessor can only return a primitive value. I hope this is ok to slip in, as it will only ever matter when `val` is an object or a function, where previously it would give an unhelpful error (sorry I forget what it was)

edmofro (changes requested):
Hey @Alastair-L, unfortunately we'll need to update this migration to use the new and improved syntax, which is detailed [here](https://beyond-essential.slab.com/posts/report-server-transform-config-design-change-proposals-k8l3tch0) - thanks!

PR #3187, opened 2021-09-09
SOL-83 Pacmossi mosquito species district

billli0:
### Issue #:
SOL-83

### Changes:
- Support transparent polygon value config
- Migration


edmofro (changes requested):
@billli0 unfortunately the report within this will need to be updated to use our (very cool) new format - could you please make those updates as per https://beyond-essential.slab.com/posts/report-server-transform-config-design-change-proposals-k8l3tch0 ? Cheers!

packages/database/src/migrations/20210901022638-AddReportForMosquitoSpeciesPacmossi-modifies-data.js
```diff
@@ -0,0 +1,98 @@
...
+          numerator: '=$value',
+          denominator:
+            '=sum(where(f($otherRow) = equalText($otherRow.organisationUnit, $organisationUnit)).value)',
+          name: `=translate($row.dataElement, ${JSON.stringify(dataElementsToNames)})`,
+          organisationUnitCode: '=$organisationUnit',
+        },
+        exclude: '*',
+      },
+
+      { transform: 'excludeRows', where: '=not ($denominator > 0)' },
```
rohan-bes, line 71:
```suggestion
      { transform: 'excludeRows', where: '=$denominator <= 0' },
```

packages/database/src/migrations/20210901022638-AddReportForMosquitoSpeciesPacmossi-modifies-data.js
```diff
@@ -0,0 +1,98 @@
...
+      {
+        transform: 'updateColumns',
+        insert: {
+          value: "='exists'",
```
rohan-bes, line 75:
```suggestion
          value: "exists",
```

packages/database/src/migrations/20210901022638-AddReportForMosquitoSpeciesPacmossi-modifies-data.js
```diff
@@ -0,0 +1,98 @@
...
+          '=$name': '=formatAsFractionAndPercentage($numerator, $denominator)',
+        },
+        include: ['organisationUnitCode'],
+      },
+      { transform: 'mergeRows', using: 'last', groupBy: 'organisationUnitCode' },
```
rohan-bes, line 80:
```suggestion
      { transform: 'mergeRows', groupBy: 'organisationUnitCode' },
```
Not 100% sure if this is true in your case, but if you expect there just to be a single defined value per org unit then you can just leave `using` empty and it will default to `single`.  `single` throws an error if there's more than 1 👍

rohan-bes (approved):
Looks good! Nice work on using the new syntax! Just a couple of recommendation around the report transform config, but pre-approving 👍

packages/database/src/migrations/20210901022638-AddReportForMosquitoSpeciesPacmossi-modifies-data.js
billli0: That makes sense! 👍

PR #3189, opened 2021-09-09
Hot fix WAI-939: survey submissions with empty answers blocking sync

edmofro:
### Issue #:
WAI-939

### DIFF
--- a/packages/meditrak-app/app/assessment/actions/submitSurvey.js
+++ b/packages/meditrak-app/app/assessment/actions/submitSurvey.js
@@ -98,7 +98,7 @@ const processQuestions = async (dispatch, getState, database, userId, questions)
   const answers = getAnswers(getState());
   for (const question of questions) {
     const answer = answers[question.id];
-    if (answer === undefined) {
+    if (answer === undefined || answer === null || answer === '') {
       continue;
     }
 
--- a/packages/meditrak-app/package.json
+++ b/packages/meditrak-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tupaia/meditrak-app",
-  "version": "1.11.113",
+  "version": "1.11.114",
   "private": true,
   "description": "Android and iOS app for surveying medical facilities",
   "repository": {
--- a/packages/meditrak-server/src/apiV2/postChanges.js
+++ b/packages/meditrak-server/src/apiV2/postChanges.js
@@ -182,8 +182,10 @@ const requiresSurveyResponseTranslation = (
   const answerTranslatorFields = Object.keys(answerTranslators);
   const { answers } = surveyResponseObject;
   if (Array.isArray(answers)) {
-    return answers.some(answer =>
-      Object.keys(answer).some(field => answerTranslatorFields.includes(field)),
+    return answers.some(
+      answer =>
+        answer.body === '' ||
+        Object.keys(answer).some(field => answerTranslatorFields.includes(field)),
     );
   }
 
@@ -200,6 +202,7 @@ const constructSurveyResponseTranslators = models => ({
   user_email: userEmail => translateUserEmailToIdAndAssessorName(models.user, userEmail),
   entity_code: entityCode => translateEntityCodeToId(models.entity, entityCode),
   survey_code: surveyCode => translateSurveyCodeToId(models.survey, surveyCode),
+  answers: answers => answers.filter(a => a.body !== ''), // remove any empty answers
 });
 
 const constructAnswerTranslators = models => ({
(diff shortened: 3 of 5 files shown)


packages/meditrak-app/app/assessment/actions/submitSurvey.js
```diff
@@ -98,7 +98,7 @@ const processQuestions = async (dispatch, getState, database, userId, questions)
   const answers = getAnswers(getState());
   for (const question of questions) {
     const answer = answers[question.id];
-    if (answer === undefined) {
+    if (answer === undefined || answer === null || answer === '') {
```
edmofro, line 101: This fixes the original issue I introduced in the hot fix: answers that were previously filtered out are now making it through. Still allows the 0's to successfully make it through, which was the motivation for the original hot fix.

packages/meditrak-server/src/apiV2/postChanges.js
```diff
@@ -182,8 +182,10 @@ const requiresSurveyResponseTranslation = (
   const answerTranslatorFields = Object.keys(answerTranslators);
```
edmofro, line 182: This file applies a server-side translation, so that even people with the existing version of the app can submit data using it, and it'll strip out the empty string answers at this point

packages/meditrak-server/src/apiV2/postChanges.js
```diff
@@ -182,8 +182,10 @@ const requiresSurveyResponseTranslation = (
...
   const { answers } = surveyResponseObject;
   if (Array.isArray(answers)) {
-    return answers.some(answer =>
-      Object.keys(answer).some(field => answerTranslatorFields.includes(field)),
+    return answers.some(
+      answer =>
+        answer.body === '' ||
```
rohan-bes, line 187: So this is for just the empty string case? As the undefined/null should just be omitted from the payload by JSON?

packages/meditrak-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tupaia/meditrak-app",
-  "version": "1.11.113",
+  "version": "1.11.114",
```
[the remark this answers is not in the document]
rohan-bes: Do we need to bump version in this file too? `ios/TupaiaMediTrak/Info.plist package.json`? Just going off this: https://beyond-essential.slab.com/posts/medi-trak-app-release-processes-48g606pu

rohan-bes (approved):
Changes look good! Just a q about if we need to edit anywhere else when bumping meditrak-app version

packages/meditrak-server/src/apiV2/postChanges.js
edmofro: At this point the app will have translated anything like that into string format, so we wouldn't expect to see a body of any other type, and if we ever do, I'd like it to still error so we investigate. 

The awkward bit is that if there was a `null` response that the same bug in the app is failing to filter out, by this point on the server it would be `'null'`, but that's indistinguishable from a user actually typing in 'null' and submitting that, so we just can't handle that case. It shouldn't matter too much if there are some `'null'` answers in the db, and I will try to get the app update deployed asap so the window of them potentially making it through is short.

packages/meditrak-app/package.json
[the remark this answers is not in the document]
edmofro: Totally right, thanks for catching that!

edmofro:
After testing locally I needed to make a couple of changes, sorry @rohan-bes threw this one up a bit hastily

packages/meditrak-server/src/apiV2/utilities/translateObjectFields.js
```diff
@@ -8,8 +8,8 @@ export const translateObjectFields = async (object, objectTranslators) => {
     Object.entries(object).map(async ([field, value]) => {
       if (objectTranslators[field]) {
         const newFields = await objectTranslators[field](value);
-        Object.entries(newFields).forEach(([newField, newValue]) => (object[newField] = newValue));
         delete object[field];
```
edmofro, line 11: Move the `delete` to before the line where we add the new entries to the survey response object, so that we don't go: add the translated `answers`, then delete the field `answers`

packages/meditrak-server/src/apiV2/postChanges.js
```diff
@@ -202,7 +202,7 @@ const constructSurveyResponseTranslators = models => ({
   user_email: userEmail => translateUserEmailToIdAndAssessorName(models.user, userEmail),
   entity_code: entityCode => translateEntityCodeToId(models.entity, entityCode),
   survey_code: surveyCode => translateSurveyCodeToId(models.survey, surveyCode),
-  answers: answers => answers.filter(a => a.body !== ''), // remove any empty answers
+  answers: answers => ({ answers: answers.filter(a => a.body !== '') }), // remove any empty answers
```
edmofro, line 205: This translator had a bit of an unexpected format, where you need to return the _changed or new fields in the survey response object_ rather than just the updated field. Makes a bit of sense when considering the use cases above (swapping `survey_code` to `survey_id` etc.)

rohan-bes (approved):
Is looking good to me!

PR #3190, opened 2021-09-09
MEL-485 Allow expression parser to use all mathjs functions

IgorNadj:
MEL-485

### DIFF
--- a/packages/expression-parser/src/expression-parser/ExpressionParser.js
+++ b/packages/expression-parser/src/expression-parser/ExpressionParser.js
@@ -36,14 +36,6 @@ const translate = (value, translations) => {
   return translations[value];
 };
 
-/**
- * List of built in functions in math.js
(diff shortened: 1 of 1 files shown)


packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -68,6 +60,20 @@ export class ExpressionParser {
...
   }
 
+  /**
+   * @param {string} name
+   * @return bool
+   */
+  isBuiltInFunction(name) {
+    try {
+      // see https://mathjs.org/docs/reference/functions/help.html
+      this.math.help(name);
```
rohan-bes, line 70: Wow, I hadn't realised mathjs has `help` built into it! I wonder if you can extend the default functions with your own ones, could neat to provide in-app help in viz builder or something...

rohan-bes (approved):
Very clever solution, I love it!

PR #3192, opened 2021-09-09
2354 env banner

tcaiger:
### Changes:

- Add new EnvBanner component to ui-components
- Add branch name as an env variable for front end builds in CI/CD
- Add EnvBanner to AdminPanel, Lesmis, PSSS and Tupaia

---

### Screenshots:
See issue


packages/admin-panel/src/index.js
```diff
@@ -6,15 +6,19 @@ import React, { lazy, Suspense } from 'react';
...
 import { render as renderReactApp } from 'react-dom';
 import 'react-table/react-table.css';
+import { EnvBanner } from '@tupaia/ui-components';
 import AdminPanel from './App';
 import { AdminPanelProviders, VizBuilderProviders } from './utilities';
 import { Footer, Navbar } from './widgets';
 
 const VizBuilder = lazy(() => import('./VizBuilderApp'));
 
+const branch = process.env.REACT_APP_BRANCH;
```
tcaiger, line 16: @reviewer Maybe the branch variable should just be moved into the `EnvBanner` component? What do you think?

edmofro: I agree, at least as a default (prop could still be exposed in case someone wanted to use a different variable, but not totally necessary)

edmofro (approved):
lgtm! Up to you whether you move the env reading into the component

edmofro (commented):
@tcaiger I've just merged dev into nz-917, and am resolving conflicts here. I can't work out how the startup script build gets the `REACT_APP_BRANCH` environment variable? Does this work when the server first starts up?

PR #3193, opened 2021-09-10
Sol 442 error in data preview for cell

chris-pollard:
### Issue #: 
SOL-442
### Changes:

- Stringify report data
- Add word-break css for table row cells

---


packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
```diff
@@ -133,6 +137,13 @@ export const PreviewSection = () => {
   const columns = useMemo(() => getColumns(reportData), [reportData]);
   const data = useMemo(() => reportData, [reportData]);
 
+  const stringifiedData = data.map(row => {
+    let rowArrays = Object.entries(row).map(([key, value]) => {
```
rohan-bes, line 141:
```suggestion
    const rowArrays = Object.entries(row).map(([key, value]) => {
```

Just a point on the `let` vs `const` here. We have a strong preference towards `const` over `let` where possible. It enforces immutability, which keeps code simpler and a bit more bug free, but it's also just a stylistic thing. In my IDE I get a red underline under `let` here as we have an eslint rule set up to flag these instances. Do you get the same? If not let me know and I'll help ya get eslint set up properly for your vscode 👍

packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
```diff
@@ -48,6 +48,10 @@ const StyledTable = styled(DataTable)`
     thead {
       text-transform: none;
     }
+
+    td {
+      word-break: break-all;
+    }
```
rohan-bes, line 54: Awesome, thanks!

rohan-bes (approved):
Super clean work here @chris-pollard ! You've stolen the joy of tormenting someone on their first PR from me, it's practically perfect 😝 

One comment about using `let` over `const` here, but gonna pre-approve anyway 👍 

Do let someone know though if you're not get the eslint validation errors showing up in your IDE, it's the main tool we use to enforce code style/convention here so it's pretty important to have it working.

One last note, for frontend bugs like this it's always nice to put up a screenshot of the fixed behaviour somewhere, ideally in the linear ticket. That's useful for the reviewer, but also it helps the tester know the new desired behaviour 👍

chris-pollard:
Closing pull request, as there is another PR from [Phoenix](https://github.com/beyondessential/tupaia/pull/3211/) that addresses same issue.

PR #3203, opened 2021-09-12
Add environment variable to front end builds on startup

edmofro:
https://github.com/beyondessential/tupaia/pull/3192 introduced a great new environment banner, but because we have two build paths (one through CI/CD, the other on server startup), the same env var needs to be applied to the other place


kael89 (approved):

edmofro:
Testing this myself, then will merge

edmofro:
Tested, all working :-)

PR #3211, opened 2021-09-13
PHX-45 Allow displaying objects in vizbuilder

Alastair-L:
### Issue #: PHX-45

image

This just transforms all values to primitives, otherwise the table will error, I took a look at the react table docs (which is where this is eventually rendered) and they say that accessor can only return a primitive value. I hope this is ok to slip in, as it will only ever matter when val is an object or a function, where previously it would give an unhelpful error `Something went wrong`


IgorNadj (approved):
LGTM

PR #3212, opened 2021-09-13
WAI-35 refactor lesmis vitals

EMcQ-BES:
### Issue #:
WAI-35
WAI-903

### Changes:

- Replace various levels of vitals reports with just two: `LESMIS_school_vitals` and `LESMIS_entity_vitals`
- Update LESMIS to correctly use this two report setup
- Update LESMIS to actually pull data for country level
- Update the priority district vitals value to pull from the entity attributes


packages/database/src/migrations/20210912223650-LESMISVitalsUseEntityAggregation-modifies-data.js
```diff
@@ -0,0 +1,232 @@
...
+    code,
+    config: reportConfig,
+    permission_group_id: permissionGroupId,
+  });
+};
+
+exports.up = async function (db) {
+  for (const code of OBSOLETE_REPORTS) {
+    await deleteObject(db, 'report', { code });
+  }
```
EMcQ-BES, line 203: I didn't restore these in the `down` because that would involve rewriting the full reports into this migration

packages/database/src/migrations/20210912223650-LESMISVitalsUseEntityAggregation-modifies-data.js
```diff
@@ -0,0 +1,232 @@
...
+};
+
+// New reports
+const ENTITY_VITALS_REPORT_CONFIG = {
+  fetch: {
+    dataElements: [
+      'SDP001',
+      'nosch_total',
+      'nostu_total',
+      ...DEVELOPMENT_PARTNERS.map(dataElement => `${dataElement}_summary`),
```
EMcQ-BES, line 121: With the new indicators all of these can be requested from the `sub_district` level and aggregated up, this means we can use one report for all entity levels (except school which requires specific values)

packages/lesmis/src/api/queries/useEntitiesData.js
```diff
@@ -26,7 +26,7 @@ export const useEntitiesData = (entityCode, params = DEFAULT_PARAMS) => {
 };
 
 const PROJECT_PARAMS = {
-  fields: 'id,child_codes,code,country_code,image_url,name,type,parent_code',
+  fields: 'id,child_codes,code,country_code,image_url,name,type,parent_code,attributes',
```
EMcQ-BES, line 29: I needed to have the attributes to read whether the parent district of a school is a priority district

packages/lesmis/src/api/queries/useVitalsData.js
```diff
@@ -44,18 +44,16 @@ const getParentOfType = (entities, rootEntityCode, type) => {
   }
```
tcaiger, line 44: Wow, this file is soo much simpler now!

tcaiger (approved):
Looks good to me. It's great to have this updated.

PR #3209, opened 2021-09-13
Phx 23 7 day positivity dashboard

Alastair-L:
### Issue #: PHX-23

New config, new branch (don't think it was needed, but oh well).

Can review commit-wise if you like, may be easier

PR #3213, opened 2021-09-13
PHX-22 Positivity overlay

Alastair-L:
### Issue #: PHX-22

Original PR: https://github.com/beyondessential/tupaia/pull/3179


packages/database/src/migrations/20210818014247-AddFijiCovid7DayPositivityIndicator-modifies-data.js
```diff
@@ -0,0 +1,122 @@
...
+      },
+      builder: 'analyticArithmetic',
+    })),
+    aggregation: 'FINAL_EACH_DAY',
+    defaultValues: new Array(7)
+      .fill(0)
+      .map((_, i) => `NumTests${i + 1}Day${i === 0 ? '' : 's'}Ago`)
+      .reduce((acc, code) => ({ ...acc, [code]: 0 }), {}),
+  },
+};
```
rohan-bes, line 91: Didn't these indicators already get added in another PR?  https://github.com/beyondessential/tupaia/pull/3209

packages/database/src/migrations/20210913042439-Add7DayPositivityMapOverlayCovidFiji-modifies-data.js
```diff
@@ -0,0 +1,218 @@
...
+      using: 'sum',
+    },
+    {
+      transform: 'excludeRows',
+      where: '=not(exists($Num_Tests)) or $Num_Tests <= 0 or not(exists($Pos_Tests))',
+    },
+    {
+      transform: 'updateColumns',
+      insert: {
+        value: '=divide($Pos_Tests,$Num_Tests)',
```
rohan-bes, line 84: No need to change, but wanted to let you know that as of the syntax changes, the standard ` / ` operator now can handle `undefined` 🎉 so now it's safe to do stuff like: `$Pos_Tests / $Num_Tests` even if either or both of them is `undefined` 👍

packages/database/src/migrations/20210913042439-Add7DayPositivityMapOverlayCovidFiji-modifies-data.js
```diff
@@ -0,0 +1,218 @@
...
+const FIJI_HIERARCHY_NAMES = {
+  District: 'Division',
+  SubDistrict: 'Sub-division',
+};
+const getConfig = measureLevel => ({
+  code: `FJ_Covid_7_Day_Positivity_Rate_Overlay_${FIJI_HIERARCHY_NAMES[measureLevel]}`,
+  name: `7 day % positivity rate (${FIJI_HIERARCHY_NAMES[measureLevel]})`,
+  frontEndConfig: getFrontEndConfig(measureLevel),
+  reportConfig: getReportConfig(measureLevel),
+  userGroup: 'BES Admin',
```
rohan-bes, line 103:
```suggestion
  userGroup: 'Public',
```
Not sure about this, just judging by the ticket? But I doubt it's meant to be BES Admin...

rohan-bes (changes requested):
Just a request to get rid of the indicators migration as it's already part of a different PR, and I think the usergroup needs to be something else?

Aside from that totally happy with this so once those are fixed feel free to aggressively ping me until I smash that approve button 👍

packages/database/src/migrations/20210913042439-Add7DayPositivityMapOverlayCovidFiji-modifies-data.js
Alastair-L: You are totally right, but all of these vizzes are to be released as BES Admin and then changed post-release to Public when ready.

packages/database/src/migrations/20210913042439-Add7DayPositivityMapOverlayCovidFiji-modifies-data.js
Alastair-L: Oh awesome! Nice one 🎉

packages/database/src/migrations/20210818014247-AddFijiCovid7DayPositivityIndicator-modifies-data.js
Alastair-L: Yes... they did. My hope was that because they are the same commit they would just merge nicely? And if not that we could cross that bridge when we come to it. What should I do instead?

rohan-bes: Ummm if you squash n merged your other commit then I don't think it'll be a clean merge... might be if the files are identical I guess...

I'd usually recommend rebasing in this scenario, to clean up the commit history. Happy to give you a quick explanation if you like?

Otherwise, just test locally first to ensure the merge goes through cleanly?

packages/database/src/migrations/20210913042439-Add7DayPositivityMapOverlayCovidFiji-modifies-data.js
rohan-bes: Ahh interesting, sounds good 👍

rohan-bes (approved):

packages/database/src/migrations/20210818014247-AddFijiCovid7DayPositivityIndicator-modifies-data.js
Alastair-L: Sounds good, how about I rebase whichever one gets merged second, and test it locally as well?

rohan-bes: Sounds good to me! Also, might just mention that indicators can actually be created in the Admin panel, so if you'd wanted to do that it would've been fine. I prefer to do that when I'm sharing an indicator between multiple vizes 👍

Alastair-L: Ah, that sounds like it would have been the best way. I'll keep it in mind for next time!

Alastair-L:
@rohan-bes could you please re-review the last commit? Minimal changes, but would appreciate it 🙏

rohan-bes (approved):
LGTM!

packages/database/src/migrations/20210818014247-AddFijiCovid7DayPositivityIndicator-modifies-data.js
Alastair-L: > might be if the files are identical I guess
For reference, it was a clean merge. Still not sure it was the best way to go commit-history-wise, but always good to know

PR #3214, opened 2021-09-13
RN-81: Added unit tests for entity-server

rohan-bes:
### Issue RN-81:

Long overdue unit tests for the entity-server! These are more of our 'integration style' unit tests, that push data into the database and then use `supertest` to run the entity-server app and hit the endpoints. I went with this approach since entity-server makes heavy use of middlewares, so ensuring that the routing is correctly composed in express is almost as important as the code itself working correctly.

To get some of the modern typescript syntax (4.2+) used in entity-server to compile during the tests I had to bump the `babel-jest` version to 27.x (and consequently `jest` itself, since it seems the versions aren't cross compatible). This lead to a bit of tedious yack shaving process to changing broken integrations with `jest` in a few places.

I also had to delete and re-create the node_modules directory in my local to get the tests to run...

It's always nice when creating unit tests that you catch some bugs in the actual code in the process. There was a couple of changes made to address permissions issues in entity-server that I think are good fixes!


package.json
```diff
@@ -81,8 +83,7 @@
     "cypress-dotenv": "^1.2.2",
     "eslint": "^7.9.0",
     "eslint-plugin-cypress": "^2.11.1",
-    "jest": "^26.1.0",
-    "jest-environment-jsdom-sixteen": "^1.0.3",
```
rohan-bes, line 85: No longer needed as jsdom-16 is default in jest 26+

packages/database/src/modelClasses/Entity.js
```diff
@@ -442,7 +445,13 @@ export class EntityModel extends MaterializedViewLogDatabaseModel {
...
       return Promise.all(relations.map(async r => r.getData()));
     });
-    return Promise.all(relationData.map(async r => this.generateInstance(r)));
+
+    const uniqueEntities = entityArray => {
+      const entitiesById = keyBy(entityArray, 'id');
+      return Object.values(entitiesById);
+    };
+
+    return Promise.all(uniqueEntities(relationData).map(async r => this.generateInstance(r)));
```
rohan-bes, line 454: Bug fix here! Previously if the requested entities had overlapping relatives, we'd return a relative for each entity, leading to duplicates in the responses from entity-server.

packages/entity-server/package.json
```diff
@@ -22,7 +22,7 @@
     "start": "node dist",
     "start-dev": "../../scripts/bash/backendStartDev.sh 9996 -ts",
     "start-verbose": "LOG_LEVEL=debug npm run start-dev",
-    "test": "jest",
+    "test": "jest --runInBand",
```
rohan-bes, line 25: The tests need to run in band since they perform DB operations

packages/entity-server/src/__tests__/__integration__/fixtures.ts
```diff
@@ -0,0 +1,307 @@
...
+    .map(entityCode => getEntityWithFields(entityCode, fields));
+
+export const sortedByCode = (entities: { code: string }[]) =>
+  entities.sort((a, b) => (a.code > b.code ? 1 : -1));
+
+export const PROJECTS = [{ code: 'redblue' }, { code: 'goldsilver' }];
+
+export const COUNTRIES = ['KANTO', 'JOHTO'];
+
+export const ENTITIES: Entity[] = [
```
rohan-bes, line 40: Okay plz don't judge me, but I may have modelled all the dummy data for these tests off the first two pokemon games 😂 I'm a big pokemon nerd, so it brings me great joy, but if reviewer prefers I can make it something less esoteric!

Some really nice features about this data set though. 
- In the first pokemon game, you explore a country which can easily be organised into a hierarchy. 
- In the second pokemon you explore a new country, however you can also travel back to the country which has had some changes since the first game
- This allows us to test some of the trickier corner cases with alternate hierarchies effectively (eg. what happens if an entity in one hierarchy has a different parent in another hierarchy)

packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
```diff
@@ -13,8 +13,12 @@ const throwNoAccessError = (entityCodes: string[]) => {
...
 };
 
-const userCanAccessEntity = (entity: EntityType, allowedCountries: string[]) =>
-  entity.isProject() ||
+const userCanAccessEntity = (
+  entity: EntityType,
+  allowedCountries: string[],
+  rootEntity: EntityType,
+) =>
+  (entity.isProject() && entity.code === rootEntity.code) ||
```
rohan-bes, line 21: Bug fix here! Previously this check would pass if you requested for multiple project entities. This is impossible, since there can only be one project entity in the hierarchy, so checking here to ensure the project entity is the root hierarchy entity

packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
```diff
@@ -46,13 +50,16 @@ const validateEntitiesAndBuildContext = async (
     throwNoAccessError(entityCodes);
   }
 
-  if (!entities.every(entity => userCanAccessEntity(entity, allowedCountries))) {
+  const allowedEntities = entities.filter(entity =>
+    userCanAccessEntity(entity, allowedCountries, rootEntity),
+  );
```
rohan-bes, line 55: I was under the false assumption previously that we'd return only the entities with access if requesting for multiple entities. Turns out we'd thrown an error if any of the entities couldn't be accessed... I believe we wish to only return entities with access though... So I've made that change

packages/web-frontend/jest.config.js
```diff
@@ -1,15 +1,17 @@
...
-  testEnvironment: 'node',
   testURL: 'http://localhost',
   transform: {
-    '^.+\\.(js|jsx)$': '../../node_modules/babel-jest',
+    ...baseConfig.transform,
     '^.+\\.css$': '<rootDir>/config/jest/cssTransform.js',
     '^(?!.*\\.(js|jsx|css|json)$)': '<rootDir>/config/jest/fileTransform.js',
   },
-  transformIgnorePatterns: ['[/\\\\]node_modules[/\\\\].+\\.(js|jsx)$'],
+  transformIgnorePatterns: [`/node_modules/(?!(@?react-leaflet))`],
```
rohan-bes, line 14: Some syntax in react-leaflet was causing the new jest version to fail, so allowed it to be transformed

packages/entity-server/src/__tests__/__integration__/EntityDescendantRoutes.test.ts
```diff
@@ -0,0 +1,215 @@
...
+  let app: TestableEntityServer;
+
+  beforeAll(async () => {
+    app = await setupTestApp();
+    app.grantAccessToCountries(COUNTRIES);
+  });
+
+  afterAll(() => tearDownTestApp(app));
+
+  describe('/hierarchy/<hierarchyName>/<entityCode>/descendants', () => {
```
kael89, line 19: Existing tests (eg in `meditrak-server`) generally use express endpoint syntax in their descriptions

```suggestion
  describe('/hierarchy/:hierarchyName/:entityCode/descendants', () => {
```

I slightly prefer this syntax too, but that's a detail - happy either way!

packages/entity-server/src/__tests__/__integration__/EntityRelationshipsRoutes.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    describe('errors', () => {
+      it('throws error if no type in descendant_filter', async () => {
+        const { text } = await app.get('hierarchy/redblue/redblue/relationships', {
+          query: { descendant_filter: 'name==Kanto' },
+        });
+
+        const error = JSON.parse(text);
+
+        expect(error).toEqual({
+          error: 'Internal server error: descendant_filter:type url parameter must be present',
```
kael89, line 29: As an optional suggestion, I think it's generally good to not use the exact error messages since they are prone to change. My personal rule of thumb is make them as generic as possible while they still convey their basic message.

```suggestion
          error: 'descendant_filter:type url parameter must be present',
```

packages/entity-server/src/__tests__/__integration__/EntityRelationshipsRoutes.test.ts
```diff
@@ -0,0 +1,176 @@
...
+        });
+      });
+
+      it('throws if type in ancestor_filter is non-equality', async () => {
```
kael89, line 33: Love the thorough test cases here! 💯

packages/entity-server/src/__tests__/__integration__/fieldsAndFilters.test.ts
```diff
@@ -0,0 +1,255 @@
...
+
+      const error = JSON.parse(text);
+
+      expect(error).toEqual({
+        error:
+          'Internal server error: Invalid single field requested fake_field, must be one of: id,code,name',
+      });
+    });
+
+    it('throws error if requesting field for not flat property', async () => {
```
kael89, line 68: 👌

packages/entity-server/src/__tests__/__integration__/fieldsAndFilters.test.ts
```diff
@@ -0,0 +1,255 @@
...
+      expect(entity).toBe('Kanto');
+    });
+
+    it('can multiple entities as single fields', async () => {
+      const { text } = await app.post('hierarchy/redblue', {
+        query: { field: 'code' },
+        body: { entities: ['redblue', 'KANTO', 'VIRIDIAN', 'PKMN_TOWER'] },
+      });
+
+      const entity = JSON.parse(text);
```
kael89, line 96:
```suggestion
      const entities = JSON.parse(text);
```

packages/entity-server/src/__tests__/__integration__/fieldsAndFilters.test.ts
```diff
@@ -0,0 +1,255 @@
...
+    it('can fetch an entity with specific fields', async () => {
+      const { text } = await app.get('hierarchy/redblue/KANTO', {
+        query: { fields: 'name,attributes' },
+      });
+
+      const entity = JSON.parse(text);
+      expect(entity).toEqual(getEntityWithFields('KANTO', ['name', 'attributes']));
+    });
+
+    it('can multiple entities with specific fields', async () => {
```
kael89, line 124:
```suggestion
    it('can fetch multiple entities with specific fields', async () => {
```

packages/entity-server/src/__tests__/__integration__/fieldsAndFilters.test.ts
```diff
@@ -0,0 +1,255 @@
...
+    it('throws error for unknown field', async () => {
+      const { text } = await app.get('hierarchy/redblue/KANTO', {
+        query: { filter: 'fake_field==fake_value' },
+      });
+
+      const error = JSON.parse(text);
+
+      expect(error).toEqual({
+        error:
+          'Internal server error: Unknown filter key: fake_field, must be one of: id,code,country_code,name,image_url,type,attributes_type',
```
kael89, line 187: Here is another error message example - apart from the message format, entity props are prone to adjustment in the future. While it's probably simple to update this test case when that happens, I'd use a bit more generic error message like the following:

```suggestion
          'Unknown filter key: fake_field',
```

packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
rohan-bes: It sort of does... however my theory is that we're just making it more lenient here? Previously it would have thrown an error, and given that we don't see that error anywhere then we ought not be encountering this scenario anywhere in practice currently? 

However, all the same this seems like better behaviour to me...

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
```diff
@@ -73,14 +73,6 @@ describe('DataBroker', () => {
...
   describe('pull()', () => {
     describe('input validation', () => {
-      beforeAll(() => {
-        jest.spyOn(global.console, 'warn').mockImplementation();
-      });
-
-      afterAll(() => {
-        global.console.warn.restore();
-      });
-
```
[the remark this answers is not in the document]
rohan-bes: Ah good question! I removed it since I was getting an error about there being no `.restore()` on `global.warn`. I think this is all related to the jest upgrade. Truthfully I should've tracked down exactly why this didn't work anymore (I tried for a bit and then gave up), and how to support it in the new version, but I felt that maybe this wasn't the most important thing? I think it just suppresses a single warning that originates from a test where we do an unhappy path check...

packages/database/src/modelClasses/Entity.js
```diff
@@ -442,7 +445,13 @@ export class EntityModel extends MaterializedViewLogDatabaseModel {
       );
       return Promise.all(relations.map(async r => r.getData()));
     });
-    return Promise.all(relationData.map(async r => this.generateInstance(r)));
+
+    const uniqueEntities = entityArray => {
```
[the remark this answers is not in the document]
rohan-bes: Good call, will change!

packages/entity-server/src/__tests__/testUtilities/setupTestApp.ts
```diff
@@ -0,0 +1,91 @@
...
+import { findOrCreateDummyRecord, getTestModels, EntityHierarchyCacher } from '@tupaia/database';
+
+import { TestModelRegistry } from '../types';
+import { TestableEntityServer } from './TestableEntityServer';
+import {
+  PROJECTS,
+  ENTITIES,
+  ENTITY_HIERARCHIES,
+  ENTITY_RELATIONS,
+} from '../__integration__/fixtures';
```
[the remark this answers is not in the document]
rohan-bes: yeah I like the `__` too!

packages/entity-server/src/__tests__/types.ts
```diff
@@ -0,0 +1,23 @@
...
+} from '@tupaia/database';
+
+export interface TestModelRegistry extends ModelRegistry {
+  readonly entity: EntityModel;
+  readonly entityHierarchy: EntityHierarchyModel;
+  readonly entityRelation: EntityRelationModel;
+  readonly user: UserModel;
+  readonly userEntityPermission: UserEntityPermissionModel;
+  readonly project: ProjectModel;
+}
```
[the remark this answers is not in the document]
rohan-bes: I don't mind ModelRegistry being an opt-in kind of thing, since it helps with information hiding! But it would be really cool for it to be generic or something, eg. `type TestModelRegistry = ModelRegistry<[entity, entityHierarchy, entityRelation, user, userEntityPermission, project]>`

packages/entity-server/src/__tests__/testUtilities/setupTestApp.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const models = getTestModels() as TestModelRegistry;
+const entityCodesToIds: Record<string, string> = {};
+const entityHierarchyNamesToIds: Record<string, string> = {};
```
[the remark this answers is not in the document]
rohan-bes: Good point! I feel I flip flop on which makes more sense to me all the time haha. It's like one of those optical illusions that look like 2 things at once

packages/entity-server/src/__tests__/testUtilities/TestableEntityServer.ts
```diff
@@ -0,0 +1,104 @@
+/**
+ * Tupaia MediTrak
+ * Copyright (c) 2017 Beyond Essential Systems Pty Ltd
+ */
+import * as dotenv from 'dotenv';
+
+/**
+ * Start the server
+ */
+// eslint-disable-next-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
rohan-bes: Yeah I didn't know why this wasn't working... I will take another look, good catch!

packages/entity-server/src/__tests__/testUtilities/TestableEntityServer.ts
```diff
@@ -0,0 +1,104 @@
...
+  `/v${apiVersion}/${endpoint}`;
+
+export const getAuthorizationHeader = () => {
+  const credentials = `${process.env.CLIENT_USERNAME}:${process.env.CLIENT_SECRET}`;
+  return `Basic ${Buffer.from(credentials).toString('base64')}`;
+};
+
+type RequestOptions = { headers?: Record<string, any>; query?: Record<string, any>; body?: any };
+
+export class TestableEntityServer {
```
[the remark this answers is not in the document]
rohan-bes: Definitely! I was a bit worried that I'd need to change this class a lot for supertest in typescript, but it turned out to not be so bad...

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.js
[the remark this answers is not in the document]
kael89: Sure thing - happy to leave this as is. Thanks for the explanation!

packages/entity-server/src/__tests__/testUtilities/setupTestApp.ts
```diff
@@ -0,0 +1,91 @@
...
+const hierarchyCacher = new EntityHierarchyCacher(models);
+
+export const setupTestApp = async () => {
+  await Promise.all(
+    ENTITIES.map(async ({ code, name, country_code, type, ...restOfEntity }) => {
```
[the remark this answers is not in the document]
rohan-bes: Thanks for letting me know about this @kael89! I refactored that util and these tests to use it!

packages/entity-server/src/__tests__/__integration__/fixtures.ts
```diff
@@ -0,0 +1,307 @@
...
+  };
+};
+
+export const getEntitiesWithFields = (entityCodes: string[], fields: (keyof Entity)[]) =>
+  ENTITIES.map(e => e.code)
+    .filter(entityCode => entityCodes.includes(entityCode))
+    .map(entityCode => getEntityWithFields(entityCode, fields));
+
+export const sortedByCode = (entities: { code: string }[]) =>
+  entities.sort((a, b) => (a.code > b.code ? 1 : -1));
```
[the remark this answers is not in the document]
rohan-bes: Thanks for this! I had a hard time finding the correct jest assertion for what I wanted, so I wrote my own logic, but this is exactly what I'm after 🎉 

Have switched to using it 👍

packages/entity-server/src/__tests__/__integration__/fixtures.ts
```diff
@@ -0,0 +1,307 @@
...
+  },
+];
+
+export const ENTITY_HIERARCHIES = [{ name: 'redblue' }, { name: 'goldsilver' }];
+
+export const ENTITY_RELATIONS = [
+  { parent: 'redblue', child: 'KANTO', hierarchy: 'redblue' },
+  { parent: 'KANTO', child: 'PALLET', hierarchy: 'redblue' },
+  { parent: 'PALLET', child: 'BLUE', hierarchy: 'redblue' },
+  { parent: 'KANTO', child: 'VIRIDIAN', hierarchy: 'redblue' },
```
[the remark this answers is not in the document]
rohan-bes: As you wish ❤️

packages/entity-server/src/__tests__/__integration__/EntityDescendantRoutes.test.ts
rohan-bes: Yeah I like it too, switched 👍

packages/entity-server/src/__tests__/__integration__/EntityRelationshipsRoutes.test.ts
rohan-bes: Good call! Switched!

packages/entity-server/src/__tests__/__integration__/permissions.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    app.grantAccessToCountries(['JOHTO']);
+  });
+
+  afterAll(() => tearDownTestApp(app));
+
+  describe('no access', () => {
+    it('throws error if fetching project with no access to any of its countries', async () => {
+      const { text } = await app.get('hierarchy/redblue/redblue', {
+        query: { fields: 'code,name,type' },
+      });
```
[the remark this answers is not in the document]
rohan-bes: Oh ummm probably... but I'm fine with how we're doing it here?

packages/entity-server/src/__tests__/testUtilities/TestableEntityServer.ts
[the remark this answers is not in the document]
rohan-bes: Okay, it's still not working and I still don't know why 😝 

I think it's got something to do with typescript since it's complaining about the types not being listed in the project dependencies... I'm gonna do a big typescript cleanup soon to get rid of all the errors across all packages so is it okay if I do it then?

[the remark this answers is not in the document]
kael89: I see! Could it be that adding `@types/supertest` in the root package.json solves this?

It's a detail so happy to leave this anyway 👍

packages/entity-server/src/__tests__/__integration__/permissions.test.ts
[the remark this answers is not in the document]
kael89: Up to you, I guess it's not a biggie... The change would be:

```ts
   const { error } = await app.get('hierarchy/redblue/redblue', {
        query: { fields: 'code,name,type' },
      });

  expect(error.error).toContain('No access to requested entities: redblue')
```

So you wouldn't have to extract `text` and then parse its json. Happy to leave this as is though

packages/entity-server/src/__tests__/testUtilities/TestableEntityServer.ts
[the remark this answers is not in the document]
rohan-bes: It's there already! Very strange...

rohan-bes:
@kael89 since you were curious 😉 
image

Pretty happy with that 🎉 Still a few more edge cases that could be covered, but majority is tested!

kael89:
> @kael89 since you were curious 😉 test coverage:
> image
> 
> Pretty happy with that 🎉 Still a few more edge cases that could be covered, but majority is tested!

Amazing!!!

kael89 (approved):
Thanks for taking the time to address the suggestions! Kudos for the excellent code coverage 🧪 🚀 

Approving so that you can proceed after CI/CD tests are fixed

PR #3218, opened 2021-09-14
RN-83: Convert 'sum' in report-server over to a mathjs typed function

rohan-bes:
### Issue RN-83:

Few changes involved here:
1. The titular converting of the `sum` function to a mathjs typed function
2. Adding support for extending and overriding mathjs typed function implementations in `ExpressionParser`
3. Switched to importing functions into mathjs with `wrap=true`. This means mathjs will coerce its more esoteric `Matrix` types into simple arrays when executing those functions!
4. File location shuffling and switching `function` unit tests to testing the result of parsing each function than than directly executing it


packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -63,8 +63,10 @@ export class ExpressionParser {
     };
 
     this.customScope = customScope;
-    this.customFunctions = this.getCustomFunctions();
-    this.math.import(this.customFunctions);
+    this.customFunctionNames = Object.keys(this.getCustomFunctions());
+    this.math.import(this.getCustomFunctions(), { wrap: true });
```
rohan-bes, line 67: `wrap: true`
> If true, the functions will be wrapped in a wrapper function which converts data types like Matrix to primitive data types like Array. The wrapper is needed when extending math.js with libraries which do not support these data type. False by default.

https://mathjs.org/docs/reference/functions/import.html

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
```diff
@@ -0,0 +1,172 @@
...
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { TransformParser } from '../../../../reportBuilder/transform/parser';
+
+describe('functions', () => {
+  describe('basic', () => {
+    describe('value', () => {
+      it('returns a string when passed to it', () =>
+        expect(new TransformParser().evaluate("=value('some string')")).toBe('some string'));
```
rohan-bes, line 12: All these tests are basically the same as the existing `functions.test.ts` with a few added. However evoking them via the `TransformParser` is a much better test as it validates how these will evaluate when used in reports

packages/report-server/src/reportBuilder/transform/parser/functions/math.ts
```diff
@@ -0,0 +1,42 @@
...
+  number: (num: number) => num,
+  undefined: (undef: undefined) => undefined,
+  Array: (arr: unknown[]) =>
+    arr.every(item => item === undefined)
+      ? undefined
+      : arr
+          .filter(item => item !== undefined)
+          .map(enforceIsNumber)
+          .reduce((total, item) => total + item, 0),
+});
```
rohan-bes, line 42: I am loving typed functions! mathjs.org/docs/reference/functions/typed.html

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+      it('returns sum if input is at least partially defined array', () =>
+        expect(new TransformParser().evaluate('=sum([undefined, 1])')).toBe(1));
+
+      it('returns sum if inputs are at least partially defined', () =>
+        expect(new TransformParser().evaluate('=sum(undefined, 1)')).toBe(1));
+
+      it('returns input if input is defined', () =>
+        expect(new TransformParser().evaluate('=sum(1)')).toBe(1));
+    });
```
kael89, line 170: It seems very trivial but I suggest adding a test case for summing 2 numbers, and one for summing multiple (eg 4-5) numbers

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        expect(() => new TransformParser().evaluate("=sum([1, 'cat'])")).toThrow());
+
+      it('throws error for non-numeric inputs', () =>
+        expect(() => new TransformParser().evaluate("=sum(1, 'cat')")).toThrow());
+
+      it('returns undefined if input is undefined', () =>
+        expect(new TransformParser().evaluate('=sum(undefined)')).toBe(undefined));
+
+      it('returns undefined if input is array of undefined', () =>
+        expect(new TransformParser().evaluate('=sum([undefined, undefined])')).toBe(undefined));
```
kael89, line 157: Suggestion: it may be a bit subjective but for repeated test cases I often find [it.each](https://jestjs.io/docs/api#testeachtablename-fn-timeout) very helpful, and easier to maintain. Most of all, it encourages you to think of more edge cases, since you can add them very easily!

We have a lot of examples in our code base, eg in [array.test.js](https://github.com/beyondessential/tupaia/blob/dev/packages/utils/src/__tests__/array.test.js#L74). This section describe block would look something like the following:

```ts
describe('sum', () => {
  const testData: [string, string, unknown][] = [
    ['returns undefined if input is undefined', '=sum(undefined)', undefined],
    ['returns undefined if input is array of undefined', '=sum([undefined, undefined])', undefined],
    ['returns undefined if inputs are undefined', '=sum(undefined, undefined)', undefined],
    ['returns sum if input is at least partially defined array', 1],
    // ...etc
  ];

  it.each(testData)('%s', (_, expression: string, expected: unknown) => {
    expect(new TransformParser().evaluate(expression)).toBe(expected);
  });
});

```

kael89 (approved):
Great work @rohan-bes ! It was an interesting introduction to the inner workings of report server transforms.

`report-server` + `math.js` = ♥

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
kael89: On the other hand, your test cases here are already easy to read and elegant, so I wouldn't necessarily "recommend" using `it.each` here. I'd probably do it because I like the concept of [parameterized tests](https://www.baeldung.com/parameterized-tests-junit-5), but that's probably a matter of taste.

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
rohan-bes: Good call, will add!

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
kael89: I agree, it's easy to go overboard with parameterized tests!_

kael89:
Closing this in favour of https://github.com/beyondessential/tupaia/pull/3264

PR #3219, opened 2021-09-14
RN-75 Household vis perform

nick-beyondessential:
Issue: [RN-75] 

To improve performance on mapOverlays with a large number of survey responses, we've introduced a new `presentationOptions` config of `fetchOnlyDisplayLevel` a boolean to work with `displayOnLevel`, when set to true (default: false) will only fetch data at the entity level specified in `displayOnLevel`. i.e. district/sub_district etc instead of default of country. This means more fetches but reduced fetched time for when country level fetch times are prohibitive.
MapOverlays need fetchOnlyDisplayLevel=true and a displayOnLevel type set to see in action.


packages/web-frontend/src/sagas.js
```diff
@@ -884,12 +886,32 @@ function* fetchMeasureInfo(measureId) {
...
+  const { fetchOnlyDisplayLevel = false } = measureParams;
+  const measureDataAncestor = fetchOnlyDisplayLevel
+    ? selectDisplayLevelAncestor(state)
+    : countryCode;
+
+  const shouldFetchCurrentLevel = selectShouldFetchCurrentLevel(state);
+
+  if (!shouldFetchCurrentLevel) {
+    // Don't try and fetch null measures
+    yield put(cancelFetchMeasureData());
```
rohan-bes, line 901: The behaviour of this feels a bit buggy. For instance, if we select a regularly map overlay, and then switch to one where the `displayOnLevel` is lower than the current org unit, it just leaves the previous map overlay showing...

I've tried a couple quick things to see if this is easily remedied, but I think it's a bit tricky... We basically need to yield an action that sets the currently selected measure in the state properly, but doesn't actually fetch data...

Happy to chat on this if ya like 👍

rohan-bes (changes requested):
Nice work @nick-beyondessential ! I'm impressed by how clean this all came out in the end 🎉 

Had a few issues with this branch when I was testing locally, lemme know if you wanna chat about any of it 👍

rohan-bes:
Also, couple q's:
1. What's the massive households json file for?
2. Was there meant to be a migration to set the `fetchOnlyDisplayLevel` key for this overlay? I'm somewhat of the opinion that we should make this new logic the standard behaviour for `displayOnLevel`, but if we don't then I think there needs to be a migration or something...

nick-beyondessential:
Thanks for the review @rohan-bes. 
The two questions above:
1. This was the original migration add household data which has already been applied, I'll delete this file.
2. fetchOnlyDisplayLevel can be set via admin panel. Can be a migration to bundle together. Happy to consider making it standard for displayOnLevel, this way just felt a bit safer with more backwards compatibility with existing overlays.

rohan-bes:
Ahh good point about setting this via Admin Panel, yup that's a much better approach 👍

edmofro:
Closing as stale

PR #3220, opened 2021-09-14
Bump api client / change api-client to use tsc for the builder 

IgorNadj:


nick-beyondessential (approved):
LGTM

packages/server-boilerplate/package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
     "@tupaia/access-policy": "3.0.0",
-    "@tupaia/api-client": "2.0.1",
+    "@tupaia/api-client": "^2.0.1",
```
edmofro, line 29: Ah sorry @IgorNadj can we leave this as exact matching? It means it gets it from the relative path in the mono-repo rather than npm, which is really handy as well as being a bit more efficient.

IgorNadj: Ah sorry that was a mistake! I have to fix something else, will revert it in the same change.

IgorNadj:
Extended this PR to include "change api-client to use tsc for the builder"

Background: https://github.com/beyondessential/tamanu/pull/1409

No real reason to use babel here as far as I can tell, `tsc` gives us everything we need in this package. @tcaiger can you confirm this please?

tcaiger:
> Extended this PR to include "change api-client to use tsc for the builder"
> 
> Background: [beyondessential/tamanu#1409](https://github.com/beyondessential/tamanu/pull/1409)
> 
> No real reason to use babel here as far as I can tell, `tsc` gives us everything we need in this package. @tcaiger can you confirm this please?

Hey Igor, yes I believe you are right that compiling with tsc can replace the need for babel. I don't have first hand experience with it though.

PR #3226, opened 2021-09-15
Hack to get front-end builds running again

rohan-bes:
Front-end builds are currently failing due to not being able to resolve the eslintc react-native package (or something like that).

image

Since the new flow involves new builds deleting the existing builds directories, this means that Tupaia ends up 'Under Maintenance' each time a new build comes in. 

This is a short term hack to get builds working again.

We need to follow up by:
1. Fixing up the eslint dependencies
2. Ensure that if building the frontend packages fails, the build fails and we don't deploy over the existing build

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -132,9 +132,9 @@
     {
       "files": ["packages/meditrak-app/**"],
       "plugins": ["react-native"],
-      "env": {
-        "react-native/react-native": true
-      },
+      // "env": {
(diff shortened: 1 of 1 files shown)


kael89 (approved):
Happy to merge this for now then investigate! Hopefully the linter can still work for `meditrak-app`, or if it doesn't work at least it won't break the build!

PR #3228, opened 2021-09-15
RN-82: Auto refresh analytics when any source table changes

rohan-bes:
### Issue RN-82:
Due to the performance improvements in the latest pg-mv-fast-refresh version, I think we can try allowing the auto refresh behaviour when any source table changes. Previously we were refraining from refreshing some of the more sensitive tables as it seemed to have poor performance and we didn't want to refresh in the middle of a large import (entities for instance). Now that the behaviour seems fast across all tables, we can try always auto refreshing. The benefit of this is that changes will be integrated faster and more reliably, and large log table backlogs are less likely to build up.

### Changes:
- Bump pg-mv-fast-refresh version to include performance improvements
- Now we can try refresh if any of the source tables change

---

### DIFF
--- a/packages/database/src/tests/changeHandlers/AnalyticsRefresher.test.js
+++ b/packages/database/src/tests/changeHandlers/AnalyticsRefresher.test.js
@@ -36,10 +36,13 @@ describe('AnalyticsRefresher', () => {
   const analyticsRefresher = new AnalyticsRefresher(models);
   analyticsRefresher.setDebounceTime(REFRESH_DEBOUNCE_TIME);
 
-  const assertAnalyticsMatch = async expectedAnalytics => {
+  const assertAnalyticsMatch = async (
+    expectedAnalytics,
+    codesToFetch = TEST_DATA.question.map(q => q.code),
+  ) => {
     await models.database.waitForAllChangeHandlers();
     const analyticsInDb = await models.analytics.find({
-      data_element_code: TEST_DATA.question.map(q => q.code),
+      data_element_code: codesToFetch,
     });
     const formattedAnalyticsInDb = analyticsInDb.map(analytic => {
       const formattedAnalytic = {};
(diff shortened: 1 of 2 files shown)


packages/database/src/tests/changeHandlers/AnalyticsRefresher.test.js
```diff
@@ -190,4 +193,77 @@ describe('AnalyticsRefresher', () => {
     });
     await assertAnalyticsMatch(updatedAnalytics);
   });
+
+  it('refreshes analytics table if surveys are updated', async () => {
+    // Update an answer, make sure analytic is updated
```
kael89, line 198:
```suggestion
    // Update a survey, make sure analytic is updated
```

similarly for the other test cases

packages/database/src/tests/changeHandlers/AnalyticsRefresher.test.js
```diff
@@ -190,4 +193,77 @@ describe('AnalyticsRefresher', () => {
...
+    await assertAnalyticsMatch(updatedAnalytics);
+  });
+
+  it('refreshes analytics table if entities are updated', async () => {
+    // Update an answer, make sure analytic is updated
+    await models.entity.update(
+      {
+        id: 'entity001_test',
+      },
+      { code: 'HAPPIER_LAND', name: 'The happiest land of all' },
```
kael89, line 223: 🌈 🌞 🎶

packages/database/src/tests/changeHandlers/AnalyticsRefresher.test.js
```diff
@@ -190,4 +193,77 @@ describe('AnalyticsRefresher', () => {
...
+  it('refreshes analytics table if data sources are updated', async () => {
+    // Update an answer, make sure analytic is updated
+    await models.dataSource.update(
+      {
+        id: 'dataSource001_test',
+      },
+      { service_type: 'dhis' },
+    );
+    const updatedAnalytics = ANALYTICS.filter(analytic => analytic.data_element_code !== 'Q001');
+    await assertAnalyticsMatch(updatedAnalytics);
```
kael89, line 267: Nice test case!

kael89 (approved):
Nice one - I've got a good feeling about your analytics improvements! Handy tests too

PR #3233, opened 2021-09-16
Fail codeship build on any failure of internal scripts

edmofro:
### Issue #:
WAI-949

### Changes:
Standardised the use of `-e` (which exits if an internal script returns a non-zero exit code) in almost all bash scripts, so we avoid running into this again


packages/devops/scripts/ci/testInternalDependencies.sh
```diff
@@ -1,13 +1,11 @@
...
 for PACKAGE in $(${DIR}/../../../../scripts/bash/getInternalDependencies.sh); do
     # skip the following packages - they get tested separately as they require db access
     if [[ "$PACKAGE" == "database" || "$PACKAGE" == "data-api" || "$PACKAGE" == "auth" || "$PACKAGE" == "indicators" || "$PACKAGE" == "server-boilerplate" ]]; then
         continue
     fi
     echo Testing ${PACKAGE}
-    if ! yarn workspace @tupaia/${PACKAGE} test; then
-        exit 1 # the tests for this internal depencency failed
-    fi
+    yarn workspace @tupaia/${PACKAGE} test
```
edmofro, line 10: No longer need the conditional given that the `-e` means it'll exit the script early anyway

scripts/bash/getInternalDependencies.sh
```diff
@@ -1,11 +1,13 @@
...
+if [ "$1" != "" ]; then
+  # pop the package_path off, and interpret the rest as dependencies that have been checked earlier
+  # in the recursion
+  package_path=$1
+  shift
+fi
 
-# pop the package_path off, and interpret the rest as dependencies that have been checked earlier
-# in the recursion
-shift
```
edmofro, line 8: This was actually throwing up a `1` exit code if no `package_path` was provided, so now we do it conditionally

scripts/bash/validate.sh
```diff
@@ -1,12 +1,7 @@
...
+#!/bin/bash -e
 
 DIR=$(dirname "$0")
 . ${DIR}/../../packages/devops/scripts/ci/utils.sh
 
 yarn -s workspace @tupaia/devops -s validate-branch-name
 yarn -s workspace @tupaia/devops -s validate-tests
-if [[ $? > 0 ]]; then
-    log_warn "Exclusive (.only) tests found in your code. The CI/CD build will fail if you push those"
-fi
```
edmofro, line 12: Both previously (with the `set -e`) and now (with the `-e` in the shebang) this code is unreachable

packages/data-api/scripts/buildAnalyticsMaterializedView.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
kael89, line 1: I always used `set -e` as a separate command - didn't know you can also use it this way, nice!

packages/devops/scripts/ci/testInternalDependencies.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -le
```
kael89, line 1: `-l` was not used here - just double checking that its addition is on purpose

By the way I'm not familiar with `-l`, what does it do?

scripts/bash/validate.sh
kael89: Good catch - I guess we shouldn't need this anyway, since `validateTests` logs more specific errors

kael89 (commented):
Looks good @edmofro ! Just have a question about `-l`

packages/devops/scripts/ci/testInternalDependencies.sh
edmofro: Jeez you've got good eyes! This was me testing that they'd work together, it's not harmful but also not required here, just tells bash to run the script in a regular user session (used during startup scripts called from crontab to make sure PATH, environment variables etc. are set up as normal)

edmofro: Removed :+1:

kael89 (approved):

PR #3236, opened 2021-09-16
Allow api-client build errors

IgorNadj:
Is causing CI build to fail now that we are stricter about this. This is a workaround, proper fix incoming which fixes the type errors.

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -6,7 +6,7 @@
   "types": "dist/index.d.ts",
   "license": "SEE LICENSE IN LICENSE",
   "scripts": {
-    "build": "tsc; echo Done",
+    "build": "tsc; echo Done TODO fix type errors",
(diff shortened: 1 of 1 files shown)


packages/api-client/package.json
```diff
@@ -6,7 +6,7 @@
   "types": "dist/index.d.ts",
   "license": "SEE LICENSE IN LICENSE",
   "scripts": {
-    "build": "tsc",
+    "build": "tsc; echo Done",
```
edmofro, line 9:
```suggestion
    "build": "tsc; echo TODO fix type errors",
```

edmofro (approved):
lgtm

packages/api-client/package.json
IgorNadj: Added 👍

edmofro (approved):

PR #3237, opened 2021-09-16
WAI-950: Write temporary files to system tmp directory

edmofro:
### Issue #:
- WAI-950

### Changes:
- Uploads and exports are written to the temporary folder defined by the OS

Note that on most systems this means they will be deleted on restart, so I've also tweaked the instructions for setting up a new EC2 instance to preserve them for 20 days

### DIFF
--- a/packages/devops/Set up server.md
+++ b/packages/devops/Set up server.md
@@ -425,6 +425,23 @@ export LASTPASS_PASSWORD=xxx
 
 (get actual credentials from LastPass)
 
+# Temporary files
+
+Add a rule so that temporary files (like uploads and exports) are only deleted after 20 days, rather
+than immediately on any reboot
+
+`sudo vi /etc/tmpfiles.d/tmp.conf`
+
+Paste in
+
+```
+#/etc/tmpfiles.d/tmp.conf
+
+d /tmp 1777 root root 20d
+```
+
+And save.
+
 # ssh (optional)
 
 ### On local machine, set .ssh/config to ignore known_hosts for dev ip (it changes every day so gets annoying to remove and re-add it)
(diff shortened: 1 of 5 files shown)


packages/meditrak-server/src/apiV2/export/getExportPathForUser.js
```diff
@@ -3,10 +3,11 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 import fs from 'fs';
+import { getTempDirectory } from '../../utilities';
 
-const EXPORT_BASE_PATH = 'exports';
 export const getExportPathForUser = userId => {
```
kael89, line 8: I think this will also work:

```js
export const getExportPathForUser = userId => getTempDirectory(`exports/${userId}`);
```

... but don't bother changing it if we want to merge a fix quickly!

packages/devops/Set up server.md
```diff
@@ -425,6 +425,23 @@ export LASTPASS_PASSWORD=xxx
...
+than immediately on any reboot
+
+`sudo vi /etc/tmpfiles.d/tmp.conf`
+
+Paste in
+
+```
+#/etc/tmpfiles.d/tmp.conf
+
+d /tmp 1777 root root 20d
```
kael89, line 440: Didn't know you could define this in the OS level nice one!!

kael89 (approved):
Looks good!

packages/meditrak-server/src/apiV2/export/getExportPathForUser.js
edmofro: Ah yes, nice!

edmofro: Tested and confirmed it worked, thanks Kostas :-)

PR #3238, opened 2021-09-16
WAI-951: Minimise likelihood of concurrent builds conflicting

edmofro:
### Issue #:
- WAI-951

### Changes:
Use a specific `incoming_builds` folder, suffixed with the CI build timestamp. There is still the opportunity for clashes, if they both hit the specific bit of the push that moves the unzipped folder over the top of the `tupaia` at the same time. That's a relatively fast step though (I'd say a second or two), whereas the entire `Push builds` step takes around 10 minutes, so it will greatly minimise the chance of conflicts causing issues

### DIFF
--- a/packages/devops/scripts/ci/pushBuilds.sh
+++ b/packages/devops/scripts/ci/pushBuilds.sh
@@ -3,15 +3,16 @@ set -x
 
 PACKAGE=$1
 DIR=$(dirname "$0")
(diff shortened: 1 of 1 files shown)


IgorNadj (approved):
LGTM! 💯

edmofro:
Already tested, merging to dev :-)

PR #3239, opened 2021-09-16
Hotfix release 2021-37-1

tcaiger:


edmofro (approved):
Not exactly sure why we were trying to use `.color` before, but assuming you do, the change makes sense to me!

PR #3240, opened 2021-09-16
Ignore non-zero status codes from pm2 delete all

edmofro:
The startup script is now running a stricter mode, where any failures of internal steps will fail the whole script. On startup, there are generally no processes running, so `pm2 delete all` throws a non zero exit code. This ignores that.

### DIFF
--- a/packages/devops/scripts/deployment/deployPackages.sh
+++ b/packages/devops/scripts/deployment/deployPackages.sh
@@ -4,13 +4,11 @@
 echo "Existing processes:"
 pm2 list
 echo "Deleting existing backend processes..."
-pm2 delete all
+pm2 delete all || : # return 0 exit code even if delete failed because there were no running processes
 echo "Deleting existing frontend builds..."
 cd ${HOME_DIRECTORY}
-# delete all "served_build" folders, which is where most front ends are served from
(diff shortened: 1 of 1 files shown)


packages/devops/scripts/deployment/deployPackages.sh
```diff
@@ -4,13 +4,11 @@
...
-pm2 delete all
+pm2 delete all || : # return 0 exit code even if delete failed because there were no running processes
 echo "Deleting existing frontend builds..."
 cd ${HOME_DIRECTORY}
-# delete all "served_build" folders, which is where most front ends are served from
-find ./packages -type d -name "served_build" -maxdepth 2 -exec rm -rf {} \;
-# delete "builds" from web-frontend, which is where those builds are served from
-rm -rf ./packages/web-frontend/builds
+# delete all "build" folders, which is where front ends are served from
+find ./packages -type d -name "build" -maxdepth 2 -exec rm -rf {} \;
```
edmofro, line 11: Snuck this in too - the result of a soft conflict, where the name of the `build` folder had changed on an incoming branch, and in the meantime we'd added the early delete using the old folders on dev

EMcQ-BES (approved):
lgtm

PR #3242, opened 2021-09-16
[HOT FIX]: Don't consider CI build time a production environment

edmofro:
The test suite will bail early if it gets run in a "production environment". Now that we're pulling the appropriate .env files for the current branch in CI/CD, builds to `master` get the production environment variables, so the tests can't run, even on the build machine. This tweaks what is considered a "production environment" to exclude when we are inside a CI/CD process on codeship.

See [this failed build](https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/e3edb781-ba9f-489c-9881-5fc419e97803?component=step_Validate_and_test_Test_Test_meditrak-server) for context

### DIFF
--- a/packages/meditrak-server/src/devops/getIsProductionEnvironment.js
+++ b/packages/meditrak-server/src/devops/getIsProductionEnvironment.js
@@ -3,4 +3,5 @@
  * Copyright (c) 2017 Beyond Essential Systems Pty Ltd
  */
 
-export const getIsProductionEnvironment = () => process.env.IS_PRODUCTION_ENVIRONMENT === 'true';
+export const getIsProductionEnvironment = () =>
+  process.env.IS_PRODUCTION_ENVIRONMENT === 'true' && !process.env.CI_BUILD_ID;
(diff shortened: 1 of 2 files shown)


packages/meditrak-server/src/devops/getIsProductionEnvironment.js
```diff
@@ -3,4 +3,5 @@
  * Copyright (c) 2017 Beyond Essential Systems Pty Ltd
  */
 
-export const getIsProductionEnvironment = () => process.env.IS_PRODUCTION_ENVIRONMENT === 'true';
+export const getIsProductionEnvironment = () =>
+  process.env.IS_PRODUCTION_ENVIRONMENT === 'true' && !process.env.CI_BUILD_ID;
```
rohan-bes, line 7: Do we need to change this in `web-config-server/src/utils/getIsProductionEnvironment.js` too?

rohan-bes (approved):
I think this is looking good to me! Just a comment about whether we need this in the web-config-server one too?

packages/meditrak-server/src/devops/getIsProductionEnvironment.js
edmofro: We 100% do! Great catch :+1:

rohan-bes: Weird the web-config-server build passed on that build you linked... but it should have failed??? Maybe the env var isn't being set correctly for web-config-server container...

PR #3250, opened 2021-09-20
[HOT FIX] WAI-952: Add remote tracking of correct branch, if it exists

edmofro:
### Issue #:
- WAI-952


kael89 (approved):
Nice fix!

edmofro:
I've tested this through calls to it using branches that both exist and do not, so will merge straight to master

kael89:
@edmofro should we merge this in `dev` too? (Even if it doesn't affect its functionality for now)

edmofro:
I'm just debugging dev and feature deployment startups and would rather hold off until after that

kael89:
> I'm just debugging dev and feature deployment startups and would rather hold off until after that

Sure thing!

PR #3252, opened 2021-09-20
WAI 16 Gross Enrolment Ration Map Overlays

tcaiger:
### Changes:

- add mapOverlay utils for migrations
- add LESMIS GER map overlays

---

### Screenshots:
See issue


packages/database/src/utilities/mapOverlays.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+export const getMapOverlayGroupId = async (db, code) => {
```
EMcQ-BES, line 6: There's a function under `database/src/utilities/migration.js` called `codeToId` which does this generically across any table

packages/database/src/utilities/mapOverlays.js
```diff
@@ -0,0 +1,19 @@
...
+  const results = await db.runSql(`select "id" from "map_overlay_group" where "code" = '${code}';`);
+
+  if (results.rows.length > 0) {
+    return results.rows[0].id;
+  }
+
+  throw new Error('MapOverlayGroup not found');
+};
+
+export const getPermissionGroupId = async (db, name) => {
```
EMcQ-BES, line 16: I think this exact function has been declared individually in about 20 different migrations, so probably a good call to split it out to a util function, but I wouldn't expect to find this in a file called `mapOverlays` 🤔 

I reckon get rid of the mapOverlay function for the generic codeToId and then rename this file to match this util function

EMcQ-BES (changes requested):
The main body of the migration looks good, just a couple notes on the utils you added.

EMcQ-BES (approved):

packages/database/src/utilities/mapOverlays.js
tcaiger: Thanks for that. It's a handy util to be aware of.

PR #3253, opened 2021-09-20
PHX-47: Allow using org unit names in report-server analytics

kael89:
### Issue #:
PHX-47

Based off [rn-83-convert-sum-to-typed-func](https://github.com/beyondessential/tupaia/tree/rn-83-convert-sum-to-typed-func) since it has the latest in `report-server`

### DIFF
--- a/packages/expression-parser/src/__tests__/expression-parser/ExpressionParser.test.js
+++ b/packages/expression-parser/src/__tests__/expression-parser/ExpressionParser.test.js
@@ -62,22 +62,4 @@ describe('ExpressionParser', () => {
       expect(functions).toStrictEqual(expected);
     });
   });
-
-  describe('Using custom calculation prefix', () => {
-    const parser = new (class extends ExpressionParser {
-      CALCULATION_PREFIX = '=';
-    })();
-
-    it('getVariables()', () => {
-      const variables = parser.getVariables('=x + y');
-      const expected = ['x', 'y'];
-      expect(variables).toStrictEqual(expected);
-    });
-
-    it('getFunctions()', () => {
-      const functions = parser.getFunctions('=cos(90) + sin(90)');
-      const expected = ['cos', 'sin'];
-      expect(functions).toStrictEqual(expected);
-    });
-  });
 });
--- a/packages/expression-parser/src/__tests__/expression-parser/ExpressionParser.test.js
+++ b/packages/expression-parser/src/__tests__/expression-parser/ExpressionParser.test.js
@@ -25,4 +25,59 @@ describe('ExpressionParser', () => {
       expect(result).toEqual(0);
     });
   });
+
+  describe('getVariables()', () => {
+    const testData = [
+      ['single variable', 'x', ['x']],
+      ['multiple variables', '2 * x + y + 1', ['x', 'y']],
+      ['expression with parentheses', '(x + y) / z', ['x', 'y', 'z']],
+      ['expression with functions', 'sqrt(cos(x)^2 + sin(y)^2)', ['x', 'y']],
+      ['variable is repeated', '(x - y) / x', ['x', 'y']],
+      ['longer variable names', 'alpha + beta - delta', ['alpha', 'beta', 'delta']],
+      ['left side variable', 'x = 1', ['x']],
+      ['variables at both sides', 'x = y', ['x', 'y']],
+    ];
+
+    const parser = new ExpressionParser();
+
+    it.each(testData)('%s', (_, expression, expected) => {
+      const variables = parser.getVariables(expression);
+      expect(variables).toStrictEqual(expected);
+    });
+  });
+
+  describe('getFunctions()', () => {
+    const testData = [
+      ['single function', 'cos(90)', ['cos']],
+      ['multiple functions', 'cos(90) + sin(90)', ['cos', 'sin']],
+      ['expression with parentheses', '(cos(3) + sin(30)) / sqrt(4)', ['cos', 'sin', 'sqrt']],
+      ['same function is repeated', 'sin(45) / sin(30)', ['sin']],
+      ['two-sided expression', 'x = sin(45)', ['sin']],
+    ];
+
+    const parser = new ExpressionParser();
+
+    it.each(testData)('%s', (_, expression, expected) => {
+      const functions = parser.getFunctions(expression);
+      expect(functions).toStrictEqual(expected);
+    });
+  });
+
+  describe('Using custom calculation prefix', () => {
+    const parser = new (class extends ExpressionParser {
+      CALCULATION_PREFIX = '=';
+    })();
+
+    it('getVariables()', () => {
+      const variables = parser.getVariables('=x + y');
+      const expected = ['x', 'y'];
+      expect(variables).toStrictEqual(expected);
+    });
+
+    it('getFunctions()', () => {
+      const functions = parser.getFunctions('=cos(90) + sin(90)');
+      const expected = ['cos', 'sin'];
+      expect(functions).toStrictEqual(expected);
+    });
+  });
 });
--- a/packages/expression-parser/src/expression-parser/ExpressionParser.js
+++ b/packages/expression-parser/src/expression-parser/ExpressionParser.js
@@ -15,14 +15,6 @@ import { create, all, factory } from 'mathjs';
  */
 
 /**
- * @typedef {Object} FactoryParams
- * @property {string[]} dependencies
- * @property {(dependencies: Record<string, Function>) => Function} create
- *
- * @see https://mathjs.org/docs/core/extension.html#factory-functions
- */
-
-/**
  * Usage:
  * const expressionParser = new ExpressionParser();
  *
@@ -57,12 +49,6 @@ const ADDITIONAL_ALPHA_CHARS = ['@'];
 
 export class ExpressionParser {
   /**
-   * Override in child classes to use a custom prefix
-   * @protected
-   */
-  CALCULATION_PREFIX = '';
-
-  /**
    * Can pass in a custom scope
    * @param {Scope} customScope
    */
@@ -83,13 +69,13 @@ export class ExpressionParser {
     this.validExpressionCache = new Set();
   }
 
-  isCalculation = expression =>
-    expression.length > 0 && expression.startsWith(this.CALCULATION_PREFIX);
-
-  readExpression = expression =>
-    this.CALCULATION_PREFIX
-      ? expression.replace(new RegExp(`^${this.CALCULATION_PREFIX}`), '')
-      : expression;
+  /**
+   * Can override in child classes to allow custom expression formats
+   * @protected
+   */
+  readExpression(input) {
+    return input;
+  }
 
   /**
    * @param {string} name
@@ -105,11 +91,13 @@ export class ExpressionParser {
     }
   }
 
-  isFunctionName = name => this.customFunctionNames.includes(name) || this.isBuiltInFunction(name);
+  isFunctionName(name) {
+    return this.customFunctionNames.includes(name) || this.isBuiltInFunction(name);
+  }
 
   /**
    * Return the variable names in an expression.
-   * @param {string} expression
+   * @param {*} expression
    */
   getVariables(expression) {
     return this.extractSymbols(expression, node => !this.isFunctionName(node.name));
@@ -117,7 +105,7 @@ export class ExpressionParser {
 
   /**
    * Return the function names in an expression.
-   * @param {string} expression
+   * @param {*} expression
    */
   getFunctions(expression) {
     return this.extractSymbols(expression, node => this.isFunctionName(node.name));
@@ -136,7 +124,7 @@ export class ExpressionParser {
 
   /**
    * Validate an expression and throw an error if it's invalid.
-   * @param {string} expression
+   * @param {*} expression
    */
   validate(expression) {
     const expr = this.readExpression(expression);
@@ -157,7 +145,7 @@ export class ExpressionParser {
 
   /**
    * Evaluate an expression. Also validate the expression beforehand
-   * @param {string} expression
+   * @param {*} expression
    */
   evaluate(expression) {
     const expr = this.readExpression(expression);
@@ -169,7 +157,7 @@ export class ExpressionParser {
    * Evaluate an expression and cast boolean results to 1 or 0. Also validate the expression beforehand.
    *
    * Note this could also return a string, array, or object (potentially others too)
-   * @param {string} expression
+   * @param {*} expression
    */
   evaluateToNumber(expression) {
     const expr = this.readExpression(expression);
@@ -250,6 +238,6 @@ export class ExpressionParser {
   }
 
   factory(...args) {
-    return factory(...args);
+    return this.math.factory(...args);
   }
 }
--- a/packages/expression-parser/src/expression-parser/ExpressionParser.js
+++ b/packages/expression-parser/src/expression-parser/ExpressionParser.js
@@ -3,7 +3,24 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-import { create, all } from 'mathjs';
+import { create, all, factory } from 'mathjs';
+
+/**
+ * @typedef {Object} Scope
+ * @property {(s: string) => unknown} get
+ * @property {(s: string, v: unknown) => void} set
+ * @property {(s: string) => boolean} has
+ * @property {(s: string) => void} delete
+ * @property {() => void} delete
+ */
+
+/**
+ * @typedef {Object} FactoryParams
+ * @property {string[]} dependencies
+ * @property {(dependencies: Record<string, Function>) => Function} create
+ *
+ * @see https://mathjs.org/docs/core/extension.html#factory-functions
+ */
 
 /**
  * Usage:
@@ -40,8 +57,12 @@ const ADDITIONAL_ALPHA_CHARS = ['@'];
 
 export class ExpressionParser {
   /**
-   * @typedef { { get: (s: string) => unknown, set: (s: string, v: unknown) => void, has: (s: string) => boolean, keys: () => string[], delete: (s: string) => void, clear: () => void } } Scope
-   *
+   * Override in child classes to use a custom prefix
+   * @protected
+   */
+  CALCULATION_PREFIX = '';
+
+  /**
    * Can pass in a custom scope
    * @param {Scope} customScope
    */
@@ -62,6 +83,14 @@ export class ExpressionParser {
     this.validExpressionCache = new Set();
   }
 
+  isCalculation = expression =>
+    expression.length > 0 && expression.startsWith(this.CALCULATION_PREFIX);
+
+  readExpression = expression =>
+    this.CALCULATION_PREFIX
+      ? expression.replace(new RegExp(`^${this.CALCULATION_PREFIX}`), '')
+      : expression;
+
   /**
    * @param {string} name
    * @return bool
@@ -76,34 +105,46 @@ export class ExpressionParser {
     }
   }
 
+  isFunctionName = name => this.customFunctionNames.includes(name) || this.isBuiltInFunction(name);
+
   /**
-   * Return the variable names from an expression.
-   * @param {*} expression
+   * Return the variable names in an expression.
+   * @param {string} expression
    */
   getVariables(expression) {
-    this.validate(expression);
-    const nodeTree = this.math.parse(expression);
-    const variables = nodeTree
-      .filter(
-        node =>
-          node.isSymbolNode &&
-          !this.customFunctionNames.includes(node.name) &&
-          !this.isBuiltInFunction(node.name),
-      )
+    return this.extractSymbols(expression, node => !this.isFunctionName(node.name));
+  }
+
+  /**
+   * Return the function names in an expression.
+   * @param {string} expression
+   */
+  getFunctions(expression) {
+    return this.extractSymbols(expression, node => this.isFunctionName(node.name));
+  }
+
+  extractSymbols(expression, extractor) {
+    const expr = this.readExpression(expression);
+    this.validate(expr);
+    const nodeTree = this.math.parse(expr);
+    const items = nodeTree
+      .filter(node => node.isSymbolNode && extractor(node))
       .map(({ name }) => name);
-    return [...new Set(variables)];
+
+    return [...new Set(items)];
   }
 
   /**
    * Validate an expression and throw an error if it's invalid.
-   * @param {*} expression
+   * @param {string} expression
    */
   validate(expression) {
-    if (this.validExpressionCache.has(expression)) {
+    const expr = this.readExpression(expression);
+    if (this.validExpressionCache.has(expr)) {
       return;
     }
 
-    const nodeTree = this.math.parse(expression);
+    const nodeTree = this.math.parse(expr);
     nodeTree.traverse(node => {
       if (node.isOperatorNode && node.op === '*' && node.implicit) {
         throw new Error(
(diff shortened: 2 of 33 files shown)


packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -62,6 +83,14 @@ export class ExpressionParser {
     this.validExpressionCache = new Set();
   }
 
+  isCalculation = expression =>
+    expression.length > 0 && expression.startsWith(this.CALCULATION_PREFIX);
+
+  readExpression = expression =>
+    this.CALCULATION_PREFIX
+      ? expression.replace(new RegExp(`^${this.CALCULATION_PREFIX}`), '')
+      : expression;
```
rohan-bes, line 92: Hmmm while I'm not opposed to us doing this, the whole `=` at the start of expressions is really just a report-server thing for now... Indicators and Meditrak App Arithmetic questions both use expression parser, and don't append the `=` to the start to indicate it's a calculation.

To me it feels appropriate to keep all logic about this encapsulated in TransformParser until we've standardised the behaviour across all users of the ExpressionParser?

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -3,7 +3,24 @@
...
+import { create, all, factory } from 'mathjs';
+
+/**
+ * @typedef {Object} Scope
+ * @property {(s: string) => unknown} get
+ * @property {(s: string, v: unknown) => void} set
+ * @property {(s: string) => boolean} has
+ * @property {(s: string) => void} delete
+ * @property {() => void} delete
+ */
```
rohan-bes, line 15: ❤️

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -205,4 +248,8 @@ export class ExpressionParser {
   getFunctionOverrides() {
     return {};
   }
+
+  factory(...args) {
+    return factory(...args);
```
rohan-bes, line 253:
```suggestion
    return this.math.factory(...args);
```
Thoughts on this? It will allow us to use the factory function from our existing math instance, which means if we've already imported stuff into that math instance we can use it (custom types for example)

packages/expression-parser/src/__tests__/expression-parser/ExpressionParser.test.js
```diff
@@ -25,4 +25,59 @@ describe('ExpressionParser', () => {
...
+      const expected = ['x', 'y'];
+      expect(variables).toStrictEqual(expected);
+    });
+
+    it('getFunctions()', () => {
+      const functions = parser.getFunctions('=cos(90) + sin(90)');
+      const expected = ['cos', 'sin'];
+      expect(functions).toStrictEqual(expected);
+    });
+  });
```
rohan-bes, line 82: Nice tests!!!

rohan-bes (changes requested):
Awesome stuff @kael89! Really nice work 🎉 

Few comments on naming and some code structure suggestions, but am really happy with things overall!

packages/report-server/src/reportBuilder/context/buildContext.ts
```diff
@@ -0,0 +1,71 @@
...
+    reqContext.hierarchy,
+    getUniqueEntries(orgUnitCodes),
+    { fields: ['code', 'name'] },
+  );
+
+  return keyBy(entities, 'code');
+};
+
+const contextBuilders: Record<ContextProp, ContextBuilder<ContextProp>> = {
+  orgUnitMap: buildOrgUnitMap,
```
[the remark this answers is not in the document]
kael89: Oh yes, good call, happy to use another name! I can use something simple eg `orgUnits`

packages/report-server/src/reportBuilder/context/extractContextProps.ts
```diff
@@ -0,0 +1,57 @@
...
+    .map(([, config]) => config.contextProps)
+    .flat();
+
+  return getUniqueEntries(contextProps);
+};
+
+export const extractContextProps = (transform: unknown): ContextProp[] => {
+  if (!Array.isArray(transform)) {
+    return [];
+  }
```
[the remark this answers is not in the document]
kael89: Yes that's what I thought too! An approach would be:

1. Expand `this.config.transform` to an array of objects (eg by inlining aliases to objects)
2. Validate the expanded transform
3. Use the expanded transform to find all functions that use context
4. Use the expanded transform to build the output

So validation would happen first, and the rest of the functions would expected a validated & expanded `Transform` object as an input. This will make function extraction simpler, since the `transform` in its input will have a more specific schema.

We use a similar approach in `indicators` (we expand the aggregation input)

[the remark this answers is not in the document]
kael89: I can submit a PR implementing the above at a later stage, but yeah for now let's proceed with what we have

packages/report-server/src/reportBuilder/context/extractContextProps.ts
```diff
@@ -0,0 +1,57 @@
...
+        title,
+        description,
+        ...restOfTransformParams
+      } = transformStep;
+
+      return Object.values(restOfTransformParams).map(param => {
+        return typeof param !== 'object' ? [param] : Object.entries(param);
+      });
+    })
+    .flat(3)
```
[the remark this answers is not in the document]
kael89: Yeah good point... I guess there is no harm using a bigger value here (eg `5`, even 10!), another solution would be using [lodash.flattenDeep](https://lodash.com/docs/4.17.15#flattenDeep).

I'm not sure why the default implementation uses `1` as depth... I'd assume it is simple/performant to flatten with a generous depth by default (similar to how recursions use a max call stack size)

packages/report-server/src/reportBuilder/context/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './buildContext';
+export * from './types';
```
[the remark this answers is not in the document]
kael89: Haha I am happy to respect and use the existing code style in the package! I generally try to do that (although habit is strong)

Let me know what you prefer!

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -131,6 +152,21 @@ export class TransformParser extends ExpressionParser {
...
+        fnName,
+        this.factory(fnName, ['getContext'], create),
+      ]),
+    );
+
+    return { functions, dependencies };
+  }
+
   public static isExpression(expression: string) {
     return expression.startsWith('=');
```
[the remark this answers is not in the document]
kael89: Actually we don't need this method anymore, since `ExpressionParser` handles this - but I forgot to update the child class! Thanks for catching this

packages/expression-parser/src/expression-parser/ExpressionParser.js
kael89: I see what you mean... I just tried to avoid having to override each method in `TransformParser`, to make them aware of the `=` symbol. Not from laziness, but trying to figure out a correct abstraction for this concept...

It seems to me that if a child class does alter how expressions (or a subset of them, eg `calculations`) are defined, then the parent class should expect and allow that... If not, then we are "violating" the contract of the parent class, and is an indication that we shouldn't do it in the first place.

For this specific case: an expression parser does depend on symbols to do its job. Thus, it seems fair to me that a specific implementation of the parser, either via inheritance or via constructor configuration, could use its symbols of choice.

packages/expression-parser/src/expression-parser/ExpressionParser.js
kael89: Good call!

packages/report-server/src/reportBuilder/transform/parser/functions/context.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Context, ContextProp } from '../../../context';
+
+/**
+ * `contextProps`: props that will be required by the function
+ * `create`: factory method that creates the function by injecting the context
+ */
+type ContextParams = {
+  contextProps: ContextProp[];
+  create: (deps: { getContext: () => Context }) => (...args: any[]) => unknown;
+};
```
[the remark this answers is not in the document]
kael89: I used `params` since it is used in a seemingly similar vein throughout the project: get an object (eg from `transform`/`fetch` config), and construct a function (the `transform`/`fetch` function). I guess that my bias towards "shorter names where possible" played a role too!

In retrospect, those are not quite the same thing, so happy to change that - it feels like equating oranges and apples here 😄 

About `contextDependencies`: that was the term I used initially. I also tried to inject the `context` object directly (instead of a getter for it: `getContext`), but math.js wouldn't let me. I think I quite like `contextProps` eventually, because that's what this is: a prop of the context - the fact that `context` is eventually injected as a "dependency", through a getter, is probably an implementation detail. It also reminds me a bit of `react` props which I feel is an advantage - you props are drilled down your components, and you don't care where they came from...

Happy to discuss this further though!

packages/report-server/src/reportBuilder/transform/functions/excludeColumns.ts
```diff
@@ -20,8 +21,8 @@ const paramsValidator = yup.object().shape({
   where: yup.string(),
 });
 
-const excludeColumns = (rows: Row[], params: ExcludeColumnsParams): Row[] => {
-  const parser = new TransformParser(rows);
+const excludeColumns = (rows: Row[], params: ExcludeColumnsParams, context: Context): Row[] => {
```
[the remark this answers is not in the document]
kael89: I agree that while there are solutions to inject `context` in those functions, they will probably make the code more complicated, as it will be harder to follow where the context is coming from... And you have established a nice functional pattern here!

There is a tipping point where the complexity of a wrapper would be preferred over the complexity/clutter of an additional parameter like context, but I believe we haven't reached that stage yet 🙂

packages/report-server/src/reportBuilder/context/index.ts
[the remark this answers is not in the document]
kael89: Changed them to explicit exports to keep consistency with the rest of the project

kael89:
@rohan-bes I have addressed all comments apart from the following:

* `contextDependencies`: left it as `contextProps`

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -65,14 +77,22 @@ export class TransformParser extends ExpressionParser {
...
+  public isExpression(input: unknown) {
+    return typeof input === 'string' && input.startsWith(this.EXPRESSION_PREFIX);
+  }
 
-    return expression;
+  protected readExpression(input: unknown) {
+    return this.isExpression(input)
+      ? (input as string).replace(new RegExp(`^${this.EXPRESSION_PREFIX}`), '')
+      : input;
+  }
```
kael89, line 92: Eventually I followed your suggestion and moved most of the logic in `TransformParser` 🙂 Not doing so would result in some interesting questions which we probably cannot answer consistently at this stage:
  * What happens if something is not an expression? Do we throw an error or just use the input?
   * Similarly: is the prefix optional or not? And if it is optional, what's the point in using it?

The compromise I thought of is adding a `readExpression` method in the base class, which can be overridden to support custom expression formats. The caveat here is that future devs must remember to call this function when an expression needs to be read in `ExpressionParser`. It's not ideal, but extending all present (and future) `ExpressionParser` methods in `TransformParser` so that our custom expression parsing is used seems even worse!

Another advantage of this pattern is that the custom format specified in sub classes is not restricted to a prefix, but could be anything.

kael89: Regarding semantics, I still believe that anything that can be evaluated is an _expression_. Following that definition, we should treat all expressions equally:
* for the literal number `2`: use `=2`
* for the literal string `MyString`: use `='MyString'`, or `=MyString` if our parser can easily distinguish between variables, functions and string literals

**But** I guess we had a valid reason for not doing so - I remember that in the past we used expressions like the following in the left sides of params, which didn't read fluently:
`"'name'":  "orgUnit"`

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -65,14 +77,22 @@ export class TransformParser extends ExpressionParser {
...
+
+  public evaluate(input: unknown) {
+    return this.isExpression(input) ? super.evaluate(input) : input;
```
kael89, line 95: The condition retains the existing functionality, which just returns the input (and avoids validation + evaluation) if the input does not start with `=`

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
kael89: I believe that left-side evaluation is what what makes a difference in `report-server` in relation to other languages we are (I am?) used into, and may be the reason - and justification! of more custom expression parsing.

Eg in Javascript:
```js
const valid = { result: 2 + 3 };
const invalid = { 2 + 3: 'result' };
```

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
```diff
@@ -75,6 +75,27 @@ describe('functions', () => {
     });
   });
 
+  describe('context', () => {
+    describe('orgUnitCodeToName', () => {
+      const context = {
+        orgUnits: [
```
rohan-bes, line 81: Totally happy with `orgUnits` for the name here.

At the same time it makes me sad... I feel we're losing the war on the 'Rename orgUnits to entities' fight. Org units is the right choice here, since the analytic has 'organisationUnit' in it, but the terminology is just so entrenched 😢

kael89:
> @rohan-bes I have addressed all comments apart from the following:
> 
> * `contextDependencies`: left it as `contextProps`

@rohan-bes I forgot to mention: happy to discuss about this and be convinced otherwise too. I didn't do it for the reasons I mentioned in my comment, but it's not like "I won't do this" 😛

packages/report-server/src/reportBuilder/context/buildContext.ts
```diff
@@ -0,0 +1,67 @@
...
+  reqContext: ReqContext,
+  data: FetchResponse,
+) => Promise<Context[K]>;
+
+const isEventResponse = (data: FetchResponse) => data.results.some(result => 'event' in result);
+
+const buildOrgUnits = async (reqContext: ReqContext, data: FetchResponse) => {
+  const orgUnitCodes = isEventResponse(data)
+    ? data.results.map(d => d.orgUnit)
+    : data.results.map(d => d.organisationUnit);
```
rohan-bes, line 28: It really bugs me that events use `orgUnit` and analytics use `organisationUnit`! Not saying we do this now, but I wonder how hard it'd be to migrate events to be consistent...

rohan-bes:
> @rohan-bes I have addressed all comments apart from the following:
contextDependencies: left it as contextProps

I'm cool with contextProps 👍

packages/report-server/src/routes/helpers.ts
```diff
@@ -3,8 +3,10 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import type { EntityApi } from '@tupaia/api-client';
-import type { AccessPolicy } from '@tupaia/access-policy';
+import { AccessPolicy } from '@tupaia/access-policy';
+import { MicroServiceRequestContext } from '@tupaia/server-boilerplate';
+
+type EntityApi = MicroServiceRequestContext['services']['entity'];
```
rohan-bes, line 9: I don't mind this change, but am also not opposed to the previously behaviour of directly importing `EntityApi` from `api-client`? I think some degree of awareness that services are instances of various `api-client`s is fine throughout the system?

packages/report-server/src/reportBuilder/transform/transform.ts
```diff
@@ -6,10 +6,20 @@
...
 
-type TransformParams = {
+export type TransformParams =
+  | keyof typeof aliases
+  | {
+      [key: string]: string | Record<string, string> | undefined;
+      transform: string;
+      title?: string;
+      description?: string;
+    };
```
rohan-bes, line 20: Not major, but I usually find that if I set up the yup validator properly I don't need to explicitly set up a type?

packages/report-server/src/reportBuilder/transform/transform.ts
```diff
@@ -68,13 +78,13 @@ const buildParams = (params: unknown): TransformParams => {
   return {
     title,
     name: transformStep,
-    apply: transformBuilders[transformStep](restOfTransformParams),
+    apply: transformBuilders[transformStep](restOfTransformParams, context as Context),
```
rohan-bes, line 81: Is this `as Context` required? If it being optional is an issue perhaps lets just provide a default empty context?

packages/report-server/src/__tests__/reportBuilder/context/buildContext.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    getEntities: async (
+      hierarchyName: keyof typeof ENTITIES,
+      entityCodes: string[],
+      queryOptions: { fields?: string[] } = {},
+    ) => {
+      const entities = ENTITIES[hierarchyName]?.filter(e => entityCodes.includes(e.code));
+      const { fields } = queryOptions;
+      return fields ? entities.map(e => pick(e, fields)) : entities;
+    },
+  };
```
rohan-bes, line 31: Really nice to see the hand written mock over using a mocking framework. Give that JS lets us do it, I'm quite partial to hand written mocks as I find them much easier to reason about and extend/debug!

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
```diff
@@ -65,14 +77,22 @@ export class TransformParser extends ExpressionParser {
...
     }
+
+    this.context = context;
   }
 
-  public evaluate(expression: string) {
-    if (TransformParser.isExpression(expression)) {
-      return super.evaluate(expression.substring(1));
-    }
+  public isExpression(input: unknown) {
```
rohan-bes, line 84: I know I'm being nitpicky here, but mind if we make this `static` again? Given that this function independnant of the instance state, it makes sense for it to be static to me? It also means that we can check if a input is an expression without instantiating a `TransformParser` (we do this in `sortRows` for instance)

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
rohan-bes: Also, wanna note that I love how the architecture allows us to write tests like this! Super clean 👍

rohan-bes (approved):
Awesome work @kael89! Really love how this all came together. Also thanks for adding such thorough tests 🎉 

A few comments on typescript things, and a few suggestions, but nothing too major so happy to approve 👍

packages/report-server/src/__tests__/reportBuilder/transform/parser/functions.test.ts
kael89: Yeah I think we came up with something clean! 🙌

packages/report-server/src/reportBuilder/context/buildContext.ts
kael89: I agree with you we should definitely do this at some point!

packages/report-server/src/routes/helpers.ts
kael89: To do that, I'd typically need to declare `@tupaia/api-client` as a dependency of this package.... I changed this to fix an existing EsLint error.

What I could do is expose some `@tupaia/api-client` types to `@tupaia/server-boilerplate` consumers through the latter... I will try that

rohan-bes: Ah, so typescript has something that works really nicely for that scenario. There's an `import type` command, which you can use to import a type from somewhere without declaring that package as a dependency 👍 Since the typescript stuff is all just compile time all `import type` statements get removed when code is built!

kael89: Ah that's really nice! Turns out the `import type` syntax was used in the first place, and the error was a bit different (`EntityApi` is not exported in that package)

packages/report-server/src/reportBuilder/transform/transform.ts
kael89: Great point! Will use that 👍

packages/report-server/src/__tests__/reportBuilder/context/buildContext.test.ts
kael89: I agree! Plus I am not a big fan of `jest`'s mocking api: too many options which is confusing, expects a `string` to the path of the mocked module (instead of importing it in the test)... I miss good ol' `sinon.createStubInstance`!

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
kael89: Now that `EXPRESSION_PREFIX` is not inherited from the base class, we could. I'll do that.

Disclaimer: I generally avoid static methods unless if there is a very specific reason to use one, since they are [hard to mock](https://testing.googleblog.com/2008/12/static-methods-are-death-to-testability.html) + don't fit right in an inheritance model. Of course, as with almost everything, there are justifiable use cases for them!

(The article I linked has a very ominous, click-baity title 😛 But I have also confirmed that static methods are hard to mock, at least with `sinon` )

kael89: Having said that, I've noticed in the past that instantiating `ExpressionParser` took much more than one would expect (something like `30ms`). In fact, they accounted for 70% of the slow performance in the initial implementation of [nested indicators](https://github.com/beyondessential/tupaia/pull/1835)!

I don't know if this is still the case, but I am flagging it since `TransformParser` is instantiated multiple times in `report-server`. If it is, a simple solution around that would be using a singleton.

packages/report-server/src/reportBuilder/transform/transform.ts
kael89: That's a bit tricky. My understandin is that it is required in some non-aliased transforms, which makes it generally required in their "type". At this code path we are calling a non-aliased transform.

Happy to have a call and see if there is a better way around that!

kael89:
@rohan-bes thanks for the insightful suggestions! I've applied the changes. Let me know if you want to chat about https://github.com/beyondessential/tupaia/pull/3253#discussion_r713471400

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
rohan-bes: I agree that it's good to be wary about static methods. You really need to be careful that they don't mutate state, and yes they are very hard to mock. That being said, in this instance, and in general, I do find them appropriate for places where they perform simple, functional logic, and that mocking wouldn't be appropriate.

For instance you could in theory mock `isExpression` to always return true when testing various expressions in report-server, to save yourself the trouble of putting `=` at the start of them. But I think that would be a misuse of mocking. Leaving it as is and adding the `=` improves your tests as it aligns your tests with how the code actually operates.

packages/report-server/src/reportBuilder/transform/transform.ts
rohan-bes: I think the general approach with validation in report-server has been that the higher levels make type assertions to the lowest common denominator, and then it's up to the children to validate up to their required level. 

In this instance I guess that would mean that context can be optional in this scope, as it's not required by all. And then for the transforms that do require it there's a secondary validation check to assert it is defined?

Yup, happy to call on this if you want 👍

packages/report-server/src/routes/helpers.ts
```diff
@@ -30,7 +28,7 @@ export const getAccessibleOrgUnitCodes = async (
 export const getRequestedOrgUnitObjects = async (
   hierarchy: string,
   orgUnitCodes: string | string[],
-  entityApi: EntityApi,
+  entityApi: TupaiaApiClient['entity'],
```
rohan-bes, line 31: Very clean 👍

packages/report-server/src/reportBuilder/transform/parser/TransformParser.ts
rohan-bes: As for the performance concerns around instantiating `TransformParser`. I share them too, but I think for now it's acceptable. We insatiate at once per transform I believe, which could be up to 10 or so? Singleton pattern could work but would require a bit more overhead of cleaning up scope between uses...

kael89: 10 times 30 is 300ms, a significant overhead for something that doesn't access the file system...

kael89:
@rohan-bes squashed and merged in the base branch, let's remember to mention this card in the merge commit and release PR when `rn-83` is merged in dev

PR #3254, opened 2021-09-20
Performance fix for in-efficient code in mergeRows

rohan-bes:
O(n^2) was causing noticable slowness for 50k+ rows

### DIFF
--- a/packages/report-server/src/reportBuilder/transform/functions/mergeRows/getMergeStrategy.ts
+++ b/packages/report-server/src/reportBuilder/transform/functions/mergeRows/getMergeStrategy.ts
@@ -19,40 +19,57 @@ export const buildGetMergeStrategy = (
   using?: keyof typeof mergeStrategies | Record<string, keyof typeof mergeStrategies>,
 ) => {
   const isGroupByField = buildIsGroupByField(groupBy);
+  const knownMergeStrategies: Record<string, keyof typeof mergeStrategies> = {};
+  const setAndReturnStrategy = (field: string, strategy: keyof typeof mergeStrategies) => {
+    knownMergeStrategies[field] = strategy;
+    return strategy;
+  };
 
   if (using === undefined) {
     return (field: string) => {
-      if (isGroupByField(field)) {
-        return 'group';
+      if (knownMergeStrategies[field]) {
+        return knownMergeStrategies[field];
       }
-      return 'default';
+
+      if (isGroupByField(field)) {
+        return setAndReturnStrategy(field, 'group');
+      }
+      return setAndReturnStrategy(field, 'default');
     };
   }
 
   if (typeof using === 'string') {
     return (field: string) => {
-      if (isGroupByField(field)) {
-        return 'group';
+      if (knownMergeStrategies[field]) {
+        return knownMergeStrategies[field];
       }
-      return using;
+
+      if (isGroupByField(field)) {
+        return setAndReturnStrategy(field, 'group');
+      }
+      return setAndReturnStrategy(field, using);
     };
   }
 
   const anyFieldMergeStrategyName = using['*'];
   return (field: string) => {
+    if (knownMergeStrategies[field]) {
+      return knownMergeStrategies[field];
+    }
+
     if (isGroupByField(field)) {
-      return 'group';
+      return setAndReturnStrategy(field, 'group');
     }
 
     const mergeStrategyName = using[field];
     if (mergeStrategyName === undefined) {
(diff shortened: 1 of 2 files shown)


packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
```diff
@@ -70,19 +70,22 @@ const groupRows = (rows: Row[], params: MergeRowsParams) => {
...
   });
 
   return { groups: Object.values(groupsByKey), ungroupedRows };
 };
 
-const addRowToGroup = (group: Group, row: Row): Group => {
-  const newGroup = { ...group };
+const addRowToGroup = (group: Group = {}, row: Row): Group => {
   Object.keys(row).forEach((field: string) => {
-    newGroup[field] = [...(newGroup[field] || []), row[field]];
```
rohan-bes, line 83: If I had a nickel for (each time I made this mistake x each time I made this mistake) I would be a very rich crappy developer 😝

packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
```diff
@@ -70,19 +70,22 @@ const groupRows = (rows: Row[], params: MergeRowsParams) => {
...
+    if (!group[field]) {
+      // eslint-disable-next-line no-param-reassign
```
rohan-bes, line 83: I know it's naughty, but it's so much speedier...

packages/report-server/src/reportBuilder/transform/functions/mergeRows/getMergeStrategy.ts
```diff
@@ -19,40 +19,57 @@ export const buildGetMergeStrategy = (
   using?: keyof typeof mergeStrategies | Record<string, keyof typeof mergeStrategies>,
 ) => {
   const isGroupByField = buildIsGroupByField(groupBy);
+  const knownMergeStrategies: Record<string, keyof typeof mergeStrategies> = {};
+  const setAndReturnStrategy = (field: string, strategy: keyof typeof mergeStrategies) => {
+    knownMergeStrategies[field] = strategy;
+    return strategy;
+  };
```
rohan-bes, line 26: This optimisation is probably a bit unnecessary, but given that there's usually many many more rows than there are columns it would save a bit of time just to cache the known values for each column

packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
```diff
@@ -70,19 +70,22 @@ const groupRows = (rows: Row[], params: MergeRowsParams) => {
       return;
     }
     const groupKey = params.createGroupKey(row);
-    groupsByKey[groupKey] = addRowToGroup(groupsByKey[groupKey] || {}, row);
+    groupsByKey[groupKey] = addRowToGroup(groupsByKey[groupKey], row);
```
edmofro, line 73: I reckon I'd prefer a function signature that emphasised that it was mutating internally, because the assignment here makes you think it wouldn't be
```suggestion
    addRowToGroup(groupsByKey, groupKey, row); // mutates groupsByKey
```

packages/report-server/src/reportBuilder/transform/functions/mergeRows/getMergeStrategy.ts
edmofro: I like it :-)

edmofro (changes requested):
One little request to make the mutation more obvious - naughtiness should be as explicit as possible ;-)

packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
rohan-bes: Ah good call, will do!

edmofro: Great optimisation btw, thanks for improving this!

edmofro (approved):

packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
rohan-bes: I was testing with a report that had 70k rows, and the performance was really bad (+1 min to return). Initially I went pale thinking that mathjs expression parsing was too slow... moral of the story is to check yourself before you blame the 3rd party lib 😂

PR #3258, opened 2021-09-21
PHX-57 refactor to use mapoverlayids

billli0:
### Issue #:
PHX-57
### Changes:
- Web config server measureData endPoint changes: 
  - Remove `measureIds` in the response of  `measure` and `measureData` api endpoint. 
  - Query param `measureIds` to `mapOverlayId`, so frontend don't need to be aware of measure ids. Config server will return a union measure data if multiple mapOverlay ids is provided.  

- Front end changes:
  - Take `mapOverlayIds` in the state as which map overlay is selected, instead of `measureIds`.
  - A few function and variable name changes across frontend including cypress.

- Migration:
  - Update 'project' table's 'default_measure' to use map overlay id instead of measure ids.


kael89:
@billli0 could you use Tom's branch as the `base` branch for this PR, to make the actual diff clearer?

billli0:
@kael89 That is the actual diff since Tom's branch is already merged. Happy to chat and walk through all the changes because it seems to be large.

kael89:
> @kael89 That is the actual diff since Tom's branch is already merged. Happy to chat and walk through all the changes because it seems to be large.

All good @billli0 - it's pretty normal for a change like this to span over 30+ files... I thought that Tom's code was included in the diff by looking at the commit history, but I guess it is caused by the other branch being squashed-and-merged 🙂 

I will have a general look first and will let you know if we should do a walk-n-talk review!

billli0:
@kael89 My fault, the commit history is such a mess now, I should rebase next time.

kael89:
> @kael89 My fault, the commit history is such a mess now, I should rebase next time.

It's not a big deal since the diff is actual clean, that's the most important part 🙌

billli0:
For now I haven't tested the changes in tupaia mobile app, if the changes is solid I will build one and check.

packages/web-frontend/src/selectors/mapOverlaySelectors.js
```diff
@@ -0,0 +1,80 @@
...
+);
+
+export const selectCurrentMapOverlay = createSelector(
+  [state => selectMapOverlayById(state, selectCurrentMapOverlayId(state))],
+  currentMapOverlay => currentMapOverlay || {},
+);
+
+export const selectIsMapOverlayInHierarchy = createSelector(
+  [state => state.measureBar.measureHierarchy, (_, ids) => ids],
+  (measureHierarchy, ids) => !!getMapOverlayFromHierarchy(measureHierarchy, ids),
```
kael89, line 35:
```suggestion
  [state => state.measureBar.measureHierarchy, (_, id) => id],
  (measureHierarchy, id) => !!getMapOverlayFromHierarchy(measureHierarchy, id),
```

packages/web-frontend/src/actions.js
```diff
@@ -541,10 +541,10 @@ export function setMeasure(measureId) {
  *
  * @param {object} measureConfig
  */
-export function updateMeasureConfig(measureId, measureConfig) {
+export function updateMeasureConfig(mapOverlayIds, measureConfig) {
```
kael89, line 544:
```suggestion
export function updateMeasureConfig(mapOverlayId, measureConfig) {
```

packages/web-frontend/src/selectors/mapOverlaySelectors.js
```diff
@@ -0,0 +1,80 @@
...
+);
+
+export const selectCurrentPeriodGranularity = createSelector(
+  [selectCurrentMapOverlay],
+  mapOverlay => mapOverlay.periodGranularity,
+);
+
+export const selectMapOverlayGroupById = createSelector(
+  [state => state.measureBar.measureHierarchy, (_, ids) => ids],
+  (measureHierarchy, ids) => {
```
kael89, line 61:
```suggestion
  [state => state.measureBar.measureHierarchy, (_, id) => id],
  (measureHierarchy, id) => {
```

etc in the function body

packages/web-frontend/src/selectors/mapOverlaySelectors.js
```diff
@@ -0,0 +1,80 @@
...
+    let mapOverlayGroupIndex = {};
```
kael89, line 62: The name was a bit inaccurate in the existing code (an index is generally an integer, not an object), let's take the chance to fix it:

```suggestion
    let mapOverlayGroup = {};
```

packages/web-frontend/src/selectors/mapOverlaySelectors.js
```diff
@@ -0,0 +1,80 @@
...
+    measureHierarchy.forEach(({ name, children }, categoryIndex) => {
+      const selectedMapOverlayGroupIndex = children.findIndex(
+        mapOverlay => mapOverlay.mapOverlayId === ids,
+      );
+      if (selectedMapOverlayGroupIndex > -1) {
+        mapOverlayGroupIndex = {
+          name,
+          categoryIndex,
+          mapOverlayGroupIndex: selectedMapOverlayGroupIndex,
+          mapOverlays: children[selectedMapOverlayGroupIndex],
```
kael89, line 73:
```suggestion
          groupIndex,
          mapOverlayIndex: selectedMapOverlayIndex,
          mapOverlay: children[selectedMapOverlayIndex],
```

We'll also need to update the usages of `selectMapOverlayGroupById` to use the new field names

packages/web-frontend/src/selectors/mapOverlaySelectors.js
```diff
@@ -0,0 +1,80 @@
...
+  mapOverlay => mapOverlay.periodGranularity,
+);
+
+export const selectMapOverlayGroupById = createSelector(
+  [state => state.measureBar.measureHierarchy, (_, ids) => ids],
+  (measureHierarchy, ids) => {
+    let mapOverlayGroupIndex = {};
+
+    measureHierarchy.forEach(({ name, children }, categoryIndex) => {
+      const selectedMapOverlayGroupIndex = children.findIndex(
```
kael89, line 65:
```suggestion
    measureHierarchy.forEach(({ name, children }, groupIndex) => {
      const selectedMapOverlayIndex = children.findIndex(
```

kael89 (changes requested):
Great work @billli0 ! I think we were overusing the measures concept, the front-end code makes much more sense now!

I left a note about a few more renames we'll need to do. Also, I'm not sure about the query name you used (`mapOverlayIds`). I think I'd prefer it to stay as it were: simply `overlay`, which could be pluralised to `overlays`. It may not be a big deal, I'll leave a note in the card to confirm.

For future reference, it's generally preferred to leave a note in the card and tag other devs about a decision like a query param name. This will give you the chance to reach a consensus early on, rather than having to change a lot of code after the PR is up 🙂

packages/web-frontend/src/containers/MapTableModal/MapTableModal.js
```diff
@@ -66,24 +66,25 @@ MapTableModalComponent.propTypes = {
...
   measureData: null,
   currentCountry: null,
-  currentMeasure: null,
+  currentMapOverlay: null,
 };
 
 const mapStateToProps = state => {
   const { measureOptions } = state.map.measureInfo;
-  const currentMeasure = selectCurrentMeasure(state);
+  // TODO: select multiple map overlays
```
kael89, line 81: @billli0 tip (no need to change this): you can include the card number in the TODO for easier tracking, eg

```suggestion
  // TODO: select multiple map overlays - PHX-1
```

billli0:
> I left a note about a few more renames we'll need to do. Also, I'm not sure about the query name you used (`mapOverlayIds`). I think I'd prefer it to stay as it were: simply `overlay`, which could be pluralised to `overlays`. It may not be a big deal, I'll leave a note in the card to confirm.
> 
> For future reference, it's generally preferred to leave a note in the card and tag other devs about a decision like a query param name. This will give you the chance to reach a consensus early on, rather than having to change a lot of code after the PR is up 🙂

Thanks @kael89! I also think overlays looks better after all the changes are done.. I will change it back.

packages/web-frontend/src/containers/Map/Map.js
```diff
@@ -75,10 +75,6 @@ class MapComponent extends Component {
     } = this.props;
     // Only updates/re-renders when the measure has changed or the orgUnit has changed.
     // These are the only cases where polygons or area tooltips should rerender.
-    if (nextProps.measureInfo.measureId !== measureInfo.measureId) {
-      return true;
-    }
-
```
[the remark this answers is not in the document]
billli0: That's a good point @kael89 , I feel that is a trade off because `measureId` is not being used in this component. However I'm happy to put it back for future usage, I guess it can be used for another map overlay refactoring.

packages/web-frontend/src/containers/MeasureBar/index.js
```diff
@@ -195,39 +207,44 @@ const mapStateToProps = state => {
   const { isLoadingOrganisationUnit } = state.global;
 
   const currentOrganisationUnit = selectCurrentOrgUnit(state);
-  const currentMeasure = selectCurrentMeasure(state);
+  const currentMapOverlay = selectCurrentMapOverlay(state);
+
+  // TODO failed prop type
+  const currentMapOverlayId = currentMapOverlay.mapOverlayId;
+  // TODO: current measures in Measure Bar
+  // const currentMeasures = selectCurrentMeasures(state);
```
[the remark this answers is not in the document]
billli0: It is related to the {} return from selectCurrentMapOverlay(). Issue just fixed!

packages/web-frontend/src/selectors/mapOverlaySelectors.js
```diff
@@ -0,0 +1,80 @@
...
+import { getMapOverlayFromHierarchy } from '../utils';
+import { flattenMapOverlayHierarchy, isMeasureHierarchyEmpty } from '../utils/mapOverlays';
+
+import { selectCurrentProject } from './projectSelectors';
+import { selectLocation } from './utils';
+
+export const selectMapOverlayById = createSelector(
+  [state => state.measureBar.measureHierarchy, (_, id) => id],
+  (measureHierarchy, id) => {
+    return getMapOverlayFromHierarchy(measureHierarchy, id) || {};
```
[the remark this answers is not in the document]
billli0: That's right we don't need it now, previously it was caused by `Control.js`, I did a bit refactoring and now it doesn't need to return {}.

packages/web-frontend/src/screens/mobile/RegionScreen/index.js
```diff
@@ -291,7 +291,7 @@ const mapStateToProps = state => {
 };
 
 const mapDispatchToProps = dispatch => ({
-  onChangeMeasure: measureId => dispatch(setMeasure(measureId)),
+  onChangeMapOverlay: mapOverlayId => dispatch(setMapOverlay(mapOverlayId)),
   onClearMeasure: () => dispatch(clearMeasure()),
```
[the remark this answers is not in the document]
billli0: @kael89 TOGGLE_MEASURE_EXPAND is used with onToggleMeasureExpand(), will delete other two that you've found.

I left these measure variables for PHX-1 because I feel like they will be handled and changed slightly different from map overlay for multiple map overlay checkbox (but `onClearMeasure` is not this case..). Do you think we can still keep others for now?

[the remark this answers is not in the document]
kael89: Fair enough! Let's leave them as they are for now.

Good point about `onToggleMeasureExpand`, I just realised it's used in `RegionScreen` 👍

packages/database/src/modelClasses/MapOverlay.js
```diff
@@ -17,6 +17,25 @@ export class MapOverlayModel extends DatabaseModel {
...
+        columns: [
+          { id: `${TYPES.MAP_OVERLAY}.id` },
+          { linkedMeasures: `${TYPES.MAP_OVERLAY}.linkedMeasures` },
+        ],
+      },
+    );
+    const measureIds = new Set();
+    overlays.forEach(o => {
+      const { id, linkedMeasures = [] } = o;
+      measureIds.add(linkedMeasures);
```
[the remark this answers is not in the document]
billli0: Good catch🙌

packages/web-frontend/src/utils/mapOverlays.js
```diff
@@ -0,0 +1,40 @@
...
+export const getMapOverlayFromHierarchy = (mapOverlayHierarchy, targetMapOverlayId) => {
+  if (!targetMapOverlayId) {
+    return null;
+  }
+
+  const flattenMapOverlays = flattenMapOverlayHierarchy(mapOverlayHierarchy);
+  return flattenMapOverlays.find(({ mapOverlayId }) => targetMapOverlayId === mapOverlayId);
+};
+
+export function flattenMapOverlayHierarchy(measureHierarchy) {
```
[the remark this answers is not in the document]
billli0: Easy👍

packages/web-frontend/src/actions.js
billli0: Will leave that for now because we might need to update configs for both selected map overlays in PHX-1?

kael89: The code is used with a single id: https://github.com/beyondessential/tupaia/blob/dev/packages/web-frontend/src/sagas.js#L934

Let's use `mapOverlayId` so that we are consistent with the current code, and refactor it if indeed we will need to change it to receive an array of ids

kael89: The state may look a bit different when multiple overlays can be selected, and it's a bit hard to predict from now, so it's safer to be consistent with the existing code and refactor at a later stage

billli0: Sounds good, let's keep it consistent.

packages/web-frontend/src/containers/MapTableModal/MapTableModal.js
billli0: Good idea! It helps to find in the future.

packages/web-frontend/src/sagas.js
```diff
@@ -924,6 +924,13 @@ function* watchTryUpdateMeasureConfigAndWaitForHierarchyLoad() {
   );
 }
 
+function* updateMapOverlayOnceHierarchyLoads() {
+  yield take(FETCH_MEASURES_SUCCESS);
+  const state = yield select();
+  const currentMapOverlayId = selectCurrentMapOverlayId(state);
+  yield put(setMapOverlay(currentMapOverlayId));
+}
+
```
billli0, line 933: Previously when we refresh or hit the URL with map overlay id, we fetch measure data before the map overlay hierarchy is not loaded, which cannot include the period in the request and shows all the data. Now we always fetch measure data after hierarchy is loaded.

kael89 (approved):
Looks great @billli0, thanks for taking the time to address all suggestions!

PR #3260, opened 2021-09-21
remove old laos schools map overalays

tcaiger:
### Changes:

Remove old laos schools map overalys to make way for the new ones as part of this cycle of new overlays. I didn't remove all the old ones as they might be good to have for future reference. They can always be deleted later on when they get replaced.

---


kael89 (approved):
Looks good! I assume that we are keeping the groups and that the new overlays will be added there 👍

PR #3261, opened 2021-09-21
Fix startup script: don't use `exec` from cron job

edmofro:
Will solve the fact that dev and feature deployments are not coming online

### DIFF
--- a/scripts/bash/downloadEnvironmentVariables.sh
+++ b/scripts/bash/downloadEnvironmentVariables.sh
@@ -19,9 +19,7 @@ for PACKAGE in $PACKAGES; do
     ENV_FILE_PATH=${DIR}/../../packages/${PACKAGE}/.env
 
     # checkout branch specific env vars, or dev as fallback, temporarily redirecting stderr
-    exec 3> /dev/stderr 2> /dev/null
-    lpass show --notes ${PACKAGE}.${BRANCH}.env > ${ENV_FILE_PATH} || lpass show --notes ${PACKAGE}.dev.env > ${ENV_FILE_PATH}
-    exec 2>&3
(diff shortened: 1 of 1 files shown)


scripts/bash/downloadEnvironmentVariables.sh
```diff
@@ -19,9 +19,7 @@ for PACKAGE in $PACKAGES; do
     ENV_FILE_PATH=${DIR}/../../packages/${PACKAGE}/.env
 
     # checkout branch specific env vars, or dev as fallback, temporarily redirecting stderr
-    exec 3> /dev/stderr 2> /dev/null
```
edmofro, line 22: It was actually the `exec` that was breaking things when run from a cron job

IgorNadj (approved):
Ah that's what it was! Good work finding that!

scripts/bash/downloadEnvironmentVariables.sh
```diff
@@ -19,9 +19,7 @@ for PACKAGE in $PACKAGES; do
...
-    lpass show --notes ${PACKAGE}.${BRANCH}.env > ${ENV_FILE_PATH} || lpass show --notes ${PACKAGE}.dev.env > ${ENV_FILE_PATH}
-    exec 2>&3
+    lpass show --notes ${PACKAGE}.${BRANCH}.env 2> /dev/null > ${ENV_FILE_PATH} || lpass show --notes ${PACKAGE}.dev.env > ${ENV_FILE_PATH}
```
edmofro, line 22: Stderr now redirected specifically for the first "branch specific" one, which often fails before we fall back to dev

edmofro:
> Ah that's what it was! Good work finding that!

Thanks...you'd be amazed at how long I spent trying to nail it down 😅

edmofro:
Tested on feature deployment, works :+1:

PR #3265, opened 2021-09-22
WAI-9 & WAI-10: Lesmis Gross Intake Ratio Overlays

tcaiger:
### Changes:

- Gross Intake Rate map overlays for Primary, Lower Secondary and Upper Secondary

---

### Screenshots:
See issue


EMcQ-BES (approved):
Looks good to me. Keep in mind repeating that `nameToId` change might cause some merge issues against the other branch where you added it.

PR #3267, opened 2021-09-23
WAI-13 Allow lesmis to request map overlays directly from report server

EMcQ-BES:
### Issue #:
WAI-13

Also added these map overlays as proof of concept:
WAI-12
WAI-4
WAI-5

### Changes:

- Migration to add enrolment rate map overlays
- Update the report route in `lesmis-server` to allow legacy and new map overlay requests
- Allow `useMapOverlayReportData` to translate between the two styles of map overlay
- Update the measures route on `web-config-server` to return some extra pieces of necessary data
- Update `useVitalsData` to be more consistent with other report usages

---

### Screenshots:
See WAI-944 for a screenshot of one of these map overlays displaying


packages/lesmis/src/api/queries/useVitalsData.js
```diff
@@ -85,23 +85,19 @@ const useEntityReport = entity =>
...
 
 const useSchoolVitals = entity => {
-  const { data, isLoading } = useSchoolReport(entity);
+  const { data: results, isLoading } = useSchoolReport(entity);
 
   return {
-    data: {
-      ...data?.results[0],
-    },
+    data: results ? results[0] : undefined,
```
edmofro, line 91: I think you could use 
```suggestion
    data: results?[0],
```
but also clear and good as is

packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -159,17 +159,29 @@ const sortMapOverlayItems = (mapOverlayItems, relations) => {
...
+        report_code: reportCode,
+        dataElementCode,
+        legacy,
+      }) => {
+        const idString = [id, ...(linkedMeasures || [])].sort().join(',');
+
+        return {
+          id, // just for sorting purpose, will be removed later
+          measureId: idString,
+          code: legacy ? idString : reportCode, // if we're not a legacy overlay, return the report code
```
edmofro, line 177: Hmmm...with the dashboard items, do we return the report code or the dashboard item code under the `code` key? It seems a little misleading here

edmofro (changes requested):
PR review excluding migration, will get back to that!

packages/database/src/migrations/20210923011806-LESMISEnrolmentRateMapOverlays-modifies-data.js
```diff
@@ -0,0 +1,235 @@
...
+const getReport = (reportCode, dataElements) => ({
+  code: reportCode,
+  config: {
+    fetch: {
+      dataElements,
+      aggregations: [
+        {
+          type: 'FINAL_EACH_YEAR',
+          config: {
+            dataSourceEntityType: reportCode.includes('Province') ? 'district' : 'sub_district',
```
edmofro, line 28:
```suggestion
            dataSourceEntityType: sub_district',
```

packages/database/src/migrations/20210923011806-LESMISEnrolmentRateMapOverlays-modifies-data.js
```diff
@@ -0,0 +1,235 @@
...
+  isDataRegional: true,
+  measureBuilder: 'useReportServer',
+  measureBuilderConfig: {
+    dataSourceType: 'custom',
+    reportCode,
+  },
+  presentationOptions: {
+    scaleType: 'performance',
+    displayType: 'shaded-spectrum',
+    measureLevel: reportCode.includes('Province') ? 'District' : 'SubDistrict',
```
edmofro, line 63:
```suggestion
    measureLevel: 'SubDistrict',
```

packages/database/src/migrations/20210923011806-LESMISEnrolmentRateMapOverlays-modifies-data.js
```diff
@@ -0,0 +1,235 @@
...
+  },
+  {
+    parentCode: `${OVERLAY_CODE}_Group`,
+    children: [
+      {
+        name: 'District Level',
+        code: `${OVERLAY_CODE}_District_Group`,
+      },
+    ],
+  },
```
edmofro, line 138: This seems a bit redundant if we're not adding province level reports. Perhaps there will be for other map overlays?

edmofro (changes requested):
And migration (I have one eye on this and the other on the symposium, so hopefully I haven't missed anything)

packages/database/src/migrations/20210923011806-LESMISEnrolmentRateMapOverlays-modifies-data.js
EMcQ-BES: I don't think there's any province level reports for this particular category (ECE/Pre-Primary Enrolment), but I think it's better to keep it so that nested structure is consistent between all the different categories.

edmofro: Sure!

packages/lesmis/src/api/queries/useVitalsData.js
EMcQ-BES: I tried this and it didn't work so I swapped it out quickly, just took another look an realised the format is just slightly unexpected for arrays `results?.[0]` seems to work

edmofro: Ah right! That is just slightly unexpected.

packages/lesmis/src/api/queries/useMapOverlayReportData.js
```diff
@@ -122,19 +148,21 @@ export const useMapOverlayReportData = ({ entityCode, year }) => {
...
-
+  const overlay = findOverlay(overlaysData, selectedOverlay);
   const { startDate, endDate } = yearToApiDates(year);
 
   const params = {
     startDate,
     endDate,
     shouldShowAllParentCountryResults: false,
     type: 'mapOverlay',
+    legacy: overlay?.legacy,
```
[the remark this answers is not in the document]
EMcQ-BES: Because it's pulled from the `useMapOverlaysData` react hook which doesn't return a value right away

The query after this checks `!!overlay` before running so it's guaranteed to be set before it's actually used, but not to be set in this particular line where we assign

At least that's how my limited understanding of react goes, I do know that this line crashes if you don't account for it 😅

packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
EMcQ-BES: Ah fair enough. Dashboard items have a separate field for reportCode so I've updated this to match

packages/lesmis/src/api/queries/useMapOverlayReportData.js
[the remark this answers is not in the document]
edmofro: Ah yep that makes sense. In that case I think we should avoid running any of the code below `useMapOverlaysData` while that's still loading. I'd jump to an early return after line 152 if `overlaysLoading` was true with just `return { isLoading: true };`, but that expects that the downstream consumers are ok with not having the other keys defined while it's loading, and I'm not sure whether that's true. Could you check it out?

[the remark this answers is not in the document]
EMcQ-BES: Can't do that because react needs to render the same number of hooks everytime. We can't exit before we've run all the `use` functions. That's why we have to pass it into the `enabled` param to check if we should actually perform the action.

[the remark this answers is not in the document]
edmofro: Ah yes, of course! Cool :+1:

edmofro (approved):

PR #3269, opened 2021-09-23
Add support for GPI map overlays

tcaiger:
### Changes:

- Add new GPI Spectrum type

This is quite a specific type but considered common enough to have its own type. We did look into making it more generic but we deemed that substantially harder to implement.

---

### Screenshots:
See issue

### DIFF
--- a/packages/ui-components/src/components/Map/constants/colors.js
+++ b/packages/ui-components/src/components/Map/constants/colors.js
@@ -56,18 +56,7 @@ export const DEFAULT_COLOR_SCHEME = 'default';
 export const REVERSE_DEFAULT_COLOR_SCHEME = 'default-reverse';
 export const PERFORMANCE_COLOR_SCHEME = 'performance';
 export const TIME_COLOR_SCHEME = 'time';
-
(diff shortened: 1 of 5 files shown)


packages/ui-components/src/components/Map/constants/colors.js
```diff
@@ -56,18 +56,7 @@ export const DEFAULT_COLOR_SCHEME = 'default';
 export const REVERSE_DEFAULT_COLOR_SCHEME = 'default-reverse';
 export const PERFORMANCE_COLOR_SCHEME = 'performance';
 export const TIME_COLOR_SCHEME = 'time';
-
-export const HEATMAP_DEFAULT_RGB_SET = [
```
tcaiger, line 60: This is not being used anywhere

edmofro (approved):
Nice work guys! The overlays look super cool with this in place. Some of the stuff in here gets extremely specific (like the neutral thresholds) which means it wouldn't work to reuse this "tri-colour" scale for any other overlay, which is a bit of a shame but we can always add in a more generic one that's normalised to percentage later on.

PR #3272, opened 2021-09-24
WAI-946 Part 1: change mapOverlays table to snake_case

edmofro:
### Issue #:
WAI-946

PR #3273, opened 2021-09-24
WAI-946 Part 1: change mapOverlays table to snake_case

edmofro:
Issue #:
WAI-946

Replaces https://github.com/beyondessential/tupaia/pull/3272 with valid branch name


EMcQ-BES:
@edmofro This should probably be pointing to `lesmis-map-overlays-epic` right?

edmofro:
Indeed! This is the one you already reviewed.

PR #3274, opened 2021-09-26
WAI-946 Part 2: Add migration to split out legacy reports from map overlay table

edmofro:
### Issue #:
WAI-946


packages/database/src/migrations/20210924062930-SplitMapOverlayLegacyReportsOut-modifies-schema.js
```diff
@@ -0,0 +1,82 @@
...
+  await db.runSql(`
+    UPDATE map_overlay
+    SET report_code = id || '_map'
+    WHERE legacy = true;
+  `);
+  await db.runSql(`
+    INSERT INTO legacy_report (id, code, data_builder, data_builder_config, data_services)
+    FROM (
+      SELECT generate_object_id(), id || '_map', "measureBuilder", "measureBuilderConfig", data_services
+      FROM map_overlay
```
EMcQ-BES, line 32:
```suggestion
      FROM map_overlay
      WHERE legacy = true
```

EMcQ-BES (changes requested):
One small change, otherwise migration looks good from a read through.

Obviously this is going to break everything until we actually write the supporting code.

packages/database/src/migrations/20210924062930-SplitMapOverlayLegacyReportsOut-modifies-schema.js
edmofro: Good catch

edmofro:
Absolutely, I'm hoping you can pick that part up today?

EMcQ-BES:
> Absolutely, I'm hoping you can pick that part up today?

Yep, that's the plan!

PR #3280, opened 2021-09-26
Sanitize jsonb arrays

edmofro:
### Issue #:
https://github.com/beyondessential/tupaia/pull/3273 is failing tests because of an issue inserting test data into the db. The problem is that knex doesn't handle json/jsonb columns where the root of the object is an array, see https://knexjs.org/#Schema-json

### DIFF
--- a/packages/database/src/DatabaseModel.js
+++ b/packages/database/src/DatabaseModel.js
@@ -201,7 +201,15 @@ export class DatabaseModel {
   }
 
   generateInstance = async (fields = {}) => {
-    const data = await this.getDatabaseSafeData(fields);
+    const data = {};
+
+    // add values for standard fields
+    const fieldNames = await this.fetchFieldNames();
+    fieldNames.forEach(fieldName => {
+      data[fieldName] = fields[fieldName];
+    });
+
+    // add values for joined fields
     this.joins.forEach(({ fields: joinFields }) => {
       Object.values(joinFields).forEach(fieldName => {
         data[fieldName] = fields[fieldName];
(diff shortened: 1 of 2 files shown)


packages/database/src/DatabaseModel.js
```diff
@@ -201,7 +201,15 @@ export class DatabaseModel {
   }
 
   generateInstance = async (fields = {}) => {
-    const data = await this.getDatabaseSafeData(fields);
```
edmofro, line 204: I am 95% sure the `getDatabaseSafeData` method was just called out of convenience here. What we actually want is just an object with the expected records filled out, but on _retrieving_ records we shouldn't need to do the same sanitization as when we're _inserting_ records into the db (and in fact, with the addition of stringifying json arrays, we wouldn't want to)

packages/database/src/TupaiaDatabase.js
```diff
@@ -513,6 +513,10 @@ function buildQuery(connection, queryConfig, where = {}, options = {}) {
     query.returning('*');
   }
 
+  if (process.env.DB_VERBOSE === 'true') {
+    winston.info(query.toString());
+  }
```
edmofro, line 518: This is unrelated but helpful. Allows running e.g. `DB_VERBOSE=true yarn workspace @tupaia/meditrak-server start-dev` to see all queries logged to the console. To see the same in tests, you do also need to update the winston transports in `scaffolding.test.js` like so:
image

kael89: Nice one! The other day I also realised we should not silence error logging in tests by default. Some times they indicate an issue in the tested code (which is not picked by an assertion).

We can then silence errors selectively (ideally silence specific types of errors, or test the code in a level where errors are thrown rather than logged)

packages/database/src/DatabaseModel.js
kael89: Sounds good!

kael89 (approved):
Thanks for fixing this! Great timing, I just stumbled upon this issue on Thursday, while I was trying to persist/retrieve a value for `legacy_report.data_services`  in some new tests!

kael89:
@edmofro could we change the base branch to `dev`?

edmofro:
Can do @kael89! I'll do that but will only merge after today's release cut off, so it gets a week of passive regression testing.

packages/database/src/TupaiaDatabase.js
edmofro: True! Left the tests log handling as out of scope here, but if you felt like doing that go for it. We'd have to be fairly assiduous in silencing expected logs so that logs became more meaningful - at the moment I feel I see logs in the tests and just ignore them because most of the time they're just a side effect

kael89: Started doing that a few weeks ago, took more time than would be justified at that stage, and left it in the middle. Ideally we'd catch specific error types, and log everything else

PR #3284, opened 2021-09-26
RN-104 Fix duplicate fetch

IgorNadj:
RN-104

### DIFF
--- a/packages/data-broker/src/services/dhis/DhisService.js
+++ b/packages/data-broker/src/services/dhis/DhisService.js
@@ -138,11 +138,13 @@ export class DhisService extends Service {
 
     const entityCodes = organisationUnitCodes || [organisationUnitCode];
     const pullData = this.pullers[type];
-    const apis = dataServices.map(({ isDataRegional }) =>
-      getDhisApiInstance({ entityCodes, isDataRegional }, this.models),
-    );
 
-    return pullData(apis, dataSources, options);
+    const apis = new Set();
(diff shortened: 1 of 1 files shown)


packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -138,11 +138,13 @@ export class DhisService extends Service {
...
     const pullData = this.pullers[type];
-    const apis = dataServices.map(({ isDataRegional }) =>
-      getDhisApiInstance({ entityCodes, isDataRegional }, this.models),
-    );
 
-    return pullData(apis, dataSources, options);
+    const apis = new Set();
+    dataServices.forEach(({ isDataRegional }) => {
+      apis.add(getDhisApiInstance({ entityCodes, isDataRegional }, this.models));
+    });
```
rohan-bes, line 145: Feels a tad risky just to rely on object reference to ensure we're not doubling up here (usually I prefer to have some kind of id or something here). I suppose that we could implement a `getServerName()` function on the `dhisApi` and key by that here? But that probably feels a bit much... and is a it bit of a duplication of logic as well...

Am cool with this as is, just wanted tor raise the consideration 👍

rohan-bes (approved):
LGTM!

packages/data-broker/src/services/dhis/DhisService.js
IgorNadj: Agreed. To cover this I thought I'd add some tests, will submit the PR soon

PR #3286, opened 2021-09-27
Switched output over to using yup validation for report-server output step

rohan-bes:
Just a small refactor here to use `yup` object validation over our original hand crafted version. Will be using these changes later down the line when adding functionality for: https://linear.app/bes/issue/RN-28/date-selector-shows-latest-data-by-default

### DIFF
--- a/packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
@@ -116,6 +116,29 @@ describe('output', () => {
     });
 
     describe('handle categoryField', () => {
+      it('throws error if categoryField matches any columns', () => {
+        const output = buildOutput({
+          type: 'matrix',
+          categoryField: 'InfrastructureType',
+          rowField: 'FacilityType',
+          columns: ['InfrastructureType', 'Laos', 'Tonga'],
+        });
+        expect(() => {
+          output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES);
+        }).toThrow();
+      });
+
+      it('throws error if categoryField matches rowField', () => {
+        const output = buildOutput({
+          type: 'matrix',
+          categoryField: 'FacilityType',
+          rowField: 'FacilityType',
+        });
+        expect(() => {
+          output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES);
+        }).toThrow();
+      });
+
       it('can perform with categoryField', () => {
         const expectedData = {
           columns: [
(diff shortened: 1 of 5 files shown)


packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
```diff
@@ -116,6 +116,29 @@ describe('output', () => {
...
+      it('throws error if categoryField matches any columns', () => {
+        const output = buildOutput({
+          type: 'matrix',
+          categoryField: 'InfrastructureType',
+          rowField: 'FacilityType',
+          columns: ['InfrastructureType', 'Laos', 'Tonga'],
+        });
+        expect(() => {
+          output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES);
+        }).toThrow();
```
IgorNadj, line 128: I like to specify the message we're expecting, so we don't get false negatives if it starts throwing something else like undefined is not a function or whatever.

packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
```diff
@@ -213,6 +236,89 @@ describe('output', () => {
...
+      });
+
+      it('throws error if rowField not provided ', () => {
+        const output = buildOutput({
+          type: 'matrix',
+          categoryField: 'FacilityType',
+        });
+        expect(() => {
+          output(MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES);
```
IgorNadj, line 269: Maybe it's just me + Monday morning, but it looks like the function argument MULTIPLE_TRANSFORMED_DATA_WITH_CATEGORIES is never used in any of these calls added in the PR?

IgorNadj (approved):
LGTM, couple minor comments

Much cleaner with yup, nice one!

packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
rohan-bes: No, you're totally right. We need to pass in some kind of array of values to this function, but given that we're just testing config validation here it could be an empty array for all intents and purposes. I'll change that 👍

packages/report-server/src/__tests__/reportBuilder/output/output.test.ts
rohan-bes: I did think of this, and given you feel that way too I'm happy to update these tests. Sometimes I worry it makes the tests a bit too brittle, but it's probably worth it especially since these can be user facing errors

PR #3289, opened 2021-09-27
RN-104 Fix duplicate fetch - add tests

IgorNadj:
This PR adds tests for previous PR #3284

Easiest way to review this PR is commit by commit. 
- Refactor commits are straightforward renames/code moving around with no changes.
- Change commits do not contain refactors. Sometimes adding ?w=1 helps view these commits if there are indentation changes


packages/data-broker/src/__tests__/services/dhis/DhisService.test.js
```diff
@@ -0,0 +1,286 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import * as GetDhisApiInstance from '../../../services/dhis/getDhisApiInstance';
+import { DATA_SOURCES, DATA_VALUES, DHIS_REFERENCE, SERVER_NAME } from './DhisService.fixtures';
+import { DhisService } from '../../../services/dhis';
+import { createModelsStub, stubDhisApi } from './DhisService.stubs';
+
+describe('DhisService', () => {
```
IgorNadj, line 11: Moving back in a single file has two benefits:
- Does not hide the fact that DhisService is doing a lot. If this test suite becomes too long that's a code smell that the class is doing too much and needs to be broken up. We have done a little with breaking out pullers to (IMO) good effect, we could in the future do the same for Pushers and Deleters.
- Lets us more easily reason about how test setups and teardowns interact. Previously you had to look at 3 separate files to check the initialisers of the DhisService under test and the mocks, and run the risk of having tests interfere with each other, now these are all inline.

packages/data-broker/src/services/dhis/pullers/DeprecatedEventsPuller.js
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * This is a deprecated puller which invokes a slow DHIS2 api ('/events')
+ * and returns an obsolete data structure (equivalent to the raw DHIS2 events).
+ * It is invoked using the `options.useDeprecatedApi` flag
+ *
+ * TODO Delete this puller as soon as all its past consumers have migrated over to
+ * the new (non-deprecated) method
+ */
+export class DeprecatedEventsPuller extends EventsPuller {
```
IgorNadj, line 16: Split this out into its own class because EventsPuller.test.js was too long and hard to read.

packages/data-broker/src/__tests__/services/dhis/DhisService.test.js
```diff
@@ -0,0 +1,286 @@
...
+        expect(getDhisApiInstanceSpy)
+          .toHaveBeenNthCalledWith(2, { entityCodes: ['TO'], isDataRegional: true, }, expect.anything());
+
+        // expect those apis to be passed to pull
+        expect(dhisService.analyticsPuller.pull)
+          .toHaveBeenCalledOnceWith([{ myDhisApi: 1 }, { myDhisApi: 2 }], expect.anything(), expect.anything())
+      });
+
+      it('only passes unique apis it gets to pull', async () => {
+        // (See RN-104)
```
IgorNadj, line 258: This is the new test for RN-104

rohan-bes (approved):
Really nice @IgorNadj !! Thanks for taking the time to clean up the test structure as well as add these tests! I feel safer about the slightly risky logic for checking if we're using the same dhis api now that's it's guarded by a test 👍

PR #3293, opened 2021-09-27
Hotfix map components regresssion

tcaiger:
### Changes:

- fix Transparent shaded polygons not showing
- update AreaTooltips to new format

---

### DIFF
--- a/packages/ui-components/src/components/Map/AreaTooltip.js
+++ b/packages/ui-components/src/components/Map/AreaTooltip.js
@@ -8,39 +8,16 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Tooltip } from 'react-leaflet';
 import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/components/Map/AreaTooltip.js
```diff
@@ -81,7 +37,23 @@ export const AreaTooltip = ({
       permanent={permanent}
       interactive={permanent}
     >
-      <div style={{ display: 'grid' }}>{displayText}</div>
+      {text ? (
+        <Grid>{text}</Grid>
```
tcaiger, line 41: This text prop is only used in lesmis. I think it would be better to refactor lesmis to use a simpler Tooltip component but I will do that in another PR after the release.

billli0 (approved):
LGTM!

PR #3294, opened 2021-09-27
Hotfix: Meditrak-App AppCenter build fix

rohan-bes:
### Issue https://beyondessential.slack.com/archives/C01MAA3NKM1/p1632710469006000

So I believe this issue was caused by the release of NDK 23 (https://developer.android.com/ndk/downloads/revision_history)

I'm not sure which version our appcenter builds were using previously, but I think we've been bumped up to the latest version which does play nicely with gradle tools lower than 4.0.

I tried manually setting our NDK version to 22, but I couldn't get this to work. Instead I went with bumping our gradle tools version to 4.0.1 which was easier and seems the better fix 👍

### DIFF
--- a/packages/meditrak-app/android/app/src/main/AndroidManifest.xml
+++ b/packages/meditrak-app/android/app/src/main/AndroidManifest.xml
@@ -12,7 +12,6 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
     <uses-sdk
-        android:minSdkVersion="19"
         android:targetSdkVersion="26" />
 
     <application
--- a/packages/meditrak-app/android/app/build.gradle
+++ b/packages/meditrak-app/android/app/build.gradle
@@ -133,7 +133,7 @@ android {
         ndk {
             abiFilters "armeabi-v7a", "x86", "arm64-v8a", "x86_64"
         }
-        ndkVersion 20.1.5948944
+        ndkVersion "20.1.5948944"
     }
     signingConfigs {
         debug {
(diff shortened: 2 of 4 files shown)


packages/meditrak-app/android/app/src/main/AndroidManifest.xml
```diff
@@ -12,7 +12,6 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
 
     <uses-sdk
-        android:minSdkVersion="19"
```
rohan-bes, line 15: Not exactly sure why but gradle 4.0 doesn't support this key here... Instead it said to specify it in our build.gradle file, which it turns out we were already doing... Weirdly enough our `minSdkVersion` in build.gradle is set to 21 whereas here it was 19...

I don't know which config was being observed previously...

edmofro (approved):
The irony is that we'll definitely need to regression test meditrak-app now! But we can hit the release button before that's complete, given it's a manual release process for the app 

Thanks for battling this one out, lgtm!

rohan-bes:
> The irony is that we'll definitely need to regression test meditrak-app now! But we can hit the release button before that's complete, given it's a manual release process for the app

Good point @edmofro, do you think it would be best just to wait til the release is over before merging this? Either way might be safer to QA the APK I built from this branch before merging 👍

edmofro:
Yep let's go with waiting, and actually we could just target dev with this instead

rohan-bes:
Closing in favour of: https://github.com/beyondessential/tupaia/pull/3296

PR #3298, opened 2021-09-28
WAI-956: Reduce feature branch deployment time by 30 minutes

edmofro:
### Issue #:
WAI-956

Changes:
- Use a common `lodash` version across all dependencies. To do this, I combined lines in the `yarn.lock` file where `lodash` was being unnecessarily resolved using different versions, despite all dependents accepting any version within the same minor. This saved the majority of time, the reason being that it was having to copy across the (rather bloated) lodash library into every dependency that used it. Now, just one version is installed in the root `node_modules`, which is then utilised everywhere else. This is the default behaviour `yarn` uses to hoist nested dependencies, but it will only hoist one version, so when there are many being resolved, you end up with a lot of duplication.
- Changed legacy @beyondessential/tupaia-access-policy so that it doesn't unnecessarily bring in dev dependencies when included as a dependency. This saved a further 2 minutes. See https://github.com/beyondessential/tupaia-access-policy/commit/519e3c708593f401490c59651a9df7f0a4dc5c9b for the actual change

Together these save around 30 minutes of the 45 minutes spent on `yarn install` when run as part of the startup script on fresh EC2 deployments. Similar proportional savings wherever else we do `yarn install`, though smaller raw time savings (seems particularly slow on those EC2 machines).

Other strategies attempted:
- Completely wipe `yarn.lock` so that like lodash, all other dependencies that can have versions consolidated, are. This saves a further 8 minutes, but also creates build issues with some incompatible modules that would take a while to resolve. I hope to do this, but it was taking a while and I'd prefer to get the quick win of the initial 30 minutes committed.
- Remove `storybook` from dependencies, as it can be installed globally by devs who use it. This saved a further 5 minutes on top of this PR, but actually only about 20 seconds if we can do the above point of completely wiping `yarn.lock`, so I'm hesitant to do it without first putting some more effort into that (better) strategy.
- Remove `meditrak-app` from being managed through yarn workspaces, and manage its dependencies completely independently. This saves a further 2 minutes off the completely wiped `yarn.lock` version, which doesn't seem worth it for the confusion it might create.

Timings recorded [here](https://docs.google.com/spreadsheets/d/1czv9uwI55OFuVX2fPmN8cyEPbo5A4UyZFzvgvLDC5JU/edit?usp=sharing)

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -26,7 +26,7 @@
     "randomatic": "^3.1.1"
   },
   "devDependencies": {
-    "@beyondessential/tupaia-access-policy": "^2.5.0",
+    "@beyondessential/tupaia-access-policy": "^2.5.1",
     "@tupaia/access-policy": "3.0.0",
     "@tupaia/database": "1.0.0"
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -111,15 +111,6 @@
   resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.13.11.tgz#9c8fe523c206979c9a81b1e12fe50c1254f1aa35"
   integrity sha512-BwKEkO+2a67DcFeS3RLl0Z3Gs2OvdXewuWjc1Hfokhb5eQWP9YRYH1/+VrVZvql2CfjOiNGqSAFOYt4lsqTHzg==
 
-"@babel/compat-data@^7.9.6":
-  version "7.9.6"
-  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.9.6.tgz#3f604c40e420131affe6f2c8052e9a275ae2049b"
-  integrity sha512-5QPTrNen2bm7RBc7dsOmcA5hbrS4O2Vhmk5XOL4zWW/zD/hV0iinpefDlkm+tBBy8kDtFaaeEvmAqt+nURAV2g==
-  dependencies:
-    browserslist "^4.11.1"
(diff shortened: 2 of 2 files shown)


kael89 (approved):
Awesome change @edmofro !! It always amazes me when such small and simple to implement (but not to detect) changes result in such big gains!

I guess we should draw some conclusions from this experience!
* ~Don't use `lodash` at all~ - haha just kidding, because we've discussed this in the past 😛 
* Somehow "lock" dependency versions...  Perhaps start using _exact_ versions instead of version ranges? Version ranges sound like a great idea in principle, but perhaps they don't work out well in a monorepo context. It would be interesting to see if there is any relevant industry knowledge/best practices on the subject
* Find another way to instruct `yarn` to always resolve dependencies in a common compatible version

edmofro:
> Somehow "lock" dependency versions... Perhaps start using exact versions instead of version ranges? Version ranges sound like a great idea in principle, but perhaps they don't work out well in a monorepo context. It would be interesting to see if there is any relevant industry knowledge/best practices on the subject

Good thought, but actually we can't really do this because new dependencies we introduce may have a nested dependency (e.g. on lodash) that is a higher version than the one used by ourselves or any other dependency already included, which is when the extra lodash gets added to our `yarn.lock`. In this case we aren't in control of that third party dependency's requested lodash version. 

We can chat more about other strategies on the ticket - I see you've left a big comment there! For now I'll merge this so it has some time to get passive regression testing across it before release.

PR #3299, opened 2021-09-28
RN-28: Added support for 'useDatesOfData' as the defaultTimePeriod for the Date Picker

rohan-bes:
### Issue RN-28:

Bit of a roundabout implementation here. For RN-27, we want to be able to automatically detect the latest available data and to show that in the dashboard. In order to do this we will perform the following steps:
1. Fetch data for all time
2. Aggregate data to the MOST_RECENT and then only return records that are match the highest value in their batch
3. Attach metadata to the response stating the period the data is for (this must be done as we don't wish to show the period in the response data itself)
4. Have the frontend determine how to set the date for the date picker based on the metadata in the response

Changes were made in the frontend to support (1), and (4). Changes were made in the report-server to support (3). (2) will be done in the report itself for this dashboard.

Some of the decisions made for the report-server changes I'm not entirely happy with, so would appreciate feedback if the reviewer is also uneasy with them.

**Note:** Couple bugs in the latest merge conflicts so this may not run locally atm. Will fix asap but wanted to get review up!

### DIFF
--- a/packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
+++ b/packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
@@ -5,9 +5,9 @@
 
 import { yup } from '@tupaia/utils';
 
-import { Row } from '../../../types';
+import { Response } from '../../types';
 import { MatrixBuilder } from './matrixBuilder';
-import { Matrix, MatrixParams } from './types';
+import { MatrixParams } from './types';
 
 const paramsValidator = yup.object().shape(
   {
@@ -16,6 +16,7 @@ const paramsValidator = yup.object().shape(
         ? yup.array().of(yup.string().required())
         : yup.mixed<'*'>().oneOf(['*'], "columns must be either '*' or an array"),
     ),
+    excludeFields: yup.array().of(yup.string().required()),
     rowField: yup
       .string()
       .required()
@@ -55,24 +56,30 @@ const paramsValidator = yup.object().shape(
   [['rowField', 'categoryField']],
 );
 
-const matrix = (rows: Row[], params: MatrixParams): Matrix => {
-  return new MatrixBuilder(rows, params).build();
+const matrix = (response: Response, params: MatrixParams) => {
+  return new MatrixBuilder(response, params).build();
 };
 
 const buildParams = (params: unknown): MatrixParams => {
   const validatedParams = paramsValidator.validateSync(params);
-  const { columns, categoryField, rowField } = validatedParams;
-  const includeFields = columns || '*';
+  const {
+    columns = '*',
+    excludeFields: excludeFieldsParams = [],
+    categoryField,
+    rowField,
+  } = validatedParams;
 
-  const excludeFields = categoryField ? [categoryField, rowField] : [rowField];
+  const excludeFields = categoryField
+    ? [categoryField, rowField, ...excludeFieldsParams]
+    : [rowField, ...excludeFieldsParams];
 
   return {
-    columns: { includeFields, excludeFields },
+    columns: { includeFields: columns, excludeFields },
     rows: { categoryField, rowField },
   };
 };
 
 export const buildMatrix = (params: unknown) => {
   const builtParams = buildParams(params);
-  return (rows: Row[]) => matrix(rows, builtParams);
+  return (response: Response) => matrix(response, builtParams);
 };
--- a/packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
+++ b/packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
@@ -1,5 +1,6 @@
 import pick from 'lodash.pick';
-import { Row } from '../../../types';
+import { Row, PeriodMetadata } from '../../../types';
+import { Response } from '../../types';
 import { MatrixParams, Matrix } from './types';
 
 /** TODO: currently we are using 'dataElement' as a key in rows to specify row field (name),
@@ -12,22 +13,27 @@ const NON_COLUMNS_KEYS = [CATEGORY_FIELD_KEY, ROW_FIELD_KEY];
 export class MatrixBuilder {
   private rows: Row[];
 
+  private metadata: PeriodMetadata;
+
   private matrixData: Matrix;
 
   private params: MatrixParams;
 
-  constructor(rows: Row[], params: MatrixParams) {
-    this.rows = rows;
+  constructor(response: Response, params: MatrixParams) {
+    const { results, ...metadata } = response;
+    this.rows = results;
+    this.metadata = metadata;
     this.params = params;
     this.matrixData = { columns: [], rows: [] };
   }
 
   public build() {
     this.buildColumns();
+    this.buildPeriod();
     this.buildRows();
     this.adjustRowsToUseIncludedColumns();
     this.buildCategories();
-    return this.matrixData;
+    return { results: this.matrixData, ...this.metadata };
   }
 
   private buildColumns() {
@@ -58,20 +64,37 @@ export class MatrixBuilder {
     assignColumnSetToMatrixData(columnsInArray);
   }
 
+  private buildPeriod() {
+    const isString = (val: unknown): val is string => typeof val === 'string';
+    const periods = this.rows
+      .map(row => row.period)
+      .filter(isString)
+      .sort();
+
+    if (periods.length > 0) {
+      const earliestPeriod = periods[0];
+      const latestPeriod = periods[periods.length - 1];
+      this.metadata.period = {
+        ...this.metadata.period,
+        reportStart: earliestPeriod,
+        reportEnd: latestPeriod,
+      };
+    }
+  }
+
   private buildRows() {
     const rows: Row[] = [];
+    const { excludeFields } = this.params.columns;
     const { rowField, categoryField } = this.params.rows;
 
     this.rows.forEach(row => {
-      let newRows: Row;
+      const newRow = { ...row };
+      excludeFields.forEach(field => delete newRow[field]);
       if (categoryField) {
-        const { [rowField]: rowFieldData, [categoryField]: categoryId, ...restOfRow } = row;
-        newRows = { [ROW_FIELD_KEY]: rowFieldData, [CATEGORY_FIELD_KEY]: categoryId, ...restOfRow };
-      } else {
-        const { [rowField]: rowFieldData, ...restOfRow } = row;
-        newRows = { [ROW_FIELD_KEY]: rowFieldData, ...restOfRow };
+        newRow[CATEGORY_FIELD_KEY] = row[categoryField];
       }
-      rows.push(newRows);
+      newRow[ROW_FIELD_KEY] = row[rowField];
+      rows.push(newRow);
     });
 
     this.matrixData.rows = rows;
--- a/packages/report-server/src/reportBuilder/output/types.ts
+++ b/packages/report-server/src/reportBuilder/output/types.ts
@@ -1,0 +1,10 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { Row, PeriodMetadata } from '../types';
+
+export interface Response extends PeriodMetadata {
+  results: Row[];
+}
--- a/packages/report-server/src/reportBuilder/output/types.ts
+++ b/packages/report-server/src/reportBuilder/output/types.ts
@@ -3,8 +3,8 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { Row, PeriodMetadata } from '../types';
+import { Row, DataElementsMetadata, PeriodMetadata } from '../types';
 
-export interface Response extends PeriodMetadata {
+export interface Response extends DataElementsMetadata, PeriodMetadata {
   results: Row[];
 }
--- a/packages/utils/src/object.js
+++ b/packages/utils/src/object.js
@@ -264,3 +264,6 @@ export const camelKeys = object =>
 
 export const snakeKeys = object =>
   Object.fromEntries(Object.entries(object).map(([key, value]) => [snake(key), value]));
+
+export const isNullishOrEmptyObject = obj =>
+  obj === null || obj === undefined || (typeof obj === 'object' && Object.keys(obj).length < 1);
--- a/packages/web-config-server/src/apiV1/report.js
+++ b/packages/web-config-server/src/apiV1/report.js
@@ -1,4 +1,9 @@
-import { CustomError, replaceValues, convertDateRangeToPeriodQueryString } from '@tupaia/utils';
+import {
+  CustomError,
+  replaceValues,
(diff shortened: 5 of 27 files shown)


packages/report-server/src/reportBuilder/output/types.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { Row, PeriodMetadata } from '../types';
+
+export interface Response extends PeriodMetadata {
+  results: Row[];
+}
```
rohan-bes, line 10: Decided to now return the period metadata along with results here. Should be reminiscent of the metadata that is returned from data-broker itself when we fetch analytics/events.

I believe it also returns a `dataElementCodesToNames` payload, so I will add that to the typings...

packages/web-config-server/src/apiV1/report.js
```diff
@@ -67,8 +72,18 @@ export class ReportHandler extends DataAggregatingRouteHandler {
...
+    const reportMetadata = {};
+    if (period?.reportStart) {
+      reportMetadata.dataStartDate = periodToDateString(period.reportStart, false);
+    }
+    if (period?.reportEnd) {
+      reportMetadata.dataEndDate = periodToDateString(period.reportEnd, true);
+    }
+    return Array.isArray(results)
+      ? { data: results, ...reportMetadata }
+      : { ...results, ...reportMetadata };
```
rohan-bes, line 86: As of this change, we now receive the following when requesting data for a dashboard:
```
{
  results, // data for report
  startDate, // requested start date
  endDate, // requested end date
  dataStartDate, // date of earliest data from report
  dataEndDate, // date of latest data from report
}
```

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
```diff
@@ -222,23 +222,33 @@ export class MatrixWrapper extends Component {
...
       return null; // Not using a period selector
     }
 
-    const datePickerLimits = getLimits(viewContent.periodGranularity, viewContent.datePickerLimits);
+    const { startDate: defaultStartDate, endDate: defaultEndDate } = getDefaultDatePickerDates(
+      viewContent,
+    );
+
+    const startDate = urlStartDate || defaultStartDate;
+    const endDate = urlEndDate || defaultEndDate;
```
rohan-bes, line 238: The URL reigns supreme! No matter what we respect the url dates, but in their absence we must show the default date.

This logic is duplicated in a couple places, so will refactor to a common util

packages/web-frontend/src/utils/periodGranularities.js
```diff
@@ -259,19 +261,40 @@ const getDefaultDatesForRangeGranularities = (periodGranularity, defaultTimePeri
...
 
   // we need a valid granularity to proceed
   if (!periodGranularity) {
     return {};
   }
 
+  // Fetch data for specified range and determine default dates based on that
+  const { useDatesOfData, ...restOfDefaultTimePeriod } = defaultTimePeriod || {};
+  if (useDatesOfData) {
+    return getDefaultDatesForRangeGranularities(periodGranularity, restOfDefaultTimePeriod);
```
rohan-bes, line 275: If using `useDatesOfData` always fetch for a range regardless of the period granularity. It's up to the report to implement logic to ensure that it only returns data for a single period.

Not ideal to have this implicit coupling between the report and the date picker behaviour, but I do feel it's somewhat unavoidable...

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
```diff
@@ -58,20 +64,37 @@ export class MatrixBuilder {
...
+    if (periods.length > 0) {
+      const earliestPeriod = periods[0];
+      const latestPeriod = periods[periods.length - 1];
+      this.metadata.period = {
+        ...this.metadata.period,
+        reportStart: earliestPeriod,
+        reportEnd: latestPeriod,
+      };
+    }
+  }
```
rohan-bes, line 83: So for now we're just doing this for matrixes, as that's what's required for RN-27. However I feel we really ought to implement output steps for the other tupaia chart types and we can make this logic common across them 👍

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
```diff
@@ -16,6 +16,7 @@ const paramsValidator = yup.object().shape(
         ? yup.array().of(yup.string().required())
         : yup.mixed<'*'>().oneOf(['*'], "columns must be either '*' or an array"),
     ),
+    excludeFields: yup.array().of(yup.string().required()),
```
rohan-bes, line 19: So I couldn't think of a clean way to have the output step determine the start/end dates for the data without including the `period` column in the report and treating it as a magic field that the output step reads. A bit hacky, but also a bit pragmatic I feel?

The real downside to me is that if we wish to exclude the `period` column from the report output then we have to manually exclude it here... Tedious and unexpected.

I might try and brainstorm some more to see if there's a cleaner pattern we could use here, but would love if the reviewer had some thoughts?

packages/utils/src/object.js
```diff
@@ -264,3 +264,6 @@ export const camelKeys = object =>
 
 export const snakeKeys = object =>
   Object.fromEntries(Object.entries(object).map(([key, value]) => [snake(key), value]));
+
+export const isNullOrEmptyObject = obj =>
+  obj === null || obj === undefined || !(typeof obj === 'object' && Object.keys(obj).length > 0);
```
edmofro, line 269: Nice tests! They picked this up. 
```suggestion
  obj === null || obj === undefined || (typeof obj === 'object' && !Object.keys(obj).length > 0);
```

packages/web-frontend/src/containers/MapOverlayBar/Control.js
```diff
@@ -137,6 +141,7 @@ export const Control = ({
 
   // Map overlays always have initial dates, so DateRangePicker always has dates on initialisation,
   // and uses those rather than calculating it's own defaults
+  console.log(mapOverlay, defaultDates);
```
edmofro, line 144: The last commit on this is called `WIP` and this is still here, shall I hold off review? Perhaps you could delete from the reviews channel and re-post once it's ready?

rohan-bes: Ah sorry, I just planned on dealing with that one case there as it showed up due to changes in the latest merge from dev, but felt the rest of the code was ready for review? Can hold off if you'd prefer til it's all finished though 👍

packages/utils/src/object.js
rohan-bes: Ah woops!! Hooray for tests 👍

packages/web-frontend/src/containers/MapOverlayBar/Control.js
edmofro: Oh I should have read right through your PR description, no that's all good I can review as is :-)

packages/utils/src/object.js
rohan-bes: Btw, curious to know if you think checking for `undefined` is correct here? I suppose it could say `isNullOrUndefinedOrEmptyObject` but that just feels long winded to me 😛

edmofro: I haven't actually got back to reviewing this, but seeing as you ask I'd go with `isNullishOrEmptyObject`

edmofro: Nullish seems pretty accepted as "null or undefined", see e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
```diff
@@ -58,20 +64,37 @@ export class MatrixBuilder {
...
+      .filter(isString)
+      .sort();
+
+    if (periods.length > 0) {
+      const earliestPeriod = periods[0];
+      const latestPeriod = periods[periods.length - 1];
+      this.metadata.period = {
+        ...this.metadata.period,
+        reportStart: earliestPeriod,
+        reportEnd: latestPeriod,
```
rohan-bes, line 80: I thought about overriding the `earliestAvailable` and `latestAvailable` values that we get from the intial fetch here, but then felt that maybe retaining both pieces of information could be valuable?

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
edmofro: Definitely agree!

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
edmofro: Did you want to come back to this?

packages/web-frontend/src/utils/periodGranularities.js
edmofro: I don't even see it as a requirement that the report returns data for a single period, that would depend on the type of viz being displayed. If it was a time series chart, go ahead and return all data as different periods, and it'll be fine! So in that way it's more of a coupling between the report and the display type, which is not quite as bad. 

I did think there was scope for a `usePeriodOfMostRecentData` type behaviour, where it respects a period granularity like `one_month_at_a_time` and trims all data prior to the most recent. However I see this would push the trimming logic into the report server in some way. Probably the cleanest way to do that is to define a new aggregation of `MOST_RECENT_PERIOD` which takes a config like `{ periodType: 'month' }`, and trims all data before that. That puts into the report building process but in a place that a) can be applied pretty consistently and easily to every report type that might use it, and b) can be fairly easily implemented in the db, so at least internally fetched data can have the trimming done at that level. Caveat being it would only work when the period wasn't being transformed during the report building process!

packages/web-frontend/src/utils/periodGranularities.js
```diff
@@ -259,19 +261,44 @@ const getDefaultDatesForRangeGranularities = (periodGranularity, defaultTimePeri
...
   }
-  return getDefaultDatesForRangeGranularities(periodGranularity, defaultTimePeriod);
+
+  const {
+    defaultTimePeriod,
+    startDate: requestedStartDate,
+    endDate: requestedEndDate,
+    dataStartDate,
+    dataEndDate,
+  } = viewConfig;
```
edmofro, line 296: I really don't like that post-data-build fields exist on an object called `config`, which I would think should refer to the static configuration options that are defined in the db and available to web-frontend earlier on. That's a problem that exists long before your PR though, so just a mention without any need to do anything here

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
edmofro: I actually feel there are a few too many ways to refer to dates flying around the place, especially when you factor in `dataStartDate` and `dataEndDate` that these get turned into before going to the front end

What about bringing some consistency with something like:
- `earliestRawData`, `latestRawData` for what's returned from initial fetch (will require some refactor...)
- `earliestReportData`, `latestReportData` for what's calculated here
- `earliestDataDate`, `latestDataDate` for what is passed to the front end (doesn't sit inside a `period` object, so needs the extra `Date` context)

I quite like maintaining the divide between `start`/`end` as a directive or request from the client, vs. `earliest`/`latest` as information about the response from the server

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
edmofro: Yeah I can't think of a way without a radically different architecture. Let's walk and talk this though in case we come up with something by bouncing off each other!

edmofro (changes requested):
A lot of work here! Nice one @rohan-bes 

Because this is a PR that spans quite a bit of the stack, it'd be a good one for a walk and talk so we can go through the overall architecture. I'm just slightly uncomfortable about the complexity vs. feature benefit tradeoff here, so would be keen to talk about that at the same time!

packages/web-frontend/src/components/View/MatrixWrapper/MatrixWrapper.js
rohan-bes: Ah thanks for the reminder! I forgot 🤦  will do 👍

packages/web-frontend/src/utils/periodGranularities.js
rohan-bes: That's a good point, the coupling isn't too bad, as the report being responsible for generating the type of data we want to display is already a bit of a requirement 👍 

As to using an aggregation vs report-server transforms to 'trim' the data, I can definitely see the benefit of this being done in aggregator, however in the specific case that I need this behaviour for it's a little bit awkward.

I need to display data for any facility that returned data in the same month as the most recent response from any facility, and I want to fetch at the country level, for multiple data elements. Our aggregations largely work independently across dimensions, whereas the filtering we'd need to do would be intentionally across all dimensions. Not saying it wouldn't be possible, but feels a bit of a breach of how aggregations are supposed to behave. However I could see the benefit of an aggregations that say 'returns all analytics in the same month as the most recent response per org unit / data element' as an independent thing 👍 

So I had actually planned on doing this in the report-server transforms instead. Was tossing up between just using an generic `excludeRows` to do this, or to add an alias transform called something like `excludeRowsNotInMostRecentMonth` (probably can come up with a better name 😛)

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
rohan-bes: Sweet, yes lets!

packages/report-server/src/reportBuilder/output/functions/matrix/matrixBuilder.ts
rohan-bes: Ah good call, I hadn't considered the semantics of `start` vs `earliest` but it's a poignant distinction 👍 Happy to make these changes 😄

packages/web-frontend/src/utils/periodGranularities.js
rohan-bes: Yeah good to point this out. Honestly wading through web-frontend proved to be the trickiest part of these changes by far. II think this area of the code is the one that has been most punished by our tendency to not quite deprecate legacy code 😅

I think a good refactor is in order, but out of the scope for this

rohan-bes:
@edmofro very glad to hear you have some concerns about the complexity tradeoff with all this, honestly I feel a similar way. Very happy for us to walk-n-talk this one, I'm in on thursday?

edmofro:
Aww man. Yeah I really appreciate your willingness to opt for the bigger picture rather than pushing this PR through, I think that's the right call but it is impressive you're willing to put it to the side!

It gives us some extra motivation to do more front end refactors around that date picker and data fetching flow in Tupaia, so let's both keep an eye out for the opportunity to slip that in alongside anything else that would benefit from it or touch the same components.

PR #3302, opened 2021-09-29
Multi map overlay selection checkbox

billli0:
### Issue #:
PHX-1's basic functionalities, now support:
- Multi map overlay checkbox.
- Show multi map overlay data in flippa table and download. 
- If multi map overlay ids in URL, automatically switch to multi map overlay checkbox mode.
- Show no map overlay selected message in flippa table.


packages/web-frontend/src/historyNavigation/historyMiddleware.js
```diff
@@ -57,54 +61,66 @@ export const historyMiddleware = store => next => action => {
...
+          { startDate, endDate },
+          periodGranularity,
+        ),
+      });
+      break;
+    }
+    case UNSELECT_MAP_OVERLAY: {
+      const { routing } = state;
+      const currentMapOverlayIdsArray = getLocationComponentValue(
+        routing,
```
kael89, line 101: We can use the `selectCurrentMapOverlayIds` selector here

packages/web-frontend/src/components/HierarchyItem.js
```diff
@@ -20,83 +20,73 @@
...
+      if (onClick) {
+        onClick();
+      }
+      setIsOpen(!isOpen);
+    };
 
-    const IconComponent = isOpen ? OpenIcon : ClosedIcon;
-    return <IconComponent style={styles.buttonIcon} />;
-  }
+    const OpenClosedIcon = () => {
```
kael89, line 57: The convention we use across the codebase for elements rendered inside the same element is using `render...` methods:

```suggestion
    const renderOpenClosedIcon = () => {
```

packages/web-frontend/src/components/HierarchyItem.js
```diff
@@ -20,83 +20,73 @@
...
+    dispatch,
+    ...otherProps
+  }) => {
+    const [isOpen, setIsOpen] = useState(false);
 
-    const hasChildren = hasNestedItems || (Array.isArray(nestedItems) && nestedItems.length > 0);
-    if (!hasChildren) {
-      return null;
-    }
+    const buttonOnClick = () => {
```
kael89, line 50: Prefixing props accepting callbacks with `on` (eg `onCick)` and their handlers with `handle`(`handleClick`) is a standard React convention:

```suggestion
    const handleClick = () => {
```

packages/web-frontend/src/components/HierarchyItem.js
```diff
@@ -20,83 +20,73 @@
...
-    if (this.props.onClick) {
-      this.props.onClick();
-    }
-    this.setState(state => ({ isOpen: !state.isOpen }));
-  }
-
-  renderOpenClosedIcon() {
-    const { nestedItems, hasNestedItems } = this.props;
-    const { isOpen } = this.state;
+export const HierarchyItem = React.memo(
```
kael89, line 33: Just double checking that the changes in this file are related to the card's scope? It seems to me that they are not (still worthy refactoring!), but it would be good to confirm for reviewing purposes

kael89 (changes requested):
Awesome work @billli0 , looking forward to this feature! Code structure is fine, my suggestions are mainly about renaming things 🚀

packages/web-frontend/src/sagas.js
```diff
@@ -905,16 +911,56 @@ function* fetchMeasureInfo(mapOverlayId) {
...
+        ...currentMapOverlayIds,
+        action.mapOverlayId,
+      ])
+    : [action.mapOverlayId];
+  yield put(setMapOverlay(mapOverlayIds.join(',')));
+}
+
+function* watchSelectMapOverlayChange() {
+  yield takeLatest(SELECT_MAP_OVERLAY, callSetMapOverlay);
+}
```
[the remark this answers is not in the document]
billli0: That's a good point @kael89 ! `callSetMapOverlay()` will call `SET_MAP_OVERLAY` triggered by `SELECT_MAP_OVERLAY`, I feel our code convention is to keep these similar logics (calling between actions) in `sagas.js` instead of `reducers.js`, which is not the best approach I would say.

However after taking a step back, I felt `selectMapOverlay()` can be a component function within `MapOverlayBar.js` instead of a global action, because it doesn't get called anywhere else. Do you feel that is better approach?

packages/web-frontend/src/selectors/mapOverlaySelectors.js
```diff
@@ -8,38 +8,46 @@
...
+  const mapOverlayIdsInString = getLocationComponentValue(location, URL_COMPONENTS.MAP_OVERLAY);
+  return mapOverlayIdsInString ? mapOverlayIdsInString.split(',') : [];
+});
 
-export const selectIsMapOverlayInHierarchy = createSelector(
-  [state => state.mapOverlayBar.mapOverlayHierarchy, (_, id) => id],
-  (mapOverlayHierarchy, id) => !!getMapOverlayFromHierarchy(mapOverlayHierarchy, id),
+export const selectCurrentMapOverlays = createSelector(
+  [state => selectMapOverlayByIds(state, selectCurrentMapOverlayIds(state))],
+  currentMapOverlays => currentMapOverlays,
```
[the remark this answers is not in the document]
billli0: Nice one!

packages/web-frontend/src/utils/mapOverlays.js
```diff
@@ -5,13 +5,28 @@
...
 
-export const getMapOverlayFromHierarchy = (mapOverlayHierarchy, targetMapOverlayId) => {
-  if (!targetMapOverlayId) {
-    return null;
+export const getMapOverlaysFromHierarchy = (mapOverlayHierarchy, targetMapOverlayIds) => {
+  if (!targetMapOverlayIds) {
+    return [];
   }
 
   const flattenMapOverlays = flattenMapOverlayHierarchy(mapOverlayHierarchy);
```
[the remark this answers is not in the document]
billli0: works for me👍

packages/web-frontend/src/containers/MapOverlayBar/Control.js
```diff
@@ -110,18 +110,23 @@ const ExpandedContent = styled.div`
...
   isMeasureLoading,
   onUpdateMeasurePeriod,
   children,
 }) => {
   const [isExpanded, setIsExpanded] = useState(false);
-  const mapOverlay = selectedMapOverlay || {};
+  const [isMeasureSelected, setIsMeasureSelected] = useState(false);
+  useEffect(() => {
+    setIsMeasureSelected(selectedMapOverlays.length > 0);
+  }, [selectedMapOverlays]);
```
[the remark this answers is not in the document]
billli0: That's correct🙏. I use it for fixing a bug but actually can do without it.

packages/web-frontend/src/sagas.js
```diff
@@ -905,16 +911,51 @@ function* fetchMeasureInfo(mapOverlayId) {
...
+
+function* fetchMeasureInfoOnlyIfMapOverlayIsSelected() {
+  const state = yield select();
+  const currentMapOverlayIds = selectCurrentMapOverlayIds(state);
+  // Clear measure if no mapOverlay is selected
+  if (currentMapOverlayIds.length === 0) {
+    yield put(clearMeasure());
+    return;
```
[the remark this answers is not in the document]
billli0: I was thinking the same but hesitate to put all logic together, however it looks nice and simpler to be together with less code paths. And we can decompose if needed later on.

packages/web-frontend/src/components/HierarchyItem.js
billli0: Yes you are right, it is out of scope of this card, but using class component produces more code comparing to functional component (e.g.: constructor, state and componentDidUpdated). Plus it's also easier if later we want to use useContext() or useReducer().

packages/web-frontend/src/sagas.js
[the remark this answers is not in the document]
billli0: Same as `UNSELECT_MAP_OVERLAY `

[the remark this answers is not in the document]
kael89: Oh I see what you mean... I like your suggestion for keeping it as a local function, if possible. Actions that call other actions that result in sagas... Things can get really complicated with redux and redux-saga 😄 !

packages/web-frontend/src/components/HierarchyItem.js
kael89: I agree - React functional components are more elegant! Just double checking to make sure I didn't miss reviewing any other aspect of the refactoring 🙂

packages/web-frontend/src/sagas.js
```diff
@@ -874,7 +872,13 @@ function* fetchMeasureInfo() {
   const organisationUnitCode = selectCurrentOrgUnitCode(state);
   const measureParams = mapOverlayIds.length > 0 && selectMapOverlayById(state, mapOverlayIds[0]);
 
-  if (mapOverlayIds.length === 0 || !organisationUnitCode || !measureParams) {
+  if (mapOverlayIds.length === 0) {
+    yield put(clearMeasure());
+    yield put(cancelFetchMeasureData());
```
kael89: Nice!

packages/web-frontend/src/screens/mobile/RegionScreen/index.js
```diff
@@ -293,22 +293,22 @@ const mapStateToProps = state => {
...
-  const { onChangeMapOverlay, onUnselectMapOverlay } = dispatchProps;
+  const { onChangeMapOverlay, onClearMeasure } = dispatchProps;
 
   return {
     ...stateProps,
     ...dispatchProps,
     ...ownProps,
     onChangeMapOverlay: mapOverlay =>
-      mapOverlay ? onChangeMapOverlay(mapOverlay.mapOverlayId) : onUnselectMapOverlay(),
+      mapOverlay ? onChangeMapOverlay(mapOverlay.mapOverlayId) : onClearMeasure(),
```
kael89: Just flagging that this will clear all currently selected measures here - which may be what we want, but I'm not sure

packages/web-frontend/src/containers/MapOverlayBar/MapOverlayBar.js
```diff
@@ -208,23 +209,44 @@ const mapStateToProps = state => {
...
-  const { dispatch } = dispatchProps;
-  const { currentMapOverlayIds } = stateProps;
+  const { dispatch, onSetMapOverlay, onClearMeasure } = dispatchProps;
+  const { currentMapOverlayIds, mapOverlayHierarchy } = stateProps;
+  const onSelectMapOverlay = mapOverlayId => {
+    const mapOverlayIds = sortMapOverlayIdsByHierarchyOrder(mapOverlayHierarchy, [
+      ...currentMapOverlayIds,
+      mapOverlayId,
+    ]);
+    onSetMapOverlay(mapOverlayIds.join(','));
```
kael89, line 229: I think I definitely prefer props and actions to be combined in such a way here, rather than having to add new sagas! This way we can have a few well-defined actions and sagas, then client components can combine them in any way they see fit.

kael89 (approved):
Nice work @billli0, approved!

packages/web-frontend/src/screens/mobile/RegionScreen/index.js
billli0: Yeah in [mobile.tupaia.org](mobile.tupaia.org) I couldn't find any map overlays, but correct me if I'm wrong... If so I think it should be fine to not support multi map overlay checkbox here.

kael89: Good point - let's not worry about it then!

PR #3306, opened 2021-09-30
Wai 61 lesmis admin panel

tcaiger:
### Changes:

- Example

---

### Screenshots:
See issue


packages/server-boilerplate/src/utils/useForwardUnhandledRequests.ts
```diff
@@ -0,0 +1,58 @@
...
+export const useForwardUnhandledRequests = (
+  app: Express,
+  target: string,
+  prefix?: string,
+  attachSession: Middleware = defaultAttachSession,
+) => {
+  const options = {
+    target,
+    changeOrigin: true,
+    pathRewrite: (path: string) => {
```
tcaiger, line 35: @chris-bes - I've extracted this function from admin-panel-server to server-boilerplate. I replaced app.path with path as it has the same value. Does that make sense to you?

chris-bes: Yep I like it! Makes more sense that it sits in `server-boilerplate`

packages/server-boilerplate/src/utils/useForwardUnhandledRequests.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import { Express, NextFunction, Request, Response } from 'express';
+import { IncomingMessage, ServerResponse } from 'http';
+import { createProxyMiddleware, fixRequestBody } from 'http-proxy-middleware';
```
rohan-bes, line 7: Do we need to move this dependency into `server-boilerplate` package.json?

packages/lesmis/src/utils/getApiUrl.js
```diff
@@ -0,0 +1,31 @@
...
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+export const getApiUrl = () => {
+  // if no env var, use sensible defaults based on the front end url
+  const { hostname } = window.location; // eslint-disable-line no-undef
+
+  // localhost becomes http://localhost:8060
+  if (hostname === 'localhost') {
```
rohan-bes, line 11: Not a major thing, but recently I experience a bug with this kind of logic in tests using the `supertest` framework as it uses `127.0.0.1` instead of `localhost`. Just a heads up in case you wanted to add it here 👍

packages/admin-panel/src/library.js
```diff
@@ -0,0 +1,8 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import 'react-table/react-table.css';
```
rohan-bes, line 5: Don't quite understand the reason for this import? Does it re-style all the components that are being exported here?

.eslintrc
```diff
@@ -120,6 +120,12 @@
       }
     },
     // Package overrides
+    {
+      "files": ["packages/admin-panel/**"],
+      "rules": {
+        "import/no-extraneous-dependencies": "off"
```
rohan-bes, line 126: Was there a particular import that was causing the need to disable this rule?

rohan-bes (approved):
Overall happy with this, so approving 👍 Most comments are questions, only one that might need to be addressed is the missing dependency in server-boilerplate.

Also, only if you have time, I'd be kinda keen to get a walkthrough of how the UX/UI works for this? Largely for curiosities' sake hehe

packages/lesmis-server/src/app/createApp.ts
```diff
@@ -39,4 +44,9 @@ export function createApp() {
     .post('/v1/report/:entityCode/:reportCode', handleWith(ReportRoute))
     .put('/v1/userEntityPermission', handleWith(UpdateUserEntityPermissionRoute))
     .build();
+
+  // Forward any unhandled request to meditrak-server
+  useForwardUnhandledRequests(app, MEDITRAK_API_URL, '/admin', attachSession);
```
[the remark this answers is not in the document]
tcaiger: Yeah I thought it best to stick to the orchestrator pattern if possible

packages/admin-panel/src/library.js
tcaiger: It literally just adds the react-table stylesheet to the bundle. It doesn't actually style the components per say, it's just a bunch of old school css selectors that apply to the whole page. Not sure if that answers your question? Without it there are no default react table styles.

packages/server-boilerplate/src/utils/useForwardUnhandledRequests.ts
tcaiger: Yes - good catch thanks

.eslintrc
tcaiger: Good question. It was for the dev dependencies which are imported in rollup. However when I thought about it a bit more I realised it was only one file so I have just disabled that rule in the file instead which is much more targeted.

PR #3307, opened 2021-09-30
Multi overlay checkbox other essentials

billli0:
### Issue # 
Other essential features in PHX-1 
 
-  Only two map overlays can be selected at the same time

### Changes:

- Example

---


packages/web-config-server/src/apiV1/utils/findAccessibleGroupedMapOverlays.js
```diff
@@ -158,14 +158,25 @@ const sortMapOverlayItems = (mapOverlayItems, relations) => {
...
+        id,
+        name,
+        linkedMeasures,
+        presentationOptions,
+        report_code: reportCode,
+        dataElementCode,
+        legacy,
+      }) => ({
+        id, // just for sorting purpose, will be removed later
+        measureIds: [id, ...(linkedMeasures || [])],
```
billli0, line 172: Need to add back `measureIds` in `measures` end points to work with `displayedMapOverlay` state and switch buttons.

packages/web-frontend/src/selectors/index.js
```diff
@@ -58,9 +58,9 @@ export {
 export {
   selectHasPolygonMeasure,
   selectAllMeasuresWithDisplayInfo,
-  selectAllMeasuresWithDisplayAndOrgUnitData,
```
billli0, line 61: selectAllMeasuresWithDisplayAndOrgUnitData() is not used by other files.

packages/web-frontend/src/selectors/index.js
```diff
@@ -58,9 +58,9 @@ export {
...
+  selectMeasureOptionsBySelectedMapOverlays,
   selectRenderedMeasuresWithDisplayInfo,
-  selectRadiusScaleFactor,
```
billli0, line 63: Delete selectRadiusScaleFactor() since it is duplicated in `ui-components`.

packages/ui-components/src/components/Map/Legend/Legend.js
```diff
@@ -47,43 +54,54 @@ const getLegendComponent = measureType => {
...
-  );
-});
+    return (
+      <>
+        {displayedLegends.map(series => {
+          const { type } = series;
+          const LegendComponent = getLegendComponent(type);
+          const isDisplayed = displayedMeasureIds.includes(series.key);
+          const hasIconLayer = type === MEASURE_TYPE_ICON;
+          const hasRadiusLayer = type === MEASURE_TYPE_RADIUS;
```
billli0, line 81: Hope that doesn't break any existing visuals🤞

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -0,0 +1,157 @@
...
+import DownArrow from '@material-ui/icons/ArrowDropDown';
+import MuiSwitch from '@material-ui/core/Switch';
+import CircularProgress from '@material-ui/core/CircularProgress';
+import { DateRangePicker } from '../../components/DateRangePicker';
+import { getDefaultDates, getLimits, GRANULARITY_CONFIG } from '../../utils/periodGranularities';
+import { Content, ContentText } from './Styles';
+import { TUPAIA_ORANGE } from '../../styles';
+import { setDisplayedMapOverlay } from '../../actions';
+
+const Switch = withStyles({
```
kael89, line 16: Not sure if we need `withStyles` here -perhaps we can use `styled` instead?

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -0,0 +1,157 @@
...
+  displayedMapOverlays,
+  onSetDisplayedMapOverlay,
+  // TODO: PHX-103: show both date pickers
+  showDatePickerOnlyAfterSecondTitle,
+}) => {
+  const [isSwitchedOn, setIsSwitchedOn] = useState(true);
+  useEffect(() => {
+    setIsSwitchedOn(displayedMapOverlays.includes(mapOverlay?.mapOverlayId));
+  }, [JSON.stringify(displayedMapOverlays)]);
+  const switchOnChange = () => {
```
kael89, line 63:
```suggestion

  const handleSwitchChange = () => {
```

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -0,0 +1,157 @@
...
+    if (isSwitchedOn) {
+      onSetDisplayedMapOverlay(
+        displayedMapOverlays.filter(mapOverlayId => mapOverlayId !== mapOverlay?.mapOverlayId),
+      );
+    } else {
+      onSetDisplayedMapOverlay([...displayedMapOverlays, mapOverlay?.mapOverlayId]);
+    }
```
kael89, line 70: This makes the logic flow a bit clearer:

```suggestion
    const newDisplayedOverlays = isSwitchedOn
      ? displayedMapOverlays.filter(mapOverlayId => mapOverlayId !== mapOverlay?.mapOverlayId)
      : [...displayedMapOverlays, mapOverlay?.mapOverlayId];
    onSetDisplayedMapOverlay(newDisplayedOverlays);
```

Also, I see we use `mapOverlay?.mapOverlayId` but I wonder what will happened if mapOverlayId is `undefined`? Can we set the displayed overlays to `undefined`, and will this eg be reflected in the url?

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -0,0 +1,157 @@
...
+  const showDatePicker = !!(isTimePeriodEditable && periodGranularity);
+  const updateMeasurePeriod = (startDate, endDate) => {
+    const period = GRANULARITY_CONFIG[periodGranularity].momentUnit;
+    onUpdateMeasurePeriod(moment(startDate).startOf(period), moment(endDate).endOf(period));
+  };
+  // Map overlays always have initial dates, so DateRangePicker always has dates on initialisation,
+  // and uses those rather than calculating it's own defaults
+  const startDate = mapOverlay?.startDate || defaultDates.startDate;
+  const endDate = mapOverlay?.endDate || defaultDates.endDate;
+  return (
```
kael89, line 85: Empty lines make the code breathe, similar to paragraphs in text:

```suggestion

  return (
```

I suggest restructuring this component using the following structure:
1. All non-function variable declarations go at the top (`useState`/other hooks go first, then other variable destructuring/assignments). Use a blank line at the end of this block. Can use more blank lines inside the block to group related operations if desired (may not be required here)
2. We then place function declarations. Most of the times we separate each of them with blank lines. `useEffect` should go either first or last here (not in between), to make it stand out

We don't define the above in a specific style guide, but you'll see them used as common practices in our functional components

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -75,13 +76,43 @@ export const selectHasPolygonMeasure = createSelector(
...
+    .flat();
+});
+
+export const selectMeasureOptionsByDisplayedMapOverlays = createSelector(
+  [state => state],
+  state => {
+    const { measureInfo } = state.map;
+    const { measureOptions } = measureInfo;
+    if (!measureOptions) {
+      return undefined;
```
kael89, line 98: Not sure if this should be `undefined` or `[]` - let's discuss

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -0,0 +1,157 @@
...
+    <>
+      <Content expanded={isExpanded} selected={isMeasureSelected} period={periodGranularity}>
+        <ContentText>{isMeasureLoading ? <CircularProgress size={22} /> : name}</ContentText>
+        <ToolsWrapper>
+          <Switch
+            checked={isSwitchedOn}
+            onChange={() => {
+              switchOnChange();
+            }}
```
kael89, line 94:
```suggestion
            onChange={switchOnChange}
```

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -0,0 +1,157 @@
...
+    if (isSwitchedOn) {
+      onSetDisplayedMapOverlay(
+        displayedMapOverlays.filter(mapOverlayId => mapOverlayId !== mapOverlay?.mapOverlayId),
+      );
+    } else {
+      onSetDisplayedMapOverlay([...displayedMapOverlays, mapOverlay?.mapOverlayId]);
+    }
+    setIsSwitchedOn(!isSwitchedOn);
+  };
+  const { periodGranularity, isTimePeriodEditable = true, name } = mapOverlay;
```
kael89, line 73: This is from the existing code:

```suggestion
  const { periodGranularity, isTimePeriodEditable = true, name } = mapOverlay || {};
```

You use `?.` with `mapOverlay`, so I assume that we can't be sure that it is actually defined.

kael89 (changes requested):
Nice one @billli0 ! This is a really cool feature that's taking form - looking forward to delivering this!

I left a few suggestions and questions. Let's discuss some of them in a call

packages/ui-components/src/components/Map/Legend/Legend.js
```diff
@@ -47,43 +54,54 @@ const getLegendComponent = measureType => {
...
+    }
 
-  return (
-    <LegendFrame className={className}>
-      {displayedLegends.map(series => {
-        const { type } = series;
-        const LegendComponent = getLegendComponent(type);
+    const legendTypes = displayedLegends.map(displayedLegend => displayedLegend.type);
+    const legendsHaveSameType =
+      legendTypes.length > 1 && legendTypes.every(val => val === legendTypes[0]);
```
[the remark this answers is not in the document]
billli0: Nice!🙌

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -76,6 +77,18 @@ export const selectHasPolygonMeasure = createSelector(
...
+export const selectCurrentMeasureIds = createSelector([state => state], state => {
+  const currentMapOverlayIds = selectCurrentMapOverlayIds(state);
+  const { mapOverlayHierarchy } = state.mapOverlayBar;
+  const hierarchyMapOverlays = flattenMapOverlayHierarchy(mapOverlayHierarchy);
+
+  return hierarchyMapOverlays.map(hierarchyMapOverlay =>
+    currentMapOverlayIds.includes(hierarchyMapOverlay.mapOverlayId)
+      ? hierarchyMapOverlay.measureIds
+      : [],
+  );
```
kael89, line 89: This reminds me of `selectDisplayedMeasureIds`... The difference is that `selectDisplayedMeasureIds` uses `.flat()` in the end. Perhaps we don't need it theere?

packages/web-frontend/src/containers/MapOverlayBar/MapOverlayBar.js
```diff
@@ -212,28 +213,31 @@ const mapStateToProps = state => {
...
-  const { currentMapOverlayIds, mapOverlayHierarchy } = stateProps;
+  const { currentMapOverlayIds } = stateProps;
   const onSelectMapOverlay = mapOverlayId => {
-    const mapOverlayIds = sortMapOverlayIdsByHierarchyOrder(mapOverlayHierarchy, [
-      ...currentMapOverlayIds,
+    onSetMapOverlay([
+      // Only two map overlays can be selected at the same time
+      ...(currentMapOverlayIds.length === 2 ? [currentMapOverlayIds[1]] : currentMapOverlayIds),
       mapOverlayId,
     ]);
```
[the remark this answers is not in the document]
billli0: Nice! Code is much cleaner.

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
billli0: I was thinking to use `styled` but that's how material ui want us to do for switch's custom colour: https://v4.mui.com/components/switches/#CustomizedSwitches.js 

I haven't tried very hard on use `styled`, but I think that should be fine to use `withStyles`?

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
billli0: It could be `{}` in `useEffect()` .

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
kael89: I believe that material-ui exposes its own custom styling helper (`withStyles`), but we generally prefer `styled` over it.

Please give `styled` a go, if it's too tricky then let's stick with `withStyles`

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
kael89: Cool, let's not worry about this then if it doesn't affect the user experience (happy to stay as is if front-end works as expected)

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
billli0: mapOverlay will always be an object in this component. I can assign `{}` to `mapOverlay` at props to make it clearer.

kael89: That would be great!

packages/web-frontend/src/selectors/measureSelectors.js
billli0: That's a bit confusing but they are slightly different. `selectCurrentMeasureIds` is what have been selected in checkbox while `selectDisplayedMeasureIds` is what have been switch on.

A particular example can be `selectCurrentMeasureIds` = [[measure1, measure2], [measure3]] `selectDisplayedMeasureIds` = [measure1, measure2], 

But I feel that we can filter the result of `selectCurrentMeasureIds` by displayed map overlay ids. Hope that make more sense?

billli0: I feel this part of logic can be simpler when we  are working on double date picker because there will be two responses which is separated by map overlay id. At that time we don't need to group measure ids here in reselector.

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
billli0: Switch use it's only logic to render check and uncheck colour, seems like styled can't easily override it's colour.. Or perhaps I make something wrong on css. Would you like to pair on it?

kael89: Yeap, happy to have a a look!

packages/web-frontend/src/containers/MapOverlayBar/MapOverlayBar.js
```diff
@@ -221,11 +221,12 @@ const mergeProps = (stateProps, dispatchProps, ownProps) => {
   const { dispatch, onSetMapOverlay, onClearMeasure } = dispatchProps;
   const { currentMapOverlayIds } = stateProps;
   const onSelectMapOverlay = mapOverlayId => {
-    onSetMapOverlay([
-      // Only two map overlays can be selected at the same time
-      ...(currentMapOverlayIds.length === 2 ? [currentMapOverlayIds[1]] : currentMapOverlayIds),
-      mapOverlayId,
-    ]);
+    const MAX_SELECTED_OVERLAYS = 2;
```
kael89, line 224: (Detail) let's move this in the file-scope: after the imports and before other code. Symbolic constants like this one (all uppercase) usually go there, so that devs can have a look at code "configuration" at a glance

PR #3309, opened 2021-09-30
Fix dropout and repetition overlays

chris-pollard:
### Issue #:
fix problems with Wai 6 & 7, dropout rate and repetition rate overlays.
Problem seemed to be that original overlay groups no longer exist, but could not verify.

### Changes:

- Add overlay groups.

---

### DIFF
--- a/packages/database/src/migrations/20210921051858-LESMISRepititionRatesDistrict-modifies-data.js
+++ b/packages/database/src/migrations/20210921051858-LESMISRepititionRatesDistrict-modifies-data.js
@@ -1,6 +1,6 @@
 'use strict';
 
-import { insertObject, generateId } from '../utilities';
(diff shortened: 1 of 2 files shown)


packages/database/src/migrations/20210921051858-LESMISRepititionRatesDistrict-modifies-data.js
```diff
@@ -182,6 +181,27 @@ const getGpiMapOverlay = (name, reportCode) => ({
...
+    parentCode: 'LESMIS_Repetition_Rate_Group',
+    children: [
+      {
+        name: 'District Level',
+        code: 'LESMIS_Repetition_Rate_District_Group',
+      },
+    ],
+  },
+];
+
```
EMcQ-BES, line 204: I'm not sure this is correct. Those overlay groups still exist in the database from what I can see:

image

PR #3311, opened 2021-10-01
WAI-5: Fix lesmis map spectrum colors

tcaiger:
### Issue #: Relates to all lesmis map overlays but the issue was raised on WAI-5

There are 2 commits on this PR. The first one is a simple refactor which I did as part of debugging the issue. The second one is the actual fix.

### Changes:
Always get measureData for the whole country so that the displayed area has a color relative to the rest of the country.

---

### Screenshots:
See WAI-5 Issue

### DIFF
--- a/packages/lesmis/src/api/queries/useMapOverlayReportData.js
+++ b/packages/lesmis/src/api/queries/useMapOverlayReportData.js
@@ -47,14 +47,10 @@ const getMeasureDataFromResponse = (overlay, measureDataResponse) => {
   };
 };
 
-const processMeasureInfo = ({ serieses, measureData, ...rest }) => {
-  const processedSerieses = serieses.map(series => {
+const processSerieses = (serieses, measureData) =>
+  serieses.map(series => {
     const { values: mapOptionValues, type } = series;
-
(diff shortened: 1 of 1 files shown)


packages/lesmis/src/api/queries/useMapOverlayReportData.js
```diff
@@ -154,15 +155,15 @@ export const useMapOverlayReportData = ({ entityCode, year }) => {
   };
 
   const { data: measureDataResponse, isLoading: measureDataLoading } = useQuery(
-    ['mapOverlay', entityCode, selectedOverlay, params],
+    ['mapOverlay', COUNTRY_CODE, selectedOverlay, params],
     () =>
-      get(`report/${entityCode}/${reportCode}`, {
+      get(`report/${COUNTRY_CODE}/${reportCode}`, {
```
edmofro, line 160: Funny thing is, this probably results in _less_ overall data being fetched, because you often view a map overlay at the country level first before zooming in

edmofro (approved):
That little refactor is bloody great, music to my eyes. Actual fix is :+1: too

PR #3312, opened 2021-10-01
RN-121 Fix report server SUM_PREVIOUS_EACH_DAY aggregation returns slice

IgorNadj:
RN-121
The issue was that SUM_PREVIOUS_EACH_DAY only works if you provide a period. Report server was not providing a period, only start/end dates.


edmofro (approved):
lgtm, and @rohan-bes has been making noises about the benefits of periods over start/end date recently, so this is aligned with that too :-)

PR #3314, opened 2021-10-01
WAI-956 Part 2: Speed up feature branch deployment by a further 8 minutes

edmofro:
### Issue #:
WAI-956

### Changes:
Same strategy as with https://github.com/beyondessential/tupaia/pull/3298, but this time with all of the `@babel` dependencies. Higher risk, so would need more regression testing.


packages/ui-components/src/components/Chart/Table/useChartDataExport.js
```diff
@@ -3,8 +3,8 @@
  *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { useDataTableExport } from '@tupaia/ui-components';
 import { getChartTableData } from './getChartTableData';
+import { useDataTableExport } from '../../DataTable/useDataTableExport';
```
edmofro, line 7: @tcaiger this was throwing an error, is there any issue with importing the file directly?

packages/ui-components/src/components/Map/utils/markerColors.js
```diff
@@ -15,8 +15,6 @@ import {
   TIME_COLOR_SCHEME,
 } from '../constants';
 
-export const DEFAULT_DISASTER_COLOR = 'orange';
```
edmofro, line 18: Same const is exported from `colors.js`

packages/ui-components/package.json
```diff
@@ -47,7 +47,6 @@
     "@rollup/plugin-commonjs": "^18.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-multi-entry": "^4.0.0",
-    "@svgr/rollup": "^5.5.0",
```
edmofro, line 50: A heavy package that's no longer needed if we use the svgs as components specified in js rather than .svg files

packages/ui-components/src/components/Chart/Table/useChartDataExport.js
tcaiger: Yes, we can't import the file directly because the Chart (and Map) directory gets bundled separately to the rest of the ui-components. It's definitely a trip hazard the way it's set up at the moment.

edmofro:
Closing in favour of https://github.com/beyondessential/tupaia/pull/3326

PR #3315, opened 2021-10-01
Fix dropout and repetition overlay bugs

chris-pollard:
### Issue #:
Fixing bugs on epic branch for WAI-6 and WAI-7
### Changes:
includes async addition, small text changes, new migration file for province level dropout.

- Example

---


packages/database/src/migrations/20211001060307-LESMISDropOutRateProvinceOverlay-modifies-data.js
```diff
@@ -0,0 +1,311 @@
...
+
+const getOverlayNames = (gradeNum, educationLevel, GPI) => {
+  if (gradeNum !== null) {
+    return {
+      name: `Grade ${educationLevel === 'p' ? gradeNum : gradeNum + 5} Dropout Rate ${
+        GPI ? 'GPI' : ''
+      }`,
+      dataElement: [
+        `dor_province_${educationLevel.concat(gradeNum)}_${GPI ? 'f' : 't'}`,
+        `${GPI ? 'dor_province_s1_m' : ''}`,
```
EMcQ-BES, line 30:
```suggestion
        `${GPI ? 'dor_province_s1_m' : ''}`,
        ... GPI ? ['dor_province_s1_m'] : [],
```

The way this is written currently these reports will wind up with an empty string in their dataElements array, I'm not sure if that's actually a problem but it's a little messy

packages/database/src/migrations/20211001060307-LESMISDropOutRateProvinceOverlay-modifies-data.js
```diff
@@ -0,0 +1,311 @@
...
+        insert: {
+          value: `=divide($${dataElement[0]},$${dataElement[1]})`,
+        },
+        exclude: [dataElement[0], dataElement[1]],
+      },
+    ],
+  },
+});
+
+const getTargetDistrictReport = (reportCode, dataElement) => ({
```
EMcQ-BES, line 125: This isn't actually being called anywhere right? Can't have a a target district report at province level. It's generally preferable to delete unused code because it adds unnecessary confusion

EMcQ-BES (changes requested):
The fixes to the old files are good! A couple of notes on the new migration but nothing too major.

packages/database/src/migrations/20211001060307-LESMISDropOutRateProvinceOverlay-modifies-data.js
chris-pollard: Thanks for the tip, will update that.

packages/database/src/migrations/20211001060307-LESMISDropOutRateProvinceOverlay-modifies-data.js
chris-pollard: Yes, you're right, will remove.

EMcQ-BES (approved):
Looks good, thanks for making those changes

PR #3320, opened 2021-10-04
RN-124: Fix error that appears in console when opening viz-builder

rohan-bes:
Bug occurred due to attempting to fetch available locations before selecting a hierarchy.

### Changes:
- Added handler for when project is undefined

---

### DIFF
--- a/packages/admin-panel/src/VizBuilderApp/api/queries/useLocations.js
+++ b/packages/admin-panel/src/VizBuilderApp/api/queries/useLocations.js
@@ -9,7 +9,10 @@ import { DEFAULT_REACT_QUERY_OPTIONS } from '../constants';
 export const useLocations = (project, search) =>
   useQuery(
     ['hierarchy', project, search],
-    () => get(`hierarchy/${project}/${project}`, { params: { search, fields: 'name,code' } }),
+    () =>
+      project
+        ? get(`hierarchy/${project}/${project}`, { params: { search, fields: 'name,code' } })
+        : [],
     {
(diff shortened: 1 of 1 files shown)


packages/admin-panel/src/VizBuilderApp/api/queries/useLocations.js
```diff
@@ -9,7 +9,10 @@ import { DEFAULT_REACT_QUERY_OPTIONS } from '../constants';
 export const useLocations = (project, search) =>
   useQuery(
     ['hierarchy', project, search],
-    () => get(`hierarchy/${project}/${project}`, { params: { search, fields: 'name,code' } }),
+    () =>
+      project
+        ? get(`hierarchy/${project}/${project}`, { params: { search, fields: 'name,code' } })
+        : [],
```
rohan-bes, line 15: Could've fixed this in a number of places:
1. Check before fetching in `admin-panel/src/VizBuilderApp/components/PreviewOptions.js` 
2. Check at the request (the approach shown here)
3. Check at the server if the hierarchy is defined

I don't like 3. at all, since it is ineffecient and allows requesting for `hierarchy/null/null` which seems incorrect. Option 1. seems decent too, but the interface is a bit awkward since we need to stub the response for `useQuery` which is a bit trickier...

Overall I think this is simplest?

kael89: Yes I like that!

kael89 (approved):
Thanks for fixing this, it was annoying! 🙂

PR #3321, opened 2021-10-04
Fix tupaia mobile not showing page

billli0:
### Issue #:
https://beyondessential.slack.com/archives/C01MAA3NKM1/p1633310619027900

### DIFF
--- a/packages/web-frontend/src/screens/mobile/RegionScreen/index.js
+++ b/packages/web-frontend/src/screens/mobile/RegionScreen/index.js
@@ -256,7 +256,7 @@ const mapStateToProps = state => {
   const { mapOverlayHierarchy, isExpanded } = state.mapOverlayBar;
   const { measureInfo, isMeasureLoading } = state.map;
   const { isGroupSelectExpanded } = state.dashboard;
-  const currentMapOverlay = selectCurrentMapOverlay(state);
+  const currentMapOverlay = selectCurrentMapOverlay(state) || {};
   const orgUnit = selectCurrentOrgUnit(state);
 
   const mobileListItems = getListItemsFromOrganisationUnitChildren(


kael89 (approved):

packages/web-frontend/src/screens/mobile/RegionScreen/index.js
```diff
@@ -256,7 +256,7 @@ const mapStateToProps = state => {
   const { mapOverlayHierarchy, isExpanded } = state.mapOverlayBar;
   const { measureInfo, isMeasureLoading } = state.map;
   const { isGroupSelectExpanded } = state.dashboard;
-  const currentMapOverlay = selectCurrentMapOverlay(state);
+  const currentMapOverlay = selectCurrentMapOverlay(state) || {};
```
kael89, line 259: On a second thought, should we do this in the selector level (as is the case eg in https://github.com/beyondessential/tupaia/blob/dev/packages/web-frontend/src/selectors/orgUnitSelectors.js#L131)? Do you think this may break other existing logic which checks whether the overlay is defined or not?

billli0: That's a good point. I felt a bit struggled to choose between `null` and `{}` to represent no map overlay found in this case. Personally I incline to use `null` because it is not straightforward to check with `{}`, e.g. in [MapTableModal.js](https://github.com/beyondessential/tupaia/blob/3a92bcf892c66276b8e0044083a4812ed9baeb7a/packages/web-frontend/src/containers/MapTableModal/MapTableModal.js#L36).

@kael89 I remember we discussed the standard for null data, but not object format specifically.. Do you think we should raise this in dev chat?

kael89: I see that a "falsy" value (like `undefined`, or `null`) is required to check for non existing overlays in other places, so happy with your above solution then

kael89 (approved):

billli0:
Branch name invalidated, changes have merged with https://github.com/beyondessential/tupaia/pull/3322

packages/web-frontend/src/screens/mobile/RegionScreen/index.js
kael89: On the `null` vs `{}` dilemma, I generally use the following rule of thumb:
* `null` for empty class instances (eg a model not found in the DB)
* `{}` for empty "plain" objects (with no methods), eg an empty config

In some cases it's a bit harder to make the call, since the object we're working with does not strictly fall into any of the above categories. Here I also feel that `null` would be more appropriate. On the other hand, this is the result of refactoring existing code, so the most convenient approach (less chance to introduce new bugs) would be following the existing approach on empty values.

billli0: > On the `null` vs `{}` dilemma, I generally use the following rule of thumb:
> 
> * `null` for empty class instances (eg a model not found in the DB)
> * `{}` for empty "plain" objects (with no methods), eg an empty config
> 

This is very useful! And yeah following the existing approach is true because I did create some bugs in development process and it cost time to fix..

PR #3324, opened 2021-10-04
Hoftix: Fix new viz save

kael89:
### Issue #:
https://linear.app/bes/issue/RN-126/vizbuilder-save-error-invalid-content-for-field-legacy


rohan-bes (approved):
LGTM

packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationExtractor.ts
```diff
@@ -77,7 +77,7 @@ export class DashboardVisualisationExtractor<
         name,
       },
       reportCode: code,
-      legacy,
+      legacy: !!legacy,
```
rohan-bes, line 80: So this means it's defaults to `not legacy`? That seems correct...

PR #3325, opened 2021-10-04
RN-125: Add 'dataElementCodeToName' function to report-server transforms

rohan-bes:
### Issue RN-125:

Basically just importing the `dataElementCodeToName` from the analytics fetch into the context, and then using a function to reference it from there. I thought about allowing users to directly reference the `metadata` object in the transform itself, and I think that could still have merit, but this method is cleaner and easier for users to work with 👍 

Note: only works with fetch analytics for now. If we want to add support when using fetch events I'd recommend we support that by returning metadata from fetch events also

A couple of cleanups in this PR as well, hopefully they aren't too distracting.

### DIFF
--- a/packages/report-server/src/reportBuilder/context/buildContext.ts
+++ b/packages/report-server/src/reportBuilder/context/buildContext.ts
@@ -15,7 +15,7 @@ export type ReqContext = {
   services: MicroServiceRequestContext['services'];
 };
 
-type ContextBuilder<K extends keyof Context> = (
+type ContextBuilder<K extends ContextProp> = (
   reqContext: ReqContext,
   data: FetchResponse,
 ) => Promise<Context[K]>;
@@ -34,8 +34,13 @@ const buildOrgUnits = async (reqContext: ReqContext, data: FetchResponse) => {
   );
 };
 
+const buildDataElementCodeToName = async (reqContext: ReqContext, data: FetchResponse) => {
+  return data.metadata?.dataElementCodeToName || {};
+};
+
 const contextBuilders: Record<ContextProp, ContextBuilder<ContextProp>> = {
   orgUnits: buildOrgUnits,
+  dataElementCodeToName: buildDataElementCodeToName,
 };
 
 function validateContextProp(key: string): asserts key is keyof typeof contextBuilders {
@@ -46,6 +51,14 @@ function validateContextProp(key: string): asserts key is keyof typeof contextBu
   }
 }
 
+const setContextValue = <Key extends ContextProp>(
+  context: Context,
+  key: Key,
+  value: Context[Key],
+) => {
+  context[key] = value;
+};
+
 export const buildContext = async (
   transform: unknown,
   reqContext: ReqContext,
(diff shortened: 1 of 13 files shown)


packages/report-server/src/reportBuilder/context/buildContext.ts
```diff
@@ -34,8 +34,13 @@ const buildOrgUnits = async (reqContext: ReqContext, data: FetchResponse) => {
   );
 };
 
+const buildDataElementCodeToName = async (reqContext: ReqContext, data: FetchResponse) => {
+  return data.metadata?.dataElementCodeToName || {};
```
rohan-bes, line 38: Right now this `metadata` is throwing a typescript error since it's not expected to be on the `FetchResponse` type. That's about to change as of: https://github.com/beyondessential/tupaia/pull/3299 so I'm happy to leave this as an error for now and then when that gets merged in it will go away 👍

packages/report-server/src/reportBuilder/context/buildContext.ts
```diff
@@ -46,21 +51,26 @@ function validateContextProp(key: string): asserts key is keyof typeof contextBu
   }
 }
 
+const setContextValue = <Key extends ContextProp>(
+  context: Context,
+  key: Key,
+  value: Context[Key],
+) => {
+  context[key] = value;
+};
```
rohan-bes, line 60: This is a little trick I've had to employ a few times. Typescript will complain if you do something like:
```
context[key] = await contextBuilders[key](reqContext, data))
```
This is because it can't be sure that `await contextBuilders[key](reqContext, data))` returns a value that's compatible with `context[key]` since it treats `key` as being all of the values for `ContextProp` at once. However, if you use a generic function like this it is able to perform the correct type assertion 👍

packages/report-server/src/reportBuilder/context/extractContextProps.ts
```diff
@@ -32,7 +32,7 @@ export const extractContextProps = (transform: unknown): ContextProp[] => {
     return [];
   }
 
-  const expressions = (transform as TransformParams[])
+  const expressions = transform
```
rohan-bes, line 35: This type cast wasn't mandatory, so decided to remove it and get rid of the `TransformParams` dependency

packages/report-server/src/reportBuilder/context/types.ts
```diff
@@ -4,7 +4,8 @@
  */
 
 export interface Context {
-  orgUnits: { code: string; name: string }[];
+  orgUnits?: { code: string; name: string }[];
```
rohan-bes, line 7: Decided to make all these optional, rather than needing to maintain a default value for each one. It does mean that the actual context dependant functions now need to assert that a value is present for their dependencies, but I don't actually mind that as it provides a check to ensure that the developer has correctly wired up the dependencies 👍

packages/report-server/src/reportBuilder/transform/index.ts
```diff
@@ -3,5 +3,5 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-export { contextConfig, TransformParser } from './parser';
-export { buildTransform, TransformParams } from './transform';
+export { buildTransform } from './transform';
```
rohan-bes, line 6: There is an (existing) dependency cycle here that I have been trying to debug to no avail. I'll admit it's definitely not my forte 😢 Have left it for now, but hopefully will clean it when we do: RN-115

EMcQ-BES (approved):
All looks really clean, appreciate all the comments explaining your thoughts. 👍

PR #3326, opened 2021-10-04
WAI-961: Speed up feature branch deployment by a further 8 minutes

edmofro:
### Issue #:
WAI-961

### Changes:
- Consolidate all `@babel` nested dependencies that are comfortable using the same version range, so we have less total versions of babel needing to be installed
- Remove `svgr`


packages/ui-components/src/components/Chart/Table/useChartDataExport.js
```diff
@@ -3,8 +3,8 @@
  *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { useDataTableExport } from '@tupaia/ui-components';
```
edmofro, line 6: @tcaiger this was throwing an error, is there any issue with importing the file directly?

packages/ui-components/src/components/Map/utils/markerColors.js
```diff
@@ -15,8 +15,6 @@ import {
   TIME_COLOR_SCHEME,
 } from '../constants';
 
-export const DEFAULT_DISASTER_COLOR = 'orange';
```
edmofro, line 18: Same const is exported from colors.js

packages/ui-components/src/components/Chart/Table/useChartDataExport.js
tcaiger: Yes, we can't import the file directly because the Chart (and Map) directory gets bundled separately to the rest of the ui-components. It's definitely a trip hazard the way it's set up at the moment.

packages/ui-components/src/components/Map/Table/useMapDataExport.js
```diff
@@ -3,8 +3,8 @@
  *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { useDataTableExport } from '@tupaia/ui-components';
 import { getMapTableData } from './getMapTableData';
+import { useDataTableExport } from '../../DataTable/useDataTableExport';
```
tcaiger, line 7: Same issue here with the relative import

packages/ui-components/package.json
```diff
@@ -47,7 +47,6 @@
     "@rollup/plugin-commonjs": "^18.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-multi-entry": "^4.0.0",
-    "@svgr/rollup": "^5.5.0",
```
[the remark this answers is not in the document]
tcaiger: Great call getting rid of it!

tcaiger (changes requested):
The main changes for babel and the svgs look good. Just need to revert the relative imports to `useDataTableExport`

packages/ui-components/src/components/Chart/Table/useChartDataExport.js
edmofro: Right. What kind of issues does it cause? The way I've set it up here works in lesmis and tupaia.org, and the size of the file is still nice and small (e.g. `lib/map/index.js` is 55 KB after the change), so I'm not sure where to look for what might need fixing tbh!

edmofro: Just checked and it was 54 KB before, so seems like a pretty acceptable increase 😀

tcaiger: Oh I mi-understood your message. I thought it was throwing an error after the change. The only issue that it causes is that it breaks the encapsulation between the 2 bundles which may be needed if we split them into different packages. However since that's just theoretical at this point, it is all good to pull in the file directly.

tcaiger (approved):
LGTM

packages/ui-components/src/components/Chart/Table/useChartDataExport.js
edmofro: Ah right! True, I see your point. Does fix an eslint issue and avoids confusion for now though, so let's split it properly if needed later. Thanks!

PR #3336, opened 2021-10-04
Update map overlay support for legacy introduction

EMcQ-BES:
### Issue #:
Resolves WAI-13 WAI-946 WAI-945

This is the final of a chain of branches, once it's approved we can merge the chain down into `wai-946-part-one` and do the bulk of the testing against that branch

### Changes:

- Update web config server to serve both legacy and new style map overlays
- Update usages of mapOverlayId to instead refer to map overlay code
- Update database usage to match new schema


packages/web-config-server/src/apiV1/measureData.js
```diff
@@ -80,18 +80,21 @@ const accessDeniedForMeasure = {
...
-    });
-  });
+  const measureDataResponses = overlays.map(
+    ({ code, data_builder_config: measureBuilderConfig }) => {
+      const { dataElementCode = 'value' } = measureBuilderConfig ?? {};
+      const { data } = measureDataResponsesByMeasureCode[code];
+
+      return data.map(obj => {
+        const { [dataElementCode]: value, ...restData } = obj;
+        return { [code]: value, ...restData };
```
edmofro, line 94: I found this section a bit hard to parse, but can't think of a better way - it's just quite clever with the dynamic destructuring that gets renamed to a new variable. Perhaps avoiding reuse of the word `value` would help a little
```suggestion
        const { [dataElementCode]: measureValue, ...restData } = obj;
        return { [code]: measureValue, ...restData };
```

packages/web-config-server/src/apiV1/measureData.js
```diff
@@ -330,18 +330,42 @@ export default class extends DataAggregatingRouteHandler {
...
-    const buildMeasure = async (measureId, ...args) => ({
-      [measureId]: await getMeasureBuilder(measureBuilder)(...args),
-    });
 
+    if (legacy) {
+      const {
+        data_builder: measureBuilder,
+        data_builder_config: measureBuilderConfig,
+      } = mapOverlay;
+      const { dataElementCode = 'value' } = measureBuilderConfig;
```
edmofro, line 339: Most other times with this pattern you've had safety around `measureBuilderConfig` being defined - is that not needed here?

packages/web-config-server/src/apiV1/measureData.js
```diff
@@ -257,7 +263,7 @@ export default class extends DataAggregatingRouteHandler {
       ...restOfPresentationConfig
     } = config;
 
-    const { dataSourceType = DATA_SOURCE_TYPES.SINGLE, periodGranularity } =
+    const { dataSourceType = DATA_SOURCE_TYPES.CUSTOM, periodGranularity } =
```
edmofro, line 266: What's the `CUSTOM` about?

packages/web-config-server/src/apiV1/measureData.js
```diff
@@ -330,18 +330,42 @@ export default class extends DataAggregatingRouteHandler {
...
       this.models,
       this.aggregator,
       dhisApi,
-      { ...this.query, dataElementCode },
-      { ...measureBuilderConfig, dataServices },
+      { ...this.query, dataElementCode: 'value' },
+      { reportCode: mapOverlay.report_code },
       entity,
       this.req,
     );
```
edmofro, line 371: We don't need to maintain consistency with the legacy way here - I think it'd be cleaner to just call into the report server builder directly (e.g. we're otherwise passing in two arguments that are unused, and I also can't see that the report server builder uses `dataElementCode` at all, etc.)

Something like the following might work, but needs testing:
```
    const builder = new ReportServerBuilder(
      this.req,
      this.models,
      { reportCode: mapOverlay.report_code },
      { ...this.query },
      entity,
    );
    return {
      [measureCode]: await builder.build(),
    };
```

edmofro (changes requested):
Looking good, this cleanup is going to bring me a lot of joy. I have left a few questions inline

packages/web-config-server/src/apiV1/measureData.js
EMcQ-BES: Usually I'm doing that because I'm protecting against the fact a new style report doesn't have a `measureBuilderConfig`, but this occurrence is within an `if (legacy)` block so we should expect that `measureBuilderConfig` _is_ defined at this point.

Possibly still worth the protection though since technically it's possible for the config to be null in the database

edmofro: No in that case totally leave it unprotected! I'm a fan of non-defensive programming in situations where you have a reasonable expectation, so that if that expectation (slash assumption) isn't met, things blow up. If you were to change anything, I'd make it an assertion that throws a nice error message rather than a defense against it being null. 

See `validateSurveyFields.js` for an example of succinct assertions of preconditions

edmofro:
Also some references to map overlay id in the tests still, which is why we've got a red x

packages/web-config-server/src/apiV1/measureData.js
EMcQ-BES: Oh good catch. This is because when `dataSourceType` is `single` the server will attempt to fetch metadata through the aggregator (Fetching from DHIS2 according to the comments) Since new style reports don't have a `measureBuilderConfig` they always take the default `dataSourceType` and error during that process, so I swapped it to `custom` which is more accurate for them.

This line is still wrong though because it will break the actual `single` type overlays which assume this default is in place so don't define `dataSourceType` in their configs. We could add a migration to add that `dataSourceType` to the configs of those and keep this defaulting to `custom`

Alternatively we could just make the default conditional `dataSourceType = legacy ? DATA_SOURCE_TYPES.SINGLE : DATA_SOURCE_TYPES.CUSTOM`

Let me know if you have any preference between those. I think the first solution is probably more correct.

edmofro: Ah cool. I actually think we should go for the second solution, because that maintains status quo for devs who are familiar with developing legacy dashboard reports, which will probably continue for some time until the report server completely catches up.

It also doesn't look like you've added the `CUSTOM` type to the definition?

EMcQ-BES: Oh wow I had just assumed it was already defined and ran with it. Good catch.

packages/web-config-server/src/apiV1/measureData.js
```diff
@@ -89,9 +91,10 @@ const buildMeasureData = (overlays, resultData) => {
       const { dataElementCode = 'value' } = measureBuilderConfig ?? {};
       const { data } = measureDataResponsesByMeasureCode[code];
 
+      // Rename the the value field from the dataElementCode to the measureCode
```
edmofro, line 94: Helpful :+1:

packages/web-config-server/src/apiV1/measureData.js
```diff
@@ -263,8 +267,10 @@ export default class extends DataAggregatingRouteHandler {
       ...restOfPresentationConfig
     } = config;
 
-    const { dataSourceType = DATA_SOURCE_TYPES.CUSTOM, periodGranularity } =
-      measureBuilderConfig || {};
+    const {
+      dataSourceType = legacy ? DATA_SOURCE_TYPES.SINGLE : DATA_SOURCE_TYPES.CUSTOM,
```
edmofro, line 271: You know I just looked into this some more, and I can't see it being used anywhere except the bit that fetches options from DHIS2 below. Let's just pop that into a `legacy === true` along with this destructuring, and then we can avoid introducing the semi-confusing `CUSTOM` type.

edmofro (changes requested):
All the changes look great, but I have one more update request

edmofro (approved):

PR #3339, opened 2021-10-05
RN-132: Fixed entity-server not returning any relationships when ancestorType = descendantType

rohan-bes:
### Issue RN-132:

Pretty straightforward bug, was occurring because `models.entity.fetchAncestorDetailsByDescendantCode()` doesn't return values that map to themselves (and it probably shouldn't)

### Changes:

- Example

---


edmofro (approved):
lgtm :-)

PR #3342, opened 2021-10-05
Wai 960 fix storybook

tcaiger:
### Changes:

- update storybook version and add support for nullish-coalescing-operator

---


yarn.lock
```diff
@@ -18,7 +18,7 @@
   optionalDependencies:
     chokidar "^2.1.8"
 
-"@babel/code-frame@7.10.4":
+"@babel/code-frame@7.10.4", "@babel/code-frame@^7.10.4":
```
edmofro, line 21: Could you please revert the changes to `yarn.lock`, then re-run `yarn install`? See https://beyondessential.slack.com/archives/CDGS6P56K/p1633403340044000 for more

edmofro (changes requested):
Everything looks good but I think the `yarn.lock` has lost my changes

yarn.lock
tcaiger: Woops I thought I did that but must have cleared it with my many yarn installs. I have reset and re built it and it's looking good now.

edmofro (approved):
Looks good!

PR #3344, opened 2021-10-06
RN-130: Support string concatenation in add operation in reportBuilder transform

chris-pollard:
### Issue #:
RN-130: Support string concatenation via '+' operator in viz-builder transforms
### Changes:
- Add new rule for two strings in the add typed function.
---


chris-pollard:
Closing this PR because another has been made with compliant branch name.

PR #3348, opened 2021-10-07
RN-122 Allow start/end time to be specified in POST survey response api

IgorNadj:


packages/meditrak-server/src/apiV2/surveyResponse.js
```diff
@@ -85,8 +92,8 @@ function buildResponseRecord(user, entitiesByCode, body) {
     user_id: user.id,
     entity_id: entityId || entitiesByCode[entityCode].id,
     data_time: stripTimezoneFromDate(time),
-    start_time: time,
-    end_time: time,
+    start_time: inputStartTime ? inputStartTime : time,
+    end_time: inputEndTime ? inputEndTime : time,
```
chris-bes, line 96: Ah I was going to make a comment on the default timestamp change for this 😄 .
Do we also want to convert `inputStartTime` and `inputEndTime` to ISO string (similar to `time`, line 87) to keep everything consistent?

IgorNadj: Good point, updated

chris-bes (approved):
LGTM!

PR #3353, opened 2021-10-08
RN-136: Fixed entity-server returning 403 when fetching for empty list of entities

rohan-bes:
### Changes:

- Example

---

### DIFF
--- a/packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
+++ b/packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
@@ -36,6 +36,11 @@ const validateEntitiesAndBuildContext = async (
     throw new Error(`Cannot find root entity for hierarchy: ${req.params.hierarchyName}`);
   }
 
+  if (entityCodes.length === 0) {
+    // No entities requested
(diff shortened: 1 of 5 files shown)


packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
```diff
@@ -36,6 +36,11 @@ const validateEntitiesAndBuildContext = async (
     throw new Error(`Cannot find root entity for hierarchy: ${req.params.hierarchyName}`);
   }
 
+  if (entityCodes.length === 0) {
+    // No entities requested
+    return { entities: [], allowedCountries: [] };
```
rohan-bes, line 41: Previously we were carrying on from here and latter doing a check to see if there were any entities found that we had permission to... If the list is empty we'd throw a 403.

I still think the 403 is appropriate if the user does request for entities even if the codes are incorrect, as it does seem to be unexpected behaviour, and we'd be leaking information about which entity codes exist in our database. However it seems fine and expected to return empty for an empty list of entity codes 👍

kael89 (approved):
Looks good!

PR #3354, opened 2021-10-08
WAI-61: lesmis survery responses view

tcaiger:
### Changes:

- Split Survey Responses table into 2 seperate tables for Review and Approved (using a test filter at the moment until the BE is complete)
- Add new Approve action
- Extend the delete action to be a Reject action

---

### Screenshots:
See issue


packages/admin-panel/src/importExport/SurveyResponsesExportModal.js
```diff
@@ -44,6 +44,7 @@ export const SurveyResponsesExportModal = () => {
         allowMultipleValues
       />
       <RadioGroup
+        name="survey responses mode"
```
tcaiger, line 47: fix prop error

packages/admin-panel/src/pages/resources/ResourcePage.js
```diff
@@ -84,6 +87,7 @@ ResourcePage.propTypes = {
 };
 
 ResourcePage.defaultProps = {
+  ConfirmDeleteModalComponent: undefined,
```
tcaiger, line 90: Needs to be undefined as opposed to null so that the DataFetchingModal component uses the default if nothing is passed in here.

packages/admin-panel/src/table/ExpansionContainer.js
```diff
@@ -24,14 +24,50 @@ export const ExpansionContainer = styled.div`
   .rt-tr {
     padding-right: 5px;
   }
+`;
 
-  &:after {
```
tcaiger, line 28: fix a z-index front end bug where the box shadow was blocking clicking of components under it

packages/lesmis-server/src/routes/ApproveSurveyResponse.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  constructor(req: Request, res: Response, next: NextFunction) {
+    super(req, res, next);
+
+    this.meditrakConnection = new MeditrakConnection(req.session);
+  }
+
+  async buildResponse() {
+    const { id } = this.req.params;
+    console.log('Approve Survey Response. Id: ', id);
```
tcaiger, line 22: This is a place holder routes for the BE work

packages/lesmis/src/views/SurveyResponsesView.js
```diff
@@ -0,0 +1,88 @@
...
+const ConfirmRejectModal = ({ isOpen, onConfirm, onCancel }) => {
+  return (
+    <ConfirmModal
+      onClose={onCancel}
+      isOpen={isOpen}
+      handleAction={onConfirm}
+      isLoading={false}
+      title="Reject Survey Response"
+      mainText="Are you sure you want to reject this Survey Response?"
+      description="Rejecting a Survey Response will also delete it. Once deleted this can’t be undone."
```
edmofro, line 72: Great helper message

packages/lesmis/src/views/SurveyResponsesView.js
```diff
@@ -0,0 +1,88 @@
...
+} from '@tupaia/admin-panel/lib';
+import { ConfirmModal } from '@tupaia/ui-components';
+import { ApproveButton } from '../components';
+
+// Todo: Replace base filter with real filter @see WAI-832
+export const ApprovedSurveyResponsesView = props => (
+  <ResourcePage
+    title="Approved Survey Responses"
+    baseFilter={{ 'survey.code': { comparator: 'ILIKE', comparisonValue: '%_Confirmed_WNR' } }}
+    {...props}
```
edmofro, line 18: Does this need columns passsed in?

packages/lesmis/src/components/ApproveButton.js
```diff
@@ -0,0 +1,78 @@
...
+  const [isApproved, setIsApproved] = useState(false);
+  const [showAlert, setShowAlert] = useState(false);
+  const { mutate, isLoading, isError, isSuccess } = useApproveSurveyResponse();
+
+  const handleClose = () => {
+    setShowAlert(false);
+  };
+
+  const handleClickAccept = async () => {
+    mutate(id, {
```
edmofro, line 31: I find the word `mutate` a little generic here, but is it the convention for this type of custom hook?

packages/lesmis/src/components/ApproveButton.js
```diff
@@ -0,0 +1,78 @@
...
+        setShowAlert(true);
+      },
+    });
+  };
+
+  return (
+    <>
+      {isApproved ? (
+        <FlexCenter>
+          <CheckIcon style={{ color: GREEN }} />
```
edmofro, line 46: Will the approved state ever show up in the page that has the approve button? I would have thought it might just disappear as the data table refreshes the data and that survey response no longer matches the filter, though that could be an uncomfortable experience for the user

edmofro (approved):
Again no required changes, so approving. Nice work!

packages/lesmis/src/components/ApproveButton.js
tcaiger: It's the plain react-query function rather than anything I've written so it is a convention in the react-query context but it could also make sense to name it to a more local context. I don't mind either way.

packages/lesmis/src/components/ApproveButton.js
tcaiger: Yes good question. It does stay in the list until the page is refreshed or until you revisit the tab (no FE cache here). The approve button uses a lesmis query rather than dispatching a table action so the normal admin panel table refresh is not triggered. I did do a bit of playing around to see if I could get the refresh to work and it is possible in a couple of ways by either adding redux-connect to lesmis or by moving the ApproveButton component to AdminPanel and connecting to the store there. I thought that:
a) it's not great for maintenance as it adds lesmis concepts to admin panel or it adds redux to lesmis and
b) it wasn't necessarily the best user experience anyway

And so opted for some obvious visual feedback instead using the toast and green tick.

packages/lesmis/src/views/SurveyResponsesView.js
tcaiger: No it doesn't as `ResourcePage` here is actually the whole `SurveyResponsesPage` from admin-panel. The way I've named it `ResourcePage` is confusing though so I've changed it to simply be `SurveyResponsesPage`

PR #3361, opened 2021-10-11
hotfix issue with phx-105 for this weeks release

avaek:
Fixes up an issue that was found during regression testing where a change to permission group name also had to be addressed in the `dashboard_relation` table

Original issue: https://linear.app/bes/issue/PHX-105/change-name-of-community-health-senior-permission-group


rohan-bes (approved):
LGTM!

Raises a good issue though... We've eschewed using foreign keys and join tables in a fair few scenarios in order to make things easier and more human readable, however this does come at a cost of data consistency...

I think really what we need going forward is to rely more on the admin panel, and actually push our db back to using more of the best practices for data modelling (foreign keys and join tables, etc.)

Anyways I'm rambling 😝 Nice work 👍

PR #3362, opened 2021-10-11
Wai-61: Remove old users page from lesmis

tcaiger:
### Changes:

- Remove old users page from lesmis because now we are re-using the admin-panel tables.

---


packages/lesmis-server/src/routes/UsersRoute.ts
```diff
@@ -1,64 +0,0 @@
...
-    this.meditrakConnection = new MeditrakConnection(req.session);
-  }
-
-  async buildResponse() {
-    const userData = await this.meditrakConnection.getUserEntityPermissions({
-      pageSize: '10000', // retrieve all entries, as we don't handle pagination
-      columns: JSON.stringify(Object.keys(FIELDS)),
-      sort: JSON.stringify(['user.creation_date DESC']),
-      filter: JSON.stringify({
-        'entity.code': 'LA',
```
edmofro, line 54: Does the new users table within LESMIS filter to only Laos users at all?

packages/lesmis/src/components/DataGrid/DataGrid.js
```diff
@@ -1,178 +0,0 @@
...
- * Tupaia
- * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
- *
- */
-import React from 'react';
-import PropTypes from 'prop-types';
-import styled from 'styled-components';
-import { useTable, useFilters, useSortBy, usePagination } from 'react-table';
-import { Table, TableHead, TableRow, TableBody, TableSortLabel } from '@material-ui/core';
-import { TableCell, StyledTableRow } from '@tupaia/ui-components';
```
edmofro, line 11: Was there anything left in `ui-components` that becomes unused across all packages after this?

edmofro (approved):
So much deleting 🎉 

I have questions but they don't block the PR at all, so approved :-)

packages/lesmis-server/src/routes/UsersRoute.ts
tcaiger: No there isn't. That reminds me that there also isn't any filtering of any of the data such as survey responses either. I assumed that we would do that at the permissions level and I saw some refinement notes to that extent on the BE ticket but I was meaning to check.

edmofro: Yep for survey responses they'll only see those they have permission to, but that does include all responses to public surveys (e.g. Basic Clinic Data etc.) which will be a bit weird. @enunan how much effort should we put into this? Obviously once it's their own deployment in Laos it wouldn't have any of that data on it anyway (if we ever do that), but in the meantime, the stuff they see through the LESMIS admin panel will include any Tupaia data that they have permission to see

PR #3365, opened 2021-10-11
RN-102: Added support for fetching for additional organisation units in report-server

rohan-bes:
### Issue RN-102:

In order to implement RN-27, we want to be able to run a report that fetches data for the requested organisation unit and one other specific organisation unit. 

Piggy backed off the work @kael89 did recently to allow overriding the `startDate`/`endDate` parameters, here we're overriding the `organisationUnitCodes` parameter.

Went with a basic syntax as so:
```
{
  "fetch": {
    "dataElements": [...],
    "organisationUnits": ['$requested', 'TO'] // '$requested' refers to the original org units from the request
  }
}
```

This allows for completely overriding the organisation units, or for extending them by using `$requested`. 

In the future I feel it could be really cool to support query parameters here rather than just explicit codes, eg.
```
organisationUnits: [{ country_code: 'LA', type: 'school'}]
```

It would be pretty easy to implement since we are already going via entity-server to check permissions on the extra org units. However held off introducing it now as I feel it would be a tad over engineering 👍 

**NOTE:**
First 2 commits are just refactors, so if reviewer prefers to just focus on functionality changes, see 9b08d7b


packages/report-server/src/reportBuilder/query/buildOrganisationUnitParams.ts
```diff
@@ -0,0 +1,47 @@
...
+      codesToFetch.push(organisationUnit);
+    }
+  });
+
+  const fetchedEntities = await ctx.services.entity.getEntities(hierarchy, codesToFetch, {
+    fields: ['code', 'country_code', 'type'],
+  });
+
+  const fetchedCodesWithAccess = fetchedEntities
+    .filter(({ type }) => type !== 'project') // Can't fetch at project level due to permissions, not supported for now
```
rohan-bes, line 37: So this is a bit annoying. We strictly speaking don't support fetching data at the project level, as we have no way of enforcing permissions on it. Instead if such a request is passed to report-server atm, we map the project code to the underlying countries. I'm not the biggest fan of that behaviour and it can lead to some unexpected things, so I chose to not do that here... Perhaps it would be worth it tho... not sure...

packages/report-server/src/reportBuilder/query/buildPeriodParams.ts
```diff
@@ -0,0 +1,73 @@
+/**
```
rohan-bes, line 1: Moved out of `QueryBuilder` as I was worried that class was gonna get pretty busy

packages/report-server/src/reportBuilder/fetch/functions/dataElement.ts
```diff
@@ -15,6 +16,8 @@ type DataElementFetchParams = {
   aggregations?: Aggregation[];
 };
 
+const dataElementsValidator = yup.array().of(yup.string().required()).min(1).required();
```
edmofro, line 19: Cool to know about the `.min` feature! Yup's so great 😀

packages/report-server/src/reportBuilder/query/buildOrganisationUnitParams.ts
edmofro: I think it's best to just do an easy unsupported check like you've done rather than wrangling it to awkwardly work, and if we need it to work later then there's just a bunch of refactor in different places to do it properly

In this case I think the `country_code !== null` is doing enough work for us though. There's one case in my local db where it's a project with a country code set, and that's a Fiji specific project, so that seems to give us the opportunity to opt-in to fetching data against a project if it's just for one country

packages/report-server/src/__tests__/reportBuilder/query/QueryBuilder.test.ts
```diff
@@ -220,22 +259,109 @@ describe('QueryBuilder', () => {
...
+    });
+
+    describe('organisation unit specs in config', () => {
+      type TestParams = [
+        string,
+        { config: { fetch: Record<string, unknown> }; query: FetchReportQuery },
+        Record<string, unknown>,
+      ];
+
+      const testData: TestParams[] = [
```
edmofro, line 299: This is a silly thing because I know it'll work, but I'd like to see a case where none of the org units is a country

packages/report-server/src/reportBuilder/query/buildOrganisationUnitParams.ts
```diff
@@ -0,0 +1,47 @@
...
+    .filter(({ country_code }) => country_code !== null)
+    .filter(({ country_code }) =>
+      accessPolicy.allows(country_code as Exclude<typeof country_code, null>, permissionGroup),
+    )
```
edmofro, line 41: It might be worth pulling all of the entity permission checking into this file (out of FetchReportRoute.ts) so there is a single place for any changes in rules to be applied (i.e. checking permission for $requested entities here as well as the additional ones)

packages/report-server/src/__tests__/reportBuilder/query/QueryBuilder.test.ts
rohan-bes: Sure thing!

packages/report-server/src/reportBuilder/query/buildOrganisationUnitParams.ts
edmofro: Yeah sounds great :-)

edmofro (approved):
Thanks for updating, lgtm!

packages/report-server/src/__tests__/reportBuilder/testUtils/entityApiMock.ts
```diff
@@ -15,4 +18,38 @@ export const entityApiMock = (entities: Record<string, Record<string, string>[]>
     const { fields } = queryOptions;
     return fields ? foundEntities.map(e => pick(e, fields)) : foundEntities;
   },
+  getDescendantsOfEntities: async (
```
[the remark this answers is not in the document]
rohan-bes: Hehe thanks 😄 idk why but queue's are on of my fav things to write, takes me back to my uni days 😂  

I thought about moving it to a more central place so that it can be re-used. I think perhaps implementing a complete `MockTupaiaApi` class would be really good at some point so that when we make changes to any of the microservices we only need to update the mocks in one place.

I'll create a card for it but I think it's out of scope for this ticket? 👍

packages/report-server/src/routes/helpers.ts
```diff
@@ -3,54 +3,12 @@
...
-    ({ country_code: countryCode }) =>
-      countryCode !== null && accessPolicy.allows(countryCode, permissionGroupName),
-  );
-  if (accessibleOrgUnits.length === 0) {
-    throw new Error(
-      `No '${permissionGroupName}' permissions to any one of entities: ${foundOrgUnits.map(
-        orgUnit => orgUnit.code,
-      )}`,
-    );
+export const parseOrgUnitCodes = (organisationUnitCodes: string | string[] | undefined) => {
```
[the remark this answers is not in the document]
rohan-bes: Ahh good call, will do 👍

packages/report-server/src/__tests__/reportBuilder/testUtils/entityApiMock.ts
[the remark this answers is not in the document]
edmofro: Yeah I was thinking something similar, a card would be good thanks

PR #3374, opened 2021-10-13
phx-106 delete test accounts

avaek:
### Issue #:
- https://linear.app/bes/issue/PHX-106/delete-these-old-test-accounts

This migration deletes the three test accounts specified in that card (one of which is an `api_client`)


kael89 (approved):
LGTM!

PR #3378, opened 2021-10-14
Wai 934 Lesmis FE Internationalisation

tcaiger:
### Changes:
- Url structure for internationalisation
- Update links and routes to handle locale in url
- Add locale to network requests
- Add button to switch between locales

This does not include adding any actual front end translations. There will be another PR in future for that.

---


packages/lesmis/src/api/api.js
```diff
@@ -38,6 +37,25 @@ const getApiUrl = () => {
...
+
+  if (locale !== DEFAULT_LOCALE) {
+    requestOptions = {
+      ...options,
+      params: {
+        ...options.params,
+        locale,
+      },
+    };
+  }
```
edmofro, line 54: Seems like we might as well send the locale through to the backend even if it's the default, so that
- the front end default isn't assuming the back end default is the same, so we could change them independently
- we drop a little bit of code and a `let` here
```suggestion
  return { timeout, ...options, params: { ...options.params, locale } };
```

There's a risk this causes the back end to run through "translate to English" unnecessarily, but I think the back end should be set up to avoid that if its default locale is passed through (cc @EMcQ-BES)

packages/lesmis/src/api/mutations/useLogin.js
```diff
@@ -7,9 +7,11 @@ import { useMutation, useQueryClient } from 'react-query';
 import { useHistory } from 'react-router-dom';
 import { post } from '../api';
 import { useUser } from '../queries';
+import { useHomeUrl } from '../../utils/useHomeUrl';
 
 export const useLogin = () => {
-  const history = useHistory();
+  const { pushToHomeUrl } = useHomeUrl();
+  const { push, location } = useHistory();
```
edmofro, line 14: Good call destructuring here, makes the code below read more cleanly for sure

packages/lesmis/src/components/LocaleMenu.js
```diff
@@ -0,0 +1,154 @@
...
+
+export const LocaleMenu = () => {
+  const [anchorEl, setAnchorEl] = useState(null);
+  const history = useHistory();
+  const { locale, entityCode, view } = useUrlParams();
+  const { search } = useLocation();
+  const queryClient = useQueryClient();
+
+  const handleChange = (event, code) => {
+    const link = `${makeEntityLink(code, entityCode, view)}${search}`;
```
edmofro, line 96:
```suggestion
  const handleChange = (event, newLocale) => {
    const link = `${makeEntityLink(newLocale, entityCode, view)}${search}`;
```

packages/lesmis/src/utils/makeEntityLink.js
```diff
@@ -5,8 +5,9 @@
  */
 import { generatePath } from 'react-router-dom';
 
-export const makeEntityLink = (entityCode, view = 'dashboard') =>
-  generatePath('/:entityCode/:view', {
+export const makeEntityLink = (locale, entityCode, view = 'dashboard') =>
+  generatePath(`/:locale/:entityCode/:view`, {
```
edmofro, line 9: Is it possible to use `useUrlParams` within this function, and get the `locale` there, rather than requiring it to be passed through from every place that calls `makeEntityLink`? I'm guessing not or you would have done it! Probably that would require that `makeEntityLink` itself would need to be treated like a hook, and only called outside of loops and conditionals etc.

packages/lesmis/src/utils/useHomeUrl.js
```diff
@@ -0,0 +1,23 @@
...
+
+export const useHomeUrl = () => {
+  const { params } = useRouteMatch();
+  const { push } = useHistory();
+  const { locale } = params;
+  const homeUrl = `/${locale}`;
+
+  const isHomeUrl = path => path.replace(/\/$/, '') === homeUrl;
+
+  const pushToHomeUrl = push(homeUrl);
```
edmofro, line 16: This name makes me think we're pushing something onto the home url, rather than pushing the home url into history. What do you think about this
```suggestion
  const navigateToHomeUrl = push(homeUrl);
```

packages/lesmis/src/utils/makeEntityLink.js
edmofro: Oh yeah well that does make things bloody easy... let's just do that! I have a sneaking feeling we might come to regret it if react-router and window.location somehow come out of sync (even for a moment in the component lifecycle), but it's so easy we should just do it and fix it if that happens.

tcaiger:
closing as is duplicate

PR #3379, opened 2021-10-14
RN-147: Refactor 'fetch' config in viz-builder

rohan-bes:
### Issue RN-147:

Bit of a refactor here to support adding new parameters to the `fetch` section. Currently we spread the fetch values in the config as so:
```
{
  dataElements: ["BCD1"],
  dataGroups: ["HP1"],
  aggregations": ["SUM"],
  transform: ["keyValueByDataElement"]
}
```

The change here is to:
1. Nest the fetch parameters within a fetch object
2. Rename "aggregations" to "aggregate" to match the viz-builder UI

So now we get:
```
{
  fetch: {
    dataElements: ["BCD1"],
    dataGroups: ["HP1"]
  },
  aggregate": ["SUM"],
  transform: ["keyValueByDataElement"]
}
```

Also took the liberty to do some refactors in the `admin-panel-server` and to add some unit tests.

**NOTE:** This change will break existing viz exports from being importable. It should be fairly simple to manually fix up those json files, and we'll throw an error if someone tries to upload an old one.


packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationExtractor.ts
```diff
@@ -92,11 +90,22 @@ export class DashboardVisualisationExtractor<
   private vizToReport(previewMode?: PreviewMode): Record<keyof Report, unknown> {
     const { code, permissionGroup, data, presentation } = this.visualisation;
 
-    const fetch = extractFetch(data);
+    const { fetch: vizFetch, aggregate, transform } = data;
+    const validatedVizFetch = yup
+      .object()
+      .required('fetch is a required field')
+      .validateSync(vizFetch);
```
kael89, line 97: Could we instead update `baseVisualisationValidator` to make its validation more thorough?

packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationExtractor.ts
```diff
@@ -48,10 +47,10 @@ export class DashboardVisualisationExtractor<
     this.reportValidatorContext = context;
   };
 
-  public getDashboardVisualisationResource = (previewMode?: PreviewMode) => {
+  public getDashboardVisualisationResource = () => {
     // Resources (like the ones passed to meditrak-server for upsert) use snake_case keys
     const dashboardItem = this.getDashboardItem();
-    const report = this.getReport(previewMode);
+    const report = this.getReport(PreviewMode.PRESENTATION); // always fetch full report when building resource
```
kael89, line 53: Thanks for simplifying the usage of PreviewMode- it was a bit hard to follow!

packages/admin-panel-server/src/__tests__/viz-builder/dashboardVisualisation/DashboardVisualisationExtractor.test.ts
```diff
@@ -0,0 +1,417 @@
...
+
+    describe('getDraftReport', () => {
+      it('throws error if viz does not have data.fetch', () => {
+        expect(() =>
+          new DashboardVisualisationExtractor(
+            { code: 'viz', data: { transform: [] }, presentation: {} },
+            yup.object(),
+            draftReportValidator,
+          ).getReport(),
+        ).toThrow('fetch is a required field');
```
kael89, line 40: In the past I'd follow this format too and try to combine as many things as possible in the `expect` statement. I changed my approach after I read https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices#arranging-your-tests, and I think I prefer it that way now. I had a discussion with @edmofro that Tamanu devs also follow an explicit Arrange/Act/Assert structure in their tests.

Nothing to change here, of course 🙂 Our codebase is filled with this assertion style, just food for thought!

packages/admin-panel-server/src/__tests__/viz-builder/dashboardVisualisation/DashboardVisualisationExtractor.test.ts
```diff
@@ -0,0 +1,417 @@
...
+            fetch: {
+              dataElements: ['BCD1'],
+            },
+            transform: [],
+          },
+        });
+      });
+    });
+
+    describe('getLegacyReport', () => {
```
kael89, line 99: Camel case in `getLegacyReport` slightly implies that this is a function (or a property) while this is not the case. Another approach would be using something like the following:

```suggestion
    describe('getReport() - legacy report', () => {
```

I like using `functionName()` to denote that the scope of a test block is this specific function. However, this is a personal preference and not something we have decided as a team

kael89 (approved):
Excellent work @rohan-bes , thank you for the clean up! I ❤️ those unit tests, we should probably get into the habit of adding integration/unit tests as soon as a server package is established! 👏

packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationExtractor.ts
rohan-bes: Unfortunately this isn't quite true anymore now that this same class handles legacy vizes which don't have a fetch step...

We could define `baseVizReportValidator` and `baseLegacyVizReportValidator` though? And validate those at the start of `vizToReport()` and `vizToLegacyReport()` respectively?

packages/admin-panel-server/src/__tests__/viz-builder/dashboardVisualisation/DashboardVisualisationExtractor.test.ts
rohan-bes: Oh wow, I'd never heard of this but it's actually such a good suggestion! Thanks for raising it 😄 I'm happy to change these, and will try to remember this rule in the future 👍

packages/admin-panel-server/src/viz-builder/dashboardVisualisation/DashboardVisualisationExtractor.ts
kael89: Oh I see, good point! We have validation in a lot of places so I wouldn't worry too much, happy for either what you have here or your above approach

PR #3382, opened 2021-10-15
PHX-113 request mapoverlay separately

biaoli0:
### Issue #:
PHX-113

### Changes:
- Request two sets of map overlay data separately.
- [Map state formate changes.](https://linear.app/bes/issue/PHX-113#comment-dc05784c)

### DIFF
--- a/packages/lesmis/src/api/api.js
+++ b/packages/lesmis/src/api/api.js
@@ -9,13 +9,13 @@ import FetchError from './fetchError';
 const timeout = 45 * 1000; // 45 seconds
 
 const getApiUrl = () => {
-  // if no env var, use sensible defaults based on the front end url
-  const { hostname } = window.location; // eslint-disable-line no-undef
-
   const { REACT_APP_LESMIS_API_URL } = process.env;
   if (REACT_APP_LESMIS_API_URL) {
     return REACT_APP_LESMIS_API_URL;
   }
+
+  // if no env var, use sensible defaults based on the front end url
+  const { hostname } = window.location; // eslint-disable-line no-undef
 
   // localhost becomes http://localhost:8060
   if (hostname === 'localhost') {


packages/ui-components/src/components/Map/Legend/Legend.js
```diff
@@ -56,40 +56,48 @@ const getLegendComponent = measureType => {
...
-    serieses,
     className,
+    measureInfo: baseMeasureInfo,
     setValueHidden,
     hiddenValues,
-    currentMeasureIds,
-    displayedMeasureIds,
+    currentMapOverlayIds,
+    displayedMapOverlayIds,
+    seriesesKey = 'serieses',
```
billli0, line 65: lesmis changes and uses `serieses` while tupaia uses `measureOptions` from backend response. We can do a clean up in tupaia to get rid of `measureOptions`.

packages/web-frontend/src/containers/Map/ConnectedPolygon.js
```diff
@@ -46,29 +43,19 @@ Polygon.defaultProps = {
...
+  const { displayedMapOverlays } = state.map;
+  const currentMapOverlayIds = selectCurrentMapOverlayIds(state);
+  const measureOptions = selectMeasureOptions(state, currentMapOverlayIds) || [];
   const measureOrgUnits = selectHasPolygonMeasure(state)
-    ? selectAllMeasuresWithDisplayInfo(state)
+    ? selectMeasuresWithDisplayInfo(state, displayedMapOverlays)
     : [];
-  const measureData = selectMeasureDataWithCoordinates(
-    selectRenderedMeasuresWithDisplayInfo(state),
-  );
```
billli0, line 65: measureData is not used in component

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -34,139 +32,143 @@ const displayInfoCache = createCachedSelector(
...
-export const selectCurrentMeasureIds = createSelector(
-  [selectCurrentMapOverlayIds, state => state.mapOverlayBar.mapOverlayHierarchy],
-  (currentMapOverlayIds, mapOverlayHierarchy) => {
-    const hierarchy = flattenMapOverlayHierarchy(mapOverlayHierarchy);
-
-    return hierarchy.map(mapOverlay =>
-      currentMapOverlayIds.includes(mapOverlay.mapOverlayId) ? mapOverlay.measureIds : [],
-    );
+export const selectMeasureOptions = createSelector(
+  [state => state.map.measureInfo, (_, mapOverlayIds) => mapOverlayIds],
```
billli0, line 47: selectMeasureOptions by either `displayedMapOverlayIds` or `currentMapOverlayIds`.

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -34,139 +32,143 @@ const displayInfoCache = createCachedSelector(
...
-const selectDisplayLevelAncestor = createSelector(
-  [
-    state => selectCountryHierarchy(state, selectCurrentOrgUnitCode(state)),
-    selectCurrentOrgUnitCode,
-    selectMeasureOptionsByDisplayedMapOverlays,
-  ],
-  (country, currentOrganisationUnitCode, measureOptions) => {
-    if (!country || !currentOrganisationUnitCode || !measureOptions) {
-      return undefined;
-    }
```
billli0, line 95: Move it down for a clearer flow.

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -34,139 +32,143 @@ const displayInfoCache = createCachedSelector(
...
-    selectMeasureOptionsByDisplayedMapOverlays,
-    state => state.map.measureInfo.hiddenMeasures,
-    state => state.map.measureInfo.currentCountry,
-    state => state.map.measureInfo.measureLevel,
+    selectActiveProjectCountries,
+    selectCurrentProjectCode,
+    state => selectCountryHierarchy(state, state.map.currentCountry),
+    (state, mapOverlayIds) => selectMeasureData(state, mapOverlayIds),
+    (state, mapOverlayIds) => selectMeasureOptions(state, mapOverlayIds),
+    (state, mapOverlayIds) => selectMeasureLevel(state, mapOverlayIds),
```
billli0, line 125: select by either `displayedMapOverlayIds` or `currentMapOverlayIds`.

packages/web-frontend/src/utils/measures.js
```diff
@@ -204,6 +205,16 @@ export function processMeasureInfo(response) {
...
+  // remove measure units with no coordinates
+  // for circle heatmap remove empty values or values that are not of positive float type
+  if (displayType === MARKER_TYPES.CIRCLE_HEATMAP) {
+    measureData = measureData.filter(({ value }) => {
+      if (!value || value === '') return false;
+      const parsedValue = parseFloat(value);
+      return !Number.isNaN(parsedValue) && parsedValue >= 0;
+    });
+  }
+
```
billli0, line 217: This was from `mapReducer.js`, moving to `processMeasureInfo` makes more sense to me

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -34,139 +32,143 @@ const displayInfoCache = createCachedSelector(
...
+  (measureInfo = {}, mapOverlayIds) => {
+    const selectedMeasureOptions = mapOverlayIds.reduce((results, mapOverlayId) => {
+      const { measureOptions = [] } = measureInfo[mapOverlayId] || {};
+      return [...results, ...measureOptions];
+    }, []);
+
+    const measureOptionsByKey = selectedMeasureOptions.reduce((results, measureOption) => {
+      const { key } = measureOption;
+      return results[key] ? results : { ...results, [key]: measureOption };
+    }, {});
```
billli0, line 57: Don't override existing measureOption if same key.
e.g.: two map overlays have same measure, we don't want to show the same legend in map and same item in flippa table.

packages/web-frontend/src/containers/MapOverlayBar/MapOverlayBar.js
```diff
@@ -49,53 +49,60 @@ export class MapOverlayBarComponent extends Component {
...
+    }
+
+    const isSelected = currentMapOverlayIds.includes(defaultMapOverlay.mapOverlayId);
+
     return (
       <HierarchyItem
         nestedMargin="0px"
         label={defaultMapOverlay.name}
-        isSelected={currentMapOverlayIds.includes(defaultMapOverlay.mapOverlayId)}
-        onClick={() => this.handleSelectMapOverlay(defaultMapOverlay)}
```
billli0, line 70: It was broken before 😅

packages/ui-components/src/components/Map/Legend/Legend.js
billli0: Sure: https://linear.app/bes/issue/PHX-120/change-measureoptions-to-serieses

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -34,139 +32,143 @@ const displayInfoCache = createCachedSelector(
...
-    return displayedMeasureOptions.length > 0 ? displayedMeasureOptions : undefined;
+    const measureData = {};
+    mapOverlayIds.forEach(mapOverlayId => {
+      const selectedMeasureData = measureInfo[mapOverlayId]?.measureData;
+      if (selectedMeasureData) {
+        selectedMeasureData.forEach(measure => {
+          measureData[measure.organisationUnitCode] = {
+            ...measureData[measure.organisationUnitCode],
+            ...measure,
+          };
```
kael89, line 79: `...` creates a new reference/value for each spread field, which can slow down things significantly when done a few thousand times (I've seen this happening in the past).

A way to avoid this could be something like the following:
```suggestion
          measureData[measure.organisationUnitCode] = Object.assign(
            measureData[measure.organisationUnitCode] || {},
            measure,
          );
```

`...` is nice in most cases, where
* we care about immutability
* we don't care about performance

In this specific case
* we do care about performance, since each overlay can have lots of data points
* we don't care as much about immutability in `measureData`: we "own" it (it was created inside the function), and the local scope of the variable is fairly limited

kael89: Based on the above, it would be good to confirm that `...` doesn't degrade performance here. To do so, we can select 2 overlays which both have lots of data (eg in Laos Schools) and take some measurements with `...` and with `Object.assign`.

 I'm happy for this to stay as is if there is no noticeable difference from a user perspective. `...` is commonly in modern JavaScript, and it reads nicely!

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -34,139 +32,143 @@ const displayInfoCache = createCachedSelector(
...
     if (!currentCountry || !measureData || !country) {
       return [];
     }
 
-    const listOfMeasureLevels = measureLevel.split(',');
-    let allOrgUnitsOfLevel = selectAllOrgUnitsInCountry(country).filter(orgUnit => {
-      return listOfMeasureLevels.includes(orgUnit.type);
-    });
-    if (currentCountry === projectCode) allOrgUnitsOfLevel = projectCountries;
+    const allOrgUnitsOfLevel =
```
kael89, line 143: Nice refactoring turning this into a `const`!

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -34,139 +32,143 @@ const displayInfoCache = createCachedSelector(
...
-    state => selectCountryHierarchy(state, state.map.measureInfo.currentCountry),
-    state => state.map.measureInfo.measureData,
-    selectMeasureOptionsByDisplayedMapOverlays,
-    state => state.map.measureInfo.hiddenMeasures,
-    state => state.map.measureInfo.currentCountry,
-    state => state.map.measureInfo.measureLevel,
+    selectActiveProjectCountries,
+    selectCurrentProjectCode,
+    state => selectCountryHierarchy(state, state.map.currentCountry),
+    (state, mapOverlayIds) => selectMeasureData(state, mapOverlayIds),
```
kael89, line 123: Btw no need to change it, but this should also work:

```suggestion
    selectMeasureData,
```

All arguments will be passed to the callbacks by JS

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -177,26 +179,47 @@ const selectAllMeasuresWithDisplayAndOrgUnitData = createSelector(
...
+      return undefined;
+    }
+
+    return selectAncestors(country, currentOrganisationUnitCode, displayOnLevel).find(
+      ancestor => ancestor.type === displayOnLevel,
+    );
+  },
+);
+
+export const selectRenderedDMeasuresWithDisplayInfo = createSelector(
```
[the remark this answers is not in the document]
billli0: Nice catch!

packages/lesmis/src/api/api.js
```diff
@@ -12,6 +12,11 @@ const getApiUrl = () => {
   // if no env var, use sensible defaults based on the front end url
   const { hostname } = window.location; // eslint-disable-line no-undef
 
+  const { REACT_APP_LESMIS_API_URL } = process.env;
+  if (REACT_APP_LESMIS_API_URL) {
+    return REACT_APP_LESMIS_API_URL;
+  }
```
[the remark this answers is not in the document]
billli0: It is very helpful to me because I'm coding in mac using a linux home server (everything is in my local network so it is secure). I also saw `REACT_APP_LESMIS_API_URL` in `.env.sample`. Hopefully it doesn't break anything?

packages/web-frontend/src/selectors/measureSelectors.js
billli0: Nice call Kostas! I've compared the difference and it seems they both rendered fast. But I would love to use Object.assign in case `...` will cost more time rendering in the future.

packages/web-frontend/src/actions.js
```diff
@@ -746,6 +747,15 @@ export function fetchMeasureInfoSuccess(response, countryCode) {
   };
 }
 
+/**
+ * All measure data have been arrived, changes isMeasureLoading to TRUE.
```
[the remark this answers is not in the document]
billli0: This comment is removed.

packages/lesmis/src/api/api.js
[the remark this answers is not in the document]
edmofro: Actually it used to be this way, I'm not sure when it changed. Happy to have this back in place @billli0, but could you please move it above the comment that refers to "if no env var" which used to live after that check? That way the comment still makes sense too. Thanks!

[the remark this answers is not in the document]
billli0: Make sense. 👍

PR #3390, opened 2021-10-18
RN-189 empty survey resp

IgorNadj:
RN-189

This change means that we can now push "deletes" of answers by submitting an empty survey response.

This change brings the API endpoint POST /surveyResponse in line with the behaviour of the importer. 

This only affects periodic surveys. For periodic surveys, the newest (non outdated) survey response informs the data that ends up in the analytics table. By pushing an empty survey response, we are in effect saying there is no data for that entity-day combination.

The specific use case this is for, is for a subsequent data push correcting a previously incorrect push where data for a COVID Samoa village was given, but after further data cleanup, it should not have been, so a subsequent push can mark it as No Data.

### DIFF
--- a/packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
+++ b/packages/meditrak-server/src/apiV2/export/exportSurveyResponses/exportResponsesToFile.js
@@ -250,13 +250,6 @@ export async function exportResponsesToFile(
         exportData[exportRow][exportColumn] = answer?.text || '';
       });
 
-    // If there is no data, add a message at the top
-    if (answers.length === 0) {
-      exportData.unshift([
-        `No data for ${survey.name} ${getExportDatesString(startDate, endDate)}`,
-      ]);
-    }
-
     if (easyReadingMode) {
       return addExportedDateAndOriginAtTheSheetBottom(exportData, timeZone);
     }


packages/meditrak-server/src/tests/apiV2/surveyResponse.test.js
```diff
@@ -225,7 +226,7 @@ describe('surveyResponse endpoint', () => {
     expect(body.results[1].answerIds).to.have.length(2);
   });
 
-  it('Should throw if a submission has an invalid survey code', async () => {
+  it('Should throw if a submission has an invalid survey id', async () => {
```
IgorNadj, line 229: Minor unrelated cleanup

packages/meditrak-server/src/tests/apiV2/surveyResponse.test.js
```diff
@@ -315,19 +316,6 @@ describe('surveyResponse endpoint', () => {
     expectError(response, /Could not find question/);
   });
 
-  it('Should throw if a survey has no answers', async () => {
```
IgorNadj, line 318: Moved into a describe block below

packages/meditrak-server/src/apiV2/surveyResponse.js
```diff
@@ -37,14 +37,16 @@ async function validateResponse(models, userId, body) {
...
+
   if (!body.entity_id && !body.entity_code) {
     throw new Error('Must provide one of entity_id or entity_code');
   }
 
   const surveyQuestions = await findQuestionsInSurvey(models, body.survey_id);
 
   const { answers } = body;
-  if (Object.keys(answers).length === 0) {
+  if (survey.can_repeat && Object.keys(answers).length === 0) {
```
rohan-bes, line 49: Honestly I do still feel this is acceptable even if the survey isn't periodic? @edmofro mind if I ask for your opinion on this?

The way I see it, it's a valid survey response even if there's no answers with it. It mightn't mean much, but it doesn't seem invalid..?

packages/meditrak-server/src/tests/apiV2/surveyResponse.test.js
```diff
@@ -441,6 +429,57 @@ describe('surveyResponse endpoint', () => {
...
+      const response = await app.post('surveyResponse', {
+        body: {
+          survey_id: surveyId,
+          entity_id: ENTITY_ID,
+          timestamp: 123,
+          answers: {},
+        },
+      });
+
+      expectError(response, /must contain at least one answer/);
```
rohan-bes, line 443: Looks like this test is failing for some reason?

IgorNadj: It was some test cross-contamination due to reusing the same test survey code, fixed

packages/meditrak-server/src/apiV2/surveyResponse.js
IgorNadj: Removed the answers validation altogether, making it allowed for all types of surveys.

IgorNadj:
@rohan-bes added 949bbe2 to fix an issue Sima found, happy with that?

rohan-bes:
Looks good to me code wise 👍 Trying to think if this change might have any side effects but I think it should be fine 😄

PR #3400, opened 2021-10-20
PHX-111 show data for all checked ols

billli0:
### Issue #:
PHX-111

### Changes:
1. Delete `ConnectedPolygon` and call `InteractivePolygon` directly in `Map.js`
2. In `Map.js`, we pass `measureData`, `measureOptions/serieses`, `allMeasureData` and `allMeasureOptions/Allserieses` to `InteractivePolygon` and `MarkerLayer`
3. Apply changes in storybook


tcaiger (changes requested):
Great work Biao. The multi-mapoverlays feature is very complex and the changes are looking good. As far as I can tell everything looks good to me from a logical point of view. I just have a couple of suggestions about naming to improve readability.

1. Can we use something like `multiOverlaySerieses` and `multiOverlayMeasureData`  instead of `allSerieses` and `allMeasureData`? It is not that clear to me that "all" refers to the multi-mapoverlays
2. Can we replace `measureOptions` with `serieses` everywhere you have used it. We are trying to phase our measureOptions. Sorry this suggestion is annoying since there are already measureOptions in some places.

billli0:
Thanks @tcaiger ! Yeah `all` is not clear I agree. I am happy to rename `allSerieses` and `allMeasureData` to `multiOverlaySerieses` and `multiOverlayMeasureData` 

We also agree to replace `measureOptions` to `serieses` across the tupaia codebase after this feature is merged. https://linear.app/bes/issue/PHX-121/change-measureoptions-to-serieses

PR #3401, opened 2021-10-20
WAI-993 LESMIS BackEnd Internationalisation

EMcQ-BES:
### Changes:

- Make translationKeys work as a minimised json schema
- Configure i18n to me attached to individual requests
- Add translations to more routes
- Move specific location of locale files


packages/lesmis-server/locales/en.json
```diff
@@ -0,0 +1,197 @@
+{
```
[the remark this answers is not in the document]
edmofro: Cool - let's just empty this I reckon, it seems like it'll inevitably require maintaining as text changes so better to have one less spot

packages/lesmis-server/locales/lo.json
```diff
@@ -0,0 +1,198 @@
...
+		"Khongxedone District": "ຄົງເຊໂດນ",
+		"Xaysetha District": "ໄຊເສດຖາ",
+		"Khop District": "ຄອບ",
+		"May District": "ໃໝ່",
+		"Sisattanak District": "ສີສັດຕະນາກ",
+		"Keo oudom District": "ແກ້ວອຸດົມ",
+		"Khamkeuth District": "ຄຳເກີດ",
+		"Naxaithong District": "ນາຊາຍທອງ",
+		"Xienghone District": "ຊຽງຮ່ອນ",
+		"Mad District": "Mad District",
```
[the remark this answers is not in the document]
edmofro: Thanks for doing the matching! @enunan could you ask them for translations of those two districts please?

packages/lesmis-server/src/routes/TranslatableRoute.ts
```diff
@@ -0,0 +1,66 @@
...
+export class TranslatableRoute extends Route {
+  translationSubGroup: string = '';
+  translationKeys: TranslationKey = { type: 'string' };
+
+  respond(responseBody: TranslationValue, statusCode: number) {
+    const translatedResponse = this.translateResponse(this.translationKeys, responseBody);
+    super.respond(translatedResponse, statusCode);
+  }
+
+  // Overload this function so that translationKey implies the shape of translationValue
```
edmofro, line 40: Nice to see the eventual solution to this problem!

packages/lesmis-server/src/routes/TranslatableRoute.ts
```diff
@@ -0,0 +1,66 @@
...
+  translateResponse(translationKey: ObjectKey, translationValue: { [key: string]: TranslationValue }): TranslationValue;
+  translateResponse(translationKey: ArrayKey, translationValue: TranslationValue[]): TranslationValue;
+  translateResponse(translationKey: TranslationKey, translationValue: TranslationValue): TranslationValue;
+  translateResponse(translationKey: any, translationValue: any): any {
+    if (!translationValue) {
+      return undefined;
+    }
+    switch(translationKey.type) {
+      case 'string':
+        return this.res.__(`${this.translationSubGroup}.${translationValue}:${translationValue}`);
```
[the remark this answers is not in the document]
edmofro: That's probably a handy comment tbh, it would be hard to pick up without going to the docs.

An unrelated point: I'm not a fan of the `__` mystery function

Can we please use the `api` config option? From the docs:
```
// hash to specify different aliases for i18n's internal methods to apply on the request/response objects (method -> alias).
  // note that this will *not* overwrite existing properties with the same name
  api: {
    __: 't', // now req.__ becomes req.t
    __n: 'tn' // and req.__n can be called as req.tn
  },
```

packages/lesmis-server/src/routes/TranslatableRoute.ts
```diff
@@ -0,0 +1,66 @@
+/**
```
edmofro, line 1: Does anything about this file make it specific to LESMIS? I'd be keen to stick it in `server-boilerplate`, I'm sure we'll need it there at some point so might as well avoid the risk of coupling it too tightly to lesmis-server in the meantime

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -33,6 +35,8 @@ export class ApiBuilder {
 
   private verifyAuthMiddleware?: Middleware;
 
+  private translatorConfigured: boolean = false;
```
edmofro, line 38: This is very subjective so entirely optional, but to me initialised just speaks more to the idea that it happens once on first use
```suggestion
  private translatorInitialised: boolean = false;
```

packages/lesmis-server/src/routes/TranslatableRoute.ts
```diff
@@ -0,0 +1,66 @@
...
+  type: 'array';
+  items: TranslationKey;
+  where?: (entry: any) => boolean;
+}
+
+type TranslationKey = StringKey | ObjectKey | ArrayKey;
+type TranslationValue = undefined | string | TranslationValue[] | { [key: string]: TranslationValue };
+
+export class TranslatableRoute extends Route {
+  translationSubGroup: string = '';
```
edmofro, line 32: This is clear to me now that I have read this file, but wasn't at all clear when I was reading the route implementations. First impression was that it might be used where a route class responds to multiple api paths (e.g. the collection and the individual record), to specify which of those paths ran through the translation. I'm sitting here trying to think up a name or a different place of putting it that might avoid that, but not coming up with much. Could you make some suggestions? My only thought so far is that the `translationSchema` has a top level field `domain` or `scope`, so it'd look like this for the dashboard route:

```
    this.translationSchema = {
      domain: 'dashboards',
      type: 'array',
      items: {
        type: 'object',
        properties: {
          dashboardName: {
            type: 'string'
          },
        }
      }
    };
```

edmofro (changes requested):
Brilliant stuff, this is gonna be awesome and I'm still impressed with how little weight you've added to the codebase to implement such a fundamental feature 🎉 

Some questions and suggestions inline, happy to talk through any if you like

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
EMcQ-BES: Usually I would agree with this particular language. I used "configured" here to match the actual functions we're calling against the i18n library. We call `i18n.configure` once, but `i18n.init` against each req.

edmofro:
Pretty sure we can close this now @EMcQ-BES

PR #3407, opened 2021-10-21
Wai 934 lesmis fe translations

tcaiger:
### Changes:

- Add I18n component
- Add useI18n hook for contexts where a jsx component can't be used (eg. form inputs)
- Add first batch of translations (home page and main site menu)

---

### Screenshots:
See issue


EMcQ-BES (approved):
Exciting stuff. I like how clean the code for the contents pages look without the content directly in them

PR #3409, opened 2021-10-22
RN-117 Allow dev-only env config

IgorNadj:
RN-117


edmofro (approved):
Works for me, nice one :-)

PR #3430, opened 2021-10-27
Phx 122 fix survey export condition arithmetic

Alastair-L:
### Changes:
-Added config builders for exporting condition and arithmetic question types
-Added tests for this functionality
-Allow codes to be used more than once in an arithmetic formula
-Renamed `getExpressionQuestionCodes` to `getDollarPrefixedExpressionVariables` because it made more sense to me (as it's now also used to get expression question **Ids**), but happy to revert that


Alastair-L:
Thanks for the review @rohan-bes!
1. Oops! Sorry about that, deleted now
2. I think you are absolutely right, and thanks for offering to review it, but I'm not sure I can justify the time unless we plan to add many new question types.

PR #3440, opened 2021-10-28
MAUI-117: Add 'insertNumberOfFacilitiesColumn' alias transform

rohan-bes:
### Issue MAUI-117:

Fairly short term fix here to add a 'insertNumberOfFacilitiesColumn' alias transform. If this proves too un-flexible in the short term we can easily switch it out for a context function, however I prefer to alias transform for now since they're a bit easier to add/remove.

I feel the longer term fix for these kinds of requests would be to support pulling data from the entity-server itself...

### Changes:
- Bit of a refactor to the naming conventions in report-server. While I still think it can do with some work, we're now referring to the functions/alias transforms that require context data as `contextFunctions` and `contextAliases`. These in turn have `dependencies` which are a list of dependencies that the context can provide (`orgUnits`, `dataElementCodeToName`, `facilityCountByOrgUnit`, etc.)


IgorNadj (approved):
LGTM!

PR #3447, opened 2021-11-01
wai-1020 export code instead of id for entities

avaek:
Issue: https://linear.app/bes/issue/WAI-1020/export-answers-to-entity-questions-as-codes-rather-than-ids

This exports Entity answers as the Entity code instead of the Entity id


edmofro (approved):
Wow this is absolutely great stuff!

edmofro:
@avaek can you merge this and test it on the release branch plx?

PR #3450, opened 2021-11-02
WAI-963, Part 2: Connect new CI/CD flow to codeship

edmofro:
### Issue #:
WAI-963

### Changes:
- Codeship no longer handles any of the building and pushing, instead just triggers the same lambda function used by us to deploy feature branches, which will deploy a new instance on top of the old one, and then swap the gateway over to the new one and terminate the old


packages/devops/scripts/ci/triggerRedeploy.sh
```diff
@@ -0,0 +1,58 @@
...
+      else
+        echo "Still waiting for instance startup build to complete. To watch detailed progress, connect to instance ${NEW_INSTANCE_ID} and run tail -f logs/deployment_log.txt"
+        WAIT_ATTEMPTS=$((WAIT_ATTEMPTS+1))
+      fi
+    fi
+  done
+
+else
+  echo "No deployment running, skipping redeploy"
+fi
```
rohan-bes, line 58: Not major at all, but might be nice to do exit early here with:
```
echo "No deployment running, skipping redeploy"
exit 0
```
?

packages/devops/scripts/lambda/actions/redeploy_tupaia_server.py
```diff
@@ -35,7 +35,7 @@
 
 def redeploy_tupaia_server(event):
     instance_filters = [
-        { 'Name': 'tag:SubdomainsViaDns', 'Values': ['ssh'] }, # the main server rather than db instance
+        { 'Name': 'tag:SubdomainsViaDns', 'Values': ['ssh'] }, # the main server rather than db instance (can be removed after RN-195)
```
rohan-bes, line 38: Ah thanks for marking this here 👍

packages/devops/scripts/ci/copyInEnvironmentVariables.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -e
+
+# So long as setupEnvironmentVariables has run, the common volume "environment_variables"
+# will have all .env files in the same directory structure as they're expected within the
+# tupaia repo, so we just rsync the whole volume to copy the .env files in
+
+rsync -a /environment_variables/ /tupaia/
```
rohan-bes, line 7: I never knew about `rsync`! Seems really cool 🎉

packages/devops/scripts/ci/triggerRedeploy.sh
```diff
@@ -0,0 +1,58 @@
...
+    cat $RESPONSE_FILE
+    exit 1
+  fi
+
+  NEW_INSTANCE_ID=$(cat $RESPONSE_FILE | cut -d "\"" -f 2)
+
+  echo "Waiting for instance to run its startup build script. To watch detailed progress, connect to instance ${NEW_INSTANCE_ID} and run tail -f logs/deployment_log.txt"
+  WAIT_ATTEMPTS=0
+  while true; do
+    aws ec2 wait instance-exists \
```
rohan-bes, line 32: Not 100% sure if I'm reading the correct docs, but it seems to suggest that this operation polls by itself, and will return after 40 attempts (5 seconds wait between attempts): https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ec2/wait/instance-exists.html

Would that mean we don't need our own polling logic here?

rohan-bes (approved):
Approving since it's all looking good and my suggestion is probably just me being a bit confused about the correct use of the aws functions 👍 

So nice to see so much red 😄 Looking forward to these changes!

packages/devops/scripts/ci/triggerRedeploy.sh
edmofro: Yeah good idea, will do

packages/devops/scripts/ci/copyInEnvironmentVariables.sh
edmofro: It is! Made this quite straightforward

packages/devops/scripts/ci/triggerRedeploy.sh
edmofro: You're definitely reading the correct docs, but we need to wait even longer than those 40 attempts for the build to finish...meaning we have to wrap the polling logic in polling logic 😅

edmofro:
@rohan-bes I realised I hadn't accounted for multiple deployments of the same branch in this PR. Have now added support for that, along with the early return change

packages/devops/ci/deployment.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM amazon/aws-cli
+
+COPY ./packages/devops/scripts/ci/triggerRedeploy.sh .
+
+RUN curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 > jq
+RUN chmod +x jq
+RUN mv jq /bin
```
rohan-bes, line 7: This is by no means my area of expertise, but is it possible for us to use docker to install jq like we do for our other dependencies? eg. 
```
# install features not available in base alpine distro
RUN apk --no-cache add \
  bash \
  curl \
  git \
  lastpass-cli \
  openssh \
  postgresql-client \
  rsync \
  jq
```
This post suggests it's possible: https://stackoverflow.com/questions/56452639/how-to-use-httpie-and-jq-within-docker
and there's a docker image for it here: https://hub.docker.com/r/stedolan/jq

rohan-bes (approved):
It's looking good to me! Just one suggestion on how we could maybe use docker to install `jq` over doing a manual installation?

Also kinda curious to know the use case for multiple deployments per branch? Sounds handy!

packages/devops/ci/deployment.Dockerfile
edmofro: Yeah I would prefer to be using `apk add` too...unfortunately it seems the aws-cli base docker image strips that out of alpine or something, because `apk` just isn't available. 

Also worth noting that I don't think `apk add` uses docker images under the hood, I'm pretty sure it's more like homebrew on mac and pulls from a repository of binaries.

edmofro:
The main use case for multiple deployments per branch is `viz-sandbox`, `viz-testing`, and `e2e` - no more merge from master back in to those branches on release 🎉

PR #3451, opened 2021-11-02
[HOT FIX]: Sanitize entity answers before they are imported

edmofro:
Fixes an issue with import now that we're exporting entity questions as codes rather than ids


IgorNadj (approved):
LGTM! 

One small thing, the name. I usually think of sanitisers as like trim white space and stuff. This is more of an import data "transformer" or something like that. Happy with this though, names are hard.

edmofro:
Good call, I agree!

PR #3452, opened 2021-11-02
Release v2021-44-1

EMcQ-BES:
### Issue #:

A couple of hotfixes that need to be released outside of our usual schedule


packages/web-config-server/src/apiV1/dataBuilders/generic/dataDownload/rawDataValues.js
```diff
@@ -148,10 +147,6 @@ class RawDataValuesBuilder extends DataBuilder {
   sortEvents = async (events, sortKeys) => {
     if (!sortKeys) return events;
 
-    // If we are merging sorting on that key takes precedence
-    // for performance of merge and avoid ancestor lookup
-    if (sortKeys.mergeRowKey) return this.sortEventsByDataValue(events, sortKeys.mergeRowKey);
```
edmofro, line 153: We definitely need to regression test any viz that has a `mergeRowKey` and uses this data builder before releasing

edmofro (approved):
lgtm

PR #3453, opened 2021-11-02
Wai-875 lesmis dashboard search

tcaiger:
### Changes:

- Example

---

### Screenshots:
see issue


packages/lesmis/src/components/DashboardSearch/DashboardSearch.js
```diff
@@ -0,0 +1,199 @@
...
+    limit: LIMIT,
+    onChange: (event, option, reason) => {
+      if (reason === 'select-option' && option && option.code) {
+        resetSearch();
+      }
+    },
+    muiProps: {
+      debug: true, // Keeps the options in memory after blur event
+      onClose: (event, reason) => {
+        if (reason === 'escape') {
```
edmofro, line 139: As opposed to which other reasons for closing?

packages/lesmis/src/components/DashboardSearch/DashboardSearch.js
```diff
@@ -0,0 +1,199 @@
...
+  const [isActive, setIsActive] = useState(false);
+  const options = useDashboardItems();
+
+  const resetSearch = () => {
+    onToggleSearch(false);
+    setIsActive(false);
+    setInputValue('');
+  };
+
+  const searchResults = useAutocomplete({
```
edmofro, line 125: When seeing `searchResults` in later contexts I kept assuming it would be an array of the results, where is this is more of a complex object including prop getters etc. Perhaps we should call it something like `autocompleteResponse` to avoid that confusion

edmofro (approved):
This all looks great! I got a little confused about the handling of clear and exit states etc., but I think that's just because there are a few different flows that need to be taken care of, so don't have any suggestions there :-)

packages/lesmis/src/components/DashboardSearch/DashboardSearch.js
tcaiger: This is just the best way that I figured with material-ui to add an escape button event. The reason prop can be one of: "toggleInput", "escape", "select-option", "blur".

packages/lesmis/src/components/DashboardSearch/DashboardSearch.js
tcaiger: Yes, good suggestion!

PR #3458, opened 2021-11-03
Phx 130 loading component

billli0:
### Issue #:
PHX-130

### Changes:
Loading skeleton component

### Screenshots:
https://linear.app/bes/issue/PHX-130#comment-4c6c2f91

### DIFF
--- a/packages/web-frontend/src/containers/MapOverlayBar/Content.js
+++ b/packages/web-frontend/src/containers/MapOverlayBar/Content.js
@@ -24,7 +24,7 @@ export const EmptyContentText = styled(ContentText)`
 
 export const ExpandedContent = styled.div`
   pointer-events: auto;
-  background: ${MAP_OVERLAY_SELECTOR.subBackGround};
+  background: ${MAP_OVERLAY_SELECTOR.subBackground};
   color: #fff;
   border-bottom-left-radius: 5px;
   border-bottom-right-radius: 5px;
--- a/packages/web-frontend/src/containers/MapOverlayBar/Content.js
+++ b/packages/web-frontend/src/containers/MapOverlayBar/Content.js
@@ -6,14 +6,15 @@ export const Content = styled.div`
   justify-content: space-between;
   align-items: flex-start;
   pointer-events: auto;
-  margin: 21px 16px 20px 16px;
+  padding: 21px 16px 15px 16px;
   color: #ffffff;
 `;
 
 export const ContentText = styled.div`
   font-weight: 500;
   font-size: 16px;
-  line-height: 19px;
+  line-height: 21px;
+  padding-right: 15px;
 `;
 
 export const EmptyContentText = styled(ContentText)`
--- a/packages/web-frontend/src/containers/MapOverlayBar/Content.js
+++ b/packages/web-frontend/src/containers/MapOverlayBar/Content.js
@@ -6,7 +6,7 @@ export const Content = styled.div`
   justify-content: space-between;
   align-items: flex-start;
   pointer-events: auto;
-  padding: 21px 16px 15px 16px;
+  padding: 21px 16px 15px 18px;
   color: #ffffff;
 `;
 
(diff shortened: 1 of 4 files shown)


packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -17,6 +21,10 @@ const FlexSpaceBetween = styled(FlexSpaceBetweenCenter)`
   align-items: flex-start;
 `;
 
+const SkeletonHeader = styled(FlexCenter)`
```
tcaiger, line 24: You shouldn't need a custom component to add padding to this. FlexCenter extends [Material UI Box](https://v4.mui.com/components/box/#box) component and it is possible to use that to set the margin.
eg `<FlexCenter mt="10px">`

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -100,28 +117,42 @@ export const TitleAndDatePickerComponent = ({
...
-            startDate={startDate}
-            endDate={endDate}
-            min={datePickerLimits.startDate}
-            max={datePickerLimits.endDate}
-            onSetDates={updateMeasurePeriod}
-            isLoading={isMeasureLoading}
-          />
-        </MeasureDatePicker>
+      {isMeasureLoading ? (
+        <SkeletonHeader mt="10px">
```
tcaiger, line 121: The loading section appears to be a different height to the content section and so the side bar moves up and down when it starts and stops loading. I think adding 10px margin bottom will fix it.

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
```diff
@@ -54,6 +62,15 @@ const Pin = styled(PinBase)`
   margin: 3px 12px 0 7px;
 `;
 
+const CircularProgress = styled(MuiCircularProgress)`
+  color: #2196f3;
+`;
+
+const Skeleton = styled(MuiSkeleton)`
+  margin-left: ${({ ml = 0 }) => ml}px;
+  margin-top: ${({ mt = 0 }) => mt}px;
```
tcaiger, line 71:
```suggestion
  margin-top: ${({ mt = 0 }) => mt}px;
  background: none;
```
To make the fading effect from the design, you can turn off the background color on the skeleton.

tcaiger (changes requested):
Hey Biao, this looks cool with the skeleton loader. I just left a suggestion to match the height of the loader to the content but apart from that it looks good.

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
billli0: Nice, will remove `SkeletonHeader`.

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
billli0: Sounds good, these component have their own padding and it costs me some times to fix it. Is it ok to use 'min-height' to make them consistent?

packages/web-frontend/src/containers/MapOverlayBar/Content.js
```diff
@@ -6,14 +6,15 @@ export const Content = styled.div`
...
+  padding: 21px 16px 15px 16px;
   color: #ffffff;
 `;
 
 export const ContentText = styled.div`
   font-weight: 500;
   font-size: 16px;
-  line-height: 19px;
+  line-height: 21px;
+  padding-right: 15px;
```
billli0, line 17: As discussed with David, we want to add a padding between the overlay name and switch button, see screenshot: https://linear.app/bes/issue/PHX-130#comment-3097748d

packages/web-frontend/src/containers/MapOverlayBar/TitleAndDatePicker.js
tcaiger: yep no problem

PR #3462, opened 2021-11-05
RN-196: Few fixes for heavy memory usage when fetching /organisationUnits

rohan-bes:
### Issue RN-196:

This isn't an exhaustive investigation/fix by any means, but addresses a few low hanging fruit.

It seems that memory consumption when loading the /organisationUnits route in a project with a large hierarchy can be very high. During this process the memory usage of the web-config-server can spike to over 1GB. Initially the memory seemed to be leaking as it stayed high and grew on each subsequent request, but it turns out that was just the Node garbage collection not cleaning things up as yet. 

Overall to stop such issues in the future we probably need to tune our max memory limits for the various servers to something more safe given the total RAM of the feature instances is only 4GB.

Actual fixes here:
1. Pushed the filtering for unique entities in Entity.js inside the cached function. This avoid accidentally caching duplicate entities, as removes the need to perform this filter every time
2. Wrapped the the utility functions for getting parent/child relations on AncestorDescendantRelation.js inside cached functions. This saves us some work recalculating these every time
3. Made a few optimisations to the logic of the /organisationUnits route 

Still doing a bit of proper testing on a feature instance to guage just how much difference they make (seemed to make a 40% reduction in peak memory usage on my local), will post results once that's completed

### DIFF
--- a/packages/database/src/modelClasses/AncestorDescendantRelation.js
+++ b/packages/database/src/modelClasses/AncestorDescendantRelation.js
@@ -25,6 +25,10 @@ export class AncestorDescendantRelationType extends DatabaseType {
       fields: { code: 'ancestor_code' },
     },
   ];
+
+  get cacheDependencies() {
+    return [TYPES.ENTITY_RELATION, TYPES.ENTITY_HIERARCHY];
+  }
 }
 
 export class AncestorDescendantRelationModel extends DatabaseModel {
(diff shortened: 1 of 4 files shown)


packages/database/src/modelClasses/AncestorDescendantRelation.js
```diff
@@ -41,22 +45,34 @@ export class AncestorDescendantRelationModel extends DatabaseModel {
   }
 
   async getChildIdToParentId(hierarchyId) {
-    const relationRecords = await this.getImmediateRelations(hierarchyId);
-    return reduceToDictionary(relationRecords, 'descendant_id', 'ancestor_id');
+    const cacheKey = this.getCacheKey(this.getChildIdToParentId.name, hierarchyId);
+    return this.runCachedFunction(cacheKey, async () => {
+      const relationRecords = await this.getImmediateRelations(hierarchyId);
+      return reduceToDictionary(relationRecords, 'descendant_id', 'ancestor_id');
+    });
```
rohan-bes, line 52: Decided to cache these individual functions rather than the base `getImmediateRelations()` function since that function returns a pretty large payload and caching it all would greatly increase the server memory footprint

packages/entity-server/src/routes/hierarchy/format/formatEntityForResponse.ts
```diff
@@ -85,12 +85,13 @@ export async function formatEntitiesForResponse(
...
-      ? await models.ancestorDescendantRelation.getImmediateRelations(hierarchyId, {
-          'descendant.country_code': ctx.allowedCountries,
-        })
-      : [];
+  const childCodeToParentCode: Record<string, string> = fields.includes('parent_code')
+    ? await models.ancestorDescendantRelation.getChildCodeToParentCode(hierarchyId)
+    : {};
+  const parentCodeToChildCodes: Record<string, string[]> = fields.includes('child_codes')
+    ? await models.ancestorDescendantRelation.getParentCodeToChildCodes(hierarchyId)
+    : {};
```
rohan-bes, line 93: Just a bit of a cleanup here to re-use those utility functions 👍

packages/web-config-server/src/apiV1/organisationUnit.js
```diff
@@ -21,7 +21,7 @@ export default class extends RouteHandler {
 
     // If this is a project entity, get entitiesWithAccess and set them on the entity
     if (this.entity.isProject()) {
-      const orgUnits = await this.entity.getNearestOrgUnitDescendants();
+      const orgUnits = await this.entity.getChildren(project.entity_hierarchy_id);
```
rohan-bes, line 24: This **_is_** actually a functionality change here... The previous `getNearestOrgUnitDescendants()` returns the nearest descendants of the valid 'orgUnit' types. I might have to double check this but I don't think we actually have any project with non-orgUnit entities as the immediate children of a project.

Switching to `getChildren()` here greatly reduces the entityHierarchyCache size as otherwise we end up caching the entire project hierarchy...

packages/web-config-server/src/apiV1/organisationUnit.js
```diff
@@ -50,11 +55,8 @@ export default class extends RouteHandler {
           },
         })
       : [];
-    const orgUnitHierarchy = await this.filterForAccess([
-      projectEntity,
-      country,
-      ...countryDescendants,
-    ]);
```
rohan-bes, line 57: Removed this permissions check here, as for all entities other than the country entity we were immediately returning `true`. Instead we just check the country entity at the start and return an empty object if we don't have access to that country.

packages/database/src/modelClasses/Entity.js
```diff
@@ -443,11 +443,11 @@ export class EntityModel extends MaterializedViewLogDatabaseModel {
           sort: ['generational_distance ASC'],
         },
       );
-      return Promise.all(relations.map(async r => r.getData()));
+      const relationData = await Promise.all(relations.map(async r => r.getData()));
+      const uniqueEntities = Object.values(keyBy(relationData, 'id'));
```
edmofro, line 447: I realise this is not a new thing, but it seems a shame not to have guaranteed order preservation here when the db query has a sort in it

edmofro (approved):
lgtm!

packages/database/src/modelClasses/Entity.js
rohan-bes: Ahh that's a great point, hadn't thought of that... I think it should still preserve order though? I think lodash keyby will insert the keys as per the order they appear in the original array. I was testing with the entity-server then and it seems they still come out in generational distance order 😄

edmofro: Javascript objects don't guarantee any order, so it's possible we could see the order change unexpectedly, but in practice yeah we probably will end up with preserved order, just because that's the easiest thing for the lazy-bones computer to do 😛

rohan-bes: I think as of ES2015 the non integer like keys are kept in insertion order, and thus should consistently be correct? @kael89 loves to remind me of this as I have made similar statements recently too 😝

edmofro: Oh I totally remember @kael89 mentioning this to me too now! Thanks for tagging him in here to throw me under the bus haha (sorry for forgetting Kostas!!)

PR #3476, opened 2021-11-09
RN-85: Disable 'Save' and 'Export' button when viz config is invalid

rohan-bes:
### Issue RN-85:

The only change here is this commit: b819256

The other commit is just a file rename. (though feel free to give feedback on the rename 👍)

Created a new context to handle the error state. We have a similar context for `PreviewData` however that seems more related to the data pulled for the data table, and the scope of that context is a bit smaller so I felt creating a new one was appropriate.


edmofro (approved):
Really helpful little feature, and nice clean solve 🌟

PR #3477, opened 2021-11-09
Maui 125 insert summary row and column alias

chris-pollard:
### Issue #:
As part of the matrix visual, create transform functionality for summary row and column based on Y and N
### Changes:
New alias in summaryAlias.ts and test for insertSummaryRowAndColumn


packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+      }
+    })
+    return futureValue
+  },{})
+};
+
+const getSummaryRow = (numeratorRow: Row, denominatorRow: Row) => {
+  return Object.fromEntries(Object.entries(denominatorRow).map(([key,value]) => {
+    const numeratorValue = numeratorRow[key] as number;
```
rohan-bes, line 61: If we expect `numeratorRow` and `denominatorRow` to be a `Record<string, number>` type, feel free to define a type for this in this file 👍 Will save us needing to cast explicitly here which is a bit of a bad practice

rohan-bes (commented):
Halfway through reviewing, will finish up tomorrow morning 👍

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (typeof value === 'number') {
+      const percentage = `${(numeratorValue / value * 100).toFixed(1)}%`;
```
rohan-bes, line 63: Probably worth encapsulating this formatting in a function as we use it twice in this file. eg.
```
const formatSummary = (numerator: number, denominator: number) => `${(numerator / denominator * 100).toFixed(1)}%`;
```

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,77 @@
...
+    return [key, value]
+  }))
+}
+
+export const insertSummaryRowAndColumn = () => (rows: Row[]) => {
+  const rowsWithSummaryColumn = getRowsWithSummaryColumn(rows);
+  const numeratorRow = getFractionRow(rows,true);
+  const denominatorRow = getFractionRow(rows,false)
+  const summaryRow = getSummaryRow(numeratorRow,denominatorRow)
+  return [...rowsWithSummaryColumn,summaryRow]
```
rohan-bes, line 75: Apologies as I'm being nit picky here, and feel free to push back on this if you'd rather just push this through, however I felt you might appreciate my perspective here?

Tbh I find the logical flow of the algorithm here a bit unexpected. We're presented with a set of rows, which we plan on adding a summary column to each, and then finally adding a summary row to the table. Knowing which columns we are summarising seems to be a key concept here, but it's sort of handled implicitly in the if/else logic of `getRowsWithSummaryColumn()` and `getFractionRow()`.

Knowing which columns to summarise does feel a touch tricky, but a safeish bet seems to be all columns whose values are all either `'Y' | 'N' | undefined`? That's not technically what we have here, as we do it on a cell by cell basis which should largely be fine, however it could lead to small bugs in some circumstances (I recently did a matrix where we had a column whose value was the first letter of the medicine, which could be `'Y'` or `'N'`).

I think the way I intuit this algorithm would be based on how I might do this process manually, and goes a bit like:
```
const columnsToSummarise = detectColumnsToSummarise(rows); // Works out which columns are all Y/N/undefined
const rowsWithSummaryColumn = rows.map(row => addSummaryColumn(row, columnsToSummarise));
const summaryRow = getSummaryRow(rows, columnsToSummarise); 
return [...rowsWithSummaryColumn, summaryRow];
```

Very happy to chat on this as design ideas like this can be very subjective 👍

rohan-bes (changes requested):
Few suggestions and some redesign recommendations. Happy to chat on any of it if ya like 👍 

Massive thanks for adding tests btw! Makes the behaviour of the alias transform much more apparent!

chris-pollard:
@rohan-bes have re-worked logic based on your suggestion. Ready for your review, don't mind the lint, as I will fix that once I get my eslint working properly on typescript files.

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+const detectColumnsToSummarise = (rows: Row[]): string[] => {
+  console.log(rows)
+  const anyRelevantColumns = rows.reduce((previousValue, currentValue) => {
+    const columnsWithYorN = Object.entries(currentValue).filter(([,value]) => value === 'Y' || value === 'N').map(([column,]) => column);
+    return previousValue.concat(columnsWithYorN);
+  },[] as string[])
+  const columnsToSummarise = anyRelevantColumns.reduce((previousValue: string[], currentValue: string): string[] => {
+    if (previousValue.indexOf(currentValue) > -1) {
```
rohan-bes, line 21:
```suggestion
    if (previousValue.includes(currentValue)) {
```
`Array.includes()` is probably what you want to use here (achieves the same goal but reads a little nicer). 

However, overall it looks like you're just de-duplicating the array using this reduce? Probably my favourite way to do that these days is using a `Set`, eg. `const columnsToSummarise = [... new Set(anyRelevantColumns)];`

Better still, you can use a `Set` in the first reduce like so:
```
  const setOfColumnsToSummarise = rows.reduce((columnsToSummarise, row) => {
    const columnsWithYorN = Object.entries(currentValue).filter(([,value]) => value === 'Y' || value === 'N').map(([column,]) => column);
    columnsWithYorN.forEach(column => columnsToSummarise.add(column));
    return columnsToSummarise;
  }, new Set<string>());
  return [...setOfColumnsToSummarise];

File: packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts [hunk 5]
```diff
@@ -0,0 +1,61 @@
...
+    }
+    previousValue.push(currentValue)
+    return previousValue
+  },[] as string[])
+
+  return columnsToSummarise
+};
+
+const addSummaryColumn = (row: Row, columnsToSummarise: string[]): Row => {
+  const numerator = Object.entries(row).filter(([key,value]) => columnsToSummarise.indexOf(key) > -1 && value === 'N').length
```
rohan-bes, line 32:
```suggestion
  const numerator = Object.entries(row).filter(([column,value]) => columnsToSummarise.includes(column) && value === 'N').length
```

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,61 @@
...
+  const denominator = Object.entries(row).filter(([key,]) => columnsToSummarise.indexOf(key) > -1).length
+  const summaryColumn = `${(numerator / denominator * 100).toFixed(1)}%`
```
rohan-bes, line 34: I still think it's worth extracting this ``${(numerator / denominator * 100).toFixed(1)}%`` line out into a function to be re-used in `getSummaryRow()` 👍

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,61 @@
...
+    previousValue.push(currentValue)
+    return previousValue
+  },[] as string[])
+
+  return columnsToSummarise
+};
+
+const addSummaryColumn = (row: Row, columnsToSummarise: string[]): Row => {
+  const numerator = Object.entries(row).filter(([key,value]) => columnsToSummarise.indexOf(key) > -1 && value === 'N').length
+  const denominator = Object.entries(row).filter(([key,]) => columnsToSummarise.indexOf(key) > -1).length
```
rohan-bes, line 33:
```suggestion
  const denominator = Object.entries(row).filter(([column,value]) => columnsToSummarise.includes(column) && (value === 'Y' || value === 'N')).length
```
I think there may have been a bug here previously, what happens if the `value` was `undefined`?

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,61 @@
...
+  const arrayOfColumns = columnsToSummarise.map((column: string) => {
+    const [numerator, denominator] = rows.reduce((previousValue, currentValue) => {
+      if (currentValue[column] === 'N') {
+        previousValue[0] += 1
+        previousValue[1] += 1
+      } else if (currentValue[column] === 'Y') {
+        previousValue[1] += 1
+      }
+      return previousValue
+    },[0,0])
```
rohan-bes, line 49:
```suggestion
    const { numerator, denominator } = rows.reduce((accumulator, row) => {
      if (currentValue[column] === 'N') {
        accumulator.numerator += 1
        accumulator.denominator += 1
      } else if (currentValue[column] === 'Y') {
        accumulator.denominator += 1
      }
      return accumulator;
    },{numerator: 0, denominator: 0})
```
I'm often a fan of tuples but in this instance I'd probably lean towards an object in favour of readability. 'accumulator' (or even 'accum') is a bit of a jargon-y word often used to refer to the value being accumulated as you iterate through a reduce, but feel free to pick your own word if you prefer 👍

packages/report-server/src/__tests__/reportBuilder/transform/aliases.test.ts
```diff
@@ -88,4 +89,14 @@ describe('aliases', () => {
...
   });
-});
+
+  it('insertSummaryRowAndColumn', () => {
+    const transform = buildTransform(['insertSummaryRowAndColumn']);
+    expect(transform(MULTIPLE_ANALYTICS_SUMMARY_BINARY)).toEqual([
+      { summaryColumn: '75.0%', dataElement: 'Male condoms', TO: 'N', FJ: 'N', NR: 'Y', KI: 'N' },
+      { summaryColumn: '25.0%', dataElement: 'Female condoms', TO: 'N', FJ: 'Y', NR: 'Y', KI: 'Y' },
+      { summaryColumn: '0.0%', dataElement: 'Injectable contraceptives', TO: 'Y', FJ: 'Y', NR: undefined, KI: undefined },
+      { TO: '66.7%', FJ: '33.3%', NR: '0.0%', KI: '50.0%' },
```
rohan-bes, line 99: I may have gotten a little confused somewhere but I thought the numerator is meant to be `# of Ys / # of Ys and # of Ns`?

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,61 @@
...
+  const columnsToSummarise = anyRelevantColumns.reduce((previousValue: string[], currentValue: string): string[] => {
+    if (previousValue.indexOf(currentValue) > -1) {
+      return previousValue
+    }
+    previousValue.push(currentValue)
+    return previousValue
+  },[] as string[])
+
+  return columnsToSummarise
+};
```
rohan-bes, line 29: One slight concern with this function, I think it would classify 'any columns that have a `Y` or an `N` in them' as a column to summarise? If that how we want it to behave then that's okay, but perhaps it's more correct for it to be 'any columns that have only `Y`s or `N`s in them'?

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { Row } from '../../types';
+
+/**
+ * Inserts a new column with a value that is the calculation of the percentage of 1s out of 1s and 0s in a row.
+ * { rowFieldName: 'Availability of Male Condoms', facilityNameA: 'N', facilityNameB: 'N', facilityNameC: 'Y', facilityNameD: 'N'  }
+ *  => { rowFieldName: 'Availability of Male Condoms', facilityNameA: 'N', facilityNameB: 'N', facilityNameC: 'Y', facilityNameD: 'N', rowSummary: 0.75 }
+ */
+
+const detectColumnsToSummarise = (rows: Row[]): string[] => {
```
rohan-bes, line 14: One neat thing about typescript is that it performs type inference:
```suggestion
const detectColumnsToSummarise = (rows: Row[]) => {
```
You don't need to be explicit that this function returns a `string[]` as typescript will infer that based on the return value of the function. Likewise, later on when doing `const columnsToSummarise = detectColumnsToSummarise(rows);` you don't need to specify that `columnsToSummarise` is a `string[]` as typescript will infer that as well.

The advantage of letting type inference do its magic is that in the scenario that you want to change the interface for something (eg. lets say we want this function to return a `Set` or a `Map`) then we just need to change the return value and everything automatically updates. Ideally, we are only explicit about types in the scenarios that it matters (parameters into functions, or in some scenarios that we want to be explicit for some reason).

rohan-bes (changes requested):
Looking heads better @chris-bes !! Nice work.

A few more design suggestions and a couple possible bugs that need addressing 👍

packages/report-server/src/__tests__/reportBuilder/transform/aliases.test.ts
chris-pollard: @rohan-bes it's actually the opposite. It's the percentage of Ns, because they want to display how much isn't in stock.

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
chris-pollard: @rohan-bes how would we treat undefined. If we looped through each row and some columns were undefined, we would not want to exclude them. It's really only columns with anything other than 'Y' 'N' or undefined that we would exclude.

rohan-bes: We could do it that way as well 👍 'any column that has only `Y`s, `N`s, or `undefined`'

packages/report-server/src/__tests__/reportBuilder/transform/aliases.test.ts
rohan-bes: Ah! I see, all good then 👍

packages/report-server/src/reportBuilder/transform/aliases/summaryAliases.ts
chris-pollard: @rohan-bes I've created a filter within detectColumnsToSummarise.

rohan-bes (approved):
Awesome work @chris-pollard thanks for your patience on this one! 🎉

PR #3481, opened 2021-11-11
RN-199: Allow using region in viz-builder

rohan-bes:
### Issue RN-199:

Would have been nice to use the `includeRootEntity` flag on the `/descendants` route. However, that flag _**always**_ includes the root entity, even if it doesn't match the query parameters. Perhaps it's not right that it always includes the root entity, but I'm in two minds on it... so happy to use multiple fetches here 👍 

### Changes:
- Added fetching of project entity to getEntities() in admin-panel-server


IgorNadj (approved):
👍

PR #3484, opened 2021-11-11
wai-1030-verification issue solved

c-sara:
### Changes:

- Fixed error message
- Pre-populated verification field


rohan-bes (approved):
Awesome work @c-sara! Great fix 👍 

Just one thing to note, in a few places here the code doesn't quite match out code style. Mainly scenarios were we'd include a trailing comma. Most of us manage our code style automatically through a couple of Vs Code plugins: Eslint and Prettier. Are you using Vs Code? If so install those two plugins and just turn on the `Format on Save` option 👍

rohan-bes:
So think just 2 more files are requiring a `prettier` makeover:
`packages/web-frontend/src/actions.js`
`packages/web-frontend/src/reducers.js`

Aside from that it's all good to go 👍

rohan-bes (approved):

PR #3492, opened 2021-11-15
PHX-148 Delete empty dashboard group "COVID-19 Surveillance"

c-sara:
### Changes: 

- Create migration script to delete dashboard record

### DIFF
--- a/packages/database/src/migrations/20211115030833-DeleteTongaCovid19SurveillanceGroup-modifies-data.js
+++ b/packages/database/src/migrations/20211115030833-DeleteTongaCovid19SurveillanceGroup-modifies-data.js
@@ -1,0 +1,39 @@
+('use strict');
+
+import { codeToId, deleteObject, generateId, insertObject } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+/**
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20211115030833-DeleteTongaCovid19SurveillanceGroup-modifies-data.js
```diff
@@ -0,0 +1,39 @@
...
+  await deleteObject(db, 'dashboard', { id: dashboardId });
+};
+
+exports.down = async function (db) {
+  await insertObject(db, 'dashboard', {
+    id: generateId(),
+    code: DASHBOARD_GROUP,
+    name: 'COVID 19 Surveillance',
+    root_entity_code: 'TO',
+  });
```
edmofro, line 34: Technically this `down` migration should fully reverse the `up`, which would include inserting the dashboard relation(s) as well. We're not too strict on `down` migrations though, because we barely ever use them except in local testing, so not a blocker :-)

edmofro (approved):
Looks great, nice work @c-sara!

PR #3493, opened 2021-11-15
Fix PHX-1 testing failed cases

billli0:
### Issue #:
Fix PHX-1 testing failed cases


packages/ui-components/src/components/Map/Legend/Legend.js
```diff
@@ -134,5 +135,5 @@ Legend.defaultProps = {
   className: null,
   hiddenValues: {},
   setValueHidden: null,
-  seriesesKey: null,
+  seriesesKey: 'serieses',
```
billli0, line 138: I thought defaultProps only does warning, but it did set default value for us.

packages/ui-components/src/components/Map/PopupDataItemList.js
```diff
@@ -28,31 +28,23 @@ PopupDataItem.propTypes = {
...
+      const { formattedValue, valueInfo } = getFormattedInfo(data, series, {
+        key,
+        metadata,
+        ...otherConfigs,
+      });
+      return valueInfo.hideFromPopup ? null : (
+        <PopupDataItem key={name} measureName={name} value={formattedValue} />
+      );
+    })
+    .filter(popupItem => popupItem !== null);
```
billli0, line 47: Mainly removing `showNoDataLabel` here, we hand back the logic to web-config-server ([link](https://github.com/beyondessential/tupaia/blob/404b0c49b2fdbd656bc55623934fc25c93f5db4f/packages/web-config-server/src/apiV1/measureData.js#L239)) to tell us when show no data label on tooltips.

packages/web-frontend/src/sagas.js
```diff
@@ -204,9 +205,8 @@ function* handleUserPage(userPage, initialComponents) {
 const URL_REFRESH_COMPONENTS = {
   [URL_COMPONENTS.PROJECT]: setProject,
   [URL_COMPONENTS.ORG_UNIT]: setOrgUnit,
-  [URL_COMPONENTS.MAP_OVERLAY]: setMapOverlays,
+  [URL_COMPONENTS.MAP_OVERLAY]: setMapOverlaysOnceHierarchyLoads,
   [URL_COMPONENTS.REPORT]: openEnlargedDialog,
-  [URL_COMPONENTS.OVERLAY_PERIOD]: updateCurrentOverlayConfigOnceHierarchyLoads,
```
billli0, line 209: Previously we fetch twice request when we refresh page, here is why because we loop every entry in URL_REFRESH_COMPONENTS to set state, which will call requests. 
I've removed `OVERLAY_PERIOD` and keep `MAP_OVERLAY`. Now `setMapOverlaysOnceHierarchyLoads` will set overlay and overlay period before senting a request for measureData.

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -65,7 +65,7 @@ export const selectMeasureData = createSelector(
   [state => state.map.measureInfo, (_, mapOverlayCodes) => mapOverlayCodes],
   (measureInfo, mapOverlayCodes) => {
     if (!measureInfo || mapOverlayCodes.length === 0) {
-      return undefined;
+      return [];
```
billli0, line 68: Revert it to [], because undefined cause bug.

packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -136,7 +136,7 @@ export const selectMeasuresWithDisplayInfo = createSelector(
     hiddenMeasures,
     currentCountry,
   ) => {
-    if (!currentCountry || !measureData || !country) {
```
billli0, line 139: That's where the bug I mentioned above comes from. We don't actually need to check that with [];

packages/web-frontend/src/sagas.js
```diff
@@ -908,47 +907,66 @@ function* fetchMeasureInfo() {
...
+
+    const previousDisplayedOverlays = displayedMapOverlays.filter(code =>
+      mapOverlayCodes.includes(code),
+    );
+    const newSelectedMapOverlays = mapOverlayCodes.filter(code => !measureInfo[code]);
+
+    yield fetchMeasureInfo(mapOverlayCodes, [
+      ...previousDisplayedOverlays,
+      ...newSelectedMapOverlays,
+    ]);
```
kael89, line 930: The filtering logic above seems legit, but would like to have a quick chat with you so that I can fully understand it

kael89 (approved):
Awesome work @billli0 !! Really excited about this feature, I love the recent improvements, it looks awesome! 🎉 🚀

packages/web-frontend/src/containers/MapOverlayBar/Control.js
```diff
@@ -135,7 +136,7 @@ export const Control = ({
   );
 
   return (
-    <Container>
+    <Container $isExpanded={isExpanded}>
```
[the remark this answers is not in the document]
billli0: In styled component, the dollar sign $ helps us to pass the parameter to the DOM through Material ui component. 
This is called the transient props, it is more like a defender to prevent passing everything to DOM: https://medium.com/@probablyup/introducing-transient-props-f35fd5203e0c
https://styled-components.com/docs/api#transient-props

packages/web-frontend/src/containers/MapOverlayBar/MapOverlayBar.js
```diff
@@ -43,28 +44,41 @@ const pinnedOverlayCodeReducer = (currentOverlayCode, newOverlayCode) => {
...
   }, [currentMapOverlayCodes]);
+  useEffect(() => {
+    if (maxSelectedOverlays === 1) {
+      setPinnedOverlay(null);
+      onSetDisplayedMapOverlays(currentMapOverlayCodes);
+    }
+    if (maxSelectedOverlays < currentMapOverlayCodes.length) {
+      onSetMapOverlay(currentMapOverlayCodes.slice(0, maxSelectedOverlays));
+    }
+  }, [maxSelectedOverlays]);
```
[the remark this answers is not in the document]
billli0: My understanding on `useEffect` is it usually do fetch or async function in it. 

One example is we do `fetch(A,B,C)` in useEffect, we want to put `[A,B,C]` as dependency because we want to fetch when one of them has been changed. But if we put [A,B] in the dependency array, when `A` is changed, fetch should use the latest `C` (if `C` has been updated). 

In our case `currentMapOverlayCodes` is more like the `C`. We don't want to run this useEffect if `currentMapOverlayCodes` has been changed. 

I've also tested it locally to make sure it is using the latest `currentMapOverlayCodes`.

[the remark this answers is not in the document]
kael89: Oh I didn't know that we don't want to re-fetch when `currentMapOverlayCodes` changes. In that case, what you did seems perfect, thanks for clarifying @billli0 !

packages/web-frontend/src/containers/MapOverlayBar/Control.js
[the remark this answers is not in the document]
kael89: I didn't know that, that's awesome! 💯

PR #3508, opened 2021-11-18
WAI-963: Wait a full minute for the database to become available

edmofro:
### Issue #:
If the database instance and server instance are started at the same time, sometimes the pm2 processes will start up before the database is available. This means the servers will be a little more patient before giving up


nick-beyondessential (approved):
LGTM

edmofro:
Replaced with https://github.com/beyondessential/tupaia/pull/3509 because of the bad branch name, could you review that @nick-beyondessential?

PR #3510, opened 2021-11-18
Only add default Start/StopAtUTC tags for non-prod

IgorNadj:
Tested with a feature branch `spin_up_tupaia_deployment`, production release through CI.

### DIFF
--- a/packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
+++ b/packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
@@ -71,11 +71,20 @@ def spin_up_tupaia_deployment(event):
         delete_after = datetime.now() + timedelta(hours=event['HoursOfLife'])
         extra_tags.append({ 'Key': 'DeleteAfter', 'Value': format(delete_after, "%Y-%m-%d %H:%M") })
 
-    if 'StartAtUTC' in event:
-        extra_tags.append({ 'Key': 'StartAtUTC', 'Value': event['StartAtUTC'] })
+    if deployment_name == 'production':
+        if 'StartAtUTC' in event or 'StopAtUTC' in event:
+            raise Exception('Production deployment cannot have StartAtUTC/StopAtUTC')
+    else:
+        if 'StartAtUTC' in event:
+            extra_tags.append({ 'Key': 'StartAtUTC', 'Value': event['StartAtUTC'] })
+        else:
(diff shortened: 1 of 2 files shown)


packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
```diff
@@ -71,11 +71,20 @@ def spin_up_tupaia_deployment(event):
...
         extra_tags.append({ 'Key': 'DeleteAfter', 'Value': format(delete_after, "%Y-%m-%d %H:%M") })
 
-    if 'StartAtUTC' in event:
-        extra_tags.append({ 'Key': 'StartAtUTC', 'Value': event['StartAtUTC'] })
+    if deployment_name == 'production':
+        if 'StartAtUTC' in event or 'StopAtUTC' in event:
+            raise Exception('Production deployment cannot have StartAtUTC/StopAtUTC')
+    else:
+        if 'StartAtUTC' in event:
+            extra_tags.append({ 'Key': 'StartAtUTC', 'Value': event['StartAtUTC'] })
```
nick-beyondessential, line 79: Wondering if we should check for a valid start/stop time here? Like would it cause issues later when trying to read these tags if invalid time/empty string entered by accident? No biggy as this is how it was working previously.

nick-beyondessential (approved):
LGTM

packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
IgorNadj: Yeah good point. Maybe in the future if we run across any issues, it's mostly automatically set

packages/devops/scripts/lambda/helpers/creation.py
```diff
@@ -68,12 +68,6 @@ def get_instance_creation_config(
       tags = tags + extra_tags
       extra_tag_keys = [extra_tag['Key'] for extra_tag in extra_tags]
 
-    if 'StopAtUTC' not in extra_tag_keys:
-      tags.append({ 'Key': 'StopAtUTC', 'Value': '09:00'}) # 9am UTC is 7pm AEST, 8pm AEDT, 9pm NZST, 10pm NZDT
-
-    if 'StartAtUTC' not in extra_tag_keys:
```
edmofro, line 74: This does mean that new deployments of DHIS2 won't get the StarAt/StopAt tags, but that's probably ok considering we can add them through the event payload

edmofro (commented):
Thanks for this @IgorNadj!

PR #3512, opened 2021-11-21
WAI-963: Additional tweaks

edmofro:
### Issue #:
WAI-963 (already released, these are tweaks only)

### Changes:
Commit names show the changes. The main change is the final commit, which adds a "DeploymentType" tag on instances, gateways, etc. so that we can distinguish between tupaia, regional-dhis2, and tonga-dhis2 deployments. Currently there are some that share a DeploymentName (e.g. dev, e2e) so you can unintentionally tear down instances you don't mean to.

### DIFF
--- a/packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
+++ b/packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
@@ -91,7 +91,9 @@ def spin_up_tupaia_deployment(event):
     # do this after the server has started because it will take a while to run its startup script, so
     # we might as well be cloning the db instance at the same time, so long is it is available before
     # the server first tries to connect
+    deployment_type='tupaia'
     clone_instance(
+        deployment_type,
         clone_db_from,
         deployment_name,
         instance_type,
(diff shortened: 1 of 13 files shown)


packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
```diff
@@ -91,7 +91,9 @@ def spin_up_tupaia_deployment(event):
     # do this after the server has started because it will take a while to run its startup script, so
     # we might as well be cloning the db instance at the same time, so long is it is available before
     # the server first tries to connect
+    deployment_type='tupaia'
     clone_instance(
+        deployment_type,
```
IgorNadj, line 105: deployment_type='tupaia' should go in arg not above it?

packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
```diff
@@ -106,7 +106,7 @@ def spin_up_tupaia_deployment(event):
         clone_db_from,
         deployment_name,
         instance_type,
-        extra_tags=extra_tags,
+        extra_tags=extra_tags + [{ 'Key': 'DeploymentComponent', 'Value': 'db' }],
```
IgorNadj, line 109: Happy with this, but could also do DeploymentType = 'TupaiaDb' ?

IgorNadj (approved):
LGTM

packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
IgorNadj: Discussed, it does work as is.

IgorNadj:
Merging for Edwin, afk

PR #3514, opened 2021-11-22
add map table modal to lesmis

tcaiger:
### Changes:

- Add MapTable Modal to lesmis
- When on MapView add the modal

---

### Screenshots:
See issue


packages/lesmis/src/components/LocationHeader.js
```diff
@@ -116,9 +131,21 @@ export const LocationHeader = () => {
...
+                title={exportTitle}
+                overlayReportData={overlayReportData}
+                Button={props => (
+                  <IconButton {...props} startIcon={<AssignmentIcon />}>
+                    <I18n t="dashboards.export" />
+                  </IconButton>
+                )}
+              />
+            ) : (
+              <IconButton startIcon={<GetApp />}>
```
edmofro, line 145: Now that we're close to releasing it without including full dashboard export, we should probably hide this button that does nothing right @enunan?

edmofro (approved):
lgtm

PR #3515, opened 2021-11-22
WAI-963: Fail build when committing to a deployed branch out of hours

edmofro:
### Issue #:
WAI-963 (already released)

### Changes:
If someone pushes to a branch out of hours, the redeploy will fail, but the build will still go green. This is because when the servers try to start up, the db instance will be sleeping. This change doesn't actually make it work (if we want to fix this properly, we should wait until the RDS transition is complete), but does mean that if the deployment is stopped, codeship won't trigger a redeploy, but will mark the build as failed so the dev knows to retry it during office hours.


IgorNadj (approved):
Good solution, nice one

PR #3518, opened 2021-11-22
[HOT FIX] Avoid multiple redeploy attempts

edmofro:
### Issue #:
Now that we have four deployments all connected to the master branch, the lambda redeploy function was taking longer than the default AWS CLI timeout. The AWS CLI then kicks into "helpful" retry attempts, which resulted in multiple deployments coming online for production, e2e, viz-testing and viz-sandbox.

This change:
- Increases the AWS CLI timeout to match lambda's 15 min max execution time
- Turns off the retry attempt behaviour, lest it try to be helpful again in future

### DIFF
--- a/packages/devops/scripts/lambda/actions/redeploy_tupaia_server.py
+++ b/packages/devops/scripts/lambda/actions/redeploy_tupaia_server.py
@@ -85,7 +85,7 @@ def redeploy_tupaia_server(event):
             deployment_name=get_tag(existing_instance, 'DeploymentName'),
             branch=get_tag(existing_instance, 'Branch'),
             instance_type=event.get('InstanceType', existing_instance['InstanceType']),
-            extra_tags=extra_tags,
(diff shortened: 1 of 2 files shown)


packages/devops/scripts/lambda/actions/redeploy_tupaia_server.py
```diff
@@ -85,7 +85,7 @@ def redeploy_tupaia_server(event):
             deployment_name=get_tag(existing_instance, 'DeploymentName'),
             branch=get_tag(existing_instance, 'Branch'),
             instance_type=event.get('InstanceType', existing_instance['InstanceType']),
-            extra_tags=extra_tags,
+            extra_tags=extra_tags + [{ 'Key': 'DeploymentComponent', 'Value': 'app-server' }], # TODO remove deployment component stuff after move to RDS
```
edmofro, line 88: Unrelated nice to have

nick-beyondessential (approved):
Lgtm

PR #3519, opened 2021-11-22
wai-352: Restrict analytics to production

tcaiger:
### Changes:

- Use REACT_APP_DEPLOYMENT_NAME value to only run analytics on live deployments

---


edmofro (approved):
LGTM!

PR #3520, opened 2021-11-23
update column title to match legend

c-sara:
### Changes:

- Updated title to take the label so the title does not distort


packages/ui-components/src/components/Chart/Table/getChartTableData.js
```diff
@@ -90,8 +90,10 @@ const processData = viewContent => {
 export const getChartTableData = viewContent => {
   const columns = useMemo(() => processColumns(viewContent), [JSON.stringify(viewContent)]);
   const data = useMemo(() => processData(viewContent), [JSON.stringify(viewContent)]);
+  const { chartConfig } = viewContent;
```
billli0, line 93: I played around within this file and found we actually can add the label in here specifying the `Header` config as a function in  `processColumns()`. API reference: https://react-table.tanstack.com/docs/api/useTable

I feel it would be better to put the logic here along with other logics. What I did is to put the following code after line 64:

>   const configColumns = Object.keys(chartDataConfig).map(columnKey => {
    return {
      Header: ({ column }) => {
        return chartConfigs[column.id]?.label || column.id;
      },
      id: columnKey, 


Happy to have a chat to talk it through 😬.

billli0 (changes requested):
@c-sara Sorry to request the changes that goes with a slight different path from what we've discussed before. Happy to talk it through if it is confusing!

billli0 (approved):
Give a pre approve! We still need to restore some configs for `store.json` but overall looks good to me! 

Please let tester know we also want to smoke test excel export and if other tables have been impacted.

PR #3523, opened 2021-11-23
RN-79 see viz using data element

IgorNadj:
### Changes:
- Implement used by for Data Elements
- Implement used by for Questions

---


packages/admin-panel/src/usedBy/UsedBy.js
```diff
@@ -0,0 +1,121 @@
+/**
+ * Tupaia MediTrak
+ * Copyright (c) 2018 Beyond Essential Systems Pty Ltd
```
nick-beyondessential, line 3:
```suggestion
 Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
```
and files below.

nick-beyondessential: Though we should update to 2022 soon?

nick-beyondessential (approved):
Looks good to me!
Small copyright suggestion.

PR #3524, opened 2021-11-23
PHX-199 Delete Test Country

avaek:
### Issue #:
- https://linear.app/bes/issue/PHX-199/there-is-a-country-in-the-country-table-called-klaus-test

This migration removes an accidental country from the country table as per the above ticket


kael89 (approved):
Looks good, thanks for fixing this!

PR #3525, opened 2021-11-23
PHX-165 Insert PenFa'a Samoa project

billli0:
### Issue #:
PHX-165 PenFa'a Samoa project creation

Note: have not confirmed we need to exclude cases or other entity types in `frontendExcludedTypes`. If so we can easily add it later in admin-panel.


kael89 (approved):
Looks good!

PR #3526, opened 2021-11-24
Add map overlay group for project pen fa'a

billli0:
### Issue #:
No issue

### Changes:
That's the only overlay group we need for project `pen fa'a`. Since it is just an overlay group, I assume that we can put no testing required label.


kael89 (approved):
👍

PR #3532, opened 2021-11-25
Fix yarn install during deployment

edmofro:
### Issue #:
Deployments have been failing because:
- The `yarn.lock` file in dev is out of date with the actual dependencies required
- At the end of `yarn install` during startup, yarn tries to update that `yarn.lock` to reflect the current status
- The `tupaia` repo is owned by the root user on the gold master

This PR fixes all three of those issues, as well as two unrelated nits that I came across during debugging.

### DIFF
--- a/packages/devops/scripts/deployment/buildDeployablePackages.sh
+++ b/packages/devops/scripts/deployment/buildDeployablePackages.sh
@@ -12,7 +12,7 @@ PACKAGES=$(${TUPAIA_DIR}/scripts/bash/getDeployablePackages.sh)
 
 # Install external dependencies and build internal dependencies
 cd ${TUPAIA_DIR}
-yarn install
+yarn install --non-interactive --frozen-lockfile
 
 # Inject environment variables from LastPass
 LASTPASS_EMAIL=$($DIR/fetchParameterStoreValue.sh LASTPASS_EMAIL)
--- a/packages/devops/scripts/deployment/setupGoldMaster.sh
+++ b/packages/devops/scripts/deployment/setupGoldMaster.sh
@@ -51,7 +51,7 @@ mkdir -p /home/ubuntu/.local/share/lpass
 
 # clone our repo
 cd /home/ubuntu
-git clone https://github.com/beyondessential/tupaia.git
+sudo -Hu ubuntu git clone https://github.com/beyondessential/tupaia.git
 
 # build all packages once using dev to speed up future branch-specific builds
-/home/ubuntu/tupaia/packages/devops/scripts/deployment/buildDeployablePackages.sh gold-master-image-builder
+sudo -Hu ubuntu /home/ubuntu/tupaia/packages/devops/scripts/deployment/buildDeployablePackages.sh gold-master-image-builder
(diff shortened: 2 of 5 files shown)


packages/devops/scripts/deployment/buildDeployablePackages.sh
```diff
@@ -12,7 +12,7 @@ PACKAGES=$(${TUPAIA_DIR}/scripts/bash/getDeployablePackages.sh)
 
 # Install external dependencies and build internal dependencies
 cd ${TUPAIA_DIR}
-yarn install
+yarn install --non-interactive --frozen-lockfile
```
edmofro, line 15: This matches what we do during codeship builds: don't update dependency versions within their accepted range or try to consolidate versions, or do any other fancy `yarn` stuff...just accept the current version of the lockfile and install the dependencies recorded there.

packages/devops/scripts/deployment/setupGoldMaster.sh
```diff
@@ -51,7 +51,7 @@ mkdir -p /home/ubuntu/.local/share/lpass
 
 # clone our repo
 cd /home/ubuntu
-git clone https://github.com/beyondessential/tupaia.git
+sudo -Hu ubuntu git clone https://github.com/beyondessential/tupaia.git
```
edmofro, line 54: Building new gold master images takes a while so I'm actually still testing this, won't merge until I have confirmed it works (or rolled it back)

IgorNadj (approved):
Great changes! Thanks!

yarn.lock
```diff
@@ -5223,17 +5223,7 @@
...
-  resolved "https://registry.yarnpkg.com/@typescript-eslint/eslint-plugin/-/eslint-plugin-2.34.0.tgz#6f8ce8a46c7dea4a6f1d171d2bb8fbae6dac2be9"
-  integrity sha512-4zY3Z88rEE99+CNvTbXSyovv2z9PNOVffTWD2W8QF5s2prBQtwN2zadqERcrHpcR7O/+KMI3fcTAmUUhK/iQcQ==
-  dependencies:
-    "@typescript-eslint/experimental-utils" "2.34.0"
-    functional-red-black-tree "^1.0.1"
-    regexpp "^3.0.0"
-    tsutils "^3.17.1"
-
-"@typescript-eslint/eslint-plugin@^4.18.0", "@typescript-eslint/eslint-plugin@^4.5.0":
+"@typescript-eslint/eslint-plugin@^2.25.0", "@typescript-eslint/eslint-plugin@^4.18.0", "@typescript-eslint/eslint-plugin@^4.5.0":
```
edmofro, line 5226: This is the change that broke everything - `yarn.lock` in dev wasn't fully up to date.

packages/devops/scripts/deployment/setupGoldMaster.sh
edmofro: For future reference @IgorNadj this doesn't seem to have changed the owner of the tupaia repo, but it also did no harm, so I just left it in.

PR #3540, opened 2021-11-26
Wai 1039 lesmis email verification

tcaiger:
### Changes:
- add primary_platform column to user_account table
- update registration handler to send email depending on primary platform
- update user entity permission notifier to send email depending on primary platform
- add verify-email route on lesmis frontend

---

### Screenshots:
See issue


packages/meditrak-server/src/apiV2/requestPasswordReset.js
```diff
@@ -28,7 +28,8 @@ export const requestPasswordReset = async (req, res) => {
     user_id: user.id,
   });
 
-  const resetUrl = process.env.PASSWORD_RESET_URL.replace('{token}', token);
+  const passwordResetUrl = `${process.env.TUPAIA_FRONT_END_URL}/reset-password?passwordResetToken={token}`;
```
tcaiger, line 31: Replace PASSWORD_RESET_URL  and VERIFY_EMAIL_URL  with TUPAIA_FRONT_END_URL to save duplication.

packages/lesmis-server/src/routes/VerifyEmailRoute.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  async buildResponse() {
+    const { emailToken } = this.req.params;
+
+    const response = await this.meditrakConnection.verifyEmail(emailToken);
+    if (response.emailVerified === 'true') {
+      return response;
+    }
+
+    throw Error('Email was not correctly verified');
```
edmofro, line 27: I wonder if the error message meditrak-server returns might be more informative?

packages/lesmis/src/api/queries/useEmailVerification.js
```diff
@@ -0,0 +1,17 @@
...
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import { useQuery } from 'react-query';
+import { useUrlSearchParams } from '../../utils/useUrlSearchParams';
+import { get } from '../api';
+
+export const useEmailVerification = () => {
+  const [{ verifyEmailToken }] = useUrlSearchParams();
+
+  return useQuery(['user', verifyEmailToken], () => get(`verify/${verifyEmailToken}`), {
```
edmofro, line 12: Does `web-frontend` use a `get` as well? I would have thought this would be a `post`

packages/meditrak-server/src/database/models/UserEntityPermission.js
```diff
@@ -11,6 +11,28 @@ export class UserEntityPermissionModel extends CommonUserEntityPermissionModel {
...
+  },
+  lesmis: {
+    subject: 'LESMIS Permission Granted',
+    body: (userName, permissionGroupName, entityName) =>
+      `Hi ${userName},\n\n` +
+      `This is just to let you know that you've been added to the ${permissionGroupName} access group for ${entityName}. ` +
+      'This allows you to see reports and map overlays on lesmis.la.\n\n' +
+      "Please note that you'll need to log out and then log back in to get access to the new permissions.\n\n" +
+      'Feel free to get in touch if you have any questions.\n',
+    signOff: 'Best regards,\nThe LESMIS Team',
```
edmofro, line 32: I like that the LESMIS team is more formal 😀

packages/meditrak-server/src/apiV2/userAccounts/CreateUserAccounts.js
```diff
@@ -76,7 +78,7 @@ export class CreateUserAccounts extends CreateHandler {
         });
       }
 
-      return { userId: user.id, secretKey };
+      return { ...user, userId: user.id, secretKey };
```
edmofro, line 81: Bit of a danger here that we'll expose the hashed password to the front end, because it's a generic helper that might be used in a bunch of places. Let's just pass back what we need, and if we need the hashed password from the user in one of the places that calls this, refetch the user from the db at that point.

packages/lesmis-server/src/routes/VerifyEmailRoute.ts
edmofro: Have checked and it's also very generic (`'Email address could not be verified'`), but might as well try to use it in case we change that logic centrally later. I'm also fairly sure that it will throw an error on line 22 if the verification doesn't work properly, rather than having a response that doesn't have `emailVerified` - could you test to make sure the failing case works? E.g. pass in an invalid token

packages/meditrak-server/src/apiV2/verifyEmail.js
```diff
@@ -78,7 +54,7 @@ export const requestResendEmail = async (req, res) => {
     throw new FormValidationError(`Unable to send verification email to ${emailAddress}`);
   }
 
-  sendEmailVerification(user);
+  await sendEmailVerification(user);
```
edmofro, line 57: I'd say this was originally intentional, as we don't need to wait for the email to be sent before we respond, but it's probably a good call in case there's an error in sending it.

edmofro (changes requested):
Was definitely a good call adding that additional column, makes all the logic quite clear and simple

packages/lesmis-server/src/routes/VerifyEmailRoute.ts
tcaiger: I updated it to just return the response which definitely makes more sense. I tested the failing case and works well.

packages/lesmis/src/api/queries/useEmailVerification.js
tcaiger: Yes web frontend is a get too

packages/meditrak-server/src/database/models/UserEntityPermission.js
tcaiger: Haha I'm not really sure why I did that but it seemed like I needed to

edmofro (approved):
lgtm!

PR #3542, opened 2021-11-26
PHX-206 Get columns using header groups

billli0:
### Issue #:
PHX-206

### Changes:
Bug caused by PHX-154 that `Header` have been modified as a function or string.

### DIFF
--- a/packages/ui-components/src/components/DataTable/useDataTableExport.js
+++ b/packages/ui-components/src/components/DataTable/useDataTableExport.js
@@ -9,14 +9,18 @@ import { useTable } from 'react-table';
 import moment from 'moment';
 
 export const useDataTableExport = (columns, data, title) => {
-  const { rows: tableData, columns: tableColumns } = useTable({
+  const { headerGroups, rows: tableData, columns: tableColumns } = useTable({
     columns,
     data,
   });
 
   const doExport = () => {
     // Get data from react table properties
-    const header = [tableColumns.map(col => col.Header)];
+    const header = headerGroups.map(({ headers }) =>
+      headers.map(({ Header: getHeader, id }) =>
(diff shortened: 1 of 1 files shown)


packages/ui-components/src/components/DataTable/useDataTableExport.js
```diff
@@ -9,14 +9,16 @@ import { useTable } from 'react-table';
...
+  const { headerGroups, rows: tableData, columns: tableColumns } = useTable({
     columns,
     data,
   });
 
   const doExport = () => {
     // Get data from react table properties
-    const header = [tableColumns.map(col => col.Header)];
+    const header = headerGroups.map(({ headers }) =>
+      headers.map(({ Header: getHeader, id }) => getHeader({ column: { id } })),
```
edmofro, line 20: Could you make this polymorphic, so that it can handle `Header` being a string or a function? In the style of e.g. https://github.com/beyondessential/tupaia/blob/dev/packages/ui-components/src/components/Table/TableRow.js#L32

packages/ui-components/src/components/Chart/Table/getChartTableData.js
```diff
@@ -23,7 +23,8 @@ const FirstColumnCell = styled.span`
 `;
 
 const makeFirstColumn = (header, accessor) => ({
-  Header: header,
+  id: header,
+  Header: () => header,
```
edmofro, line 27: If you follow the other suggestion, this can be reverted

edmofro (changes requested):
A suggestion to make the simple string definition of headers work as well 🙏

packages/ui-components/src/components/DataTable/useDataTableExport.js
billli0: Sounds good!

edmofro (approved):
lgtm!

PR #3543, opened 2021-11-26
WAI-1051: Fail when some part of the build errors

edmofro:
### Issue #:
WAI-1051

### DIFF
--- a/packages/devops/scripts/ci/triggerRedeploy.sh
+++ b/packages/devops/scripts/ci/triggerRedeploy.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 
 STOPPED_INSTANCES=$(aws ec2 describe-instances \
       --filters Name=tag:Branch,Values=${CI_BRANCH} Name=tag:DeploymentType,Values=tupaia Name=tag:DeploymentComponent,Values=app-server Name=instance-state-name,Values=stopped \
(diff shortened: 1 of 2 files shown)


packages/devops/scripts/ci/triggerRedeploy.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
```
edmofro, line 1: Fail the whole redeploy if any of the scripts within exit with a non-zero code

packages/devops/scripts/lambda/resources/startup.sh
```diff
@@ -1,7 +1,9 @@
-#!/bin/bash -le
+#!/bin/bash -leE
```
edmofro, line 1: The capital E means errors within called scripts bubble up to be caught by the top level `trap` (see below)

packages/devops/scripts/lambda/resources/startup.sh
```diff
@@ -19,9 +28,10 @@ echo "Starting up ${DEPLOYMENT_NAME} (${BRANCH})"
 mkdir -p $LOGS_DIR
 
 # Turn on cloudwatch agent for prod and dev (can be turned on manually if needed on feature instances)
-if [[ $DEPLOYMENT_NAME == "production" || $DEPLOYMENT_NAME == "dev" ]]; then
-    $DEPLOYMENT_SCRIPTS/startCloudwatchAgent.sh | while IFS= read -r line; do printf '\%s \%s\n' "$(date)" "$line"; done  >> $LOGS_DIR/deployment_log.txt
-fi
+# TODO currently broken
```
edmofro, line 31: This was already broken, but now that we're using pipefail on this script, it matters. We haven't really got a good solution for pushing telemetry to cloudwatch now that our instances are short-lived.

IgorNadj (approved):
LGTM!

PR #3545, opened 2021-11-28
PHX-202: Merge Flutracking responses

kael89:
### Issue #:
PHX-202

This not your usual migration!

It takes 1 hour to process FLWV responses, and 2-3 hours (not sure) for FPWV. Because of this, we could consider a slightly different way to deploy it:
1. Deploy separately as a hotfix after hours (I volunteer to do this)
2. This migration commits changes in batches, so it doesn't have to run all at once. Even if we stop and start it (or it stops for a different reason), the changes that have been processed already will have been committed to the db.


packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
```diff
@@ -0,0 +1,189 @@
...
+
+  console.log(
+    'Processing response groups in batches. Each batch will be processed as a transaction',
+  );
+  console.log(`Batches found: ${responseGroups.length}`);
+  await processInBatches(responseGroups, RESPONSE_GROUP_BATCH_SIZE, async (responseGroup, i) => {
+    await db.runSql('START TRANSACTION');
+    printProgress(i, responseGroups.length);
+    await processResponseGroups(db, responseGroup);
+    await db.runSql('COMMIT');
```
kael89, line 172: This is kind of required for two reasons:
1. If we don't commit the current changes, memory builds up and quickly brings the system to its knees
2. This is a long running migration so it's probably desired to commit changes for each batch (data for each of them is independent). This way, even if the migration stop halfway through, we can run `yarn  migrate` again and it will continue from the last "checkpoint".

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
```diff
@@ -0,0 +1,189 @@
...
+
+  return { responseIdsToDelete, answersToCreate };
+};
+
+const selectAnswersForResponseGroups = async (db, responseGroups) => {
+  const responseIds = responseGroups.map(group => group.map(r => r.id)).flat();
+  const { rows: answers } = await db.runSql(`
+    select * from answer where survey_response_id IN (${arrayToDbString(responseIds)});
+  `);
+  return groupBy(answers, 'survey_response_id');
```
edmofro, line 95: Efficient!

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
```diff
@@ -0,0 +1,189 @@
...
+};
+
+const processResponseGroups = async (db, responseGroupBatch) => {
+  const a = Date.now();
```
edmofro, line 99: Timing tests artefact - no longer needed?

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
```diff
@@ -0,0 +1,189 @@
...
+  const { rows } = await db.runSql(`
+    select
+      sr.id as id1,
+      sr2.id as id2,
+      sr.end_time as end_time1,
+      sr2.end_time as end_time2,
+      sr.entity_id,
+      concat(date_part('year', sr.data_time), 'W', date_part('week', sr.data_time)) as iso_week
+    from survey_response sr
+    join survey_response sr2 on sr2.id > sr.id and sr2.survey_id = sr.survey_id and sr2.entity_id = sr.entity_id
```
edmofro, line 146: `>` is such a clever way of only getting the duplicates in one direction, not double matching!

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
edmofro: Yes great pattern! I understood the migration as a whole to be wrapped in a transaction by `pg-migrate`...how do these inner transactions interact with that?

edmofro (approved):
I have one question about how the transactions interact, but given you've tested it I'm sure it is fine, so this all looks great and ready to go!

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
kael89: You are correct, this is also my understanding! But internal `start transaction`/`commit` queries seem to work too, which comes handy in data-heavy migrations like this one

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
kael89: Oops yes - good catch!

edmofro (approved):

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
```diff
@@ -165,14 +157,24 @@ const mergeFlutrackingResponses = async (db, surveyCode) => {
...
   console.log(`Batches found: ${responseGroups.length}`);
+
+  let prevPercentage = -1;
   await processInBatches(responseGroups, RESPONSE_GROUP_BATCH_SIZE, async (responseGroup, i) => {
     await db.runSql('START TRANSACTION');
-    printProgress(i, responseGroups.length);
+
+    // Print progress
+    const percentage = getPercentage(i, responseGroups.length);
+    if (parseInt(percentage - prevPercentage) > 0) {
```
kael89: The previous logic in `printProgress` allows us to update the last console output line with a changing percentage, which is cool, but for some reason doesn't work on the deployment logs.

What we now do here is printing an integer percentage every time the integer part changes (i.e. 100 times max per process)

kael89:
@edmofro I rebased from dev so it may be harder to see what has changed, you have reviewed everything except for the last 2 commits (starting from "Process a subset of data at each run")

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
```diff
@@ -158,19 +158,20 @@ const mergeFlutrackingResponses = async (db, surveyCode) => {
...
-    // Print progress
-    const percentage = getPercentage(i, responseGroups.length);
-    if (parseInt(percentage - prevPercentage) > 0) {
-      // Print percentage when its integer part changes
-      console.log(`${Math.round(percentage)}%`);
-      prevPercentage = percentage;
-    }
-
-    await processResponseGroups(db, responseGroup);
+    // Print percentage when its integer part changes
```
edmofro: This comment is no longer needed - I assume because it's a smaller total number now, we don't need to stress about the spam of printing every time

edmofro (approved):
lgtm!

packages/database/src/migrations/20211128090124-MergeFlutrackingResponses-modifies-data.js
kael89: Indeed! 👍

PR #3547, opened 2021-11-29
WAI-832: Complete Kobo Data Approval Front End

tcaiger:
### Changes:

- Update survey response table filters to use approval status
- Add tab for all survey responses
- hook up approve button to meditrak server

---


packages/lesmis/src/routes/AdminPanelRoutes.js
```diff
@@ -34,36 +34,32 @@ import {
   DraftSurveyResponsesView,
 } from '../views/AdminPanel/SurveyResponsesView';
 
-const ADMIN_URL = '/admin';
```
tcaiger, line 37: I'm also taking this chance to fix a regression in the admin panel url that was introduced during internationalisation. The admin segment is now added upstream

packages/ui-components/src/components/Inputs/RadioGroup.js
```diff
@@ -17,6 +17,7 @@ const FormControl = styled(MuiFormControl)`
 `;
 
 const FormLabel = styled(MuiFormLabel)`
+  position: relative;
```
tcaiger, line 20: fix a weird chrome bug where legend elements don't display

packages/lesmis-server/src/routes/ApproveSurveyResponse.ts
```diff
@@ -19,8 +19,6 @@ export class ApproveSurveyResponse extends Route {
 
   async buildResponse() {
     const { id } = this.req.params;
-    console.log('Approve Survey Response. Id: ', id);
-    await sleep(2000);
-    return true;
+    return this.meditrakConnection.updateSurveyResponse(id);
```
edmofro, line 22: I would either go with `approveSurveyResponse`, or provide the mutation that's being made as a second parameter `{ approval_status: 'approved' }`

packages/lesmis/src/views/AdminPanel/SurveyResponsesView.js
```diff
@@ -11,7 +11,7 @@ import { ApproveButton } from '../../components';
 export const ApprovedSurveyResponsesView = props => (
   <SurveyResponsesPage
     title="Approved Survey Responses"
-    baseFilter={{ 'survey.code': { comparator: 'ILIKE', comparisonValue: '%_Confirmed_WNR' } }}
+    baseFilter={{ approval_status: { comparisonValue: 'approved' } }}
```
edmofro, line 14: So much nicer!

packages/lesmis/src/views/AdminPanel/SurveyResponsesView.js
```diff
@@ -78,7 +78,7 @@ export const DraftSurveyResponsesView = props => (
   <SurveyResponsesPage
```
edmofro, line 78: Comment about lines 50 - 74: Could you please change the reject button from a straight delete to a mutation of `{ approval_status: 'rejected' }`?

edmofro (changes requested):
Glad these changes were nice and simple - flexible initial design, and good early call to not put too much work into cloning the survey responses across to the approved version!

packages/lesmis-server/src/app/createApp.ts
```diff
@@ -41,8 +41,8 @@ export function createApp() {
     .get('/v1/map-overlays/:entityCode', handleWith(MapOverlaysRoute))
     .get('/v1/entity/:entityCode', handleWith(EntityRoute))
     .get('/v1/report/:entityCode/:reportCode', handleWith(ReportRoute))
+    .put('/v1/updateSurveyResponse/:id', handleWith(UpdateSurveyResponseRoute))
```
edmofro, line 44: Sorry I realise I suggested the name, but have just realised that the http verb `put` implies the `update`, so we can just call this `/v1/surveyResponse/:id`. All the other `Update` places are appropriate, just the endpoint should change 🙏

packages/lesmis/src/views/AdminPanel/SurveyResponsesView.js
tcaiger: Good idea. I also updated the all survey responses tab to now be rejected so there is a nice three tab flow (Under Review, Approved, Rejected)

packages/lesmis/src/views/AdminPanel/SurveyResponsesView.js
```diff
@@ -49,37 +55,19 @@ const COLUMNS = [
...
-  );
-};
-
-// Todo: Replace base filter with real filter @see WAI-832
 export const DraftSurveyResponsesView = props => (
   <SurveyResponsesPage
     {...props}
     title="Survey Responses For Review"
-    baseFilter={{ 'survey.code': { comparator: 'NOT ILIKE', comparisonValue: '%_Confirmed_WNR' } }}
+    baseFilter={{ approval_status: { comparisonValue: 'pending' } }}
```
edmofro, line 70: We probably do need a fourth page to show those that don't go through an approval workflow at all

packages/lesmis/src/views/AdminPanel/SurveyResponsesView.js
```diff
@@ -49,37 +55,19 @@ const COLUMNS = [
   {
     Header: 'Reject',
     source: 'id',
+    Cell: RejectButton,
     type: 'delete',
```
edmofro, line 59: Can this type be dropped now?

edmofro (changes requested):
A few more small changes requested 🙏

packages/lesmis-server/src/app/createApp.ts
tcaiger: Good point. Should it be snake case too?

packages/lesmis/src/views/AdminPanel/SurveyResponsesView.js
tcaiger: hmmm maybe the not required ones should just go into the approved tab?

edmofro: Good thought! I just chatted to @enunan and she prefers having a separate table for those that don't require approval

packages/lesmis-server/src/app/createApp.ts
edmofro: Oh yeah looks like it based on the other endpoints in here, didn't even notice that 👀

packages/lesmis/src/views/AdminPanel/SurveyResponsesView.js
tcaiger: actually the type brings in the styling for the column so I will leave it as is

edmofro (approved):
lgtm!

PR #3549, opened 2021-11-29
Support fixed line in bar chart

billli0:
### Issue #:
Requirement from PHX-168

### Changes:
Support fixed line in bar chart


kael89 (approved):
Looks good!

PR #3556, opened 2021-12-01
WAI-1054: Fix timeouts when processing access requests

edmofro:
### Issue #:
Was caused by a build up of queries against the entity table, which is huge and possibly poorly indexed, so the queries were taking a long time to return. Filtering by country is both a UX improvement, and means the query returns very quickly.

### Changes:
-  Only show countries when granting access requests or permissions

### DIFF
--- a/packages/admin-panel/src/autocomplete/Autocomplete.js
+++ b/packages/admin-panel/src/autocomplete/Autocomplete.js
@@ -153,6 +153,7 @@ const mapDispatchToProps = (
     onChange,
     parentRecord = {},
     allowMultipleValues,
+    baseFilter,
   },
 ) => ({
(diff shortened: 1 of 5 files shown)


kael89 (approved):
Great fix @edmofro!

Another potential improvement is to disable editing the Country field in an access request, since the user has already selected the desired country. A counter-argument would be that since we often know the users personally, there may be cases where we want to edit an access request on their behalf. A detail any way, so no need to bother!

edmofro:
Yeah true...will leave it as is for now, but you're right that it's probably not necessary to have it editable. Another more advanced fix is that we are possibly over-eager with our initial fetch of options for an autocomplete; it could be better to only fetch after the user focuses the field, or types an initial letter. All possibilities for the future :-)

PR #3535, opened 2021-11-25
Wai 832 Add Data approval status to survey responses

EMcQ-BES:
### Changes:

- Add data `approval_status` to survey responses
- Add `approval_required` to surveys
- Update `admin_panel` to display and edit new columns
- Update analytics table to ignore unapproved data

---

PR #3558, opened 2021-12-01
WAI-1045: Permissions Update

EMcQ-BES:
### Issue #: WAI-1045

**This is the base branch for the permissions update work, more work will be added against it**

### Changes:

- Split `dataGroups` away from the `dataSource` table
- Add new table to the database model
- Update all the code that touches the `dataSource` table

---

### DIFF
--- a/packages/database/src/migrations/20211125201506-SplitDataSourceTable-modifies-schema.js
+++ b/packages/database/src/migrations/20211125201506-SplitDataSourceTable-modifies-schema.js
@@ -1,0 +1,123 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const updateForeignKey = async (
+  db,
+  tableName,
+  oldFKTable,
+  oldFKColumn,
+  newFKTable,
+  newFKColumn,
+) => {
+  await db.runSql(`
+    ALTER TABLE ${tableName}
+    ADD COLUMN ${newFKColumn} TEXT;
+
+    UPDATE ${tableName}
+    SET ${newFKColumn} = ${newFKTable}.id
+    FROM ${newFKTable}
+    INNER JOIN ${oldFKTable}
+      ON ${newFKTable}.code = ${oldFKTable}.code
+    WHERE ${oldFKTable}.id = ${oldFKColumn};
+
+    ALTER TABLE ${tableName}
+    ADD FOREIGN KEY (${newFKColumn})
+    REFERENCES ${newFKTable}(id);
+
+    ALTER TABLE ${tableName}
+    ALTER COLUMN ${newFKColumn} SET NOT NULL;
+
+    ALTER TABLE ${tableName}
+    DROP COLUMN ${oldFKColumn};
+  `);
+};
+
+exports.up = async function (db) {
+  // Create new table for data groups (renamed to events)
+  await db.runSql(`
+    CREATE TABLE event (
+      id TEXT PRIMARY KEY,
+      code TEXT NOT NULL,
+      service_type service_type NOT NULL,
+      config JSONB NOT NULL DEFAULT '{}'
+    );
+  `);
+  // Copy data groups info into new table
+  await db.runSql(`
+    INSERT INTO event (id, code, service_type, config)
+    SELECT generate_object_id(), code, service_type, config
+    FROM data_source
+    WHERE type = 'dataGroup';
+  `);
+  // Update foreign keys that refer to data groups
+  await updateForeignKey(
+    db,
+    'data_element_data_group',
+    'data_source',
+    'data_group_id',
+    'event',
+    'event_id',
+  );
+  await updateForeignKey(db, 'survey', 'data_source', 'data_source_id', 'event', 'event_id');
+
+  // Drop data groups from data source table
+  // Drop type column since it's no longer relevant
+  await db.runSql(`
+    DELETE FROM data_source
+    WHERE type = 'dataGroup';
+
+    ALTER TABLE data_source
+    DROP COLUMN type;
+  `);
+};
+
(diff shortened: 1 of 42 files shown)


rohan-bes (commented):
Hey @EMcQ-BES, would we be able to do a walk-n-talk on this one maybe? I think getting an overview on the broader direction we're going with this permissions work would be really helpful for me? (Apologies for not being more involved in the initial design phase 🙏)

Also, a more specific note, I was a bit thrown off by the naming change of `DataGroup`->`Event`. As I've understood it, an `Event` is a single instance of data of the `DataGroup` type. ie: a survey response to a survey.

But I could just be misunderstanding the change here, let's chat on it? 👍

packages/data-api/scripts/patches/1_0_0/20211207200858-AddRenameMaterializedViewLog.js
```diff
@@ -0,0 +1,155 @@
+let dbm;
```
EMcQ-BES, line 1: I've written equivalents to this patch in `mvBESFunctions.sql` directly, but I don't currently have permissions to push to the `pg-mv-refresh` repo

packages/database/src/migrations/20211125201506-SplitDataSourceTable-modifies-schema.js
```diff
@@ -0,0 +1,300 @@
...
+          answer ON answer.survey_response_id = survey_response.id
+        INNER JOIN
+          entity ON entity.id = survey_response.entity_id
+        INNER JOIN
+          survey ON survey.id = survey_response.survey_id
+        INNER JOIN
+          question ON question.id = answer.question_id
+        INNER JOIN
+          ${newTableName} ON ${newTableName}.id = question.${newTableName}_id
+        WHERE ${newTableName}.service_type = ''tupaia'' AND survey_response.outdated IS FALSE';
```
EMcQ-BES, line 235: Updating the analytics function means this may need to change depending on when the approval workflow work is merged, since that also updated this function

#3535

packages/database/src/index.js
```diff
@@ -19,3 +19,4 @@ export { DatabaseChangeChannel } from './DatabaseChangeChannel';
 export { DatabaseModel } from './DatabaseModel';
 export { DatabaseType } from './DatabaseType';
 export * from './testUtilities';
+export { getConnectionConfig } from  './getConnectionConfig';
```
EMcQ-BES, line 22: So we can use it in `patchMvRefresh.js`

packages/data-api/scripts/patchMvRefresh.js
```diff
@@ -22,19 +23,8 @@ const migrationInstance = DBMigrate.getInstance(
...
-        database: requireEnv('DB_NAME'),
         schema: requireEnv('DB_MV_USER'),
-        ssl:
-          process.env.DB_ENABLE_SSL === 'true'
-            ? {
-                // Test server cannot turn on ssl, so sets the env to disable it
-                rejectUnauthorized: false,
-              }
-            : null,
+        ...getConnectionConfig(),
```
EMcQ-BES, line 27: Load CI/CD specific env when running on codeship

rohan-bes (changes requested):
Just a couple of small changes requested for the installing of the mv-refresh patches. But aside from that it's looking awesome!!

Sorry for taking so long to get through this review 🙏 it's really nice work and thanks for going the extra mile with adding the rename function and renaming the data sources table 🎉

packages/data-api/scripts/patches/1_0_0/20211207200858-AddRenameMaterializedViewLog.js
```diff
@@ -0,0 +1,155 @@
...
+    Returns:                TEXT    The version
+    
+    ************************************************************************************************************************************
+    Copyright 2021 Beyond Essential Systems Pty Ltd
+    ***********************************************************************************************************************************/
+    DECLARE
+    
+    BEGIN
+    
+        RETURN '1_0_1';
```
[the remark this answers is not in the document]
EMcQ-BES: Gotten rid of this :+1:

packages/data-api/scripts/patchMvRefresh.js
```diff
@@ -22,19 +23,8 @@ const migrationInstance = DBMigrate.getInstance(
       defaultEnv: 'tupaia',
       tupaia: {
         driver: 'pg',
-        host: requireEnv('DB_URL'),
-        port: getEnvVarOrDefault('DB_PORT', 5432),
-        user: requireEnv('DB_MV_USER'),
-        password: requireEnv('DB_MV_PASSWORD'),
```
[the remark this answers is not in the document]
EMcQ-BES: Good catch, I'm now passing through the user/pass as overwrites to the `getConnectionConfig`

packages/database/src/getConnectionConfig.js
```diff
@@ -5,18 +5,18 @@
 
 import { requireEnv, getEnvVarOrDefault } from '@tupaia/utils';
 
-const getServerConfig = () => ({
+const getServerConfig = ({ userEnv = 'DB_USER', passEnv = 'DB_PASSWORD' }) => ({
```
rohanport, line 8: Oooh love this change, very clean and would be nicely extensible!

rohan-bes: ^The real @rohan-bes agrees with everything this imposter says 😂

rohan-bes (approved):
Looks like there's some conflicts to resolve, but I'm approving anyway 👍 

Awesome work on this, love the changes!

edmofro:
@EMcQ-BES the second to last commit failed to deploy, and I had a quick look why. The error is:
image

This seems a strange error given the change, but I reckon it might be just that the same migration attempted to run twice:
- First commit triggers migrations to run. Checks the (common) db instance, finds the latest migration hasn't run, sets it off
- Second commit triggers the same. First commit's migration hasn't finished yet because it's long running, so checks the (common) db instance, finds the latest migration hasn't run, sets it off
- Two versions of same migration run. Second one finishes first, because you disabled the trigger to make it shorter running, meaning first one is now trying to commit a transaction that does a bunch of the same stuff, so throws an error

I'm going to terminate the errored instance :+1:

PR #3566, opened 2021-12-02
RN-195: Move Tupaia database to RDS

rohan-bes:
### Issue RN-195:

Update to the deployment scripts to no longer spin up an ec2 instance for the Tupaia database instance, but instead create a new RDS instance which is cloned from a snapshot (of production by default).


packages/devops/scripts/deployment/startBackEnds.sh
```diff
@@ -24,6 +25,10 @@ for PACKAGE in ${PACKAGES[@]}; do
             # reset cwd back to `/tupaia`
             cd ${TUPAIA_DIR}
 
+            # ensure db master password is set correctly
+            echo "Setting RDS master password for ${DEPLOYMENT_NAME}"
+            source packages/data-api/.env && aws rds modify-db-instance --db-instance-identifier $DEPLOYMENT_NAME --master-user-password $DB_PG_PASSWORD
```
rohan-bes, line 30: Not quite happy with this, but keen to get some input from the reviewer.

Basically, when an RDS instance is spun up, it will have a random generated master user password. So we need to manually set it to be our postgres user password. The postgres user password is stored in the `.env` files, so we need to be in a context that has access to those. Unfortunately that means we can't do this within the lambda scripts, and instead need to do it in the startup process.

Overall I've come to accept that, and this place seems appropriate to me, however I'm still not quite happy with it. We're `source`ing the `data-api` environment variables in order to get the postgres user password which feels really ugly...

My current plan is to make this a node script in the database package, something like `set-rds-master-password`. So this would be: `yarn workspace @tupaia/database set-rds-master-password $DEPLOYMENT_NAME`

I still don't love it though, as it pollutes our generally infrastructure independent `database` package code with logic solely for RDS... Was hoping maybe the reviewer had a cleverer solution?

packages/devops/scripts/lambda/helpers/creation.py
```diff
@@ -157,3 +167,43 @@ def create_instance(
...
+        DBInstanceClass=instance_type,
+        Port=5432,
+        PubliclyAccessible=True,
+        VpcSecurityGroupIds=security_group_ids,
+        Tags=[
+            {
+                'Key': 'DeploymentName',
+                'Value': instance_name
+            },
+        ],
```
rohan-bes, line 194: Not actually sure if this is necessary, since the db identifier is just the deployment name anyway... but felt consistent with the way we handle tagging of ec2 instances 👍

packages/devops/scripts/lambda/helpers/creation.py
```diff
@@ -157,3 +167,43 @@ def create_instance(
...
+):
+    security_group_ids = get_security_group_ids_config(security_group_code, security_group_id)
+
+    snapshot_id = get_latest_db_snapshot(snapshot_name)
+    print('Starting to clone db instance from snapshot')
+    rds.restore_db_instance_from_db_snapshot(
+        DBInstanceIdentifier=instance_name,
+        DBSnapshotIdentifier=snapshot_id,
+        DBInstanceClass=instance_type,
+        Port=5432,
```
rohan-bes, line 186: Technically this should be controlled by environment variable as well... but we never actually use non-default ports in practice. Happy to leave for now and cross that hurdle when we get to it?

packages/devops/scripts/deployment/startBackEnds.sh
edmofro: There's one other option I can think of, which also has its downsides: add the db password as a secret on parameter store, and have the lambda scripts grab it from there during db setup. The downsides are it exposes the secret to another place (lambda), and means there's an unusual spot to update if we ever need to. Not terrible, but I don't think it's any better or worse than what you've already got, so might as well stick with that.

That does get me thinking about the automatic password rotation provided by AWS Secrets Manager, which integrates directly with RDS...there could be a nice way of using that, and then the tricky bit becomes setting up a process that gets the rotated password through into our environment variables. A future task though :-)

packages/devops/scripts/lambda/actions/spin_up_tupaia_deployment.py
```diff
@@ -96,18 +97,15 @@ def spin_up_tupaia_deployment(event):
...
+    # so we might as well be cloning the db instance at the same time, so long is it is available before
     # the server first tries to connect
-    deployment_type='tupaia'
-    clone_instance(
-        deployment_type,
-        clone_db_from,
+    create_db_instance_from_snapshot(
         deployment_name,
-        instance_type,
-        extra_tags=extra_tags + [{ 'Key': 'DeploymentComponent', 'Value': 'db' }],
```
edmofro, line 109: We can remove all the other `DeploymentComponent` stuff too (some should have todos on it)

packages/devops/scripts/lambda/helpers/creation.py
```diff
@@ -157,3 +167,43 @@ def create_instance(
         setup_subdomains_via_gateway(deployment_type, new_instance_object, subdomains_via_gateway, deployment_name)
 
     return new_instance_object
+
+def create_db_instance_from_snapshot(
+  instance_name,
```
edmofro, line 172:
```suggestion
  deployment_name,
  deployment_type,
```
Although we're not using RDS for anything other than Tupaia atm, I feel safer bringing in the deployment type stuff so that it's easily distinguished if we do (actual example: we may be setting up the EMIS database for Laos in RDS)

packages/devops/scripts/lambda/helpers/creation.py
edmofro: Yep and can we add the deployment type tag for consistency too?

packages/devops/scripts/lambda/helpers/creation.py
edmofro: Agreed!

packages/devops/scripts/lambda/helpers/teardown.py
```diff
@@ -57,3 +58,15 @@ def teardown_instance(instance):
       delete_gateway(deployment_type, deployment_name)
 
     terminate_instance(instance)
+
+def teardown_db_instance(
+  instance_name
+):
+    deleted_instance = rds.delete_db_instance(
+        DBInstanceIdentifier=instance_name,
```
edmofro, line 66: I reckon we should make the identifier either `deployment_type: instance_name` like on the EC2 instances, or look up the db instance by tags here (including deployment type)

edmofro (changes requested):
Nice! Glad to see this is relatively lightweight and fits in smoothly with the existing processes. As you mentioned on slack, we do still need to deal with overnight refreshes before merging, so that's the big change requested here, and otherwise it's just a few minor things. Awesome stuff getting us moved over to RDS @rohan-bes 🎉

rohan-bes:
@edmofro on the 'still need to do the overnight refresh stuff' I think I might make this a pt1 PR, and re-direct it to a feature branch? Might just be nice to keep this one fairly small 👍

edmofro:
> @edmofro on the 'still need to do the overnight refresh stuff' I think I might make this a pt1 PR, and re-direct it to a feature branch? Might just be nice to keep this one fairly small 👍

Yeah good call :+1:

packages/devops/scripts/deployment/startBackEnds.sh
rohan-bes: I like your suggestion, but I'm with ya that it's kinda on par with the suggestion I had, and I feel I'll have an easier time implementing mine.

Definitely with you on the rotating password though... which will be tricky be rather valuable to get working!

packages/devops/scripts/lambda/helpers/teardown.py
rohan-bes: So it'd be like `tupaia: rn-195_move_to_rds`? I'm down with that 👍

rohan-bes: So due to restrictions in the naming convention for RDS databases, it's gonna have to be `tupaia-<deployment-name>`
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Limits.html

packages/devops/scripts/deployment/fetchEnvVarsFromLastpass.sh
```diff
@@ -0,0 +1,15 @@
...
+
+cd ${TUPAIA_DIR}
+
+# Initialise NVM (which sets the path for access to npm, yarn etc. as well)
+. $HOME/.nvm/nvm.sh
+
+# Inject environment variables from LastPass
+LASTPASS_EMAIL=$($DIR/fetchParameterStoreValue.sh LASTPASS_EMAIL)
+LASTPASS_PASSWORD=$($DIR/fetchParameterStoreValue.sh LASTPASS_PASSWORD)
+LASTPASS_EMAIL=$LASTPASS_EMAIL LASTPASS_PASSWORD=$LASTPASS_PASSWORD yarn download-env-vars $DEPLOYMENT_NAME
```
rohan-bes, line 15: Moved this logic out of `buildDeployablePackages.sh` so that the RDS password can be set without first building the packages

packages/devops/scripts/deployment/setupGoldMaster.sh
```diff
@@ -54,4 +54,5 @@ cd /home/ubuntu
 sudo -Hu ubuntu git clone https://github.com/beyondessential/tupaia.git
 
 # build all packages once using dev to speed up future branch-specific builds
+sudo -Hu ubuntu /home/ubuntu/tupaia/packages/devops/scripts/deployment/fetchEnvVarsFromLastpass.sh gold-master-image-builder
```
rohan-bes, line 57: @edmofro do we actually need the env vars here? If we're just building the packages?

edmofro: Most of them are redundant, but the front end builds do bake in some of the env vars to the build (e.g. the back end server url), plus there's a `SKIP_PREFLIGHT_CHECK` one that's needed. It's a good point though, and it would be more secure to be more specific, and either work those out without going to lastpass, or just get the front end package env vars from lastpass.

packages/devops/scripts/lambda/actions/redeploy_tupaia_server.py
```diff
@@ -85,7 +85,6 @@ def redeploy_tupaia_server(event):
             deployment_name=get_tag(existing_instance, 'DeploymentName'),
             branch=get_tag(existing_instance, 'Branch'),
             instance_type=event.get('InstanceType', existing_instance['InstanceType']),
-            extra_tags=extra_tags + [{ 'Key': 'DeploymentComponent', 'Value': 'app-server' }], # TODO remove deployment component stuff after move to RDS
```
edmofro, line 88: We do need the extra tags in this and the next file for things other than the DeploymentComponent

edmofro (changes requested):
I like the change to set the master password outside of the build process

Just need to reinstate those extra tags and this is ready to go as the base branch :-)

packages/devops/scripts/ci/triggerRedeploy.sh
```diff
@@ -22,7 +22,7 @@ fi
 echo "At least one running deployment, triggering redeploy of any tagged with Branch ${CI_BRANCH}"
 RESPONSE_FILE=lambda_redeploy_response.json
 AWS_MAX_ATTEMPTS=1 aws lambda invoke \
-  --function-name deployment \
+  --function-name testRDS \
```
rohan-bes, line 25: Will switch these back before merging the epic

packages/devops/scripts/deployment/setupGoldMaster.sh
rohan-bes: I'll create a ticket for us to clean up this, since I feel it would be nice for the build to not rely on any `.env` vars, but will leave for now 👍

PR #3570, opened 2021-12-03
RN-226: Fixed error where location search wasn't working as intended

rohan-bes:
### Issue RN-226:

Not sure how long this has been broken hehe 😅 Basically due incorrectly treating the string input as an object we were never actually searching for values (however the `Autocomplete` component automatically searches through it's known data so it had the appearance of searching...)


kael89 (approved):
Wow that was fast!  👏

PR #3571, opened 2021-12-03
Wai 1000 bulk access requests

tcaiger:
### Changes:

- Make Meditrak-Server Edit Handler polymorphic
- Add User -> AccessRequest endpoint to Meditrak-Server
- Add BulkEditButton to AdminPanel
- Update AccessRequestPage to use bulk endpoints

---

### Screenshots:
See Issue


packages/admin-panel/src/pages/resources/SocialFeedPage.js
```diff
@@ -115,9 +115,7 @@ export const SocialFeedPage = ({ getHeaderEl }) => (
     columns={SOCIAL_FEED_COLUMNS}
     editConfig={EDIT_CONFIG}
     createConfig={CREATE_CONFIG}
-    onProcessDataForSave={data => {
-      data.type = 'markdown'; // eslint-disable-line no-param-reassign
-    }}
+    onProcessDataForSave={data => ({ ...data, type: 'markdown' })}
```
IgorNadj, line 118: 👍

packages/meditrak-server/src/apiV2/EditHandler/EditHandler.js
```diff
@@ -18,11 +18,21 @@ export class EditHandler extends CRUDHandler {
 
   async handleRequest() {
     await this.validate();
-    await this.editRecord();
+
```
IgorNadj, line 21: I like the simplicity of this, however I don't think it works. 

The two main problems I see are 
1) a failed record update half way through the loop will mean some are persisted and some are not, leaving the database in an unintended state. 

And 2) the current edit handler subclass implementations, e.g. EditUserAccounts are under the assumption there is one this.recordId and one this.updatedFields object, which there is, but there can also be multiple if you just POST the endpoint with an array. There _will_ be an error thrown, but in a strange place, probably at the database level when it tries to update the row but passes in an array instead of an object. Or worse, it might allow it, and do something strange like take the last item in the array and use that. I'm not sure, but the behaviour being unknown is pretty concerning.

At this point I think there are two options:
a) Refactor EditHandler to work with multiple records only, e.g. remove this.recordId, replace with this.recordIds which could be an array with 1 item.
b) Create a new MultiEditHandler which handles multiple edits, and removes this.recordId for this.recordIds, has editRecords instead of editRecord etc.

Personally I think (a) is a tonne of work, if you want feel free to do this. I think (b) would be the quicker solution.

tcaiger: Very good points. Thanks for picking up those.

In regards to 1. I found that there is already a way of handling this using the transactingModels concept so I wrapped that around the set of updates, tested it and it works well.

In regards to 2. I think b is a good option. I was keen to construct it so that the same code could be run for single and bulk updates so I didn't replace this.recordId with this.recordIds. Also when Edwin and I kicked off the card we discussed using the term Bulk and also having the endpoint handle the data as polymorphic so I kept those concepts in place as they seemed more about preference rather than major concerns. In saying that I could be wrong about that.

IgorNadj: >In regards to 1. I found that there is already a way of handling this using the transactingModels concept so I wrapped that around the set of updates, tested it and it works well.
Perfect

>In regards to 2. ... having the endpoint handle the data as polymorphic 
Yeah I think that's the refactor of EditHandler I was mentioning. The initial change did make it kind of polymorphic but to make it fully polymorphic it would mean refactoring to get rid of recordId, and replace it with recordIds array etc.

IgorNadj (approved):
Nice one!

PR #3572, opened 2021-12-03
RN-225: Fix Viz-Builder UI hanging when fetching a large number of rows 

rohan-bes:
### Changes:
- Added `rowLimit` prop to DataTable
- Limited viz-builder to 100 rows


tcaiger (approved):
I was initially going to suggest moving this code into the VizBuilder itself because it is a bit specific but it probably could be useful as a quick fix in other use cases of the data table. The ideal solution to this would be pagination in the data table, so maybe we should make a card to add it?

rohan-bes:
@tcaiger good call! Have created: https://linear.app/bes/issue/RN-228/add-pagination-to-datatable-ui-component

nick-beyondessential (approved):
LGTM!

PR #3578, opened 2021-12-06
Refactor tests for survey response import

kael89:
[No issue]

Not strictly required but in preparation for PHX-152, I felt that it would be worth refactoring the survey response import tests. Reasons:
* I found it a bit hard to get an overview of the existing test cases
* These tests use some older patterns. Refactoring to use newer setup and test data patterns should make them more maintainable

**Changes overview**
1. Used semantic test ids (eg `duplicate_quest_id1_test`) instead of the existing, more realistic ids (eg `fdfcc42a44705c032a8_test`)
2. Removed redundant test data (multiple responses testing the same logic, unnecessary models eg assign a `Public` permission group to a survey when permissions are not tested, etc)
3. Extracted validation tests in their own file and parameterised them
4. Updated the import spreadsheets according to the new changes + restructured them a bit


IgorNadj (approved):
Thanks for cleaning that up!

PR #3579, opened 2021-12-07
RN-238: Fix error when trying to save a large viz

rohan-bes:
### Issue RN-238:

Dropped notification trigger from report table to allow saving large reports


nick-beyondessential (approved):
LGTM!

PR #3585, opened 2021-12-09
WAI-1070: Migrate database when meditrak-server starts via pm2

edmofro:
### Issue #:
WAI-1070

### Changes:
- Migrations are run whenever meditrak-server is started via pm2, meaning they'll be run when the server comes back online after being off overnight, and if there's been a clone from production in the meantime, any migrations on that branch will be applied


packages/database/.babelrc.js
```diff
@@ -0,0 +1,38 @@
+const env = {
+  test: {
+    plugins: ['istanbul'],
+  },
+};
+
+const includedDateOffset = 90 * 24 * 60 * 60 * 1000; // include migrations up to 90 days old
```
edmofro, line 7: Because meditrak-server needs to run the babelified migrations, we needed to take them off the ignore list. However, there are so many that it makes the compilation takes 3x the time. This js based babel config has a custom ignore function, that ignores any migrations which are older than 90 days.

packages/database/src/getDbMigrator.js
```diff
@@ -0,0 +1,54 @@
+/**
```
edmofro, line 1: Most of this can be diff'd locally against `migrate.js`. The only differences are the different callback signature when running it from the app rather than cli, and the change of migration directory (which is specified on the command line when run via yarn anyway)

packages/database/.babelrc.js
kael89: Clever!!

packages/database/package.json
```diff
@@ -11,7 +11,7 @@
   "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
   "main": "dist/index.js",
   "scripts": {
-    "build": "babel src --out-dir dist --source-maps --ignore \"**/migrations\",\"src/tests/**\" --config-file \"../../babel.config.json\" && tsc",
+    "build": "babel src --out-dir dist --source-maps --config-file \"../../babel.config.json\" && tsc",
```
kael89, line 14: I think we should still ignore `src/tests` when building

packages/database/.babelrc.js
```diff
@@ -0,0 +1,38 @@
...
+const includedMigrationsDate = new Date().setTime(Date.now() - includedDateOffset);
+const checkMigrationWithinIncludedRange = function (migrationName) {
+  const yearPart = migrationName.substring(0, 4);
+  const monthPart = migrationName.substring(4, 6);
+  const dayPart = migrationName.substring(6, 8);
+  const migrationDate = new Date(`${yearPart}-${monthPart}-${dayPart}`);
+  return migrationDate < includedMigrationsDate;
```
kael89, line 14: This works but the function name is a bit misleading, I suggest either renaming it to `checkMigrationOutsideIncludedRange`, or reversing the operator to `>` and adding a `!` where it's called.

packages/database/.babelrc.js
```diff
@@ -0,0 +1,38 @@
...
+      const filename = filepathComponents[filepathComponents.length - 1];
+      return checkMigrationWithinIncludedRange(filename);
+    }
+    const parentDirectory =
+      filepathComponents.length > 3 ? filepathComponents[filepathComponents.length - 3] : null;
+    if (parentDirectory === 'migrationData') {
+      // Some migration data is broken into separate subfiles within a folder named with the
+      // same name as the migration
+      return checkMigrationWithinIncludedRange(directory);
+    }
```
kael89, line 33: Optional suggestion: using a stack can make the logic a bit simpler, since you can deconstruct the components in one go without checking for their length:

```js
const components = filepath.split('/');
const fileName = components.pop(); // will be always on top
const directory = components.pop(); // may be `undefined`
const parentDirectory = components.pop(); // may be `undefined`
```

packages/database/.babelrc.js
```diff
@@ -0,0 +1,38 @@
...
+    const filepathComponents = filepath.split('/');
+    const directory =
+      filepathComponents.length > 2 ? filepathComponents[filepathComponents.length - 2] : null;
+    if (directory === 'migrations' || directory === 'migrationData') {
+      const filename = filepathComponents[filepathComponents.length - 1];
+      return checkMigrationWithinIncludedRange(filename);
+    }
+    const parentDirectory =
+      filepathComponents.length > 3 ? filepathComponents[filepathComponents.length - 3] : null;
+    if (parentDirectory === 'migrationData') {
```
kael89, line 29: Hopefully devs will not nest files more than one level in `migrationData`!

kael89 (changes requested):
Looks great @edmofro ! One rename change requested

packages/database/package.json
edmofro: They should get ignored by the dynamic rule within the .babelrc.js :-)

packages/database/.babelrc.js
edmofro: Oh yep good call, will rename

packages/database/.babelrc.js
edmofro: True, that's nice!

packages/database/.babelrc.js
edmofro: Ha, yes hopefully!

packages/database/package.json
kael89: Oh yes you are right, I read this file first 👍

packages/database/.babelrc.js
edmofro: Ah, but there's no `/` in the string...I remember getting to this point when I looked at it initially

kael89: I think that's still fine?

```js
'fileName'.split('/') // ['test']
'/fileName'.split('/') // ['', 'fileName']

'directory/fileName'.split('/') // ['directory', 'fileName']
'/directory/fileName'.split('/') // ['', 'directory', 'fileName']
```

Unless if you mean something else

edmofro: Oh yes this is a different spot! I was looking at the date string. Shouldn't try to multi-task while I'm in a call. I'll come back to this at the end of my meetings for the day

kael89 (approved):
Approved! You may have noticed it already but the build fails:

image

edmofro:
@kael89 I actually hadn't noticed the failing test, thanks for pointing that out. Fixed now, though it did require a change of structure in the babel config file, so will need a quick re-review

kael89 (approved):
Looks good!

edmofro:
Made one more change based on testing @kael89

kael89 (approved):
Looks good

PR #3594, opened 2021-12-13
Wai 1047 Add permissions column to data element table

EMcQ-BES:
### Changes:

- Migration to calculate permission groups on data elements and add them in a new column


packages/database/src/migrations/20211201024558-AddPermissionGroupsToDataSources-modifies-schema.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  // Update to wildcard if the value is in public permission group
+  // or if we haven't found any permissions so far
+  const publicPermissionId = await nameToId(db, 'permission_group', 'Public');
+  await db.runSql(`
+    UPDATE data_element
+      SET permission_groups = '{"*"}'
+      WHERE '${publicPermissionId}' = ANY(permission_groups)
+      OR permission_groups = '{}'
```
EMcQ-BES, line 101: At this point the only data elements with no permissions against them should be the `weather` service data elements and a couple of edge case indicators such as the `target` indicators in LESMIS (where the formula is just a value)

PR #3595, opened 2021-12-13
Fix bar chart legend overlap

tcaiger:
### Changes:

- Trigger re-render on bar charts so that the layout is calculated correctly everytime

---


edmofro (approved):
lgtm, thanks for getting that sorted

PR #3597, opened 2021-12-13
HotFix 1 for Release 2021-50

tcaiger:
### Changes:

- Fix logic in getIsUsingDots for Matrix Tables

---


kael89 (approved):
Not totally familiar with these props but fix seems legit, thanks for addressing this!

PR #3603, opened 2021-12-14
WAI-1083: Fix preaggregations not running overnight

edmofro:
### Issue #:
WAI-1083

### Changes:
- Allow cron job to write to the logs folder, which was write protected and preventing the script from running
- Fix date handling in cron job output pipeline, which was being evaluated early 
- Store cron job under ubuntu user so it is easier to find, and not running as root
- Use same PATH as regular bash scripts so `yarn` is available, etc.


kael89 (approved):
Unix level 💯 - well done!! 👏 😎 🌈 🎸

edmofro:
Haha thanks! From the number of commits you can see it took some trial and error, but got there in the end 😎

PR #3604, opened 2021-12-14
Wai 1048 Update Data Source CRUD handlers

EMcQ-BES:
### Changes:

- Add new CRUD handlers for Data Element and Data Group
- Update Question edit handler to respect data element rules


packages/meditrak-server/src/apiV2/dataElements/DeleteDataElements.js
```diff
@@ -0,0 +1,29 @@
...
+
+export class DeleteDataElements extends DeleteHandler {
+  async assertUserHasAccess() {
+    // User has access to all permission groups for the data_element, plus tupaia admin panel
+    // Or is a BES admin
+    const dataElementPermissionChecker = accessPolicy =>
+      assertDataElementPermissions(accessPolicy, this.models, this.recordId);
+
+    await this.assertPermissions(
+      assertAnyPermissions([
```
IgorNadj, line 23: Probably out of scope for this PR but I suggest moving common permission checks to the base classes. 

A call with "BES Admin" always passes. Else, a call without "admin panel" always fails.

packages/meditrak-server/src/apiV2/dataElements/assertDataElementPermissions.js
```diff
@@ -0,0 +1,45 @@
...
+  if (hasBESAdminAccess(accessPolicy)) {
+    return criteria;
+  }
+  const dbConditions = { ...criteria };
+  const userPermissions = await getPermissionIdListWithWildcard(accessPolicy, models);
+
+  // Permission groups on the data element overlap with our permission groups
+  // Wildcard is added to our list so it will be included
+  dbConditions.permission_groups = {
+    comparator: '&&',
```
IgorNadj, line 41: Oh interesting, I didn't know && was part of SQL. Maybe use AND for consistency?

packages/meditrak-server/src/apiV2/dataElements/assertDataElementPermissions.js
```diff
@@ -0,0 +1,45 @@
...
+
+import { hasBESAdminAccess } from '../../permissions';
+
+const getPermissionIdListWithWildcard = async (accessPolicy, models) => {
+  // Get the users permission groups as a list of ids
+  const userPermissionGroupNames = accessPolicy.getPermissionGroups();
+  const userPermissionGroups = await models.permissionGroup.find({
+    name: userPermissionGroupNames,
+  });
+  return ['*', ...userPermissionGroups.map(permission => permission.id)];
```
IgorNadj, line 14: What is the purpose of wildcard?

IgorNadj (changes requested):
Good change! Couple questions and suggestions

IgorNadj:
Also looks like we have a few tests for other things like map overlays/surveys/questions etc. that have test cases for permissions. Would be good to have some tests for CRUD of data elements and data groups, including a test for these new finer grained permission checks, since it's such a core behaviour.

packages/meditrak-server/src/apiV2/dataElements/DeleteDataElements.js
EMcQ-BES: There's a [ticket](https://linear.app/bes/issue/WAITP-597/obfuscate-permissions-error-messages) for cleaning up all these asserts, so in the meantime I'm leaving them consistent

packages/meditrak-server/src/apiV2/dataElements/EditDataElements.js
```diff
@@ -0,0 +1,38 @@
...
+    await this.assertPermissions(
+      assertAnyPermissions([
+        assertBESAdminAccess,
+        assertAllPermissions([assertAdminPanelAccess, dataElementPermissionChecker]),
+      ]),
+    );
+  }
+
+  async editRecord() {
+    const dataElementEditPermissionChecker = accessPolicy =>
```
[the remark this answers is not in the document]
EMcQ-BES: It absolutely doesn't need to you're right

packages/meditrak-server/src/apiV2/dataElements/assertDataElementPermissions.js
EMcQ-BES: Added a comment

packages/meditrak-server/src/apiV2/dataElements/assertDataElementPermissions.js
EMcQ-BES: We wanted a sort of greater scope idea of a data element that anyone can access, which is not alway necessarily the same concept as `Public`
Edwin had some examples but honestly it's been a month since I wrote this and I don't 100% remember the justification

EMcQ-BES:
> Also looks like we have a few tests for other things like map overlays/surveys/questions etc. that have test cases for permissions. Would be good to have some tests for CRUD of data elements and data groups, including a test for these new finer grained permission checks, since it's such a core behaviour.

I've done all the fixups against https://github.com/beyondessential/tupaia/pull/3608 just to avoid too much merging up the chain, I've added unit tests for the CRUD handlers in there as well

packages/meditrak-server/src/apiV2/dataElements/assertDataElementPermissions.js
IgorNadj: Got it! Thanks

IgorNadj:
LGTM.

Changes in other PR address comments here

IgorNadj (approved):

PR #3608, opened 2021-12-15
Wai-1049 Check data source permissions when pulling via data broker

EMcQ-BES:
### Changes:

- Pass in an access policy when creating a data broker instance
- Check permissions before pulling data
- Update some unit tests to keep working

### DIFF
--- a/packages/data-broker/src/DataBroker.js
+++ b/packages/data-broker/src/DataBroker.js
@@ -33,7 +33,9 @@ export class DataBroker {
   constructor(context = {}) {
     this.context = context;
     this.models = getModelRegistry();
-    this.userPermissions = getPermissionIdListWithWildcard(this.context.accessPolicy, this.models);
+    this.userPermissions = this.context.accessPolicy
+      ? getPermissionIdListWithWildcard(this.context.accessPolicy, this.models)
+      : ['*'];
     this.resultMergers = {
       [this.getDataSourceTypes().DATA_ELEMENT]: this.mergeAnalytics,
       [this.getDataSourceTypes().DATA_GROUP]: this.mergeEvents,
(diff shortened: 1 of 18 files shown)


packages/data-broker/src/DataBroker.js
```diff
@@ -123,6 +177,11 @@ export class DataBroker {
 
   pullForServiceAndType = async (dataSources, options, type) => {
     const { service_type: serviceType } = dataSources[0];
+    const permissionChecker = this.permissionCheckers[type];
+    const userHasPermission = await permissionChecker(dataSources);
+    if (!userHasPermission) {
+      throw new Error(`Missing permissions for some or all ${type}s for ${serviceType}`);
```
IgorNadj, line 183: I suspect this error message will puzzle a lot of devs/users 😆 
If possible lets list the data elements / data groups the user doesn't have permission for.

packages/data-broker/src/DataBroker.js
```diff
@@ -34,6 +46,21 @@ export class DataBroker {
       [this.getDataSourceTypes().DATA_GROUP]: this.fetchFromDataGroupTable,
       [this.getDataSourceTypes().SYNC_GROUP]: this.fetchFromSyncGroupTable,
     };
+    this.permissionCheckers = {
```
IgorNadj, line 49: It's not entirely obvious why data broker is doing permission checks. 

You could think that it should be done before or after the data broker. 
- Before: different callers may use data broker differently. In some cases it's for fetching data to show to a user, in other cases it may be a machine user which needs access to all data.
- After: DHIS already has permission checks, why do it twice?

I'm guessing the reasoning is that we want the services to expose all data, and for data broker to be user-aware and only expose the data the user is allowed to see. If so, this would be really good to comment here somewhere for the next devs coming across this.

IgorNadj (changes requested):
Just a couple little things, otherwise looks great!

packages/data-broker/src/DataBroker.js
EMcQ-BES: Added some better error messaging

IgorNadj: Beautiful, thank you!

packages/data-broker/src/DataBroker.js
IgorNadj: I think it makes sense if you think about it more. Happy with the small comment.

IgorNadj (approved):

IgorNadj:
LGTM!

PR #3612, opened 2022-01-06
NOT-903 fix dataset autocompletion

avaek:
Issue: https://linear.app/bes/issue/NOT-903/dhis2-data-sets-no-longer-completing-automatically

Basically this PR makes it so that data submitted against data sets is "completed" automatically when the data syncs to DHIS2
- "Completed" basically just means added to the `completedatasetregistration` table

For context - this used to happen automatically but stopped working in mid December (coincidentally right after we patched DHIS2 as mentioned in the issue linked above). This PR seems to make it work again


IgorNadj (approved):
LGTM

PR #3614, opened 2022-01-07
Rn 317 allow creating a map overlay group

billli0:
### Issue #:
RN-317

### Changes:
Basically just copy and paste from PR https://github.com/beyondessential/tupaia/pull/3048. Extra work is to set map overlay group code as a unique key in db.

### DIFF
--- a/packages/database/src/migrations/20220107061813-SetMapOverlayGroupCodeAsUniqueKey-modifies-schema.js
+++ b/packages/database/src/migrations/20220107061813-SetMapOverlayGroupCodeAsUniqueKey-modifies-schema.js
@@ -1,0 +1,37 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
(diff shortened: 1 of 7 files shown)


packages/database/src/migrations/20220107061813-SetMapOverlayGroupCodeAsUniqueKey-modifies-schema.js
```diff
@@ -0,0 +1,37 @@
...
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  // 1. Delete duplicated map overlay group with code 'Root'.
+  //    Have checked it doesn't have any children in overlay relation table.
+  // 2. Set map overlay group code as unique key.
```
tcaiger, line 20: Good work on removing this, I had no idea that there were 2 root mapOverlayGroups. I double checked and you are correct that the one with Id Root has no children so good riddance.

tcaiger (approved):
Looks good to me

packages/database/src/migrations/20220107061813-SetMapOverlayGroupCodeAsUniqueKey-modifies-schema.js
billli0: Thanks @tcaiger, and yes it's a bit weird. Hope that doesn't break anything.

PR #3622, opened 2022-01-16
Unable to edit survey code on certain surveys

billli0:
### Issue #:
RN-253

### Changes:
Reason: `service_type` plays a role in survey update validation. While we only want to change the survey code in admin panel, it doesn't have `service_type` in updated field from front end.

Change: Add `service_type` from `dataGroup` if `service_type` is not set.


nick-beyondessential (approved):
LGTM!

PR #3624, opened 2022-01-17
Display latest overlay date in a format

billli0:
### Issue #:
RN-250

### Changes:
`moment.js` is not able to import date with format 'yyyymm', thus we need to use our customised translate function `periodToMoment` in front end.


rohan-bes (approved):
Nice simple fix!

PR #3629, opened 2022-01-18
RN-239: Set up overnight clone of production DB for RDS

rohan-bes:
### Issue RN-239:

Added `redeploy_tupaia_database` lambda action which is responsible for running each night and re-creating the RDS instance with the latest snapshot.

### Changes:
- Added the postgres master password to AWS Secrets Manager which allows us to easily access it via lambda scripts. This does have a major issue of no longer having a single point of truth for the password, however it was proving too unruly to try and pull the password from lastpass. What's more, we really ought to move to using secrets manager for database credentials, so this can be the first step

### Merge tasks:
- [ ] Change event bridge configs to use new lambda function names


packages/devops/scripts/lambda/resources/startup.sh
```diff
@@ -60,7 +60,6 @@ git reset --hard origin/${BRANCH_TO_USE}
 # Deploy each package, including injecting environment variables from LastPass
 sudo -Hu ubuntu $DEPLOYMENT_SCRIPTS/fetchEnvVarsFromLastpass.sh $DEPLOYMENT_NAME |& while IFS= read -r line; do printf '\%s \%s\n' "$(date)" "$line"; done  >> $LOGS_DIR/deployment_log.txt
 sudo -Hu ubuntu $DEPLOYMENT_SCRIPTS/buildDeployablePackages.sh $DEPLOYMENT_NAME |& while IFS= read -r line; do printf '\%s \%s\n' "$(date)" "$line"; done  >> $LOGS_DIR/deployment_log.txt
-sudo -Hu ubuntu $DEPLOYMENT_SCRIPTS/setRDSMasterPassword.sh tupaia-$DEPLOYMENT_NAME |& while IFS= read -r line; do printf '\%s \%s\n' "$(date)" "$line"; done  >> $LOGS_DIR/deployment_log.txt
```
rohan-bes, line 63: No longer needed as we now handle this via the initial `spin_up_tupaia_deployment` lambda script

packages/devops/scripts/lambda/actions/spin_up_tupaia_database.py
```diff
@@ -0,0 +1,37 @@
...
+#   "Action": "spin_up_tupaia_database",
+#   "User": "rohan",
+#   "DeploymentName": "rn-195-epic",
+#   "CloneDbFrom": "dev",
+#   "DbInstanceType": "db.t4g.medium"
+# }
+
+from helpers.creation import create_db_instance_from_snapshot
+
+def spin_up_tupaia_database(event):
```
rohan-bes, line 16: Strictly speaking we don't need this action as part of the deployment process, however it's proven quite useful to have around during development.

packages/devops/scripts/lambda/actions/redeploy_tupaia_database.py
```diff
@@ -0,0 +1,48 @@
...
+    db_id = 'tupaia-' + deployment_name
+    db_instance = get_db_instance(db_id)
+    # get manual input parameters, or default for any not provided
+    db_instance_type = db_instance['DBInstanceClass']
+    security_group_id = db_instance['VpcSecurityGroups'][0]['VpcSecurityGroupId']
+    clone_db_from = next(filter(lambda item: item['Key'] == 'ClonedFrom', db_instance['TagList']))['Value']
+
+    # rename then delete existing db
+    teardown_db_instance(deployment_name, 'tupaia')
+    loop.run_until_complete(wait_for_db_instance(db_id, 'deleted'))
```
rohan-bes, line 36: This does create a bit of a database downtime, but given it happens overnight and our feature instance should be offline then anyway it ought to be fine. The alternative is to spin up the new one under a new name and once it's up drop the old one and rename, but the timing for that was proving tricky within the 15 min timeout window for lambdas.

packages/devops/scripts/lambda/actions/redeploy_tupaia_database.py
```diff
@@ -0,0 +1,48 @@
...
+    if 'DeploymentName' not in event:
+        raise Exception('You must include the key "DeploymentName" in the lambda config, e.g. "dev".')
+    deployment_name = event['DeploymentName']
+
+    db_id = 'tupaia-' + deployment_name
+    db_instance = get_db_instance(db_id)
+    # get manual input parameters, or default for any not provided
+    db_instance_type = db_instance['DBInstanceClass']
+    security_group_id = db_instance['VpcSecurityGroups'][0]['VpcSecurityGroupId']
+    clone_db_from = next(filter(lambda item: item['Key'] == 'ClonedFrom', db_instance['TagList']))['Value']
```
edmofro, line 32: Clever functional programming!

packages/devops/scripts/lambda/actions/spin_up_tupaia_database.py
```diff
@@ -0,0 +1,37 @@
...
+    deployment_name = event['DeploymentName']
+
+    # get manual input parameters, or default for any not provided
+    db_instance_type = event.get('DbInstanceType', 'db.t4g.medium')
+    security_group_code = event.get('SecurityGroupCode', 'tupaia-dev-sg') # Use security group tagged with code
+    clone_db_from = event.get('CloneDbFrom', 'production') # Use volume snapshot tagged with deployment name
+
+    # create db instance from a snapshot
+    # do this after the server has started because it will take a while to populate the db from the snapshot, 
+    # so we might as well be cloning the db instance at the same time, so long is it is available before
```
edmofro, line 29:
```suggestion
    # so we might as well be cloning the db instance at the same time as the app server's build scripts are running, so long is it is available before
```

packages/devops/scripts/lambda/helpers/secrets.py
```diff
@@ -0,0 +1,34 @@
+# Use this code snippet in your app.
+# If you need more information about configurations or implementing the sample code, visit the AWS docs:   
+# https://aws.amazon.com/developers/getting-started/python/
```
edmofro, line 3:
```suggestion
```

edmofro (approved):
LGTM! Nice work. A couple of little comments on comments, but nothing substantive to suggest

packages/devops/scripts/lambda/helpers/secrets.py
rohan-bes: Haha I thought I might keep this so as to suggest that this really is the basic aws suggestion! I'll clean it 👍

rohan-bes:
@edmofro followed up on your suggestions, but also included a change to stop using the term 'instance' without either stating if it's a db instance, or an app server instance. Previous usages of 'instance' have been replaced by 'server' where appropriate.

I did this because 'instance' is the term used by AWS to describe both 'ec2' instances, and 'rds' instances, and it was rather confusing. So now we have:
- database: refers to RDS database
- server: refers to ec2 application server
- deployment: refers to the whole deployment (server AND database)

edmofro (approved):
Good call! We'll just have to remember to modify the cron jobs in EventBridge that call those actions with changed names

rohan-bes:
> Good call! We'll just have to remember to modify the cron jobs in EventBridge that call those actions with changed names

Great point! I'll add it to the merge tasks.

Sorry I think the last push is the final one! I basically scrapped `redploy_tupaia_database` in favour of `refresh_cloned_databases` which is very similar to the `refresh_cloned_servers` we use for the overnight clone currently. Then had to re-write a bunch of the code for it to run successfully in parallel. 🙏

PR #3630, opened 2022-01-19
Use 'rbd' to achieve drag and drop functionality

billli0:
### Issue #:
RN-153 Sort functionality 

### Changes:
Transfer lists in material ui cannot support drag and drop functionality. Found anothe tool [react-beautiful-dnd](https://www.npmjs.com/package/react-beautiful-dnd) that is very handy to achieve this requirement.

### DIFF
--- a/packages/ui-components/src/components/DataLibrary/DataLibrary.js
+++ b/packages/ui-components/src/components/DataLibrary/DataLibrary.js
@@ -7,6 +7,7 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import MuiContainer from '@material-ui/core/Container';
 import { CreateNewFolder } from '@material-ui/icons';
+import { DragDropContext } from 'react-beautiful-dnd';
 import { FlexColumn } from '../Layout';
 import { InputField } from './InputField';
 import { ResultsList } from './ResultsList';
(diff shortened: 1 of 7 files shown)


packages/ui-components/src/components/DataLibrary/DataLibrary.js
```diff
@@ -158,7 +178,7 @@ DataLibrary.defaultProps = {
 
 DataLibrary.propTypes = {
   // options: either an array of options if single data type or map of dataType -> options
-  options: PropTypes.oneOf([PropTypes.object, PropTypes.arrayOf(optionPropType)]).isRequired,
+  options: PropTypes.oneOfType([PropTypes.object, PropTypes.arrayOf(optionPropType)]).isRequired,
```
billli0, line 181: oneOf is passing an exact value from an array, but not for type. https://jaketrent.com/post/react-oneof-vs-oneoftype

packages/ui-components/src/components/DataLibrary/ResultsList.js
```diff
@@ -44,12 +44,16 @@ export const ResultsList = ({
...
+                  key={option.code}
+                  option={option}
+                  onSelect={onSelect}
+                />
               );
             } else {
               return (
                 <SelectableOption
-                  key={index}
+                  key={option.code}
```
billli0, line 56: When you use index of an array as a key, React will optimize and not render as expected. https://stackoverflow.com/questions/46735483/error-do-not-use-array-index-in-keys

packages/ui-components/src/components/DataLibrary/SelectedDataList.js
```diff
@@ -6,15 +6,21 @@
...
+import { Droppable } from 'react-beautiful-dnd';
 
 const StyledSelectedDataList = styled.div`
   padding-top: 10px;
 `;
 
 export const SelectedDataList = ({ value, onRemove }) => (
-  <StyledSelectedDataList>
-    {value.map((option, index) => {
-      return <SelectedDataCard key={index} option={option} onRemove={onRemove} />;
```
billli0, line 17: Same as above.

IgorNadj (approved):
LGTM

PR #3631, opened 2022-01-19
Rn 186 toggle back to json

billli0:
### Issue #:
RN-186

### Screenshots:
https://linear.app/bes/issue/RN-186#comment-fe55f5f7

### DIFF
--- a/packages/admin-panel/src/VizBuilderApp/components/Panel.js
+++ b/packages/admin-panel/src/VizBuilderApp/components/Panel.js
@@ -11,7 +11,8 @@ import { FlexColumn, FlexSpaceBetween } from '@tupaia/ui-components';
 import { TabPanel } from './TabPanel';
 import { JsonEditor } from './JsonEditor';
 import { PlayButton } from './PlayButton';
-import { useVizConfig, useVizConfigError } from '../context';
+import { JsonToggleButton } from './JsonToggleButton';
+import { useEditPanel, useVizConfig, useVizConfigError } from '../context';
 import { DataElementDataLibrary } from './DataLibrary';
 
 const Container = styled(FlexColumn)`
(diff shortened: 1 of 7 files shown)


packages/admin-panel/src/VizBuilderApp/views/Main.js
```diff
@@ -88,7 +93,9 @@ export const Main = () => {
         <Toolbar />
         <Container maxWidth="xl">
           <PreviewDataProvider>
-            <Panel />
+            <JsonEditorProvider>
```
billli0, line 96: Added a new context here, in case we need more global state for JsonEditor in the future.

packages/admin-panel/src/VizBuilderApp/components/Panel.js
```diff
@@ -73,7 +74,9 @@ const PanelTabPanel = styled.div`
 
 export const Panel = () => {
   const { hasDataError, setDataError } = useVizConfigError();
+  const { jsonToggleEnabled } = useJsonEditor();
   const [tab, setTab] = useState(0);
+  // const [jsonToggle, setJsonToggle] = useState(false);
```
IgorNadj, line 79: Commented out code

packages/admin-panel/src/VizBuilderApp/context/JsonEditor.js
```diff
@@ -0,0 +1,25 @@
...
+  const [jsonToggleEnabled, setJsonToggleEnabled] = useState(false);
+
+  return (
+    <JsonEditorContext.Provider value={{ jsonToggleEnabled, setJsonToggleEnabled }}>
+      {children}
+    </JsonEditorContext.Provider>
+  );
+};
+
+export const useJsonEditor = () => useContext(JsonEditorContext);
```
IgorNadj, line 21: I don't love useJsonEditor as the name here, as all this is is the toggle. It is unrelated to the actual json editor. What I mean is, if the actual json editor was exposed as a context, you could do stuff like `{ currentValue, isValid } = useJsonEditor()`. What you actually have here is more like `{ jsonToggleEnabled } = useJsonToggle()`.

IgorNadj (approved):
LGTM. Couple suggestions, pre-approving.

packages/admin-panel/src/VizBuilderApp/context/JsonEditor.js
billli0: That makes sense. in case we need more global state for the tab panel in the future, I will change the name to useTabPanel().

PR #3642, opened 2022-01-21
RN-205: OrgUnit URL error for test data in Viz Builder

chris-pollard:
### Issue #:
RN-205: OrgUnit URL error for test data in Viz Builder
### Changes:
Set default organisationUnits for testData reportBuild.


packages/report-server/src/routes/TestReportRoute.ts
```diff
@@ -29,7 +29,7 @@ export class TestReportRoute extends Route<TestReportRequest> {
   async buildResponse() {
     const { query, body } = this.req;
     const { testData, testConfig, ...restOfBody } = body;
-    const { hierarchy = 'explore', organisationUnitCodes, ...restOfQuery } = query;
+    const { hierarchy = 'explore', organisationUnitCodes = 'explore', ...restOfQuery } = query;
```
rohan-bes, line 32: While this is a clever way to resolve this issue, I feel perhaps it's a tad misleading. In actual fact, there's no need to build the query if test data is provided, as we'll just return the test data anyway right?

If that's the case, I feel we could address this issue a little closer to the source? In ReportBuilder, we can change the logic of the `fetch` function to only build the query when needed. 
```
const fetch = this.testData
      ? () => ({ results: this.testData } as FetchResponse)
      : async () => {
          const builtQuery = await new QueryBuilder(this.reqContext, this.config, query).build();
          return buildFetch(this.config?.fetch)(aggregator, builtQuery);
        };
```
^this is a bit ugly and I think has some ts errors so would need to be cleaned up, but hopefully it makes sense. Lemme know what ya think?

rohan-bes (changes requested):

packages/report-server/src/routes/TestReportRoute.ts
chris-pollard: Nice! Yeah, I thought it was too good to be true!

edmofro:
Does this need re-review @chris-pollard, or is it no longer required?

chris-pollard:
@edmofro this one was not a priority, but still needs to be finished.

chris-pollard:
Superseded PR>

PR #3648, opened 2022-01-25
rn-344 remove tamanu server from meditrak leaderboard

avaek:
Issue: https://linear.app/bes/issue/RN-344/remove-tamanu-server-user-from-meditrak-leaderboard

This (in theory) removes Tamanu Server from the MediTrak leaderboard. I haven't tested it locally but I feel confident it will work


rohan-bes (approved):
No coconuts for you!

PR #3652, opened 2022-01-27
MAUI-367: Fully qualify URL for latest data downloads

edmofro:
### Issue #:
MAUI-367

### Changes:
[This hot fix](https://github.com/beyondessential/tupaia-backlog/issues/3208) worked for standard raw data downloads, but not for latest data downloads. This PR brings the same strategy of fetching against the correct URL for web-config-server to latest data downloads.

### DIFF
--- a/packages/web-config-server/README.md
+++ b/packages/web-config-server/README.md
@@ -74,27 +74,6 @@ API connect for Tupaia App server
 
 Deals with oauth2 authentication and request from dhis2
 
-## Exporting charts and Lambda
-
-In order to export charts from a local instance of the config server you'll need to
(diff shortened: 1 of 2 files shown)


packages/web-config-server/README.md
```diff
@@ -74,27 +74,6 @@ API connect for Tupaia App server
 
 Deals with oauth2 authentication and request from dhis2
 
-## Exporting charts and Lambda
```
edmofro, line 77: Unrelated cleanup, this is outdated documentation about a system that we no longer use

tcaiger (approved):
Looks good for the specific use case as long as the value is either a fully qualified url or a resource endpoint. Obviously there are values that won't work but since we control the value it should be ok.

PR #3653, opened 2022-01-27
WAITP-990: add question permission filter

tcaiger:
### Changes:

- add filter for questions based on survey permission group and country access

---


EMcQ-BES (approved):
I'm always glad when someone else is able to pick up permissions and roll with it, means I didn't write it _too_ dense

Looks good, a lot of the times with permissions defining the rules themselves is more trouble than actually enforcing them.

PR #3655, opened 2022-01-31
RN-347 enable aggregate tab to use transfer

billli0:
### Issue #:
RN-347 

### Changes:
1. Enable aggregate tab to use transfer
2. Extend `useSearchDataSources` to fetch `aggregationTypes` from `report-server` 
3. Add a tooltip to show description when hover on result list

### Screenshot:
https://linear.app/bes/issue/RN-347#comment-232e5cf7

### DIFF
--- a/packages/admin-panel/src/VizBuilderApp/api/queries/useSearchDataSources.js
+++ b/packages/admin-panel/src/VizBuilderApp/api/queries/useSearchDataSources.js
@@ -7,22 +7,31 @@ import { get } from '../api';
 import { DEFAULT_REACT_QUERY_OPTIONS } from '../constants';
 import { stringifyQuery } from '@tupaia/utils';
 
-export const useSearchDataSources = ({ search, type = 'dataElement', maxResults = 100 }) =>
-  useQuery(
-    ['dataSources', type, search],
-    async () => {
-      const endpoint = stringifyQuery(undefined, 'dataSources', {
-        columns: JSON.stringify(['code', 'type']),
-        filter: JSON.stringify({
-          type,
-          code: { comparator: 'ilike', comparisonValue: `${search}%`, castAs: 'text' },
-        }),
-        pageSize: maxResults,
-      });
-      return get(endpoint);
-    },
-    {
-      ...DEFAULT_REACT_QUERY_OPTIONS,
-      keepPreviousData: true,
-    },
-  );
+export const useSearchDataSources = ({
+  getQueryFn,
+  search,
+  type = 'dataElement',
+  maxResults = 100,
+}) =>
+  useQuery(['dataSources', type, search], getQueryFn({ type, search, maxResults }), {
+    ...DEFAULT_REACT_QUERY_OPTIONS,
+    keepPreviousData: true,
+  });
+
+export const getQueryFnFromDatabase = ({ type, search, maxResults }) => async () => {
+  const endpoint = stringifyQuery(undefined, 'dataSources', {
+    columns: JSON.stringify(['code', 'type']),
+    filter: JSON.stringify({
+      type,
+      code: { comparator: 'ilike', comparisonValue: `${search}%`, castAs: 'text' },
+    }),
+    pageSize: maxResults,
+  });
+  return get(endpoint);
+};
+
+export const getQueryFnFromReportServer = ({ search }) => async () => {
+  const endpoint = stringifyQuery(undefined, 'fetchAggregationOptions', {});
+  const result = await get(endpoint);
+  return result.filter(({ code }) => code.toLowerCase().includes(search.toLowerCase()));
+};
--- a/packages/admin-panel/src/VizBuilderApp/api/queries/useSearchDataSources.js
+++ b/packages/admin-panel/src/VizBuilderApp/api/queries/useSearchDataSources.js
@@ -2,36 +2,40 @@
  * Tupaia
  *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
+import { stringifyQuery } from '@tupaia/utils';
 import { useQuery } from 'react-query';
 import { get } from '../api';
 import { DEFAULT_REACT_QUERY_OPTIONS } from '../constants';
-import { stringifyQuery } from '@tupaia/utils';
 
-export const useSearchDataSources = ({
-  getQueryFn,
-  search,
-  type = 'dataElement',
-  maxResults = 100,
-}) =>
-  useQuery(['dataSources', type, search], getQueryFn({ type, search, maxResults }), {
-    ...DEFAULT_REACT_QUERY_OPTIONS,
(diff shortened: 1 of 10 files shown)


packages/admin-panel/src/VizBuilderApp/components/DataLibrary/AggregationDataLibrary.js
```diff
@@ -0,0 +1,56 @@
...
+  aggregate: PropTypes.oneOfType([
+    PropTypes.arrayOf(
+      PropTypes.shape({
+        type: PropTypes.string.isRequired,
+        config: PropTypes.object,
+      }),
+    ),
+    PropTypes.string,
+  ]).isRequired,
+  onAggregatchange: PropTypes.func.isRequired,
```
IgorNadj, line 55: Typo onAggregatchange -> onAggregateChange

Looking at what it is in config, maybe we should call it onFetchAggregationsChange? or just onAggregationsChange?

packages/ui-components/src/components/DataLibrary/DataLibrary.js
```diff
@@ -157,11 +157,11 @@ export const DataLibrary = ({
   );
 };
 
-// FIXME: del id, name not needed
+// FIXME: del id not needed
 const optionPropType = PropTypes.shape({
   id: PropTypes.string,
   code: PropTypes.string,
-  name: PropTypes.string,
+  description: PropTypes.string,
```
IgorNadj, line 164: Happy with this :) Just missed one above on line 91

packages/admin-panel/src/VizBuilderApp/components/Panel.js
```diff
@@ -140,11 +140,21 @@ export const Panel = () => {
...
+            value={aggregate}
+            onChange={value => setTabValue('aggregate', value)}
+            onInvalidChange={handleInvalidChange}
+          />
+        ) : (
+          <AggregationDataLibrary
+            aggregate={aggregate}
+            onAggregatchange={value => {
+              setTabValue('aggregate', value);
+              setFetchJsonEditorKey(fetchJsonEditorKey + 2);
```
IgorNadj, line 154: Ah this just invalidates the fetch tab JsonEditor. To have this work properly you would need to add a separate variable and a `key` property to the aggregate tab JsonEditor, on line 144. 

Background: for some reason JsonEditor doesn't seem to reflect changes when you change its input props, so above I had to just invalidate the entire JsonEditor component using the key property when the value changes to get react to drop and recreate and reinitialise the JsonEditor component entirely. <-- this might be a good comment above on line 137, and maybe a `// see above` for this line.

IgorNadj (changes requested):
Looking good! Some wording changes I think would be good, and a couple changes and should be perfect :)

IgorNadj:
Thanks for popping that screen recording in the card, so helpful!

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/AggregationDataLibrary.js
```diff
@@ -0,0 +1,56 @@
...
+
+import React, { useState } from 'react';
+import { useSearchDataSources, getQueryFnFromReportServer } from '../../api';
+import { DataLibrary } from '@tupaia/ui-components';
+import PropTypes from 'prop-types';
+import { useDebounce } from '../../../utilities';
+
+const MAX_RESULTS = 10;
+
+const aggregateToValue = aggregate =>
```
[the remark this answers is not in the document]
billli0: Good suggestion to add a comment. I know it is a bit confusing, it is actually handling `config.aggregate` data, so I suggest to keep calling it `aggregateToValue`?

packages/admin-panel/src/VizBuilderApp/api/queries/useSearchDataSources.js
```diff
@@ -7,22 +7,31 @@ import { get } from '../api';
...
-        pageSize: maxResults,
-      });
-      return get(endpoint);
-    },
-    {
-      ...DEFAULT_REACT_QUERY_OPTIONS,
-      keepPreviousData: true,
-    },
-  );
+export const useSearchDataSources = ({
```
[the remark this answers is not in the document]
billli0: Nice one. I agree it is not a good practice to handle two different tab queries using the same key `dataSources`. Will change it later.

packages/admin-panel/src/VizBuilderApp/api/queries/useSearchDataSources.js
```diff
@@ -7,22 +7,31 @@ import { get } from '../api';
...
+    }),
+    pageSize: maxResults,
+  });
+  return get(endpoint);
+};
+
+export const getQueryFnFromReportServer = ({ search }) => async () => {
+  const endpoint = stringifyQuery(undefined, 'fetchAggregationOptions', {});
+  const result = await get(endpoint);
+  return result.filter(({ code }) => code.toLowerCase().includes(search.toLowerCase()));
```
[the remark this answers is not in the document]
billli0: Ah sorry I didn't know that. Thanks for clarifying.

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/AggregationDataLibrary.js
billli0: My bad, but Eslint should do this work for us.. :(
About the renaming, same reason as above, my preference is to use `onAggregateChange` .

packages/admin-panel/src/VizBuilderApp/components/Panel.js
billli0: Emm no wonder it doesn't reflect changes when I change the input. However, after I add the json toggle, everythings work fine. The reason is that while we toggle back and forward, JsonEditor will rerender itself, which will reflect the change. I did try to remove `setFetchJsonEditorKey` on `fetch` tab and it works perfectly.
  
So I think it is fine to do so?

IgorNadj: Perfect, much simpler that way!

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/AggregationDataLibrary.js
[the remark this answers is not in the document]
IgorNadj: Aggregate is correct! 👍

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/AggregationDataLibrary.js
IgorNadj: 👍

IgorNadj (approved):
Perfect 🏌️‍♂️

PR #3656, opened 2022-01-31
MAUI-390: Base the latest responses on data time rather than end time

edmofro:
### Issue #:
MAUI-390

### Changes:
During a disaster, users are likely more interested in data that relates to the most recent time period rather than the data that was most recently submitted. This changes the raw data download to sort by date of data rather than the time the survey was completed, and fixes a mismatch between the export and the visual, reported **Issue 2** in MAUI-390


tcaiger (approved):
I had a quick read through the ticket and it makes sense to me.

PR #3667, opened 2022-02-08
RN-336: Created DataLakeApi to pull data from the data lake

rohan-bes:
### Issue RN-336:

This is the bare minimum code to an API connection to the data-lake database up and running for Tupaia to allow pulling of analytics and events from it.

Later in RN-242 we will improve at least the following issues:
- Refactoring code duplication with `@tupaia/data-api` and `@tupaia/database`
- Switch to typescript
- Add unit tests
- Extend to support data types other than Number 

### Changes:
- Added new `@tupaia/data-lake-api` package with basic interface to pull data from the data-lake database
- Added `DataLakeService` in `@tupaia/data-broker` to use `@tupaia/data-lake-api`
- Added `data-lake` type to `service_type` enum in the database

### Merge tasks:
- [ ] Update .env vars in lastpass for dev/prod


packages/data-lake-api/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "url": "git+https://github.com/beyondessential/tupaia.git",
+    "directory": "packages/data-lake-api"
+  },
+  "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "babel src --out-dir dist --source-maps --ignore \"src/__tests__/**\" --config-file \"../../babel.config.json\"",
+    "lint": "eslint --ignore-path ../../.gitignore .",
+    "lint:fix": "yarn lint --fix",
+    "test": "jest --passWithNoTests",
```
rohan-bes, line 17: No tests for now, will add in RN-242

packages/data-lake-api/src/DataLakeApi.js
```diff
@@ -0,0 +1,75 @@
...
+  );
+
+let dataLakeDatabase;
+
+const getDatabase = () => {
+  if (!dataLakeDatabase) {
+    dataLakeDatabase = new DataLakeDatabase();
+  }
+  return dataLakeDatabase;
+};
```
rohan-bes, line 35: Used this singleton pattern so that there's just a single connection to the data-lake database. Not entirely happy with it, but there is some trickiness around managing a long-lived connection when the whole data-broker->api chain gets re-created each data fetch call...

packages/data-lake-api/src/DataLakeEventsFetchQuery.js
```diff
@@ -0,0 +1,104 @@
...
+    const sqlQuery = new SqlQuery(query, params);
+
+    return sqlQuery.executeOnDatabase(this.database);
+  }
+
+  getAliasedColumns() {
+    const aliasedColumns = [
+      'date',
+      'entity_code AS "entityCode"',
+      'entity_code AS "entityName"',
```
rohan-bes, line 42: Just using `entity_code` as `name` here. Personally I'm a bit torn on how we want to handle this in the future.

We could fairly easily query the database to get the entity names from the code, however that alone really highlights to me how redundant the `entityName` field really is... Going forward I feel I'd prefer we omitted it entirely... thoughts?

packages/database/src/migrations/20220203064851-AddDataLakeServiceType-modifies-schema.js
```diff
@@ -0,0 +1,45 @@
...
+        'tupaia',
+        'indicator',
+        'weather',
+        'kobo',
+        'data-lake'
+    );
+    ALTER TABLE data_source ALTER COLUMN service_type TYPE service_type USING (service_type::service_type);
+    ALTER TABLE data_service_sync_group ALTER COLUMN service_type TYPE service_type USING (service_type::service_type);
+    ALTER TABLE sync_service ALTER COLUMN service_type TYPE service_type USING (service_type::service_type);
+    ALTER TABLE sync_service_log ALTER COLUMN service_type TYPE service_type USING (service_type::service_type);
```
rohan-bes, line 35: I wanted to use postgres's `ALTER TYPE ... ADD VALUE ...` syntax, however that doesn't work with pg-migrate as it can't run in a transaction: https://stackoverflow.com/questions/53149484/error-alter-type-add-cannot-run-inside-a-transaction-block

packages/data-lake-api/src/DataLakeApi.js
IgorNadj: Seems like the singleton decision should be made by whoever instantiates a DataBroker, and then databroker should create services once, which create apis once. So I would move this to the constructor.

packages/data-lake-api/src/DataLakeEventsFetchQuery.js
IgorNadj: Agree, lets force the issue by returning empty string instead?

packages/data-lake-api/src/DataLakeEventsFetchQuery.js
```diff
@@ -0,0 +1,104 @@
...
+    ];
+    if (this.hasDataElements) {
+      aliasedColumns.push('data_element_code AS "dataElementCode"');
+    }
+    return aliasedColumns.join(', ');
+  }
+
+  getInnerJoinsAndParams() {
+    let params = this.entityCodes;
+    const joins = [SqlQuery.innerJoin('analytics', 'entity_code', this.entityCodes)];
```
IgorNadj, line 55: How come this query has joins on the same table?

packages/data-lake-api/src/DataLakeAnalyticsFetchQuery.js
```diff
@@ -0,0 +1,74 @@
+/**
```
IgorNadj, line 1: Can we have this be a typescript package please :) JS is so 1990s

packages/data-lake-api/src/DataLakeDatabase.js
```diff
@@ -0,0 +1,48 @@
...
+      return true;
+    };
+    this.connectionPromise = connectToDatabase();
+  }
+
+  async closeConnections() {
+    return this.connection.destroy();
+  }
+
+  async waitUntilConnected() {
```
IgorNadj, line 31: Do we need this? This is useful for our primary db because we want to keep trying to connect to the db in case it blips. But for simple data fetching, the connection should either be there or show an error to the user I'd say.

packages/data-lake-api/src/DataLakeDatabase.js
```diff
@@ -0,0 +1,48 @@
...
+    await this.connectionPromise;
+  }
+
+  /**
+   * Runs an arbitrary SQL query against the database.
+   *
+   * Use only for situations in which Knex is not able to assemble a query.
+   */
+  async executeSql(sqlString, parametersToBind) {
```
IgorNadj, line 40: I see this is used below, but not sure why? Seems like a simple enough SELECT * FROM analytics query?

packages/data-lake-api/src/SqlQuery.js
```diff
@@ -0,0 +1,33 @@
...
+  constructor(baseQuery, baseParameters = []) {
+    this.query = baseQuery;
+    this.parameters = baseParameters;
+  }
+
+  async executeOnDatabase(database) {
+    return database.executeSql(this.query, this.parameters);
+  }
+
+  loggableQuery() {
```
IgorNadj, line 27: Knex has query logging built in :) 
Otherwise something like this works wonders https://github.com/beyondessential/tupaia/blob/dev/packages/database/src/TupaiaDatabase.js#L516

packages/data-lake-api/src/getConnectionConfig.js
```diff
@@ -0,0 +1,21 @@
...
+import { requireEnv, getEnvVarOrDefault } from '@tupaia/utils';
+
+export const getConnectionConfig = () => ({
+  host: requireEnv('DATA_LAKE_DB_URL'),
+  port: getEnvVarOrDefault('DATA_LAKE_DB_PORT', 5432),
+  user: requireEnv('DATA_LAKE_DB_USER'),
+  password: requireEnv('DATA_LAKE_DB_PASSWORD'),
+  database: requireEnv('DATA_LAKE_DB_NAME'),
+  ssl:
+    process.env.DATA_LAKE_DB_ENABLE_SSL === 'true'
```
IgorNadj, line 15: It seems like we use both DB_ENABLE_SSL and DB_DISABLE_SSL patterns currently... my preference is for the default to be on and you have to pass a flag DISABLE_SSL to turn it off. Not a biggie though.

IgorNadj (changes requested):
Nice work! Left some comments

packages/data-lake-api/src/DataLakeEventsFetchQuery.js
rohan-bes: Oooh good idea! Will do 👍

packages/data-lake-api/src/DataLakeAnalyticsFetchQuery.js
rohan-bes: Hahaha I'm with ya 😂 But I was kinda hoping to get this PoC merged in (with all it's code issues) and then fix them up shortly in [RN-242](https://linear.app/bes/issue/RN-242/add-data-lake-data-service-in-tupaia)? 

More than just switching to typescript, I also wanna refactor out common code into utils packages, and take a bit more time to consider how the api should work.

The reason for merging this in its current state would just be to allow Juliana to start working with it, as I don't think any changes I'd make would impact her side of things 👍

packages/data-lake-api/src/DataLakeDatabase.js
rohan-bes: Hmmm yeah fair point! I'll scrap it 👍

packages/data-lake-api/src/DataLakeEventsFetchQuery.js
rohan-bes: Yeah the joins here a bit a surprising, but it's a performance optimisation we came across when noticing slowness in the analytics table.

Basically what we want to be saying here is: `SELECT * from analytics WHERE entity_code IN ${this.entityCodes}`
However, when there are looots of entity codes (which happens sometimes with large hierarchies, ie. all schools in Laos) this can be quite slow.

The recommendation seems to be to instead use a `VALUES` expression and join it as so:
```
SELECT * FROM analytics
INNER JOIN (
  VALUES ${this.entityCodes}
) entity_codes(code) on entity_codes.code = analytics.entity_code
```

https://dba.stackexchange.com/questions/91247/optimizing-a-postgres-query-with-a-large-in

packages/data-lake-api/src/SqlQuery.js
rohan-bes: This is all copy pasted from the `SqlQuery.js` in `@tupaia/data-api`, however I think in my experience with `query.toString()` it didn't handle logging query parameters correctly, it just left them as `?`... I'll have to double check that now as my memory is a bit foggy on it

packages/data-lake-api/src/getConnectionConfig.js
rohan-bes: Come to think of it, since nobody will be running the data-lake in local, perhaps we will always be using SSL? I might just scrap this key 👍

packages/data-lake-api/src/DataLakeDatabase.js
rohan-bes: Not exactly sure what you mean by this? Perhaps this answer sheds some light though? https://github.com/beyondessential/tupaia/pull/3667/files#r803160664

packages/data-lake-api/src/DataLakeEventsFetchQuery.js
IgorNadj: Ah cool, can we pop that in as a comment, super cool trick

packages/data-lake-api/src/DataLakeAnalyticsFetchQuery.js
IgorNadj: Ha I get it, I think it will be relatively quick though no? Especially since right now we are pretty loose on anys

packages/data-lake-api/src/DataLakeDatabase.js
IgorNadj: That answers it thanks

packages/data-lake-api/src/SqlQuery.js
IgorNadj: No worries, minor comment

packages/data-lake-api/src/getConnectionConfig.js
IgorNadj: That works too

packages/data-lake-api/src/DataLakeApi.js
IgorNadj: Ok sounds good

IgorNadj (approved):
🏌️‍♂️

PR #3682, opened 2022-02-14
Fix maximum call stack size exceeded with array.push

billli0:
### Issue #:
[Issue from slack](https://beyondessential.slack.com/archives/C02HES8HFQF/p1644452399914279)

We were doing an `arrray.push` over a large number of values. It will put a large amount of number (number > ~135000) as function arguments on the stack, which will cause maximum call stack size exceeded error.

### Changes:
Using spread `...` can fix this.

### DIFF
--- a/packages/indicators/src/Builder/helpers.ts
+++ b/packages/indicators/src/Builder/helpers.ts
@@ -32,12 +32,12 @@ export const fetchAnalytics = async (
   // Group data elements per aggregationList to minimise aggregator calls
   const aggregationJsonToElements = groupKeysByValueJson(aggregationLisByElement);
 
-  const analytics: Analytic[] = [];
+  let analytics: Analytic[] = [];
   await Promise.all(
     Object.entries(aggregationJsonToElements).map(async ([aggregationJson, elements]) => {
       const aggregations = JSON.parse(aggregationJson);
       const { results } = await aggregator.fetchAnalytics(elements, fetchOptions, { aggregations });
-      analytics.push(...results);
+      analytics = [...analytics, ...results];
     }),
   );
 
--- a/packages/indicators/src/Builder/helpers.ts
+++ b/packages/indicators/src/Builder/helpers.ts
@@ -37,7 +37,7 @@ export const fetchAnalytics = async (
     Object.entries(aggregationJsonToElements).map(async ([aggregationJson, elements]) => {
       const aggregations = JSON.parse(aggregationJson);
       const { results } = await aggregator.fetchAnalytics(elements, fetchOptions, { aggregations });
-      analytics = [...analytics, ...results];
+      analytics = analytics.concat(results);
     }),
   );
 


packages/indicators/src/Builder/helpers.ts
```diff
@@ -32,12 +32,12 @@ export const fetchAnalytics = async (
...
   const aggregationJsonToElements = groupKeysByValueJson(aggregationLisByElement);
 
-  const analytics: Analytic[] = [];
+  let analytics: Analytic[] = [];
   await Promise.all(
     Object.entries(aggregationJsonToElements).map(async ([aggregationJson, elements]) => {
       const aggregations = JSON.parse(aggregationJson);
       const { results } = await aggregator.fetchAnalytics(elements, fetchOptions, { aggregations });
-      analytics.push(...results);
+      analytics = [...analytics, ...results];
```
edmofro, line 40: @billli0 nice work! Because this is in quite a central piece of code that deals with a lot of data, we'd want to be confident that the change to using spread here doesn't slow things down. I suspect it may, as we are creating a new array and inserting all of the original items (`analytics`) each time, as well as inserting the new items, whereas the old way kept the same array in place and just appended the new items. Could you take a look into whether this has a performance hit? You could take some measurements with large amounts of data to check. 

If it is slower this way, we could test the size of the array, and vary our strategy based on its length (potentially slicing then `.push`ing each batch)

edmofro (commented):
Really great find here @billli0! I wouldn't have suspected it. I'm happy with the code, but want to make sure there's no performance impact first.

packages/indicators/src/Builder/helpers.ts
billli0: Thanks @edmofro, I haven't thought about that. Another way to avoid copy and paste with `...analytics` is to use `analytics = analytics.concat(results);`, so that we don't need to do more work on chopping the data into piece.

I've also done the performance measurements on 13500 analytics plus 13500 analytics (two aggregations in a list), and I couldn't tell any difference on the timeout between `.push` and `...`. Maybe it will come to an issue while we have many aggregations fetching in indicator, which I don't think we have for now.

edmofro: Cool, thanks for that @billli0. That's a really good point that we don't tend to have so many aggregations, maybe only half a dozen at most, so there aren't too many iterations. Nice work!

edmofro (approved):

packages/indicators/src/Builder/helpers.ts
billli0: Awesome! I've also found `.push(...results)` has been used in many place in tupaia. I think it is ok to update them in this PR to make the system more robust?

edmofro: I would prefer to wait until we know each place is an issue before changing them, as we would need to think through performance implications more carefully in other places where there are more iterations. I understand `.concat` is slower than `.push`, so it's better to stick with push where the array we're pushing is small enough (see e.g. https://www.measurethat.net/Benchmarks/Show/4223/0/array-concat-vs-spread-operator-vs-push)

billli0: Good point! And this benchmark website is pretty cool! Thanks for sharing.

PR #3687, opened 2022-02-14
WAITP-1103: Include category headers for nested data when exporting

edmofro:
### Issue #:
- WAITP-1103

### Changes:
In https://github.com/beyondessential/tupaia/pull/279 we updated matrixes to allow nested row categories, but we didn't add handling for this to the export to excel formatter. This PR handles those nested row categories during export to excel.


nick-beyondessential (approved):
Took me while to wrap my head around the recursion, thrown a little by the manipulation of the outer context data. But now I get it, looks good!

PR #3688, opened 2022-02-15
RN-362 Use survey codes instead of names when importing responses

IgorNadj:
### Changes:

See card (RN-362) for import/export logic changes

--

Note this change is separate from the base branch (phx-152-allow-merging-answers), which adds functionality but does not break existing functionality. I based this PR on that branch because it also contains a huge test refactor which was easier to work on top of.

### Screenshot
https://linear.app/bes/issue/RN-362#comment-40464fab


packages/meditrak-server/src/apiV2/import/importSurveyResponses/importSurveyResponses.js
```diff
@@ -103,25 +99,38 @@ export async function importSurveyResponses(req, res) {
...
+          );
+        }
+      }
+    }
+
+    const entitiesBySurveyCode = await getEntitiesBySurveyCode(
       models,
       workbook.Sheets,
-      surveyNames,
+      querySurveyCodes,
```
edmofro, line 126: I would have thought this needs to take `tabNames` if no survey codes are passed in the query?

edmofro (commented):
Thanks @IgorNadj, this touches quite a few places! My main question is whether we should ask the user to enter surveys in the import modal at all? As far as I can tell, this will restrict them if they enter something that's not in the spreadsheet, but doesn't restrict the other way around (i.e. all tabs will be imported, even if they're not included in the list they provide)? I could be missing where that happens though. Either way, I'm channeling you and wondering whether the feature of being able to specify surveys warrants the extra UI and code clutter?

That point aside, all the changes look really good - thanks for simplifying this down!

packages/meditrak-server/src/tests/apiV2/import/importSurveyResponses/importSurveyResponses.test.js
```diff
@@ -8,7 +8,7 @@ import { testFunctionality } from './testFunctionality';
 import { testOutdatedStatusUpdate } from './testOutdatedStatusUpdate';
 import { testValidation } from './testValidation';
 
-describe('importSurveyResponses(): POST import/surveyResponses', () => {
+describe.only('importSurveyResponses(): POST import/surveyResponses', () => {
```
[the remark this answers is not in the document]
IgorNadj: D: I should really start passing this as a flag instead of modifying code

packages/meditrak-server/src/tests/apiV2/import/importSurveyResponses/testPermissions.js
```diff
@@ -166,11 +166,13 @@ export const testPermissions = async () => {
...
     await app.grantAccess(DEFAULT_POLICY);
     const response = await importFile('importResponsesFromMultipleSurveys.xlsx', [
-      SURVEY_NAME_1,
-      SURVEY_NAME_2,
+      SURVEY_CODE_1,
+      SURVEY_CODE_2,
     ]);
     const { statusCode } = response;
 
+    console.log(response.body);
```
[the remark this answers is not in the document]
IgorNadj: Removed

IgorNadj:
> Thanks @IgorNadj, this touches quite a few places! My main question is whether we should ask the user to enter surveys in the import modal at all? 

I thought about this too. I left it in because it can be useful to only reimport a single tab even if your spreadsheet contains multiple - it saves having to do work with your spreadsheet to make a copy, remove tabs etc. Especially if your spreadsheet contains multiple surveys. For example, you can export all survey responses for Demo Land, and it gives you a file with 20+ tabs. No point reimporting everything when you only changed something in a single tab.

Your comment about the functionality not working as expected is correct! It wasn't! I've added a fix for this and a test.

packages/meditrak-server/src/tests/apiV2/import/importSurveyResponses/testGeneral.js
```diff
@@ -0,0 +1,83 @@
...
+    it('makes sure tabs exist matching the survey codes specified', async () => {
+      const response = await importFile(app, `general/basic.xlsx`, [
+        'A_Survey_Code_That_Is_Not_Specified_In_Spreadsheet_Tabs',
+      ]);
+      expectError(response, /specified in import but there is no tab named/, 400);
+    });
+
+    it('detects survey codes via tab names if not present in query', async () => {
+      const response = await importFile(app, `general/basic.xlsx`, []); // surveyCodes empty
+      expect(response.status).to.equal(200);
```
edmofro, line 69: Be good to check here that survey responses against all tabs now exist in the db

edmofro (commented):
Cool, fix looks good. Just waiting on a response [here](https://github.com/beyondessential/tupaia/pull/3688/files/efd8b349a40886ad1d924cb1eac945d94d20aabb#r806391192), otherwise this is ready

packages/meditrak-server/src/tests/apiV2/import/importSurveyResponses/testGeneral.js
IgorNadj: Good point, added check against num survey responses for all tests.

packages/meditrak-server/src/apiV2/import/importSurveyResponses/importSurveyResponses.js
IgorNadj: Ah this argument was obsolete after some cleanup, removed.

packages/database/src/migrations/20220211015354-AddConstraintSurveyCodeLength-modifies-schema.js
```diff
@@ -0,0 +1,27 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`ALTER TABLE survey ALTER COLUMN code type varchar(30);`);
```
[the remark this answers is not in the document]
IgorNadj: Yeah this was annoyingly tricky... I left it. On create via import it's impossible to go over 30. On edit, we don't have validation in admin panel edit modals, and didn't think the cost/benefit is worth it for this PR.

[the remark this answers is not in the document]
edmofro: Fair call

packages/meditrak-server/src/tests/apiV2/surveyResponse.test.js
```diff
@@ -77,7 +82,7 @@ describe('surveyResponse endpoint', () => {
     const [{ survey }] = await buildAndInsertSurveys(models, [
       {
         id: surveyId,
-        code: 'TEST_SURVEY_' + generateId(), // prevent test cross-pollination
```
edmofro, line 80: Why this change?

edmofro (approved):
✔️

packages/meditrak-server/src/tests/apiV2/surveyResponse.test.js
IgorNadj: Ah there is no cross-pollination anyway since each test file runs in isolation, and the total char length was >30, so I shortened it.

edmofro: Gotcha

IgorNadj:
Rebased on phx-152-allow-merging-answers

PR #3691, opened 2022-02-16
WAITP-1104: Fix single record editing on permissions page

edmofro:
### Issue #:
WAITP-1104

### Changes:
Handling the creation of permissions in bulk added some very helpful functionality, but broke editing a single record. This reinstates editing a single record.

### DIFF
--- a/packages/admin-panel/src/pages/resources/PermissionsPage.js
+++ b/packages/admin-panel/src/pages/resources/PermissionsPage.js
@@ -103,8 +103,15 @@ const CREATE_CONFIG = {
   },
 };
 
-// Return an array of records for bulk editing on the server
-const processDataForSave = fieldsToSave => {
+// When creating, return an array of records for bulk editing on the server
+// When editing, just process a single record as normal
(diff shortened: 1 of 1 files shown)


nick-beyondessential (approved):
LGTM

IgorNadj (approved):

packages/admin-panel/src/pages/resources/PermissionsPage.js
```diff
@@ -103,8 +103,15 @@ const CREATE_CONFIG = {
   },
 };
 
-// Return an array of records for bulk editing on the server
-const processDataForSave = fieldsToSave => {
+// When creating, return an array of records for bulk editing on the server
+// When editing, just process a single record as normal
+const processDataForSave = (fieldsToSave, recordData) => {
```
nick-beyondessential, line 108: Would recordData ever be undefined?

edmofro: Good question - it shouldn't be, at the point we're processing data for save.

PR #3692, opened 2022-02-17
WAITP-1106: Properly delete route53 entries on teardown

edmofro:
### Issue #:
WAITP-1106

### Changes:
- Use boto3's paginator (and undocumented `build_full_result` feature) to get all route53 entries when checking which are appropriate to delete, so we don't just get the first page and miss the ones we're after
- Only delete subdomains via dns if the instance is tagged as having them (which the app server isn't - didn't realise this was buggy because we've never actually been deleting route53 entries due to the pagination issue!)


IgorNadj (approved):
Nice

PR #3693, opened 2022-02-17
RN-258: Show loading state on admin panel login

edmofro:
### Issue #:
- RN-258

### Changes:
Although the ticket says to use a loading spinner, our common ui-component Button uses the "Loading..." text, so I've gone with that here too. That keeps the interface consistent with PSSS, LESMIS, etc.

image


billli0 (approved):
LGTM, finally this ticket can move across. Thanks @edmofro !

PR #3695, opened 2022-02-18
MAUI-429: Create project Covid-19 Kiribati

chris-pollard:
### Issue #:
MAUI-429: Create project Covid-19 Kiribati
### Changes:
Add migration file with the following functions:
- add entity
- add dashboard
- add map overlay group
- add entity hierarchy
- add entity relation for country
- add project record


packages/database/src/migrations/20220218050602-AddCovidKiribatiProject-modifies-data.js
```diff
@@ -0,0 +1,140 @@
...
+  });
+};
+
+const addEntity = async db => {
+  const id = generateId();
+  const { bounds } = (await getWorldBounds(db)).rows[0];
+  const parentId = await codeToId(db, 'entity', 'World');
+  await insertObject(db, 'entity', {
+    id,
+    bounds,
```
edmofro, line 79: Are you sure you want the bounds to be the same as World, not the same as Kiribati? Or potentially a slightly further zoomed out view than what we have for Kiribati at the moment

packages/database/src/migrations/20220218050602-AddCovidKiribatiProject-modifies-data.js
```diff
@@ -0,0 +1,140 @@
...
+const DASHBOARD_CODE = 'covid_ki';
+const DASHBOARD_NAME = 'COVID-19 Kiribati';
+const MAP_OVERLAY_GROUP_CODE = 'covid_ki';
+const MAP_OVERLAY_GROUP_NAME = 'COVID-19 Kiribati';
+const PROJECT = {
+  code: PROJECT_CODE,
+  description: 'COVID-19 case surveillance and data visualisations in Kiribati',
+  sort_order: 16,
+  image_url:
+    'https://tupaia.s3.ap-southeast-2.amazonaws.com/thumbnails/uploads/covid_kiribati_background.jpeg',
```
edmofro, line 30: The `thumbnails` folder on S3 is actually supposed to be populated with smaller versions of images uploaded to the main `uploads` folder (there's a lambda script that detects the upload and puts a compressed version in `thumbnails/uploads`)

edmofro (changes requested):
All the code looks great, just a couple of operational things to suggest

packages/database/src/migrations/20220218050602-AddCovidKiribatiProject-modifies-data.js
chris-pollard: I was copying some other project creations here, so didn't really understand the usage. I will change that.

packages/database/src/migrations/20220218050602-AddCovidKiribatiProject-modifies-data.js
chris-pollard: aha! Thanks.

packages/database/src/migrations/20220218050602-AddCovidKiribatiProject-modifies-data.js
edmofro: Oh yep, makes sense! This refers to the geographic bounds that Tupaia should zoom to when you visit that entity. For projects where there are multiple countries involved, it makes sense to use bounds that capture all of those countries, so the World bounds is an easy shortcut (as it's actually just a decent view of the Pacific, despite being called World). For this project though, we likely want a tighter view that's zoomed in on just Kiribati (which is admittedly a very stretched out nation!)

edmofro (approved):
lgtm!

PR #3698, opened 2022-02-18
Maui 517 add data elements from datalake

billli0:
### Issue #:
MAUI-517 create-data-elements-in-tupaia-for-datalakes-data-elements

### Changes:
Got all data elements from data lake prod. If more coming up, we need to add it by migration for now.

- Example

---


packages/database/src/migrations/20220218015022-AddDataLakeDataElementsIntoTupaiaDataSource-modifies-data.js
```diff
@@ -0,0 +1,39 @@
...
+exports.up = async function (db) {
+  const formattedValues = DATA_ELEMENT_CODES.dataElementCodes
+    .map(code => `('${generateId()}', '${code}','dataElement','data-lake')`)
+    .join(',');
+  return db.runSql(`
+    INSERT INTO data_source (id, code, type, service_type)
+    VALUES
+    ${formattedValues}
+    ON CONFLICT ON CONSTRAINT data_source_code_type_key
+    DO UPDATE SET service_type = 'data-lake';
```
billli0, line 29: One code (covid_n_test) has been found in `data-lake` as same as one of the data element code in tupaia. Have a chat with Juliana, and we decided to delete the survey it is associated to (done) and update its `service_type`.

rohan-bes: If this is being deleted we shouldn't need this right? I'm slightly worried that if another clash happened, this migration would move that data element to the data-lake with cleaning up the requisite question/survey. But I think I'm just being paranoid 😉 Should be fine 👍

rohan-bes (approved):
Looks good, nice one!

packages/database/src/migrations/20220218015022-AddDataLakeDataElementsIntoTupaiaDataSource-modifies-data.js
billli0: Emm that's right. Let me double check.

billli0: Got another test to make sure there is not another clash. It is good to go. Very good thought!

billli0:
I've seen the number of data element codes in data lake has been drop from ~1000 to ~670, so I've updated the json file.

PR #3699, opened 2022-02-18
RN-368: Added overnight shutdown for RDS

rohan-bes:
### Issue RN-368:

Added deployment scripts to stop and start RDS instances based on the `StartAtUTC` and `StopAtUTC` tags.

Will add EventBridge configurations to trigger these automatically


edmofro (approved):
Thanks for doing this, will save some $!

PR #3700, opened 2022-02-18
RN-161 Fetch transform schema endpoints

billli0:
### Issue #:
Part of RN-161

### Changes:
Slightly different approach from discussed with @IgorNadj . We can actually expose the type objects that are used by `yup.js`, which is a constant object.

### DIFF
--- a/packages/report-server/src/reportBuilder/transform/functions/index.ts
+++ b/packages/report-server/src/reportBuilder/transform/functions/index.ts
@@ -6,13 +6,22 @@
 import { Context } from '../../context';
 import { Row } from '../../types';
 
-import { buildInsertColumns } from './insertColumns';
-import { buildExcludeColumns } from './excludeColumns';
-import { buildUpdateColumns } from './updateColumns';
-import { buildMergeRows } from './mergeRows';
-import { buildSortRows } from './sortRows';
-import { buildExcludeRows } from './excludeRows';
-import { buildInsertRows } from './insertRows';
+import {
+  buildInsertColumns,
+  paramsValidator as insertColumnsParamsValidator,
+} from './insertColumns';
+import {
+  buildExcludeColumns,
+  paramsValidator as excludeColumnsParamsValidator,
+} from './excludeColumns';
+import {
+  buildUpdateColumns,
+  paramsValidator as updateColumnsParamsValidator,
+} from './updateColumns';
+import { buildMergeRows, paramsValidator as mergeRowsParamsValidator } from './mergeRows';
+import { buildSortRows, paramsValidator as sortRowsParamsValidator } from './sortRows';
+import { buildExcludeRows, paramsValidator as excludeRowsParamsValidator } from './excludeRows';
+import { buildInsertRows, paramsValidator as insertRowsParamsValidator } from './insertRows';
 
 type TransformBuilder = (params: unknown, context: Context) => (rows: Row[]) => Row[];
 
@@ -24,4 +33,20 @@ export const transformBuilders: Record<string, TransformBuilder> = {
   sortRows: buildSortRows,
   excludeRows: buildExcludeRows,
   insertRows: buildInsertRows,
+};
+
+export const transformSchemas: Record<
+  string,
+  {
+    type: string;
+    fields: Record<string, unknown>;
+  }
+> = {
+  insertColumns: insertColumnsParamsValidator.describe(),
+  excludeColumns: excludeColumnsParamsValidator.describe(),
+  updateColumns: updateColumnsParamsValidator.describe(),
+  mergeRows: mergeRowsParamsValidator.describe(),
+  sortRows: sortRowsParamsValidator.describe(),
+  excludeRows: excludeRowsParamsValidator.describe(),
+  insertRows: insertRowsParamsValidator.describe(),
 };
--- a/packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
+++ b/packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
@@ -20,7 +20,7 @@ type MergeRowsParams = {
   where: (parser: TransformParser) => boolean;
 };
 
-const paramsValidator = yup.object().shape({
+export const paramsValidator = yup.object().shape({
   groupBy: starSingleOrMultipleColumnsValidator,
   using: yup.lazy((value: unknown) => {
     const optionalMergeStrategyNameValidator = yup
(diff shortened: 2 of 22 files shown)


packages/report-server/src/routes/FetchTransformSchemaRoute.ts
```diff
@@ -0,0 +1,48 @@
...
+      ([transformKey, config]) => {
+        const { fields } = config;
+        return {
+          name: transformKey,
+          code: transformKey,
+          ...fields,
+          where: {
+            type: 'string',
+            optional: true,
+          },
```
rohan-bes, line 33: I don't think this is quite accurate. The `where` clause doesn't apply to all transforms (I think `sortRows` doesn't use it). Moreso, if it does apply, it ought to be in the `paramsValidator` for that transform right? So we can omit this?

packages/report-server/src/reportBuilder/transform/functions/index.ts
```diff
@@ -25,3 +34,19 @@ export const transformBuilders: Record<string, TransformBuilder> = {
...
+    fields: Record<string, unknown>;
+  }
+> = {
+  insertColumns: insertColumnsParamsValidator.describe(),
+  excludeColumns: excludeColumnsParamsValidator.describe(),
+  updateColumns: updateColumnsParamsValidator.describe(),
+  mergeRows: mergeRowsParamsValidator.describe(),
+  sortRows: sortRowsParamsValidator.describe(),
+  excludeRows: excludeRowsParamsValidator.describe(),
+  insertRows: insertRowsParamsValidator.describe(),
```
rohan-bes, line 51: Wow! I didn't know we could use `yup.describe` like this! So cool!

packages/report-server/src/reportBuilder/transform/functions/index.ts
```diff
@@ -25,3 +34,19 @@ export const transformBuilders: Record<string, TransformBuilder> = {
...
+  string,
+  {
+    type: string;
+    fields: Record<string, unknown>;
+  }
+> = {
+  insertColumns: insertColumnsParamsValidator.describe(),
+  excludeColumns: excludeColumnsParamsValidator.describe(),
+  updateColumns: updateColumnsParamsValidator.describe(),
+  mergeRows: mergeRowsParamsValidator.describe(),
```
rohan-bes, line 48: Hmmm so one thing I've noticed is that it seems `yup.describe()` returns `null` for any lazy validators. This is totally understandable, as the type of validator can't be known until a value is given, but does present a slight problem for us as we'd like to get some sort of schema for that validator here...

Ways we could address this might be:
- Re-writing those validators to not use lazy validation. Not too sure if this is possible though...
- Prodividing some schema definitions for the lazy validators to combine with the default `yup.describe()`?

rohan-bes (changes requested):
Looks great @billli0, I love using `yup.describe()`.

A few comments in there but here's a couple more:
1. It would be great to get entries for `/fetchTransformSchemas` and `/fetchAggregationOptions` in the report-server examples.http file
2. I have a concern around the fact that the schema returned from `/fetchAggregationOptions` is totally different to the one returned from `/fetchTransformSchemas`. It would be nice if these were largely consistent. I'm not too fussed which we make more like the other. Probably the best approach is to come up with our own format that we'd like and then to align both of them with that?

packages/report-server/src/reportBuilder/transform/functions/index.ts
billli0: I've tried these two options and it didn't work. Another hacky way is to extend class `Lazy` and modify the `describe()`. More info is coming up from next comment.

packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
```diff
@@ -20,32 +20,36 @@ type MergeRowsParams = {
...
 
-    throw new yup.ValidationError(
-      'mergeUsing must be either a single merge strategy, or a mapping between columns and merge strategies',
-    );
-  }),
+      throw new yup.ValidationError(
+        'mergeUsing must be either a single merge strategy, or a mapping between columns and merge strategies',
+      );
+    },
+    [optionalMergeStrategyNameValidator, yup.object().shape({})],
```
billli0, line 51: We need to pass in all validators from lazy function (`mergeStrategyNameValidator` and `optionalMergeStrategyNameValidator` are the same, so only pass in one of them), so that we can convert them into json schema.

packages/utils/src/validation/yup.js
```diff
@@ -127,9 +127,52 @@ const testSync = (schema, createError) =>
...
+
+    if (convertedSchema.length === 1) {
+      return convertedSchema[0];
+    }
+
+    return {
+      oneOf: convertedSchema,
+    };
+  }
+}
```
billli0, line 166: This is not fully capable of converting all yup schema, but it should work for all our transform functions.

packages/report-server/src/reportBuilder/transform/functions/index.ts
billli0: Thanks @rohan-bes for pushing me to do the schema! I did try before but I gave up..

packages/report-server/src/reportBuilder/transform/functions/mergeRows/mergeRows.ts
billli0: If that requires more explanation, we can have a chat on it.

packages/utils/src/validation/yup.js
```diff
@@ -127,9 +127,52 @@ const testSync = (schema, createError) =>
     },
   });
 
+class CustomLazy extends Lazy {
```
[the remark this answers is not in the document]
rohan-bes: Also, can we please break out into a separate file 🙏

packages/utils/src/validation/yup.js
```diff
@@ -127,9 +127,52 @@ const testSync = (schema, createError) =>
...
+}
+
+const lazy = (builder, yupSchemas) => {
+  return new CustomLazy(builder, yupSchemas);
+};
+
 export const yupUtils = {
   oneOfType,
   oneOrArrayOf,
+  lazy,
```
rohan-bes, line 175: I'd also call this `describableLazy` to differentiate from the standard `lazy` 👍

packages/utils/src/validation/yup.js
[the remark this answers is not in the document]
rohan-bes: One last request, could we possibly get a couple of unit tests for this class? I think it would be really helpful to give people a better understanding of what it's for, and will make extending it easier in the future 👍

rohan-bes (approved):
Awesome work @billli0 ! Really like the extended `lazy` implementation, now the schema looks really complete!

Pre-approving with a few requests for unit tests, code structure, and renaming. 👍 

Still wanna confirm that the aggregation schema configs are in-line with the transform ones? But that can be a separate chat 👍

billli0:
@rohan-bes Yep the schemas aer in line with aggregation schema. We can use it to validate in json editor.

PR #3704, opened 2022-02-21
RN-262: Fix array based dashboard relations filters

edmofro:
### Issue #:
RN-262

### Changes:
- Adds a new filter type for array columns to use, making use of the autocomplete component

Screenshots [in the ticket](https://linear.app/bes/issue/RN-262#comment-5ad72afa)


packages/meditrak-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -42,28 +42,36 @@ export const generateLinkHeader = (resource, pageString, lastPage, originalQuery
...
 
-export const processColumnSelector = (models, unprocessedColumnSelector, baseRecordType) => {
+export const fullyQualifyColumnSelector = (models, unprocessedColumnSelector, baseRecordType) => {
   const [resource, column] = unprocessedColumnSelector.includes('.')
     ? unprocessedColumnSelector.split('.')
     : [baseRecordType, unprocessedColumnSelector];
-
   const recordType = resourceToRecordType(resource);
-  const model = models.getModelForDatabaseType(recordType);
-  const customSelector = model?.customColumnSelectors?.[column];
```
edmofro, line 52: This `customSelector` was previously being applied to both sides of the `as`, where we actually only want it on the left, e.g. `select ST_AsGeoJSON(entity.region) as entity.region` rather than `select ST_AsGeoJSON(entity.region) as ST_AsGeoJSON(entity.region)`

packages/database/src/TupaiaDatabase.js
```diff
@@ -45,7 +45,7 @@ export const JOIN_TYPES = {
 };
 
 // list valid behaviour so we can validate against sql injection
-const VALID_CAST_TYPES = ['text', 'text[]', 'date'];
+const VALID_CAST_TYPES = ['text', 'text[]', 'date', 'entity_type[]'];
```
edmofro, line 48: I have mixed feelings about this...adding a piece of very Tupaia specific metadata into the otherwise generic `TupaiaDatabase`. The name of the class would suggest it's ok, but it is the first example we have of breaking the genericism

packages/database/src/modelClasses/Entity.js
```diff
@@ -31,8 +31,8 @@ const PROJECT = 'project';
 const CITY = 'city';
 const POSTCODE = 'postcode';
 
-// Note: if a new type is not included in `ORG_UNIT_ENTITY_TYPES`,
-// a corresponding tracked entity type must be created in DHIS
+// Note: if a new type is not included in `ORG_UNIT_ENTITY_TYPES`, but data is to be stored against
+// it on DHIS2, a corresponding tracked entity type must be created in DHIS2
```
edmofro, line 35: Unrelated, just slipping in this clarification

packages/database/src/TupaiaDatabase.js
IgorNadj: Maybe 'enum' is the cast type so it can stay generic, and specify an enum type `castAsEnum: 'entity_type'`?

packages/meditrak-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -42,28 +42,36 @@ export const generateLinkHeader = (resource, pageString, lastPage, originalQuery
   return formatLinkHeader(linkHeader);
 };
 
-export const processColumnSelector = (models, unprocessedColumnSelector, baseRecordType) => {
+export const fullyQualifyColumnSelector = (models, unprocessedColumnSelector, baseRecordType) => {
```
IgorNadj, line 45: <3 descriptive names

packages/database/src/TupaiaDatabase.js
IgorNadj: N/m, removed in later commit

IgorNadj (approved):
LGTM!

IgorNadj:
One minor thing, do we want that cell rendering to put a space between array items? Seems to run into each other, e.g. "countrydistrict"

edmofro:
Good spot! I actually added exactly that, but didn't bother updating the screenshots. The array cells now render as a bullet list (which was already implemented for the map overlays page)

PR #3705, opened 2022-02-22
Add data element from data lake to tupaia

billli0:
### Issue #:
Urgently need to add data element from data lake to tupaia for visuals that needs to be done before this week.


rohan-bes (approved):
LGTM

PR #3707, opened 2022-02-22
[Hot fix]: Include json migrationData files in babel output directory

edmofro:
In today's release, we changed to running migrations within the meditrak-server process when it starts up. Subsequently, we deployed the first migration since that change as a hot-fix, and it met an edge case: json files were not included in the `dist` folder that meditrak-server reads migrations from. 

See https://beyondessential.slack.com/archives/CT0KTFSHZ/p1645492338632989?thread_ts=1645491301.510779&cid=CT0KTFSHZ

### DIFF
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -11,7 +11,7 @@
   "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
   "main": "dist/index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


rohan-bes (approved):
LGTM! But perhaps worth confirming it works on a feature instance?

edmofro:
Yeah that's a good call after we've just broken production. Spinning up now!

edmofro:
Seems to have worked :-) 
image

PR #3709, opened 2022-02-22
RN-289: Enable sorting of date column in table flippa

edmofro:
### Issue #:
- RN-289

### Changes:
Use a custom sort function for the special Date column, sorting by timestamp rather than the displayed date text


packages/ui-components/src/components/Chart/Table/getChartTableData.js
```diff
@@ -92,7 +95,15 @@ const processData = viewContent => {
 };
 
 export const getChartTableData = viewContent => {
-  const columns = useMemo(() => processColumns(viewContent), [JSON.stringify(viewContent)]);
+  // Because react-table wants its sort function to be memoized, it needs to live here, outside of
+  // the other useMemo hooks
+  const sortByTimestamp = useMemo(() => (rowA, rowB) => {
+    const rowAMoreRecent = rowA.original.timestamp > rowB.original.timestamp;
+    return rowAMoreRecent ? 1 : -1;
+  });
```
rohan-bes, line 103: Little awkward that we need to memoize it here and thus it has to be passed through as a parameter, but if it can't be helped then so be it 👍

rohan-bes (approved):
Nice and simple one!

packages/ui-components/src/components/Chart/Table/getChartTableData.js
edmofro: Yeah I agree...but they've bolded the bit that says you need to memoize it in the react-table docs, so I reckon we'd better do what they say!

packages/ui-components/src/components/Chart/Table/getChartTableData.js
```diff
@@ -22,13 +22,19 @@ const FirstColumnCell = styled.span`
...
+const makeFirstColumn = (header, accessor, sortRows) => {
+  const firstColumn = {
+    Header: header,
+    accessor,
+    // eslint-disable-next-line react/prop-types
+    Cell: ({ value }) => <FirstColumnCell>{String(value)}</FirstColumnCell>,
+  };
+  if (sortRows) {
+    firstColumn.sortType = sortRows;
+  }
```
edmofro, line 34: This fixes a bug where having `sortType` set to `undefined` broke sorting for non-date columns

packages/ui-components/src/components/DataTable/DataTable.js
```diff
@@ -27,6 +39,13 @@ export const DataTable = ({ columns, data, className, rowLimit }) => {
     {
       columns,
       data,
+      initialState: {
+        sortBy: [
+          {
+            id: getColumnId(columns[0]),
+          },
+        ],
+      },
```
edmofro, line 48: Enables sorting on first column by default, per https://linear.app/bes/issue/RN-289#comment-4d5f3393

rohan-bes (approved):
Nice fixes!

PR #3710, opened 2022-02-22
RN-195: Epic

rohan-bes:
### Issue RN-195:
Epic branch merge for the RN- Individual changes have been approved along the way. Will merge after switching over to RDS so that redeploy will work correctly.

### Merge Tasks:
- [ ] Update EventBridge configs to use new lambda action names


packages/devops/scripts/ci/triggerRedeploy.sh
```diff
@@ -21,8 +21,9 @@ fi
 
 echo "At least one running deployment, triggering redeploy of any tagged with Branch ${CI_BRANCH}"
 RESPONSE_FILE=lambda_redeploy_response.json
+# TODO: Set the 'testRDS' to a correct name before merging rn-195-epic
```
rohan-bes, line 24: So the current plan is to keep this as is. Otherwise we'd need to replace the existing deployment lambda with the new one, which will put us in a tricky spot if anything goes wrong with the merge (which it may very well...). So I was thinking we keep both lambda functions around, and then when we've confirmed that the new function is working correctly we switch this to `deployment`?

nick-beyondessential (approved):
Approving Epic branch PR as has been approved along the way.

PR #3715, opened 2022-02-23
MAUI-530: Remove invalid Kiribati entities

chris-pollard:
### Issue #:
MAUI-530

### Changes:
Create database migration that will:
- Remove invalid district and sub_district entity records
- Reassign a valid parent id to an orphan facility entity

### DIFF
--- a/packages/database/src/migrations/20220223052221-DeleteInvalidKiribatiEntities-modifies-data.js
+++ b/packages/database/src/migrations/20220223052221-DeleteInvalidKiribatiEntities-modifies-data.js
@@ -29,7 +29,6 @@ exports.up = async function (db) {
   for (const code of ENTITY_CODES) {
     const entityId = await codeToId(db, 'entity', code);
     await deleteObject(db, 'entity_relation', { child_id: entityId });
-    await deleteObject(db, 'survey_response', { entity_id: entityId });
     await deleteObject(db, 'entity', { id: entityId });
   }
 };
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20220223052221-DeleteInvalidKiribatiEntities-modifies-data.js
```diff
@@ -0,0 +1,43 @@
...
+
+exports.up = async function (db) {
+  const parentId = await codeToId(db, 'entity', VALID_DISTRICT_CODE);
+  const facilityId = await codeToId(db, 'entity', FACILITY_CODE);
+  await updateValues(db, 'entity', { parent_id: parentId }, { id: facilityId });
+
+  for (const code of ENTITY_CODES) {
+    const entityId = await codeToId(db, 'entity', code);
+    await deleteObject(db, 'entity_relation', { child_id: entityId });
+    await deleteObject(db, 'survey_response', { entity_id: entityId });
```
edmofro, line 32: I reckon we remove this line - I've checked and confirmed there are no survey responses against the entities being deleted, and as a bit of code that others will copy from, I'd prefer people have to manually deal with any data against an entity (by moving it or deleting it using the admin panel) rather than deleting it without looking

edmofro (changes requested):
One minor change requested :)

packages/database/src/migrations/20220223052221-DeleteInvalidKiribatiEntities-modifies-data.js
chris-pollard: Yep, thanks.

edmofro (approved):

PR #3722, opened 2022-02-23
migration add missing facilitiues

nick-beyondessential:
Add missing facilities from RN-301


edmofro:
Are these still missing @nick-beyondessential? Should we merge this?

nick-beyondessential:
These have been imported thanks @edmofro!

PR #3729, opened 2022-02-27
Remove db change channel timeout

IgorNadj:
Remove db change channel timeout to fix issue of db being down for ~10 minutes for the overnight sync.

Have tested on my local:
1. start servers with db down and wait 4 minutes (previous timeout), then start db 
2. start servers with db up, take db down for 4 minutes, then start db
...it reconnects and everything seems to work fine.

Not sure what the other implications of this change are.

### DIFF
--- a/packages/database/src/DatabaseChangeChannel.js
+++ b/packages/database/src/DatabaseChangeChannel.js
@@ -43,7 +43,7 @@ export class DatabaseChangeChannel extends PGPubSub {
   /**
    * Sends a ping request out to the database and listens for a response
    * @param {number} timeout - default 250ms
-   * @param {number} retries - default 240 (i.e. 1 minute total wait time)
+   * @param {number} retries - default 240 (i.e. 1 minute total wait time). Set to 0 for unlimited retries
    */
   async ping(timeout = 250, retries = 240) {
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


edmofro (approved):
Ha, this is genius in its simplicity!

I think it's totally fine :-)

edmofro:
And I think this means https://github.com/beyondessential/tupaia/pull/3730 is no longer required hey?

IgorNadj:
Tested the analytics refresh hook. Works in case 2, fails in case 1 :(

IgorNadj:
Re-tested. I think the failure was due to me not waiting for it to reconnect before updating the answers table.

Have re-tested two more times with @rohan-bes as witness and it all worked perfectly.

IgorNadj:
For posterity:

Normal knex db connections seem to reconnect just fine, haven't looked into it.

Change channel listeners re-connect as well. We looked at the source code of https://github.com/voxpelli/node-pg-pubsub/blob/main/index.js but couldn't understand how this was happening, but it is happening. Stopping a database and re-starting it, the change channel re-registers.

edmofro:
Yeah I've actually looked into that before @IgorNadj - it happens [here](https://github.com/voxpelli/node-pg-pubsub/blob/main/lib/client.js#L47)

IgorNadj:
Yep, just wasn't sure how the LISTEN query gets re-run https://github.com/voxpelli/node-pg-pubsub/blob/main/index.js#L99 Seems like black magic to me

PR #3730, opened 2022-02-27
Start DB an hour earlier than server

edmofro:
See context [on slack](https://beyondessential.slack.com/archives/CT0KTFSHZ/p1645994719332819?thread_ts=1645994457.082069&cid=CT0KTFSHZ)

### DIFF
--- a/packages/devops/scripts/lambda/actions/spin_up_tupaia_database.py
+++ b/packages/devops/scripts/lambda/actions/spin_up_tupaia_database.py
@@ -26,7 +26,11 @@ def spin_up_tupaia_database(event):
     db_instance_type = event.get('DbInstanceType', 'db.t4g.large')
     security_group_code = event.get('SecurityGroupCode', 'tupaia-dev-sg') # Use security group tagged with code
     clone_db_from = event.get('CloneDbFrom', 'production') # Use volume snapshot tagged with deployment name
-    extra_tags = build_extra_tags(event)
+    extra_tags = build_extra_tags(
+        event,
+        # Turn off overnight, but come back online an hour before the server so db is available
+        { 'StopAtUTC': '09:00', 'StartAtUTC': '18:00' }
+    )
 
     # create db instance from a snapshot
     create_db_instance_from_snapshot(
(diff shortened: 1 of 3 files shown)


rohan-bes:
@edmofro I think this one is still needed actually. That's because if the DB is stopped when we attempt the overnight clone (that current is done at 4am AEST) then it will fail. Work could be put in to finding a way to get the overnight clone working even while the db is stopped, but I think for now starting it an hour earlier and then doing the clone during that hour will prove easiest?

edmofro:
Ah right. I'll test this then :-)

packages/devops/scripts/lambda/actions/spin_up_tupaia_database.py
```diff
@@ -26,7 +26,11 @@ def spin_up_tupaia_database(event):
     db_instance_type = event.get('DbInstanceType', 'db.t4g.large')
     security_group_code = event.get('SecurityGroupCode', 'tupaia-dev-sg') # Use security group tagged with code
     clone_db_from = event.get('CloneDbFrom', 'production') # Use volume snapshot tagged with deployment name
-    extra_tags = build_extra_tags(event)
+    extra_tags = build_extra_tags(
+        event,
+        # Turn off overnight, but come back online an hour before the server so db is available
+        { 'StopAtUTC': '09:00', 'StartAtUTC': '18:00' }
```
rohan-bes, line 32: This time seems appropriate 👍 

Means we're losing an hour on the current server uptime, but that's not necessarily a bad thing.

Just wanna note though, the current eventbridge event to do reclone the databases also occurs at 18:00 UTC, which means that it will clash with this time and most likely fail. Should be fine if we bump it back by 30 mins to 18:30 UTC?

rohan-bes (approved):
Nice one! Thanks for this 👍

packages/devops/scripts/lambda/actions/spin_up_tupaia_database.py
edmofro: Yeah I figured actually we don't need them to come online so early - we set it earlier when the build time was taking over an hour, so it wouldn't be up by the time people started work. These days the server comes online really quickly.

Good call on bumping back the reclone!

edmofro:
@rohan-bes fyi this is deployed to production, and the db clone refresh is moved to 1830 UTC

PR #3737, opened 2022-02-28
offset from start date instead of end date

billli0:
### Issue #:
Currently in `report-server`, `offset` config in `fetch's startDate` will be applied on `endDate` instead of `startDate`.

1. I've checked we are lucky to have only one visual at the moment(`PG_Strive_PNG_Prev_7_Days_Reported_Cases`) to use this config, and it didn't break anything after the changes.

2. From `periodGranularities.test.js` we can tell that it always put start date and end date in query, so there is no way we have a `null` start date.

3. It will be also consistent to the frontend offset config `defaultTimePeriod`:

>    {
         start: { unit: 'day', offset: -1 },
         end: { unit: 'day', offset: 3 },
       }


rohan-bes (approved):
Nice. I think if we ever do need to add the 'offset start date from end date' behaviour, we can always add the `from` config as we discussed 👍 

Might just wanna get a bit of QA on this ticket to ensure the existing vizes aren't affected?

PR #3742, opened 2022-03-01
RN-376: Updated database dump script

rohan-bes:
### Issue RN-376:

Few small improvements to the database dump script post the RDS changes.

### Changes:
- No longer requires password input
- Uses compression
- Shows loading spinner to indicate progress

### DIFF
--- a/packages/database/scripts/dumpDatabase.sh
+++ b/packages/database/scripts/dumpDatabase.sh
@@ -11,6 +11,26 @@ Options:
 EOF
 }
 
+# https://stackoverflow.com/questions/12498304/using-bash-to-display-a-progress-indicator
+function show_loading_spinner() {
+    eval $2 2>/dev/null &
+    pid=$! # Process Id of the previous running command
+
+    spin='-\|/'
+
+    i=0
+    while kill -0 $pid 2>/dev/null
+    do
+    i=$(( (i+1) %4 ))
+    printf "\r$1 ${spin:$i:1}"
+    sleep .5
+    done
+    printf "\r$1  "
+    echo "" # reset prompt
+}
+
+source ".env"
+
 DUMP_FILE_NAME="dump.sql"
 
 identity_file=""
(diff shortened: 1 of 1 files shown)


packages/database/scripts/dumpDatabase.sh
```diff
@@ -11,6 +11,26 @@ Options:
 EOF
 }
 
+# https://stackoverflow.com/questions/12498304/using-bash-to-display-a-progress-indicator
+function show_loading_spinner() {
```
rohan-bes, line 15: Probably unnecessary and just having a little bit of fun. But it wasn't tricky and does make it clearer that something is actually happening as we wait the 30-40 mins it takes to dump the database

packages/database/scripts/dumpDatabase.sh
```diff
@@ -55,7 +75,11 @@ target_path="$(
     cd "$target_dir"
     pwd
 )/$DUMP_FILE_NAME"
+target_zip_path="$target_path.gz"
+
+show_loading_spinner "Dumping database to $target_zip_path" "PGPASSWORD=$DB_PG_PASSWORD pg_dump \"host=$host user=postgres dbname=tupaia sslmode=require sslkey=$identity_file\" -Z1 -f $target_zip_path"
```
rohan-bes, line 80: Adding the `-Z1` argument turns on compression of the dump file. I skipped doing this previously as it didn't really seem to speed up the process at all.

However for the folks at home saving 8GB of data download each time you dump the database is probably worth it

edmofro (approved):
lgtm, thanks for this!

I'm thinking of this as a helpful stopgap until we get something better in place, being either:
- A smaller db for local dev, either through generation of dummy data or stripping the live db; or
- Pointing to an RDS environment for local dev

rohan-bes:
Thanks for the review @edmofro 

Agreed about this being a stop gap for now. Hopefully after moving some of our larger projects to the data lake we won't have quite as large a database, but it's still best if we move away from devs having real data in their local environments...

rohan-bes:
@edmofro added a check for the postgres user env vars, as they probably won't be there in most people's local .env file

PR #3743, opened 2022-03-02
RN-161 Transform schema and transform tab

billli0:
### Issue #:
RN-161

### Changes:
1. Transfer tab 
2. Verify input by transform schema
3. Add `id` to each step (aggregate & transform) so that we can handle multiple selected option properly

### Screenshot:
image

### DIFF
--- a/packages/admin-panel/src/VizBuilderApp/components/DataLibrary/TransformDataLibrary.js
+++ b/packages/admin-panel/src/VizBuilderApp/components/DataLibrary/TransformDataLibrary.js
@@ -17,14 +17,14 @@ const DATA_TYPES = {
 
 const transformToValue = transform =>
   transform.map(({ id, name, transform: code, ...restOfConfig }) => ({
-    id: id || generateId(),
+    id: id || generateId(), // transform config in existing report does not have id.
     code,
     ...restOfConfig,
   }));
 
 const valueToTransform = value =>
   value.map(({ id, code, isDisabled = false, ...restOfConfig }) => ({
-    id: id || generateId(),
+    id: id || generateId(), // option from selectable options does not have id.
     transform: code,
     isDisabled,
     ...restOfConfig,
--- a/packages/admin-panel/src/VizBuilderApp/components/DataLibrary/TransformDataLibrary.js
+++ b/packages/admin-panel/src/VizBuilderApp/components/DataLibrary/TransformDataLibrary.js
@@ -35,9 +35,6 @@ export const TransformDataLibrary = ({ transform, onTransformChange, onInvalidCh
 
   const [dataType, setDataType] = useState(DATA_TYPES.TRANSFORM);
   const [inputValue, setInputValue] = useState('');
-  const [isDisabledAll, setIsDisabledAll] = useState(
-    value ? !value.some(option => !option.isDisabled) : false,
-  );
 
   const { data, isFetching } = useSearchTransformSchemas();
   const options = {
@@ -75,7 +72,7 @@ export const TransformDataLibrary = ({ transform, onTransformChange, onInvalidCh
       onInputChange={(event, newInputValue) => (event ? setInputValue(newInputValue) : false)}
       isLoading={isFetching}
       onMouseEnter={prefetchTransformSchemas}
-      optionComponent={(option, setEdittingOption) => {
+      optionComponent={(option, setIsDragDisabled) => {
         const optionWithMetaData = {
           ...Object.values(options)
             .flat()
@@ -93,12 +90,11 @@ export const TransformDataLibrary = ({ transform, onTransformChange, onInvalidCh
             optionMetaData={optionWithMetaData}
             onChange={newValue => onChangeInOption(newValue, option)}
             onRemove={onRemove}
-            setEdittingOption={setEdittingOption}
+            setIsDragDisabled={setIsDragDisabled}
             onInvalidChange={onInvalidChange}
           />
         );
       }}
-      headerConfig={{ isDisabledAll, setIsDisabledAll }}
       allowAddMultipleTimes
       supportsDisableAll
     />
--- a/packages/admin-panel/src/VizBuilderApp/components/DataLibrary/TransformDataLibrary.js
+++ b/packages/admin-panel/src/VizBuilderApp/components/DataLibrary/TransformDataLibrary.js
@@ -1,0 +1,119 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import React, { useState } from 'react';
+import generateId from 'uuid/v1';
+import PropTypes from 'prop-types';
+import { DataLibrary } from '@tupaia/ui-components';
+import { prefetchTransformSchemas, useSearchTransformSchemas } from '../../api';
+import { TransformSelectedOptionWithJsonEditor, TransformSelectedOption } from './component';
+
+const DATA_TYPES = {
+  TRANSFORM: 'Transforms',
+  ALIAS: 'Aliases',
+};
+
+const transformToValue = transform =>
+  transform.map(({ id, name, transform: code, ...restOfConfig }) => ({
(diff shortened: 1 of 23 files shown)


packages/report-server/src/routes/FetchTransformSchemaRoute.ts
```diff
@@ -20,15 +20,32 @@ export type FetchTransformSchemaRequest = Request<
...
+        if (typeof value === 'object') {
+          updatedFields[key] = removeRedundantConfigs(value);
+        }
+        const isEmptyArray = Array.isArray(value) && value.length === 0;
+        if (isEmptyArray) {
+          delete updatedFields[key];
+        }
+      });
+      return updatedFields;
+    };
```
billli0, line 38: Remove schema with empty oneOf array. It will cause error in `json-editor` validation.

packages/report-server/src/routes/FetchTransformSchemaRoute.ts
```diff
@@ -20,15 +20,32 @@ export type FetchTransformSchemaRequest = Request<
...
+
     const formattedTransformSchema = Object.entries(transformSchemas).map(
       ([transformKey, config]) => {
         const { fields } = config;
         return {
-          name: transformKey,
```
billli0, line 27: `name` is no needed.

packages/utils/src/__tests__/validation/DescribableLazy.test.js
```diff
@@ -20,20 +20,10 @@ describe('DescribableLazy', () => {
     );
     expect(paramsValidator.describe()).toStrictEqual({
       oneOf: [
-        {
-          oneOf: [
-            { const: 'human', type: 'string' },
-            { const: 'zombie', type: 'string' },
-          ],
-        },
+        { enum: ['human', 'zombie'] },
```
billli0, line 23: Use `enum` can show options in validation error in json editor as screenshot.

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/component/TransformSelectedOptionWithJsonEditor.js
```diff
@@ -0,0 +1,138 @@
...
+  .icon-wrapper {
+    cursor: pointer;
+  }
+  .MuiSvgIcon-root {
+    transition: transform 0.3s ease;
+    transform: rotate(${({ $expanded }) => ($expanded ? '0deg' : '-90deg')});
+  }
+`;
+
+export const TransformSelectedOptionWithJsonEditor = ({
```
rohan-bes, line 56: In fact, most of this component is almost identical to the `AggregateSelectedOptionWithJsonEditor`. Could we look at a way that we could refactor them into common code?

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/component/TransformSelectedOption.js
```diff
@@ -0,0 +1,54 @@
...
+        checkedIcon={<CheckBoxOutlinedIcon />}
+        checked={!option.isDisabled}
+        onChange={() => {
+          const newOption = { ...option };
+          newOption.isDisabled = !option.isDisabled;
+          onChange(newOption);
+        }}
+        disableRipple
+        size="small"
+      />
```
rohan-bes, line 38: I think this checkbox component is also common, no?

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/component/AggregateSelectedOptionWithJsonEditor.js
```diff
@@ -95,15 +95,15 @@ export const SelectedOptionWithJsonEditor = ({
       </FlexBetweenPanel>
       {isExpanded && (
         <JsonEditorPanel
-          onMouseOver={() => setEdittingOption(option.code)}
+          onMouseOver={() => setEdittingOption(option.id)}
           onMouseLeave={() => setEdittingOption(null)}
```
rohan-bes, line 99: I took me a really long time to work out what this `setEdittingOption` property was all about. It's just being used to control if the option is draggable or not? If that's the case can we rename it and have it behave a bit more clearly:
```suggestion
          onMouseOver={() => setIsDragDisabled(true)}
          onMouseLeave={() => setIsDragDisabled(false)}
```

rohan-bes (changes requested):
Nice work @billli0 really love how this feature is coming together!

A few comments about how we could reduce the amount of code by re-using in a few places, and some renaming suggestions.

Also, in playing around on the branch I noticed that the default values for all the fields are null (eg. `groupBy: null`). I was wondering if we could have these default to an empty value for the appropriate data type, eg. empty string for string keys?

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/component/AggregateSelectedOptionWithJsonEditor.js
billli0: Yeah it makes it clearer. Thanks!

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/component/TransformSelectedOptionWithJsonEditor.js
billli0: Have refactored it to reuse the code. Nice call!

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/TransformDataLibrary.js
```diff
@@ -0,0 +1,119 @@
...
+  }));
+
+export const TransformDataLibrary = ({ transform, onTransformChange, onInvalidChange }) => {
+  const value = transformToValue(transform);
+
+  const [dataType, setDataType] = useState(DATA_TYPES.TRANSFORM);
+  const [inputValue, setInputValue] = useState('');
+  const [isDisabledAll, setIsDisabledAll] = useState(
+    value ? !value.some(option => !option.isDisabled) : false,
+  );
```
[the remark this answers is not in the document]
billli0: Good catch!

billli0:
@rohan-bes Thanks for the review! Latest update:
1. The code has been refactored; 
2. All fields are able to set default value based on it's type.
3. Drawer is added.
4. title and description can be set.

packages/ui-components/src/components/DataLibrary/DataLibrary.js
```diff
@@ -100,12 +136,24 @@ export const DataLibrary = ({
     onChange(event, newValue);
   };
 
+  const handleDrawerOpen = () => {
+    setOpen(true);
+  };
+  const handleDrawerClose = () => {
+    setOpen(false);
+  };
```
rohan-bes, line 144: I really like how you broke these out, makes it super readable 👍

rohan-bes (approved):
Thanks so much @billli0! This is really exciting stuff 🎉

packages/admin-panel/src/VizBuilderApp/components/DataLibrary/component/SelectedOptionWithJsonEditor.js
```diff
@@ -86,7 +89,11 @@ export const SelectedOptionWithJsonEditor = ({
         >
           <DownArrow />
         </DownArrowIconWrapper>
-        <BaseSelectedOption option={basicOption} onRemove={onRemove} />
+        <EditableSelectedOption
```
rohan-bes, line 92: I have a feeling we might wanna make this config driven, through some kind of `supportsTitleEditing` property? Currently we can edit titles of aggregations as well which doesn't quite feel right. What do you think?

billli0: Yep we can add a feature flag in this case.

rohan-bes (approved):
Awesome work @billli0 so cool!

PR #3746, opened 2022-03-04
RN-395 Swap incorrect parent/child relation

EMcQ-BES:
### Issue #: RN-395

I probably shouldn't submit against other team's tickets but it was small and that's where Georgia made the ticket

### Changes:

- Move get via parent from `GETUserAccount` to `GETAccessRequest`
- Update route to use `GETAccessRequest`


edmofro (approved):
Fast work!

PR #3747, opened 2022-03-04
RN-396: Remove failed access policies  from the cache if they fail to build

rohan-bes:
### Issue RN-396:

This is a pretty obvious problem when you look at it. We always cache the first time we try to build the access policy for a user, however if that fails for any reason (eg. database connection timeout) we keep that cached result. That can be seen happening here: https://beyondessential.slack.com/archives/C01JD93J4E6/p1646349708959339

This seems to be a fairly general case problem. In the past, we've made a conscious effort to try and cache promises rather than results (in order to avoid many requests for the same expensive resource). However the natural downside of doing that is that we don't know if the promise resolves successfully or not...

Happy to chat with the reviewer and decide if this should be fixed in other areas (namely `DatabaseModel.js`)

### DIFF
--- a/packages/auth/src/AccessPolicyBuilder.js
+++ b/packages/auth/src/AccessPolicyBuilder.js
@@ -47,11 +47,17 @@ export class AccessPolicyBuilder {
     }
 
     const cacheKey = getCacheKey(userId, useLegacyFormat);
-    if (!this.cachedPolicyPromises[cacheKey]) {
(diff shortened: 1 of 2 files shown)


edmofro (approved):
Great find, fix, and test case! Yeah heck this is a pretty general problem throughout our codebase huh...seems like a bit of an ad hoc effort to launch into fixing all the places now, but definitely something we should create a card for, and also worth raising at dev chat so that people can avoid this in future

PR #3751, opened 2022-03-04
Switched to building frontend modules in serial rather than parallel

rohan-bes:
Building in parallel caused out of memory errors on feature instances

### DIFF
--- a/packages/web-frontend/package.json
+++ b/packages/web-frontend/package.json
@@ -12,7 +12,7 @@
   "main": "dist/index.js",
   "scripts": {
     "analyze": "source-map-explorer build/static/js/main.*",
-    "build": "rm -rf dist && run-p \"build:* {@}\" --",
+    "build": "rm -rf dist && run-s \"build:* {@}\" --",
     "build:desktop": "SKIP_PREFLIGHT_CHECK=true BUILD_PATH='./build/desktop/' GENERATE_SOURCEMAP=false react-scripts build",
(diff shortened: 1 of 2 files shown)


packages/web-frontend/package.json
```diff
@@ -12,7 +12,7 @@
   "main": "dist/index.js",
   "scripts": {
     "analyze": "source-map-explorer build/static/js/main.*",
-    "build": "rm -rf dist && run-p \"build:* {@}\" --",
+    "build": "rm -rf dist && run-s \"build:* {@}\" --",
```
rohan-bes, line 15: We are still using `run-p` for most other `build` commands, as they're just building types and source code. Figured it was safe to leave them, but an argument could be made to switch to `run-s` for all I suppose...

nick-beyondessential (approved):
LGTM

scripts/bash/getDeployablePackages.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 
-echo "admin-panel" "lesmis" "psss" "web-frontend" "admin-panel-server" "entity-server" "lesmis-server" "meditrak-server" "psss-server" "report-server" "web-config-server"
+echo "admin-panel" "lesmis" "psss" "web-frontend" "report-server" "admin-panel-server" "entity-server" "lesmis-server" "meditrak-server" "psss-server" "web-config-server"
```
rohan-bes, line 3: Turns out this is also an issue. Now `admin-panel-server` depends on `report-server` so they need to be built in the correct order. I think when we get `lerna` these kinds of issues will resolve themselves, right @IgorNadj ?

IgorNadj: Yeah there's a few circular dependancies at the moment which are untangled in the followup PR.

PR #3752, opened 2022-03-06
RN-388: Add map tab to mobile

edmofro:
### Issue #:
RN-388 (subtask of RN-373)

### Changes:
- Add tabs to navigate between dashboard and map overlay on the root screen
- Deep linked using a query parameter (makes sense to live in the path, but it's already tightly structured and would break things on desktop)

Screenshots on [the ticket](https://linear.app/bes/issue/RN-388#comment-0431fd4f)


packages/web-frontend/src/screens/mobile/RootScreen/index.js
```diff
@@ -26,21 +34,66 @@ const EntityName = styled.p`
...
+          TabIndicatorProps={{
+            style: { display: 'none' },
+          }}
+        >
+          <StyledTab label="Dashboard" value="dashboard" disableRipple />
+          <StyledTab label="Map" value="map" disableRipple />
+        </StyledTabs>
+        <SearchBar />
+        {selectedTab === 'dashboard' && <Dashboard />}
+        {selectedTab === 'map' && <ComingSoon>Coming soon</ComingSoon>}
```
rohan-bes, line 89: Love the use of this 'Coming Soon' place holder 👍 Out of curiosity are you planning on releasing with that at any point, or just using it internally on the RN-373 branch?

rohan-bes (approved):
Looks great to me!

One consideration that has occurred to me is about how we might handle links that work across desktop/mobile. Now if a user shares a link from desktop where they wish to highlight the map overlay data, the link won't have a `tab` param, and thus if the receiver is on mobile they won't see it without switching tabs. Not a major issue, and I don't necessarily see an easy way around it, but I thought I might mention it... 

Perhaps we add a `share` button in the desktop map overlay section which add the correct parameters to the url for mobile..?

packages/web-frontend/src/screens/mobile/RootScreen/index.js
edmofro: Yeah that's just internal for now - it's added in a later PR :)

edmofro:
Good point about the URLs not being isomorphic! It could be pretty hard to tell what the desktop user intends to share - were you thinking there'd be a separate share button on the map overlay somehow, and another on each dashboard report? It's a cool idea :-)

PR #3755, opened 2022-03-06
WAITM-108: add logos to footers

tcaiger:
### Changes:

- Add logos to homepage and general footer

---


packages/lesmis/src/views/HomeView.js
```diff
@@ -107,12 +114,16 @@ const Footer = styled(MuiContainer)`
...
-const FooterInner = styled(FlexEnd)`
+const FooterInner = styled(FlexSpaceBetween)`
   color: white;
   border-top: 1px solid rgba(255, 255, 255, 0.1);
   padding-top: 1rem;
   padding-bottom: 1rem;
 
+  @media screen and (min-width: 600px) {
+    padding-left: 1.5rem;
+  }
```
edmofro, line 125: What's this bit doing?

edmofro (approved):
lgtm! The logos themselves look a bit ugly, so if we could get nicer assets that have backgrounds that work better with LESMIS it would improve things, but obviously that's not on you

edmofro:
@tcaiger could target the release directly with this I reckon, so that it goes with tomorrow's merge to master per https://linear.app/bes/issue/WAITM-108#comment-2e1858fd

packages/lesmis/src/views/HomeView.js
tcaiger: It's aligning the logos on desktop screens to line up with the info text. The layout could be shuffled around so that the padding is using the same number as the 50px in the main container above but since it's a one off layout I am happy to have this specific little bit in here for now.

tcaiger:
> lgtm! The logos themselves look a bit ugly, so if we could get nicer assets that have backgrounds that work better with LESMIS it would improve things, but obviously that's not on you

Yes - they are quite ugly so hopefully we can swap them out soon.

PR #3758, opened 2022-03-07
Hotfix: Release 2022-9 - Fix report-server not responding to requests

rohan-bes:
The issue here is that the `next()` function was never being called when the middleware was appended to the request. As a result requests were never actually being handled.

### DIFF
--- a/packages/report-server/examples.http
+++ b/packages/report-server/examples.http
@@ -6,6 +6,12 @@
 # see: https://marketplace.visualstudio.com/items?itemName=humao.rest-client#environment-variables
 @authorization  = Basic {{email}}:{{password}}
 
+
+### Test
+GET {{baseUrl}}/test HTTP/1.1
+content-type: {{contentType}}
+Authorization: {{authorization}}
+
 ### Fetch PSSS Weekly Report
 GET {{baseUrl}}/fetchReport/PSSS_Weekly_Report?organisationUnitCodes=TO&hierarchy=psss&period=2020 HTTP/1.1
 content-type: {{contentType}}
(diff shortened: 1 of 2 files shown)


packages/report-server/examples.http
```diff
@@ -6,6 +6,12 @@
 # see: https://marketplace.visualstudio.com/items?itemName=humao.rest-client#environment-variables
 @authorization  = Basic {{email}}:{{password}}
 
+
+### Test
+GET {{baseUrl}}/test HTTP/1.1
+content-type: {{contentType}}
+Authorization: {{authorization}}
```
rohan-bes, line 13: Always handy to have this, as it allows us to easily check if the server is available

packages/report-server/src/app/createApp.ts
```diff
@@ -5,14 +5,19 @@
 
 import { MicroServiceApiBuilder, handleWith } from '@tupaia/server-boilerplate';
 import { TupaiaDatabase } from '@tupaia/database';
+import {
+  ForwardingAuthHandler,
+  getBaseUrlsForHost,
+  LOCALHOST_BASE_URLS,
+  TupaiaApiClient,
+} from '@tupaia/api-client';
```
rohan-bes, line 13: Cross pacakge `@tupaia` imports should come before package internal imports

packages/report-server/src/app/createApp.ts
```diff
@@ -31,6 +36,7 @@ export function createApp() {
       };
       req.ctx = context;
       res.ctx = context;
+      next();
```
rohan-bes, line 39: This was the culprit!

edmofro (approved):
Stamped!

PR #3759, opened 2022-03-07
Hotfix: Release 2022-9 - Fixed login failing on PSSS due to poorly loaded .env vars

rohan-bes:
This was a very strange issue and tbh I still don't quite understand why this was happening.

Whenever we attempted to login to PSSS, we'd send across an auth header of `Basic dW5kZWZpbmVkOnVuZGVmaW5lZA==` which decodes to username and password as `undefined`. This would fail to authenticate and the user couldn't log in.

The auth header is meant to be defined by 2 .env vars:
`MEDITRAK_API_CLIENT_NAME` and `MEDITRAK_API_CLIENT_PASSWORD`

Both of these .env vars are present in the PSSS .env file, with correct values. PSSS is loading its .env vars _**exactly**_ the same as how lesmis does. When logging out the .env vars with a console log they had the correct values.

So I'm pretty stumped, but applying these changes fixed the issue:
1. Correctly listing `dotenv` in psss-server's `package.json`
2. Restructuring the code a little so that we rebuild the auth header at request time rather than when the server first starts. This is slightly inefficient, but it's fine and feels a little safer anyway

### DIFF
--- a/packages/psss-server/package.json
+++ b/packages/psss-server/package.json
@@ -37,6 +37,7 @@
     "body-parser": "^1.18.3",
     "client-sessions": "^0.8.0",
     "cors": "^2.8.5",
+    "dotenv": "^8.2.0",
     "express": "^4.16.2",
     "winston": "^3.2.1"
   }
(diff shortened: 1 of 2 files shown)


billli0 (approved):
Emm that looks very strange.. Approve as long as it works! Thanks for the fix man.

edmofro:
Weird! And weird that it would just start failing now 😕

rohan-bes:
@edmofro we only just started using this code in psss-server this release, so it is a new code path. But yeah it's been working fine for lesmis-server forever, so not sure what the difference is...

PR #3760, opened 2022-03-07
Import multiple dataElement into data source table

billli0:
### Issue #:
RN-372 Import multiple dataElement into data source table

Limitation: Do not update existing data source since it might have existing answer in tupaia or being used in reports.

Furthermore, we need to discuss the permission to add data element after WAITP-1045 (permission update in data source).


packages/meditrak-server/src/apiV2/import/importDataSources/createDataSources.js
```diff
@@ -0,0 +1,30 @@
...
+
+      const exsitingDataSource = await transactingModels.dataSource.find({ code });
+      if (exsitingDataSource.length > 0) {
+        throw new Error(`Data source code '${code}' already exists`);
+      }
+
+      await transactingModels.dataSource.create(dataSource);
+    } catch (e) {
+      throw new ImportValidationError(e.message, excelRowNumber);
+    }
```
edmofro, line 28: Rather than a try/catch here, I'd just directly throw `ImportValidationError` in the two places they're thrown. Otherwise other errors (such as a db error during the `create` query) would also be thrown as a validation error, despite not relating to validation

packages/database/src/modelClasses/DataSource.js
```diff
@@ -10,10 +10,13 @@ import { TYPES } from '../types';
 const DATA_ELEMENT = 'dataElement';
 const DATA_GROUP = 'dataGroup';
 const SYNC_GROUP = 'syncGroup';
+const DATA_LAKE = 'data_lake';
+
 const DATA_SOURCE_TYPES = {
   DATA_ELEMENT,
   DATA_GROUP,
   SYNC_GROUP,
+  DATA_LAKE,
```
edmofro, line 19: Should it be a `DATA_SOURCE_TYPE`?

edmofro (changes requested):
Nice clean work here @billli0!

Unfortunately it's a bit awkwardly timed, as @EMcQ-BES's permissions change splits the data source table in two: `data_element` and `data_group`, which would have an impact on the core functionality here. I'm happy for this to go in ahead of that, because I know it'll help out with getting changes to the data lake processed quickly, but I don't want to push the merging work onto Ethan now he's focused on Tamanu. Once this is merged to dev, could you do the merge from dev into [this PR](https://github.com/beyondessential/tupaia/pull/3558) and rewrite the bits of this that need to account for the split in tables?

billli0:
> Thanks Biao! Unfortunately because the `try` is surrounding all the other errors, they will be captured and turned into `DatabaseError`. Good call to capture `DatabaseError`, but you'd need the `try` to just be around the specific db queries. It's probably not totally necessary, as they're rare and will ultimately bubble up to an `InternalServerError` anyway, so up to you whether to try/catch or just leave them uncaught.

Thanks @edmofro, it looks right that `InternalServerError` will handle the job. I'll leave them uncaught.

packages/meditrak-server/src/apiV2/import/importDataSources/importDataSources.js
```diff
@@ -0,0 +1,49 @@
...
+      await createDataSources(transactingModels, dataSources);
+    });
+    respond(res, { message: 'Imported data sources' });
+  } catch (error) {
+    if (error.respond) {
+      throw error; // Already a custom error with a responder
+    } else {
+      throw new DatabaseError('importing data sources', error);
+    }
+  }
```
billli0, line 48: @edmofro If `DatabaseError` will eventually be `InternalServerError`, I think we don't need try catch here?

edmofro: True! Yep I would remove it

billli0:
I made a few stupid mistakes in this PR, thanks @edmofro for spotting. Once this PR is released, I will merge from dev into Ethan's PR.

edmofro (approved):
That would be great, thanks @billli0

PR #3764, opened 2022-03-08
Fix cypress not found when running e2e

IgorNadj:
Cypress installation requires postinstall hook to run

This fixes cypress being installed correctly. ~But it also removes the convenience behaviour for local development of rebuilding internal dependancies after running yarn install.~

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "migrate-create": "yarn workspace @tupaia/database migrate-create",
     "migrate-down": "yarn workspace @tupaia/database migrate-down",
     "package-json-lint": "npmPkgJsonLint --configFile .npmpackagejsonlintrc.json .",
+    "postinstall": "test -n \"$SKIP_BUILD_INTERNAL_DEPENDENCIES\" || yarn build:internal-dependencies",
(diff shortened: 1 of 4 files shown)


rohan-bes:
Great find! But yeah that is annoying :\

I think it'll be safe but do you think it'd break any of the build processes that may have relied on the internal deps already being build after the `yarn install`?

It looks like we had a `postinstall` script previously? Do we know why it's suddenly stopped working?

IgorNadj:
Ah yeah in the tsc PR (maybe scope creeped a bit...) I did a little clean up, removing our custom flag `SKIP_BUILD_INTERNAL_DEPENDENCIES` and replacing it with `--ignore-scripts` to achieve the same result. However I didn't know that --ignore-scripts affects other packages too... whoops.

AFAIK there's nowhere that relies on the behaviour of internal dependencies being built via a postinstall hook. But to be safe, I'm happy to put SKIP_BUILD_INTERNAL_DEPENDENCIES back in for now and do the rework in the lerna PR.

rohan-bes (approved):
Nice, yep lets fix it in the lerna PR 👍

PR #3774, opened 2022-03-11
RN-390: Mobile overlay library

edmofro:
### Issue #:
RN-390


packages/web-frontend/src/actions.js
```diff
@@ -98,7 +98,6 @@ export const TOGGLE_INFO_PANEL = 'TOGGLE_INFO_PANEL';
 export const TOGGLE_LOCATION_ITEM_EXPANDED = 'TOGGLE_LOCATION_ITEM_EXPANDED';
 export const OPEN_USER_DIALOG = 'OPEN_USER_DIALOG';
 export const CLOSE_USER_DIALOG = 'CLOSE_USER_DIALOG';
-export const TOGGLE_MEASURE_EXPAND = 'TOGGLE_MEASURE_EXPAND';
```
edmofro, line 101: Was unused, just cleaning up because I noticed it

packages/web-frontend/src/components/MapOverlayHierarchy/MapOverlayHierarchy.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
+import { HierarchyItem } from './HierarchyItem';
+
+export const MapOverlayHierarchy = ({
```
edmofro, line 9: This is pulling out lines [149-204](https://github.com/beyondessential/tupaia/pull/3774/files#diff-ec94902c8f0c53a834ef9b6e3a00222079b45acc2e735ea8b165e6c1fd4cb64dL149-L204) from MapOverlayBar.js so they can be shared by the mobile version

packages/web-frontend/src/containers/MapOverlayBar/MapOverlayBar.js
```diff
@@ -129,79 +125,6 @@ const MapOverlayBarComponent = ({
     onSetMaxSelectedOverlays(maxNumOfSelectedOverlays);
   };
 
-  const renderDefaultMapOverlay = () => {
```
edmofro, line 132: Felicity suggested we get rid of the default map overlay showing at the top, as it's often confused for representing the currently selected map overlay

packages/web-frontend/src/screens/mobile/RootScreen/index.js
```diff
@@ -72,11 +72,23 @@ const RootScreen = ({
     setSelectedTab,
   ]);
 
+  // maintain the header height in state, so the overlay div can sit below it
```
billli0, line 75: It makes sense to use useRef() here. While I was trying things, I've removed the `useEffect` and the overlay still sit below the header. Just wondering what will make it overlapped with header?

billli0 (approved):
Great work and thanks for do some clean up! Just have a question and a small request, but can give a pre-approve.

packages/web-frontend/src/containers/MapOverlayBar/MapOverlayBar.js
edmofro: Good call! `selectDefaultMapOverlayCode` is still used elsewhere, but got rid of the first

packages/web-frontend/src/screens/mobile/RootScreen/index.js
edmofro: You're such a good reviewer going to that effort! This isn't used to make sure the overlay sits below the header, but that the overlay fills the remaining space on the screen, but without creating a scrollbar. 
image
This was actually a little tricky to get right, especially on safari for iOS, so this is a little bit of a compromise solution. There were a couple of options but they all seemed worse:
- Have it always set to `min-height: 100vh;`. This would make sure it filled the screen area, but would overfill it and create a scrollable area, which is a bit weird (though much much simpler code, so I almost went with it)
- Use a react portal here which fills the space, and render the library into that. I tried this too, and it felt like overkill - involved more complex code that what we end up with here

billli0:
> Have it always set to min-height: 100vh;. This would make sure it filled the screen area, but would overfill it and create a scrollable area,

@edmofro Sorry I might just be annoying to you, but what does 'a scrollable area' mean? I couldn't tell the different between `min-height: 100vh;` and `min-height: calc(100vh - ${p => p.$appHeaderHeight}px);`

edmofro:
Haha yeah I was kinda keen on leaving the scrolling area, but Tom talked me into prioritising the user rather than the code in this case. I'll add that comment!

edmofro:
@billli0 I've also posted https://github.com/beyondessential/tupaia/pull/3786 for you to take a look at the option using a portal. It has the same result as this, but the responsibility for positioning the modal below the header and filling the screen is all contained at the root. It does result in a bit more complexity. Interested to hear which you feel is more clear as an outside reader :-)

billli0:
@edmofro Portal is really cool! Thanks for posting a separate PR to guide me through. 

I agree it will make code more complex. But I think once you know Portal is rendering components, such as modals, tooltips, popup messages, the code will speak for itself. 

Perhaps [portal from material-ui](https://mui.com/components/portal/) can reduce the code. I'm ok with either way using Portal.

edmofro:
@billli0 thanks for pointing out the mui portal api, I like that a bit more as it feels more like regular react. Have changed to that and merged the portal stuff in here :-)

packages/web-frontend/src/screens/mobile/RootScreen/index.js
```diff
@@ -72,8 +95,18 @@ const RootScreen = ({
...
-  // below it
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [modalElement, setModalElement] = useState(null);
+  const useModal = useCallback(
+    () => [
+      ({ children }) => modalElement && <Portal container={modalElement}>{children}</Portal>,
+      setIsModalOpen,
+    ],
+    [modalElement],
+  );
```
billli0, line 106: Just realise this is a custom useState, that's pretty cool!

billli0 (approved):
Approve with the love of your passion to produce cleaner and more readable code!

PR #3776, opened 2022-03-11
RN-394: Convert @tupaia/data-api package to typescript

rohan-bes:
### Issue RN-394:

Prior to refactoring out common code between `@tupaia/data-api` and `@tupaia/data-lake-api` it is easiest if they're both written in typescript. Have avoided doing any refactors here, and just converted to typescript.


packages/data-api/jest.config.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import baseConfig from '../../jest.config-ts.json';
+
+module.exports = async () => ({
+  ...baseConfig,
+  rootDir: '.',
+  setupFilesAfterEnv: ['../../jest.setup.js', './jest.setup.ts'],
+  preset: 'ts-jest',
```
rohan-bes, line 12: Switched to using `ts-jest` for running tests as now we're no longer using `babel` to compile typescript packages

packages/data-api/scripts/patchMvRefresh.ts
```diff
@@ -3,13 +3,14 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
+// @ts-ignore
```
rohan-bes, line 6: `db-migrate` has no types :(

packages/data-api/package.json
```diff
@@ -12,12 +12,10 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "build": "rm -rf dist && run-p \"build:* {@}\" --",
-    "build:src": "babel src --out-dir dist --source-maps --ignore \"src/__tests__/**\" --config-file \"../../babel.config.json\"",
-    "build:types": "tsc",
+    "build": "tsc -p tsconfig-build.json",
```
rohan-bes, line 15: We are using a separate `tsconfig` file when building. This is because when linting/compiling we want to include the `__tests__` directory, but we want to exclude them from the build

packages/data-api/src/TupaiaDataApi.ts
```diff
@@ -188,25 +214,37 @@ export class TupaiaDataApi {
...
-      try {
-        // options coming from question.options are JSON format strings
-        // options coming from option_set are actual JSON objects
-        const optionObject = typeof option === 'string' ? JSON.parse(option) : option;
-        const { value, label } = optionObject;
+      if (typeof option === 'string') {
+        try {
+          // options coming from question.options are JSON format strings
+          // options coming from option_set are actual JSON objects
+          const optionObject = optionsObjectValidator.validateSync(JSON.parse(option));
```
rohan-bes, line 238: Still feel a bit awkward that this logic uses exception catching like it does... And technically the behaviour has slightly changed by using a validator here as it will throw an exception if the expected fields aren't present... Might rethink this...

packages/data-api/src/__tests__/testFetchDataElements.test.ts
```diff
@@ -6,14 +6,9 @@ import { getTestDatabase } from '@tupaia/database';
...
 import { TupaiaDataApi } from '../TupaiaDataApi';
 
-export const testFetchDataElements = () => {
+describe('fetchDataElements()', () => {
   const api = new TupaiaDataApi(getTestDatabase());
 
-  it('throws an error with invalid parameters', async () => {
-    await expect(api.fetchDataElements()).toBeRejectedWith(/data element codes/);
-    await expect(api.fetchDataElements(null)).toBeRejectedWith(/data element codes/);
-  });
```
rohan-bes, line 15: No longer need these kinds of tests, thanks to TS doing type checking 😄

packages/data-api/src/__tests__/testFetchDataGroup.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+  });
+
+  it('returns a data group with data elements in the correct format', async () => {
+    await expect(
+      api.fetchDataGroup('BCDTEST', ['BCD1TEST', 'BCD57TEST'], {
+        includeOptions: false,
+      }),
+    ).resolves.toStrictEqual(
+      expect.objectContaining({
```
rohan-bes, line 25: Only just learnt you can nest jest matchers like this, it's super handy! 🎉

packages/data-api/src/AnalyticsFetchQuery.ts
```diff
@@ -65,8 +76,66 @@ const supportsConfig = (aggregationSwitch, config) => {
...
+
+type QueryAggregation = {
+  config: { entityMap: Record<string, string> } & Record<string, unknown>;
+  switch: AggregationSwitch;
+  stackId: number;
+};
+
 export class AnalyticsFetchQuery {
-  constructor(database, options) {
+  private readonly database: TupaiaDatabase;
```
IgorNadj, line 131: SO GOOD to see private! Yay encapsulation :)

packages/data-api/src/TupaiaDataApi.ts
rohan-bes: Decided to leave it for now as it feels appropriate... happy to hear reviewer's thoughts though 👍

packages/data-api/src/__tests__/testFetchDataElements.test.ts
IgorNadj: Wouldn't this be validation on the api interface itself? Rather than inside-the-package type checking? `fetchDataElements()` is a public method.

Validation code which throws the error:
https://github.com/beyondessential/tupaia/pull/3776/files#diff-afd345b68ba1a10d317d519770e12d7674c23a54f14a37f90945c6c263b7e013R77

packages/data-api/src/SqlQuery.ts
```diff
@@ -0,0 +1,38 @@
+/**
```
IgorNadj, line 1: Minor: it looks like Github isn't smart enough to know this is a renamed file, it sees it as an Added file and a Deleted file. To make it easier to review, maybe in the future we should do this in two separate commits, one for the rename only, and the other to make changes?

packages/data-api/src/TupaiaDataApi.ts
```diff
@@ -6,18 +6,24 @@
...
+  hasOwnProperties,
+} from './utils';
 import { validateEventOptions, validateAnalyticsOptions } from './validation';
 import { sanitiseFetchDataOptions } from './sanitiseFetchDataOptions';
 
 const EVENT_DATE_FORMAT = 'YYYY-MM-DDTHH:mm:ss';
 
-const buildEventDataValues = resultsForEvent =>
-  resultsForEvent.reduce(
+const buildDataValuesFromAnswers = (answersForEvent: EventAnswer[]) =>
```
IgorNadj, line 20: Minor: similarly, this looks like a great rename for clarity. Love it, but would be easier to have this separate refactor be in a separate commit if possible.

packages/data-api/src/TupaiaDataApi.ts
IgorNadj: Yeah I don't really get the throw-catch logic... maybe pop your thinking into the code for a future cleanup, otherwise happy with it.

packages/data-api/src/__tests__/testFetchDataGroup.test.ts
IgorNadj: Oh, very cool!

IgorNadj (approved):
Thank you for doing this! So happy more of our critical data-flow path is typed! Just one comment on a test removal, otherwise LGTM!

packages/data-api/src/AnalyticsFetchQuery.ts
rohan-bes: I love to using accessors too! It's such a joy when your editor only shows you public methods for auto completion 😻 

Kostas and I chatted about this and we were both quite keen to turn on [this rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md) that enforces using them.

I'll add it to the dev-chat agenda since I think it affects the tamanu folks as well, but I'd be happy for us to start enforcing it 👍

packages/data-api/src/SqlQuery.ts
rohan-bes: Ah good point. Tbh I always do my reviews as bulk rather than by commit, so I don't often think about the reviewer's perspective like that, but I'll try to be more mindful from now on 👍

packages/data-api/src/AnalyticsFetchQuery.ts
IgorNadj: Yeah that would be great! We probably need to add linting to CI first :)

packages/data-api/src/__tests__/testFetchDataElements.test.ts
rohan-bes: That's a good point! I think if our entire codebase were typescript, then I'd argue it'd be worth removing that validation code and just making the type signature a bit stricter. But since this gets called from `@tupaia/data-broker` which is still js, I'll re-add the tests back 👍

IgorNadj: Makes sense, more motivation to go full TS :)

packages/data-api/src/SqlQuery.ts
IgorNadj: Yeah me too, but in this case even in full-change view it shows SqlQuery twice, so commit by commit might be easier. We don't typically rename files like this so not a biggie.

PR #3777, opened 2022-03-11
Merge from dev to wai-1045 permission update

billli0:
### Changes:

- Example

---

### DIFF
--- a/packages/meditrak-server/src/apiV2/import/importDataElements/createDataElements.js
+++ b/packages/meditrak-server/src/apiV2/import/importDataElements/createDataElements.js
@@ -1,0 +1,30 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { ImportValidationError } from '@tupaia/utils';
+
+export async function createDataElements(transactingModels, dataElements) {
+  const codes = [];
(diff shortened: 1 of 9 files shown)


packages/meditrak-server/src/apiV2/import/importDataElements/createDataElements.js
```diff
@@ -0,0 +1,30 @@
...
+    if (codes.includes(code)) {
+      throw new ImportValidationError(
+        `Data element code '${code}' is not unique in the uploaded file`,
+        excelRowNumber,
+      );
+    }
+    codes.push(code);
+
+    const exsitingDataElements = await transactingModels.dataElement.find({ code });
+    if (exsitingDataElements.length > 0) {
```
EMcQ-BES, line 24:
```suggestion
    const existingDataElements = await transactingModels.dataElement.find({ code });
    if (existingDataElements.length > 0) {
```

EMcQ-BES (approved):
I didn't take too deep of a look at the specific logic, since it's already been reviewed for dev. But the changes for the table split look fine to me

I did notice a typo though.

packages/meditrak-server/src/apiV2/import/importDataElements/createDataElements.js
billli0: Good spotting! Thanks @EMcQ-BES .

PR #3589, opened 2021-12-10
RN-115 cleanup typescript

IgorNadj:
### Changes:

- Reconfigure
  - Ts packages build using tsc
  - All packages have "build" script
  - Consolidate build scripts for js and ts
- Fix type errors
- Fix dependencies
  - Remove server-boilerplate dependancy on api-client to prevent typescript import circular dependancy
  - Api client no longer imports types without depending on the package
- CI
  - No longer downloads env vars, not needed  
- Misc 
  - Remove some legacy code paths in viz builder

### Release tasks
- [ ] Remove CI specific env vars from lastpass

PR #3778, opened 2022-03-11
Fix viz export

IgorNadj:
Fixes issue where admin-panel was exporting map overlay vizes without `reportPermissionGroup`, which would then fail validation on import.

I broke this accidentally in https://github.com/beyondessential/tupaia/pull/3589. Also adds in some flags for local dev.


chris-pollard (approved):
LGTM!

IgorNadj:
Replaced by https://github.com/beyondessential/tupaia/pull/3779

PR #3786, opened 2022-03-17
Modal as a portal

edmofro:


packages/web-frontend/src/screens/mobile/RootScreen/index.js
```diff
@@ -60,6 +61,28 @@ const StyledTab = styled(Tab)`
   max-width: none;
 `;
 
+const ModalDiv = styled.div`
```
edmofro, line 64: Also played with some css to make it animate in from the bottom

PR #3793, opened 2022-03-22
RN-413: Switch to using ts-jest for all typescript tests

rohan-bes:
### Issue RN-413:

We made the switch to compiling all typescript packages using typescript a little while ago. But we were still running the tests using `babel-jest`. `ts-jest` is the [preferred option](https://kulshekhar.github.io/ts-jest/docs/babel7-or-ts/) here, and now we can completely remove `babel` from all our typescript packages 🎉

Pretty much the same change set here for all packages:
- Add `tsconfig-build.json` which excludes the `__tests__` directory, make the `yarn build` command use this config file
- Change the default `tsconfig.json` to no longer emit anything, and to include the `__tests__` directory and other root level typescript files
- Rename/add `jest.config.ts` and possibly `jest.setup.ts`
- Fixup any typescript or linting errors in the `__tests__` directory
- Remove `.babelrc` file


packages/indicators/src/__tests__/__integration/integration.test.ts
```diff
@@ -24,7 +24,7 @@ interface TestCase {
   throws?: boolean;
 }
 
-const models = getTestModels();
+const models = (getTestModels() as unknown) as ModelRegistry;
```
rohan-bes, line 27: eslint-typescript forces this awkward casting as unknown when it can't easily coerce the types. I think until we switch the `@tupaia/database` package to typescript we'll have to stick with something like this...

IgorNadj (approved):
Heck yeah!

PR #3794, opened 2022-03-23
RN-369 new type gauge viz

billli0:
### Issue #:
RN-369


packages/ui-components/src/components/Chart/GaugeChart.js
```diff
@@ -0,0 +1,162 @@
...
+          <Label
+            content={props =>
+              getContent({
+                x: props.viewBox.cx,
+                y: props.viewBox.cy,
+                value: formatDataValueByType({ value: elements[0].value }, 'percentage'),
+                fontSize: 30,
+              })
+            }
+          />
```
billli0, line 120: To add a customised label, other properties (position, value, etc) in Label will be skipped.

billli0, line 120: To add a customised label, other properties (position, value, etc) in Label will be skipped. We have to manually specify it at content component.

packages/ui-components/src/components/Chart/GaugeChart.js
```diff
@@ -0,0 +1,162 @@
...
+    for (let i = 0; i < numOfElements; i++) {
+      elements.push({ value: denominator });
+      cellComponents.push(<Cell key={`cell-${i}`} fill="#eaeaea" />);
+    }
+
+    return { elements, cellComponents };
+  };
+
+  const { elements, cellComponents } = useMemo(() => generateElements(), [data]);
+  const height = useMemo(() => getHeight(isExporting, isEnlarged), [isExporting, isEnlarged]);
```
edmofro, line 70: Really nice use of memoization

packages/ui-components/src/components/Chart/GaugeChart.js
```diff
@@ -0,0 +1,162 @@
...
+
+  const responsiveStyle = !isEnlarged && !isMobile() && !isExporting ? 1 : 1.5;
```
edmofro, line 72: Positive conditions are generally easier to understand than negative
```suggestion
  const responsiveStyle = isEnlarged || isMobile() || isExporting ? 1.5 : 1;
```

packages/ui-components/src/components/Chart/GaugeChart.js
```diff
@@ -0,0 +1,162 @@
...
+  const innerRadius = 60 * responsiveStyle;
+  const outerRadius = innerRadius * 1.4;
+
+  const getContent = useCallback(({ value, x, y, fontSize }) => {
```
edmofro, line 76:
```suggestion
  const getLabelContent = useCallback(({ value, x, y, fontSize }) => {
```

packages/ui-components/src/components/Chart/GaugeChart.js
```diff
@@ -0,0 +1,162 @@
...
+              const fontSize = 15;
+              return getContent({
+                x: props.viewBox.cx - outerRadius + (outerRadius - innerRadius) / 2,
+                y: props.viewBox.cy + fontSize,
+                value: 0,
+                fontSize,
+              });
+            }}
+          />
+          {/* 100% label on the left */}
```
edmofro, line 133:
```suggestion
          {/* 100% label on the right */}
```

packages/ui-components/src/components/Chart/GaugeChart.js
```diff
@@ -0,0 +1,162 @@
...
+          paddingAngle={2}
+          isAnimationActive={!isExporting && isEnlarged}
+          onClick={item => {
+            onItemClick(item.originalItem);
+          }}
+          {...restOfConfigs}
+        >
+          {cellComponents}
+          {/* Coordinate graph for Gauge chart: https://beyond-essential.slab.com/posts/gauge-chart-2f6rrxnj#h5p3r-how-to-add-labels-into-pie-chart-in-rechart */}
+          {/* Value label on the center */}
```
edmofro, line 110: I recommend we put this JSX between the left and right labels, so the JSX reads in the same order as the visual elements

packages/ui-components/src/components/Chart/GaugeChart.js
```diff
@@ -0,0 +1,162 @@
...
+          data={elements}
+          paddingAngle={2}
+          isAnimationActive={!isExporting && isEnlarged}
+          onClick={item => {
+            onItemClick(item.originalItem);
+          }}
+          {...restOfConfigs}
+        >
+          {cellComponents}
+          {/* Coordinate graph for Gauge chart: https://beyond-essential.slab.com/posts/gauge-chart-2f6rrxnj#h5p3r-how-to-add-labels-into-pie-chart-in-rechart */}
```
edmofro, line 109: Helpful diagram!! So cool you documented that on slab

packages/ui-components/src/components/Chart/GaugeChart.js
edmofro: I see now the left and right sit lower than the central label, so happy to be pushed back on this

packages/ui-components/src/components/Chart/GaugeChart.js
```diff
@@ -0,0 +1,162 @@
...
+  const { viewContent, isExporting, isEnlarged, onItemClick } = props;
+  const { data, color = BLUE, ...restOfConfigs } = viewContent;
+
+  const generateElements = () => {
+    const denominator = 0.05;
+    const elements = [...data];
+    const cellComponents = [<Cell fill={color} />];
+    const numOfElements = Math.floor((1 - data[0].value) / denominator);
+
+    for (let i = 0; i < numOfElements; i++) {
```
edmofro, line 61: If we wanted this to be perfectly accurate, we would add an additional little slice to represent the remainder of the division. However, I reckon this will be visually "good enough"!

edmofro (changes requested):
Nice code and well explained throughout! Just a couple of very minor suggestions

packages/ui-components/src/components/Chart/GaugeChart.js
billli0: I feel that the remainder will make it looks a bit awkward because of the gap. Should be good enough to leave it.
image

packages/ui-components/src/components/Chart/GaugeChart.js
billli0: That makes sense!

packages/ui-components/src/components/Chart/GaugeChart.js
edmofro: Fair call!

edmofro (approved):

PR #3795, opened 2022-03-23
RN-387: De-genericise ControlBar

edmofro:
While looking at the desktop styles, I noticed that the ControlBar was made in a generic way but is now only being used by the SearchBar, so there was room to simplify it.


rohan-bes (approved):
So much red 😍 

Nice work!

PR #3796, opened 2022-03-23
RN-387: Improve entity hierarchy desktop styles

edmofro:
I noticed that the change to a full black entity hierarchy looked a bit wrong on desktop, so this changes it to have a more subtle background.


rohan-bes (approved):
Nice!

PR #3799, opened 2022-03-24
Update fields if data element existed

billli0:
Issue #:
RN-419

Changes:
Currently we block the update if code existed. If we can update the existing date elements in bulks, we can reuse the data elements for lesmis in the datalake and the existing reports.

### DIFF
--- a/packages/meditrak-server/src/apiV2/import/importDataSources/createDataSources.js
+++ b/packages/meditrak-server/src/apiV2/import/importDataSources/createDataSources.js
@@ -22,9 +22,9 @@ export async function createDataSources(transactingModels, dataSources) {
 
     const exsitingDataSource = await transactingModels.dataSource.find({ code });
     if (exsitingDataSource.length > 0) {
-      throw new ImportValidationError(`Data source code '${code}' already exists`, excelRowNumber);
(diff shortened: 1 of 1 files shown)


packages/meditrak-server/src/apiV2/import/importDataSources/createDataSources.js
```diff
@@ -22,9 +22,9 @@ export async function createDataSources(transactingModels, dataSources) {
 
     const exsitingDataSource = await transactingModels.dataSource.find({ code });
     if (exsitingDataSource.length > 0) {
-      throw new ImportValidationError(`Data source code '${code}' already exists`, excelRowNumber);
+      await transactingModels.dataSource.update({ code }, dataSource);
+    } else {
+      await transactingModels.dataSource.create(dataSource);
     }
```
billli0, line 28: Surprisingly found that code in data_source table is not unique. So we can't use .createOrUpdate(). I believe we can resolve this by Ethan's work on splitting data source table.

IgorNadj (approved):

billli0:
Sorry @Klaus-S I didn't mention on the card. We don't have to merge the latest dev since it will merge into prod.

PR #3800, opened 2022-03-24
Use the latest image, even on redeploy

edmofro:
This not only meant that if we deleted old images, we couldn't redeploy to servers based on those images, but _we weren't actually including the latest security patches on production_ 🤦


rohan-bes (approved):
LGTM, I'll test it and merge if it works 👍

PR #3803, opened 2022-03-25
Rn-410 display two different types of overlays separately 

billli0:
### Issue #:
RN-410 

### Changes:
1. Refactor `Map.js`, Each new components are all extracted from `Map.js`
2. Use `displayedMapOverlayCodes.map()` to avoid combining two overlays.


packages/web-frontend/src/containers/Map/Map.js
```diff
@@ -188,38 +133,11 @@ class MapComponent extends Component {
...
-        {displayedChildren?.map(area => (
-          <InteractivePolygon
-            area={area}
-            key={area.organisationUnitCode}
-            organisationUnitChildren={getChildren(area.organisationUnitCode)}
-            isChildArea
-            {...basicPropsForInteractivePolygon}
-          />
+        {displayedMapOverlayCodes.map(displayedMapOverlayCode => (
+          <DataVisualsLayer displayedMapOverlayCode={displayedMapOverlayCode} />
```
billli0, line 137: Instead of selecting/combining two sets of measure data and series, here we manage to handle it separately.

packages/web-frontend/src/containers/Map/DataVisualsLayer/DataVisualsLayer.js
```diff
@@ -0,0 +1,108 @@
+/*
```
IgorNadj, line 1: Thanks for doing this refactor, so much cleaner now! Great to break up big components into smaller ones.

Minor: I can guess at what the refactor is from the code change, but would be beneficial to name the commit something like "Refactor: extract Map marker and polygon rendering out into separate component", makes it easier to review. No need to change it this time though!

packages/web-frontend/src/containers/Map/Map.js
```diff
@@ -188,38 +135,21 @@ class MapComponent extends Component {
...
         <ZoomControl sidePanelWidth={sidePanelWidth} />
         <DemoLand />
-        {currentOrganisationUnit && organisationUnitIsArea(currentOrganisationUnit) && (
-          <InteractivePolygon
-            area={currentOrganisationUnit}
-            organisationUnitChildren={getChildren(currentOrganisationUnit.organisationUnitCode)}
-            isActive
-            {...basicPropsForInteractivePolygon}
-          />
+        {areMeasuresOnTheSameEntityLevel ? (
```
IgorNadj, line 138: Would be good to add a comment here for why we do this. It's not obvious to me why in the case of the visuals, all being on the same entity level is different.

IgorNadj (approved):
LGTM, one minor comment

packages/web-frontend/src/containers/Map/DataVisualsLayer/DataVisualsLayer.js
billli0: Good point. Can do better naming on the commit.

edmofro:
@billli0 is this one ready for a re-review?

billli0:
@edmofro It has been reviewed but it is testing failed. Need to fix it later. If it is urgent I can have a look now.

edmofro:
Not urgent but never great to leave it to go too stale, so whenever you next have a gap it would make sense to finish off :-)

billli0:
Agree, thanks for reminding!

PR #3807, opened 2022-03-28
Add underline as invalid chars in branch name

billli0:
### Issue #:
Underline in branch name will fail to tear down an instance.


edmofro (approved):
Thanks for adding the rule!

PR #3814, opened 2022-03-29
RN-439: Entity list performance optimisations

edmofro:
### Issue #:
- RN-439

### Changes:
There are two optimisations here:
- Search the entity list locally within the component, so that we don't have to run the permissions check every time we search from the db (which was the longest part)
- Don't render all of the entities, instead limit it to the first 100 results. This is the key change which fixes the lag, but comes at the expense of not all entities showing up when you first hit the entity menu, so users will have to type in a search term if their entity doesn't appear after scrolling. Ideally we should indicate this somehow, but for a quick fix I think it's good enough


rohan-bes (approved):
Nice changes! Thanks for picking this up 🙏 

Initially when I looked at the code I was most scared by the hefty looking sort function, but as always with react it seems things were more affected by the number of components rendered 😅 

Couple of things to note:
1. I couldn't comment here since it's outside the PR scope, but reducing the `initialNumToRender` on the `FlatList` component might be good. Accord to their docs, it should be '[enough to fill a screen but not much more](https://reactnative.dev/docs/flatlist#initialnumtorender)'
2. We can probably implement something for the [onEndReached](https://reactnative.dev/docs/flatlist#onendreached) or similar to get scroll loading functionality working so that users can see all entities if they keep scrolling (not now but in the next release?)

edmofro:
@rohan-bes thanks for pushing this through the deployment process - that's harder than fixing the bug!

> 1. I couldn't comment here since it's outside the PR scope, but reducing the `initialNumToRender` on the `FlatList` component might be good. Accord to their docs, it should be '[enough to fill a screen but not much more](https://reactnative.dev/docs/flatlist#initialnumtorender)'

I actually played around with some of the parameters here, and despite making changes, it still seemed to be rendering all (or at least absolutely loads) of the rows - i.e. it was calling the `renderEntityCell` function with thousands of different ids, no matter what parameters I set.

> 2. We can probably implement something for the [onEndReached](https://reactnative.dev/docs/flatlist#onendreached) or similar to get scroll loading functionality working so that users can see all entities if they keep scrolling (not now but in the next release?)

That's a great idea!

PR #3815, opened 2022-03-29
MAUI-697: Fix end date offset

chris-pollard:
### Changes:
- Add now function in Utils package
- End date defaults to now
- Offset start date from 'today' in new 'from' param


packages/report-server/src/__tests__/reportBuilder/query/QueryBuilder.test.ts
```diff
@@ -82,23 +82,33 @@ describe('QueryBuilder', () => {
...
-    endDate: '2020-12-31',
+    endDate: '2020-12-15',
     ...input,
   });
 
   const getPeriodString = (
     [startYear, startMonth]: [number, number],
     [endYear, endMonth]: [number, number],
+    endDay?: number,
+    startDay?: number,
```
rohan-bes, line 95: Personally I'd prefer to use this syntax, which allows passing an array of numbers of length 1,2, or 3
```suggestion
    [startYear, startMonth, startDay]: [number, number?, number?],
    [endYear, endMonth, endDay]: [number, number?, number?],
```

However it has one drawback which seems like a bug in Typescript to me tbh. It treats the type of `startMonth` and `startDay` as `number` when  really they should be `number | undefined` imo... It's fine cos we can just manually check for them in the function but still...

packages/report-server/src/__tests__/reportBuilder/query/QueryBuilder.test.ts
```diff
@@ -82,23 +82,33 @@ describe('QueryBuilder', () => {
...
   ) => {
     const periods = [];
 
     for (let year = startYear; year <= endYear; year++) {
       const currentStartMonth = year === startYear ? startMonth : 1;
       const currentEndMonth = year === endYear ? endMonth : 12;
       for (let month = currentStartMonth; month <= currentEndMonth; month++) {
-        periods.push(`${year}${month.toString().padStart(2, '0')}`);
+        if (month === currentEndMonth && year === endYear && endDay) {
+          for (let day = startDay || 1; day <= endDay; day++) {
```
rohan-bes, line 104: Just be a tad careful using the `||` operator here. If `startDay` were `0` then it'd still result it in being `1` here. While that works fine in this specific code so I guess its fine it is always a bit surprising 😮 Falsey behaviour is a bit risky

rohan-bes (changes requested):
Great stuff @chris-pollard ! Main thing is that we can probably do something different with the `getDefaultPeriod()` thing given it's nowhere else (maybe double check me on that) and just a couple small things. Love it 🎉

packages/utils/src/period/getDefaultPeriod.js
```diff
@@ -11,12 +11,12 @@ export const EARLIEST_DATA_DATE = utcMoment(EARLIEST_DATA_DATE_STRING);
...
 
 // Assemble a default date range using monthly granularity.
-export const getDefaultPeriod = () => {
+export const getDefaultPeriod = (granularity = 'month') => {
   const minimumStartDate = utcMoment().subtract(MAXIMUM_MONTHS_TO_LOOK_BACK, 'month');
   const startDateToUse =
     EARLIEST_DATA_DATE > minimumStartDate ? EARLIEST_DATA_DATE : minimumStartDate;
   const startDate = startDateToUse.date(1);
-  const endDate = utcMoment().endOf('month');
+  const endDate = utcMoment().endOf(granularity);
```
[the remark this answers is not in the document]
chris-pollard: Ah, nice. That makes it a bit cleaner.

chris-pollard:
@rohan-bes I've updated the branch based on your suggestions. Could you please take a look? Thanks.

packages/utils/src/period/getDefaultPeriod.js
```diff
@@ -16,7 +16,7 @@ export const getDefaultPeriod = () => {
   const startDateToUse =
     EARLIEST_DATA_DATE > minimumStartDate ? EARLIEST_DATA_DATE : minimumStartDate;
   const startDate = startDateToUse.date(1);
-  const endDate = utcMoment().endOf('month');
+  const endDate = utcMoment();
```
rohan-bes, line 19: Nice! One small request, given that we've deviated from the `getDefaultPeriod()` behaviour in web-config-server, can we add a comment either here or in that function explaining the difference. I think it's pretty easy for people to get confused otherwise 👍 

Even better could add a `@deprecated` [jsdoc](https://jsdoc.app/tags-deprecated.html) to the one in web-config-server!

packages/report-server/src/__tests__/reportBuilder/query/QueryBuilder.test.ts
```diff
@@ -82,23 +82,31 @@ describe('QueryBuilder', () => {
...
-        periods.push(`${year}${month.toString().padStart(2, '0')}`);
+        if (month === currentEndMonth && year === endYear && endDay > 1) {
+          for (let day = startDay; day <= endDay; day++) {
+            periods.push(
+              `${year}${month.toString().padStart(2, '0')}${day.toString().padStart(2, '0')}`,
+            );
+          }
+        } else {
+          periods.push(`${year}${month.toString().padStart(2, '0')}`);
+        }
```
rohan-bes, line 109: I think maybe the logic here is not _quite_ right. What would happen if someone called: 
`getPeriodString([2020, 11, 13], [2020, 12, 14])`?

Happy to pair with ya on this if you want, though I'll give a bit of a hint and suggest that you may need to use [this](https://stackoverflow.com/questions/1184334/get-number-days-in-a-specified-month-using-javascript)

yarn.lock
```diff
@@ -5274,17 +5274,7 @@
...
-  resolved "https://registry.yarnpkg.com/@typescript-eslint/eslint-plugin/-/eslint-plugin-2.34.0.tgz#6f8ce8a46c7dea4a6f1d171d2bb8fbae6dac2be9"
-  integrity sha512-4zY3Z88rEE99+CNvTbXSyovv2z9PNOVffTWD2W8QF5s2prBQtwN2zadqERcrHpcR7O/+KMI3fcTAmUUhK/iQcQ==
-  dependencies:
-    "@typescript-eslint/experimental-utils" "2.34.0"
-    functional-red-black-tree "^1.0.1"
-    regexpp "^3.0.0"
-    tsutils "^3.17.1"
-
-"@typescript-eslint/eslint-plugin@^4.18.0", "@typescript-eslint/eslint-plugin@^4.5.0":
+"@typescript-eslint/eslint-plugin@^2.25.0", "@typescript-eslint/eslint-plugin@^4.18.0", "@typescript-eslint/eslint-plugin@^4.5.0":
```
rohan-bes, line 5277: Kinda surprised to see this change honestly... But think it seems fine?

rohan-bes (changes requested):
Just a small request to fix up the test logic a bit. If you want to get this in asap feel free to bug me and I'll help ya with it 👍

yarn.lock
chris-pollard: I'm not sure how that came in, but if you think it's ok.

chris-pollard:
@rohan-bes I've reworked the getPeriodString function in the test suite. It's pretty bloated but it works. Would appreciate any refactoring tips.

chris-pollard:
@rohan-bes I think this reads a bit better. Let me know there's anything to adjust, cheers.

rohan-bes (approved):
Nice one! Great stuff @chris-pollard !! And thanks for being so patient with the review comments

Btw have you tried and tested this with the Tonga vizes? Did they end up needing anything for the timezone? (Can be a separate PR if they do)

PR #3817, opened 2022-03-30
MAUI-685: Change Single Value Data Colour

chris-pollard:
### Changes:

- Add use of config 'dataColor' for Single Value dashboard items, which changes the text color of the data.

### DIFF
--- a/packages/web-frontend/src/components/View/SingleValueWrapper.js
+++ b/packages/web-frontend/src/components/View/SingleValueWrapper.js
@@ -1,57 +1,46 @@
-/**
- * Tupaia Web
- * Copyright (c) 2019 Beyond Essential Systems Pty Ltd.
- * This source code is licensed under the AGPL-3.0 license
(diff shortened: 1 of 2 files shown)


tcaiger (changes requested):
Hey Crispy,  the change is technically correct but I'm going to be really annoying here and request that we take the chance to bring this component up to scratch with the modern conventions of the code base. It's quite an old component so there are lots of small changes. I have put them in [another PR](https://github.com/beyondessential/tupaia/pull/3821) to hopefully make it clearer. Please have a look and let me know if you have any questions or if you disagree with any of the suggestions. Also I didn't test my changes locally so if you end up agreeing with them, we'll need to test it as well.

chris-pollard:
@tcaiger the refactoring looks good, and it's working in my local. Do you want to merge into this branch?

tcaiger (approved):
Great - yep thanks for merging. Cheers

PR #3818, opened 2022-03-30
Hotfix: RN-440 - Fix IOS builds in AppCenter

rohan-bes:
### Issue RN-440:

Seems like IOS builds have been failing in AppCenter for the past couple months 😅 I think it's caused by this [bug in the AppCenter SDK](https://githubhot.com/repo/microsoft/appcenter-sdk-react-native/issues/955)

Fix was just to bump the AppCenter version to explicitly `^4.4.3` and the builds seem to be passing now


billli0 (approved):
LGTM. Thanks for the fix!

PR #3820, opened 2022-03-31
Maui-717: map entity type id to name

chris-pollard:
### Changes:
- Add function to parse analytics data, changing entity type responses from id to code.


edmofro:
We can close this right @chris-pollard? I believe the code to make it work has already made its way to master?

chris-pollard:
yes

PR #3821, opened 2022-03-31
refactor single value wrapper

tcaiger:
Issue [MAUI-685](https://linear.app/bes/issue/MAUI-685)

### DIFF
--- a/packages/web-frontend/src/components/View/SingleValueWrapper.js
+++ b/packages/web-frontend/src/components/View/SingleValueWrapper.js
@@ -1,57 +1,46 @@
-/**
- * Tupaia Web
- * Copyright (c) 2019 Beyond Essential Systems Pty Ltd.
- * This source code is licensed under the AGPL-3.0 license
- * found in the LICENSE file in the root directory of this source tree.
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
  */
-/**
- * SingleValueWrapper
- *
- * Renders view with single value from data provided by viewContent object
- * @prop {object} viewContent An object with the following structure
-   {
-    "type": "view",
-    "viewType": "singleValue",
-    "name": "Total Stock On Hand",
-    "value": 24063409.4
-  }
- * @return {React Component} a view with one value and its title
- */
-import React, { PureComponent } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
-import { VIEW_STYLES } from '../../styles';
-import { formatDataValue } from '../../utils';
+import styled, { css } from 'styled-components';
+import { VIEW_CONTENT_SHAPE } from './propTypes';
 import { ViewTitle } from './Typography';
+import { VIEW_STYLES, WHITE } from '../../styles';
+import { formatDataValue, isMobile } from '../../utils';
 
-export class SingleValueWrapper extends PureComponent {
-  render() {
-    const {
-      name,
-      valueType,
-      dataColor,
-      value,
-      total,
-      value_metadata: valueMetadata,
-    } = this.props.viewContent;
-    const metadata = valueMetadata || this.props.viewContent[`${name}_metadata`];
-    const { style } = this.props;
-    if (dataColor) {
-      VIEW_STYLES.data.color = dataColor;
-    }
+const DataWrapper = styled.div`
+  font-size: 50px;
+  font-weight: bold;
+  text-align: center;
+  color: ${WHITE};
 
-    return (
-      <div style={VIEW_STYLES.viewContainer}>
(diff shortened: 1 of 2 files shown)


packages/web-frontend/src/components/View/SingleValueWrapper.js
```diff
@@ -1,57 +1,46 @@
-/**
- * Tupaia Web
- * Copyright (c) 2019 Beyond Essential Systems Pty Ltd.
- * This source code is licensed under the AGPL-3.0 license
- * found in the LICENSE file in the root directory of this source tree.
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
```
tcaiger, line 3: This is the current copyright

packages/web-frontend/src/components/View/SingleValueWrapper.js
```diff
@@ -1,57 +1,46 @@
...
  */
-/**
- * SingleValueWrapper
- *
- * Renders view with single value from data provided by viewContent object
- * @prop {object} viewContent An object with the following structure
```
tcaiger, line 11: If we use prop types correctly then we don't also need this doc block and just having the one place with the shape of the viewContent is easier to maintain.

packages/web-frontend/src/components/View/SingleValueWrapper.js
```diff
@@ -1,57 +1,46 @@
...
 import PropTypes from 'prop-types';
-import { VIEW_STYLES } from '../../styles';
-import { formatDataValue } from '../../utils';
+import styled, { css } from 'styled-components';
+import { VIEW_CONTENT_SHAPE } from './propTypes';
 import { ViewTitle } from './Typography';
+import { VIEW_STYLES, WHITE } from '../../styles';
+import { formatDataValue, isMobile } from '../../utils';
+
+const DataWrapper = styled.div`
```
tcaiger, line 13: styled-components is now the convention. There are some advantages such as the styles being written in native css, better code autocomplete and more.

packages/web-frontend/src/components/View/SingleValueWrapper.js
```diff
@@ -1,57 +1,46 @@
...
-      value,
-      total,
-      value_metadata: valueMetadata,
-    } = this.props.viewContent;
-    const metadata = valueMetadata || this.props.viewContent[`${name}_metadata`];
-    const { style } = this.props;
-    if (dataColor) {
-      VIEW_STYLES.data.color = dataColor;
-    }
+  ${!isMobile() &&
```
tcaiger, line 19: Refactoring like this also saves me duplicating the core styles.

packages/web-frontend/src/components/View/SingleValueWrapper.js
```diff
@@ -1,57 +1,46 @@
...
-      <div style={VIEW_STYLES.viewContainer}>
-        <ViewTitle>{name}</ViewTitle>
-        <div style={{ ...VIEW_STYLES.data, ...(style || {}) }}>
-          {formatDataValue(value, valueType, { ...metadata, total })}
-        </div>
-      </div>
-    );
-  }
-}
+export const SingleValueWrapper = ({ viewContent, style }) => {
```
tcaiger, line 28: When working on old components that are the PureComponent class I try to refactor them into the new functional component pattern as it's new new convention (which is used because it's easier to read, is compatible with new react features and is more performant)

packages/web-frontend/src/components/View/SingleValueWrapper.js
```diff
@@ -1,57 +1,46 @@
...
+  const { name, valueType, dataColor, value, total, value_metadata: valueMetadata } = viewContent;
+  const metadata = valueMetadata || viewContent[`${name}_metadata`];
+
+  return (
+    <div style={VIEW_STYLES.viewContainer}>
+      <ViewTitle>{name}</ViewTitle>
+      <DataWrapper style={{ color: dataColor, ...style }}>
```
tcaiger, line 35: This should work out the same as before for determining which color is set. Eg. the hierarchy least important to most.
- the original css above (eg white)
- viewContent.dataColor if it's set
- the style prop if set on a component that uses this component

packages/web-frontend/src/styles/index.js
```diff
@@ -385,23 +385,6 @@ export const VIEW_STYLES = {
     lineHeight: '16px',
     textAlign: 'left',
   },
-  data: isMobile()
```
tcaiger, line 388: I checked and this is only used in our component so it's safe to move the styles into our file and delete this shared const. The advantage of having it locally defined is that we can change the styles without having to worrying about if it affects something else and it's just less cognitive load than switching between files.

PR #3825, opened 2022-04-01
RN-448 link after import

IgorNadj:
RN-448
---

### Screenshots:
https://linear.app/bes/issue/RN-448#comment-d100d940


edmofro (approved):
Nice clean solution! Given it's a UI change, we should probably follow the protocol of getting Felicity's review on it, but that depends on how quickly we need to get this out?

PR #3831, opened 2022-04-04
Bumped ts-eslint version to enforce access modifier

rohan-bes:
- Added modifiers to all existing ts classes
- Fixed up a couple linting errors here and there

### DIFF
--- a/packages/admin-panel-server/src/routes/FetchReportPreviewDataRoute.ts
+++ b/packages/admin-panel-server/src/routes/FetchReportPreviewDataRoute.ts
@@ -27,19 +27,19 @@ export type FetchReportPreviewDataRequest = Request<
     previewConfig: Record<string, unknown>;
     testData?: unknown[];
   },
-  { entityCode: string; hierarchy: string; previewMode?: PreviewMode, vizType: VizType; }
+  { entityCode: string; hierarchy: string; previewMode?: PreviewMode; vizType: VizType }
 >;
 
 export class FetchReportPreviewDataRoute extends Route<FetchReportPreviewDataRequest> {
   private readonly reportConnection: ReportConnection;
 
-  constructor(req: FetchReportPreviewDataRequest, res: Response, next: NextFunction) {
+  public constructor(req: FetchReportPreviewDataRequest, res: Response, next: NextFunction) {
     super(req, res, next);
 
     this.reportConnection = new ReportConnection(req.session);
   }
 
-  async buildResponse() {
+  public async buildResponse() {
     this.validate();
 
     const { entityCode, hierarchy } = this.req.query;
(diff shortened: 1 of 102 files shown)


packages/admin-panel-server/src/routes/FetchReportPreviewDataRoute.ts
```diff
@@ -95,16 +95,18 @@ export class FetchReportPreviewDataRoute extends Route<FetchReportPreviewDataReq
         draftDashboardItemValidator,
         draftReportValidator,
       );
-    } else if (vizType === VIZ_TYPE_PARAM.MAP_OVERLAY) {
+    }
+
+    if (vizType === VIZ_TYPE_PARAM.MAP_OVERLAY) {
```
rohan-bes, line 100: Linter complains about `else if` after a `return` as `if` would suffice

packages/api-client/src/connections/EntityApi.ts
```diff
@@ -37,7 +35,7 @@ const recurseFilter = (
     const value = filter.comparisonValue;
     filterArray.push([
       nestedKeys,
-      // @ts-ignore
```
rohan-bes, line 40: `@ts-ignore` gives a linting error, but `ts-expect-error` doesn't as long as we provide and explanation. So we should always use that 👍

packages/entity-server/src/routes/hierarchy/relationships/ResponseBuilder.ts
```diff
@@ -137,23 +146,12 @@ export class ResponseBuilder {
     return formattedEntitiesByCode;
   }
 
-  isDefaultFilter = (filter: EntityFilter) => {
```
rohan-bes, line 140: Cosmetic but in this case I felt the function made a bit more sense outside the class (as it didn't rely on class members and is kind of generic)

packages/data-lake-api/src/DataLakeDatabase.ts
```diff
@@ -3,9 +3,9 @@
  * Copyright (c) 2017-2020 Beyond Essential Systems Pty Ltd
  */
 
-// @ts-ignore
+// @ts-expect-error pg has not types
```
edmofro, line 6:
```suggestion
// @ts-expect-error pg has no types
```

packages/psss-server/src/routes/reports/ConfirmWeeklyReportRoute.ts
```diff
@@ -3,8 +3,10 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
edmofro, line 6: Why was this needed?

edmofro (approved):
Thanks for doing this chore! And cleaning up a few bits and pieces along the way, including some of those wasteful extra lines between fields 😀

packages/psss-server/src/routes/reports/ConfirmWeeklyReportRoute.ts
rohan-bes: Good question! I added `lodash.groupby` to the package dependencies instead 👍

PR #3835, opened 2022-04-05
Hot fix WAITP-1112: Sync permissions check broken

edmofro:
### Issue #:
WAITP-1112

### Changes:
Today's release had a change to a function that didn't update all callers of that function. This updates the caller that was missed


IgorNadj (approved):

edmofro:
Thanks @IgorNadj. Haven't tested this yet so just want to prove it works before releasing.

IgorNadj:
Good catch! I had no idea this code path was used by external apis, I assumed it was only admin panel. Sorry about that.

packages/meditrak-server/src/apiV2/import/importSurveyResponses/assertCanImportSurveyResponses.js
```diff
@@ -31,6 +31,9 @@ export const assertCanImportSurveyResponses = async (
       'code',
       surveyResponseCountryCodes,
     );
+    if (surveyResponseCountries.length !== surveyResponseCountryCodes.length) {
```
edmofro, line 34: @IgorNadj added this check because previously "Demo Land" didn't exist in tests, so they were (wrongly) making it through the permissions check

PR #3836, opened 2022-04-05
Setup local test database for running integration tests

rohan-bes:
Generally when running integration tests in Tupaia (ie. the tests that require a DB connection) devs have just been using their local DB instance. This has a few downsides:
- It's very slow, often timing out, for some tests as inserts/deletes can take a long time for larger tables
- It can pollute your local database if tests fail to clean themselves up properly
- Tests can behave differently if there's existing data in the database, which means the behaviour in local mightn't match CI/CD

This change is to update the `setupTestDatabase.sh` script which CI/CD uses to work locally, and then to switch the `test` command in all packages that use the database for their tests to use the test database.

~Merge task:~
- [x] ~Update .env vars to have the `TEST_DB_NAME` where needed~


codeship-services.yml
```diff
@@ -35,8 +35,9 @@ e2e:
 
 # 'db' service provides a postgres db with postgis
 db:
-  image: mdillon/postgis:10-alpine
+  image: postgis/postgis:13-3.2-alpine
```
rohan-bes, line 38: Switched to using [postgis image](https://registry.hub.docker.com/r/postgis/postgis/) over [mdillion image](https://hub.docker.com/r/mdillon/postgis/) as it looks like mdillion is no longer being maintained (last update was 3 years ago and has no postgres13 support).

codeship-services.yml
```diff
@@ -35,8 +35,9 @@ e2e:
...
   encrypted_dockercfg_path: ./packages/devops/ci/dockercfg.encrypted
+  encrypted_env_file: ./packages/devops/ci/db.env.encrypted
```
rohan-bes, line 40: The postgis image requires `PG_PASSWORD` env var be set

package.json
```diff
@@ -32,8 +32,8 @@
     "postinstall": "test -n \"$SKIP_BUILD_INTERNAL_DEPENDENCIES\" || yarn build:internal-dependencies",
     "refresh-database": "yarn workspace @tupaia/database refresh-database --root ../../",
     "test-all": "node ./scripts/node/testAll --silent",
-    "update-codeship-env-vars": "./packages/devops/scripts/ci/updateCiEnvVars.sh",
```
rohan-bes, line 35: The `./packages/devops/scripts/ci/updateCiEnvVars.sh` script no longer exists, so I removed this command

packages/api-client/package.json
```diff
@@ -9,7 +9,7 @@
     "build": "tsc -p tsconfig-build.json",
     "lint": "eslint . --ext .ts",
     "lint:fix": "yarn lint --fix",
-    "test": "jest --runInBand",
+    "test": "USE_TEST_DB=true jest --runInBand",
```
rohan-bes, line 12: Controlling whether to use the test database or the usual database everywhere using this `USE_TEST_DB` env var. Honestly I'm not too happy with it, ideally we'd just be passing through the `DB_NAME` env var to be something else.

However, there are some scripts that source the `.env` file, so we need logic that controls it that can't be overriden by that... This is the best I could come up with, any suggestions?

packages/database/scripts/setupTestDatabase.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash -e
```
rohan-bes, line 1: Moved from `packages/devops/scripts/ci/setupTestDatabase.sh`

package.json
edmofro: Thanks for tidying up!

packages/api-client/package.json
edmofro: Yeah this was my thought too. The scripts that source the `.env` file almost exclusively just need the db variables, so instead of calling `source .env` directly they could call a custom `sourceDbEnv.sh` which:
- Stores any that are already defined in custom variables, e.g. `EXISTING_DB_NAME=$DB_NAME`
- Sources .env
- Restores any that were existing

Changing the places that source .env to do this seems like no more added logic in specific places than the requirement to check for `USE_TEST_DB` in all those places.

It would also replace our existing strategy below
```
# if env vars are not already defined (e.g. by script caller during CI/CD), pull them in from .env
if [ "$DB_URL" == "" ]; then
    source .env
fi
```

edmofro: And actually it could be more straightforward to just store the whole environment temporarily to avoid having to declare each db variable explicitly, as in [this stack overflow answer](https://stackoverflow.com/a/69127685/1988532)

edmofro (changes requested):
Nice! Thanks for doing this - such a simple change really but solves a longstanding ticket (which I've just assigned to you: RN-463)

packages/database/scripts/setupTestDatabase.sh
```diff
@@ -0,0 +1,32 @@
...
+USE_TEST_DB=true yarn workspace @tupaia/data-api install-mv-refresh
+echo "Patching mvrefresh"
+USE_TEST_DB=true yarn workspace @tupaia/data-api patch-mv-refresh up
+echo "Installing Analytics table"
+USE_TEST_DB=true yarn workspace @tupaia/data-api build-analytics-table
+
+echo "Deleting migrations that target data modifications, as there is no data to migrate on the test database"
+rm ./src/migrations/*modifies-data.js
+USE_TEST_DB=true yarn migrate
+git checkout -- './src/migrations/*modifies-data.js'
```
edmofro, line 32: Clever addition 🎉

edmofro (commented):
I don't want to make the chore too much more onerous, but this could also do with a change to the readme or some CLI warnings if someone tries to test without having run `yarn setup-test-database`, or tries to setup the test database without defining the pg user credentials

edmofro:
Interestingly, I ran into the opposite on Tuesday, where tests were incorrectly passing on the test database, but failing (as they should have been - there was a bug) on a fully populated local database. I don't think this is common, but it does highlight that it can be hard to recreate realistic data in a test environment.

packages/database/scripts/setupTestDatabase.sh
rohan-bes: Thanks! In truth we really ought to do a check here to see if there are any edits existing on files that match the `./src/migrations/*modifies-data.js` pattern, as those edits would be lost by the `rm` and `checkout` actions. I'll see what I can do for that 👍

packages/api-client/package.json
rohan-bes: Thanks for the suggestion, it's much cleaner! Have gone with the 'store the whole environment temporarily to avoid having to declare each db variable explicitly' option 👍

packages/database/scripts/setupTestDatabase.sh
rohan-bes: Switched to just backing up the folder somewhere and re-instating it at the end of this script. Simpler and more robust 👍

rohan-bes:
> I don't want to make the chore too much more onerous, but this could also do with a change to the readme or some CLI warnings if someone tries to test without having run yarn setup-test-database, or tries to setup the test database without defining the pg user credentials

No this is a totally valid request! I've added a `check-test-database-exists` script which runs before the tests now that enforces we create it. Having a bit less luck finding an easy way to enforce users are using a non-trust auth method though... Will keep looking...

packages/data-api/scripts/dropAnalyticsMaterializedView.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash -e
-source .env
+
+# Use whatever existing .env vars have been specified
+curenv=$(declare -p -x)
+test -f .env && source .env
+eval "$curenv"
```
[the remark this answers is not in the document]
rohan-bes: Hello 🤖

edmofro (approved):
Who does this beyondessential-bot think he is, coming in and approving this from underneath me!

PR #3838, opened 2022-04-05
RN-439: Improve PrimaryEntity and Entity question UX and performance

edmofro:
### Issue #:
RN-439
Also fixes RN-465, RN-458 and RN-460

### Changes:
The hot fix we released last week works, but isn't the best user experience, because they can only see the first 100 items no matter how far they scroll. This fix instead makes use of FlatList's inbuilt optimisations to allow lazy rendering of all items as they scroll through the list. To do this, we needed to do a few things, each caused by the previous:
- Set a fixed height. This is because the FlatList sits within an infinite ScrollView, so it can't tell how long it is and therefore can't optimise rendering based on what is within its bounds. Simply doing this suddenly gives us all of the FlatList optimisations, and removes the lag without needing to limit the results.
- Hand over scroll control from the ScrollView to the FlatList. Once you set a fixed height, it doesn't seem to handle scroll within another scrollable thing. Additionally, we already had a bug reported (RN-458) that this wasn't working on some Android devices ever.
- Hide the full list of entities if it is on a page with more than one question, until the user starts to type a search term. This is because we can't work out whether the overall page or the internal entity results should have scroll control until the user is definitely interacting with the entity question (i.e. when they start typing in the search bar)
- Hide the search bar and results, and show just the selected entity once one is chosen. This is so we don't need to keep showing a scrollable list once the user has finished with the component, and can therefore hand scroll control back to the overall page. Additionally, we already had a ticket to make the selected entity more obvious (RN-460), see screenshots [in that ticket](https://linear.app/bes/issue/RN-460#comment-89f61313)


packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -119,54 +223,38 @@ export class EntityList extends PureComponent {
...
 }
 
 EntityList.propTypes = {
-  filteredEntities: PropTypes.array.isRequired,
+  filteredEntities: PropTypes.array,
   selectedEntityId: PropTypes.string,
   onRowPress: PropTypes.func.isRequired,
+  onClear: PropTypes.func.isRequired,
   onMount: PropTypes.func,
+  isOnlyQuestionOnScreen: PropTypes.bool.isRequired,
```
rohan-bes, line 238: I know it's a bit more code, but could we maybe move the isOnlyQuestionOnScreen prop up the stack by one to `EntityQuestion`? Then we could have it pass it in as a `handlesScrollControl` (or something better named) prop? It's cosmetic, but I feel the naming of this prop doesn't make the most sense in this component's context?

rohan-bes (approved):
Awesome stuff! Glad you got that FlatList rendering logic working properly 🎉 (tbh I'd be quite confused/worried when the slow load time bug came in since it really seemed like FlatList was made to handle such cases...)

Just a minor request around prop naming, but it's looking good to me!

packages/meditrak-app/app/entityMenu/EntityList.js
edmofro: I'm totally on board with that!

edmofro: Really happy with how that came out, also led me to make the handling of scroll control within the component a bit more contained and obvious

rohan-bes (approved):
Great stuff man!

PR #3840, opened 2022-04-07
MAUI-717: Change entity answer from id to name

chris-pollard:


edmofro:
Same here: We can close this right @chris-pollard? I believe the code to make it work has already made its way to master?

PR #3841, opened 2022-04-07
Maui 717 test inner join

chris-pollard:
### Changes:

- Example

---


edmofro:
And here: We can close this right @chris-pollard? I believe the code to make it work has already made its way to master?

PR #3843, opened 2022-04-07
MAUI-739: Enable specific start and end dates

chris-pollard:
### Issue MAUI-739
- add functionality to handle date strings in web-frontend


edmofro (approved):
Great change @chris-pollard! The only thing that could bring polish is validation of the date string format. We wouldn't want to a badly formatted date string to be interpreted as best possibly by moment, but come up with a different date to what the user expected. It also might be nice if the date strings are all one consistente format, even if moment would allow a wide range (e.g. for doing migrations or search the db and being able to count on a format, etc.). That's optional though, so this is approved as is :-)

packages/web-frontend/src/utils/periodGranularities.js
```diff
@@ -164,6 +164,13 @@ export const momentToDateString = (date, granularity, format) =>
  * @param {*} periodGranularity
  * @param {*} defaultTimePeriod
  */
+
+const validateDateString = date => {
+  if (!moment(date).isValid()) {
```
edmofro, line 169: Good test about an invalid date! That's even more important than what I was thinking, which is just around using a consistent format to represent the date. `moment()` will accept a wide range of formats, but with some it will be doing a "best guess", and anyway I think it'd be good for us to use a single consistent format. I think you could get this format checking by adding a single format parsing argument, but not totally sure
```suggestion
  if (!moment(date, 'YYYY-MM-DD').isValid()) {
```

edmofro (approved):
Nice, thanks!

chris-pollard:
Thanks for the feedback @edmofro, it's nice getting to know the moment package.

PR #3844, opened 2022-04-07
RN-452 Unable to download tonga testing site

billli0:
### Issue #:
RN-452

### Changes:
1. Enable raw data download to use report server
2. Updated Json: 
https://linear.app/bes/issue/RN-452#comment-86960c92

### Limitation:
1. 0/1 data need to transform to Yes/No in report server
2. We still need to use `surveyDataExport` as legacy report, but we fetch data from report server. Because data in placeholder in dashboard is required and it is different. See [line 16 in surveyDataExport.js](https://github.com/beyondessential/tupaia/blob/e313efa89ae2904a547fe380024b4fc04c592ff2/packages/web-config-server/src/apiV1/dataBuilders/generic/dataDownload/surveyDataExport.js#L16)
3. Currently answers as entity id are not able to be converted as entity name, because it is tricky to add its entities dependency in report server for `entityIdToName()`. But it will be solved with the migration Crispy is working on to manage that in analytics table. https://linear.app/bes/issue/MAUI-717/mapping-entity-type-dataelements-to-entity-name-in-survey-responses


rohan-bes (changes requested):
Nice work on this @billli0!

A few comments on things that seem a tad off, but overall looking good!

I did wanna chat though about what could be done in order to fully support export data reports from the report-server. It seems to me that if we just add a couple pieces of functionality we could achieve it? Namely:
1. A new 'output' type for 'rawDataExport'
2. The ability to make substitutions in the report config based on URL parameters

There's still some complexity around different reports per survey, but I feel it's worth talking things through. Obviously this is a bit of an urgent bug, so it's best if we just release this as is, but I think we should add the functionality soon!

packages/report-server/src/aggregator/ReportServerAggregator.ts
```diff
@@ -47,15 +47,28 @@ export class ReportServerAggregator {
...
     periodParams: PeriodParams,
-    dataElementCodes?: string[],
+    dataElementCodesInConfig?: string[],
   ): Promise<Event[]> {
     const { period, startDate, endDate } = periodParams;
+
+    const dataElementCodes =
+      ((!dataElementCodesInConfig || dataElementCodesInConfig.length === 0) &&
+        (await this.getDateElementCodes(programCode, organisationUnitCodes))) ||
+      dataElementCodesInConfig;
```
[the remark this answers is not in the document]
billli0: Yeah good point! I just found we got 5 reports that fetch without data elements. Will need to take care of those once this is deployed.

packages/web-config-server/src/apiV1/dataBuilders/generic/dataDownload/surveyDataExport.js
```diff
@@ -61,6 +65,87 @@ class SurveyDataExportBuilder extends DataBuilder {
...
+      requestQuery.endDate = endDate;
+    }
+
+    const surveyData = {};
+    await Promise.all(
+      surveys.map(async survey => {
+        const { reportCode, name: surveyName, codes, code } = survey;
+        if (surveys.includes(codes.toString() && surveys.includes(code))) {
+          return;
+        }
```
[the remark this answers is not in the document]
billli0: Ah what a mistake. Have updated.

packages/report-server/src/aggregator/ReportServerAggregator.ts
```diff
@@ -47,15 +47,28 @@ export class ReportServerAggregator {
     );
   }
 
+  private async getDateElementCodes(programCode: string, organisationUnitCodes: string[]) {
+    const { dataElements } = (await this.aggregator.fetchDataGroup(programCode, {
+      organisationUnitCodes,
```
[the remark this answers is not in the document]
billli0: Ah cool! It works without it. So `organisationUnitCodes` only needed for `dhis2`? I've seem it got passed in [here](https://github.com/beyondessential/tupaia/blob/c38ef1fc5dbf4f62830d2a335295c67042d8205c/packages/web-config-server/src/apiV1/dataBuilders/DataBuilder.js#L119)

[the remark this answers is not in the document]
rohan-bes: Yeah so I think it's just the `DhisService` that uses it to determine which DHIS2 instance to pull from. Given that we're looking to migrate away from doing that as of the data-services cleanup, I think it's fine for us to just not bother passing it in. If it causes issues later we can address that 👍

rohan-bes (changes requested):
Looking good! Thanks for making these changes. What do you think about taking a crack at decomposing the matrix builder stuff a bit so that we can share code without directly calling `matrix`?

packages/report-server/src/reportBuilder/output/functions/matrix/rawDataExport.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+import { ReportServerAggregator } from '../../../../aggregator';
+import { Row } from '../../../types';
+import { buildParams, matrix } from './matrix';
+
+export const buildRawDataExport = (params: unknown) => {
+  const builtParams = buildParams(params);
+  return (rows: Row[], aggregator: ReportServerAggregator) =>
+    matrix(rows, { ...builtParams, attachAllDataElementsToColumns: true }, aggregator);
```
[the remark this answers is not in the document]
billli0: Yeah that's a good point! We could make them share code instead of calling others.

packages/report-server/src/reportBuilder/reportBuilder.ts
```diff
@@ -51,11 +51,19 @@ export class ReportBuilder {
...
+    );
+    const transform = buildTransform(this.config.transform, transformContext);
     const transformedData = transform(data.results);
 
-    const output = buildOutput(this.config.output);
+    const outputContext = await buildOutputContext(
+      { ...this.config.output, ...this.config.fetch },
+      aggregator,
+    );
+    const output = buildOutput({ ...this.config.output, ...this.config.fetch }, outputContext);
```
rohan-bes, line 66: I find this line a little confusing, why are we combining the fetch and output params here?
```suggestion
    const output = buildOutput(this.config.output, outputContext);
```

packages/report-server/src/reportBuilder/context/buildOutputContext.ts
```diff
@@ -0,0 +1,67 @@
...
+    )
+  ).flat();
+  return { dataElementsMetadata };
+};
+
+const outputContextBuilders = {
+  rawDataExport: rawDataExportContextBuilder,
+};
+
+export const buildOutputContext = async (
```
rohan-bes, line 48: Hmmm I'm not too sure we actually want to go with the whole 'context' pattern for the output step. The reason we did it for the transform step is simply that we can't do async logic in the transforms, so we need to compute any async data upfront. 

But we can run the output step as async, so perhaps we just move this logic to get the `dataElementsMetadata` to the `rawDataExportBuilder` instead?

packages/report-server/src/reportBuilder/output/functions/matrix/matrix.ts
```diff
@@ -51,24 +51,26 @@ const paramsValidator = yup.object().shape(
             )
           : schema,
       ),
+    dataGroups: yup.array().of(yup.string()).notRequired(),
   },
   [['rowField', 'categoryField']],
 );
 
-const matrix = (rows: Row[], params: MatrixParams): Matrix => {
+export const matrix = (rows: Row[], params: MatrixParams): Matrix => {
```
rohan-bes, line 59: These exports are no longer needed yeah? Lets try and keep things encapsulated 👍

rohan-bes (changes requested):
Nice work @billli0, really love how as this is decoupled from the matrix stuff.

I think though that the output context pattern mightn't be needed here (we may add it down the line at some point). Wouldn't it be simpler if the rawDataExportBuilder just pulled all the dataElementsMetadata itself?

packages/report-server/src/reportBuilder/output/functions/rawDataExport/rawDataExport.ts
```diff
@@ -0,0 +1,42 @@
...
+  params: unknown,
+  outputContext: RawDataExportContext,
+  aggregator: ReportServerAggregator,
+): Promise<RawDataExport> => {
+  return new RawDataExportBuilder(rows, params, outputContext, aggregator).build();
+};
+
+const buildParams = (params: unknown): unknown => {
+  return params;
+};
```
billli0, line 29: Still wanna keep buildParams for consistent format over the places.

rohan-bes: Good call 👍

packages/report-server/src/reportBuilder/output/functions/rawDataExport/rawDataExportBuilder.ts
```diff
@@ -0,0 +1,85 @@
...
+  private async attachAllDataElementsToColumns() {
+    const dataElementValidator = yup
+      .array()
+      .of(
+        yup.object().shape({
+          code: yup.string().required(),
+          text: yup.string().required(),
+        }),
+      )
+      .required();
```
rohan-bes, line 59: If you are just using a validator here for typescript, you can also just add a jsdoc to the `fetchDataGroup` method on the base `Aggregator` class if you want 👍 Totally optional tho

rohan-bes (approved):
Nice work @billli0 ! Thanks for being so patient with the review. This is a great change 😄

PR #3848, opened 2022-04-08
Maui 717 entity type answer to name

chris-pollard:
### Issue MAUI-717

Changes:
- Add migration to update build_analytics_table to replace with entity type answer with name instead of id.


rohan-bes (approved):
Nice one @chris-pollard !

One q, did you ever check to see if renaming an entity will update the name in the analytics table? I think there is probably a way for us to get that working if it doesn't...

chris-pollard:
No didn't end up check that @rohan-bes

PR #3859, opened 2022-04-12
RN-466: Meditrak entity load performance improvement

edmofro:
### Issue #:
RN-466

### Changes:
Most of the time spent loading the entities from the db was in running a permissions check. After optimising that permissions check, I realised we actually didn't need it at all!

I would recommend looking at commit 7d4a6e0 on its own before reviewing this as a whole - that's the (very simple) meat and potatoes of the PR. The rest is just nice-to-have simplifications (mainly renaming filteredEntities to entities, because I always found it confusing that the full list of entities handed to the search component was called filteredEntities)

### DIFF
--- a/packages/meditrak-app/app/database/types/User.js
+++ b/packages/meditrak-app/app/database/types/User.js
@@ -22,24 +22,12 @@ export class User extends RealmObject {
 
   /**
    * Whether this survey is available to the given user. Checks whether that user has access to the
-   * permission group required by this survey.
+   * permission group required by this survey, in the selected country.
    */
-  hasAccessToSurveyInEntity(survey, entity) {
+  hasAccessToSurveyInCountry(survey, country) {
     const { permissionGroup } = survey;
     if (!permissionGroup) return false; // this survey is not fully synced yet, don't show it
-    return this.hasEntityAccess(entity, permissionGroup.name);
-  }
-
-  // User has access if they have access to the entity itself, or some ancestor entity
-  hasEntityAccess(entity, permissionGroupName = '') {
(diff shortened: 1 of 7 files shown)


packages/meditrak-app/app/database/types/User.js
```diff
@@ -22,24 +22,12 @@ export class User extends RealmObject {
...
-  hasEntityAccess(entity, permissionGroupName = '') {
-    const entities = [];
-    let currentEntity = entity;
-    while (currentEntity && currentEntity.type !== 'world') {
-      entities.push(currentEntity.code);
-      currentEntity = currentEntity.parent;
-    }
-
-    return this.accessPolicy.allowsSome(entities, permissionGroupName);
+    return this.accessPolicy.allows(country.code, permissionGroup.name);
```
rohan-bes, line 30: Ah the good ol' country level permissions check 🎉 

Part of me sweats that one day we'll be in real trouble when we actually do need to support more granular entity permissions. But for now it's the business logic, so this is a good change 👍

packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -91,10 +91,10 @@ export class EntityList extends PureComponent {
       return;
     }
 
-    const { filteredEntities } = this.props;
+    const { entities } = this.props;
```
rohan-bes, line 94: Great rename, I found this rather confusing as well. The EntityList doesn't need to know if the entities are filtered or not!

rohan-bes (approved):
Nice changes! LGTM

PR #3860, opened 2022-04-12
RN-404: Moved SqlQuery to @tupaia/database

rohan-bes:
### Issue RN-404:

One of possibly many cleanup PRs. Just moved the `SqlQuery` class into the `@tupaia/database` package so that it can be used across packages. A bit of a shame to migrate it back to JS, but with the JSdoc we can preserve all the typescript types 👍


edmofro (approved):
Imma approve this, but I do wonder about whether we need a different approach to sharing code between `data-api` and `data-lake`. Possibilities include:
- Another module, used by both
- One imports utilities from the other
- They get combined into a single module with some deeper namespacing

Because not all of the common code lives as obviously in an existing common module right?

In fact, even `SqlQuery`, while designed to be generic, was created for the specific use cases here, and it might be nice to keep it contained. I don't mind it going to `@tupaia/database`, but keeping it out of there would avoid it starting to be used more widely over the existing models, which would inevitably lead to its features and complexity increasing and eventually the maintenance of two database abstraction layers (catastrophising a bit here)

rohan-bes:
Thanks for raising the concerns on this one. I've had similar thoughts honestly. Initially I was gonna go with a bit more of a larger refactor and really tighten up the coupling between `data-api` and `data-lake-api`. However, the more I thought about it, the more I felt this coupling wasn't actually that appropriate. 

I felt like the idea of having some 3rd module, or somehow combining both packages into one would actually be more confusing than convenient. I like that in it's current state, if I need to debug issues in the `data-api`, there's a clear and obvious place to go look for it. And over time, I like the idea of the `data-lake-api` having the freedom to move away from the `data-api` without causing any painful refactors.

So for now I was planning on just refactoring out small common logic like this, and then cleaning up things where I can.

Definitely not 100% on this approach though, so if you feel a 3rd module, code sharing, or combining into one is a better approach please let me know.

As for not wanting to expose the `SqlQuery` too broadly. I definitely agree, though I feel it should be safe to rely on convention here rather than anything explicit? Also the static methods on `SqlQuery` could actually be re-used a bit throughout the `database` package as well!

edmofro:
All fair reasoning, am happy with that!

PR #3868, opened 2022-04-19
RN-439: Show recently used entities in entity selection questions

edmofro:
### Issue #:
Fixes the awkward UX Sima mentions [here](https://linear.app/bes/issue/RN-439#comment-e61bc161)

### Changes:
- Show a list of recently used entities, rather than pre-selecting the last used entity when starting a new survey

See the screenshot [here](https://linear.app/bes/issue/RN-439#comment-a332a845)

### DIFF
--- a/packages/meditrak-app/app/entityMenu/EntityList.js
+++ b/packages/meditrak-app/app/entityMenu/EntityList.js
@@ -128,7 +128,7 @@ export class EntityList extends PureComponent {
     }
 
     const { entities, recentEntities } = this.props;
-    if (recentEntities && recentEntities.length > 0) {
+    if (recentEntities?.length > 0) {
       return [
         { title: 'Recently used', data: recentEntities },
         { title: 'All entities', data: entities },
--- a/packages/meditrak-app/app/entityMenu/EntityList.js
+++ b/packages/meditrak-app/app/entityMenu/EntityList.js
@@ -5,7 +5,7 @@
 
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import { Dimensions, FlatList, StyleSheet, Text, View } from 'react-native';
+import { Dimensions, SectionList, StyleSheet, Text, View } from 'react-native';
 
 import { Icon, TextInput, TouchableOpacity } from '../widgets';
 import {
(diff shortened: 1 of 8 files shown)


packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -120,6 +121,22 @@ export class EntityList extends PureComponent {
...
   };
 
+  getListSections = () => {
+    const { searchResults } = this.state;
+    if (searchResults) {
+      return [{ data: searchResults }];
+    }
+
+    const { entities, recentEntities } = this.props;
+    if (recentEntities && recentEntities.length > 0) {
```
tcaiger, line 131:
```suggestion
    if (recentEntities?.length > 0) {
```
I'm not sure if mobile can use this but maybe it can?

edmofro: I had the idea it couldn't, but just tried and it works!

packages/meditrak-app/app/entityMenu/actions.js
```diff
@@ -60,9 +61,30 @@ export const loadEntitiesFromDatabase = (isPrimaryEntity, questionId) => (
...
+  recentEntityIds.forEach((id, index) => {
+    recentEntityIdToIndex[id] = index;
+  });
+  const recentEntities =
+    recentEntityIds.length === 0
+      ? []
+      : database
+          .getEntities({ ...filters, id: recentEntityIds })
+          .slice()
+          .sort((a, b) => recentEntityIdToIndex[a.id] - recentEntityIdToIndex[b.id]);
```
tcaiger, line 82: I found this bit quite confusing to get my head around. I think I was confused because recentEntityIdToIndex is created but it's not always used. Perhaps it could be done in a slightly more procedural way like this? But I think it's a pretty subjective preference.

```suggestion
  let recentEntities = [];

  if (recentEntityIds.length > 0) {
    const recentEntityIdToIndex = {};
    recentEntityIds.forEach((id, index) => {
      recentEntityIdToIndex[id] = index;
    });

    recentEntities = database
      .getEntities({ ...filters, id: recentEntityIds })
      .slice()
      .sort((a, b) => recentEntityIdToIndex[a.id] - recentEntityIdToIndex[b.id]);
  }
```

tcaiger (approved):
Looks good. The recent searches is a great UX improvement and all the changes look good to me. I let a comment but that is really just me working it out in my head so feel free to ignore that.

packages/meditrak-app/app/entityMenu/actions.js
edmofro: Good call. I've gone with a slightly different option that I think makes it more readable still. No point making code denser than it needs, so thanks for the push back!

PR #3873, opened 2022-04-19
MAUI-750: Allow mean function to discount undefined arguments

chris-pollard:
### Changes:
- add tests for 'mean'
- new overriding function for mean, removing undefined args from denominator.


IgorNadj (approved):
LGTM

PR #3878, opened 2022-04-20
RN-404: Switch to yup validation in data-api and data-lake-api

rohan-bes:
### Issue RN-404:

Removed the old validation using `ObjectValidator` in favour of `yup`.

Ended up adding this custom `yupTest` and `yupTestAny` wrapper for our existing object validator functions, so that we can re-use that functionality as it's pretty valuable. It is a bit awkward though, so happy to hear reviewer's thoughts.


packages/data-api/src/validators.ts
```diff
@@ -0,0 +1,51 @@
...
+    .test(
+      yupUtils.yupTestAny([hasNoContent, takesDateForm], 'startDate should be in ISO 8601 format'),
+    ),
+  endDate: yup
+    .string()
+    .test(
+      yupUtils.yupTestAny([hasNoContent, takesDateForm], 'endDate should be in ISO 8601 format'),
+    ),
+  dataElementCodes: yup.array().of(yup.string().required()).strict(),
+  dataGroupCode: yup.string().strict().required(),
```
rohan-bes, line 21: Needing to use `strict` here really surprised me. It turns out that by default `yup` will attempt to cast its input to the schema time (ie. `1` would be cast to `'1'` and then accepted by this validator).

I found this pretty unintuitive, and definitely gave me a bit of discomfort as I've probably not realised it was needed for other validators when using `yup`.

Overall, I'm a bit on the fence about `yup`. It's great to have an OOTB package to handle validation for us, and it plays fairly well with typescript. But this input modification, it's poor support for dynamic types, and its syntax aren't the best imo...

packages/data-api/src/validators.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+import { hasNoContent, takesDateForm, takesIdForm, yup, yupUtils } from '@tupaia/utils';
+
+export const eventOptionsValidator = yup.object().shape({
+  organisationUnitCodes: yup.array().of(yup.string().required()).strict().required(),
+  startDate: yup
+    .string()
+    .test(
+      yupUtils.yupTestAny([hasNoContent, takesDateForm], 'startDate should be in ISO 8601 format'),
```
rohan-bes, line 13: Introduced this `yupUtils.yupTestAny([...tests], message?)` function as `yup` has not support for a validator that passes if one of its tests passes. In theory we might want a `yupTestAll` function at some point, but happy to leave that for now.

packages/utils/src/validation/yup.js
```diff
@@ -126,11 +126,51 @@ const testSync = (schema, createError) =>
     },
   });
 
+/**
+ * Wrap a test function that throws an error upon failure for using in yup
+ * @param {(...args: any[]) => any} testFunction
+ * @param {string} [message]
+ * @returns
+ */
+const yupTest = (testFunction, message) => (...args) => {
```
rohan-bes, line 135: The motivation for adding this function is that `yup.test` wants to take a function that returns either a boolean or a `yup.ValidationError`.

We have a fairly comprehensive suite of validation checks in `utils/src/validation`. Unfortunately, these all throw an error if they fail, and they're designed to work with our `ObjectVadidator` class. That class is pretty heavily used still, so I didn't really want to touch anything. This is my workaround to be able to share that functionality.

packages/data-api/src/validators.ts
edmofro: Oh wow, that is surprising! There's no way I would have expected that. I think it's even worth highlighting at dev chat, as it's pretty widely used but I don't know how widely this quirk would be known

packages/data-api/src/validators.ts
```diff
@@ -0,0 +1,51 @@
...
+    .test(
+      yupUtils.yupTestAny([hasNoContent, takesDateForm], 'endDate should be in ISO 8601 format'),
+    ),
+  dataElementCodes: yup.array().of(yup.string().required()).strict().required(),
+  aggregations: yup
+    .array()
+    .of<typeof aggregationStringValidator | typeof aggregationObjectValidator>(
+      yup.lazy((value: unknown) =>
+        typeof value === 'string' ? aggregationStringValidator : aggregationObjectValidator,
+      ) as any,
```
edmofro, line 49: This is a bit of a mind boggler

edmofro (approved):
Thanks for doing this, it's good to be consolidating towards using one library (even if it does have some faults!)

packages/data-api/src/validators.ts
rohan-bes: Good point! I'll add it to the agenda!

packages/data-api/src/validators.ts
rohan-bes: Yeah I kinda hate it 😞 It's sort of a typescript error in `yup` where it can't properly infer the type of lazy validators when they're used in `yup.array().of()`. It's documented [here](https://github.com/jquense/yup/issues/1190), and seems like maybe it's been fixed in the current beta version?

rohan-bes: I'll add a comment and a TODO to upgrade yup once the fix goes out in a mainline release 👍

PR #3879, opened 2022-04-21
RN-458: Take scroll control when autocomplete in use

edmofro:
### Issue #:
RN-458

### Changes:
A UX change was required here, as we can only have either the autocomplete results _or_ the surrounding screen scrollable. 
See [here](https://linear.app/bes/issue/RN-458#comment-19a2299a) for a description with screenshots, but in brief:
- The autocomplete will now start "closed"
- Pressing the "down" arrow on the right, or just tapping in the searchbox will open the full set of options
- If you start typing, it filters the options and replaces the open/close arrow with a clear button
- Whenever the full list of options is open, it will be scrollable (but the screen around won't be)
- If you select an option or click the "up" button on the right, it will close and hand scroll control back to the surrounding screen


rohan-bes (approved):
LGTM

PR #3884, opened 2022-04-22
Use date-fns to manipulate date in transform

billli0:
### Issue #:
RN-447

### Changes:
- `date-fns` has been used in other places in codebase, it would be great to have it for data transforming. 
- Use `date-fns` to manipulate date in transform, there are plenty handy functions that we can use: https://date-fns.org/docs/Getting-Started

### Example:
- For calculating age 
-- Before:
`age = string(floor((date($currentDate.substring(0,10)).getTime() - date(string($DOB).substring(0,10)).getTime())/(1000 * 3600 * 24* 365)))`
-- Now:
`age = dateUtils().differenceInYears(date($currentDate), date($DOB))`
---

### DIFF
--- a/packages/expression-parser/src/expression-parser/ExpressionParser.js
+++ b/packages/expression-parser/src/expression-parser/ExpressionParser.js
@@ -4,6 +4,7 @@
  */
 
 import { create, all } from 'mathjs';
+import * as fns from 'date-fns';
 
 import { customFunctions } from './customFunctions';
 
@@ -49,6 +50,7 @@ export class ExpressionParser {
     this.math.import(this.getFunctionExtensions());
     this.math.import(this.getFunctionOverrides(), { wrap: true, override: true });
     this.validExpressionCache = new Set();
+    this.set('dateUtils', fns); // Support all FNS date modification functions. https://date-fns.org/v2.16.1/docs/Getting-Started
   }
 
   /**
--- a/packages/expression-parser/src/expression-parser/ExpressionParser.js
+++ b/packages/expression-parser/src/expression-parser/ExpressionParser.js
@@ -4,9 +4,9 @@
  */
 
 import { create, all } from 'mathjs';
-import * as fns from 'date-fns';
 
 import { customFunctions } from './customFunctions';
+import { customNamespaces } from './customNamespaces';
 
 /**
  * @typedef {Object} Scope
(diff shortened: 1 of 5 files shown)


packages/expression-parser/src/expression-parser/customFunctions.js
```diff
@@ -30,9 +32,12 @@ const translate = (value, translations) => {
...
 const date = (...argumentList) => new Date(...argumentList);
 
+const dateUtils = () => fns;
+
 export const customFunctions = {
   avg: average,
   firstExistingValue,
   translate,
   date,
+  dateUtils,
```
billli0, line 42: Tried to spread `...fns` in here but it has other functions as we do, such as `format`.

rohan-bes: Honestly I think namespacing behind `dateUtils` is a better move anway 👍 I've been thinking we may want to add namespacing to functions in expressions for a while, eg. `math.sum()`, `entity.codeToId()` or the like. Otherwise over time things will get messy :S

rohan-bes (approved):
Super cool stuff! Great addition @billli0 I love the creativity to add a whole libaries' worth of functions. Who knows, maybe in the future we can find all kinds of nifty js libraries to add 😄 

One request, can we please get a couple unit tests (just small ones) for this! Should be easy to extend the existing tests in expression-parser 👍

packages/expression-parser/src/expression-parser/customFunctions.js
billli0: That is a great idea! Actually it is super easy to support that. See the latest changes.

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -49,6 +50,7 @@ export class ExpressionParser {
     this.math.import(this.getFunctionExtensions());
     this.math.import(this.getFunctionOverrides(), { wrap: true, override: true });
     this.validExpressionCache = new Set();
+    this.set('dateUtils', fns); // Support all FNS date modification functions. https://date-fns.org/v2.16.1/docs/Getting-Started
```
rohan-bes, line 53: Small thing, but might be nice to put `dateUtils` into a `getNamespaces()` function (a bit like how we do for `getCustomFunctions()`. Just makes it a little clearer why we're injecting them into the scope, and makes extending this list a little easier 👍

PR #3895, opened 2022-05-02
RN-487: Rename meditrak-server

rohan-bes:
### Issue RN-487:

So we need to rename `meditrak-server` so that the name can be taken by the upcoming orchestration server for the `meditrak-app` (RN-486)

I've gone with `central-server` here. Honestly not entirely happy with the name, but the idea is for it to be a safe placeholder for the server which is responsible for miscellaneous behaviour and acting as a bit of a central coordination service. There's a [discussion in slack](https://beyondessential.slack.com/archives/C032DCHKTGR/p1651448334458359) about the naming and still very open to suggestions

#### Merge tasks:

- [ ] publish new `@tupaia/api-client` version
- [ ] update .env vars


packages/admin-panel-server/.env.example
```diff
@@ -3,8 +3,8 @@ DB_PASSWORD=
 DB_URL=
 DB_PORT=
 DB_USER=
+API_CLIENT_NAME=
+API_CLIENT_PASSWORD=
 ENTITY_API_URL=
-MEDITRAK_API_CLIENT_NAME=
-MEDITRAK_API_CLIENT_PASSWORD=
```
rohan-bes, line 8: Chose to rename these `API_CLIENT` env vars as they were a little unclear. Really these variables correspond the server's api client credentials (`admin-panel-server` in this instance).

packages/admin-panel-server/src/connections/CentralConnection.ts
```diff
@@ -25,8 +25,8 @@ const translateParams = (queryParameters?: Record<string, unknown>) => {
 /**
  * @deprecated use @tupaia/api-client
  */
-export class MeditrakConnection extends ApiConnection {
-  public baseUrl = MEDITRAK_API_URL;
+export class CentralConnection extends ApiConnection {
```
rohan-bes, line 28: These 'connection' classes are probably my most irksome feeling about this rename. I feel like `CentralConnection` really tells me nothing. Bot on the other hand, I feel that we've sort of gone down that route by design, the server itself is really just a generic place. What we could do is create explicit APIs around it in the api-client package (say `CrudApi`, `AuthApi`, etc.)

packages/psss-server/src/connections/index.ts
```diff
@@ -4,5 +4,5 @@
  */
 
 export { EntityConnection } from './EntityConnection';
-export { MeditrakConnection } from './MeditrakConnection';
+export { CentralConnection as CentralConnection } from './CentralConnection';
```
billli0, line 7: No need to rename.

packages/api-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tupaia/api-client",
-  "version": "2.1.1",
+  "version": "3.0.0",
```
billli0, line 3: Does it trigger anything to run?

billli0 (changes requested):
Thanks for the big effort on this renaming job @rohan-bes ! :tada:

Every changes are legit. Other than that I've found two functions need to be renamed as well in web-config-server `fetchFromMediTrakServer()` and `fetchFromMeditrakServerUsingTokens()`, unless we agree legacy should be legacy as untouchable. :)

Also one minor change request plus a question for myself.

packages/admin-panel-server/src/connections/CentralConnection.ts
billli0: I like this idea. That could be an easier way to manage permission for each api.

packages/psss-server/src/connections/index.ts
rohan-bes: Good catch! Will fix

packages/api-client/package.json
rohan-bes: No, we have to do a manual publish to npm. Given that nobody other than Tamanu is actually using the client though, I'm not too sure what the protocol is. I was planning on publishing a new version all the same after the code is released 👍

rohan-bes:
@billli0 ah good catch on the function names in web-config-server. While generally I'd advice against changing legacy code, I think in this instance it's okay since it's just a rename. And leaving the names as is will be even more confusing

billli0:
> @billli0 ah good catch on the function names in web-config-server. While generally I'd advice against changing legacy code, I think in this instance it's okay since it's just a rename. And leaving the names as is will be even more confusing

Yeah just realise we got have a doppelganger `meditrak-server` as an orchestration server. Definitely need to update that.

billli0 (approved):

PR #3898, opened 2022-05-03
Remove un-used @babel references in typescript packages

rohan-bes:
Noticed all these `@babel` dependencies in typescript packages even after we have transitioned away from using babel to transpile the typescript. So just cleaning them up.


packages/admin-panel-server/webpack.config.js
```diff
@@ -1,15 +0,0 @@
-module.exports = {
```
rohan-bes, line 1: No idea why we had webpack files in a server package... 🤷

billli0 (approved):
Thanks for cleaning that up!

PR #3900, opened 2022-05-03
Refactored logging to be more DRY

rohan-bes:
### Changes:

Moved winston configuration logic into server-boilerplate


packages/admin-panel-server/src/index.ts
```diff
@@ -8,9 +8,11 @@ import '@babel/polyfill';
 import * as dotenv from 'dotenv';
 import http from 'http';
 
+import winston from 'winston';
```
billli0, line 11: Do we need to import winston as we do it in boilerplate?

billli0 (approved):
Looks good!

packages/admin-panel-server/src/index.ts
rohan-bes: Yeah we still do, to use it in this file. I found it a little confusing too, but basically there's a global `winston` instance, and it can be configured from anywhere and all invocations of it after that will use the configuration.

PR #3903, opened 2022-05-05
RN-486: Introduce the meditrak-app orchestration server

rohan-bes:
### Issue RN-486:

Added the `meditrak-server` package which will act as the orchestration server for the `meditrak-app`. I hope this isn't too confusing for folks, as we're using the same name as what the now `central-server` used to be called, but it really is the appropriate title.

Pointing this PR at the `rn-487-rename-meditrak-server` branch for now as it's based off it. Will update once that's reviewed and merged to dev.


codeship-steps.yml
```diff
@@ -21,27 +21,29 @@
               command: './packages/devops/scripts/ci/testBackend.sh admin-panel-server'
             - name: Test central-server
               command: './packages/devops/scripts/ci/testBackend.sh central-server'
+            - name: Test entity-server
+              command: './packages/devops/scripts/ci/testBackend.sh entity-server'
```
rohan-bes, line 25: Just a bit of re-ordering here to make things alphabetically sorted

packages/meditrak-server/src/app/createApp.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import { TupaiaDatabase } from '@tupaia/database';
+import { MicroServiceApiBuilder } from '@tupaia/server-boilerplate';
+
+/**
+ * Set up express server with middleware,
+ */
+export function createApp() {
+  const app = new MicroServiceApiBuilder(new TupaiaDatabase()).build();
```
rohan-bes, line 12: Okay so here's the thing. While this is being dubbed an 'orchestration server' due to its role in the architecture, as of now the `meditrak-app` doesn't interact with the server in the same way all our other front-end clients do. 

They all have a simple, browser->server relationship, which requires login/logout routes, session management, etc.

The meditrak-app on the other hand handles most of that stuff locally, and just relies on the server to perform certain operations, and the mechanics of the communication is more similar to our 'microservice' style.

I think for now we'll just use a `MicroServiceApiBuilder` for this, and we'll see if we want to switch to the `OrchestrationApiBuilder` in the future. We may also want to consider renaming, or even consolidating the two ApiBuilders...

packages/meditrak-server/src/log.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+// TODO: Delete this file in favour of configureWinston() from server-boilerplate
```
rohan-bes, line 6: This file can be deleted once [this refactor PR](https://github.com/beyondessential/tupaia/pull/3900) is approved and merged into dev

billli0:
@rohan-bes Can we call it meditrak-app-server to cut down the confusion?

rohan-bes:
> Could we possibly call it meditrak-app-server to cut down the confusion?

I did consider it @billli0, and I'm not strictly opposed to it. Personally I think in a world where we never had a previous `meditrak-server`, then I'd say this package ought to be called `meditrak-server`. But since we did, it does make it a bit confusing. I suppose the tradeoff is high level of confusion now vs. low level of confusion from now onwards.

Happy to ask in slack and see what the rest of the dev team think though, I'm not totally sure which side I fall on

rohan-bes:
Closing in favor of https://github.com/beyondessential/tupaia/pull/3905

PR #3904, opened 2022-05-05
RN-333 allow account deletion in medtkapp

billli0:
### Issue #:
RN-333

### Changes:
We can initially hook up the UI to a request that comes through to the admin@tupaia.org email address. We will then manually follow up and notify the user when their account and associated data is deleted. 
If we end up fielding a lot of these requests for manual account deletion, we would then want to automate that process.

### Screenshots:
https://linear.app/bes/issue/RN-333#comment-c69cfd96


packages/meditrak-app/ios/TupaiaMediTrak/Info.plist
```diff
@@ -17,11 +17,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.11.118</string>
+	<string>1.11.120</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>118</string>
+	<string>120</string>
```
billli0, line 24: We got [a hotfix](https://beyondessential.slack.com/archives/CLVRJ3QBD/p1651706986536589) to master to bump version to 119. So this change would be 120.

packages/meditrak-app/app/requestAccountDeletion/RequestAccountDeletionPage.js
```diff
@@ -0,0 +1,203 @@
...
+              </View>
+              <Text style={{ color: THEME_COLOR_DARK }}>
+                By requesting your account to be deleted, you will still be able to login.{'\n\n'}
+                You will be contacted shortly to confirm your account deletion request.
+              </Text>
+            </ScrollView>
+          </KeyboardAvoidingView>
+          {isLoading && renderLoadingOverlay()}
+        </TupaiaBackground>
+      </MenuContext>
```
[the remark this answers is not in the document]
rohan-bes: Yeah I think that's a fine call 👍

packages/meditrak-server/src/apiV2/deleteAccount.js
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+import { respond } from '@tupaia/utils';
+import { sendEmail } from '../utilities';
+import { getUserInfoInString } from './utilities';
+
+const sendRequest = userName => {
+  const { TUPAIA_ADMIN_EMAIL_ADDRESS } = process.env;
+
+  const emailText = `${userName} has requested to delete his/her account`;
```
rohan-bes, line 12:
```suggestion
  const emailText = `${userName} has requested to delete their account`;
```
? Tiny but reads a tad clearer

packages/meditrak-server/src/apiV2/requestCountryAccess.js
```diff
@@ -84,7 +76,7 @@ export const requestCountryAccess = async (req, res) => {
   } catch (error) {
     throw new UnauthenticatedError(error.message);
   }
-  const userName = await getUserName(userId, models);
+  const userName = await getUserInfoInString(userId, models);
```
rohan-bes, line 79:
```suggestion
  const userInfo = await getUserInfoInString(userId, models);
```
And in other places this variable name is used

rohan-bes (approved):
Great stuff @billli0 super clean!

A couple of small suggestions, but nothing major 👍

billli0:
@rohan-bes I've pushed three new commits to store `isRequestedAccountDeletion` in local database and a refactor. Wanna have a quick look?

rohan-bes (approved):
Nice work man! Super simple persisting to the db. I think if we add more 'state-like' keys like this we may want to consider encapsulating them in an object somehow, but for now this is good 👍

edmofro:
Eeek sorry to wade in here after the fact, but can I suggest a (reasonably large) change here?

I get nervous of the data model in meditrak-app drifting away from a perfect match with the data model on the server, because it's set up to be a synchronising system so it should really match the source pretty closely.

But, instead of taking away the `isRequestedAccountDeletion` and storing it somewhere else, I'd like to suggest we go the other way, and _add it_ to the server. That helps us on the server end to set up for a future where account deletion requests are viewed through the admin-panel too, and gives a bit more of a robust flow because other apps logged in with that same user will correctly show the deletion request state.

What do you think @billli0?

rohan-bes:
@billli0 I chatted with @edmofro on this one and we agreed we can keep the existing implementation. Mainly because this particular table not only doesn't sync, but already has deviated from the server-side version (we store the access policy on it) 👍

PR #3905, opened 2022-05-06
RN-497: meditrak-server scaffolding

rohan-bes:
### Issue RN-497:

(duplicate of https://github.com/beyondessential/tupaia/pull/3903)

Added the `meditrak-server` package which will act as the orchestration server for the `meditrak-app`. I hope this isn't too confusing for folks, as we're using the same name as what the now `central-server` used to be called, but it really is the appropriate title.


packages/meditrak-server/src/app/createApp.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import { TupaiaDatabase } from '@tupaia/database';
+import { MicroServiceApiBuilder } from '@tupaia/server-boilerplate';
+
+/**
+ * Set up express server with middleware,
+ */
+export function createApp() {
+  const app = new MicroServiceApiBuilder(new TupaiaDatabase()).build();
```
rohan-bes, line 12: Okay so here's the thing. While this is being dubbed an 'orchestration server' due to its role in the architecture, as of now the `meditrak-app` doesn't interact with the server in the same way all our other front-end clients do. 

They all have a simple, browser->server relationship, which requires login/logout routes, session management, etc.

The meditrak-app on the other hand handles most of that stuff locally, and just relies on the server to perform certain operations, and the mechanics of the communication is more similar to our 'microservice' style.

I think for now we'll just use a `MicroServiceApiBuilder` for this, and we'll see if we want to switch to the `OrchestrationApiBuilder` in the future. We may also want to consider renaming, or even consolidating the two ApiBuilders...

billli0: Agree to use `MicroServiceApiBuilder` for now. We can always switch that to `OrchestrationApiBuilder` if new requirement comes in.

billli0 (approved):
Looks good! Love to have another TypeScript package. 👍

rohan-bes:
After discussion with @billli0 we agreed to rename this package to `meditrak-app-server`

PR #3906, opened 2022-05-09
Moved TestableApp to @tupaia/server-boilerplate

rohan-bes:
Refactored out the `TestableEntityServer` into a generic `TestableApp` class in `@tupaia/server-boilerplate`. This class is a wrapper around [supertest](https://github.com/visionmedia/supertest) which is a pretty cool library we use for testing express apps.

I'm doing this refactor so that we can add unit tests for the `meditrak-app-server`, but hopefully it paves the way for adding unit tests for our other orchestration servers as well!


packages/entity-server/jest.setup.ts
```diff
@@ -4,6 +4,11 @@
  */
 
 import { getTestDatabase, clearTestData } from '@tupaia/database';
+import { setupTestData } from './src/__tests__/testUtilities';
+
+beforeAll(async () => {
+  await setupTestData();
+});
```
rohan-bes, line 11: Just do this once at the start of all the tests rather than once per test. Initially I had assumed we might want to insert different data per test, but in practice it had proved easier and cleaner to just import the same data each time. So in the end it's faster to just import it once at the start of all tests.

packages/entity-server/src/__tests__/testUtilities/setup.ts
```diff
@@ -52,5 +55,13 @@ export const setupTestApp = async () => {
...
   );
-  return new TestableEntityServer(userAccountEmail, userAccountPassword);
+};
+
+export const setupTestApp = async () => {
+  const app = new TestableServer(createApp(getTestDatabase()));
+  app.setDefaultHeader(
+    'Authorization',
+    `Basic ${Buffer.from(`${userAccountEmail}:${userAccountPassword}`).toString('base64')}`,
+  );
```
rohan-bes, line 65: Saves us having to set the Authorization header for each request in the tests

packages/entity-server/src/__tests__/__integration__/EntityRelationshipsRoutes.test.ts
```diff
@@ -3,21 +3,19 @@
...
+import { grantAccessToCountries, revokeCountryAccess, setupTestApp } from '../testUtilities';
 import { COUNTRIES } from './fixtures';
 
 describe('relationships', () => {
   let app: TestableServer;
 
   beforeAll(async () => {
-    app = await setup();
+    app = await setupTestApp();
     grantAccessToCountries(COUNTRIES);
```
billli0, line 15: `grantAccessToCountries()` is more like `setupAccessPolicyForUserMock()` or `setupMock()`.

billli0 (approved):
Nice refactoring! Thanks @rohan-bes .

packages/entity-server/src/__tests__/__integration__/EntityRelationshipsRoutes.test.ts
rohan-bes: Hmmm I see your point, but to me that seems more like an implementation detail? I like how with the current name it just tells you what it does, not how it does it 👍

billli0: I'm happy with the current name to keep using the original one. 
I guess what I was trying to say is at my first glance I feel like `grantAccessToCountries()` should be part of `setupTestData()`, but apparently it has a different meaning. 😄

PR #3907, opened 2022-05-09
RN-493: Add AuthRoute to meditrak-app-server

rohan-bes:
### Issue RN-493:

Added `AuthRoute` to `meditrak-app-server`. First by adding an `AuthApi` to the `api-client` so that we didn't need to explicitly request from the the `central-server`. I like this move, as we've been wondering about moving the auth logic into its own server, and by wrapping it in the `api-client` like this it will make the refactor simpler.

After that we just had to add the `api-client` to the request context in the `meditrak-app-server` and then implement a simple route class that uses it.

#### Merge tasks:
- [ ] Add `AUTH_API_URL` .env vars

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tupaia/api-client",
-  "version": "3.0.0",
+  "version": "3.1.0",
   "description": "API client for connecting to Tupaia APIs",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/packages/entity-server/src/types.ts
+++ b/packages/entity-server/src/types.ts
@@ -12,9 +12,6 @@ export interface EntityServerModelRegistry extends ModelRegistry {
   readonly entityHierarchy: EntityHierarchyModel;
 }
 
-// Returns resolved type if type is promise
-export type Resolved<T> = T extends Promise<infer R> ? R : T;
-
 // Extracts keys that have object-like values from type T
 export type ObjectLikeKeys<T> = {
   [K in keyof T]: T[K] extends Record<string, unknown> ? K : never;
--- a/packages/meditrak-app-server/src/app/createApp.ts
+++ b/packages/meditrak-app-server/src/app/createApp.ts
@@ -3,13 +3,14 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 import { TupaiaDatabase } from '@tupaia/database';
-import { MicroServiceApiBuilder } from '@tupaia/server-boilerplate';
+import { handleWith, MicroServiceApiBuilder } from '@tupaia/server-boilerplate';
 import {
   ForwardingAuthHandler,
   getBaseUrlsForHost,
   LOCALHOST_BASE_URLS,
   TupaiaApiClient,
 } from '@tupaia/api-client';
+import { AuthRequest, AuthRoute } from '../routes';
 
 /**
  * Set up express server with middleware,
@@ -29,6 +30,7 @@ export function createApp() {
       res.ctx = context;
       next();
     })
+    .post<AuthRequest>('auth', handleWith(AuthRoute))
     .build();
 
   return app;
--- a/packages/meditrak-app-server/src/app/createApp.ts
+++ b/packages/meditrak-app-server/src/app/createApp.ts
@@ -4,12 +4,32 @@
  */
 import { TupaiaDatabase } from '@tupaia/database';
 import { MicroServiceApiBuilder } from '@tupaia/server-boilerplate';
+import {
+  ForwardingAuthHandler,
+  getBaseUrlsForHost,
+  LOCALHOST_BASE_URLS,
+  TupaiaApiClient,
+} from '@tupaia/api-client';
 
 /**
  * Set up express server with middleware,
  */
 export function createApp() {
-  const app = new MicroServiceApiBuilder(new TupaiaDatabase()).build();
+  const app = new MicroServiceApiBuilder(new TupaiaDatabase())
+    .middleware((req, res, next) => {
+      const baseUrls =
+        process.env.NODE_ENV === 'test' ? LOCALHOST_BASE_URLS : getBaseUrlsForHost(req.hostname);
+      const context = {
+        services: new TupaiaApiClient(
+          new ForwardingAuthHandler(req.headers.authorization),
+          baseUrls,
+        ),
+      };
+      req.ctx = context;
+      res.ctx = context;
+      next();
+    })
+    .build();
 
   return app;
 }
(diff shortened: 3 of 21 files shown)


packages/api-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tupaia/api-client",
-  "version": "3.0.0",
+  "version": "3.1.0",
```
rohan-bes, line 3: Not exactly sure how to manage this version bumping of api-client. In theory we only need to bump it when we publish a new version, which is really up to us since it's just Tamanu that is using it... I'm bumping it here, but it may be the last time I do so until all the changes for RN-486 are in...

packages/entity-server/src/types.ts
```diff
@@ -12,9 +12,6 @@ export interface EntityServerModelRegistry extends ModelRegistry {
   readonly entityHierarchy: EntityHierarchyModel;
 }
 
-// Returns resolved type if type is promise
-export type Resolved<T> = T extends Promise<infer R> ? R : T;
```
rohan-bes, line 16: Moved to `@tupaia/server-boilerplate`

packages/tsutils/src/types.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+// TODO: Switch to 'Awaited' when upgrading to typescript 4.5+
+export type Resolved<T> = T extends Promise<infer R> ? R : T;
```
rohan-bes, line 7: This is a super common type, so wanted to move it here for re-use between packages.

In typescript 4.5 they introduced an `Awaited` type that does exactly this, so when we upgrade we can remove this and replace it

packages/meditrak-app-server/src/app/createApp.ts
```diff
@@ -3,13 +3,35 @@
...
  */
 export function createApp() {
-  const app = new MicroServiceApiBuilder(new TupaiaDatabase()).build();
+  const app = new MicroServiceApiBuilder(new TupaiaDatabase())
+    .middleware((req, res, next) => {
+      const baseUrls =
+        process.env.NODE_ENV === 'test' ? LOCALHOST_BASE_URLS : getBaseUrlsForHost(req.hostname);
+      const context = {
+        services: new TupaiaApiClient(
+          new ForwardingAuthHandler(req.headers.authorization),
```
rohan-bes, line 25: Just using a forwarding auth handler for now, as the `TupaiaApi` in the `meditrak-app` manages what header to send through currently. However, in the future it might be best to make the `meditrak-app-server` responsible for passing through the appropriate header (be it the user's bearer token, or it's own api-client auth header)

packages/meditrak-app-server/src/routes/AuthRoute.ts
```diff
@@ -0,0 +1,46 @@
...
+const hasRefreshToken = (body: ReAuthenticateBody | RefreshTokenBody): body is RefreshTokenBody =>
+  'refreshToken' in body;
+
+export class AuthRoute extends Route<AuthRequest> {
+  public async buildResponse() {
+    const { body } = this.req;
+    if (hasRefreshToken(body)) {
+      const { refreshToken } = body;
+      return this.req.ctx.services.auth.refreshAccessToken(refreshToken);
+    }
```
rohan-bes, line 42: This piece of logic ignore the `grantType` parameter, and just checks if the body contains the `refreshToken` field. Honestly I wrote it this way in part because it made the typescript a bit easier... but it doesn't seem too incorrect either. But happy to change it to be a bit more robust 👍

billli0:
@rohan-bes as we manage authentication in `meditrak-app` front end code, does auth route have another purpose in `meditrak-app-server` ? Or are we going to move those request logic from front end to back end?

rohan-bes:
> as we manage authentication in meditrak-app front end code, does auth route have another purpose in meditrak-app-server ? Or are we going to move those request logic from front end to back end?

So the app actually currently does use a backend `auth` route on the first time the user logs in, and we won't be changing that functionality. This route really just ferries the request to the old `auth` route on the `central-server` 👍

billli0:
@rohan-bes if I understand correctly, maybe `login` or `auth` route is a quite common, so we can drag it to `MicroServiceApiBuilder` from `OrchestratorApiBuilder`?

rohan-bes:
> if I understand correctly, maybe login or auth route is a quite common, so we can drag it to MicroServiceApiBuilder from OrchestratorApiBuilder? * surely it is out of scope of this task or it doesn't make sense.. We can chat about it later.

Hmmm I guess it's sort of common. Honestly, this server really ought to be an `OrchestrationServer`, since they're the ones that manage logins/sessions. And if I had free-reign to refactor the meditrak-app then I'd probably make it do that. But as it stands it's sort of in-between the two... Happy to chat about this more if you like in person, it's a bit complicated to try and explain it. But I don't think other microservices will be using the `auth` route, so I think it's best to just keep it here 👍

packages/meditrak-app-server/src/routes/AuthRoute.ts
billli0: Is it easier if we bring in `yup`?

packages/meditrak-app-server/src/routes/AuthRoute.ts
```diff
@@ -0,0 +1,46 @@
...
+  devicePlatform: string;
+  installId: string;
+};
+
+export type AuthRequest = Request<
+  {
+    grantType?: 'refresh_token';
+  },
+  | Resolved<ReturnType<TupaiaApiClient['auth']['login']>>
+  | Resolved<ReturnType<TupaiaApiClient['auth']['refreshAccessToken']>>,
```
billli0, line 29: That caught me at the beginning, I thought it is another syntax of TypeScript. I feel it would be less confusing if we add a `getAuth()`.

packages/meditrak-app-server/src/routes/AuthRoute.ts
rohan-bes: Yeah probably! Nice call, I'll make the change 👍

packages/meditrak-app-server/src/routes/AuthRoute.ts
rohan-bes: Which part's confusing? The `|` symbol at the start of both lines? Yeah I find it pretty weird too... vscode just does it.

I can create a separate union type for that 👍

billli0: Ah it didn't highlight it. This part `TupaiaApiClient['auth']['refreshAccessToken']`

rohan-bes: Okay scratch that, even doing it as a separate type has the `|` at the start. I think we may just have to get used to it?

billli0 (approved):
Nice one!

PR #3911, opened 2022-05-12
[Hot fix] Entity questions not saving

edmofro:
### Issue #:
Entity questions are not saving, in the current production release of MediTrak 😱 

Need to get this out the door quickly.

### DIFF
--- a/packages/meditrak-app/app/assessment/actions/submitSurvey.js
+++ b/packages/meditrak-app/app/assessment/actions/submitSurvey.js
@@ -102,23 +102,26 @@ const processQuestions = async (dispatch, getState, database, userId, questions)
       continue;
     }
 
+    // Save any entities used in the recent list
+    if (['PrimaryEntity', 'Entity'].includes(question.type)) {
+      const { type: entityTypes } = question.config.entity;
+      addRecentEntityId(
+        database,
+        userId,
+        entityTypes,
+        getState().country.selectedCountryId,
+        answer,
+      );
+    }
+
     // Handle special question types
     switch (question.type) {
       case 'SubmissionDate':
       case 'DateOfData':
         responseFields.dataTime = moment(answer).toISOString();
         break;
-      case 'PrimaryEntity':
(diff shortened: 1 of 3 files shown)


packages/meditrak-app/app/assessment/actions/submitSurvey.js
```diff
@@ -102,23 +102,26 @@ const processQuestions = async (dispatch, getState, database, userId, questions)
...
+    if (['PrimaryEntity', 'Entity'].includes(question.type)) {
+      const { type: entityTypes } = question.config.entity;
+      addRecentEntityId(
+        database,
+        userId,
+        entityTypes,
+        getState().country.selectedCountryId,
+        answer,
+      );
+    }
```
rohan-bes, line 115: So the real difference here is that now if the question type is `PrimaryEntity` we call both `addRecentEntityId()` and `responseFields.entityId = answer` but for `Entity` we just do `addRecentEntityId()`?

And before we did both operations for both question types?

I sort of hate switch statements which fall through, but maybe this feels like the appropriate time?
```
      case 'PrimaryEntity':
        responseFields.entityId = answer;
      case 'Entity': {
        const { type: entityTypes } = question.config.entity;
        addRecentEntityId(
          database,
          userId,
          entityTypes,
          getState().country.selectedCountryId,
          answer,
        );
        break;
      }
```
?

Since this is urgent no need necessarily though... so I'm approving 👍

rohan-bes (approved):

packages/meditrak-app/app/assessment/actions/submitSurvey.js
edmofro: Ah yeah that would work, but honestly I don't like how it reads - I don't generally intuit falling through well.

Second argument is that I see this switch statement as having the responsibility of where to store the user's answer, so pulling out the conditional side effect (which is relatively unrelated to where to store the answer) seems justified.

PR #3914, opened 2022-05-15
RN-466: MediTrak entity list performance

edmofro:
### Issue #:
RN-466

### Changes:
The main change here is to stop needlessly pulling all db results into redux, and filtering them from there. Instead, the component interacts with the db directly, and ~~makes the most of one of Realm's top selling points: lazy loading queries that will pull results just-in-time as the user scrolls 🎉 ~~ then does some internal lazy loading, where it shows the first 1000 matches (prioritising start over mid string matches), and if you scroll within 2 windows of the bottom, loads the next 1000

~~Because we also show mid-string results after all the start-of-string results, the logic and behaviour gets slightly funky once we  have less than 100 start-of-string results - basically at that point we have a limited set of 100 results from both start and end of string matches, in that order. See the discussion on RN-466 for details.~~


packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -96,29 +105,29 @@ export class EntityList extends PureComponent {
...
       });
-    this.setState({
-      searchTerm,
-      searchResults,
-    });
+    } else {
+      // if there aren't enough search results for the user to look through, try including
+      // matches later in the name, or matches on the parent entity
+      const secondarySearchResults = entities.filtered(
+        `(NOT name BEGINSWITH[c] $0) AND (name CONTAINS[c] $0 OR parent.name BEGINSWITH[c] $0) LIMIT(${
```
rohan-bes, line 120:
```suggestion
        `(NOT name BEGINSWITH[c] $0) AND (name CONTAINS[c] $0 OR parent.name CONTAINS[c] $0) LIMIT(${
```
To be consistent with previous behaviour?

packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -96,29 +105,29 @@ export class EntityList extends PureComponent {
...
-        // Send entity names that start with the search term to the top.
-        const checkMatchesAtStart = name => name.toLowerCase().startsWith(lowerCaseSearchTerm);
-        if (checkMatchesAtStart(a) !== checkMatchesAtStart(b)) {
-          return checkMatchesAtStart(a) ? -1 : 1;
-        }
+    const { entities } = this.state;
 
-        // Sort alphabetically
-        return a.localeCompare(b);
+    const primarySearchResults = entities.filtered(`name BEGINSWITH[c] $0`, searchTerm);
```
rohan-bes, line 110: Are we losing the alphabetical sorting that the previous results used to have? Or are they sorted alphabetically by the db?

packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -17,27 +17,36 @@ import {
   getGreyShade,
 } from '../globalStyles';
 import { EntityItem, ITEM_HEIGHT } from './EntityItem';
+import { fetchEntities } from './helpers';
 
 const SEARCH_BOX_HEIGHT = 40;
+const ENOUGH_SEARCH_RESULTS = 20;
```
rohan-bes, line 23:
```suggestion
const ENOUGH_SEARCH_RESULTS = 100;
```
It's mentioned a couple times in the ticket that we want 100 results to be the point where we add the 'mid string ordering' results?

rohan-bes (changes requested):
Just a couple changes that feel worth making? But happy to discuss 👍

packages/meditrak-app/app/entityMenu/EntityList.js
rohan-bes: Also, what happens here if parent is `NULL`?

packages/meditrak-app/app/entityMenu/EntityList.js
edmofro: Yes sorry I was just leaving it until a chat with Flic. Had that now and she went with 500 (was using 100 as another arbitrary example in the Linear comment - apologies for the confusion)

packages/meditrak-app/app/entityMenu/EntityList.js
edmofro: Great point, we're losing it!

packages/meditrak-app/app/entityMenu/EntityList.js
edmofro: I didn't mention this decision anywhere but think the old logic gave parents way too much weight, and this makes it more like "if you really want to filter by the parent, you can bloody well type their name properly, and even then they'll only come up as secondary results!"
I should check that decision with a product owner though (which used to be me for meditrak ;-))

edmofro: Parent is null - I'll check!

packages/meditrak-app/app/entityMenu/EntityList.js
edmofro: Actually just checked and that is still retained - the sorting is done a little deeper and persists on the realm results object that this filters.

packages/meditrak-app/app/entityMenu/EntityList.js
edmofro: Tested the parent is null case and it's fine :-)

rohan-bes (approved):
Nice work! Funny how removing redux from the equation often leads to more efficient, simple solutions...

packages/meditrak-app/app/entityMenu/EntityList.js
edmofro: And followed up on the UX of filtering on parent [here](https://linear.app/bes/issue/RN-466#comment-2eb4bcea)

edmofro:
@rohan-bes unfortunately this didn't perform as well as I had hoped when put into Andrew's hands. With really fast scrolling, the entities were taking a while to show up. It seems rolling our own lazy loading (with the help of SectionList's onEndReached) actually performs better than using Realm's internal lazy loading.

See more context [here](https://linear.app/bes/issue/RN-466#comment-442fded9)

Could you please review the last three commits?

packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -91,50 +101,64 @@ export class EntityList extends PureComponent {
...
+
+    if (primarySearchResults || secondarySearchResults) {
+      const totalResultsAvailable = primarySearchResults.length + secondarySearchResults.length;
+      if (totalResultsAvailable === 0) {
+        return { noResults: true };
+      }
+      const data = [
+        ...primarySearchResults.slice(0, numberToShow),
+        ...secondarySearchResults.slice(0, numberToShow - primarySearchResults.length),
+      ];
```
rohan-bes, line 133: Hmmm won't this go a little wrong if `primarySearchResults.length > numberToShow`?
```suggestion
      const data = [
        ...primarySearchResults.slice(0, numberToShow),
        ...secondarySearchResults.slice(0, Math.max(0, numberToShow - primarySearchResults.length)),
      ];
```

packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -17,23 +17,32 @@ import {
...
 
 const SEARCH_BOX_HEIGHT = 40;
+const INITIAL_NUMBER_TO_SHOW = 1000; // will load more if they scroll to the bottom of the list
+const LOADING_MESSAGE_SECTION = { title: 'Loading more results...', data: [] }; // slightly hacky way of showing a loading message at the bottom of the list until more lazily load
+const NO_RESULTS_SECTION = { title: 'No results', data: [] };
+
 export class EntityList extends PureComponent {
   constructor(props) {
     super(props);
+    this.baseQuery = fetchEntities(
```
rohan-bes, line 30: I feel `baseQuery` is maybe a tad vague? 
```suggestion
    this.baseEntitiesQuery = fetchEntities(
```

rohan-bes (changes requested):
One change that looks like a bit of a bug. But aside from that looks great!

packages/meditrak-app/app/entityMenu/EntityList.js
edmofro: Great catch!!

rohan-bes (approved):
Nice one! Thanks!

PR #3915, opened 2022-05-16
RN-499: Add UserRewards route to meditrak-server

rohan-bes:
### Issue RN-499:

Finally sorted out the whole 'Orchestrator' vs 'Microservice' design issue with the meditrak-app-server. Thinking of sticking with the microservice builder, and using a custom `authMiddleware` to attach a user to the request if appropriate.

Basically copied over the user rewards logic from the central-server, will clean all that stuff up as part of RN-502


packages/auth/src/security.js
```diff
@@ -10,6 +10,11 @@ export function generateSecretKey() {
   return randomize('*', 20);
 }
 
+/**
+ * 'Bearer abcd123' => 'abcd123'
+ * @param {string} authHeader
+ * @returns {string} token from header
+ */
```
rohan-bes, line 17: A spoonful of JSDoc makes the Typescript go down 😄

packages/meditrak-app-server/src/app/createApp.ts
```diff
@@ -4,42 +4,30 @@
...
-      const context = {
-        services: new TupaiaApiClient(
-          new ForwardingAuthHandler(req.headers.authorization),
-          baseUrls,
-        ),
-      };
-      req.ctx = context;
-      res.ctx = context;
-      next();
-    })
```
rohan-bes, line 39: Small refactor here to make this functionality common in `ApiBuilder`

packages/server-boilerplate/src/microService/auth/middleware.ts
```diff
@@ -65,15 +61,21 @@ export const buildBasicBearerAuthMiddleware = (
...
     } else if (authHeader.startsWith('Basic')) {
-      accessPolicyObject = await getBasicAccessPolicy(authenticator, authHeader, apiName);
+      const { user, accessPolicy } = await getBasicAccessPolicy(authenticator, authHeader, apiName);
+      userObject = user;
+      accessPolicyObject = accessPolicy;
     } else {
       throw new UnauthenticatedError('Could not authenticate');
     }
 
+    req.user = userObject;
```
rohan-bes, line 78: The changes to attach the userObject here weren't actually needed in the end for this feature, as we're using a custom auth middleware. But regardless, I felt it was worthwhile to keep the changes as having access to the request's user object can be quite useful!

billli0 (approved):
Looks good to me!

PR #3918, opened 2022-05-16
RN-513: Added unit tests for meditrak-app-server

rohan-bes:
### Issue RN-513:

Added a unit testing framework and some unit tests for meditrak-app-server.

I really like using `supertest` for testing our orchestration servers (and even microservices to a degree). Given that most of the complexity with the code is about configuring the app properly (using post vs get, using params vs query, etc.) it really feels like great bang for buck, and makes it easy to write simple tests that expose the api cleanly.


packages/meditrak-app-server/src/__tests__/utilities/setupTestApp.ts
```diff
@@ -0,0 +1,31 @@
...
+import { CentralApiMock } from './CentralApiMock';
+
+jest.mock('@tupaia/api-client', () => {
+  const original = jest.requireActual('@tupaia/api-client'); // Step 2.
+  return {
+    ...original,
+    TupaiaApiClient: jest.fn().mockImplementation(() => {
+      return {
+        auth: new AuthApiMock(TEST_DATA),
+        central: new CentralApiMock(),
```
rohan-bes, line 21: Decided to go with some really simple mock classes here. I find mocking to be a pain in terms of code design honestly. Either you:
a) try and keep your mocks close to the source, which involves exposing functionality and increased coupling
b) try and keep your mocks convenient for your tests, which makes them less 'true to life'

Personally, in this case (and perhaps in general) I prefer to mock as little as possible, but to go with option b) and allow them to drift from the source. So I guess I'd anticipate each package might have their own version of mocks for each `ApiConnection` they use.

packages/meditrak-app-server/src/routes/social/SocialFeedRoute.ts
```diff
@@ -115,7 +115,7 @@ export class SocialFeedRoute extends Route<SocialFeedRequest> {
     });
 
     const hasMorePages = feedItems.length > numberPerPage;
-    const items = await Promise.all(feedItems.slice(0, numberPerPage - 1).map(f => f.getData()));
+    const items = await Promise.all(feedItems.slice(0, numberPerPage).map(f => f.getData()));
```
rohan-bes, line 118: Hooray for unit tests! They found this bug! The second parameter in `slice()` is non-inclusive, which means that the previous code was actually skipping one feed item at the end of each page 🙀

packages/meditrak-app-server/src/__tests__/utilities/setupTestApp.ts
edmofro: I'm with you here on having pretty lightweight mocking, and as a result getting more code under test (even if you lose some test specificity)

packages/meditrak-app-server/src/__tests__/utilities/CentralApiMock.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+export class CentralApiMock {
+  public async registerUserAccount(userFields: Record<string, unknown>) {
+    return { message: 'Successfully created user', id: userFields.id };
+  }
```
edmofro, line 9: I wonder if we should inject this into the mock as part of the test suite that needs it? I found it was a little hard to trace back where that expected value was coming from, and in general having more specific mocks close to the tests they're being used in makes sense to me.

packages/meditrak-app-server/src/routes/social/SocialFeedRoute.ts
```diff
@@ -115,7 +115,9 @@ export class SocialFeedRoute extends Route<SocialFeedRequest> {
     });
 
     const hasMorePages = feedItems.length > numberPerPage;
-    const items = await Promise.all(feedItems.slice(0, numberPerPage).map(f => f.getData()));
+    const items = (
+      await Promise.all(feedItems.slice(0, numberPerPage).map(f => f.getData()))
+    ).map(item => ({ ...item, creation_date: new Date(item.creation_date).toJSON() }));
```
edmofro, line 120: Was this previously broken?

edmofro (approved):
Great unit tests, thanks @rohan-bes for making things better as you go! No required changes, just a couple of minor suggestions that are open for discussion

packages/meditrak-app-server/src/__tests__/utilities/setupTestApp.ts
rohan-bes: When the mocks become as complicated and tricky to maintain the the code they're meant to stub, then I always feel like I've missed the point. Don't wanna end up in that position 👍

packages/server-boilerplate/src/models/types.ts
```diff
@@ -84,7 +84,19 @@ export type Joined<T, U extends string> = {
...
+type BaseModelOverrides<Fields = unknown, Type = unknown> = {
+  find: (filter: DbFilter<Fields>, customQueryOptions?: QueryOptions) => Promise<Type[]>;
+  findOne: (filter: DbFilter<Fields>, customQueryOptions?: QueryOptions) => Promise<Type>;
+};
+
+export type Model<BaseModel extends DatabaseModel, Fields, Type extends DatabaseType> = Omit<
+  BaseModel,
+  keyof BaseModelOverrides
+> &
+  BaseModelOverrides<Fields, Type>;
```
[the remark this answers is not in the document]
rohan-bes: I'm a bit worried when I add these super complicated types 😓 They make life a lot easier when they're working, but understanding them is really complicated. I think as we grow more mature as a typescript codebase we'll find out how best to strike the balance.

The reason for this one is to basically force using the 'overloaded' types for the model functions (`findOne`, `find`,etc.). I had to do this since the ones in base type use `any` a lot, and it was kind of messing with the types in the actual typescript packages...

packages/meditrak-app-server/src/__tests__/utilities/CentralApiMock.ts
rohan-bes: Hmmm yeah that's a good idea 💡 I had also felt the flow was a bit weird. I'll re-write things to allow injecting mocks as needed 👍

packages/server-boilerplate/src/models/types.ts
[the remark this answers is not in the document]
edmofro: Ah yep that makes sense

packages/meditrak-app-server/src/routes/social/SocialFeedRoute.ts
rohan-bes: Ummm sort of? So the `leaderboard` has always used a `toJSON()` timestamp (which looks like `2020-12-15T00:00:00.000Z`) whereas the ones pulled from the database used a different format (`2020-12-15 00:00:00.000+1100`).

The second format was making the tests hard to write as the result basically depended on the database local timezone. I decided we might as well make them consistent while fixing this. So I went with the `toJSON()` implementation?

edmofro: Oh yep sweet :+1:

PR #3920, opened 2022-05-17
Hot fix importing facilities

edmofro:
### Issue #:
See https://beyondessential.slack.com/archives/C02HES8HFQF/p1652822478304789 (first commit) and https://beyondessential.slack.com/archives/C011G5L1J12/p1652820293767959?thread_ts=1647919038.678739&cid=C011G5L1J12 (second commit)

### Changes:

- Fix missed text change from local_gov -> local_government
- Skip facility record creation if it doesn't have a valid geographical_area parent. `facility` aka `clinic` is a bit of a legacy thing, and is only maintained for people using older versions of MediTrak, meaning that most new `facility` type entities won't need a corresponding entry in the `facility`/`clinic` table

### DIFF
--- a/packages/central-server/src/apiV2/import/importEntities/updateCountryEntities.js
+++ b/packages/central-server/src/apiV2/import/importEntities/updateCountryEntities.js
@@ -28,6 +28,45 @@ function getDefaultTypeDetails(type) {
   return { categoryCode, typeName };
 }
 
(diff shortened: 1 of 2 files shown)


chris-pollard (approved):

chris-pollard:
Thanks @edmofro, I missed that yesterday!

PR #3924, opened 2022-05-18
Master to dev

IgorNadj:


IgorNadj:
@rohan-bes I check the docs every time - https://beyond-essential.slab.com/posts/release-process-hotfix-herkynw9
No idea why that's the process though.

rohan-bes:
Huh, weird... it says 'don't use use squash and merge'... but shouldn't we? If you agree I'll update the docs 👍

IgorNadj:
@rohan-bes I'm not sure who decided that, looks like it was like that since bookstack, probably for a reason?

PR #3925, opened 2022-05-19
Added logging around refresh after migrations

rohan-bes:
Once the migrations have run we do an analytics refresh. This can sometimes take a long time, and it's a bit unclear what's taking so long. Just added a simple log message to indicate that the analytics are being refreshed


chris-pollard (approved):
LGTM!

PR #3927, opened 2022-05-19
RN-512: Add ChangePasswordRoute to meditrak-app-server

rohan-bes:
### Issue RN-512:

Simple route that redirects to central-server via the CentralApi


edmofro (approved):
lgtm!

PR #3928, opened 2022-05-22
MAUI-550: Exclude Betio from COVID-19 Kiribati project

chris-pollard:
### Changes:

- Migration script to add entity relations for all sub districts except Betio against the covid_kiribati project hierarchy.


rohan-bes (approved):
LGTM!

PR #3933, opened 2022-05-26
Support string value in indicator

billli0:
### Changes:

- Example

---

### DIFF
--- a/packages/indicators/src/Builder/AnalyticArithmeticBuilder/AnalyticArithmeticBuilder.ts
+++ b/packages/indicators/src/Builder/AnalyticArithmeticBuilder/AnalyticArithmeticBuilder.ts
@@ -138,6 +138,6 @@ export class AnalyticArithmeticBuilder extends Builder {
         period,
         value: evaluateFormulaToNumber(parser, this.config.formula, dataValues),
       }))
-      .filter(({ value }) => isFinite(value));
+      .filter(({ value }) => isFinite(value) && typeof value !== 'string');
   };
 }
--- a/packages/indicators/src/Builder/AnalyticArithmeticBuilder/AnalyticArithmeticBuilder.ts
+++ b/packages/indicators/src/Builder/AnalyticArithmeticBuilder/AnalyticArithmeticBuilder.ts
@@ -14,8 +14,9 @@ import { createBuilder } from '../createBuilder';
 import {
   fetchAnalytics,
   validateConfig,
-  evaluateFormulaToNumber,
+  convertBooleanToNumber,
   replaceDataValuesWithDefaults,
+  isValidDataValues,
 } from '../helpers';
 import {
   AnalyticArithmeticConfig,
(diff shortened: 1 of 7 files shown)


packages/indicators/src/Builder/AnalyticArithmeticBuilder/AnalyticArithmeticBuilder.ts
```diff
@@ -138,6 +138,6 @@ export class AnalyticArithmeticBuilder extends Builder {
         period,
         value: evaluateFormulaToNumber(parser, this.config.formula, dataValues),
       }))
-      .filter(({ value }) => isFinite(value));
+      .filter(({ value }) => isFinite(value) && typeof value !== 'string');
```
rohan-bes, line 141: Hmmm I feel what's going on here is a bit unclear honestly. With indicators it looks like we're trying to define a slightly new spec, and it feels appropriate for us to explicit about that.

I think it'd be nice for us to have:
1. A function (eg. `isValidAnalytic(analytic)`) which has this kind of logic in it
2. Move the whole `evaluateFormulaToNumber()` stuff to this class and call it instead `convertBooleanToNumber`. Since that's all it's really doing, it's quite misleading. Nowhere else uses the `evaluateToNumber` functionality so it should be okay.

That way we can clear represent that this indicator builder returns strings, finite numbers, and booleans as a number (all of which is quite hidden atm)

rohan-bes (changes requested):
Nice work man, quick fix! 

Bit of a request to make all the logic in this area of code a bit clearer 🙏 

Also, some unit tests would be great!

packages/indicators/src/Builder/AnalyticArithmeticBuilder/AnalyticArithmeticBuilder.ts
billli0: Totally agree, not 100% sure I did the right thing, but I try to capture those validate logic together to reflex the return type.

packages/indicators/src/Builder/helpers.ts
```diff
@@ -44,7 +44,7 @@ export const fetchAnalytics = async (
   return analytics;
 };
 
-export const evaluateFormulaToNumber = (
+export const convertBooleanToNumber = (
```
rohan-bes, line 47: Ah actually, I was thinking we'd remove the `evaluateToNumber(formula)` function from expression-parser all together. And we'd implement the same logic here instead? That function being called `evaluateToNumber` is super misleading I feel, and since it's only used here, lets just remove it all together 👍

packages/indicators/src/Builder/helpers.ts
```diff
@@ -55,6 +55,9 @@ export const evaluateFormulaToNumber = (
   return value;
 };
 
+export const isValidDataValues = (value: string | number) =>
```
rohan-bes, line 58:
```suggestion
export const isValidIndicatorValue = (value: string | number) =>
```
Small rename suggestion 👍

rohan-bes (changes requested):
Thanks for adding tests @billli0! I feel we can go a little further with the refactor to clean things up from expression-parser. But happy to approve after that 👍

packages/indicators/src/Builder/helpers.ts
billli0: Nice suggestion👍 Have removed `evaluateToNumber()` in expression parser.

rohan-bes (approved):
Thanks @billli0 LGTM!

PR #3934, opened 2022-05-27
Add missing tests for meditrak-app-server

rohan-bes:
Just a couple of routes that missed having tests added amidst all the PRs. Also fixed a couple small bugs in the routes which is nice 😄

### DIFF
--- a/packages/database/src/testUtilities/buildAndInsertSurveyResponses.js
+++ b/packages/database/src/testUtilities/buildAndInsertSurveyResponses.js
@@ -67,9 +67,11 @@ const buildAndInsertSurveyResponse = async (
  */
 export const buildAndInsertSurveyResponses = async (models, surveyResponses) => {
   const user = await upsertDummyRecord(models.user);
-  return Promise.all(
-    surveyResponses.map(async surveyResponse =>
-      buildAndInsertSurveyResponse(models, user, surveyResponse),
-    ),
-  );
+  const builtResponses = [];
+  for (let i = 0; i < surveyResponses.length; i++) {
+    const surveyResponse = surveyResponses[i];
+    const builtResponse = await buildAndInsertSurveyResponse(models, user, surveyResponse);
+    builtResponses.push(builtResponse);
+  }
+  return builtResponses;
 };
--- a/packages/meditrak-app-server/src/app/createApp.ts
+++ b/packages/meditrak-app-server/src/app/createApp.ts
@@ -29,7 +29,7 @@ export function createApp(database = new TupaiaDatabase()) {
     .post<RegisterUserRequest>('user', handleWith(RegisterUserRoute))
     .get<SocialFeedRequest>('socialFeed', authMiddleware, handleWith(SocialFeedRoute))
     .get<UserRewardsRequest>('me/rewards', authMiddleware, handleWith(UserRewardsRoute))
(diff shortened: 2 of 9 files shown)


packages/database/src/testUtilities/buildAndInsertSurveyResponses.js
```diff
@@ -67,9 +67,11 @@ const buildAndInsertSurveyResponse = async (
...
-      buildAndInsertSurveyResponse(models, user, surveyResponse),
-    ),
-  );
+  const builtResponses = [];
+  for (let i = 0; i < surveyResponses.length; i++) {
+    const surveyResponse = surveyResponses[i];
+    const builtResponse = await buildAndInsertSurveyResponse(models, user, surveyResponse);
+    builtResponses.push(builtResponse);
+  }
+  return builtResponses;
```
rohan-bes, line 76: This caused what must be a rare bug on my machine: 
`error: duplicate key value violates unique constraint "entity_code_key"`

I _think_ it's due to `Promise.all` executing operations in non-deterministic order, which might've lead to a situation where two `findOrCreate()` calls were made at the same time. I guess the real underlying issue here is that `DatabaseModel.findOrCreate()` is non-atomic... but that seems like a bit too big of an issue to tackle as part of this PR...

packages/meditrak-app-server/src/app/createApp.ts
```diff
@@ -29,7 +29,7 @@ export function createApp(database = new TupaiaDatabase()) {
     .post<RegisterUserRequest>('user', handleWith(RegisterUserRoute))
     .get<SocialFeedRequest>('socialFeed', authMiddleware, handleWith(SocialFeedRoute))
     .get<UserRewardsRequest>('me/rewards', authMiddleware, handleWith(UserRewardsRoute))
-    .get<ChangePasswordRequest>(
+    .post<ChangePasswordRequest>(
```
rohan-bes, line 32: Whoops! Should've been `post` all along!

packages/meditrak-app-server/src/routes/UserRewardsRoute.ts
```diff
@@ -22,8 +22,10 @@ export class UserRewardsRoute extends Route<UserRewardsRequest> {
     const [{ coconuts, pigs }] = await models.database.executeSql<
       [{ coconuts: number; pigs: number }]
     >(
+      // The COUNT(*)::int is required here, since pg serializes count to string
+      // (https://stackoverflow.com/questions/47843370/postgres-sequelize-raw-query-to-get-count-returns-string-value)
```
rohan-bes, line 26: Funny weird bug here!

packages/database/src/testUtilities/buildAndInsertSurveyResponses.js
IgorNadj: Ah that's no good! Can you create a card for it?

IgorNadj (approved):
Nice one

packages/database/src/testUtilities/buildAndInsertSurveyResponses.js
rohan-bes: Good idea! Have created and added to refinement: https://linear.app/bes/issue/RN-536/rare-bug-where-creating-entities-fail-when-many-attempts-are-made-at

PR #3939, opened 2022-05-30
RN-476 lesmis global entity dashboard exports

billli0:
### Issue #:
RN-476

### Screenshot: 
https://linear.app/bes/issue/RN-476#comment-6e1723b4


packages/lesmis/src/views/VitalsView.js
```diff
@@ -53,9 +51,9 @@ const VitalsContainer = styled.div`
 /* eslint-disable react/prop-types */
 // Vitals data is essentially unstructured so no point checking prop types
 
-const CountryView = ({ vitals }) => (
+const CountryView = ({ vitals, variant }) => (
```
tcaiger, line 54: Is this variant option used anywhere? To me it seems un used?

packages/lesmis/src/views/DashboardView.js
```diff
@@ -43,64 +43,6 @@ TabTemplate.propTypes = {
   Body: PropTypes.string.isRequired,
 };
 
-const useDashboardDropdownOptions = () => {
```
tcaiger, line 46: There is some old `TabTemplate` component that needs to be remove.d

packages/lesmis/src/utils/export/exportImagesToPDF.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import downloadJs from 'downloadjs';
+import { Document, Page, Image, pdf } from '@react-pdf/renderer';
+
+const CostomPDFDocument = ({ pageScreenshots }) => {
```
tcaiger, line 6:
```suggestion
const CustomPDFDocument = ({ pageScreenshots }) => {
```

packages/lesmis/src/utils/export/exportImagesToPDF.js
```diff
@@ -0,0 +1,28 @@
...
+      {pageScreenshots.map(pageScreenshot => (
+        <Page key={Math.random(100)} size="A4">
+          <Image src={pageScreenshot} />
+        </Page>
+      ))}
+    </Document>
+  );
+};
+
+CostomPDFDocument.propTypes = {
```
tcaiger, line 18:
```suggestion
CustomPDFDocument.propTypes = {
```

packages/lesmis/src/utils/export/exportImagesToPDF.js
```diff
@@ -0,0 +1,28 @@
...
+  pageScreenshots: PropTypes.array.isRequired,
+};
+
+export const exportImagesToPDF = async (pageScreenshots, title) => {
+  const doc = <CostomPDFDocument pageScreenshots={pageScreenshots} />;
```
tcaiger, line 23:
```suggestion
  const doc = <CustomPDFDocument pageScreenshots={pageScreenshots} />;
```

packages/lesmis/src/utils/export/useExportToImage.js
```diff
@@ -5,38 +5,38 @@
 
 import { useRef, useState } from 'react';
 import downloadJs from 'downloadjs';
-import domtoimage from 'dom-to-image';
+
 import { sleep, toFilename } from '@tupaia/utils';
+import { getImage } from './getImages';
 
-const exportToPng = (node, filename) => {
+const exportToImage = (node, filename, formate = 'png') => {
```
tcaiger, line 12: formate should be format everywhere it is used.

packages/lesmis/src/components/LocationHeader.js
```diff
@@ -143,9 +144,14 @@ export const LocationHeader = () => {
                 )}
               />
             ) : (
-              <IconButton startIcon={<GetApp />}>
-                <I18n t="dashboards.export" />
-              </IconButton>
+              <DashboardExportModal
```
tcaiger, line 147: I think it would be good to have a loading state for the button because when it is clicked it takes a while before anything shows up. However if we are going to move the export to the server than it is ok for now without the loading state.

packages/lesmis/src/components/DashboardExportModal/pages/styles.js
```diff
@@ -0,0 +1,6 @@
+import styled from 'styled-components';
+
+export const DashboardTitleContainer = styled.div`
```
tcaiger, line 3: Good idea creating a new component for this. I would prefer to move it to the DashboardExportModal/components folder and rename the file to DashboardTitleContainer.js since that is more consistent with the conventions in lesmis.

packages/lesmis/src/components/DashboardExportModal/context/ExportOptionsContext.js
```diff
@@ -0,0 +1,39 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
```
tcaiger, line 3: Good idea to use context for the export options since the controls and the chart are so nested!

packages/lesmis/src/components/DashboardExportModal/components/OptionsBar.js
```diff
@@ -0,0 +1,55 @@
...
+          <Switch
+            checked={exportWithTable}
+            onChange={toggleExportWithTable}
+            color="primary"
+            disabled={isExporting}
+          />
+        }
+        label="Table"
+      />
+      <Pagination count={totalPage} shape="rounded" onChange={(event, value) => setPage(value)} />
```
tcaiger, line 46: Can the pagination be disabled?

tcaiger:
I got a bit confused about the structure of the files so I would like to suggest some restructuring as per below. I think the dependency cycle warning and the use of a lot of utils files are an indication that there needs to be a different structure. 
 
- Move the state for "DashboardExportModalIsOpen" to EntityView
- Move "DashboardExportModal" into DashboardView and pass down the isDashboardExportModalOpen and setIsDashboardExportModalOpen props
- Put back useDashboardDropdownOptions in DashboardView
- Pass the "dropdownOptions" into "DashboardExportModal" and do the filtering of the dashboards in that component (delete useSubDashboards.js)

packages/lesmis/src/components/DashboardExportModal/utils/useSubDashboards.js
```diff
@@ -0,0 +1,42 @@
...
+    () =>
+      subDashboards &&
+      subDashboards
+        .map(subDashboard => {
+          const { items } = subDashboard;
+          return items.length > 0 ? items.length : 1;
+        })
+        .reduce((totalNum, numOfdashboardItems) => totalNum + numOfdashboardItems, 1),
+    [JSON.stringify(subDashboards)],
+  );
```
tcaiger, line 39:
```suggestion
 const totalPage = subDashboards?.reduce(
    (totalNum, { items }) => totalNum + Math.max(1, items.length),
    1,
  );
  Perhaps this could be done in a simpler way such as this. 
```

packages/lesmis/src/components/DashboardExportModal/utils/useSubDashboards.js
```diff
@@ -0,0 +1,42 @@
+import { useMemo } from 'react';
+import { useDashboardDropdownOptions } from '../../../utils/useDashboardDropdownOptions';
+import { useUrlParams } from '../../../utils';
+import { useDashboardData } from '../../../api';
+
+export const useSubDashboards = () => {
```
tcaiger, line 6: I think there could be a better name than subDashboards because they are not part of dashboards. It could be exportableDashboards or something. Also see my main comment about moving this into the file with the dashboardModal. I think if it's in the same file then it will be more easy to understand what it is for.

packages/lesmis/src/components/DashboardExportModal/utils/useSubDashboards.js
tcaiger: Perhaps this could be done in a simpler way such as this.

tcaiger (changes requested):
Hey @billli0 First I just want to say well done. This is an extremely difficult and complex feature and it worked really well when I tested it on my local machine. It is a big PR so I have quite a few comments but my main point is about shuffling around a few of the components. Let me know if you would like to have a call to discuss it. Thanks

billli0:
Thanks @tcaiger for the review and some great suggestions. Dependency cycle warning is a tricky one to solve, thanks to your magic, now it has gone 🎉

packages/lesmis/src/components/DashboardExportModal/components/OptionsBar.js
billli0: Yes it can be. Applied thanks.

packages/lesmis/src/components/LocationHeader.js
billli0: Will put that in mind.

packages/lesmis/src/views/VitalsView.js
billli0: It has been used while I was coding, not anymore. Have been removed.

tcaiger (approved):
Thanks for all the hard work and refactoring. It looks great.

PR #3912, opened 2022-05-13
[Hot fix] RN-511: Reinstate entity answers for C19 test registration responses

edmofro:
### Issue #:
RN-511
Note that this migration does take 3 - 4 minutes to run on a standard feature branch, but I feel that's acceptable.

PR #3943, opened 2022-05-31
Rerun existing migration for new data

avaek:
This is a copy of [this migration](https://github.com/beyondessential/tupaia/pull/3912) so we can run it again to clean up new data that came in after that initial migration had been run


edmofro (approved):
Wow, this is some really great sql

PR #3948, opened 2022-06-01
RN-540 Indicator return all string instead of most recent

billli0:
### Issue #:
[RN-540](https://linear.app/bes/issue/RN-540)

### Changes:
Hacky way to return all string value. This PR can be reverted when we work on [RN-541](https://linear.app/bes/issue/RN-541)


billli0:
@rohan-bes Yep, forgot to cancel.

PR #3953, opened 2022-06-02
RN-536: Modified DatabaseModel.findOrCreate to be atomic

rohan-bes:
### Issue RN-536:

Switched to using [whereNotExists](http://knexjs.org/guide/query-builder.html#wherenotexists) ensure we don't create a record that already exists. Best practice as described here: https://stackoverflow.com/questions/4069718/postgres-insert-if-does-not-exist-already/13342031#13342031

Another option could have been use to use Postgres's `ON CONFLICT` clause (https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-upsert/) however that requires database constraints to be applied to work (something we might wanna consider for the future...)

### DIFF
--- a/packages/database/src/TupaiaDatabase.js
+++ b/packages/database/src/TupaiaDatabase.js
@@ -47,6 +47,10 @@ export const JOIN_TYPES = {
 // list valid behaviour so we can validate against sql injection
 const VALID_CAST_TYPES = ['text', 'text[]', 'date'];
 const VALID_COMPARISON_TYPES = ['where', 'whereBetween', 'whereIn', 'orWhere'];
+const WHERE_SUBQUERY_CLAUSES = {
+  EXISTS: 'exists',
+  NOT_EXISTS: 'notExists',
+};
 
 // no math here, just hand-tuned to be as low as possible while
 // keeping all the tests passing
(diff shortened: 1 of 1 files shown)


packages/database/src/TupaiaDatabase.js
```diff
@@ -553,6 +564,16 @@ function addWhereClause(connection, baseQuery, where) {
...
+    if (key === WHERE_SUBQUERY_CLAUSES.EXISTS) {
+      return querySoFar.whereExists(function () {
+        this.query(value);
+      });
+    }
+    if (key === WHERE_SUBQUERY_CLAUSES.NOT_EXISTS) {
+      return querySoFar.whereNotExists(function () {
+        this.query(value);
+      });
+    }
```
rohan-bes, line 576: I was pretty shocked to find this code worked perfectly right off the bat 🎉 I find this area of code to be pretty complex, and the lack of types really makes it hard to reason about. But on the flip side the api itself is super elegant, so doing something crazy and recursive like this just worked!

packages/database/src/TupaiaDatabase.js
```diff
@@ -572,6 +593,7 @@ function addWhereClause(connection, baseQuery, where) {
       throw new Error(`Cannot compare using ${comparisonType}`);
     }
 
+    // TODO: Replace with knex json where functions, eg. whereJsonPath
```
rohan-bes, line 596: Saw these when checking out the latest knex functions. Didn't quite wanna touch it here since this code has been a touch flaky in the past...

rohan-bes:
> Great change, thanks! Did you have a suspicion that this has caused issues, or did it come up in tests, or did you just see it and think you ought to tighten things up?

@edmofro just came up in unit tests, doing a `Promise.all()` on an array of `findOrCreate()` calls. Not aware of any cases where it's actually happened in a live environment, (not sure how often we actually use `findOrCreate()`) but it seems plausible that it may have...

edmofro:
Yeah possibly hey. Aaaah well, at least it's fixed now! 🙏

IgorNadj:
Amazing, happy it was easy!

PR #3956, opened 2022-06-03
RN-508: Upgrade to node 14.19.3

rohan-bes:
### Issue RN-508:

Bumping the node version our server instances use to the latest 14 (from 12). 

Node comes pre-installed on all of our ec2 instances in our 'Gold Master Image' that we build via ec2 image builder. So to upgrade we just need to upgrade the version that it uses. Previously we explicitly set the version in the `setupGoldMaster.sh` script, but I've changed it to use the same version as specified in `.nvmrc`.

I'll test on a branch using node 14 to ensure everything works correctly. Once that's done, we can merge this code in and then build a new gold master image using 14. We'll all dev and all feature branches using that until the next release which should give us a chance to flush out any bugs (don't think there will be any regardless)


packages/devops/scripts/deployment/setupGoldMaster.sh
```diff
@@ -51,7 +44,15 @@ mkdir -p /home/ubuntu/.local/share/lpass
 
 # clone our repo
 cd /home/ubuntu
-sudo -Hu ubuntu git clone https://github.com/beyondessential/tupaia.git
+# TODO: Remove this once RN-508 is released
```
rohan-bes, line 47: To actually get this working the `setupGoldMaster.sh` needs to be run via ec2 image builder with this branch checked out. It's a tad awkward, but the plan is to change the branch to `dev` once this code hits `dev`, and then remove entirely once it hits master.

scripts/bash/getInternalDependencies.sh
```diff
@@ -12,7 +12,7 @@ dependencies_already_visited=($@)
 
 # if no package.json entrypoint is specified, just return all internal dependencies
 if [ -z ${package_path} ]; then
-  echo "ui-components" "tsutils" "utils" "access-policy" "admin-panel" "aggregator" "api-client" "auth" "database" "data-api" "data-broker" "data-lake-api" "dhis-api" "expression-parser" "indicators" "weather-api" "server-boilerplate" "kobo-api"
+  echo "ui-components" "tsutils" "utils" "access-policy" "admin-panel" "aggregator" "api-client" "auth" "database" "data-api" "data-broker" "data-lake-api" "dhis-api" "expression-parser" "indicators" "weather-api" "server-boilerplate" "kobo-api" "report-server"
```
rohan-bes, line 15: This is actually an existing bug, that's been causing our gold master images to fail for the past couple months. Since `admin-panel-server` now depends on `report-server` for its types, we need to ensure we build it as an internal dependency.

packages/devops/scripts/deployment/setupGoldMaster.sh
```diff
@@ -51,7 +44,15 @@ mkdir -p /home/ubuntu/.local/share/lpass
...
+sudo -Hu ubuntu git clone -b rn-508-upgrade-node-to-14 https://github.com/beyondessential/tupaia.git
+
+# install node and yarn
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash
+export NVM_DIR="$HOME/.nvm"
+[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+nvm install $(sudo cat tupaia/.nvmrc)
```
rohan-bes, line 54: This is a little change to ensure that our ec2 images use the same version of node as we specify in the `.nvmrc` (ie. the one we recommend devs use). So in the future we can just bump it there 👍

packages/devops/scripts/deployment/setupGoldMaster.sh
billli0: Is this just for testing purpose?

billli0 (approved):
Looks good to me!

PR #3958, opened 2022-06-06
Bugfix regressions with data_source split

EMcQ-BES:
### Changes:

- Update viz builder search to work with split elements/groups
- Fixup createDataElements to allow updating existing data elements
- Delete extra files from bad merge

---


packages/admin-panel/src/VizBuilderApp/api/queries/useSearchDataSources.js
```diff
@@ -26,3 +25,9 @@ export const useSearchDataSources = ({ search, type = 'dataElement', maxResults
       keepPreviousData: true,
     },
   );
+  const mappedData = result?.data?.map(value => ({ ...value, type }));
```
EMcQ-BES, line 28: The actual usage of this data expects type to be included, but since we can't get that directly from the api anymore it needs to be added

We could possibly refactor the usage in `DataElementDataLibrary.js` so we don't need to do this, but for now I've just made the return from this function match the prior implementation

billli0 (approved):
Looks good to me!

PR #3959, opened 2022-06-07
Db dump show error

IgorNadj:
Show error if fails, spinner if works.

Screen Shot 2022-06-07 at 3 58 24 pm


rohan-bes (approved):
Awesome, thanks @IgorNadj!

PR #3964, opened 2022-06-13
RN-501: Move sync logic to meditrak-app-server (manage meditrak-sync-queue)

rohan-bes:
### Issue RN-501:

Part 1 of 4. 

In this PR, we're creating the `ServerChangeEnqueuer` which is a database change handler which appends all server side changes to the appropriate tables to the meditrak-sync-queue. This is the exact same behaviour as the existing `createMeditrakSyncQueue` in the `central-server`.

The first 4 commits are just moving models that exclusively existed in the `central-server` over to the `database` package so that they could be used in the `meditrak-app-server`. Basically just copy/paste jobs.

Also added a pretty rudimentary feature flag system for managing which queueing system we're using (as using both at once could cause issues).

#### Merge tasks:
- [ ] Set feature flags in `.env` vars


packages/database/src/changeHandlers/ChangeHandler.js
```diff
@@ -47,10 +47,12 @@ export class ChangeHandler {
 
   /**
    * @abstract
+   * @protected
    */
-  handleChanges = async () => {
+  // eslint-disable-next-line no-unused-vars
+  async handleChanges(change) {
```
rohan-bes, line 53: I personally quite prefer to use this notation for instance methods of classes, though I'm not entirely sure of the differences...

packages/meditrak-app-server/src/sync/ServerChangeEnqueuer.ts
```diff
@@ -0,0 +1,64 @@
...
+import { ChangeHandler } from '@tupaia/database';
+import { MeditrakSyncQueueModel } from '../models';
+import { MeditrakAppServerModelRegistry } from '../types';
+
+type Change = { record_id: string; record_type: string; type: 'update' | 'delete' };
+
+/**
+ * Adds server side changes to the meditrakSyncQueue
+ */
+export class ServerChangeEnqueuer extends ChangeHandler {
```
rohan-bes, line 15: How about that name ey? 😉

It could be better, definitely, but my motivations for going with it are:
1. All our change handlers have verb names describing their job
2. It captures how the actual system operates a bit better than `MeditrakSyncQueue` imo

I felt with the previous name it was a bit misleading, I had assumed that the `meditrakSyncQueue` was responsible both for enqueueing and de-queueing changes. But our system actually has 2 totally separate processes to manage enqueueing changes, and then passing changes to the client. So I like that the name clearly declares that this class just manages one part of the system.

packages/meditrak-app-server/src/sync/ServerChangeEnqueuer.ts
```diff
@@ -0,0 +1,64 @@
...
+    const stripDataFromChange = (
+      change: {
+        old_record: Record<string, unknown>;
+        new_record: Record<string, unknown>;
+      } & Change,
+    ) => {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      const { old_record, new_record, ...restOfRecord } = change;
+      return [{ ...restOfRecord }];
+    };
```
rohan-bes, line 31: This data needs to be stripped out, as we just keep track of the record_id for each change

packages/utils/src/featureFlags.js
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Simple feature flag functionality implemented via env vars
+ *
+ * @param {string} featureName
+ * @returns {boolean}
+ */
+export const isFeatureEnabled = featureName => {
+  return requireEnv(`FEATURE_FLAG_${featureName}`).toLowerCase() === 'on';
+};
```
rohan-bes, line 16: Decided to go with something super simple here. I actually toyed up going with a 3rd party package or something more complex, but couldn't really justify it.

Some feature that could be super useful to have:
1. Different toggle settings for prod/dev/local envs
2. Ability to switch a toggle in a live environment
3. Ability to switch a toggle for certain users/permissions

But for now just managing it via `.env` variables will be simple and fit easily with our CI/CD system 👍

packages/meditrak-app-server/src/index.ts
```diff
@@ -8,16 +8,31 @@ import * as dotenv from 'dotenv';
...
-const app = createApp();
+const app = createApp(database);
+
+/**
+ * Set up change handler for the meditrakSyncQueue
+ */
+if (isFeatureEnabled('SERVER_CHANGE_ENQUEUER')) {
+  const serverChangeEnqueuer = new ServerChangeEnqueuer(models);
+  serverChangeEnqueuer.listenForChanges();
+}
```
rohan-bes, line 35: Running the `ServerChangeEnqueuer` here rather than on the `central-server` as previous. This means that the `meditrak-app-server` would need to be running to ensure that changes get synced to the `meditrak-app`.

Also the first time we're running a change handler outside of the `central-server`. Happy to hear reviewer's opinions on it, I can definitely see pros and cons...

packages/meditrak-app-server/src/sync/ServerChangeEnqueuer.ts
edmofro: Yeah I like the change! Perhaps the `Server` bit could be changed to something like `Syncable` or `Syncing` or something? 
- Not all server side changes are enqueued
- It kinda sounds like something _about_ the server is changing, rather than a database update _on_ the server, but that could just be me and is a minor point

packages/database/src/changeHandlers/ChangeHandler.js
edmofro: The key difference is that when passed in to another function as a callback, the arrow function will bind the correct class instance as `this`, which can be super important is specific situations and usually my motivation for going with it. It must be working without though, so I guess this isn't one of those situations!

packages/utils/src/featureFlags.js
edmofro: Really really like the simplicity of the solution - thanks for not being sucked in by a flashy third party solution when really it's a straightforward and isolated thing atm!

packages/meditrak-app-server/src/index.ts
edmofro: Hmmmm....yes, I can see pros and cons too! I'll think on this...

There might be some CI/CD changes to make because of this requirement btw. For e.g. I think we wait for the meditrak-server pm2 process to post that it's ready before we run migrations.

packages/database/src/modelClasses/index.js
```diff
@@ -31,6 +31,7 @@ import { MapOverlayGroupModel } from './MapOverlayGroup';
 import { MapOverlayGroupRelationModel } from './MapOverlayGroupRelation';
 import { MapOverlayModel } from './MapOverlay';
 import { MeditrakDeviceModel } from './MeditrakDevice';
+import { MeditrakSyncQueueModel } from './MeditrakSyncQueue';
```
edmofro, line 34: I quite liked the idea that some models were entirely owned by a specific package, so didn't need to exist in the database package at all. Totally fine with this change though, and I see that it would be most devs' expectation that this would be an exhaustive list.

packages/central-server/src/index.js
```diff
@@ -37,7 +37,7 @@ const models = new ModelRegistry(database, modelClasses, true);
 /**
  * Set up change handlers e.g. for syncing
  */
-if (isFeatureEnabled('MEDITRAK_SYNC_QUEUE')) {
+if (isFeatureEnabled('MEDITRAK_SYNC_QUEUE', FEATURE_FLAG_ON)) {
```
edmofro, line 40: I'm not totally sure this is the right place to have the feature flag defaulting. With more involved feature flagging, we might have conditional checks in multiple places, and it'd be better to have one place that determined whether that feature flag was defaulted to on or off. That might look like a central repository somewhere else that defines the defaults - which in an ideal world would look like moving to https://www.npmjs.com/package/config instead of .env files, like Tamanu does :-)

edmofro (approved):
lgtm! Two non-blocking comments

packages/database/src/changeHandlers/ChangeHandler.js
rohan-bes: Ah damnit! The syntax I find uglier is the better performing one 😭 

Hmmm I have run into that issue once or twice before... Okay, I'll soften my stance on this. Thanks for mentioning it!

packages/meditrak-app-server/src/index.ts
rohan-bes: Ah good point... I guess we could just launch the `meditrak-app-server` first? But then we're starting to wade into an uncomfortable territory where various services rely on others running prior to them... Perhaps central-server could check to see if meditrak-app-server is running before it kicks off the migrations?

Lets chat on it. I don't know if I feel too strongly either way, but keeping it here certainly keeps the code cleaner 😄

packages/database/src/modelClasses/index.js
rohan-bes: I know what you mean, and I see the definitely see the value in that. Currently with all the typescript packages we limit which models they're aware of by having a custom ModelRegistry definition for each package, which seems to handle information hiding really well. But if there's a model with a whole lot of business logic, it does indeed feel a bit awkward if that logic is in say `@tupaia/database` if it really just relates to another package...

That being said the gradual pull of moving all the model classes (or atleast having a base model class) for each database table in `@tupaia/database` does seem to be happening... Ultimately I feel we may end up relying on something like the much discussed [@tupaia/types](https://linear.app/bes/issue/RN-509/add-basic-tupaiatypes-package) to manage this...

packages/central-server/src/index.js
rohan-bes: Yeah good point! I thought about maybe adding a fairly lightweight `@tupaia/feature-toggles` package, which I do still feel might be the way to go...

Alternatively, how about an even simpler option in `@tupaia/utils/src/featureToggles.js`:
```
const toggles = {
  MEDITRAK_SYNC_QUEUE: {
    default: FEATURE_FLAG_ON
  },
  SERVER_CHANGE_ENQUEUER: {
    default: FEATURE_FLAG_OFF
  },
}

const isFeatureEnabled = (featureName) => {
  if (!(featureName in toggles)) {
    throw new Error (`Unkown feature: ${featureName});
  }
  
  let toggleState;
  if (toggles[featureName].default) {
    toggleState = getEnvOrDefault(featureName, toggles[featureName].default);
  } else {
    toggleState = requireEnv(featureName);
  }

  return toggleState.toLowerCase() === FEATURE_FLAG_ON;
}
```

edmofro: Yup I like the simpler option, at least while there are only two and the behaviour is really limited

edmofro: A note to get it off my mind: I was just wondering whether we should keep the toggle definitions in the package they apply to, but it feels like over-engineering to have three separate files just to manage this small amount of toggling. It could be something to keep in mind if this grows though, but there could well be toggles that apply across packages too.

rohan-bes: @edmofro I thought about that too with keeping the toggles in their relevant packages, but I couldn't shake the feeling that it could be fairly common for a toggle to be spread across packages (this one sort of is in a way) and so didn't want to lock it in to that kind of pattern.

packages/meditrak-app-server/src/sync/ServerChangeEnqueuer.ts
rohan-bes: I like `SyncableChangeEnqueuer` 👍 definitely better than `Server` I agree!

PR #3965, opened 2022-06-13
RN-501: Move sync logic to meditrak-app-server (CountChangesRoute)

rohan-bes:
### Issue RN-501:

Part 2 of many.

In this PR, we are creating the `CountChangesRoute`, which the meditrak-app uses to determine how many changes it needs to sync. After discussing with @edmofro and Erin, we've decided to deprecate the [legacy behaviour](https://github.com/beyondessential/tupaia/blob/973ac52ae0381ad668e7f417f664ee7fb63b2927/packages/central-server/src/apiV2/countChanges/LegacyCountChangesHandler.js) which is used by app verions prior to `1.7.106` where they do not provide an `appVersion` url parameter in the request. Part of this PR will be ensuring that any users that are still using a legacy version have been notified and encouraged to upgrade.


packages/central-server/src/tests/testUtilities/index.js
```diff
@@ -6,7 +6,6 @@
 export * from './assertions';
 export * from './database';
 export { expectPermissionError, expectResponseError } from './expectResponseError';
-export { randomEmail, randomIntBetween, randomString } from './random';
```
rohan-bes, line 9: Moved to `@tupaia/utils` for re-use among other packages

packages/meditrak-app-server/src/__tests__/__integration__/sync/CountChangesRoute.test.ts
```diff
@@ -0,0 +1,299 @@
...
+  const dataElement = await upsertDummyRecord(models.dataElement, {
+    service_type: 'tupaia',
+  });
+  return upsertDummyRecord(models.question, {
+    code: dataElement.code,
+    data_element_id: dataElement.id,
+  });
+};
+
+describe('changes/count', () => {
```
rohan-bes, line 26: These tests are fantastic!! Thanks to @edmofro who wrote them in [central-server](https://github.com/beyondessential/tupaia/blob/00eb28394161dee852f873a3a452e9283d0f2877/packages/central-server/src/tests/apiV2/getChanges.test.js) I was basically able to copy them over and they worked immediately. Helped me pick out some bugs and gave me a lot more confidence in the code 👍

packages/meditrak-app-server/src/__tests__/__integration__/sync/CountChangesRoute.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    const timestampBeforeSecondUpdate = Date.now();
+    await oneSecondSleep();
+    const numberOfQuestionsToAddInSecondUpdate = randomIntBetween(1, 20);
+    const newQuestionsInSecondUpdate = [];
+    for (let i = 0; i < numberOfQuestionsToAddInSecondUpdate; i++) {
+      newQuestionsInSecondUpdate[i] = await upsertQuestion(models);
+    }
+
+    // Delete some of the questions added in the first update
+    await oneSecondSleep();
```
rohan-bes, line 128: These second long waits do add up a tiny bit... tests can take 30-60 seconds on my local machine. Personally I feel it's fine, but in an ideal world it might be worth finding a way to avoid them

packages/meditrak-app-server/src/app/createApp.ts
```diff
@@ -25,6 +28,7 @@ export function createApp(database = new TupaiaDatabase()) {
   const authMiddleware = buildAuthMiddleware(database);
   const app = new MicroServiceApiBuilder(database)
     .attachApiClientToContext(authHandlerProvider)
+    .use('*', checkAppVersion)
```
rohan-bes, line 31: This enforces that the `appVersion` parameter is always passed in for all routes

packages/meditrak-app-server/src/routes/sync/getChangesFilter.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { getHighestPossibleIdForGivenTime } from '@tupaia/database';
+import { DbFilter } from '@tupaia/server-boilerplate';
+import { getSupportedModels } from '../../sync';
+import { MeditrakSyncQueueFields } from '../../models/MeditrakSyncQueue';
```
rohan-bes, line 9: All this is copied from https://github.com/beyondessential/tupaia/blob/973ac52ae0381ad668e7f417f664ee7fb63b2927/packages/central-server/src/apiV2/utilities/getChangesFilter.js

packages/meditrak-app-server/src/__tests__/__integration__/sync/CountChangesRoute.test.ts
edmofro: Great! Love an endpoint level test, so robust to internal refactors 😀

packages/meditrak-app-server/src/__tests__/__integration__/sync/CountChangesRoute.test.ts
edmofro: Yeah...definitely an anti pattern I'm not proud of. We also run the risk of it being flakey based on timing differences - the long wait time mitigates that risk, but obviously comes with its own downside!

packages/meditrak-app-server/src/middleware/checkAppVersion.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { NextFunction, Request, Response } from 'express';
+
+export const checkAppVersion = async (req: Request, res: Response, next: NextFunction) => {
+  try {
+    const { appVersion } = req.query;
+    if (!appVersion) {
+      throw new Error('appVersion unspecified, please upgrade your app');
```
edmofro, line 12: Nice! And this just logs the user out and shows them the error message on their login screen hey? Do you know if that is supported in the old versions that the users getting this will be using? 😅

packages/server-boilerplate/src/__tests__/utilities/TestableServer.ts
```diff
@@ -47,6 +49,10 @@ export class TestableServer {
     this.defaultHeaders[name] = value;
   }
 
+  public setDefaultQueryParam(name: string, value: string) {
```
edmofro, line 52: Great addition

edmofro (approved):
Great work! Just want to check the UX of the error message on older installs, but I actually think that's been supported for ages, so should be good to go :+1:

packages/meditrak-app-server/src/__tests__/__integration__/sync/CountChangesRoute.test.ts
rohan-bes: Yeah, there probably is a way we could work around this if needs be. But the tests still run fast enough imo 👍

packages/meditrak-app-server/src/middleware/checkAppVersion.ts
rohan-bes: I'll test this before I merge! 👍

PR #3974, opened 2022-06-15
Build admin-panel-server last

edmofro:
This should fix an issue where the Gold Master EC2 Image building process is failing with the following error: 
```
src/viz-builder/types.ts(6,30): error TS2307: Cannot find module '@tupaia/report-server' or its corresponding type declarations.
```

i.e., the report server has not had its types built at the point the admin panel is being built

There's one thing that's confusing me though - the image building process _does_ sometimes succeed (see the screenshot below). Can the reviewer understand why that might be? I would've thought if this was the issue it would deterministically fail?

image

### DIFF
--- a/scripts/bash/getDeployablePackages.sh
+++ b/scripts/bash/getDeployablePackages.sh
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 
-echo "admin-panel-server" "central-server" "entity-server" "lesmis-server" "meditrak-app-server" "psss-server" "report-server" "web-config-server" "admin-panel" "lesmis" "psss" "web-frontend"
+echo "admin-panel" "lesmis" "psss" "web-frontend" "central-server" "entity-server" "lesmis-server" "meditrak-app-server" "psss-server" "report-server" "web-config-server" "admin-panel-server" # admin-panel-server last as it depends on report-server
 exit 0
--- a/scripts/bash/getDeployablePackages.sh
+++ b/scripts/bash/getDeployablePackages.sh
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 
-echo "admin-panel" "lesmis" "psss" "web-frontend" "admin-panel-server" "central-server" "entity-server" "lesmis-server" "meditrak-app-server" "psss-server" "report-server" "web-config-server"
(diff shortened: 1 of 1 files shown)


rohan-bes:
> There's one thing that's confusing me though - the image building process does sometimes succeed (see the screenshot below). Can the reviewer understand why that might be? I would've thought if this was the issue it would deterministically fail?

Ah that would be me! So I actually noticed this issue a little while ago when I was working on https://github.com/beyondessential/tupaia/pull/3956 . I fixed it then, and created a couple of images using Node14 to test it (I untagged them from the standard gold master tag so they weren't being used by regular deployments).  So that's what those 2 green builds are about 👍

scripts/bash/getDeployablePackages.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash -e
 
-echo "admin-panel" "lesmis" "psss" "web-frontend" "admin-panel-server" "central-server" "entity-server" "lesmis-server" "meditrak-app-server" "psss-server" "report-server" "web-config-server"
+echo "admin-panel-server" "central-server" "entity-server" "lesmis-server" "meditrak-app-server" "psss-server" "report-server" "web-config-server" "admin-panel" "lesmis" "psss" "web-frontend"
```
rohan-bes, line 3: Just wanted to double check this works? I thought it was the `admin-panel-server` that depended upon `report-server`? 

```suggestion
echo "report-server" "admin-panel-server" "central-server" "entity-server" "lesmis-server" "meditrak-app-server" "psss-server" "web-config-server" "admin-panel" "lesmis" "psss" "web-frontend"
```

Alternatively we could flag the `report-server` as an internal dependency, as I did in https://github.com/beyondessential/tupaia/pull/3956 but that requires a few more code changes and is confusing in its own right...

rohan-bes (commented):
Just wanted to check that the unmet dependency has been fixed? If so happy with this 👍

scripts/bash/getDeployablePackages.sh
edmofro: Oh you're right!

rohan-bes (approved):
Nice, comment helps make it clear why it's not in alphabetical order 👍

PR #3985, opened 2022-06-20
Fetch permission group by searching input in viz builder

billli0:
### Issue #:
RN-563
There is a bug that we cannot find permission group `MDA User` on viz builder while editing or creating. The reason is that as it fetches all options as a list first, the max records from GET handler is 100, https://github.com/beyondessential/tupaia/blob/b8583bdc7556b277f0438b19a2ba577a92b8e411/packages/meditrak-server/src/routes/GETHandler.js#L12, and we currently reach 108 permission group😅, which causes this issue.

I am a bit inclined to change it to search permission group by input, instead of fetching all permission groups as it keeps growing, because it requires less changes.


rohan-bes (approved):
Great call to switch to search instead of just increasing the max returned size! Nice change, thanks @billli0

PR #3992, opened 2022-06-23
RN-400: Meditrak permissions based sync (server-side changes)

rohan-bes:
### Issue RN-400:

Server side changes for the new meditrak permissions based sync.

Sync logic:
- The meditrak-app sends across a list of all previously synced countries and permission groups when syncing (none if first time sync)
- The app then just combines these lists with the countries and permissions in the user's access policy
  - For countries and permission groups that have previously been synced, just sync new changes since last sync
  - For countries and permission groups that have never been synced, sync data for all time for those
  - Note: all models that we sync can be filtered by one or both of these permissions restrictions, see comments in PR for details
- Once the sync is complete, the app then updates the list of synced countries and permission groups

This algorithm is hopefully sufficiently simple and flexible (can handle new user logins, as well as user permission changes) without syncing too much data, especially during the initial sync.

In order to filter the changes to sync by the countries and permission groups, I've created a new materialized view (`permissions_based_meditrak_sync_queue`) which the `MeditrakSyncQueue` manages.

I've tried to break this PR up into clean separate commits, if that's easier for the reviewer.


packages/central-server/src/apiV2/changesMetadata.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Permissions based sync metadata
+ * {
+ *   changeCount: number of changes since last sync
+ *   countries: countries included in the sync
+ *   permissionGroups: permissions groups included in the sync
+ * }
+ * Responds to GET requests to the /changes/metadata endpoint
+ */
+export async function changesMetadata(req, res) {
```
rohan-bes, line 19: Given that this endpoint returns more/different data to the `/changes/count` endpoint, I decided to create a new one with a new name. Also makes maintaining support for the old logic a little easier.

packages/central-server/src/apiV2/countChanges/LegacyCountChangesHandler.js
```diff
@@ -135,11 +135,16 @@ export class LegacyCountChangesHandler {
     const { models, query } = this.req;
 
     const universalTypes = getUniversalTypes(models);
-    const filter = await getChangesFilter({
-      ...this.req,
-      query: { ...query, recordTypes: universalTypes.join(',') },
-    });
-    const changesCount = await models.meditrakSyncQueue.count(filter);
+
+    const { query: dbQuery } = await getChangesFilter(
```
rohan-bes, line 139: Re-wrote this to just build up an actual SQL query rather than a query object. I find once the queries get beyond a certain level of complexity then it's just much easier to work with direct SQL. Still feels a bit odd passing in the select statement here though so I understand if reviewer wants to push back.

packages/central-server/src/apiV2/getChanges.js
```diff
@@ -34,45 +41,68 @@ async function getRecordForSync(record) {
...
-        if (action === 'delete') {
-          changeObject.record = { id: recordId };
-          if (recordType === TYPES.GEOGRAPHICAL_AREA) {
-            // TODO LEGACY Deal with this bug on app end for v3 api
-            changeObject.recordType = 'area';
-          }
+    const changes = await query.executeOnDatabase(database);
+    const changesByRecordType = groupBy(changes, 'record_type');
+    const recordTypesToSync = Object.keys(changesByRecordType);
+    const columnNamesByRecordType = Object.fromEntries(
```
rohan-bes, line 62: Rewrote this to do all the database stuff upfront in batch. Made server-side processing of changes much faster. Now the bottleneck is on the app side write to database.

packages/central-server/src/apiV2/utilities/permissionsBasedChangeFilter.js
```diff
@@ -0,0 +1,170 @@
...
+
+import { SqlQuery } from '@tupaia/database';
+import semverCompare from 'semver-compare';
+
+export const PERMISSIONS_BASED_SYNC_MIN_APP_VERSION = '1.12.123';
+
+export const supportsPermissionsBasedSync = version =>
+  semverCompare(version, PERMISSIONS_BASED_SYNC_MIN_APP_VERSION) >= 0;
+
+export const getCountriesAndPermissionsToSync = async req => {
```
rohan-bes, line 14: Here's where we do the little shimmy to work out which new countries and permission groups we need to sync. Will add more comments around this code explaining why

packages/central-server/src/database/createPermissionsBasedMeditrakSyncQueue.js
```diff
@@ -0,0 +1,78 @@
...
+
+import { SqlQuery } from '@tupaia/database';
+
+/**
+ * Creates the permissions_based_meditrak_sync_queue Materialized View
+ *
+ * This is used to improve the speed of querying the meditrak_sync_queue when
+ * doing a permissions based sync
+ */
+const query = new SqlQuery(`
```
rohan-bes, line 14: Big ol' SQL query yay! Lemme know if you'd prefer me to abstract this into something more high level? I don't mind it as it stands cos it's more readable than it would be otherwise.

packages/central-server/src/index.js
```diff
@@ -28,70 +28,72 @@ import { createApp } from './createApp';
...
+   */
+  const database = new TupaiaDatabase();
+  const models = new ModelRegistry(database, modelClasses, true);
+
+  /**
+   * Set up change handlers e.g. for syncing
+   */
+  if (isFeatureEnabled('MEDITRAK_SYNC_QUEUE')) {
+    const meditrakSyncQueue = new MeditrakSyncQueue(models);
+    await meditrakSyncQueue.createPermissionsBasedView(); // Ensure permissions based view has been setup
```
rohan-bes, line 43: Always re-create on server startup? Simple: yes, efficient: maybe no... especially since this adds about 30 seconds to the server start time...

packages/database/src/SqlQuery.js
```diff
@@ -11,9 +11,18 @@ export class SqlQuery {
...
-  static array = arr => `(${arr.map(() => '?').join(',')})`;
+  static array = (arr, type) =>
+    `ARRAY[${arr.map(() => '?').join(',')}]${type ? `::${type}[]` : ''}`;
+
+  /**
+   * @public
+   * @param {unknown[]} arr
+   * @returns {string} SQL parameter injection string for the array of values
+   */
+  static record = arr => `(${arr.map(() => '?').join(',')})`;
```
rohan-bes, line 25: Bit of a rename here, this whole time we've actually been creating 'records' rather than arrays! https://www.postgresqltutorial.com/postgresql-plpgsql/plpgsql-record-types/

packages/database/src/migrations/20220622064208-AddArrayConcatAggregationFunction-modifies-schema.js
```diff
@@ -0,0 +1,35 @@
...
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`
+  DROP AGGREGATE IF EXISTS array_concat_agg(anyarray);
+  CREATE AGGREGATE array_concat_agg(anyarray) (
```
rohan-bes, line 20: Custom aggregate function to help with the permissions based meditrak sync queue

packages/central-server/src/apiV2/utilities/permissionsBasedChangeFilter.js
```diff
@@ -0,0 +1,170 @@
...
+};
+
+export const getPermissionsWhereClause = permissionsBasedFilter => {
+  let query = '';
+  const params = [];
+  const { countries, countryIds, permissionGroups } = permissionsBasedFilter;
+  const typesWithoutPermissions = ['country', 'permission_group']; // Sync all countries and permission groups (needed for requesting access)
+  const permissionsClauses = [
+    {
+      query: `"type" = ? AND record_type IN ${SqlQuery.record(typesWithoutPermissions)}`,
```
edmofro, line 92: I'd prefer to separate these into two separate clauses, as it confused me initially thinking that we were only excluding deletes for the record types without permissions. I understand that is hard with the `OR`s between everything, so maybe we set up an initial query string outside of this first?

packages/central-server/src/database/MeditrakSyncQueue.js
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  async refreshPermissionsBasedView() {
+    try {
+      const start = Date.now();
+      await this.models.database.executeSql(
+        `REFRESH MATERIALIZED VIEW CONCURRENTLY permissions_based_meditrak_sync_queue;`,
+      );
+      const end = Date.now();
+      winston.info(`permissions_based_meditrak_sync_queue refresh took: ${end - start}ms`);
```
edmofro, line 63: How long does this generally take, out of interest?

packages/central-server/src/database/MeditrakSyncQueue.js
```diff
@@ -0,0 +1,158 @@
...
+    }));
+
+    const allChanges = [
+      surveyChange,
+      ...surveyScreenChanges,
+      ...surveyScreenComponentChanges,
+      ...questionChanges,
+      ...optionSetChanges,
+      ...optionChanges,
+    ];
```
edmofro, line 129: It would be good to run some stress testing on this - does it perform ok if e.g. all of the ssc's are being changed anyway (because they're deleted and created again) so they're also being added to the sync queue through the normal process? We don't hit too much locking if you do that with a bunch of surveys in a row?

packages/central-server/src/index.js
edmofro: But good call to go simple to start with, and optimise later if we need to

edmofro (changes requested):
This is some great work, thanks heaps for this @rohan-bes! Our users are going to absolutely love it, and it fixes a real point of embarrassment during initial sync (i.e. peoples first impression of meditrak!)

I think you've gone for a good compromise in the design decision to not bother looking the country x permission group combination, but instead just going for the simpler but mostly-just-as-good check for if the user has the permission group in any country, and any permission group in the country.

edmofro:
Oh and one more general comment: I think this could do with more inline descriptions throughout. You've articulated the strategy really clearly in the PR description, and at a minimum I'd love to see that just copied into the top of one of the files.

packages/central-server/src/apiV2/getChanges.js
```diff
@@ -34,45 +41,68 @@ async function getRecordForSync(record) {
...
+  const { limit = MAX_CHANGES_RETURNED, offset = 0, appVersion } = req.query;
 
   await req.assertPermissions(allowNoPermissions);
 
   try {
-    const filter = await getChangesFilter(req);
-    const changes = await database.find(TYPES.MEDITRAK_SYNC_QUEUE, filter, {
-      sort: ['change_time'],
+    const msqColumns = await models.meditrakSyncQueue.fetchFieldNames();
+    const changeFilterFunction = supportsPermissionsBasedSync(appVersion)
```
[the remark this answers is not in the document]
rohan-bes: I might do a bit of renaming around this area. I now recognise that the nature of this function has actually changed (it used to just build up a filter object, but now it builds a query to fetch the changes).

But I'll take note to ensure whatever happens this line reads more intuitively 👍

packages/central-server/src/apiV2/utilities/getChangesFilter.js
```diff
@@ -24,39 +28,70 @@ const getRecordTypeFilter = async req => {
...
+    OR ("type" = ? AND record_id <= ?))
+  `);
+  params.push('delete', highestPossibleSyncedId);
+
+  const recordTypeFilter = await getRecordTypeFilter(req);
+
+  if (recordTypeFilter) {
+    const { comparator, comparisonValue } = recordTypeFilter;
+    query = query.concat(`
+      AND record_type ${comparator} ${SqlQuery.record(comparisonValue)}
```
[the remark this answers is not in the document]
rohan-bes: Ah that's a really good point! Tbh, I don't know enough about param substitution, but I had assumed it wouldn't work for things like operators... I'll try it out and see 👍

packages/central-server/src/apiV2/utilities/getChangesFilter.js
```diff
@@ -68,13 +103,105 @@ const extractSinceValue = req => {
...
+    `);
+    params.push(limit);
+  }
+
+  if (offset !== undefined) {
+    query = query.concat(`
+    OFFSET ?
+    `);
+    params.push(offset);
+  }
```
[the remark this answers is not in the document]
rohan-bes: Good point, will do 👍

packages/central-server/src/apiV2/utilities/getChangesFilter.js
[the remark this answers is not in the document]
edmofro: Mmmm actually you might be right there! In which case perhaps we should validate against a list of supported comparators, exported from the db? Maybe overkill.

packages/central-server/src/database/MeditrakSyncQueue.js
rohan-bes: Good call! I had sort of assumed that this change handler would run after all the associated changes had resolved due to the debounce time, but it's definitely possible that it could kick off halfway which might make things awkward. I'll make a note in the test plan 👍

packages/central-server/src/database/models/Survey.js
```diff
@@ -16,6 +16,24 @@ class SurveyType extends DatabaseType {
...
+    return this.otherModels.surveyScreen.find({ survey_id: this.id });
+  }
+
+  async surveyScreenComponents() {
+    const questions = await this.database.executeSql(
+      `
+       SELECT ssc.* FROM survey_screen_component ssc
+       JOIN survey_screen ss ON ss.id = ssc.screen_id
+       JOIN survey s ON s.id = ss.survey_id
+       WHERE s.code = ?
```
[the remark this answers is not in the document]
rohan-bes: Ah good catch! I'll fix this up

packages/central-server/src/apiV2/utilities/permissionsBasedChangeFilter.js
rohan-bes: I think I can come up with something to address this 👍

packages/central-server/src/database/MeditrakSyncQueue.js
rohan-bes: I'll double check this, but I think I remember it taking a few seconds. Since it's a standard materialized view it basically has to fully rebuild each time, but the underlying query doesn't take too long.

If it were a fast refresh materialized view however... 😉

edmofro: Hahaha nooo don't do it!!

packages/central-server/src/database/createPermissionsBasedMeditrakSyncQueue.js
```diff
@@ -0,0 +1,89 @@
...
+DROP MATERIALIZED VIEW IF EXISTS permissions_based_meditrak_sync_queue;
+CREATE MATERIALIZED VIEW permissions_based_meditrak_sync_queue AS 
+SELECT msq.*,
```
rohan-bes, line 23: @edmofro bit of a change here in the latest commit. Before we had a separate field for each record type permissions, eg. (survey_countries, survey_permission_groups, survey_screen_countries, etc.)

This was pretty ugly and tedious, so I'm merging them into a single value using [COALESCE](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-coalesce/) here. Since there's just a single value per record type I think it should all be fine, and makes things a lot more readable!

packages/central-server/src/database/MeditrakSyncQueue.js
rohan-bes: Okay, so it takes 30 seconds to refresh, my bad. Makes sense as that's how long it takes to build the table initially. However, it's a[ concurrent refresh](https://www.postgresql.org/docs/current/sql-refreshmaterializedview.html), which means the table can be safely queried while it's refreshing

packages/database/src/migrations/20220622064208-AddArrayConcatAggregationFunction-modifies-schema.js
```diff
@@ -0,0 +1,34 @@
...
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`
+  CREATE OR REPLACE AGGREGATE array_concat_agg(anyarray) (
+    SFUNC = array_cat,
+    STYPE = anyarray
+  );
+  `);
```
rohan-bes, line 23: @edmofro there's actually a little start up problem when trying to deploy this migration.

In order to run this migration, the `MeditrakSyncQueue` must be running. 
In order to create the `MeditrakSyncQueue` the database must have the `array_concat_agg` function.

I guess this highlights a genuine dependency cycle in our code logic, however I feel it'll be acceptable for us to just manually install this function in the prod database prior to releasing this code?

edmofro: Yep I'm fine with that

edmofro (approved):
All looking great!

PR #3994, opened 2022-06-24
RN-569 Upgrade yarn to v3

IgorNadj:
### Changes:
- Upgrade yarn to v3
- Declare devDependencies of each package explicitly (see [comment below](https://github.com/beyondessential/tupaia/pull/3994#discussion_r905618090))
- Create common scripts for building, testing etc. that can be reused (see [comment below](https://github.com/beyondessential/tupaia/pull/3994#discussion_r905617224))
- Use babel instead of rollup
  - admin-panel and ui-components
  - This is simpler: can use common script, and gives the same result, see [slack thread](https://beyondessential.slack.com/archives/C032DCHKTGR/p1656288523128319?thread_ts=1656030432.164739&cid=C032DCHKTGR)
- Move e2e tests to their own package
  - There is no dependency on web-frontend (after a tiny decoupling), and this way each package can have their dependencies easier to understand and maintain
- Fix code
  - ui-components: uniquely name exported members
  - lesmis, web-frontend: use normal import from package, not `@tupaia/package/some/file`
  - lesmis: move tests to standard dir so normal test script can be used

### Outstanding issues
- yarn run is slow due to issue with yarn https://github.com/yarnpkg/berry/issues/2575
  - This will be worked around by introducing parallel builds in RN-346. Building in parallel will make the fact that there is an overhead for each build insignificant.
  - The other thing we do is use `npm run x` instead of `yarn x` for builds, this saves around 4 seconds per invocation, and because we use chaining quite a bit this saves a lot of time.

## Performance tests
| Thing |                                                                  BEFORE |     AFTER | CHANGE% | What's changed |
| -                                                                   | -         | -         | - | - |
| SKIP_BUILD_INTERNAL_DEPENDENCIES=1 yarn install (node_modules exists)  | 1.5s      | 3s         | ◽️about the same | Yarn v1 -> Yarn v3 |
| SKIP_BUILD_INTERNAL_DEPENDENCIES=1 yarn install (no node_modules)      | 104s       | 57s        | 🟩 45% faster | Yarn v1 -> Yarn v3 |
| ui-components build                                                    | 16s        | 5s         | 🟩 68% faster | rollup -> babel, remove code splitting |
| admin-panel build                                                      | 34s        | 34s         | ◽️ about the same | rollup -> babel |
| yarn build:internal-dependencies                                       | 325s       | 147s        | 🟩 54% faster | build-dev instead of build |

## Post-Release tasks
- [ ] Clean up web-frontend.env in lastpass, remove cypress vars


package.json
```diff
@@ -29,85 +29,46 @@
     "migrate-create": "yarn workspace @tupaia/database migrate-create",
     "migrate-down": "yarn workspace @tupaia/database migrate-down",
     "package-json-lint": "npmPkgJsonLint --configFile .npmpackagejsonlintrc.json .",
-    "postinstall": "test -n \"$SKIP_BUILD_INTERNAL_DEPENDENCIES\" || yarn build:internal-dependencies",
     "refresh-database": "yarn workspace @tupaia/database refresh-database --root ../../",
     "test-all": "node ./scripts/node/testAll --silent",
-    "validate": "./scripts/bash/validate.sh"
+    "validate": "./scripts/bash/validate.sh",
+    "package:build:ts": "cd $INIT_CWD && tsc -p tsconfig-build.json",
```
IgorNadj, line 35: Scripts with colon can be called from any workspace. Combined with $INIT_CWD which is the workspace calling the script, we can reuse scripts in our workspaces. The other handy thing about this is that yarn loads root level npm packages, so each package doesn't need to have a copy of e.g. eslint declared as a dependency.

packages/access-policy/package.json
```diff
@@ -13,11 +13,14 @@
...
+    "build:src": "yarn package:build:js",
+    "build:types": "yarn package:build:types",
+    "lint": "yarn package:lint:js",
     "lint:fix": "yarn lint --fix",
-    "test": "jest",
+    "test": "yarn package:test",
     "test:coverage": "yarn test --coverage"
+  },
+  "devDependencies": {
+    "npm-run-all": "^4.1.5"
```
IgorNadj, line 24: With Yarn2+, each package must explicitly declare its dependencies. 
See https://yarnpkg.com/features/workspaces/#what-does-it-mean-to-be-a-workspace

packages/central-server/package.json
```diff
@@ -13,14 +13,14 @@
...
+    "lint": "yarn package:lint:js",
     "lint:fix": "yarn lint --fix",
-    "prestart": "npm run -s build",
+    "prestart": "yarn build",
     "start": "node dist",
-    "start-dev": "../../scripts/bash/backendStartDev.sh 9999",
-    "start-verbose": "LOG_LEVEL=debug npm run start-dev",
+    "start-dev": "yarn package:start:backend-start-dev 9999",
+    "start-verbose": "LOG_LEVEL=debug yarn start-dev",
     "test": "yarn workspace @tupaia/database check-test-database-exists && DB_NAME=tupaia_test mocha",
```
IgorNadj, line 24: Packages which use non-standard scripts e.g. mocha declare them manually and have those dependencies explicitly specified.

packages/devops/ci/tupaia-ci-cd.Dockerfile
```diff
@@ -43,6 +51,8 @@ RUN mkdir -p ./packages/data-lake-api
 COPY packages/data-lake-api/package.json ./packages/data-lake-api
 RUN mkdir -p ./packages/database
 COPY packages/database/package.json ./packages/database
+RUN mkdir -p ./packages/devops
```
IgorNadj, line 54: I found that omitting some packages gives a different lockfile in Yarn2+, which means that the `yarn install --frozen-lockfile` below would fail, because it would want to change the lockfile. So I added them anyway. Adding these has no impact on CI/CD.

packages/lesmis/src/api/queries/utils.js
```diff
@@ -3,7 +3,7 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  *
  */
-import { formatDateForApi, roundStartEndDates } from '@tupaia/ui-components/lib/chart';
+import { formatDateForApi, roundStartEndDates } from '@tupaia/ui-components';
```
IgorNadj, line 6: With ui-components no longer using rollup, I changed these to normal imports.

packages/ui-components/.babelrc.js
```diff
@@ -26,5 +26,11 @@ module.exports = {
         loose: true,
       },
     ],
+    [
+      '@babel/plugin-proposal-private-property-in-object',
```
IgorNadj, line 30: Fixes annoying console output:
```
[ui-components] Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-property-in-object since the "loose" mode option was set to "true" for @babel/plugin-proposal-class-properties.
[ui-components] The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
[ui-components] 	["@babel/plugin-proposal-private-property-in-object", { "loose": true }]
[ui-components] to the "plugins" section of your Babel config.
```

packages/ui-components/package.json
```diff
@@ -11,68 +11,46 @@
...
-    "@rollup/plugin-commonjs": "^18.0.0",
-    "@rollup/plugin-json": "^4.1.0",
-    "@rollup/plugin-multi-entry": "^4.0.0",
-    "@tupaia/utils": "1.0.0",
-    "axios": "^0.21.1",
+    "@storybook/react": "^6.3.9",
     "faker": "^4.1.0",
     "fast-glob": "^3.2.5",
-    "jsoneditor": "^9.0.0",
-    "moment": "^2.29.1",
```
IgorNadj, line 58: We had a setup I don't understand here where some dependencies were marked as `devDependencies` even though they were actually used throughout src. Not sure how it worked before. I moved these to normal `dependencies`, and the final build size of our frontends is the same, e.g. web-frontend was 20MB, and it is still 20MB.

packages/ui-components/src/components/Chart/ChartTooltip.js
```diff
@@ -149,7 +149,7 @@ const NoDataTooltip = ({ payload, periodGranularity }) => {
   );
 };
 
-export const Tooltip = props => {
+export const ChartTooltip = props => {
```
IgorNadj, line 152: We had a couple different `Tooltip`s and three different `Table`s in this package, all being exported at root level, causing errors when importing. I renamed them.

billli0 (approved):
@IgorNadj That's a chuck of works, nice one! Thanks for the e2e testing new package and some dependencies clean up, the chart makes me can't wait to use the Yarn new version.
Got a few small questions but it looks great overall.

packages/web-frontend/package.json
```diff
@@ -12,19 +12,19 @@
   "main": "dist/index.js",
   "scripts": {
     "build": "rm -rf dist && run-s \"build:* {@}\" --",
-    "build:desktop": "SKIP_PREFLIGHT_CHECK=true BUILD_PATH='./build/desktop/' GENERATE_SOURCEMAP=false react-scripts build",
-    "build:mobile": "SKIP_PREFLIGHT_CHECK=true BUILD_PATH='./build/mobile/' GENERATE_SOURCEMAP=false REACT_APP_APP_TYPE=mobile react-scripts build",
+    "build:desktop": "BUILD_PATH='./build/desktop/' yarn package:build:react-scripts",
+    "build:mobile": "BUILD_PATH='./build/mobile/' yarn package:build:react-scripts",
```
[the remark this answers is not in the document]
IgorNadj: Good spotting on REACT_APP_APP_TYPE=mobile! 
The other two env vars are set in package:build:react-scripts

packages/admin-panel-server/package.json
```diff
@@ -16,6 +16,7 @@
   "private": true,
   "scripts": {
     "build": "yarn package:build:ts",
+    "build-dev": "yarn build-dev",
```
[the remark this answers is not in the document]
IgorNadj: This looks like an old version of this PR, this is fixed in latest version.

IgorNadj:
@billli0 I'm not sure if you reviewed an older version of this PR sorry, just pinged you to take a new look, thanks

billli0 (approved):
Looks good!

tcaiger:
Hey @IgorNadj I just wanted to mention for awareness that although tree shaking wasn't working well with rollup, bundling map and chart components separately was useful for saving bundle size in projects that do not use them such as PSSS and admin panel. I think those projects are not critical compared to Tupaia so I'm not worried about it. I think it could be a good idea to split out the chart components and map components into seperate packages in the mono-repo which would make this irrelevant anyway. There is [an old ticket which includes this](https://linear.app/bes/issue/WAITP-96/improve-ui-components-bundling) but it never got picked up. I'd be interested to hear if you agree and if so I might update that ticket.

PR #3995, opened 2022-06-24
RN-400: Meditrak permissions based sync (app-side changes)

rohan-bes:
### Issue RN-400:

Changes to meditrak-app to support permissions based synchronization. Much simpler changes here, refer to [the other PR](https://github.com/beyondessential/tupaia/pull/3992) for the logic


packages/meditrak-app/app/home/HomeToolbarContainer.js
```diff
@@ -16,7 +16,7 @@ const HomeToolbar = ({ onNavigateToSurveysMenu, onNavigateToTupaiaWebsite, style
       textStyle={localStyles.buttonText}
     />
     <BlueButton
-      title="SURVEY FACILITIES"
+      title="SURVEY"
```
rohan-bes, line 19: @edmofro 2 months of refactors and now I'm finally able to make this change 😉

packages/meditrak-app/app/sync/Synchroniser.js
```diff
@@ -28,7 +31,7 @@ const MEASURE_BATCH_IN_DATA = 'kilobytes';
 const KILOBYTE = 1024; // bytes
 
 const MIN_BATCH_RECORDS = 1;
-const MAX_BATCH_RECORDS = 500;
+const MAX_BATCH_RECORDS = 2000;
```
rohan-bes, line 34: Based on the api_request_log, the majority of our syncs manage to hit this 500 max, so I bumped it to 2000. This also gets hit pretty easily on my local, but the app write speed is balanced with the query time more effeciently 👍

packages/meditrak-app/app/sync/Synchroniser.js
```diff
@@ -246,61 +250,50 @@ export class Synchroniser {
    */
   async pushChanges(changesSyncJson) {
     const startTime = new Date().getTime();
-    try {
```
rohan-bes, line 249: Removed all the redundant try/catches. ESlint told me to

packages/meditrak-app/ios/TupaiaMediTrak/Info.plist
```diff
@@ -17,11 +17,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.11.121</string>
+	<string>1.12.123</string>
```
rohan-bes, line 20: Funny new app number here, it seems like we track the version just by the last number? But semver suggests we should bump the middle number for new features...

edmofro: Yes it is funny! Have just updated the slab doc [here](https://beyond-essential.slab.com/posts/releasing-medi-trak-app-via-appcenter-psgr6iak) to explain the reason, but to repost:

> Don't reset or wrap patch number, instead ALWAYS just increment it by 1 with every release.  This is because we use a shortcut where the patch number doubles as the android build number, which needs to increase monotonically.

packages/meditrak-app/app/sync/Synchroniser.js
edmofro: Yes thank you for this!

packages/meditrak-app/app/home/HomeToolbarContainer.js
edmofro: Bahahahaha 😂

packages/meditrak-app/app/sync/Synchroniser.js
```diff
@@ -309,19 +302,30 @@ export class Synchroniser {
...
-        throw new Error(responseJson.error);
-      }
-      if (typeof responseJson.changeCount !== 'number') {
-        throw new Error('Unexpected response from server');
-      }
-      return responseJson.changeCount;
-    } catch (error) {
-      throw error; // Throw error up
+  async getIncomingChangeMetadata(since, filters = {}) {
+    const countriesSynced = this.database.getSetting(COUNTRIES_SYNCED);
```
edmofro, line 306: Would be nice somewhere around here to add a comment briefly describing the way sync is filtered by permissions, so the reader doesn't need to look into the server code to understand what's going on

edmofro (approved):
Two very superficial suggestions, but both are non-blocking. Looking great :-)

packages/meditrak-app/app/sync/Synchroniser.js
rohan-bes: Great point! Will do 😄

packages/meditrak-app/app/sync/Synchroniser.js
```diff
@@ -332,25 +336,61 @@ export class Synchroniser {
...
-      return {
-        changes: responseJson,
-        requestDuration: endTime - startTime,
-      };
-    } catch (error) {
-      throw error; // Throw error up
+    const countriesSynced = this.database.getSetting(COUNTRIES_SYNCED);
+    const permissionGroupsSynced = this.database.getSetting(PERMISSION_GROUPS_SYNCED);
+    if (countriesSynced) {
+      // eslint-disable-next-line no-param-reassign
```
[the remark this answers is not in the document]
rohan-bes: Yeah... you're right huh? I get a bit too cowboy with this pattern from time to time, will fix this 👍

PR #3996, opened 2022-06-24
Hotfix RN-568 - fix issue with sync queue breaking

avaek:
Issue: [RN-568](https://linear.app/bes/issue/RN-568/submitting-an-eh01-survey-response-breaks-the-sync-queue)

This is a hotfix to fix the issue where submitting EH01 survey responses would break the sync queue 

It seems like putting `household` into the `canonical_types` of the `entity_hierarchy` for the Palau project fixes it


IgorNadj (approved):
Nice one 🏌️

PR #4001, opened 2022-06-27
Hotfix: Switched to using translated key for 'Population' in lesmis vitals

rohan-bes:
### Issue https://beyondessential.slack.com/archives/C01MAA3NKM1/p1656303472872569:

Pretty simple issue here. We added a translation for the word `Population` during this release, and the `I18n` library is set up to translate all fields of a report that it knows a translation for. 

It does feel a bit awkward to translate the key we're using on the frontend, as typically keys aren't translated in these kinds of scenarios. But given that this part of the UI is rendered from data coming from a report it seems acceptable to me.

Quite a small change here, mainly just whitespace, changes.


IgorNadj (approved):
LGTM

PR #4002, opened 2022-06-27
RN-527: Fix arithmetic questions not including all variables in expression

rohan-bes:
### Issue RN-527:

Full explanation of this issue outlined [in the ticket](https://linear.app/bes/issue/RN-527#comment-58c49107). Went with the simple fix here of just ensuring that we don't accidentally replace substrings within the question codes. A more structured re-write of this logic might be possible if we break down the expression tree, but I don't know how it build it up again using mathjs...

### DIFF
--- a/packages/central-server/src/apiV2/utilities/translateExpression.js
+++ b/packages/central-server/src/apiV2/utilities/translateExpression.js
@@ -4,6 +4,23 @@
  */
 
 /**
+ * We want to order the codes such that no code is a substring of a later code
+ * This avoids bugs when replacing the values in the regex, as we don't accidentally replace
+ * a substring
+ * @param {string[]} codes
+ */
+const sortByFewestSubstrings = codes => {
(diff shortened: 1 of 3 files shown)


packages/central-server/src/apiV2/utilities/translateExpression.js
```diff
@@ -22,7 +22,10 @@
 export const translateExpression = async (models, rawExpression, codes) => {
   const questionCodeToId = await models.question.findIdByCode(codes);
   let expression = rawExpression;
-  for (const code of codes) {
+
+  // We sort by string length (desc) so that we avoid replacing a substring in the regex
+  const codesOrderedByLength = codes.sort((a, b) => b.length - a.length);
```
rohan-bes, line 27: @IgorNadj kudos and appreciation for your insight here, my previous implementation was a lot more long-winded 😂

IgorNadj (approved):
Great fix, lovely tests

PR #4007, opened 2022-06-29
RN-559 Server side render

billli0:
### Issue #:
RN-559 Server side render

### Changes:
- Instead of exporting dashboard items as images then put together as PDF file, `lesmis-server` will access `lesmis` frontend `pdf-export` endpoint with user's session cookies, then generate PDF file.
- Performance and pdf quality are way better, and size of the file is smaller.
   
### Todo:
- Error handlers are needed for timeout and can not fetch.
- Separate package for dashboard items pdf export, ideally it can be imported by tupaia.


packages/lesmis/src/components/DashboardExportModal/components/A4Page.js
```diff
@@ -2,32 +2,22 @@ import React from 'react';
...
-  if (isFetching) {
-    return null;
-  }
-
-  const isSelected = isExporting ? false : page === currentPage;
-
+export const A4Page = ({ children, ...configs }) => {
   return (
-    <A4Container ref={addToRefs} $isSelected={isSelected}>
+    <A4Container>
```
billli0, line 20: Since we do server side PDF export, we don't need to use `ref` for screenshots exports.

packages/lesmis/src/components/DashboardExportModal/components/Logo.js
```diff
@@ -0,0 +1,32 @@
...
+`;
+
+export const Logo = () => {
+  return (
+    <Container>
+      <Img alt="logo" src="/lesmis-logo.png" />
+      <Text>Ministry of Education and Sports Lao PDR</Text>
+    </Container>
+  );
+};
```
billli0, line 32: This component is introduced because PDF export cannot render svg properly.

packages/lesmis/src/components/DashboardExportModal/components/A4Page.js
```diff
@@ -2,32 +2,22 @@ import React from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { FlexColumn } from '@tupaia/ui-components';
-import { useIsFetching } from 'react-query';
 
 import Header from './Header';
 
 const A4Container = styled.div`
-  width: 1300px;
+  width: 1192px;
```
IgorNadj, line 9: I don't think we need to specify a fixed width for printing, it will figure it out for us no? I just went to print this page in my browser and it fits nicely on A4, even though its like 2000px wide.

Actually you can delete this whole component no?

packages/lesmis/src/components/DashboardExportModal/components/Logo.js
IgorNadj: 🤔 strange, if chrome can render it normally it should be able to print it...

packages/lesmis/src/components/DashboardExportModal/components/PreviewPage.js
```diff
@@ -0,0 +1,38 @@
...
+const PreviewContainer = styled.div`
+  margin-top: 20px;
+  min-width: 900px;
+`;
+
+const Content = styled(FlexColumn)`
+  margin: 0px 150px;
+`;
+
+export const PreviewPage = ({ children, getNextPage, currentPage, isExporting, ...configs }) => {
```
IgorNadj, line 17: Yeah I'm not sure we can do preview like this anymore. Since we have a binary PDF now, the only way to render it would be to use a special iframe of some sort that has src of your /api/v1/pdf

IgorNadj (changes requested):

IgorNadj:
I think there's some stuff I don't get around previews, otherwise looking really good!

packages/lesmis/src/components/DashboardExportModal/components/Logo.js
billli0: Yeah, some svg functions isn't really compatible with PDF export..

IgorNadj: Discussed.

packages/lesmis/src/components/DashboardExportModal/components/A4Page.js
IgorNadj: Discussed. Need fixed width because of how our charts render.
Need component because of this.

packages/lesmis/src/components/DashboardExportModal/components/PreviewPage.js
IgorNadj: Discussed. PDF generation takes a while, so iframe solution is not good. This way is better.

packages/lesmis-server/src/routes/PDFExportRoute.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const browser = await puppeteer.launch();
+    const page = await browser.newPage();
+    await page.setCookie({
+      name: sessionCookieName,
+      domain: body.hostname.split(':')[0], // localhost:8030 -> localhost
+      httpOnly: true,
+      value: sessionCookieValue,
+    });
+    await page.goto(url, { waitUntil: 'networkidle0' });
```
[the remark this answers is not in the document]
billli0: The defaults timeout is 30 seconds.

billli0:
@IgorNadj A couple of nice suggestions, thanks! I've put them into the code.

packages/lesmis-server/src/routes/PDFExportRoute.ts
```diff
@@ -0,0 +1,91 @@
...
+    const cookies = new Cookies(this.req, this.res);
+    const sessionCookieName = 'sessionCookie';
+    const sessionCookieValue = cookies.get(sessionCookieName);
+    if (!sessionCookieValue) {
+      throw new Error(`'sessionCookie' is not found`);
+    }
+    return { sessionCookieName, sessionCookieValue };
+  };
+
+  private verifyBody = (body: any): Body => {
```
IgorNadj, line 43: Now that this.req is typed, you don't need to re-specify the type of the `body` arg, typescript knows it's a Body. And I don't think that return type is right? But anyway, you don't need to re-specify that either, typescript knows it's a string.

packages/lesmis-server/src/routes/PDFExportRoute.ts
```diff
@@ -0,0 +1,91 @@
...
+    const location = new URL(pdfPageUrl);
+    let browser;
+    let result;
+
+    try {
+      browser = await puppeteer.launch();
+      const page = await browser.newPage();
+      await page.setCookie({
+        name: sessionCookieName,
+        domain: location.hostname.split(':')[0], // localhost:8030 -> localhost
```
IgorNadj, line 64: Nice thinking using URL. Splitting not necessary anymore, hostname always returns the domain without the port.

IgorNadj (approved):
Couple small notes, otherwise ship it!

packages/lesmis-server/src/routes/PDFExportRoute.ts
billli0: Yeah type thing is one reason, another reason is to validate the schema of the body, do you think it is a bit unnecessary?

IgorNadj: I see, yeah, makes sense.

PR #4014, opened 2022-07-04
No issue: Add a unit test for get measure display info

billli0:
### Issue #:
Add a unit test to describe how dynamic coloured radius works.

### DIFF
--- a/packages/ui-components/src/components/Map/__tests__/markerFormats.test.js
+++ b/packages/ui-components/src/components/Map/__tests__/markerFormats.test.js
@@ -186,7 +186,8 @@ describe('measures', () => {
       expect(display3).toHaveProperty('color', 'hsl(50, 100%, 50%)');
     });
 
-    it('Coloured radius choose a colour from a spectrum, and pick radius from radius', () => {
+    it('should support coloured radius', () => {
+      // A coloured radius will choose colour from a spectrum option, and pick radius size from a radius option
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/components/Map/__tests__/markerFormats.test.js
```diff
@@ -185,6 +185,25 @@ describe('measures', () => {
       const display3 = getMeasureDisplayInfo({ questionC: 0.5 }, [optionsSpectrum]);
       expect(display3).toHaveProperty('color', 'hsl(50, 100%, 50%)');
     });
+
+    it('Coloured radius choose a colour from a spectrum, and pick radius from radius', () => {
```
IgorNadj, line 189: This test title is a little hard to understand. I suggest we call it something shorter like 'should support coloured radius' and if you want the longer description pop in in a comment inside the test case

packages/web-frontend/src/utils/measures.js
```diff
@@ -104,27 +103,6 @@ const getNullValueMapping = type => {
   return baseMapping;
 };
 
-function getFormattedValue(value, type, valueInfo, scaleType, valueType, submissionDate) {
```
IgorNadj, line 107: I love when code gets deleted, nice one!

IgorNadj (approved):

PR #4029, opened 2022-07-14
Yarn v3 upgrade fixes

IgorNadj:
Followup fixes to yarn v1 -> v3 upgrade #3994 

* Cleans up `__tests__` dirs for consistency
* Upgrade nohoist for meditrak-app
* Move default babel config flags to base package:build:js script
* Each package cleans their dist before building
* Use standard test runners and builders for consistency


billli0 (changes requested):
LGTM! Seems like there is a testing error in `web-config-server` from codeship, wanna have a quick fix?

packages/meditrak-app/.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nmHoistingLimits: workspaces
```
IgorNadj, line 1: Fixes an issue Andrew found with building meditrak-app locally. 
Screen Shot 2022-07-15 at 1 23 06 PM

yarn v3 doesn't support nohoist, but instead we have to say don't hoist anything.

packages/server-boilerplate/src/index.ts
```diff
@@ -20,4 +20,4 @@ export {
 } from './orchestrator';
 export * from './types';
 export * from './models';
-export { TestableServer } from './__tests__/utilities';
+export { TestableServer } from './testUtilities';
```
IgorNadj, line 23: Moved testUtilities out of `__tests__` because `__tests__` doesn't get ouput into dist, and other packages need these testUtilities.

packages/tsutils/package.json
```diff
@@ -12,7 +12,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "build": "rm -rf dist && tsc",
+    "build": "rm -rf dist && npm run --prefix ../../ package:build:ts",
```
IgorNadj, line 15: Use standard ts build script

packages/ui-components/src/__tests__/components/Chart/parseChartConfig.test.js
```diff
@@ -4,8 +4,11 @@
  *
  */
 
-import { parseChartConfig } from '../parseChartConfig';
-import { CHART_COLOR_PALETTE, EXPANDED_CHART_COLOR_PALETTE } from '../constants';
+import { parseChartConfig } from '../../../components/Chart/parseChartConfig';
```
IgorNadj, line 7: Update ui-components folder structure to put all tests under `__tests__` instead of alongside each component for consistency with all other packages.

packages/weather-api/package.json
```diff
@@ -13,19 +13,16 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "build": "rm -rf dist && run-p \"build:* {@}\" --",
-    "build:src": "npm run --prefix ../../ package:build:js --ignore \"src/tests/**\"",
+    "build:src": "npm run --prefix ../../ package:build:js",
     "build:types": "npm run --prefix ../../ package:build:types",
     "build-dev": "npm run build",
-    "test": "nyc --reporter=lcov mocha",
-    "test-coverage": "cross-env NODE_ENV=test nyc mocha"
+    "test": "yarn package:test",
```
IgorNadj, line 19: Use standard test runner

IgorNadj:
@billli0 Went to far in my original changes, had to roll back some packages. I was hoping we could get our old packages like web-config-server to use the normal build and test configs, but they have special needs so I rolled them back to use their old custom build and test scripts. Ready for re-review.

packages/central-server/package.json
```diff
@@ -13,7 +13,7 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "build": "rm -rf dist && run-p -c \"build:* {@}\" --",
-    "build:src": "yarn package:build:js --ignore \"**src/database/migrations\",\"src/tests/**\" --copy-files --no-copy-ignored",
+    "build:src": "yarn package:build:js",
```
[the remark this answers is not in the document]
billli0: Can we ignore them to make the bundle file smaller?

billli0 (approved):
New changes look good, thanks for the fix. Definitely it would be nice to have `web-config-server` updated to be more standardised as other packages. 
You mentioned the `web-config-server` have its own needs, what would that be? Will it be a big one for us the refactor?

IgorNadj:
> New changes look good, thanks for the fix. Definitely it would be nice to have `web-config-server` updated to be more standardised as other packages. You mentioned the `web-config-server` have its own needs, what would that be? Will it be a big one for us the refactor?

There are still a few packages that use mocha. To get these to use the standard test runner (jest) we will need to go through all tests and migrate them, some of the syntax is different, like expect().to.be instead of expect().toBe or something like that. Not too big but I thought it is probably out of scope for this PR.

packages/central-server/package.json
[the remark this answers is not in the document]
IgorNadj: I came back to this with a fresh head and moved these flags to the .babelrc.js file. This way we can use the standard build script as well as customise the build babel config for central-server. 

We are now back to 9MB builds, 3MB smaller (same as dev), and as fast as dev.

billli0:
> There are still a few packages that use mocha. To get these to use the standard test runner (jest) we will need to go through all tests and migrate them, some of the syntax is different, like expect().to.be instead of expect().toBe or something like that. Not too big but I thought it is probably out of scope for this PR.

Yeah I'm just a little bit curious. It is definitely out of scope. 
I was working on `web-config-server` unit test and I'm not really satisfied with `mocha`.. I have some interest to migrate that, will see how that goes.

PR #4030, opened 2022-07-15
MAUI-1057: Add entityImage hook to household question

chris-pollard:
### Changes:
- Add migration script to update hook field to 'entityImage' for Palau project

### DIFF
--- a/packages/database/src/migrations/20220714065133-AddEntityImageHookToHouseholdQuestion-modifies-data.js
+++ b/packages/database/src/migrations/20220714065133-AddEntityImageHookToHouseholdQuestion-modifies-data.js
@@ -1,0 +1,30 @@
+'use strict';
+
+import { codeToId, updateValues } from '../utilities';
+
+var dbm;
+var type;
+var seed;
+
+exports.setup = function (options, seedLink) {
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20220714065133-AddEntityImageHookToHouseholdQuestion-modifies-data.js
```diff
@@ -0,0 +1,30 @@
...
+  seed = seedLink;
+};
+
+const QUESTION_CODE = 'PW_EH01_008';
+const QUESTION_TABLE = 'question';
+const HOOK = 'entityImage';
+
+exports.up = async function (db) {
+  const questionId = await codeToId(db, 'question', QUESTION_CODE);
+  updateValues(db, QUESTION_TABLE, { hook: HOOK }, { id: questionId });
```
billli0, line 21:
```suggestion
  await updateValues(db, QUESTION_TABLE, { hook: HOOK }, { id: questionId });
```

billli0 (approved):
Looks good! We might need to add `await` or `return` before the async function to make sure it runs properly.

PR #4035, opened 2022-07-19
Fix puppeteer not working on feature branch

billli0:
### Changes:
- In lesmis, session cookie domain is targeting `lesmis-api.tupaia.org`, instead of `lesmis.tupaia.org`, changes will reflect that.
- Dependencies for puppeteer


billli0:
@IgorNadj I've put a commit for those issues we've chatted in the call. Ready to review now.

IgorNadj (approved):
Nice one

PR #4036, opened 2022-07-21
MAUI-859: Enable Project Creation

chris-pollard:
## Changes:

- Add config in Project Page resource component in admin-panel react app.
- Add endpoint for getting existing entity data types.
- Add endpoint for creating a project.
- Add validation rules for new project record data.

---

### Screenshots:
Screen Shot 2022-07-21 at 11 20 04 am


chris-pollard:
Have also realised that I should write up a unit test, which I have not done yet.

packages/central-server/src/apiV2/GETEntityTypes.js
```diff
@@ -0,0 +1,55 @@
...
+    let options = await this.getDbQueryOptions();
+    let criteria = this.getDbQueryCriteria();
+    if (this.permissionsFilteredInternally) {
+      ({ dbConditions: criteria, dbOptions: options } = await this.applyPermissionsFilter(
+        criteria,
+        options,
+      ));
+    }
+    const { types } = this.models.entity;
+    const entityTypes = Object.values(types)
```
IgorNadj, line 38: Technically this endpoint should return all types, including world. There may be another frontend that wants to list it.

You can then either just leave World in the dropdown and leave it to users to ignore it or add this same filter in admin-panel instead (not sure how off the top of my head but should be possible).

packages/central-server/src/apiV2/projects/CreateProject.js
```diff
@@ -0,0 +1,110 @@
...
+      description,
+      sort_order,
+      image_url,
+      logo_url,
+      permission_groups,
+      countries,
+      entityTypes,
+    } = this.newRecordData;
+
+    await this.createProjectEntity(code, name);
```
IgorNadj, line 34: It's cool that we do all this together! Maybe just need to wrap these in a transaction block so if one fails they all rollback.

Here's an example https://github.com/beyondessential/tupaia/blob/dev/packages/central-server/src/apiV2/import/importDisaster.js#L53

packages/central-server/src/apiV2/projects/CreateProject.js
```diff
@@ -0,0 +1,110 @@
...
+    await this.createEntityHierarchy(code, entityTypes);
+    await this.createProjectEntityRelations(code, countries);
+    await this.createProjectDashboard(name, code);
+
+    const [projectEntity] = await this.models.database.find('entity', { 'entity.code': code });
+    const [projectEntityHierarchy] = await this.models.database.find('entity_hierarchy', {
```
IgorNadj, line 40: Same as above

packages/central-server/src/apiV2/projects/CreateProject.js
```diff
@@ -0,0 +1,110 @@
...
+      'entity_hierarchy.name': code,
+    });
+    const defaultMapOverlayCode = '126';
```
IgorNadj, line 43: Hmmm I don't love it 😆 
Is there a way to make this not-hardcoded?

packages/central-server/src/apiV2/utilities/constructNewRecordValidationRules.js
```diff
@@ -186,6 +186,74 @@ export const constructForSingle = (models, recordType) => {
...
+            const projects = await models.project.find({
+              code: projectCode,
+            });
+            if (projects.length > 0) {
+              throw new Error('This project code already exists');
+            }
+            return true;
+          },
+        ],
+        name: [
```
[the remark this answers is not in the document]
chris-pollard: @IgorNadj I thought that perhaps given that there are a finite number of projects, and given that they all get assembled together for the user to view, it might be odd to have the same name.

packages/central-server/src/apiV2/utilities/constructNewRecordValidationRules.js
```diff
@@ -186,6 +186,74 @@ export const constructForSingle = (models, recordType) => {
...
+            const permissionGroups = await models.permissionGroup.find({
+              name: permissionGroupNames,
+            });
+            if (permissionGroupNames.length !== permissionGroups.length) {
+              throw new Error('Some provided permission groups do not exist');
+            }
+            return true;
+          },
+        ],
+        description: [hasContent],
```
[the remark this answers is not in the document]
chris-pollard: @IgorNadj  good point, thanks

packages/central-server/src/apiV2/projects/CreateProject.js
chris-pollard: @IgorNadj ah yeah, thats a great point. I was having this issue when testing locally.

chris-pollard:
@IgorNadj this is ready to review again. Note that there is no specific rollback unit test.

packages/central-server/src/apiV2/projects/CreateProject.js
```diff
@@ -0,0 +1,118 @@
...
+        entity_hierarchy_id: projectEntityHierarchyId,
+      });
+    });
+  }
+
+  async createProjectEntity(models, code, name) {
+    const worldCode = 'World';
+    const { id: worldId } = await models.entity.findOne({ 'entity.code': worldCode });
+
+    await models.entity.create({
```
IgorNadj, line 72: Minor: `.create()` returns the record with an id, so you can return that instead of querying again to get it above on line 42

packages/central-server/src/apiV2/projects/CreateProject.js
IgorNadj: Nice solution! Thanks :)

packages/central-server/src/apiV2/GETEntityTypes.js
IgorNadj: This seems to still be the case, but it's minor

IgorNadj (approved):
Only minor comments left, nice one 🏌️

chris-pollard:
Thanks for the feedback @IgorNadj

PR #4037, opened 2022-07-21
MAUI-978: Return default year to current

chris-pollard:
### Changes:

-Revert default year for reports to current .


IgorNadj:
Do we want to follow this? Or has something changed? (from card)
>As we will probably get 2023 data late in the year, let's stick with the default being 2022 not 'current year' or we will end up in the same situation early next year.

chris-pollard:
@IgorNadj yes we should follow this. Will update.

PR #4039, opened 2022-07-22
MAUI-1042: Add Household Head Attribute

chris-pollard:
## Changes:

- Add household_head attribute to entity after survey response submitted.


packages/central-server/src/hooks/utilities.js
```diff
@@ -34,3 +34,17 @@ export function parseCoordinates(answerText) {
 
   return { latitude, longitude };
 }
+
+export function updateJsonFieldValue(currentJsonValues, newKeyValues) {
+  if (Object.keys(currentJsonValues).length > 0) {
+    const updatedKeyValues = { ...currentJsonValues };
```
edmofro, line 40: Great work spreading rather than mutating the parameter :+1:

packages/central-server/src/hooks/utilities.js
```diff
@@ -34,3 +34,17 @@ export function parseCoordinates(answerText) {
...
+    const newKeys = Object.keys(newKeyValues);
+
+    newKeys.forEach(key => {
+      updatedKeyValues[key] = newKeyValues[key];
+    });
+
+    return updatedKeyValues;
+  }
+  return newKeyValues;
+}
```
edmofro, line 50: Nice solve, but it's actually even easier:
```suggestion

export function updateJsonFieldValue(currentJsonValues, newKeyValues) {
 return { ...currentJsonValues, ...newKeyValues };
}
```
Magic huh!? 
In fact, this is so straightforward I'd suggest not even making a util function here, and just doing the above line wherever it's needed :-)

edmofro: Note that this will work even if currentJsonValues is undefined ✨

packages/central-server/src/tests/hooks/questionHooks.test.js
```diff
@@ -258,6 +268,68 @@ describe('Question hooks', () => {
...
+
+        const entity = await models.entity.findById(ENTITY2_ID);
+        expect(entity.attributes).to.deep.equal(newValue);
+
+        // should be otherwise unchanged
+        const { attributes: beforeAttributes, ...beforeData } = await beforeEntity.getData();
+        const { attributes: afterAttributes, ...afterData } = await entity.getData();
+        expect(beforeData).to.deep.equal(afterData);
+      });
+    });
```
edmofro, line 331: Fantastic tests!

packages/central-server/src/hooks/entityHouseholdHead.js
```diff
@@ -0,0 +1,19 @@
+import { updateJsonFieldValue } from './utilities';
+
+export async function entityHouseholdHead({ answer, surveyResponse }) {
```
edmofro, line 3: This is a really tidy hook! There's the potential to take it further by allowing it to set any entity attribute, what do you think about having a go at that?

You'd require the question config to include a field like `entityAttribute`, and then in here you'd query the answer for its question and get the attribute to set from that.

edmofro (changes requested):
Really nice work here @chris-pollard, got a few suggestions for you, some optional :-)

packages/central-server/src/hooks/entityHouseholdHead.js
chris-pollard: @edmofro do you mean that there would be a property associated with a question called 'entityAttribute' with a value which would be end up being the 'key' in the attribute field along with the string value of the answer? Where would that property be set?

edmofro: That's right! I was thinking the property would be set in `config`, but I've just checked and that's `survey_screen_component.config` not `question.config`. We could add a new `hook_config` column to `question`, or we could abuse the hook name itself using some convention like calling the hook `entityAttributeHouseholdHead`, where any hook starting with `entityAttribute` resolves to the same hook handler

chris-pollard: @edmofro I've added another couple of commits, with the ability to add custom entity attributes. Let me know what you think.

packages/central-server/src/hooks/registry.js
```diff
@@ -5,8 +5,10 @@ export function registerHook(name, callback) {
   return callback;
 }
 
+const isEntityAttribute = name => name.substring(0, 15) === 'entityAttribute';
```
edmofro, line 8:
```suggestion
const isEntityAttribute = name => name.startsWith(ENTITY_ATTRIBUTE_HOOK_PREFIX);
```

packages/central-server/src/hooks/entityAttribute.js
```diff
@@ -0,0 +1,20 @@
+const snakeCased = name =>
```
edmofro, line 1: Nice little util, but can avoid adding more code by using
```suggestion
import { snake } from 'case';
```

packages/central-server/src/hooks/entityAttribute.js
```diff
@@ -0,0 +1,20 @@
...
+    .toLowerCase();
+
+export async function entityAttribute({ answer, surveyResponse, hookName }) {
+  const entity = await surveyResponse.entity();
+
+  if (!entity) {
+    throw new Error('Invalid entity');
+  }
+
+  const attributeKey = snakeCased(hookName.substring(15));
```
edmofro, line 14: When there's a number like this that is defined without context of where it comes from, we call it a "magic number"...and it's a bit of an anti-pattern. Instead much better to show the reader exactly why it's 15 characters long :-)
```suggestion
  const attributeKey = snakeCased(hookName.substring(ENTITY_ATTRIBUTE_HOOK_PREFIX.length));
```

packages/central-server/src/tests/hooks/questionHooks.test.js
```diff
@@ -299,8 +299,8 @@ describe('Question hooks', () => {
         expect(beforeData).to.deep.equal(afterData);
       });
 
-      it("Should add household_head to entity's attributes when there are existing attributes", async () => {
-        const TEST_HOUSEHOLD_HEAD = 'Will Smith';
+      it("Should add custom attribute to entity's attributes when there are existing attributes", async () => {
```
edmofro, line 302: Scope for a third test here to make sure it successfully overwrites any existing value with the same key. Not a requirement, but if you have time :-)

edmofro (changes requested):
You've built a really powerful feature here @chris-pollard, this will save any future requests for "can we add a hook for attribute x" 🎉 

Suggest adding a constant somewhere for the prefix, and referencing it across both files

packages/central-server/src/hooks/entityAttribute.js
chris-pollard: Great teachings, @edmofro cheers.

chris-pollard:
All changes made, with additional test. Looking much better, thanks @edmofro

edmofro (approved):
Looks great, thanks @chris-pollard!

PR #4042, opened 2022-07-25
Add upper first case function in vizbuilder

billli0:
### Issue #:
Upper first case for every word in one string.


edmofro (approved):
Thanks for just importing the specific lodash util rather than the whole (heavy) package!

PR #4043, opened 2022-07-25
Fix unable to view loading circular progress

billli0:
### Issue #:
https://linear.app/bes/issue/RN-476#comment-5192a9d2

### DIFF
--- a/packages/lesmis/src/components/DashboardExportModal/DashboardExportModal.js
+++ b/packages/lesmis/src/components/DashboardExportModal/DashboardExportModal.js
@@ -18,7 +18,7 @@ import MuiIconButton from '@material-ui/core/Button';
 import { OptionsBar } from './components';
 import { I18n, useDashboardItemsExportToPDF, useUrlParams, useUrlSearchParam } from '../../utils';
 import { DEFAULT_DATA_YEAR } from '../../constants';
-import { DASHBOARD_EXPORT_PREVIEW, ExportView } from '../../views/ExportView';
+import { DASHBOARD_EXPORT_PREVIEW, ExportView as BaseExportView } from '../../views/ExportView';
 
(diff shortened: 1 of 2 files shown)


packages/lesmis/src/components/DashboardExportModal/DashboardExportModal.js
```diff
@@ -34,6 +34,10 @@ const MuiButton = styled(MuiIconButton)`
   color: #666666;
 `;
 
+const EmptyView = styled.div`
+  height: 50vh;
```
edmofro, line 38: Where does 50vh come from? Seems pretty magic to me, and I can't see what it relates to...

edmofro (changes requested):
Got a question for you

packages/lesmis/src/components/DashboardExportModal/DashboardExportModal.js
billli0: Yeah it is a very special case for `LoadingContainer`. `EmptyView` will help us to control the height of `LoadingContainer` dynamically.
The first thing is the height of `LoadingContainer` is the height of its child component. The existing problem is `LoadingContainer` current child is`ExportView`, and its height is fixed, so when loading, if client has a small screen they cannot see the loading circular progress.
I choose to adapt the behaviour of `LoadingContainer` since it got used for many place.

edmofro: Right I get it!

How about something like this? https://github.com/beyondessential/tupaia/pull/4048/files

The `EmptyView` wasn't as clear to me as an "isLoading" prop that defines the different styles

billli0: That's much clearer @edmofro thanks! Nice one.

edmofro (approved):
Nice changes ;-)

PR #4046, opened 2022-07-26
RN-587 Add pdf export server

billli0:
### Issue #:
RN-587

### Changes:
- New micro service `pdf-export-server` for pdf export, it will be used by `lesmis` and `tupaia`


packages/lesmis-server/src/routes/PDFExportRoute.ts
```diff
@@ -25,67 +20,20 @@ export type PDFExportRequest = Request<
 >;
 
 export class PDFExportRoute extends Route<PDFExportRequest> {
+  private readonly apiClient: TupaiaApiClient;
```
IgorNadj, line 23: This should really be provided by server-boilerplate. Lesmis-server is an Orchestrator Api https://github.com/beyondessential/tupaia/blob/dev/packages/lesmis-server/src/app/createApp.ts#L46

Server-boilerplate already gives us an api-client, but only for the other type of server.
- Microservice Api: https://github.com/beyondessential/tupaia/blob/dev/packages/server-boilerplate/src/microService/api/ApiBuilder.ts#L90 <-- present
- Orchestrator Api: https://github.com/beyondessential/tupaia/blob/dev/packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts <-- missing

Can you please add the same `attachApiClientToContext` method to orchestrator api and then have lesmis-server call it in createApp instead. Then here in this route you can change `this.apiClient.pdfExport()` to `this.req.ctx.services.pdfExport()`.

packages/pdf-export-server/.npmrc
```diff
@@ -0,0 +1 @@
+scripts-prepend-node-path=true
```
IgorNadj, line 1: I don't know if this is needed?

packages/pdf-export-server/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "scripts": {
+    "build": "npm run --prefix ../../ package:build:ts",
+    "build-dev": "npm run build",
+    "lint": "yarn package:lint:ts",
+    "lint:fix": "yarn lint --fix",
+    "prestart": "npm run build",
+    "start": "node dist",
+    "start-dev": "yarn package:start:backend-start-dev 9993 -ts",
+    "start-verbose": "LOG_LEVEL=debug yarn start-dev",
+    "test": "yarn package:test:withdb --runInBand",
```
IgorNadj, line 26:
```suggestion
    "test": "yarn package:test",
```

packages/pdf-export-server/src/@types/express/index.d.ts
```diff
@@ -0,0 +1,15 @@
+/**
```
IgorNadj, line 1: Do you need this file?

IgorNadj (approved):
Small comment about apiClient but apart from that looks good!

packages/pdf-export-server/.npmrc
billli0: I've seeen it in other packages, thought that's the default setting. Do you suggest we remove it?

IgorNadj: Maybe something I added.... it's ok will clean up in a separate PR

packages/pdf-export-server/src/@types/express/index.d.ts
billli0: Nope. Have deleted it.

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -155,14 +160,29 @@ export class ApiBuilder {
...
   public use<T extends ExpressRequest<T> = Request>(
     path: string,
-    middleware: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>,
+    ...middleware: RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>[]
   ) {
     this.app.use(
       this.formatPath(path),
       this.attachSession as RequestHandler<Params<T>, ResBody<T>, ReqBody<T>, Query<T>>,
-      middleware,
+      ...middleware,
```
billli0, line 185: Be consistent with `use` in `microService`

packages/lesmis-server/src/app/createApp.ts
```diff
@@ -46,6 +47,7 @@ export function createApp() {
   const app = new OrchestratorApiBuilder(new TupaiaDatabase(), 'lesmis')
     .useSessionModel(LesmisSessionModel)
     .useAttachSession(attachSession)
+    .attachApiClientToContext(authHandlerProvider) // after useAttachSession
```
billli0, line 50: Not sure if this comment `after useAttachSession` is necessary.. But it took me a while to figure it out.

IgorNadj:
Much cleaner, thanks for making those changes!

PR #4049, opened 2022-07-28
Add logging of web-config-server requests to the api_request_log table

rohan-bes:
### Issue https://beyondessential.slack.com/archives/C02HES8HFQF/p1658966488670609:

Just added logging of api requests to the web-config-server app.

### DIFF
--- a/packages/web-config-server/src/app.js
+++ b/packages/web-config-server/src/app.js
@@ -58,9 +58,11 @@ export async function createApp() {
   // Initialise sessions
   bindUserSessions(app);
 
+  // Log api requests
+  app.use(logApiRequest(modelRegistry, 'tupaia', 1));
+
   // API router
   app.use('/api/v1', getRoutesForApiV1());
-  app.use('/api/v1', logApiRequest(modelRegistry, 'tupaia', 1));
 
   // Handle errors
   app.use(handleError);
--- a/packages/web-config-server/src/app.js
+++ b/packages/web-config-server/src/app.js
@@ -3,7 +3,6 @@ import '@babel/polyfill';
 import http from 'http';
 import express from 'express';
 import compression from 'compression';
-import morgan from 'morgan';
 import bodyParser from 'body-parser';
 import cookieParser from 'cookie-parser';
 import cors from 'cors';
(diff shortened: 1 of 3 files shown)


packages/web-config-server/src/app.js
```diff
@@ -62,6 +60,7 @@ export async function createApp() {
 
   // API router
   app.use('/api/v1', getRoutesForApiV1());
+  app.use('/api/v1', logApiRequest(modelRegistry, 'tupaia', 1));
```
rohan-bes, line 63: Strictly speaking I suppose this should be called `web-config` rather than `tupaia`. However, I think that in an ideal world this server would be called `tupaia-server`, and I wouldn't mind kickstarting that transition. Also, since non-devs may be viewing these logs, having a user friendly name seems apt. Happy to hear reviewer's thoughts though 👍

packages/web-config-server/src/utils/logApiRequest.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+export const logApiRequest = (models, apiName, version) => async (req, res, next) => {
```
edmofro, line 6: Would the util exported from `server-boilerplate` work here instead?

edmofro (approved):
lgtm, just wonder if we need to repeat that util code

packages/web-config-server/src/utils/logApiRequest.js
rohan-bes: Technically the way the `userId` is determined is different. However if I'm being honest  I was a bit worried about creating a dependency on server-boilerplate. Now that I check, there's already a dependency there! 🤦

I thiiiink it's might still be cleaner to do it separate rather than try and make the `userId` logic more generic... hopefully we switch this bad boy over to server-boilerplate sometime in the not too distant future and then we can scrap this?

edmofro: Sweet sounds good - so long as you'd considered it!

PR #4052, opened 2022-07-29
RN-597: Explicitly install ndk 21.0.6113669 in appcenter-pre-build.sh

rohan-bes:
### Issue RN-597:

Followed the advice in [this issue](https://github.com/realm/realm-js/issues/4740) in the Realm project. It's a bit of a simplistic solution, the issue's only a day old, so maybe we can keep and eye on it and see if it gets fixed by the Realm team more effectively?


edmofro (approved):
Yeah I don't love this, but does seem like a good stopgap. Could you put a reminder in your calendar to check the issue for updates in a month? 

Realm is not great at keeping devs happy so it might take a while for them to fix. Have you seen the issue about realm making debugging impossible? They've got so much hate for it over about 5 or so years, and still not done anything!

PR #4055, opened 2022-07-29
Fix lesmis cannot export as PNG file

billli0:
### Changes:
Downgrade recharts version as before.


rohan-bes:
@billli0 any background on why this version change actually broke the png exports?

billli0:
@rohan-bes No quite sure.. `recharts` doesn't have a migration guide from v1 to v2, or release note, which makes thing hard to debug.

rohan-bes (approved):
Ah that's all good then. Worth knowing about this issue if we want/need to upgrade in the future...

PR #4062, opened 2022-08-01
hotfix: Add lesmis la is valid url to pdf exporter

billli0:
### Issue #:
pdf export is broken in `lesmis.la`

### DIFF
--- a/packages/devops/scripts/deployment/setupGoldMaster.sh
+++ b/packages/devops/scripts/deployment/setupGoldMaster.sh
@@ -39,44 +39,44 @@ rm -rf lastpass-cli
 mkdir -p /home/ubuntu/.local/share/lpass
 
 # install puppeteer dependencies https://pptr.dev/15.3.0/troubleshooting#chrome-headless-doesnt-launch-on-unix
-sudo apt-get -y install \
-ca-certificates \
-fonts-liberation \
-libappindicator3-1 \
-libasound2 \
-libatk-bridge2.0-0 \
-libatk1.0-0 \
-libc6 \
-libcairo2 \
-libcups2 \
-libdbus-1-3 \
-libexpat1 \
-libfontconfig1 \
-libgbm1 \
-libgcc1 \
-libglib2.0-0 \
-libgtk-3-0 \
-libnspr4 \
-libnss3 \
-libpango-1.0-0 \
-libpangocairo-1.0-0 \
-libstdc++6 \
-libx11-6 \
-libx11-xcb1 \
-libxcb1 \
-libxcomposite1 \
-libxcursor1 \
-libxdamage1 \
-libxext6 \
-libxfixes3 \
-libxi6 \
-libxrandr2 \
-libxrender1 \
-libxss1 \
-libxtst6 \
-lsb-release \
-wget \
-xdg-utils 
+sudo apt-get -yqq install \
+  ca-certificates \
+  fonts-liberation \
+  libappindicator3-1 \
+  libasound2 \
+  libatk-bridge2.0-0 \
+  libatk1.0-0 \
+  libc6 \
+  libcairo2 \
+  libcups2 \
+  libdbus-1-3 \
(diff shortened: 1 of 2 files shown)


packages/devops/scripts/deployment/setupGoldMaster.sh
```diff
@@ -39,7 +39,7 @@ rm -rf lastpass-cli
 mkdir -p /home/ubuntu/.local/share/lpass
 
 # install puppeteer dependencies https://pptr.dev/15.3.0/troubleshooting#chrome-headless-doesnt-launch-on-unix
-sudo apt-get -y install \
+sudo apt-get -yqq install \
```
billli0, line 42: dependencies didn't install in gold master, see if this will fix.

rohan-bes: Curious to know why adding `qq` to the args would fix it? Doesn't that just affect logging?

packages/devops/scripts/deployment/setupGoldMaster.sh
```diff
@@ -39,44 +39,44 @@ rm -rf lastpass-cli
...
+  libatk1.0-0 \
+  libc6 \
+  libcairo2 \
+  libcups2 \
+  libdbus-1-3 \
+  libexpat1 \
+  libfontconfig1 \
+  libgbm1 \
+  libgcc1 \
+  libglib2.0-0 \
```
billli0, line 57: @rohan-bes And also the spaces in front of each dependency.

packages/lesmis-server/src/routes/PDFExportRoute.ts
```diff
@@ -38,12 +38,17 @@ export class PDFExportRoute extends Route<PDFExportRequest> {
   };
 
   private verifyBody = (body: any): Body => {
+    const legitLESMISHostnames = ['lesmis.la', 'www.lesmis.la'];
```
rohan-bes, line 41: I'd suggest having an explicit list of the domains that Tupaia supports

```suggestion
    const validDomains= ['.tupaia.org', '.lesmis.la'];
```

And then later do:
```
if (!validDomains.some(domain=> location.hostname.endsWith(domain)) && !location.hostname.endsWith('localhost')) {
```

Might even be best to store this list of valid hostnames in a utils file/function somewhere, but I can't see another area in the code where do we such a check, so it seems okay to me for it to be here for now 👍

rohan-bes (approved):
Minor code readability change, but aside from that looks good to me 👍

rohan-bes:
@billli0 you know the drill when it comes to making changes to the `setupGoldMaster.sh` file? How you have to import it to s3 and manually kick off a build

packages/devops/scripts/deployment/setupGoldMaster.sh
rohan-bes: Ahh whitespace issues, I love em 😂 😭

packages/lesmis-server/src/routes/PDFExportRoute.ts
billli0: Thanks for the renaming suggestion. 
This check is for security, because `pdf-export-server` will access provided `pdfPageUrl`, and without this check, hacker could easily get user's session by a malicious link as `pdfPageUrl`.

billli0:
@rohan-bes Yes I know how to update this file in s3, but I don't have permission to kick off a build. Could you grant me the access?

packages/lesmis-server/src/routes/PDFExportRoute.ts
billli0: Just found that we have to keep `lesmis.la` because it is a legit DNS for lesmis as well.

rohan-bes:
@billli0 okay have bumped your permissions, wanna give it another shot?

billli0:
Thanks @rohan-bes, it works!

PR #4065, opened 2022-08-02
Resolve merge conflict from master part 1

billli0:
### Issue #:
master and dev modified `PDFExportRoute.ts` at the same time. This PR will add back the previous code so that master can merge in without conflict.

There will be a second PR to update this file, after master merged in.

### DIFF
--- a/packages/devops/scripts/deployment/setupGoldMaster.sh
+++ b/packages/devops/scripts/deployment/setupGoldMaster.sh
@@ -39,44 +39,44 @@ rm -rf lastpass-cli
 mkdir -p /home/ubuntu/.local/share/lpass
 
 # install puppeteer dependencies https://pptr.dev/15.3.0/troubleshooting#chrome-headless-doesnt-launch-on-unix
-sudo apt-get -y install \
-ca-certificates \
-fonts-liberation \
-libappindicator3-1 \
-libasound2 \
-libatk-bridge2.0-0 \
-libatk1.0-0 \
-libc6 \
-libcairo2 \
-libcups2 \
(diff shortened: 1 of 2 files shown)


rohan-bes:
@billli0 what was your process with this?

I'd probably suggest we do the following here:
1. Create a branch off the latest dev
2. Merge master into that branch
3. Resolve conflicts (correctly so that code is working as it should)
4. Create a PR against dev
5. Once that PR is approved, merge using 'merge commit' (not 'squash n merge')

Then there should be no need to merge master into dev I think....

billli0:
@rohan-bes The difficulty is in master we add code to `PDFExportRoute.ts` at where we deleted code. With that wouldn't step 3 `resolve conflict` will create a new commit and it will not cover master new changes?

I've pushed a new commit in this PR, it should make things work.


I've done some changes to make code work.

billli0:
My process would be: 
1. Add deleted code in dev so master and dev will have no conflict
2. Merge master into dev
3. Update dev to the latest code, and remove code from step 1

rohan-bes:
>With that wouldn't step 3 resolve conflict will create a new commit and it will not cover master new changes?

Yes step 3 would create a new commit, but that's a good thing. In that commit we can ensure that the changes from master are ported to the proper location in the code 👍

Happy to do a call on this if you like?

billli0:
As discussed, once that PR is approved, merge it using 'merge commit' (not 'squash n merge').

rohan-bes (approved):
Nice @billli0, thanks for doing this 👍

PR #4067, opened 2022-08-02
RN-601: Added ability to reference query parameters in viz-builder expressions

rohan-bes:
### Issue RN-601:

I've wanted this feature for a while, as there are occasional cases where it'd be really handy to reference the original request parameters of the viz.

I see this as also paving the way for being able to have vizes take additional parameters (eg. A control to filter by Drug Type) which would be really cool.

It's also a pre-requisite for moving 'fetch' into the transform steps, which we want to have in place prior to moving on with the Data Tables epic

### DIFF
--- a/packages/report-server/src/__tests__/reportBuilder/transform/parser/parser.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/transform/parser/parser.test.ts
@@ -8,24 +8,28 @@ import { buildTransform } from '../../../../reportBuilder/transform';
 
 describe('parser', () => {
   it('can do lookups', () => {
-    const transform = buildTransform([
-      {
-        transform: 'updateColumns',
-        insert: {
-          variable: '=$BCD1',
-          current: '=@current.BCD1',
-          index: '=@index',
-          previous: '=@previous.BCD1',
-          next: '=@next.BCD1',
-          lastAll: '=last(@all.BCD1)',
-          sumAllPrevious: '=sum(@allPrevious.BCD1)',
-          sumWhereMatchingOrgUnit:
-            '=sum(where(f(@otherRow) = eq($organisationUnit, @otherRow.organisationUnit)).BCD1)',
-          tableLength: '=length(@table)',
+    const transform = buildTransform(
+      [
+        {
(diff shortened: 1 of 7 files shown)


packages/report-server/src/__tests__/reportBuilder/transform/parser/parser.test.ts
```diff
@@ -8,24 +8,28 @@ import { buildTransform } from '../../../../reportBuilder/transform';
...
+            current: '=@current.BCD1',
+            index: '=@index',
+            previous: '=@previous.BCD1',
+            next: '=@next.BCD1',
+            lastAll: '=last(@all.BCD1)',
+            sumAllPrevious: '=sum(@allPrevious.BCD1)',
+            sumWhereMatchingOrgUnit:
+              '=sum(where(f(@otherRow) = eq($organisationUnit, @otherRow.organisationUnit)).BCD1)',
+            tableLength: '=length(@table)',
+            requestParam: '= @params.animal',
```
billli0, line 26: I feel `@params` could not explain it is from request , which could be a bit confusing when we have `@params.organisationUnitCodes` and `$organisationUnitCode` at the same time. Perhaps we use `@request` or `@requestParams`? What do you think?

billli0 (approved):
Very handy function! Thanks for adding it. Just got a question about the name of `@params`.
I hope we can present and even modify request params in viz builder somedays.

packages/report-server/src/__tests__/reportBuilder/transform/parser/parser.test.ts
rohan-bes: Actually, initially I did have this as `@request` but then decided to change it to `@params`. Basically that decision was motivated by the data-tables re-architecture. We're planning on having each data-table define it's 'Parameters' (eg. entity, startDate, endDate, etc.) so that's why I went with this term, as eventually all reports are going to be converted into data-tables.

Also, for what it's worth I feel the fact that the it's a 'request' is some of an implementation detail that we should hide from the user. They don't know that we use web-requests to communicate, nor should they

PR #4068, opened 2022-08-03
Add transposeTable transform

billli0:
### Issue #:
RN602

### Screenshots:
https://linear.app/bes/issue/RN-602#comment-072dd05e


rohan-bes:
Hey @billli0 I'd love to get a bit more context around the motivation for this transform, and in what scenarios we want to use it.

Also wouldn't mind talking through how this transform works. As I understand, transposition usually refers to flipping a table's x/y. I'd expect there to need to be some kind of mapping between row index and column names then?

rohan-bes (approved):
Nice one @billli0! This is gonna be super useful

Thanks so much for the tests as well, always makes me happy to see 😄

PR #4073, opened 2022-08-08
RN-588: Part 2, support fetching events directly against Tracked Entity Instances

rohan-bes:
### Issue RN-588:

As far as I can tell there's nothing support in the [DHIS2 api](https://docs.dhis2.org/en/develop/using-the-api/dhis-core-version-master/analytics.html#webapi_event_analytics) that allows pulling against tracked entity instances. Since `tei` isn't a valid dimenion (I tried), we have to pull all events for the parent org unit, and then filter out the sibling data.

Doing this in the `EventsPuller` seemed clean enough to me.

Behaviour:
```
fetchEvents({ organisationUnitCodes: ['someTEI'] }) => [{ orgUnitCode: 'parentOfTEI', teiCode: 'someTEI' }]
```
We keep the organisationUnitCode of the responding events as the parent organisationUnitCode. This may be a tad confusing, but I feel helps make it explicit that we're dealing with tracked entity instances here which DHIS2 treats rather differently


packages/data-broker/src/__tests__/services/dhis/builders/buildEventsFromDhisEventAnalytics.test.js
```diff
@@ -82,7 +82,7 @@ describe('buildEventsFromDhisEventAnalytics()', () => {
         orgUnitName: 'Haveluloto',
         eventDate: '2020-02-07T14:33:00.000',
         dataValues: {},
-        trackedEntityId: 'tracked_entity_id_dl_household_2',
+        trackedEntityId: 'tracked_entity_id_dl_household_3',
```
rohan-bes, line 85: Since I added `tracked_entity_id_dl_household_2` as a real TEI in the mock data, I had to change this test a little

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
```diff
@@ -21,14 +18,6 @@ describe('EventsPuller', () => {
...
     dhisApi = createMockDhisApi();
     stubGetDhisApi(dhisApi);
-    buildEventsFromDhisEventAnalytics.mockReturnValue('X');
-  });
-
-  it('pulls', async () => {
-    const result = await eventsPuller.pull([dhisApi], [DATA_GROUPS.POP01_GROUP], {
-      dataElementCodes: ['POP_01'],
-    });
-    expect(result).toEqual(['X']); // not sure why but mocked return value is wrapped into an array
```
rohan-bes, line 31: Decided to scrap this test, as the tests below actually test pulling data with real results, so seems a bit redundant

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
```diff
@@ -49,20 +38,26 @@ describe('EventsPuller', () => {
     it('correctly invokes the event analytics api in DHIS', () =>
       assertEventAnalyticsApiWasInvokedCorrectly({
         dataSources: [DATA_GROUPS.POP01_GROUP],
+        options: {
+          organisationUnitCodes: ['DL_FACILITY_A'],
+        },
```
rohan-bes, line 43: I guess we did technically support not passing in `organisationUnitCodes` in the options previously. I believe DhisApi throws an error in the real code if this is the case, so I don't mind introducing this requirement in the DhisService level...

packages/data-broker/src/services/dhis/builders/buildEventsFromDhisEventAnalytics.js
```diff
@@ -61,7 +60,7 @@ const buildEvents = (dhisEventAnalytics, dataElementCodes = []) => {
     });
     events.push(event);
   });
-  return events.sort(getSortByKey(METADATA_DIMENSION_TRANSLATION.eventdate));
```
rohan-bes, line 64: Moved this sort to the higher level, which seems fine to me given this function is resposible for 'building' events (not necessarily sorting them) 👍

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
```diff
@@ -106,20 +102,25 @@ describe('EventsPuller', () => {
...
-        expect(buildEventsFromDhisEventAnalytics).toHaveBeenCalledOnceWith(
-          expect.anything(),
-          getEventAnalyticsResponse,
-          dataElementCodes,
-        );
+
+        expect(events).toEqual([
+          { dataValues: { POP01: 5, POP02: 8 } },
+          { dataValues: { POP01: 4, POP02: 3 } },
+        ]);
```
rohan-bes, line 123: Reworked these tests to actually return real data. I always prefer that over testing spy implementations where possible 👍

chris-pollard:
@rohan-bes could this also be used in the example of viz builder fetching data groups that are not programs but are tracked in dhis?

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
IgorNadj: I think the idea for this was a happy path test which is a handy way of quickly understanding a class by looking at the first test, but happy to consolidate too. We won't really need it once we have types.

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
IgorNadj: Ah cool, should we add a test to make sure it throws?

packages/data-broker/src/services/dhis/pullers/EventsPuller.js
```diff
@@ -28,17 +29,56 @@ export class EventsPuller {
...
 
-    return buildEventsFromDhisEventAnalytics(
-      this.models,
-      translatedEventAnalytics,
-      dataElementCodes,
-    );
+    return buildEventsFromDhisEventAnalytics(this.models, orgUnitEventAnalytics, dataElementCodes);
+  };
+
+  pullEventsForApi = async (api, programCode, options) => {
```
IgorNadj, line 40: Not related to your PR but could we add `@private` tags to these methods, will help Kostas' PR when adding types

packages/data-broker/src/services/dhis/pullers/EventsPuller.js
```diff
@@ -28,17 +29,56 @@ export class EventsPuller {
...
+    const { organisationUnitCodes: entityCodes, hierarchy } = options;
+
+    if (!entityCodes || entityCodes.length === 0) {
```
IgorNadj, line 43: Minor: move validation to pull()? And should we add validation and a test for hierarchy present?

packages/data-broker/src/services/dhis/pullers/EventsPuller.js
```diff
@@ -28,17 +29,56 @@ export class EventsPuller {
...
+
+    let trackedEntityEvents = [];
+    if (trackedEntities.length > 0) {
+      // Events for tracked entities cannot be fetched directly in DHIS2, instead we must fetch for the parent organisation unit and then
+      // we need to filter out the results for the siblings
+      const hierarchyId = (await this.models.entityHierarchy.findOne({ name: hierarchy })).id;
+      const parentsOfTrackedEntities = await Promise.all(
+        trackedEntities.map(trackedEntity => trackedEntity.getParent(hierarchyId)),
+      );
+      const parentEvents = await this.pullEventsForOrganisationUnits(api, programCode, {
```
IgorNadj, line 67: Nice, very clean!

rohan-bes:
> @rohan-bes could this also be used in the example of viz builder fetching data groups that are not programs but are tracked in dhis?

@chris-pollard I think we'd need to explore what's possible in the DHIS api around this. Each case is different honestly. We endeavour to wrap the DHIS api in such a way as to make it as simple as possible, but sometimes things aren't possible, or add an abstraction layer that is too inconsistent to use...

In this instance it felt like a good idea, but I can't say for pulling data groups that aren't programs...

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
IgorNadj: Ah yep this is the `buildEventsFromDhisEventAnalytics() invocation` describe block, that's why that's mocked out like that. It's not testing real data because it's more of a unit test with dependencies mocked out, so that changes to buildEventsFromDhisEventAnalytics don't fail this test, and so that buildEventsFromDhisEventAnalytics is not double-tested. 

I like proper full integration tests too, but my feeling is this area is so complicated we need to break things up if only for reducing cognitive load when working in this part of the codebase, so unit tests are great for this complexity management and processes like TDD, when you are working on a single class you can work with its tests alone and focus on what it alone does and not the half dozen or so classes it's connected to.

We do definitely need proper integration tests as well though, especially for DhisService, because it's so complicated, the rest are probably fine. Something like DhisService.integration.test.js would be a good place, following what we do in the indicators package with src/__tests__/__integration/... Otherwise for something simpler maybe a new describe block at the top of this file called something like 'integration'?

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
```diff
@@ -133,30 +134,76 @@ describe('EventsPuller', () => {
...
-        const translatedEventAnalyticsResponse = {
-          headers: [{ name: 'DIF01', column: 'Different 1', valueType: 'NUMBER' }],
-          metaData: {
-            items: {
-              DIF01: { name: 'Different 1' },
-            },
-            dimensions: { DIF01: [] },
-          },
-          rows: [],
+          rows: [[7], [4]],
```
IgorNadj, line 137: Same as above

IgorNadj (commented):
Nice one! Good to have this and so glad we went down the transparent-to-the-user path, very clean. One comment around test organisation otherwise LGTM!

packages/data-broker/src/__tests__/services/dhis/DhisService.fixtures.js
```diff
@@ -25,20 +25,42 @@ export const DATA_GROUPS = {
   }),
 };
 
+const DL_FACILITY_A = {
+  code: 'DL_FACILITY_A',
+  name: 'DL FACILITY A',
+  type: 'facility',
+  metadata: {},
+  isTrackedEntity: () => false,
```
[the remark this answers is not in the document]
rohan-bes: Mocking out `getParent` is still a bit more convenient since the real implementation relies on a whole lot of set up. But good point, I hadn't even thought of replacing these will real the `EntityType`. Will play with it and see how it looks 👍

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
rohan-bes: Ah good idea, will do 👍

packages/data-broker/src/services/dhis/pullers/EventsPuller.js
rohan-bes: Good idea, will add 👍

packages/data-broker/src/services/dhis/pullers/EventsPuller.js
rohan-bes: Yes and yes 👍

packages/data-broker/src/__tests__/services/dhis/DhisService.fixtures.js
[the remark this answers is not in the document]
IgorNadj: One thing I've noticed is we needlessly mock out the database. We could just insert the mock data and let it use the real db. But that's for a future PR. Happy with what you have if it works for now :)

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
rohan-bes: Ah I see your point. And yup, my bad for changing these without taking a look at how we've structured tests in this package.

I'm thinking of creating a separate describe block within this `EventsPuller` test file. I guess the rationale for doing so would be that ideally the logic around handing tracked entity instances is hidden within `EventsPuller`. I wouldn't necessarily want the tests to need to go in an integration test suite as it's a bit of an implementation detail. 

I'll restore the old tests to their previous behaviour 👍 

Also, fwiw I've never been too concerned about double testing areas of code. It's always felt more effort than its worth to maintain trying to just have a single test to cover each code path? Thoughts?

IgorNadj: Yeah I was thinking about where best to place those, a block in EventsPuller makes sense because it's pretty specific to that. The other options of DhisService is nice for being able to see all possible input/outputs of DhisService at a glance, but we typically only care about analytics or events not both at the same time. Then there's the option of top level / DataBroker level, which is also nice, for seeing and testing all services, would be cool being able to see which services support what, and it would be a real integration test going through all layers, but again we're not jumping between them very often so it's not super beneficial day to day. I like your idea of placement.

Double testing, yeah it's a judgement call. I'm pretty sure none of our unit tests are actually pure unit tests, there's always something we don't mock out. But the value I see of unit tests and TDD in general is that when a test is nice and neat and tests one thing it means your class/function must be neat and only do that one thing. So it kind of forces separation of concerns and clean code. You can have a function X that does two things and a test Y that tests that it does both things, but you can't have a test Y that tests one thing and a function X that does two things (if that makes sense).

packages/data-broker/src/services/dhis/pullers/EventsPuller.js
rohan-bes: Okay, I double checked it this and it turns out that the deprecated event pull api in DHIS does support no org unit codes. Since the `DeprecatedEventsPuller` extends this class, it causes issues if I move the validation to `pull`, so keeping it here.

packages/data-broker/src/__tests__/services/dhis/DhisService.fixtures.js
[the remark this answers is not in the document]
rohan-bes: I tried with using the `EntityType`, but since we're not actually using a DB connection it could be a tad confusing (since the DatabaseType requires a model, which we'd have to stub out). Thinking of going with how it was originally if that's cool?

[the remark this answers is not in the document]
IgorNadj: Sure 👍

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
rohan-bes: Yeah definitely agree that a high level place where we have a suite of tests that show the basic api (input -> output) would be really nice. This is a notoriously opaque interface imo.

But I'm always a fan of having tests that get into the details being around the code that handles those details.

>You can have a function X that does two things and a test Y that tests that it does both things, but you can't have a test Y that tests one thing and a function X that does two things (if that makes sense). I tend to think of it in those terms, rather than avoiding two tests overlapping.

That's a really good way of describing it! Thanks for sharing that perspective!

I suspect I lean towards the idea that when logic is composed hierarchically (eg. EventsPuller calls EventsTranslator, and EventsBuilder), that it seems reasonable that it should test both of those during its tests. Not explicitly, but just as a natural side effect of writing simple tests that are just aware of the interface they're testing (and not the underlying behaviour). 
Of course mocking is required at the service boundary layer (database, external package/api) but other than that I prefer to keep it to a minimum. I have a bit of an anti-mocking bias tho, which admittedly comes from some valid places but also just a bit of discomfort with the mechanics of it which isn't really a valid concern 😅

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
```diff
@@ -86,13 +95,107 @@ describe('EventsPuller', () => {
...
       assertEventAnalyticsApiWasInvokedCorrectly({
         dataSources: [DATA_GROUPS.POP01_GROUP],
-        options: { dataElementCodes: ['POP01', 'DIF01'] },
+        options: { dataElementCodes: ['POP01', 'DIF01'], organisationUnitCodes: ['DL_FACILITY_A'] },
         invocationArgs: expect.objectContaining({ dataElementCodes: ['POP01', 'DIF01_DHIS'] }),
       }));
   });
 
   describe('data building', () => {
+    describe('pulling events', () => {
```
IgorNadj, line 104: Minor: I think `data building` means just the building part of it rather than pulling as well, maybe pop `pulling events` outside of this?

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
```diff
@@ -86,13 +95,107 @@ describe('EventsPuller', () => {
...
+          organisationUnitCodes,
+        });
+
+        expect(events).toEqual([{ dataValues: { POP01: 5 } }, { dataValues: { POP01: 4 } }]);
+      });
+
+      it('can pull data for just the requested tracked entity instance', async () => {
+        // Return data for all children of the parent org unit
+        const getEventAnalyticsStub = ({ organisationUnitCodes }) => {
+          if (organisationUnitCodes[0] !== 'DL_FACILITY_A' || organisationUnitCodes.length !== 1) {
```
IgorNadj, line 133: I think I get it but not 100%, is this because its called twice, once for org units and the second time for tracked entities? If so can you add a comment

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
```diff
@@ -86,13 +95,107 @@ describe('EventsPuller', () => {
...
+          organisationUnitCodes,
+          hierarchy,
+        });
+
+        expect(events).toEqual([
+          {
+            dataValues: { DIF01: 7 },
+            trackedEntityCode: 'DL_HOUSEHOLD_1',
+            trackedEntityId: 'tracked_entity_id_dl_household_1',
+          },
```
IgorNadj, line 183: Is this response supposed to have orgUnitCode as well? Just from the PR description

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
rohan-bes: Honestly it's actually not needed, since it does only get called once. I just sort of wanted to make it a bit explicit that it's returning data for the parent (`DL_FACILITY_A`) even though we're querying for the child (`DL_HOUSEHOLD_1`). But it's a bit confusing fair point, I'll add a comment to make it more clear 👍

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
rohan-bes: The response depends on what comes from the `DhisApi`. Since we're mocking that to just return the value for `DIF01` and the tracked entity id, that what the response has. But in reality it'll come back with the org unit code as well (which would be the parent org unit's code). 

I'll add it to the test as it'll make it explicit that the tracked entity code and the org unit code don't need to match 👍

IgorNadj (approved):
Only minor stuff left, LGTM!

packages/data-broker/src/services/dhis/builders/buildEventsFromDhisEventAnalytics.js
```diff
@@ -61,7 +60,7 @@ const buildEvents = (dhisEventAnalytics, dataElementCodes = []) => {
...
+  return events;
```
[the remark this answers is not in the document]
rohan-bes: Ah great catch! I suppose we can remove that part of the name as it's no longer in the contract?

But I'll add a test to ensure events are sorted in the `EventPuller` tests as it's now part of its contract!

packages/data-broker/src/__tests__/services/dhis/pullers/EventsPuller.test.js
IgorNadj: Mocking in javascript is so painful 😢 It really is. I understand the bias completely 😆 

>writing simple tests that are just aware of the interface they're testing (and not the underlying behaviour).

Yes! This is something javascript doesn't do well, there's no inversion of control / dependency injection which would let us do this. I think having our tests/mocking know about the implementation is just a pain point we have to bear :/ It seems to be the way jest and them recommend so we're not alone.

>I suspect I lean towards the idea that when logic is composed hierarchically (eg. EventsPuller calls EventsTranslator, and EventsBuilder), that it seems reasonable that it should test both of those during its tests.

I agree for sure. With this one you definitely benefit from integration testing the parent with its children, you make sure these three parts are working well together, and like you said above, you get simplicity and not having to do this weird mocking-based-on-implementation-knowledge thing.

The downside I see is when it gets too complex to do this. You are super familiar about this codebase and a really good dev so it's no biggie for you keeping all this in your head :) but ask a junior to make a change in data-broker and their head would explode... I mean I've spent two years on this and my head is still exploding... Anyway treating the children and parent as one things just means the junior would need to understand everything at the same time, i.e. the full implementation to make a change. The TDD way would still ask a lot from them, but once they know just the point of each thing, or the single responsibility of each class, skim through the tests maybe, they can then more easily figure out where and how to make a change.

So yes I agree with you, except in this one specific part of the codebase where my head hurts thinking about everything at the same time :)

PR #4076, opened 2022-08-08
Rn 494 tupaia dashboard export in pdf part one

billli0:
### Issue #:
RN-494 tupaia dashboard exports - part 1

### Changes:
- Add a separated page for PDF exporting, example URL: `http://localhost:8088/covid_samoa/WS/COVID-19%20Samoa%20-%20Total/pdf-export`
- To achieve that, use `react-router-dom` to handle a new param `pdfExportType` in URL `/:projectCode/:countryCode/:dashboardCode/:pdfExportType`.
-- `react-router-dom` gives us ability to add a new route out of tupaia main page, and still use `sagas.js` existing fetch functions compatibly.
- The value of `:/pdfExportType` currently is `pdf-export` for dashboard items pdf export, in the future we will add a new value to support survey response pdf export with different export format.


packages/web-frontend/src/components/View/ChartWrapper.js
```diff
@@ -94,7 +100,13 @@ export const TABS = {
   TABLE: 'table',
 };
 
-export const ChartWrapper = ({ viewContent, isEnlarged, isExporting, onItemClick }) => {
+export const ChartWrapper = ({
+  viewContent,
+  isEnlarged,
+  isExporting,
```
IgorNadj, line 106: I think this needs a rework. isExporting made sense when it was just .png, now that there are two options we should rename these for clarity.

Maybe something like exportFormat = 'png' | 'pdf' | undefined. isExporting is defined on ui-components/src/components/Chart so we should make the change there.

packages/web-frontend/src/components/View/utils.js
```diff
@@ -61,6 +61,22 @@ export function getViewWrapper({ type, viewType }) {
...
+export function getExportViewWrapper({ type, viewType }) {
+  switch (type) {
+    case 'chart':
+      return ChartWrapper;
+    default: {
+      return () => (
+        <div>
+          <PDFExportViewTitle>{`'${
+            type || viewType
+          }' type visual does not support PDF export`}</PDFExportViewTitle>
```
IgorNadj, line 73: I'm surprised that there are visuals that don't support PDF export, why is this?

packages/web-frontend/src/components/PDFExport/A4Page.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
IgorNadj, line 1: This seems to be the same as the one in lesmis, can we DRY it by moving it to ui-components?

IgorNadj (changes requested):
Started on this review, will finish later

packages/web-frontend/src/components/View/ChartWrapper.js
billli0: Yeah isPDFExporting is confusing.. We can use `exportFormat` in ChartWrapper, but I suggest we leave the logic here instead of in `Chart.js`, since they are only style component different.

packages/web-frontend/src/components/View/utils.js
billli0: Matrix is not exportable, text visuals need a light background to print nicely. Discussion: https://linear.app/bes/issue/RN-494#comment-f866f20a

IgorNadj: I see. Yep that's all good, can you add that as a comment? We could also add a const like NOT_EXPORTABLE_TYPES to make it explicit, it's hard to understand what type and viewType have to do with anything.

packages/web-frontend/src/components/View/ChartWrapper.js
IgorNadj: Ah so Chart renders the same if isExporting, but ChartWrapper renders differently for png and pdf? If so, your suggestion makes sense 👍

packages/web-frontend/src/components/PDFExport/A4Page.js
billli0: Yeah good idea. I also feel `Header` in `A4page` is a bit weird. I've pull it out to address that.

IgorNadj (approved):
Really clean, love it! Nice work

PR #4081, opened 2022-08-09
RN-603 de dg multi mapping

IgorNadj:
### Changes:

- Add ability to map a single Data Element / Data Group to multiple data services based on country fetched for


packages/database/src/migrations/20220808000411-AddDataServiceMappingColumn-modifies-schema.js
```diff
@@ -0,0 +1,29 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`ALTER TABLE data_element ADD COLUMN data_service_mapping JSONB DEFAULT '[]'`);
```
rohan-bes, line 18: Not a major thing, and we are planning on removing this in the long run anyway, but I guess I had supposed this would be part of the `config` attribute? Given it's going to be used so rarely, seems it mightn't warrant a column of its own?

packages/data-broker/src/DataBroker.js
```diff
@@ -241,14 +249,70 @@ export class DataBroker {
...
+   */
+  async resolveServiceType(dataSources, orgUnitCode) {
+    const map = {};
+    const addToMap = (k, v) => {
+      if (!map[k]) map[k] = [];
+      map[k].push(v);
+    };
+
+    const entity = orgUnitCode ? await this.models.entity.findOne({ code: orgUnitCode }) : null;
+    const { countryCode } = entity ?? {};
```
rohan-bes, line 292: Do we want to throw an error here if we can't find an entity/countryCode? Feels a tad unsafe that we could follow through to to the below logic with a undefined countryCode.

Bit on the fence but just wanted to raise it

packages/data-broker/src/DataBroker.js
```diff
@@ -241,14 +249,70 @@ export class DataBroker {
...
+      if (!mappingForCountry) {
+        addToMap(dataSource.service_type, dataSource);
+        continue;
+      }
+
+      const { serviceType } = mappingForCountry;
+      if (!map[serviceType]) {
+        map[serviceType] = [];
+      }
+      addToMap(serviceType, dataSource);
```
rohan-bes, line 313:
```suggestion
      addToMap(serviceType, dataSource);
```
I think your nifty `addToService()` function handles the case if there's no entry for the service type 👍

rohan-bes (changes requested):
Really nice work man! Love how simple this was to implement 👍 

One general request, looks like with the implementation where this is a separate column, it can't be configured via the Admin Panel? I'd really love if that were possible (anything to avoid needing more migrations)

Thoughts on either moving the `dataServiceMapping` into the `config` field, or adding the appropriate boilerplate to allow configuring this as part of the Admin Panel?

Aside from that just some minor code style comments, but nothing major. Love the tests btw 👍

IgorNadj:
@rohan-bes Thanks for the review, I'm going to have to rework this to handle both country level and individual entities, will ping when ready again

IgorNadj:
Closing this, will re-create PR and link back so we don't lose feedback

PR #4083, opened 2022-08-09
Convert data broker to TS: update implentation

kael89:
### Issue #:

[No issue]

This is a follow up on the initial [Convert data-broker to TS](https://github.com/beyondessential/tupaia/pulls/kael89) PR. It _should_ fix any remaining issues and make the base branch mergeable, after conflicts are resolved.

**Note:** The build fails and I cannot see why as I have no access to Codeship... Happy to help resolve any issues but I will need some information from your end

### Changes:

* Changed a lot of test implementations to align them with the new types
* Change `data-broker` implementation in specific cases
* Slightly refactored `indicators` to avoid a circular dependency in the build
* Fixed/Improved a few types along the way - this could happen in the initial PR but it was easier for me to do it here

### Testing

I haven't tested my changes end-to-end since that would require me to run a big part of the stack locally. I suspect that this may be hard or even impossible for a Community contributor at this stage... Sounds like an interesting future task :-)

Although both `data-broker` and `indicators` have great line coverage (`82.4%` and `98.9%` respectively), to be extra safe the following could be tested manually:
* `data-broker`: Smoke test on general data pulling/pushing/deleting
* `indicators`: Smoke test on general indicator data pulling


packages/data-broker/src/services/weather/WeatherService.ts
```diff
@@ -71,6 +90,8 @@ export class WeatherService extends Service {
 
       return this.getHistoricWeather(entities, startDate, endDate, apiResultTranslator);
     }
+
+    return resultFormat === 'analytics' ? EMPTY_ANALYTICS_RESULTS : ([] as EventResults);
```
kael89, line 94: :exclamation: Impementation change

This is a breaking change with the purpose of using consistent types. The existing behaviour is returning `undefined` here, which will result in `[undefined]` being returned as the resulting analytics/events, which seems wrong.

I chose to return an empty response because this is what the other services do when no data source is found.

packages/indicators/src/types.ts
```diff
@@ -87,3 +87,5 @@ export type FetchOptions = Readonly<{
   readonly endDate: string;
   readonly period?: string;
 }>;
+
+export type DataBroker = unknown;
```
kael89, line 91: This type is enough here since `DataBroker` is just passed through from `indicators` to `aggregator`.

A more specific type can be imported from `aggregator` if that package gets migrated to TS

edmofro:
@rohan-bes @IgorNadj re-up on us needing to review and merge this. I know you're both very busy, but let's discuss in our next tech strategy mtg

PR #4084, opened 2022-08-10
RN-558: Specify DHIS Server Explicit

chris-pollard:
### Changes:
- Add migration to update null values to country server for dhisInstanceCode in data_element and data_group tables.

### DIFF
--- a/packages/database/src/migrations/20220810013153-UpdateAllNonRegionalDhisInstanceConfigDataElementsAndGroups-modifies-data.js
+++ b/packages/database/src/migrations/20220810013153-UpdateAllNonRegionalDhisInstanceConfigDataElementsAndGroups-modifies-data.js
@@ -44,13 +44,13 @@ exports.up = async function (db) {
     await db.runSql(`
     UPDATE "data_group"
     SET "config" = jsonb_set("config", '${DHIS_INSTANCE_CODE_JSONPATH}', '"${server}"')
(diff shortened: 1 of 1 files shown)


packages/database/src/migrations/20220810013153-UpdateAllNonRegionalDhisInstanceConfigDataElementsAndGroups-modifies-data.js
```diff
@@ -0,0 +1,76 @@
...
+
+const DHIS_INSTANCE_CODE_JSONPATH = '{dhisInstanceCode}';
+const prefixListKeyValuePairs = Object.entries(PREFIX_TO_SERVER_LIST);
+
+exports.up = async function (db) {
+  for (const [prefix, server] of prefixListKeyValuePairs) {
+    await db.runSql(`
+    UPDATE "data_group"
+    SET "config" = jsonb_set("config", '${DHIS_INSTANCE_CODE_JSONPATH}', '"${server}"')
+    WHERE config::text like '%"dhisInstanceCode": null%' AND substring(code from 1 for 2) = '${prefix}';
```
IgorNadj, line 47: `WHERE (config->'dhisInstanceCode') is not null and (config->>'dhisInstanceCode') is null`

First part makes sure the key exists, second part makes sure the json value is null

IgorNadj (approved):
One change for sql but otherwise looks good, thanks!

PR #4087, opened 2022-08-15
RN-611 pdf export server to library

billli0:
### Issue #:
RN-611

### Changes:
See commits 

Manual changes: 
- [ ] Update lambda scripts
- [ ] Remove pdf-export-api in route53 for master and dev 
- [ ] Remove lastpass pdf-export-server .env


packages/tsutils/src/PDFExporter.ts
```diff
@@ -0,0 +1,63 @@
...
+    name,
+    domain: apiDomain,
+    url: location.origin,
+    httpOnly: true,
+    value: cookies[name],
+  }));
+  return { pdfPageUrl, cookies: finalisedCookieObjects };
+};
+
+export const exportToPDF = async (req: any) => {
```
rohan-bes, line 40:
```suggestion
/**
 * @param pageUrl the url to visit and download as a pdf
 * @param userCookie the user's cookie to bypass auth, and ensure page renders under the correct user context
 * @returns pdf buffer
 */
export const downloadPageAsPdf = async (pageUrl: string, userCookie = '') => {
```

Might be nice to make the behaviour of this function a bit more clear, by just taking in the required parameters rather than a `req` object? Also makes it a bit more flexible in use 👍

packages/tsutils/src/PDFExporter.ts
```diff
@@ -0,0 +1,63 @@
...
+    });
+  } catch (e) {
+    throw new Error(`puppeteer error: ${(e as Error).message}`);
+  } finally {
+    if (browser) {
+      await browser.close();
+    }
+  }
+
+  return { data: buffer };
```
rohan-bes, line 62:
```suggestion
  return buffer;
```
Any reason for returning it wrapped in an object?

rohan-bes (changes requested):
Thanks so much for doing this @billli0. It was exciting to have a new server, but this is a lot slimmer and simpler to manage. Who knows, we may end up introducing a server to handle this functionality in time if it's needed.

Just a couple small changes to the `exportToPdf` function, but aside from that looks great! Love all that red 😄

rohan-bes:
Oooh one more thing, would love to see some unit tests for this!

packages/tsutils/src/PDFExporter.ts
billli0: Not a strong reason to wrap it, happy to unwrap.

billli0:
> Oooh one more thing, would love to see some unit tests for this!

Yeah that would be nice. I will do it after merge so that we can test it with tupaia pdf export.

PR #4015, opened 2022-07-04
Rn-423 import kobo survey

IgorNadj:
RN-423

### Changes:

- Adds Sync Groups tab to admin panel with CRUD

---

### Screenshots:
https://linear.app/bes/issue/RN-423#comment-e30c8c8e

PR #4088, opened 2022-08-15
RN-423: Add 'Sync Groups' to the Admin Panel (kobo sync link part 1)

rohan-bes:
### Issue RN-423:

In this PR, we've added 'Sync Groups' to the Admin Panel. This is where the user can configure the mapping between Tupaia<->Kobo in the `Config` section.

The mapping is just a JSON configuration which we accept for now is good enough for the user.

Subsequent PRs will add the functionality to view the Sync Group logs, and perform manual syncs via the Admin Panel, which should help the end user test out configuring a new Sync Group.

All credit to @IgorNadj on this one, I'm just ferrying it over.


packages/central-server/src/apiV2/syncGroups/DeleteSyncGroups.js
```diff
@@ -0,0 +1,17 @@
...
+import { DeleteHandler } from '../DeleteHandler';
+import { assertAnyPermissions, assertBESAdminAccess } from '../../permissions';
+import { assertSyncGroupEditPermissions } from './assertSyncGroupPermissions';
+
+export class DeleteSyncGroups extends DeleteHandler {
+  async assertUserHasAccess() {
+    const syncGroupChecker = accessPolicy =>
+      assertSyncGroupEditPermissions(accessPolicy, this.models, this.recordId);
+
+    await this.assertPermissions(assertAnyPermissions([assertBESAdminAccess, syncGroupChecker]));
```
rohan-bes, line 15: We may need to consider how to change this permission check... I suspect there will be some lesmis users who want to configure the sync group but don't have BES Admin...

IgorNadj (approved):
Reviewed by Rohan in #4015, I had a look as well still looks good

packages/database/src/migrations/20220630013151-LinkSyncServiceToSurvey-modifies-schema.js
```diff
@@ -0,0 +1,45 @@
...
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(
+    `DROP TRIGGER IF EXISTS data_service_sync_group_trigger ON data_service_sync_group`,
+  );
+
+  await db.runSql(`ALTER TABLE data_service_sync_group RENAME COLUMN code TO data_group_code`);
```
IgorNadj, line 22: Actually I don't understand how this works. We can't change code here without updating DataBroker.

https://github.com/beyondessential/tupaia/blob/dev/packages/data-broker/src/DataBroker.js#L82

IgorNadj (changes requested):

packages/database/src/migrations/20220630013151-LinkSyncServiceToSurvey-modifies-schema.js
rohan-bes: Yeah honestly I found this rename a bit confusing as well.
It seems like a dssg has a `data_group_code` which corresponds to a survey code in Tupaia. But should the `syncGroupCode` in the `config` actually just be the `code` of this dssg? I made that mistake a couple times when writing logic around this, as it seems really weird to me that dssg's don't have a code of their own atm...

I'd suggest changing so that:
```
current:
id  |  data_group_code | service_type |         config             | sync_cursor
-----------------------------------------------------------------------------------
123 |    LEMIS_FQS1    |      kobo    | { syncGroupCode: FQS1, ...} | 10-10-2020

becomes:
id  | code |  data_group_code | service_type | config | sync_cursor
------------------------------------------------------------------------
123 | FQS1 |     LEMIS_FQS1   |      kobo    | { ...} | 10-10-2020

```
Then when pulling from data-broker, you do:
```
const koboData = await dataBroker.pull(
  {
    code: 'FQS1',
    type: 'SYNC_GROUP',
  }
);
```
Thoughts?

You could possibly argue that the `koboSurveyCode` (which exists in the `config`) should also have its own column, but I think given that this table is meant to be agnostic in regards to the `service_type` its appropriate for it to be in the config.

rohan-bes: Also, just wanted to flag a slight concern I have around how often we rely on pseudo foriegn key relationships. Here the `data_group_code` _**must**_ correspond to a survey code. But we don't enforce that in any way via foreign key constraints. It's quiet common for us to do this in Tupaia, but it always feels a bit risky to me... Do you think I ought to add a constraint here?

IgorNadj: Discussed, like the code column, FK not necessary but worth a chat in the future.

packages/database/src/migrations/20220630013151-LinkSyncServiceToSurvey-modifies-schema.js
```diff
@@ -0,0 +1,53 @@
...
+  // Add new code column which should match current syncGroupCode
+  await db.runSql(`ALTER TABLE data_service_sync_group ADD COLUMN data_group_code text`);
+
+  await db.runSql(`
+    UPDATE data_service_sync_group 
+    SET 
+      data_group_code = trim(both '"' from config->>'internalSurveyCode'),
+      code = trim(both '"' from config->>'internalSurveyCode'),
+      config = (config || jsonb_build_object('syncGroupCode', code)) - 'internalSurveyCode'
+  `);
```
rohan-bes, line 31: @IgorNadj reworked this as discussed.

Note: the kobo sync logic won't actually work right now (due to a mismatch between the code in the `sync_group` table's `config` column and the current values) however that will be fixed up in the [pt2 PR](https://github.com/beyondessential/tupaia/pull/4089)

IgorNadj (approved):
LGTM!

PR #4090, opened 2022-08-16
RN-484: Add ability for defining 'custom reports' in the report-server 

rohan-bes:
### Issue RN-484:

There often come times where we need to rush out a viz, and the demands to get it to work are too tricky for the viz builder at the moment.

The 'Custom Report' is an option to just handcraft a single JS function that pulls and transforms all the data we need. I prefer this option over using the legacy databuilders for a few reason:
1. We will probably always need some capacity for this, so it's worthwhile to manage it in a non-legacy way
2. The legacy pattern really blurs the line between what should be config, and what should be code. This however is very clear. Either use the viz-builder transform configs, or just code. I think that makes it simpler to understand and reverse engineer these custom reports when we need to
3. They're in typescript which makes them a lot easier to work with, and the have access to the `@tupaia/api-client`

But happy to hear push back from the reviewer if they feel this really muddies the report-server code?

Thoughts on the name 'Custom Reports' btw? It feel a bit weak and non-descriptive to me... HardCodedReport? idk...

### DIFF
--- a/packages/admin-panel-server/src/viz-builder/dashboardVisualisation/combineDashboardVisualisation.ts
+++ b/packages/admin-panel-server/src/viz-builder/dashboardVisualisation/combineDashboardVisualisation.ts
@@ -3,11 +3,15 @@
  * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
  */
 
-import { LegacyReport, Report} from '../types';
+import { LegacyReport, Report } from '../types';
 import { DashboardItem, DashboardViz, DashboardVizResource } from './types';
 
 const getData = (report: Report) => {
   const { config } = report;
+  if ('customReport' in config) {
+    return { customReport: config.customReport };
+  }
+
   const { fetch, transform } = config;
   const { aggregations, ...restOfFetch } = fetch;
   return { fetch: restOfFetch, aggregate: aggregations, transform };
@@ -24,14 +28,16 @@ const getPresentation = (dashboardItem: DashboardItem, report: Report | LegacyRe
   const { type, name, ...config } = dashboardItemConfig;
 
   const presentation: Record<string, unknown> = { type, ...config };
-  if (!dashboardItem.legacy) {
+  if (!dashboardItem.legacy && 'output' in reportConfig) {
     presentation.output = reportConfig.output;
   }
 
   return presentation;
 };
 
-export function combineDashboardVisualisation(visualisationResource: DashboardVizResource): DashboardViz {
+export function combineDashboardVisualisation(
+  visualisationResource: DashboardVizResource,
+): DashboardViz {
   const { dashboardItem, report } = visualisationResource;
   const { id, code, config, legacy } = dashboardItem;
   const { name } = config;
--- a/packages/admin-panel-server/src/viz-builder/dashboardVisualisation/combineDashboardVisualisation.ts
+++ b/packages/admin-panel-server/src/viz-builder/dashboardVisualisation/combineDashboardVisualisation.ts
@@ -4,18 +4,8 @@
  */
 
 import { LegacyReport, Report } from '../types';
+import { extractDataFromReport } from '../utils';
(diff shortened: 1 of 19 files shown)


packages/admin-panel-server/src/viz-builder/dashboardVisualisation/combineDashboardVisualisation.ts
```diff
@@ -24,20 +18,22 @@ const getPresentation = (dashboardItem: DashboardItem, report: Report | LegacyRe
   const { type, name, ...config } = dashboardItemConfig;
 
   const presentation: Record<string, unknown> = { type, ...config };
-  if (!dashboardItem.legacy) {
+  if (!dashboardItem.legacy && 'output' in reportConfig) {
```
rohan-bes, line 21: Having typescript in this package was an absolute godsend! It would have been sooo bug prone to try and make such a change without type safety 😄

packages/admin-panel-server/src/viz-builder/utils.ts
```diff
@@ -0,0 +1,18 @@
...
+export const extractDataFromReport = (report: Report) => {
+  const { config } = report;
+  if ('customReport' in config) {
+    return { customReport: config.customReport };
+  }
+
+  const { fetch, transform } = config;
+  const { aggregations, ...restOfFetch } = fetch;
+  return { fetch: restOfFetch, aggregate: aggregations, transform };
+};
```
rohan-bes, line 18: Refactored this out as it was common between mapOverlay and dashboardItem logic

packages/admin-panel/src/VizBuilderApp/components/Panel.js
```diff
@@ -76,6 +76,17 @@ const PanelTabPanel = styled.div`
...
+const CustomReportName = styled.span`
+  flex: 1;
+  display: flex;
+  flex-direction: column;
+
+  > div {
+    width: 100%;
+    height: 100%;
+  }
+`;
```
rohan-bes, line 88: I am no css craftsperson 😢 You can see the ticket for an example. But given this should be for internal use only I don't mind it if it looks a bit bad

packages/report-server/src/reportBuilder/customReports/testCustomReport.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const testCustomReport = async (reqContext: ReqContext, query: FetchReportQuery) => {
+  const { organisationUnitCodes: entityCodes, hierarchy } = query;
+  const facilities = await reqContext.services.entity.getDescendantsOfEntities(
+    hierarchy,
+    entityCodes,
+    { filter: { type: 'facility' } },
+  );
+  return [{ value: facilities.length }];
+};
```
rohan-bes, line 17: Added this `testCustomReport` as an example for future use and a bit of a proof of concept. @chris-bes feel free to remove this once you add your actual implementation 👍

billli0: @chris-pollard

billli0 (approved):
Thanks @rohan-bes for adding this custom report function. Looks great!

packages/admin-panel/src/VizBuilderApp/components/Panel.js
billli0: Replace `CustomReportName` to div will do the work, don't need to attach any css.

packages/report-server/src/reportBuilder/customReports/testCustomReport.ts
chris-pollard: Thanks @rohan-bes

packages/admin-panel/src/VizBuilderApp/components/Panel.js
rohan-bes: Ah thanks @billli0 !

PR #4106, opened 2022-08-23
Hotfix: Not all surveys showing up in meditrak-app

rohan-bes:
### Issue https://beyondessential.slack.com/archives/C01MAA3NKM1/p1661205934386229:

Meditrak treats a survey with no countryIds as being available in all countries. Previously we were not syncing those surveys, so they did not show up in the app.

Personally I find the convention of no countryIds representing all countryIds a bit confusing... I won't change it as part of this PR but I'd prefer if we switched to a wild-card of sorts

### DIFF
--- a/packages/central-server/src/apiV2/utilities/meditrakSync/permissionsBasedMeditrakSyncQuery.js
+++ b/packages/central-server/src/apiV2/utilities/meditrakSync/permissionsBasedMeditrakSyncQuery.js
@@ -17,12 +17,11 @@ import {
   supportsPermissionsBasedSync,
 } from './supportsPermissionsBasedSync';
 
+const recordTypesToAlwaysSync = ['country', 'permission_group'];
 /**
  * Since all countries, permission_groups, and country entities regardless of permissions
  */
 export const permissionsFreeChanges = since => {
-  const recordTypesToAlwaysSync = ['country', 'permission_group'];
-
   return {
     query: `change_time > ? AND (record_type IN ${SqlQuery.record(
       recordTypesToAlwaysSync,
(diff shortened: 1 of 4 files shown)


packages/central-server/src/apiV2/utilities/meditrakSync/permissionsBasedMeditrakSyncQuery.js
```diff
@@ -32,14 +31,20 @@ export const permissionsFreeChanges = since => {
 };
 
 export const changesWithPermissions = (countryIds, permissionGroups, since) => {
-  let query = 'change_time > ? AND "type" = ?';
-  const params = [since, 'update'];
+  let query = `change_time > ? AND "type" = ? and record_type NOT IN ${SqlQuery.record(
+    recordTypesToAlwaysSync,
+  )}`;
```
rohan-bes, line 36: This `record_type NOT IN recordTypesToAlwaysSync` clause would have been valid previously as well honestly. It's just that with the new changes to always sync when the `countryIds` is null, we keep re-syncing permission groups that we've already synced

IgorNadj (approved):

packages/central-server/src/apiV2/utilities/meditrakSync/permissionsBasedMeditrakSyncQuery.js
```diff
@@ -89,70 +94,63 @@ export const buildPermissionsBasedMeditrakSyncQuery = async (
...
-  if (unsynced.countries) {
-    // Never before synced countries due to new country permissions
+  const addPermissionsOrClause = clause => {
     query.append(` 
-    OR (`);
-    query.append(changesWithPermissions(unsynced.countryIds, null, 0));
+     OR (`);
+    query.append(clause);
     query.append(')');
+  };
```
rohan-bes, line 108: Just made things a bit more DRY with this

packages/central-server/src/tests/apiV2/changes/getChangesMetadata.test.js
```diff
@@ -140,7 +140,7 @@ describe('GET /changes/metadata', async () => {
         });
 
         const { changeCount, countries } = response.body;
-        expect(changeCount).to.equal(8);
+        expect(changeCount).to.equal(10);
```
rohan-bes, line 143: Strictly speaking we don't need to sync down the 2 new records here. Since we've previously synced against any country, we should know we've synced them. But I didn't really wanna go into the depths of that logic as it is kinda complicated, and there's no real harm in a little 'over syncing'.

Ideally we switch from this 'NULL countryIds means all countries' logic to a proper wild card and then it should be safer to clean such things up 👍

PR #4109, opened 2022-08-24
add validation for codegenerator question

avaek:
Solves [RN-620 ](https://linear.app/bes/issue/RN-620/prevent-users-from-importing-a-codegenerator-question-without-a-config) (which was a bug caused by there not being validation ensuring CodeGenerator questions have content in the config column)


packages/central-server/src/apiV2/import/importSurveys/constructQuestionValidators.js
```diff
@@ -232,4 +232,14 @@ export const constructQuestionValidators = models => ({
       return true;
     },
   ],
+  config: [
+    (cell, row) => {
+      if (row.type === 'CodeGenerator' && isEmpty(cell)) {
```
billli0, line 237:
```suggestion
      if (row.type === ANSWER_TYPES.CODE_GENERATOR && isEmpty(cell)) {
```

Prefer to use constant variable.

billli0 (approved):
Looks good! Thanks for adding a validation. It is up to you, but I prefer to use constant variable as suggested.

packages/central-server/src/apiV2/import/importSurveys/constructQuestionValidators.js
avaek: Thanks for the suggestion, will do!

PR #4112, opened 2022-08-24
hotfix: typescript validation error

billli0:
No CDN for lesmis-server, we don't need to add `.cdn` to the host. This PR will simply revert previous changes.


avaek (approved):
Approving based on our Slack conversation

PR #4116, opened 2022-08-25
Minor renames in PDF export code

rohan-bes:
Switched to using consistent casing for PDF. Personally I've always slightly preferred using camel case for acronyms (eg. `PdfExportClass` or `downloadPdfFunction`) but it seems the prevailing convention in our code is to use consistent casing (so either all lower, or all upper). There's a friendly debate about the topic [here](https://stackoverflow.com/questions/2236807/java-naming-convention-with-acronyms) for curious folks

Also just renamed a file to match its export.


billli0 (approved):
Interesting discussion in stack overflow. `Dvd` just not right.. it looks like `dad`.

PR #4117, opened 2022-08-25
RN-619: Kobo Sync QOL improvements (part 1: display Sync Group logs in the Admin Panel)

rohan-bes:
### Issue RN-619:

Added a 'Logs' button and a modal popup that displays the latest sync logs for a Sync Group.

Screenshots available in the ticket.

Honestly, I don't love this code too much. It's really boilerplate heavy, since it uses redux. There's probably potential to switch to using react hooks and potentially react query, but I also wanted to maintain consistency with the existing pattern.

I also took a stab a consolidating things into a generic 'Modal' component, that was using by other modals (eg. Edit modals, Import/Export modals) but that quickly grew to be too complex an undertaking.

Still happy to hear pushback from reviewer if they feel this just clutters an already slightly too busy codebase...


packages/ui-components/src/components/Table/TableRow.js
```diff
@@ -46,11 +48,12 @@ export const TableRowCells = React.memo(({ columns, rowData, ExpandButton }) =>
...
       );
     },
-  ),
-);
+  );
+});
 
 TableRowCells.propTypes = {
   columns: PropTypes.arrayOf(PropTypes.shape(tableColumnShape)).isRequired,
+  rowProps: PropTypes.shape(PropTypes.shape(tableRowPropsShape)).isRequired,
```
rohan-bes, line 56: Didn't love having to do this, but there didn't seem to be another way to pass down properties to the actual row/cell level. The fact that the `rowProps` then need to be declared for all parent components was really quite a pain 😞 

Bit on the fence with this naming of `rowProps`. I wouldn't like `row`/`rows` as that would imply it's the actual data.

packages/admin-panel/src/pages/resources/ResourcePage.js
```diff
@@ -62,6 +63,7 @@ export const ResourcePage = ({
         onProcessDataForSave={onProcessDataForSave}
         displayUsedBy={displayUsedBy}
       />
+      <LogsModal />
```
rohan-bes, line 66: This is a bit concerning to me... are we just going to end up with an ever growing list of modals defined here? That's why I want to consolidate into a single Modal component and then have its contents be customizable (since there can only be a single modal open at once). But it really was too much to bite off right now...

billli0: Love the idea of consolidating them, but currently we have more than 5 modals, it is not an easy refactor.

packages/admin-panel/src/pages/resources/SyncGroupsPage.js
```diff
@@ -34,6 +38,17 @@ const FIELDS = [
 
 const COLUMNS = [
   ...FIELDS,
+  {
+    Header: 'Logs',
+    type: 'logs',
+    source: 'id',
+    actionConfig: {
+      title: '{code} sync group logs',
```
billli0, line 46: `{code}` is very cool. Didn't know that.

billli0 (approved):
Overall it looks good and clean! Yes and Redux is hard to review too, but sometimes I found it is useful that I can select state at a very deep hierarchy.
 
Got a suggestion on how we can modify the cell height, and a question of whether we can remove some of the mergeProps.

packages/database/src/modelClasses/DataServiceSyncGroup.js
```diff
@@ -25,6 +25,35 @@ class DataServiceSyncGroupType extends DatabaseType {
...
+      `
+         SELECT count(sgl.*) FROM sync_group_log sgl
+         JOIN data_service_sync_group dssg ON dssg.code = sgl.sync_group_code
+         WHERE dssg.id = ?
+       `,
+      [this.id],
+    );
+
+    return count;
+  }
```
billli0, line 40: Tested a bit and the endpoint returns string as count. Suspect count is a string in here? 
image

packages/ui-components/src/components/Table/TableRow.js
rohan-bes: Ah great call! Infact, I don't even need to add the `className` can just use `MuiTableCell-root`!

packages/database/src/modelClasses/DataServiceSyncGroup.js
rohan-bes: Ah great catch! I even encountered this same bug not long ago 🤦 will fix

packages/database/src/modelClasses/DataServiceSyncGroup.js
```diff
@@ -25,6 +25,35 @@ class DataServiceSyncGroupType extends DatabaseType {
...
+         SELECT sgl.* FROM sync_group_log sgl
+         JOIN data_service_sync_group dssg ON dssg.code = sgl.sync_group_code
+         WHERE dssg.id = ?
+         ORDER BY timestamp DESC
+         LIMIT ?
+         OFFSET ?
+       `,
+      [this.id, limit, offset],
+    );
+
```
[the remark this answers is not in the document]
rohan-bes: Yup 👍

PR #4118, opened 2022-08-25
RN-619: Kobo Sync QOL improvements (part 2: allow manual Sync Group syncing via Admin Panel)

rohan-bes:
### Issue RN-619:

Added a 'Sync' button to the Sync Groups tab of the Admin Panel. Clicking the button will trigger a manual sync for the Sync Group. The button also indicates the sync status (green and spinning for 'sync in progress', red for 'last sync attempt failed')

Screenshots available in the ticket.


packages/database/src/migrations/20220822035354-AddSyncStatusToSyncGroup-modifies-schema.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  return db.runSql(
+    `ALTER TABLE data_service_sync_group ADD COLUMN sync_status text DEFAULT 'IDLE'`,
```
billli0, line 19: Can we restrict the value to only be IDLE, SYNCING, ERROR?

rohan-bes: Fair point. I was thinking of this, but then remembered @mcccclean had mentioned something in dev-chat about pain points when working with enum types in postgres. I recall a time when you had to fully delete them and re-create the type if you wanted to add a value, but it seems postgres has supported adding types ad hoc since version 9...

I agree, enum type is nicer here, I'll do it 👍

packages/admin-panel/src/syncButton/SyncButton.js
```diff
@@ -0,0 +1,129 @@
...
+    background-color: ${props => props.theme.palette.warning.main};
+    color: white;
+  }
+`;
+
+export const SyncButton = props => {
+  const { actionConfig, original } = props;
+  const api = useApi();
+  const [status, setStatus] = useState(original.sync_status);
+  const [errorMessage, setErrorMessage] = useState(null);
```
billli0, line 66: I try a bit to make its status persisted. Thing is `DataFetchingTable` always triggers a whole refresh, even though I memo the data.. 

And surprisingly in memo the next props data is always empty.. 

I quit since it costed me for a long while debugging... Happy to keep it as it is.

billli0: One more thing is about the UX on this button. Should it give a green tick if it synced successfully?

rohan-bes: Thanks so much for looking into it man! Yeah it took a few hours out of my life too 😢 

I actually decided to go with a bit of a hacky solution to persist the latest state outside the component entirely in a map. I was happy to leave it too, but there's a really annoying bug where we lost the button status when opening the logs which can be really confusing to the user (makes it look like the sync is working again when it's really still broken).

billli0: I use `[shouldComponentUpdate()](https://reactjs.org/docs/react-component.html#shouldcomponentupdate)` in `DataFetchingTable`, set it return false, then it works fine. The thing is we need to work out the right state comparison.

Again, `nextProps.data` is always empty.. but it can be another solution if you can find out why.

rohan-bes: I feel with such a generic component, working out a standard behaviour for `shouldComponentUpdate()` will be too difficult? What do you think?

billli0: That's true, the 'always refresh' serves row expandation for survey section, the same as the matrix table in Tupaia. To persist the states correctly, it will be a huge refactor for the related components..

And that explains why in Tupaia, we store states for small components like this in global.🙂

I am satisfied with the hacky solution, I think it is a good way to go. I have raised a question whether this button or something should give a green tick if it synced successfully. Adding a success status could allow us to remove the interval waiting time, and it will be no need for an external state.

billli0: Raised [a PR](https://github.com/beyondessential/tupaia/pull/4123) to add a shouldComponentUpdate. Comparing `reduxId`, `numberOfPages` and `data` should satisfy the needs?

billli0 (approved):
LGTM! Nice UX changes. Only two requests for removing unstyled components.

packages/admin-panel/src/sync/SyncStatus.js
```diff
@@ -0,0 +1,214 @@
...
+  // First poll after 0.5 seconds, then poll each 5 seconds
+  useEffect(() => {
+    const timeout = setTimeout(pollStatus, 500);
+    return () => clearTimeout(timeout);
+  }, []);
+
+  useEffect(() => {
+    const timer = setInterval(pollStatus, 5000);
+    return () => clearInterval(timer);
+  }, []);
```
billli0, line 135: @rohan-bes Do you think we should stop polling the sync log while user is not focus on the page?

packages/admin-panel/src/sync/SyncStatus.js
```diff
@@ -0,0 +1,214 @@
...
+  display: flex;
+`;
+
+const StatusMessageContainer = styled.div`
+  display: flex;
+  align-items: center;
+  padding-left: 5px;
+`;
+
+const StatusMessage = styled.div``;
```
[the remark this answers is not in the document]
rohan-bes: Ah my bad 🤦 good catch, thanks!

packages/admin-panel/src/sync/SyncStatus.js
rohan-bes: So I believe it won't poll when a user goes to a page that isn't the Sync Groups page. However it will keep polling while the logs modal is up for instance. I feel that's okay, what do you think? Just in the spirit of keeping things simple

billli0: I've seen it keep polling on the sync groups page. My only concern is it could make the log super long if someone didn't close the page. It is not a huge problem so it is optional.

rohan-bes: I might bump the poll time up to 10 seconds.

I think polling inevitably causes a bit of noise, just the nature of things. A websockets solution is really the way to go for a dynamic front-end, but that's a lot more work 😓

PR #4121, opened 2022-08-25
MAUI-1059 Lesmis fetch by entity not country

IgorNadj:
Lesmis map overlays were fetching against the country, changed to fetch against the selected org unit.

Before: https://linear.app/bes/issue/MAUI-1059#comment-75141f0c
After: https://linear.app/bes/issue/MAUI-1059#comment-9b71d1ec


rohan-bes (commented):
Will this break existing map overlays? I think Tupaia also fetches at country level (unless there's a display at level setting...)

IgorNadj:
> Will this break existing map overlays? I think Tupaia also fetches at country level (unless there's a display at level setting...)

TBH I'm not sure, from my quick check it seems to work alright. I was going to get a tester to have a look.

rohan-bes (approved):
Nice one!

PR #4123, opened 2022-08-26
Add shouldComponentUpdate in DataFetchingTable

billli0:
### Changes:

- Example

---

### DIFF
--- a/packages/admin-panel/src/table/DataFetchingTable.js
+++ b/packages/admin-panel/src/table/DataFetchingTable.js
@@ -69,6 +69,13 @@ class DataFetchingTableComponent extends React.Component {
     }
   }
 
+  shouldComponentUpdate(nextProps) {
+    if (this.props.reduxId !== nextProps.reduxId) return true;
+    if (this.props.numberOfPages !== nextProps.numberOfPages) return true;
+    if (JSON.stringify(this.props.data) !== JSON.stringify(nextProps.data)) return true;
+    return false;
+  }
+
   renderConfirmModal() {
     const {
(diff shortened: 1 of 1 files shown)


packages/admin-panel/src/table/DataFetchingTable.js
```diff
@@ -69,6 +69,13 @@ class DataFetchingTableComponent extends React.Component {
     }
   }
 
+  shouldComponentUpdate(nextProps) {
+    if (this.props.reduxId !== nextProps.reduxId) return true;
+    if (this.props.numberOfPages !== nextProps.numberOfPages) return true;
+    if (JSON.stringify(this.props.data) !== JSON.stringify(nextProps.data)) return true;
+    return false;
+  }
```
rohan-bes, line 77: Hmmm I feel this is way to strict about when it re-renders... And I think it causes bugs!
When I use this code, and go to the Survey tab and try expand a survey, it can't open.

Tbh I'm not too comfortable about adding a `shouldComponentUpdate()` here, I feel this is a pretty complex component, and trying to control its render cycle is going to be pretty difficult. 

It seems to me that the underlying issue is that we're using redux for local state management, and the reducers often kick off re-renders for all sorts of components. Moving the admin panel over to using react context / react query will probably be the way we get out of these issues? And in the meanwhile we just bite the bullet with excessive re-renders 🤷

billli0: Yep it is very easy to cause bug when adding or changing the props comparison in this huge component.. Moving the state to parent component is a better way.

PR #4125, opened 2022-08-29
Rn 506 part 1

billli0:
### Issue #:
RN-506

### Changes:
- Enable favourite button to every dashboard item, which is only shown when logged in.

- endpoints: 
-- GET `/userFavouriteDashboardItems` used to list user's favourite dashboard items in `lesmis/src/api/queries/useFavouriteDashboardItem.js`
-- POST `/userFavouriteDashboardItem` used to create or update record.
(I didn't choose to use DELETE endpoint type because we don't sent `record_id` to frontend);

- new table `user_favourite_dashboard_item`

- Delete '[deleteMyAccount.js](https://github.com/beyondessential/tupaia/pull/4125/files#diff-ec9a5656bca7397e8c0de92d3f171b2ab73627e7b6029066ab203ae3f876b9d8)' as it has no usage.


packages/database/src/migrations/20220823234713-AddLesmisFavouriteDashboardTable-modifies-schema.js
```diff
@@ -0,0 +1,41 @@
...
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`
+    CREATE TABLE user_favourite_dashboard_item (
+      id TEXT PRIMARY KEY,
+      user_id TEXT NOT NULL,
+      dashboard_item_id TEXT NOT NULL,      
+      UNIQUE (user_id, dashboard_item_id),
```
billli0, line 23: A existing record represents a user put the dashboard item in their favourite list. 

`UNIQUE` will make sure there is no duplicated records.

packages/lesmis/src/components/Visuals/Chart.js
```diff
@@ -173,7 +182,13 @@ export const Chart = ({
   ) : (
     <>
       <VisualHeader name={name} isLoading={isFetchingInBackground}>
-        <Toggle onChange={handleTabChange} value={selectedTab} exclusive />
+        <GridContainer>
+          <Toggle onChange={handleTabChange} value={selectedTab} exclusive />
+          <FavouriteButton
+            favouriteStatus={favouriteStatus}
+            handleFavouriteStatusChange={handleFavouriteStatusChange}
+          />
```
rohan-bes, line 190: Hey I wanted to make a suggestion about the whole `favouriteStatus` thing. So as far as I understand, it seems we use the `IDLE` state just to refer to the how to treat the component when the user is not logged in?

It seems to me it could be simpler if we just handled that logic separately, eg.
```suggestion
          // Don't render FavouriteButton if user is not logged in
          userIsLoggedIn && <FavouriteButton
            favouriteStatus={favouriteStatus}
            handleFavouriteStatusChange={handleFavouriteStatusChange}
          />
```

Then we could revert the `favouriteStatus` to just being a simple boolean `isFavourite`, which I feel makes things a lot clearer. Thoughts?

packages/database/src/modelClasses/UserFavouriteDashboardItem.js
```diff
@@ -0,0 +1,29 @@
...
+  async updateRecord({ changeType, record }) {
+    switch (changeType) {
+      case 'create':
+        return this.findOrCreate(record);
+      case 'delete':
+        return this.delete(record);
+      default:
+        throw new Error(`Non supported change type: ${changeType}`);
+    }
+  }
```
rohan-bes, line 28: Thoughts on making this function use terminology that's a bit closer to the business terminology. eg.

```suggestion
  async favourite(userId, dashboardItemId) {
    return this.findOrCreate({dashboard_item_id: dashboardItemId, user_id: userId });
  }
  
  async unfavourite(userId, dashboardItemId) {
    return this.delete({dashboard_item_id: dashboardItemId, user_id: userId });
  }
```

Then we can just do:
`models.userFavouriteDashboardItem.favourite(userId, dashboardItemId);`
and
`models.userFavouriteDashboardItem.unfavourite(userId, dashboardItemId);`

packages/lesmis/src/api/mutations/useUpdateFavouriteDashboardItem.js
```diff
@@ -0,0 +1,52 @@
...
+ * Tupaia
+ *  Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+import { useMutation } from 'react-query';
+import { debounce } from '@tupaia/utils';
+
+import { IS_NOT_FAVOURITE, IS_FAVOURITE } from '../../constants';
+import { post } from '../api';
+
+const useLikeDashboardItem = () =>
```
rohan-bes, line 11:
```suggestion
const useFavouriteDashboardItem = () =>
```
Let's keep the terminology consistent

packages/lesmis/src/api/mutations/useUpdateFavouriteDashboardItem.js
```diff
@@ -0,0 +1,52 @@
...
+  useMutation(dashboardItemCode =>
+    post('userFavouriteDashboardItem', {
+      data: {
+        changeType: 'create',
+        dashboardItemCode,
+      },
+    }),
+  );
+
+const useDislikeDashboardItem = () =>
```
rohan-bes, line 21:
```suggestion
const useUnfavouriteDashboardItem = () =>
```
Likewise

packages/database/src/migrations/20220823234713-AddLesmisFavouriteDashboardTable-modifies-schema.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
rohan-bes, line 1: Could rename this file to: `20220823234713-AddUserFavouriteDashboardTable-modifies-schema.js`

No biggie though 👍

packages/utils/src/debounce.js
```diff
@@ -0,0 +1,10 @@
+let timer;
+
+export const debounce = (func, duration) => {
+  return (...args) => {
+    clearTimeout(timer);
+    timer = setTimeout(() => {
+      func(...args);
+    }, duration);
+  };
+};
```
rohan-bes, line 10: Looks like we've got a whole bunch of ways we do debouncing throughout the codebase 😅 
I'd probably argue we should just standardise to use `lodash.debounce` (we already use it in the admin-panel) rather than custom code? Sorry to be _that guy_ 🙏

packages/lesmis/src/api/queries/useDashboardReportData.js
```diff
@@ -43,16 +46,31 @@ const getReportCodesByCode = dashboards => {
...
   startDate,
   endDate,
 }) => {
+  const { userId } = useUser();
+
   const query = combineQueries({
     reportData: useDashboardReportData({ entityCode, reportCode, startDate, endDate }),
     dashboards: useDashboardData({ entityCode }),
   });
+  const favouriteDashboardItems = useFavouriteDashboardItem({ userId })?.data;
```
rohan-bes, line 73: Feels a little awkward that it's the frontend's responsibility to map which dashboards the user has favourited, ideally this is managed by the server.

However, since we just query through to central-server via lesmis-server, I feel this is the cleanest way for now 👍 

One day I'd love to switch the central-server to graphql, then we could do all the data joins we need in a single query!

rohan-bes (changes requested):
Great stuff @billli0! A few comments about renaming things, and simplifying the `favouriteStatus` stuff, keen to hear your thoughts?

packages/central-server/src/apiV2/index.js
```diff
@@ -242,6 +246,7 @@ apiV2.post('/dashboardRelations', useRouteHandler(CreateDashboardRelation));
 apiV2.post('/dashboardVisualisations', useRouteHandler(CreateDashboardVisualisation));
 apiV2.post('/mapOverlayVisualisations', useRouteHandler(CreateMapOverlayVisualisation));
 apiV2.post('/mapOverlayGroupRelations', useRouteHandler(CreateMapOverlayGroupRelation));
+apiV2.post('/userFavouriteDashboardItem', allowAnyone(handleUserFavouriteDashboardItem));
```
rohan-bes, line 249:
```suggestion
apiV2.post('/userFavouriteDashboardItems', allowAnyone(handleUserFavouriteDashboardItem));
```
I think we always use plural, even if it just edits a single record 👍

packages/lesmis/src/components/Visuals/Chart.js
billli0: Yeah thanks for that, it is clearer! Will do the changes.

packages/central-server/src/apiV2/userFavouriteDashboardItem/handleUserFavouriteDashboardItem.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+import { respond } from '@tupaia/utils';
+
+/**
+ * Handles POST endpoints:
+ * - /userFavouriteDashboardItem
+ */
+
+export const handleUserFavouriteDashboardItem = async (req, res, next) => {
```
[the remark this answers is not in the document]
billli0: This feature will be publicly accessed, so it doesn't require to have a user permission check by `assertUserHasAccess()`. Think it should be fine?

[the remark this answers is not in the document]
rohan-bes: Also, a user should only be able to change their own favourites, right? So we should determine the `userId` via the auth header I think...

packages/utils/src/debounce.js
billli0: I've tried `lodash.debounce` and it didn't work as expected. Sadly the timer in `lodash.debounce` is somehow being recreated, even with `useCallback()` .. 

Here I've pull out the timer to make sure it is persisted.

rohan-bes: Ahh that's annoying 😢 

Now that you mention it though, I've got a bit of a concern around this `debounce`. What happens if there are multiple separate uses of it? Won't they share the same `timer`, which will break one of them?

billli0: Good thinking and thanks! I just found that's an existing bug.. Have found a way to persist timer so we can use `lodash.debounce` here, see the latest commit.

packages/central-server/src/apiV2/userFavouriteDashboardItem/POSTUpdateUserFavouriteDashboardItem.js
```diff
@@ -0,0 +1,47 @@
...
+        throw new Error('User should be logged in');
+      }
+    };
+
+    return this.assertPermissions(assertAnyPermissions([assertUserIsLoggedIn]));
+  }
+
+  async handleRequest() {
+    const { models, body, userId } = this.req;
+    const { dashboardItemCode, changeType } = body;
```
rohan-bes, line 29: Instead of `changeType` being `create` or `delete` here, wanna instead stick with the `favourite` and `unfavourite` terms? 

So the request looks like:
```
POST /userFavouriteDashboardItem
{
  dashboardItemCode: 'LESMIS_viz',
  state: 'favourite'
}

or

POST /userFavouriteDashboardItem
{
  dashboardItemCode: 'LESMIS_viz',
  state: 'unfavourite'
}
```

I like the idea of hiding the fact that we're creating/deleting individual `userFavouriteDashboardItem` records. That way the client can just think about it as favouriting and unfavouriting 👍

packages/lesmis/src/api/queries/useFavouriteDashboardItem.js
```diff
@@ -0,0 +1,23 @@
+import { stringifyQuery } from '@tupaia/utils';
+import { useQuery } from 'react-query';
+import { get } from '../api';
+import { QUERY_OPTIONS } from './constants';
+
+export const useFavouriteDashboardItem = ({ userId }) => {
```
rohan-bes, line 6:
```suggestion
export const useGetFavouriteDashboardItems = ({ userId }) => {
```

Just a rename suggestion to not confusing it with the function to set a dashboard item as favourite. It's a bit confusing cos 'favourite' is both a noun (ie. a favourite dashboard) and a verb (ie. the user chose to favourite the dashboard). So this makes things a bit clearer 👍

packages/central-server/src/apiV2/userFavouriteDashboardItem/GETUserFavouriteDashboardItems.js
```diff
@@ -0,0 +1,22 @@
...
+
+  async findRecords(criteria, options) {
+    const { user_id: userId } = JSON.parse(this.req.query.filter);
+    const loggedUserId = this.req.userId;
+    if (!userId || userId !== loggedUserId) {
+      return [];
+    }
+
+    return this.database.find(this.recordType, criteria, options);
+  }
```
rohan-bes, line 21:
```suggestion
  async findRecords(criteria, options) {
    const loggedUserId = this.req.userId;
    if (!loggedUserId) {
      return [];
    }

    return this.database.find(this.recordType, { ...criteria, user_id: loggedUserId }, options);
  }
```

Seems like this is only going to return the `userFavouriteDashboardItems` of the logged in user anyway right? I like that, makes sense. So then you could change this to not need a `user_id` query parameter at all 👍

rohan-bes (changes requested):
Just a couple little change requests, aside from that it's looking awesome! Sorry if the review was before you were ready for re-review, I had some spare time and wanted to clear my plate 😅

packages/lesmis/src/api/queries/useFavouriteDashboardItem.js
billli0: Love it!

packages/web-config-server/src/apiV1/dashboards.js
```diff
@@ -88,6 +89,9 @@ export default class extends RouteHandler {
               legacy: item.legacy,
               reportCode: item.report_code,
               ...item.config,
+              isFavourite: userFavouriteDashboardItems
+                ? userFavouriteDashboardItems.includes(item.code)
+                : null,
```
billli0, line 94: @rohan-bes I feel that mapping user favourite dashboard items in frontend is becoming bigger while I am working on part 2..

Adding `isFavourite` here makes thing smooth.

In part 2 I will add a query parameter `favouriteOnly` to allow fetching favourite dashboards only.

Does it look good?

packages/web-config-server/src/authSession/authSession.js
```diff
@@ -15,6 +15,7 @@ const auth = () => async (req, res, next) => {
     // if using basic or bearer auth, check credentials and set access policy for that user
     const authHeaderUser = await getUserFromAuthHeader(req);
     if (authHeaderUser) {
+      req.userJson = { userId: authHeaderUser.id };
       req.accessPolicy = await getAccessPolicyForUser(authenticator, authHeaderUser.id);
```
billli0, line 19: It is really confused me here as we assign `userJson` and `accessPolicy` from session and auth header in this file, but I think it is required for handling permission for servers and frontend at the same time. 

I think it should be fine to assign `userJson` from header? With that `dashboards` router can access `userId` from `this.req.userJson`.

billli0:
Hey @rohan-bes thanks for the last review. I've removed `GETUserFavouriteDashboardItem` route, instead `dashboards` route can generate config `isFavourite` for every dashboard item.

packages/web-config-server/src/apiV1/dashboards.js
rohan-bes: Yup, sounds good to me 👍

packages/web-config-server/src/authSession/authSession.js
rohan-bes: Ahaha honestly this area of code always confuses me a bit. But yup, I like changing it so that we can reference by the `userJson` field 👍

rohan-bes (approved):
Nice one! It always confuses me that LESMIS pull data from web-config-server haha 😅

PR #4128, opened 2022-08-30
MAUI-1157 Restore entity name in analytics table

IgorNadj:
### Issue #: MAUI-1157

Restore entity name in analytics table.

See comment https://linear.app/bes/issue/MAUI-1157#comment-21b3a595

To see the actual changes of this PR, you have to compare to the last version of the build_analytics_table function:
Screen Shot 2022-08-30 at 10 37 13 am


rohan-bes (approved):
Thanks for fixing this @IgorNadj 

Makes me wonder though, if we need a better way of managing the `build_analytics_table` function so that we actually have it in version control in a single place (as opposed to a different migration each time we change it).

For the `permissions_based_meditrak_sync_queue` this is handled by having a script that rebuilds it, so that the logic to create it is just one one place.

Not saying it should be part of this change, but perhaps we want to add a script to install the `build_analytics_table` function in the database, and have that script run as part of the deployment process?

IgorNadj:
@rohan-bes Great idea! I've moved it to data-api
Final time we will need a screenshot for a diff 😆 :
Screen Shot 2022-08-30 at 3 21 44 pm

packages/data-api/scripts/createBuildAnalyticsTableFn.sh
```diff
@@ -0,0 +1,11 @@
...
+
+DIR=$(pwd "$0")
+source $DIR/../../scripts/bash/mergeCurrentEnvWithEnvFile.sh 
+
+# Set default port in case it wasn't in .env
+: "${DB_PORT:=5432}"
+
+cd scripts
+export PGPASSWORD=$DB_PASSWORD
+psql -p $DB_PORT -h $DB_URL -d $DB_NAME -U $DB_USER < createBuildAnalyticsTableFn.sql
```
rohan-bes, line 11: Not too opinionated, but `psql` also has the `-f` flag for passing in a file if you prefer that?

rohan-bes:
Thanks so much for switching to this being managed by a single file! Bugs should be a lot easier to avoid now.

Ah! Looks like your build is failing due to the catch22 of the analytics table requiring the migrations, and the migrations requiring the analytics table 🙀 

It's probably best solved by updating the test database, by running:
`yarn workspace @tupaia/database update-test-data`

However, I think you need to drop the analytics table from your local db first! Happy to jump on a call if you want a hand with things 👍

packages/data-api/scripts/createBuildAnalyticsTableFn.sh
IgorNadj: Actually I prefer pipes :) Generic shell syntax

IgorNadj:
> Ah! Looks like your build is failing due to the catch22 of the analytics table requiring the migrations, and the migrations requiring the analytics table 🙀

Yeah tricky one. I think I've fixed it for CI by reordering things so migrate happens before we create the fn. 

However, this has me thinking, in deployments we will have the same issue. E.g. if someone renames the data_element table and updates the fn in the same deployment... depending on how quick migrations finish [when central-server starts](https://github.com/beyondessential/tupaia/blob/dev/packages/central-server/src/index.js#L93)... it may reach the `yarn workspace @tupaia/data-api create-build-analytics-table-fn` line first.

I think our mixing of doing things in startup code + the deployment scripts isn't going to work. We will need to move all to one or the other. Since the migration logic was moved to code most recently I suggest we move everything to code. Thoughts?

rohan-bes:
> I think our mixing of doing things in startup code + the deployment scripts isn't going to work. We will need to move all to one or the other. Since the migration logic was moved to code most recently I suggest we move everything to code. Thoughts?

I like the idea of having it all in one place. And yeah, code seems easier to manage (and definitely cleaner than shell scripts).

I do think this particularly issue is a genuine circular dependency though.
- We want the analytics table installed prior to running the migrations (in order to ensure we rebuild it when migrations change any underlying data)
- However, we also want to run the migrations before we install the analytics table (to ensure that the schema matches whatever's in the `build_analytics_table` function)

IgorNadj:
Discussed
```
build_analytics_fn v1
  relies on data_element


migration renames data_element to data_element2


build_analytics_fn v2                 <------ import this from ______.v1.sql
  relies on data_element2


migration renames data_element2 to data_element3


build_analytics_fn v3
  relies on data_element3          <------------ import this from ______.v2.sql
```

IgorNadj:
I can't approve because I created the PR but looks great, thanks for that!

PR #4129, opened 2022-08-30
RN-614: Add 'Help centre' link to the User Menu

rohan-bes:
### Issue RN-614:

Pretty simple change to add a link to our public slab help page to Tupaia's user menu.

Also required re-working the menu system a little bit since there was no menu for non-logged in users previously.

Also updated the styling for the search bar, which is a bit out of scope, but may as well do it now while I'm looking at the code...

Screenshots in the ticket.


packages/web-frontend/src/containers/OverlayDiv/index.js
```diff
@@ -70,6 +76,9 @@ export const OverlayDiv = ({
 
   const components = {
     [LANDING]: () => <LandingPage isUserLoggedIn={isUserLoggedIn} transition />,
+    [VIEW_PROJECTS]: () => (
+      <LandingPage isUserLoggedIn={isUserLoggedIn} isViewingProjects transition />
+    ),
```
rohan-bes, line 81: Is this a tad awkward? Felt like the simplest way I could re-route the non-logged in user to the 'View projects' page...

packages/web-frontend/src/containers/UserMenu/index.js
```diff
@@ -118,23 +148,62 @@ class UserMenu extends Component {
...
+    const ViewProjects = () => (
+      <MenuItem
+        onClick={() => {
+          openViewProjects();
+          this.closeUserMenu();
+        }}
+      >
+        View projects
+      </MenuItem>
+    );
```
rohan-bes, line 165: Dunno how standard it is to define inline components like this, but since they use internal state, I figure this is the best way to re-use components?

billli0: Yeah if the component is small we can put it here.

billli0 (approved):
Looks great!

packages/web-frontend/src/containers/OverlayDiv/index.js
rohan-bes: Good idea! Will do 👍

PR #4143, opened 2022-09-05
RN-641: Support importing project entities without a country

rohan-bes:
### Issue RN-641:

Support being able to import project entities without a country associated with them. Project entities shouldn't have an associated country, and setting one may cause issues. 

Specify new value 'No Country' in the sheet name, or the geojson property, and the entities will be imported without a country (only works for projects).

### DIFF
--- a/packages/central-server/src/apiV2/import/importEntities/extractEntitiesByCountryName.js
+++ b/packages/central-server/src/apiV2/import/importEntities/extractEntitiesByCountryName.js
@@ -10,6 +10,16 @@ import xlsx from 'xlsx';
 
 import { convertCellToJson } from '../importSurveys/utilities';
 
+export const VALID_ENTITY_TYPES_WITH_NO_COUNTRY = ['project'];
+export const NO_COUNTRY = 'No Country';
+
+const validateNoCountryEntities = entities =>
+  entities.forEach(entity => {
+    if (!VALID_ENTITY_TYPES_WITH_NO_COUNTRY.includes(entity.entity_type)) {
+      throw new Error(`Country must be specified for entity type: ${entity.entity_type}`);
+    }
+  });
+
 const geojsonParser = filePath => {
   const fileContents = fs.readFileSync(filePath, { encoding: 'utf-8' });
   const { features } = JSON.parse(fileContents);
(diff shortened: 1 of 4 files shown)


packages/central-server/src/apiV2/import/importEntities/updateEntities.js
```diff
@@ -130,12 +131,27 @@ export async function updateCountryEntities(
...
       );
     }
     codes.push(code);
-    const { parentGeographicalArea, parentEntity } =
-      (await getOrCreateParentEntity(transactingModels, entityObject, country, pushToDhis)) || {};
+    let parentGeographicalArea;
+    let parentEntity;
+
+    // Entities without a country (ie. projects) do not have parent entities
+    if (country) {
```
rohan-bes, line 142: This felt a little hairy, but I think it's a safe enough assumption. Overall this area of the code I find really complex and hard to reason about 😢

packages/central-server/src/apiV2/import/importEntities/updateEntities.js
```diff
@@ -130,12 +131,27 @@ export async function updateCountryEntities(
...
+      const parentData = await getOrCreateParentEntity(
+        transactingModels,
+        entityObject,
+        country,
+        pushToDhis,
+      );
+      if (parentData) {
+        parentGeographicalArea = parentData.parentGeographicalArea;
+        parentEntity = parentData.parentEntity;
+      }
```
billli0, line 152: `getOrCreateParentEntity` should always return value, do you think it is better to throw error rather than return null in `getOrCreateParentEntity`?

packages/central-server/src/apiV2/import/importEntities/extractEntitiesByCountryName.js
```diff
@@ -52,13 +69,27 @@ const processXlsxRow = (row, { countryName }) => {
...
   const workbook = xlsx.readFile(filePath);
   return Object.entries(workbook.Sheets).reduce(
-    (entitiesByCountry, [countryName, sheet]) => ({
-      ...entitiesByCountry,
-      [countryName]: xlsx.utils
-        .sheet_to_json(sheet)
-        .map(row => processXlsxRow(row, { countryName })),
-    }),
-    {},
+    ({ entitiesByCountry, entitiesWithoutCountry }, [countryName, sheet]) => {
```
billli0, line 72: Is `entitiesByCountry` originally contains `entitiesByCountry` and `entitiesWithoutCountry`? Or it has something to do with the changes for `geojsonParser`?

packages/central-server/src/apiV2/import/importEntities/updateEntities.js
billli0: Yeah lots of 'if' :) No problem to have another one.

billli0 (approved):
Code looks really great! I just got a few questions on different places. 

My biggest question is do we want this feature if we can create project in admin panel powered by MAUI-859? Probably they serve different purposes though.

rohan-bes:
> My biggest question is do we want this feature if we can create project in admin panel powered by [MAUI-859](https://linear.app/bes/issue/MAUI-859)? Probably they serve different purposes though.

Great question @billli0. Honestly I had followed up on this request from someone in the Maui team asking for it, but I'll double check with @chris-pollard to confirm if it's redundant now...

rohan-bes:
Good call @billli0, I don't think this is needed anymore.

Closing it out

PR #4145, opened 2022-09-05
(hotfix) MAUI-1228 Map overlay look deeper for polygon data

IgorNadj:
### Changes:

- Map overlay look deeper for polygon data

---

### Screenshots:
See description in https://linear.app/bes/issue/MAUI-1228/look-deeper-than-grandchild-when-rendering-interactive-polygons-for


rohan-bes (approved):
Nice fix @IgorNadj! I wonder why it was hardcoded to only as deep as grandchildren before 🤔 this is much more intuitive and robust!

PR #4149, opened 2022-09-06
MAUI-1010: Add custom Palau Nursing Unit reports

chris-pollard:
### Changes:

- Add custom reports for Palau Environment project


rohan-bes (approved):
Looks great @chris-pollard ! Really love how you broke out the data into a separate json file 👍 

If I was being a real meanie, I'd request for some basic unit tests around this report. Generally makes it a bit easier for when we want to switch it to a regular report if we know the expected input/output. But it's optional 😉

chris-pollard:
Cancelled. No longer required now we have data tables.

PR #4150, opened 2022-09-07
RN-506 part 2

billli0:
### Issue #:
RN-506
 
### Changes:
- Remove `indicators` dashboard as we don't need it any more.
- Some variables with name `dashboard` is referring to `subDashboard`, have renamed them.


packages/lesmis/src/utils/useDropdownOptionsWithFavouriteDashboardItems.js
```diff
@@ -0,0 +1,32 @@
+import { useDashboardData } from '../api';
+import { useDashboardDropdownOptions } from './useDashboardDropdownOptions';
+import { useUrlParams } from './useUrlParams';
+
+const getDropdownOptionsWithFavouriteDashboardItems = (data = []) => {
+  const { favouriteDropdownOption, otherDropdownOptions } = useDashboardDropdownOptions();
+  const subDashboardsWithFavouriteDashboardItems = data.filter(
+    favouriteDropdownOption.componentProps.filterSubDashboards,
+  );
+
```
billli0, line 10: This will map `dropdownOptions` with favourite dashboard items only, which helps us not to render unfavourite dashboard items.

We don't need to do that if we move `dropdownOptions` to backend, but I feel that will generate more work..

packages/lesmis/src/views/DashboardView.js
```diff
@@ -8,27 +8,65 @@ import PropTypes from 'prop-types';
...
+  switch (tabViewType) {
+    case DASHBOARD_REPORT_TAB_VIEW:
+      return DashboardReportTabView;
+    case FAVOURITE_DASHBOARD_TAB_VIEW:
+      return FavouriteDashboardTabView;
+    case TAB_TEMPLATE:
+    default:
+      return TabTemplate;
+  }
+};
```
billli0, line 68: We got three TabComponent now😄

packages/lesmis/src/utils/useDefaultDashboardTab.js
```diff
@@ -0,0 +1,37 @@
+import { useHistory } from 'react-router-dom';
+import { useUser } from '../api';
+import { useI18n } from './I18n';
+import { useHomeUrl } from './useHomeUrl';
+
+// Gets the best default dashboard possible, and check if the selected dashboard is valid
+export const useDefaultDashboardTab = (selectedDashboard = null, options) => {
```
billli0, line 7: Originally from 'DashboardItemsView.js'

packages/lesmis/src/constants/constants.js
```diff
@@ -15,53 +15,62 @@ export const LESMIS_PERMISSION_GROUPS = {
...
-    code: 'profile',
+    value: 'profile',
+    tabViewType: DASHBOARD_REPORT_TAB_VIEW,
     useYearSelector: true,
     exportToPDF: true,
   },
   {
-    code: 'indicators',
-    label: 'dashboards.freeIndicatorSelection',
+    value: 'favourites',
```
rohan-bes, line 32: Can we make this a constant? eg.
`const FAVOURITES_DASHBOARD_CODE = 'favourites'`

I see it being re-used in packages/lesmis/src/utils/getExportableSubDashboards.js

packages/lesmis/src/api/mutations/useUpdateFavouriteDashboardItem.js
```diff
@@ -2,37 +2,75 @@
...
+import { useMutation, useQueryClient } from 'react-query';
 import { post } from '../api';
 
-const useFavouriteDashboardItem = () =>
-  useMutation(dashboardItemCode =>
-    post('userFavouriteDashboardItems', {
-      data: {
-        state: 'favourite',
-        dashboardItemCode,
+// Instead of refetching dashboard for 'isFavourite' config update, we update the current existing dashboard data.
```
rohan-bes, line 8: What's the rationale for not re-fetching the data here? It feels a little heavy handed / unsafe to manually update the state here?

packages/lesmis/src/utils/useDropdownOptionsWithFavouriteDashboardItems.js
rohan-bes: Nah I think this is a good way to go 👍 seems a front-end concern

rohan-bes (changes requested):
Awesome stuff @billli0! Had a play around with this feature on my local and it's really neat!

Just one question about how come we aren't re-fetching the dashboard when the favourites change, and a small comment about consolidating a string, but aside from that it's all looking good!

packages/lesmis/src/api/mutations/useUpdateFavouriteDashboardItem.js
billli0: I thought re-fetch has to call the re-fetch function from dashboard `fetchQuery` callback function, which will be heavy to pass that to here. 

But I just found that we can refetch a specific query from `queryClient`, that's super handy!

billli0:
Thanks @rohan-bes for the inspiration. I've added the constant for `favourite` and `profile` dashboard, and modify the update on success logic.

packages/lesmis/src/utils/useDashboardDropdownOptions.js
```diff
@@ -62,6 +53,8 @@ export const useDashboardDropdownOptions = () => {
 
   return {
     dropdownOptions,
+    favouriteDropdownOption: dropdownOptions.find(({ value }) => value === 'favourites'),
```
rohan-bes, line 56: One more spot we're using the `'favourites'` magic string

rohan-bes (approved):
Nice work @billli0. One more spot we're using a string when we could use a the `const`, but pre-approving anyway 👍

packages/lesmis/src/utils/useDashboardDropdownOptions.js
billli0: Good spotting!

PR #4152, opened 2022-09-08
RN-646: Add DataTable model to the database

rohan-bes:
### Issue RN-646:

Add the data_table table to the database.

**fields:**
id: id,
code: human readable unique identifier,
description: what the data table does
type: eg. (`sql`, `google-sheets`, `transformation`, etc.)
config: JSON for the specific data table config
permission_groups: Array of permission group names. Similar to how we do for permissions on data elements. A little conflicted about this, but honestly the permissions model for data tables still needs a bit of working out anyway, so this may just be a placeholder


IgorNadj (approved):
Happy to approve, but not sure if you want to just wait until you have a first implementation? I find that implementation details change things quite often. Up to you

rohan-bes:
> Happy to approve, but not sure if you want to just wait until you have a first implementation? I find that implementation details change things quite often. Up to you

@IgorNadj good point! Yeah I never know the best way to go with this kind of stuff 😅 I probably won't merge this one until I've got most of an initial implementation working on my local. But I like to break up the PRs rather than have one beastly one!

PR #4153, opened 2022-09-08
RN-265 Always use dhis instance code on data element / group

IgorNadj:
### Changes:

- Always use the dhisInstanceCode marked against the DE/DG when pulling via data-broker

Note: this PR is based on rn-625-disable-1-many-mapping, therefore once both are in, there will be no cases* where we use entity based resolution to figure out which DHIS instance to pull from. To do 625 cleanly, we also need to do this.

* in data-broker... After these changes, there will still be parts of the codebase which rely on entity config for this, e.g. the dhis sync logic, which will be migrated over in a future PR.

### DIFF
--- a/packages/data-broker/src/services/dhis/DhisService.js
+++ b/packages/data-broker/src/services/dhis/DhisService.js
@@ -3,12 +3,7 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-import {
-  getApiForDataSource,
-  getApiFromServerName,
-  getApisForDataSources,
-  getApisForLegacyDataSourceConfig,
-} from './getDhisApi';
+import { getApiForDataSource, getApiFromServerName, getApisForDataSources } from './getDhisApi';
 import { Service } from '../Service';
 import { DhisTranslator } from './translators';
 import {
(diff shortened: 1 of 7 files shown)


packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -3,12 +3,7 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
```
IgorNadj, line 3: Also, there's a comment in the card:
>We should:
>- Remove any client code and redundant DB data related to the above (eg legacy_report.data_services)

I left all the data alone because it's legacy and going away soon, happy with that? Or do you think we should clean it up?

packages/data-broker/src/services/dhis/DhisService.js
```diff
@@ -151,29 +143,8 @@ export class DhisService extends Service {
 
   deleteEvent = async (api, data) => api.deleteEvent(data.dhisReference);
 
-  async getPullApis(dataSources, entityCodes, detectDataServices, dataServices) {
-    const apis = detectDataServices
```
IgorNadj, line 155: The main bit of this PR, the rest is deleting stuff

packages/data-broker/src/services/dhis/DhisService.js
rohan-bes: Hahaha well wether it's going away soon or not is questionable 😉 but I'm cool with leaving it on the legacy data builders 👍

rohan-bes (approved):
Awesome stuff! Bless all this red ❤️

PR #4155, opened 2022-09-08
RN-357: Breadcrumbs through geographic hierarchy

chris-pollard:
### Changes:

- When viewing the main Tupaia interface, a set of breadcrumbs showing the hierarchy and linking to higher entities will display at the top of the side panel (note: breadcrumbs will not show if the current entity is the root, e.g. UNFPA for UNFPA project).


packages/web-frontend/src/containers/Dashboard.js
```diff
@@ -155,6 +184,7 @@ export class Dashboard extends Component {
 
     return (
       <div style={{ ...DASHBOARD_STYLES.metaImageHolder, height: mediaWidth * IMAGE_HEIGHT_RATIO }}>
+        {this.renderBreadcrumbs(currentOrganisationUnit)}
```
IgorNadj, line 187: minor: arg currentOrganisationUnit not needed?

IgorNadj (approved):
Nice once

PR #4158, opened 2022-09-09
RN-603 de dg multi mapping

IgorNadj:
### Issue #: RN-603

Replaces https://github.com/beyondessential/tupaia/pull/4081

### Changes:

- Allow mapping a Data Element to multiple data services based on country.

---

### Screenshots:
https://linear.app/bes/issue/RN-603#comment-2a6c52be


IgorNadj:
Tests were failing because I accidentally missed a commit after a rebase onto dev, cherry-picked it back in now

packages/data-broker/src/services/DataServiceResolver.js
```diff
@@ -0,0 +1,181 @@
...
+    const resolved = [];
+    for (const dataElement of dataElements) {
+      const deMappings = mappings.filter(m => m.data_element_code === dataElement.code);
+
+      if (deMappings.length > 1) {
+        // If they all resolve to the same data service, this is fine
+        const resolveToTheSameDataService = (mappingA, mappingB) => {
+          if (mappingA.service_type !== mappingB.service_type) return false;
+          return isequal(mappingA.service_config, mappingB.service_config);
+        };
```
rohan-bes, line 120: Up to you, but might be nice to move this function out to somewhere external as it doesn't require any local state 👍

packages/data-broker/src/services/DataServiceResolver.js
```diff
@@ -0,0 +1,181 @@
...
+        });
+        continue;
+      }
+
+      // No mappings, use default
+      resolved.push({
+        dataSource: dataElement,
+        service_type: dataElement.service_type,
+        config: dataElement.config,
+      });
```
rohan-bes, line 158:
```suggestion
      if (deMappings.length === 0) {
        // No mappings, use default
        resolved.push({
          dataSource: dataElement,
          service_type: dataElement.service_type,
          config: dataElement.config,
        });
        continue;
      }
      
        // If they all resolve to the same data service, this is fine
        const resolveToTheSameDataService = (mappingA, mappingB) => {
          if (mappingA.service_type !== mappingB.service_type) return false;
          return isequal(mappingA.service_config, mappingB.service_config);
        };
        for (const mappingA of deMappings) {
          for (const mappingB of deMappings) {
            if (mappingA === mappingB) continue;
            if (!resolveToTheSameDataService(mappingA, mappingB)) {
              throw new Error(
                `Conflicting mappings found for Data Source ${
                  dataElement.code
                } when fetching for orgUnits ${orgUnits.map(o => o.code).join(',')}`,
              );
            }
          }
        }
        
        // all data element mappings are the same, use first
        const [deMapping] = deMappings;
        resolved.push({
          dataSource: dataElement,
          service_type: deMapping.service_type,
          config: deMapping.service_config,
        });
```

Optional, but this cuts down the nesting and line count a bit if you like?

rohan-bes (approved):
Nice work @IgorNadj, really love how you broke out the dataService resolution logic into a separate class.

So they were cool going with a separate country for UNFPA Warehouse? That's great, makes life easier for us 😅 

Couple of code style comments, but nothing major so pre-approving

packages/data-broker/src/services/DataServiceResolver.js
```diff
@@ -0,0 +1,181 @@
...
+   * @returns {Promise<DataServiceMapping>}
+   */
+  async getMappingMultipleOrgUnits(dataSources, orgUnits = []) {
+    const dataElements = dataSources.filter(ds => ds.databaseType === TYPES.DATA_ELEMENT);
+    const dataGroups = dataSources.filter(ds => ds.databaseType === TYPES.DATA_GROUP);
+
+    const mapping = new DataServiceMapping();
+    mapping.dataElementMapping = await this.resolveDataElements(dataElements, orgUnits);
+    mapping.dataGroupMapping = await this.resolveDataGroups(dataGroups);
+    return mapping;
```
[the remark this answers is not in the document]
IgorNadj: Yeah I was thinking of just waiting until this is typescript, then can lean on that for `readonly` keyword

packages/data-broker/src/services/DataServiceResolver.js
IgorNadj: Reworked

packages/data-broker/src/services/DataServiceResolver.js
IgorNadj: Reworked

IgorNadj:
@rohan-bes I added new commits to take care of the pull-with-multiple-countries problem. We discussed extracting this out to a class like PullGrouper but it seemed simple enough to keep it inline in DataBroker.

packages/data-broker/src/DataBroker.js
```diff
@@ -303,4 +298,71 @@ export class DataBroker {
       dataServiceMapping,
     };
   }
+
+  async getPulls(dataSources, orgUnitCodes) {
+    const orgUnits = await this.models.entity.find({ code: orgUnitCodes });
+
+    // Note: each service will pull for ALL org units and ALL data sources.
```
rohan-bes, line 305: Is there a reason we can't fetch for just the countries/data elements in each mapping? We work out which countries and dataElements/groups belong to each along the way, so we'd just need to combine them when building the `uniqueMapping`?

It feels a little odd to pull for dataElements and countries that we **know** aren't in the data service (and also could create issues if they actually are in the data service but Tupaia isn't configured to fetch them from there...)

packages/data-broker/src/services/DataServiceMapping.js
```diff
@@ -70,4 +72,28 @@ export class DataServiceMapping {
     }
     return null;
   }
+
+  /**
+   * @param {DataServiceMapping} other
+   * @returns boolean
+   */
+  equals(other) {
```
rohan-bes, line 80: This warms an ex-Java dev's heart 😄 Bonus points for adding `if (this === other) return true;` at the start 👍

packages/data-broker/src/services/DataServiceResolver.js
```diff
@@ -94,50 +66,25 @@ export class DataServiceResolver {
...
-        // all data element mappings are the same, use first
-        const [deMapping] = deMappings;
-        resolved.push({
-          dataSource: dataElement,
-          service_type: deMapping.service_type,
-          config: deMapping.service_config,
-        });
-        continue;
+        throw new Error(
+          `Multiple mappings found for Data Element ${dataElement.code} for country ${countryCode}`,
```
rohan-bes, line 86: I wonder if this should be a constraint in the database? It seems like we'd never there to be multiple `dataElementDataservices` with the same `country_code` and `data_element_code`. Optional though 👍

rohan-bes (changes requested):
Requesting a change to have each `uniquePull` just pull data for the dataElements/countries that it's mapped to. But also totally happy to talk it through if this is tricky to support for some reason I'm missing.

Aside from that it's looking great!

packages/data-broker/src/DataBroker.js
IgorNadj: Discussed.

rohan-bes (approved):
Good discussion on this one @IgorNadj. Lets get the full data services refactor done and then this whole area of code will simplify itself 👍

PR #4159, opened 2022-09-09
RN-630: Add Analytics data table

rohan-bes:
### Issue RN-630:

Introducing the very first data-table, the 'analytics' data-table. This data-table pulls data from data-broker's analytics endpoint. I really wanted to break up `analytics` and `events` into separate data-tables for a couple reasons:
1. They take different parameters, which makes the config validation really tricky
2. It's currently super unclear to the user if and why they're pulling analytics or events. Now it'll be super explicit when they select which data-table to fetch from

Also the concept of an `internal` data-tables is being fleshed out here. These are data-tables that return core Tupaia data, and cannot be created/edited by end users. They still exist and entries in the database (since it'll make it easier to list out all data tables), but changes to them will be managed by the devs. I imagine these being used for all data which interacts with internal Tupaia apis (eg. data-broker, entity-api, central-api, etc.)

Also added the basic `/dataTables/:dataTableCode/fetchData` endpoint for pulling data from a data-table.

### Merge tasks:
- [x] Create route53 entries for `data-table-api`


packages/data-table-server/src/@types/express/index.d.ts
```diff
@@ -1,16 +1,18 @@
...
+import { DataTableServerModelRegistry } from '../../types';
+
 declare global {
   namespace Express {
     export interface Request {
-      ctx: Record<string, unknown>;
-    }
-
-    export interface Response {
-      ctx: Record<string, unknown>;
```
rohan-bes, line 13: I feel like we never use the `res` object, so I'm just not decorating its type here anymore

packages/data-table-server/src/app/createApp.ts
```diff
@@ -2,14 +2,24 @@
...
 
 /**
  * Set up express server with middleware,
  */
 export function createApp(database = new TupaiaDatabase()) {
-  const app = new MicroServiceApiBuilder(database, 'data-table').useBasicBearerAuth().build();
+  const app = new MicroServiceApiBuilder(database, 'data-table')
+    .attachApiClientToContext(req => new ForwardingAuthHandler(req.headers.authorization))
+    .useBasicBearerAuth()
+    .post<FetchDataTableDataRequest>(
```
rohan-bes, line 18: Decided to go with `POST` only for this endpoint. There is a rationale, here me out:
1. We have to use `POST` sometimes, since we may be passing in a large or nested query arguments, which can't be done when passing in as URL params
2. In code instances we always use `POST` for the above reason as it's safer, so the `GET` is just for developer convenience
3. Query params always come as strings and have to be parsed, whereas `POST` body is nicely formatted JSON already. Especially since data-tables will have dynamic parameters, it's hard to know which query params to parse as what type (string, string[], boolean, etc.)

packages/data-table-server/src/dataTable/AbstractDataTable.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { TupaiaApiClient } from '@tupaia/api-client';
+import { yup } from '@tupaia/utils';
+
+export abstract class AbstractDataTable<
```
rohan-bes, line 9: I know we don't usually do the whole `Abstract...` naming convention. Originally this was just called `DataTable`, but the name clashed with the model classes, which there are already a couple of, so I felt this made it a bit clearer.

packages/data-table-server/src/dataTable/AbstractDataTable.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.configSchema = configSchema;
+    this.apiClient = apiClient;
+    this.config = this.configSchema.validateSync(config);
+  }
+
+  protected validateParams(params: unknown) {
+    return this.paramsSchema.validateSync(params);
+  }
+
+  protected abstract safelyFetchData(params: yup.InferType<ParamsSchema>): Promise<RecordSchema[]>;
```
rohan-bes, line 35: Is this a dumb name? `fetchDataImpl`? 😉

packages/data-table-server/src/dataTable/AbstractDataTable.ts
```diff
@@ -0,0 +1,41 @@
...
+> {
+  protected readonly paramsSchema: ParamsSchema;
+  protected readonly configSchema: ConfigSchema;
+  protected readonly apiClient: TupaiaApiClient;
+  protected readonly config: yup.InferType<ConfigSchema>;
+
+  protected constructor(
+    paramsSchema: ParamsSchema,
+    configSchema: ConfigSchema,
+    apiClient: TupaiaApiClient,
```
rohan-bes, line 22: Passing through the `apiClient` is required for the `AnalyticsDataTable`, but it already feels like a bit of dependency pollution... I'll hold off engineering anything to handle it gracefully until we've got a few more data-table types out and we can see what's common?

packages/database/src/migrations/20220909021244-AddAnalyticsDataTable-modifies-data.js
```diff
@@ -0,0 +1,38 @@
...
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  return insertObject(db, 'data_table', {
+    id: generateId(),
+    code: 'analytics',
+    description: 'Fetches data for individual data elements',
+    type: 'internal',
+    permission_groups: '{*}',
```
rohan-bes, line 25: analytics is available to all, since data-broker will handle permissions checks to ensure users can only pull data they have access to

packages/data-table-server/src/dataTable/AbstractDataTable.ts
IgorNadj: pull -> pullData? or fetch -> fetchData? Where pull/fetch do the prep work and pullData/fetchData just fetch the data?

packages/data-table-server/src/dataTable/createDataTable.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+import { TupaiaApiClient } from '@tupaia/api-client';
+import { DataTableType } from '../models';
+import { internalDataTables } from './internal';
+
+/**
+ * Factory function for building the correct data-table for the requested dataTableObject
+ */
+export const createDataTable = (dataTableObject: DataTableType, apiClient: TupaiaApiClient) => {
```
IgorNadj, line 13: We have two concepts with the same name here. I see you've tried to differentiate by calling one `dataTableObject` with the Type suffix on the ts type. 

However, I think more consistent with the rest of the codebase would be to call this first concept a dataTable (instead of dataTableObject) with DataTable (instead of DataTableType), and the other concept something else. 

It seems pretty similar to data broker, where you can have one DhisService which can pull data for many DhisInstance objects. So I'd suggest calling the second concept here a service as well, e.g. AnalyticsDataTableService. Thoughts?

packages/data-table-server/src/dataTable/internal/AnalyticsDataTable.ts
```diff
@@ -0,0 +1,80 @@
...
+    const {
+      hierarchy,
+      dataElementCodes,
+      organisationUnitCodes,
+      startDate,
+      endDate,
+      aggregations,
+    } = params;
+
+    const aggregator = new Aggregator(new DataBroker({ services: this.apiClient }));
```
IgorNadj, line 64: No comment here just curious, do we need an Aggregator? I remember you saying something about not needing it so much?

packages/data-table-server/src/routes/FetchDataTableData.ts
```diff
@@ -0,0 +1,46 @@
...
+import { createDataTable } from '../dataTable';
+
+export type FetchDataTableDataRequest = Request<
+  { dataTableCode: string },
+  { data: unknown[] },
+  Record<string, unknown>,
+  Record<string, unknown>
+>;
+
+export class FetchDataTableDataRoute extends Route<FetchDataTableDataRequest> {
```
IgorNadj, line 18: In the data table server we can probably drop DataTable from names, it should be obvious. FetchDataRoute?

packages/data-table-server/src/app/createApp.ts
IgorNadj: Wouldn't a FetchDataTableDataRequest simply refer to the data table by code? Seems like that's the implementation of the route. So GET should be sufficient? Or are you imagining we will have some data tables in the future which will need to be heavily call-time configurable?

packages/data-table-server/src/dataTable/internal/AnalyticsDataTable.ts
rohan-bes: I'd love to bypass it ultimately, but in the short term I feel it's necessary. As it stands, we can fairly easily convert all existing reports to using data-tables if we just take their `fetch` and `aggregate` configs and pass them in to this data-table. But if we wanted to skip `Aggregator`, then we'd need to take the time to re-write each report to handle the stuff currently being done in `Aggregator` manually. It's possible, but would be a large cleanup task.

So for now lets use it, and then clean things up as we go?

packages/data-table-server/src/app/createApp.ts
IgorNadj: Discussed, may have a bunch more query params in the future, all good.

IgorNadj (approved):
Pre-approving, just some naming changes suggestions

packages/data-table-server/src/dataTable/createDataTable.ts
rohan-bes: Ah great call! I love it, much tidier. 👍

packages/data-table-server/src/routes/FetchDataTableData.ts
rohan-bes: In my head there was scope for some routes in this server that did things not relating to a specific data-table, but I think it's premature name-spacing. I like `FetchDataRoute` 👍

PR #4164, opened 2022-09-12
mSupply Setup

IgorNadj:
### Changes:

- RN-615 Create UNFPA Country
- RN-523 Create FPBS Country
- "Swap FW to FJ flag" (needed for regional map overlays)
- Register 2 new superset_instances for UNFPA project (mSupply)
- Allow superset to connect via proxy (for RN-651)
- RN-681 Admin panel for data-element-data-service mapping table
- RN-683 User permission import in admin panel

---

### Screenshots:
- See each individual card for screenshots

### Pre-release tasks
- [ ] RN-651 Add env vars for proxy
- [ ] Superset-instances: add env vars

### Post-release tasks
- [ ] Import permissions and data mapping for UNFPA - Juliana

### DIFF
--- a/packages/web-config-server/src/apiV1/measureBuilders/valueForOrgGroup.js
+++ b/packages/web-config-server/src/apiV1/measureBuilders/valueForOrgGroup.js
@@ -55,10 +55,18 @@ class ValueForOrgGroupMeasureBuilder extends DataBuilder {
       organisationUnitCode: this.entity.code,
     });
 
-    const analytics = results.map(result => ({
+    let analytics = results.map(result => ({
       ...result,
       value: result.value === undefined ? '' : result.value.toString(),
     }));
+
+    // Temp solution for RN-523
+    if (this.config.swapFwToFj) {
+      analytics = analytics.map(analytic => ({
+        ...analytic,
+        organisationUnit: analytic.organisationUnit === 'FPBS' ? 'FJ' : analytic.organisationUnit,
+      }));
+    }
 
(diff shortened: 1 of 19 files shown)


packages/web-config-server/src/apiV1/measureBuilders/valueForOrgGroup.js
```diff
@@ -55,11 +55,19 @@ class ValueForOrgGroupMeasureBuilder extends DataBuilder {
       organisationUnitCode: this.entity.code,
     });
 
-    const analytics = results.map(result => ({
+    let analytics = results.map(result => ({
       ...result,
       value: result.value === undefined ? '' : result.value.toString(),
     }));
 
+    // Temp solution for RN-523
```
IgorNadj, line 63: We decided to go for a temp solution because of time pressure + we would be able to do this easily in non-legacy visuals, and this is a throwaway legacy viz.

rohan-bes: Do we have a ticket to switch this viz to a non-legacy and remove this code? That should be a fairly simple task yeah?

rohan-bes (approved):

packages/web-config-server/src/apiV1/measureBuilders/valueForOrgGroup.js
IgorNadj: No ticket but the project is still in early stages, we need to migrate the data sources to the new data service and then after we can switch the vizes

PR #4165, opened 2022-09-12
Lesmis server minor cleanups

rohan-bes:
Just some small clean-ups I noticed needed doing when running `lesmis-server` recently


billli0 (approved):
Thanks for the clean up!

PR #4170, opened 2022-09-15
[no-issue]: Reduce flakiness in sync queue related tests

rohan-bes:
Since the change to making the ChangeHandlers run in a transaction, these sync queue related tests have all become a bit more flaky (eg. [here](https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/70d3bcf9-099a-4fcf-8b4c-2837bda2b179?component=step_Validate_and_test_Test_Test_batch_1_Test_meditrak-app-server), and [here](https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/be447781-7152-4af8-b3c7-5083f9f1d3b9?component=step_Validate_and_test_Test_Test_batch_1_Test_central-server)). They have previously been relying purely on timing at points, which isn't really all that safe. I've added a bunch of `await database.waitForAllChangeHandlers();` calls in places where they seemed necessary, which should halt the test execution until the change transaction is complete.

### DIFF
--- a/packages/central-server/src/tests/apiV2/changes/getChanges.test.js
+++ b/packages/central-server/src/tests/apiV2/changes/getChanges.test.js
@@ -19,7 +19,7 @@ describe('GET /changes/count', async () => {
 
     // Set up real sync queue for testing the /changes endpoint
     await createPermissionsBasedMeditrakSyncQueue(models.database);
-    meditrakSyncQueue.setDebounceTime(100); // Faster debounce time for tests
+    meditrakSyncQueue.setDebounceTime(50); // Faster debounce time for tests
     meditrakSyncQueue.listenForChanges();
   });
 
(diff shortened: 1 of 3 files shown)


IgorNadj (approved):
LGTM

packages/central-server/src/tests/apiV2/changes/getChanges.test.js
```diff
@@ -74,6 +75,10 @@ describe('GET /changes/count', async () => {
 
     // Add some more questions
     await oneSecondSleep();
```
IgorNadj, line 77: Can we get rid of oneSecondSleep now? :)

rohan-bes: I don't think we can 😞 The ChangeHandler won't schedule an execution until it receives a notification from the database that a change has occurred. If for some reason it takes the database longer to send a notification than it does for node to get to the next line, then calling `waitForAllChangeHandlers()` will complete instantly as no ChangeHandlers have been scheduled.

There's multiple levels of async stuff going on here 😅 It might just work though... I'm not really sure... but I don't think it'd be reliable...

PR #4175, opened 2022-09-16
RN-632: Show Data-Tables in the Admin-Panel

rohan-bes:
### Issue RN-632:

Just display data-tables for now, as we don't want users editing/deleting the internal ones. Next up is to add an interface to view the data-table data in the Admin-Panel in RN-647


packages/admin-panel/src/routes.js
```diff
@@ -121,6 +122,11 @@ export const ROUTES = [
         to: '/indicators',
         component: IndicatorsPage,
       },
+      {
+        label: 'Data-Tables',
```
rohan-bes, line 126: Do we like this naming style? With the `-`? I kinda like it as it makes clear that `Data-Tables` is a single concept, but it's not particularly convention with Tupaia to use `-`...

packages/central-server/src/apiV2/dataTables/assertDataTablePermissions.js
```diff
@@ -0,0 +1,61 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { hasBESAdminAccess } from '../../permissions';
+
+const getPermissionListWithWildcard = async accessPolicy => {
+  const userPermissionGroups = accessPolicy.getPermissionGroups();
+  return ['*', ...userPermissionGroups];
+};
```
rohan-bes, line 11: This code isn't the DRY-est... it's copy pasted from dataElements permissions stuff. It's a bit awkward to consolidate, but I could make more of an effort if the reviewer feels it's worth it

packages/admin-panel/src/routes.js
IgorNadj: I don't get it sorry 😆 
But it's just a label, - or no -, it's a + to me

packages/central-server/src/apiV2/dataTables/assertDataTablePermissions.js
IgorNadj: Sure no DRY is fine here

IgorNadj (approved):

PR #4179, opened 2022-09-19
RN-506 add year label and fix issues

billli0:
### Issue #:
RN-506

### Changes:
- Add year label ([UX design](https://linear.app/bes/issue/RN-506#comment-a7304173))
- In favourite dashboard, we want dashboard item to fetch data by default year only (no year selector). They either fetch data by year or fetch all data. To be able to do so we pass down `useYearSelector` instead of a fixed `year` from the top for generating `startDate` and `endDate`.     
- Fix [issue 4](https://linear.app/bes/issue/RN-506#comment-42fce320)


rohan-bes (approved):
It is kinda messy, but it's all gonna get cleaned up yeah? Approving, nice work!

PR #4184, opened 2022-09-21
MAUI-1269: Improve performance of Survey Response page in Admin Panel

rohan-bes:
### Issue MAUI-1269:

This page runs really slowly, and when a user navigates to it it hangs the UI for 10+ seconds.

Some basic improvements can get it loading fairly smoothly.
1. Add a debounce to the `DataFetchTable` so that it waits between inputs, currently we send an api request for every keystroke
2. Add a few indexes on the `survey_response_table`
3. Re-work how the `count()` works in the route so that it only joins the reference tables if needed

### DIFF
--- a/packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
+++ b/packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
@@ -10,6 +10,7 @@ import {
 } from './assertSurveyResponsePermissions';
 import { assertAnyPermissions, assertBESAdminAccess } from '../../permissions';
 import { assertEntityPermissions } from '../GETEntities';
+import { getQueryOptionsForColumns } from '../GETHandler/helpers';
 
 /**
  * Handles endpoints:
@@ -47,5 +48,17 @@ export class GETSurveyResponses extends GETHandler {
 
     // Apply regular permissions
     return this.getPermissionsFilter(dbConditions, options);
+  }
+
+  async countRecords(criteria) {
+    // Only join tables that we are filtering on
+    const { multiJoin } = getQueryOptionsForColumns(
+      Object.keys(criteria),
+      this.recordType,
+      this.customJoinConditions,
+      this.defaultJoinType,
+    );
+
+    return this.database.count(this.recordType, criteria, { multiJoin });
   }
 }
--- a/packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
+++ b/packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
@@ -8,7 +8,7 @@ import {
   assertSurveyResponsePermissions,
   createSurveyResponseDBFilter,
 } from './assertSurveyResponsePermissions';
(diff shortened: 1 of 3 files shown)


packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
```diff
@@ -48,4 +49,16 @@ export class GETSurveyResponses extends GETHandler {
     // Apply regular permissions
     return this.getPermissionsFilter(dbConditions, options);
   }
+
+  async countRecords(criteria) {
+    // Only join tables that we are filtering on
+    const { multiJoin } = getQueryOptionsForColumns(
```
rohan-bes, line 55: The base `countRecords()` joins all the tables required in the columns. However joining on the `entity` and `survey` tables takes the query time up to ~8 seconds (compared to < 1 second without joining).

It does mean however that this number _could_ be higher than the actual row count returned by this route, as it's possible there are survey responses with no reference entities. We could put a constraint against that in the database? Or we could just leave this as possibly saying there are more records than will appear in the Admin Panel (which isn't going to break anything)

packages/database/src/migrations/20220920062403-AddEntityIdIndexOnSurveyResponseTable-modifies-schema.js
```diff
@@ -0,0 +1,33 @@
...
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  await db.runSql(`CREATE INDEX survey_response_entity_id_idx ON survey_response (entity_id);`);
+  await db.runSql(`CREATE INDEX survey_response_outdated_id_idx ON survey_response (outdated);`);
+  await db.runSql(
+    `CREATE INDEX survey_response_data_time_idx ON survey_response (data_time DESC);`,
+  );
```
rohan-bes, line 22: We're kinda building up an 'index per column` pattern on the `survey_response` table... I don't know if that's a database anti-pattern or not... but it's hard to know how we will want to query the table...

billli0: Emm and even we have `user_id` index, it still takes some times to load in admin panel when searching by user name, that could relates to the join table though.

billli0 (approved):
Looks good! Thanks for the boost up, that makes the UI smoother.

packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
billli0: No reference entities means no entity name showing up in admin panel?

rohan-bes: No, this is just the `count` part, so it basically means that we'll count survey responses as existing even if they have no related entity 👍

billli0: Ah cool, that should be ok then.

PR #4185, opened 2022-09-21
RN-661: Prevent dead records appearing in the Meditrak Sync Queue

rohan-bes:
### Issue RN-661:

Add changes to the meditrak sync queue one by one to avoid updates being processed after deletes.

### DIFF
--- a/packages/central-server/src/database/meditrakSyncQueue/MeditrakSyncRecordUpdater.js
+++ b/packages/central-server/src/database/meditrakSyncQueue/MeditrakSyncRecordUpdater.js
@@ -15,7 +15,10 @@ export class MeditrakSyncRecordUpdater {
    * @public
    */
   async updateSyncRecords(changes) {
-    return Promise.all(changes.map(change => this.processChange(change)));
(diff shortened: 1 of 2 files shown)


packages/central-server/src/database/meditrakSyncQueue/MeditrakSyncRecordUpdater.js
```diff
@@ -15,7 +15,10 @@ export class MeditrakSyncRecordUpdater {
    * @public
    */
   async updateSyncRecords(changes) {
-    return Promise.all(changes.map(change => this.processChange(change)));
+    for (let i = 0; i < changes.length; i++) {
+      const change = changes[i];
+      await this.processChange(change);
+    }
```
rohan-bes, line 21: Just apply the changes one by one. In theory the deletes should always be in the change queue after the creates. 

Arguably a more thorough change could be made here (ie. check records exists when enqueueing an update), but historically it's been pretty stable so I feel this should suffice? Happy to be told to be less lazy by the reviewer 😉

PR #4186, opened 2022-09-21
RN-662: Support searching with underscore in the Admin Panel

rohan-bes:
### Issue RN-662:

Actually felt it might be best to make this a global change in our database layer. 

In postgres, `_` is treated as a wildcard character when doing text like comparison. I don't think we make use of the feature at all, and instead it's just confusing when we actually want to search for a text string with an underscore. There is a workaround that the user can escape the underscore using `\_` but it's not very user friendly....

Happy to hear pushback from the reviewer if they feel we're obfuscating native postgres functionality?

### DIFF
--- a/packages/database/src/TupaiaDatabase.js
+++ b/packages/database/src/TupaiaDatabase.js
@@ -53,6 +53,11 @@ const WHERE_SUBQUERY_CLAUSES = {
   NOT_EXISTS: 'notExists',
 };
 
+const COMPARATORS = {
+  LIKE: 'like',
+  ILIKE: 'ilike',
+};
+
 // no math here, just hand-tuned to be as low as possible while
 // keeping all the tests passing
 const HANDLER_DEBOUNCE_DURATION = 250;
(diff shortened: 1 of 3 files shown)


packages/database/src/TupaiaDatabase.js
```diff
@@ -555,6 +560,19 @@ function buildQuery(connection, queryConfig, where = {}, options = {}) {
   return query;
 }
 
+function sanitizeComparisonValue(comparator, comparisonValue) {
+  switch (comparator) {
+    case COMPARATORS.LIKE:
+    case COMPARATORS.ILIKE: {
+      // escape underscores, as they are treated as wildcards in postgres which we generally don't want
+      return comparisonValue.replaceAll('_', '\\_');
```
billli0, line 568:
```suggestion
      return `%${comparisonValue.replaceAll('_', '\\_'))%`;
```

Can we also add `%` to match any value contains `comparisonValue`? 
```
'abc' LIKE '%c%'     true
```

billli0 (approved):
Looks good to me! 
One idea to add `%` to support matching value contains `comparisonValue`, thinking that would be helpful?

packages/database/src/TupaiaDatabase.js
rohan-bes: Thanks for the review @billli0! On your suggestion, I think that might be trickier, as we already explicitly set the `%` in many places throughout the code. Also, I think we probably want to be able to control the ability for matching anywhere in the string vs just at the start.

But I'm happy to change it to `%${value}%` in the Admin Panel 👍 I think it would be handy

PR #4190, opened 2022-09-28
RN-665: Fix the login button in the project page project cards

rohan-bes:
### Issue RN-665:
Broken when adding the View Projects button in the UserMenu

### DIFF
--- a/packages/web-frontend/src/containers/OverlayDiv/components/LandingPage/index.js
+++ b/packages/web-frontend/src/containers/OverlayDiv/components/LandingPage/index.js
@@ -45,11 +45,11 @@ const ViewProjectsButton = styled(Button)`
 `;
 
 export const LandingPage = ({ isUserLoggedIn, isViewingProjects }) => {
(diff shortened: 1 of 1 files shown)


packages/web-frontend/src/containers/OverlayDiv/components/LandingPage/index.js
```diff
@@ -45,11 +45,11 @@ const ViewProjectsButton = styled(Button)`
 `;
 
 export const LandingPage = ({ isUserLoggedIn, isViewingProjects }) => {
-  const [isProjectsPageVisible, setIsProjectsPageVisible] = React.useState(false);
+  const [isProjectsPageVisible, setIsProjectsPageVisible] = React.useState(isViewingProjects);
```
rohan-bes, line 48: Pretty simple bug, I hadn't consider that we need to control whether we're looking at the projects page via internal state.

billli0 (approved):

PR #4191, opened 2022-09-28
RN-175: Add order columns transform (p1. switch to managing columns explicitly in report-server)

rohan-bes:
### Issue RN-175:

Part 1 of 3.

We want to introduce an ability to order the columns of the data in the viz-builder. However, before we do this, we need to actually have an explicit order to the columns in the report-server in the first place.

There are other reasons it is advantageous to switch to tracking an explicit column order as well:
- It opens opportunities for controlling how vizes are presented more easily
- It allows us to introduce excel style columns range lookups (eg. `sum(col1:col5)`)
- It makes behaviour of transforms a bit more consistent

Here we are introducing a simple data-structure called a `TransformTable`. It manages the rows of the table as before, but also separately tracks an array of columns. It has a number of utility methods for manipulating the rows/columns atomically, so that the caller doesn't need to ensure they're kept in sync manually.

Tried to make this review as digestable as possible for the reviewer by breaking up changes to each transform step into its own commit, so going commit-by-commit might be the easiest approach.

### DIFF
--- a/packages/report-server/src/__tests__/reportBuilder/transform/mergeRows.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/transform/mergeRows.test.ts
@@ -30,7 +30,7 @@ describe('mergeRows', () => {
         groupBy: 'period',
       },
     ]);
-    expect(transform(TransformTable.fromRows(SINGLE_MERGEABLE_ANALYTICS))).toEqual(
+    expect(transform(TransformTable.fromRows(SINGLE_MERGEABLE_ANALYTICS))).toStrictEqual(
       TransformTable.fromRows([{ period: '20200101', BCD1: 4, BCD2: 4, BCD3: 4 }]),
     );
   });
@@ -40,15 +40,15 @@ describe('mergeRows', () => {
       {
         transform: 'mergeRows',
         using: {
-          organisationUnit: 'exclude',
-          period: 'exclude',
+          organisationUnit: 'first',
+          period: 'first',
           '*': 'sum',
         },
       },
     ]);
-    expect(transform(TransformTable.fromRows(MERGEABLE_ANALYTICS))).toEqual(
+    expect(transform(TransformTable.fromRows(MERGEABLE_ANALYTICS))).toStrictEqual(
       TransformTable.fromRows([
-        { period: undefined, organisationUnit: undefined, BCD1: 28, BCD2: 123 },
+        { period: '20200101', organisationUnit: 'TO', BCD1: 28, BCD2: 123 },
       ]),
     );
   });
@@ -61,7 +61,7 @@ describe('mergeRows', () => {
         using: 'last',
       },
     ]);
-    expect(transform(TransformTable.fromRows(MERGEABLE_ANALYTICS))).toEqual(
+    expect(transform(TransformTable.fromRows(MERGEABLE_ANALYTICS))).toStrictEqual(
       TransformTable.fromRows([
         { period: '20200103', organisationUnit: 'TO', BCD1: 5, BCD2: 0 },
         { period: '20200103', organisationUnit: 'PG', BCD1: 2, BCD2: -1 },
@@ -77,7 +77,7 @@ describe('mergeRows', () => {
         using: 'sum',
       },
     ]);
-    expect(transform(TransformTable.fromRows(MERGEABLE_ANALYTICS))).toEqual(
+    expect(transform(TransformTable.fromRows(MERGEABLE_ANALYTICS))).toStrictEqual(
       TransformTable.fromRows([
         { period: '20200101', organisationUnit: 'TO', BCD1: 4, BCD2: 11 },
         { period: '20200102', organisationUnit: 'TO', BCD1: 2, BCD2: 1 },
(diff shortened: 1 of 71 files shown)


packages/report-server/src/__tests__/reportBuilder/transform/mergeRows.test.ts
```diff
@@ -46,7 +46,11 @@ describe('mergeRows', () => {
         },
       },
     ]);
-    expect(transform(MERGEABLE_ANALYTICS)).toEqual([{ BCD1: 28, BCD2: 123 }]);
+    expect(transform(TransformTable.fromRows(MERGEABLE_ANALYTICS))).toEqual(
+      TransformTable.fromRows([
+        { period: undefined, organisationUnit: undefined, BCD1: 28, BCD2: 123 },
+      ]),
+    );
   });
```
billli0, line 55: As we have undefined value, suggest to use `toStrictEqual` instead of `toEqual`. Difference: https://jestjs.io/docs/expect#tostrictequalvalue

Also I think we should use `toStrictEqual` in other unit tests to ensure that transforms don't add unexpected undefined value. 

As we are refactoring all unit tests in this PR, should be a good chance to switch?

packages/report-server/src/reportBuilder/transform/aliases/keyValueByFieldAliases.ts
```diff
@@ -3,37 +3,56 @@
...
+
+const keyValueByField = (table: TransformTable, field: string) => {
+  const fieldValues = table.getColumnValues(field) as string[];
+  const valueValues = table.getColumnValues('value');
+  const newColumnNames = Array.from(new Set(fieldValues));
+  const newColumns: Record<string, FieldValue[]> = Object.fromEntries(
+    newColumnNames.map(columnName => [
+      columnName,
+      // Flag all values as skip char to avoid overwriting data
+      table.hasColumn(columnName)
```
billli0, line 17: Should `table.hasColumn(columnName)` always be true? `newColumnNames` are built from `table`.

packages/report-server/src/reportBuilder/transform/aliases/keyValueByFieldAliases.ts
```diff
@@ -3,37 +3,56 @@
...
+import { TransformTable } from '../table';
+import { FieldValue } from '../../types';
+
+const keyValueByField = (table: TransformTable, field: string) => {
+  const fieldValues = table.getColumnValues(field) as string[];
+  const valueValues = table.getColumnValues('value');
+  const newColumnNames = Array.from(new Set(fieldValues));
+  const newColumns: Record<string, FieldValue[]> = Object.fromEntries(
+    newColumnNames.map(columnName => [
+      columnName,
```
billli0, line 15: Do you think it would be simpler to use the logic `get Rows -> modify Rows -> build table from Rows`?

packages/report-server/src/reportBuilder/transform/functions/updateColumns.ts
```diff
@@ -28,33 +29,54 @@ export const paramsValidator = yup.object().shape({
...
-const updateColumns = (rows: Row[], params: UpdateColumnsParams, context: Context): Row[] => {
-  const parser = new TransformParser(rows, context);
-  return rows.map(row => {
-    const returnNewRow = params.where(parser);
-    if (!returnNewRow) {
+const updateColumns = (table: TransformTable, params: UpdateColumnsParams, context: Context) => {
+  const parser = new TransformParser(table, context);
+  const newColumns: Record<string, FieldValue[]> = {};
+  const skippedRows: Record<number, Row> = {};
+  const columnNames = table.getColumns();
```
billli0, line 36: Suggest to clone table first, so the variable `table` is immuable.

packages/report-server/src/reportBuilder/transform/aliases/keyValueByFieldAliases.ts
rohan-bes: Hmmm yeah probably would be, I'll switch it it work that way 👍

packages/report-server/src/__tests__/reportBuilder/transform/mergeRows.test.ts
rohan-bes: In this specific case, I added `undefined` values more to just get the columns to appear in the correct order then to validate the data.

I think I'm generally a bit torn on using `toStrictEqual`. It would be more accurate, but it might make the tests more painful to write/maintain if there are frequent `undefined` values (which I suspect there might be...)

I'll try switching and see how things go 👍

packages/report-server/src/reportBuilder/transform/functions/updateColumns.ts
rohan-bes: `getColumns` actually returns a new array of the column names. So mutating those columns won't mutate the original table 👍

rohan-bes:
Sorry @billli0 I merged the other part PRs before noticing that you wanted to re-review this one 🤦 

I addressed the issues you mentioned in these two commits of you wanna review them?
- Make tests use strictEquals: https://github.com/beyondessential/tupaia/pull/4191/commits/9278f74769703d8c0e19e0b727e4b84c2cec0723
- Use simpler logic for keyBy aliases:  https://github.com/beyondessential/tupaia/pull/4191/commits/7c0ac43119d4857e76f268d4316bc53c819d1cc3

packages/report-server/src/__tests__/reportBuilder/transform/mergeRows.test.ts
billli0: Thanks for switching that🙏

billli0 (approved):

PR #4192, opened 2022-09-28
RN-175: Add order columns transform (p2. respect the column order in the viz-builder)

rohan-bes:
### Issue RN-175:

Part 2 of 3.

This is a fairly simple PR to switch the viz-builder to respect the column order defined during the data transformation in the report-server.

In order to be backwards compatible, we need to return data as just a simple array of objects from the report-server by default. However, we'd like the viz-builder to know the column order for this array of objects. So I added a new `output` type (`rowsAndColumns`) which returns the rows (array of objects) and the ordered list of column names. Then just a few changes in the admin-panel-server to use this new output type when previewing the data-table.

### DIFF
--- a/packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
+++ b/packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
@@ -91,6 +91,21 @@ const EditorContainer = styled.div`
   }
 `;
 
+const TABS = {
+  DATA: {
+    index: 0,
+    label: 'Data Preview',
+    previewMode: 'data',
+  },
+  CHART: {
+    index: 1,
+    label: 'Chart Preview',
+    previewMode: 'presentation',
+  },
+};
+
+const getTab = index => Object.values(TABS).find(tab => tab.index === index);
+
 const convertValueToPrimitive = val => {
   if (val === null) return val;
   switch (typeof val) {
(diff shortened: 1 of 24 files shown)


packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
```diff
@@ -141,11 +163,12 @@ export const PreviewSection = () => {
       setFetchEnabled(false);
     },
     vizType,
+    previewMode: getTab(tab).previewMode,
```
rohan-bes, line 166: This has been a bug for a while, where we haven't been passing in the `previewMode` when previewing a viz, so we only ever got the data in array of objects structure.

packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
```diff
@@ -141,11 +163,12 @@ export const PreviewSection = () => {
...
   });
-  const [tab, setTab] = useState(0);
 
   const handleChange = (event, newValue) => {
     setTab(newValue);
+    setFetchEnabled(true);
```
rohan-bes, line 171: Now we have to re-fetch when switching tabs, as the chart preview and the data preview will always fetch slightly different data.

billli0 (approved):
Looks great! Thanks for fixing up the bug and moving some frontend logic to backend. Have a thought to add a new react query for 'usePreviewConfig' so that the config and data fetching logic are separated, but I'm not very fussy about it.

PR #4193, opened 2022-09-28
RN-175: Add order columns transform (p3. order columns transform step)

rohan-bes:
### Issue RN-175:

Part 3 of 3.

The final and core PR for this feature. The actual transform.

So the idea with this transform is to allow us to re-order the columns of the table. You can either do this explicitly by specifying the column names, or by using a pre-defined sort function.

eg.
Basic explicit re-order
```
[ 
 { name: 'Bo', age: 7, car: 'Subaru' },
 { name: 'Jo', age: 12, car: 'Ferrari' }
]

transform: orderColumns
order: ['car', 'name', 'age']

=>

[ 
 { car: 'Subaru', name: 'Bo', age: 7 },
 { car: 'Ferrari', name: 'Jo', age: 12 }
]
```
Explicit re-order with wildCard
```
[ 
 { name: 'Bo', age: 7, car: 'Subaru' },
 { name: 'Jo', age: 12, car: 'Ferrari' }
]

transform: orderColumns
order: ['*', 'age']

=>

[ 
 { name: 'Bo', car: 'Subaru', age: 7 },
 { name: 'Jo', car: 'Ferrari', age: 12 }
]
```
Re-order with sortBy
```
[ 
 { name: 'Bo', age: 7, car: 'Subaru' },
 { name: 'Jo', age: 12, car: 'Ferrari' }
]

transform: orderColumns
sortBy: 'alphabetic'

=>

[ 
 { age: 7 , car: 'Subaru', name: 'Bo' },
 { age: 12 , car: 'Ferrari', name: 'Jo' }
]
```

Might also want to add a sort direction option (asc/desc)...
Also will add some unit tests

### DIFF
--- a/packages/report-server/src/reportBuilder/transform/functions/orderColumns/orderColumns.ts
+++ b/packages/report-server/src/reportBuilder/transform/functions/orderColumns/orderColumns.ts
@@ -1,0 +1,105 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { isDefined } from '@tupaia/tsutils';
+import { yup } from '@tupaia/utils';
+import { TransformTable } from '../../table';
+import { sortByFunctions } from './sortByFunctions';
+
+type OrderColumnsParams = {
+  columnOrder?: string[];
+  sorter?: (column1: string, column2: string) => number;
+};
+
+export const paramsValidator = yup.object().shape({
+  order: yup.array().of(yup.string().required()),
+  sortBy: yup
+    .mixed<keyof typeof sortByFunctions>()
+    .oneOf(Object.keys(sortByFunctions) as (keyof typeof sortByFunctions)[]),
+});
+
+const orderColumns = (table: TransformTable, params: OrderColumnsParams) => {
+  const { columnOrder = ['*'], sorter } = params;
+
+  if (sorter) {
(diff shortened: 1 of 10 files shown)


packages/report-server/src/reportBuilder/transform/functions/orderColumns/orderColumns.ts
```diff
@@ -0,0 +1,105 @@
...
+        ? indexInOrder // It's before the wildcard, just use the index
+        : indexInOrder + orderedColumns.length - 1; // It's after the wildcard, so account for all wildcard columns (minus wildcard itself)
+    newColumns[indexInNewColumns] = column;
+  });
+
+  const validatedColumns = newColumns.map(column => {
+    if (!isDefined(column)) {
+      throw new Error('Missing columns when determining new column order');
+    }
+
```
billli0, line 60: Can't really think of any example that could violate this check. Could you share one?

billli0 (approved):
`orderColumns` looks great! Also thanks for adding displayStringToMoment, it is one of the missing parts.

packages/report-server/src/reportBuilder/transform/functions/orderColumns/orderColumns.ts
rohan-bes: No I don't actually think it's possible either. But the typescript was being a bit annoying without this, and I figured it's worth checking anyway 👍

PR #4197, opened 2022-09-29
RN-674: Migrate admin-panel-server over to @tupaia/api-client

rohan-bes:
### Issue RN-674:

As part of the data-tables work we need to add another connection to the admin-panel-server, so figured we might as well do this cleanup now.

Pretty simple task in the end 🎉


packages/admin-panel-server/src/routes/UserRoute.ts
```diff
@@ -3,19 +3,18 @@
...
-  }
+const isBESAdmin = (policy: Record<string, string[]>) => {
+  return new AccessPolicy(policy).allowsSome(undefined, BES_ADMIN_PERMISSION_GROUP);
+};
 
+export class UserRoute extends Route {
   public async buildResponse() {
-    return this.centralConnection.getUser();
+    const user = await this.req.ctx.services.central.getUser();
+    return { ...camelcaseKeys(user), isBESAdmin: isBESAdmin(user.accessPolicy) };
```
rohan-bes, line 18: Had the option here to do the `camelcaseKeys` on the `CentralApi` itself, but felt it was maybe a bit of an implementation detail...

IgorNadj (approved):
LGTM

PR #4200, opened 2022-09-30
MAUI-1278: Add ability to add/remove Entity Types via the Admin Panel

rohan-bes:
### Issue MAUI-1278:

For too long, devs have had to run a migration and update the code to add a value to the `entity_type` enum. Now we can finally dust our hands of such dirty business 😉 

Honestly though this was a tad trickier than expected, and leaves a bit of complexity in the code, so would appreciate a considered review.

Listing/Adding entity types is fairly easy using simple the postgres commands. Deleting on the other hand is not so simple. To delete a value from an enum, you need to rename the old enum, recreate a new one, and switch it over where it's being used. You can't even do this inside a transaction 😭 So it feels pretty hazardous. 

According to the [postgres docs](https://www.postgresql.org/docs/current/datatype-enum.html), it seems best to leave values in an enum once they've been entered. So the hacky (or clever 😉) idea I came up with is that instead of deleting them, we rename them to `DELETED_{random uuid}`, and then when listing the types we filter out the ones prefixed with `DELETED_`. Thoughts on this? Does seem kind of hacky, but outside of switching to reference data (or dropping the constraint entirely), I can't see a more viable way of supporting user defined entity_types (which we do really want).

There turned out to be an additional complication, which is that we have a bunch of hardcoded references to the old `ENTITY_TYPES` object throughout the code. I cleaned up the ones that refer to the values dynamically (for input validation mostly) and left the existing references to specific entity_types. However for safety's sake I prevented renaming any of the hard coded entity_types 👍


packages/database/src/modelClasses/Entity.js
```diff
@@ -496,4 +505,69 @@ export class EntityModel extends MaterializedViewLogDatabaseModel {
...
+      `SELECT UNNEST(enum_range(null::${ENTITY_TYPE_TYPE})) as type;`,
+    );
+    return entityTypes
+      .map(({ type }) => type)
+      .filter(type => !type.startsWith(DELETED_ENTITY_TYPE_PREFIX));
+  }
+
+  async addEntityType(name) {
+    // Have to manually format SQL here, as postgres doesn't support parameters on enum types
+    // Using pg-format to avoid SQL injection
```
rohan-bes, line 520: This one was a little tricky to deal with 😉 As described [here](https://github.com/brianc/node-postgres/issues/2259) postgres doesn't support parameter substitution for enum types. So went with the recommendation from others to use a 3rd party SQL injection parser.

packages/central-server/src/apiV2/entityTypes/GetEntityTypes.js
```diff
@@ -0,0 +1,55 @@
...
+    const entityTypes = await this.getEntityTypes();
+    const entityType = entityTypes.find(type => type === recordId);
+    if (!entityType) {
+      return undefined;
+    }
+
+    return { id: entityType, type: entityType };
+  }
+
+  async findRecords(criteria, options) {
```
rohan-bes, line 39: Chose to not support some of the basic behaviour (filtering, sorting, etc.) since the code around it is all so database heavy. However, chose to still support pagination as that isn't easy to disable in the Admin Panel

IgorNadj:
It does seem kind of odd to have an enum be user editable. To me enums are more like state values e.g. APPROVED/NOT_APPROVED etc. What was the downside to moving this to a reference table with FK?

rohan-bes:
@IgorNadj that's a fair point, it does also feel kind of weird to me too...

Switching to reference data does have some better points, however it's also a pretty sizeable refactor I think... Due to the blended nature of our ORM model, there are still plenty of places where we would have to refactor SQL and code to support reference data for the `entity.type`. It would definitely blow out the scope of this ticket a bunch, and imo probably wouldn't justify it.

I'd argue if we're opposed to users being able to edit the `entity_type` enum directly, then we keep some of this PR (the stuff that allows `entity_types` to be dynamic rather than hard coded), and then instead just make it an OTS workflow for devs/data-engineers to add new ones (so that we don't have to do a migration each time)

Thoughts?

IgorNadj:
> I'd argue if we're opposed to users being able to edit the `entity_type` enum directly, then we keep some of this PR (the stuff that allows `entity_types` to be dynamic rather than hard coded), and then instead just make it an OTS workflow for devs/data-engineers to add new ones (so that we don't have to do a migration each time)

Yeah I like that as a first step, for sure. And then yeah we can move it to a table in a separate PR in the future. I also like the idea of having it be project based (entity_type=(id,code,project_id)), so we could probably get rid of canonical_types in project at the same time.

rohan-bes:
@IgorNadj I've reverted the ability for users to create/edit/delete entity types.

Keeping the page that lists them in the Admin Panel for convenience.

IgorNadj (changes requested):

packages/database/src/ModelRegistry.js
```diff
@@ -49,6 +49,7 @@ export class ModelRegistry {
     Object.keys(this.modelClasses).forEach(modelName => {
       const modelKey = getModelKey(modelName);
       Object.keys(this.modelClasses).forEach(otherModelName => {
+        this[modelKey].modelRegistry = this; // Add a reference to the registry
```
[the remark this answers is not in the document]
rohan-bes: Ah nice catch, yep will remove 👍

IgorNadj (approved):
LGTM!

PR #4204, opened 2022-10-03
Finalise push changes route

billli0:
### Changes:
Combined route `surveyResponse` and `postChange` in `central-server`.
Will add test for route `surveyResponse` if changes look good.


packages/meditrak-app-server/src/routes/sync/PushChangesRoute/PushChangesRoute.ts
```diff
@@ -35,27 +43,38 @@ export class PushChangesRoute extends Route<PushChangesRequest> {
...
         surveyImages.push(validatedPayload);
       }
     }
 
     // Submit survey responses
-    await this.req.ctx.services.central.createSurveyResponses(surveyResponses);
+    await this.req.ctx.services.central.upsertSurveyResponses(surveyResponses);
+    if (waitForAnalyticsRebuild) {
+      const { database } = this.req.models;
+      await AnalyticsRefresher.refreshAnalytics(database);
```
billli0, line 76: Is it fine to run refreshAnalytics() after all survey responses have been upserted? Before we do it for each survey response.

packages/central-server/src/apiV2/surveyResponses/saveResponsesToDatabase.js
```diff
@@ -0,0 +1,139 @@
...
+    question_id: yup.string().required,
+  }),
+);
+
+async function saveAnswerRecords(models, rawAnswerRecords, surveyResponseId, questionsByCode) {
+  let answers;
+  /** The format of answers between upserting and creating are different.
+   *  create answer records: [ { [questionCode]: [answerValue] } ]
+   *  upsert answer records: [ UpsertAnswersSchema ]
+   */
```
billli0, line 35: Combining upsert and create survey response logic is great, but it is a complicated when it comes to different formats of answer records...

packages/api-client/src/connections/CentralApi.ts
```diff
@@ -6,14 +6,20 @@
...
+  id?: string;
+  survey_id: string;
+  entity_id?: string;
+  entity_code?: string;
+  timestamp?: string;
+  data_time?: string;
+  start_time?: string;
+  end_time?: string;
+  approval_status?: string;
+  answers: Answers | Answers[];
```
billli0, line 18: Can we remove optional configs?

packages/meditrak-app-server/src/routes/sync/PushChangesRoute/PushChangesRoute.ts
rohan-bes: Hmmm can we actually remove this functionality? It's only used by PSSS, which won't be using this route.

Later on when we clean up the original route in central-server, we will need to decide what to do with PSSS...

packages/meditrak-app-server/src/routes/sync/PushChangesRoute/translateInboundSurveyResponse.ts
```diff
@@ -46,15 +46,16 @@ const constructSurveyResponseTranslators = (models: MeditrakAppServerModelRegist
   user_email: (userEmail: string) => translateUserEmailToIdAndAssessorName(models.user, userEmail),
   entity_code: (entityCode: string) => translateEntityCodeToId(models.entity, entityCode),
   survey_code: (surveyCode: string) => translateSurveyCodeToId(models.survey, surveyCode),
-  answers: (answers: { body: string }[]) => ({ answers: answers.filter(a => a.body !== '') }), // remove any empty answers
+  answers: (answers: { body: string }[]): Promise<SurveyResponseObject> =>
+    new Promise(() => ({ answers: answers.filter(a => a.body !== '') })), // remove any empty answers
```
rohan-bes, line 50:
```suggestion
  answers: async (answers: { body: string }[]) =>({ answers: answers.filter(a => a.body !== '') })), // remove any empty answers
```

packages/api-client/src/connections/CentralApi.ts
rohan-bes: As in flag some of these fields as non-optional? Yup, if they're required we may as well mark them as such 👍

packages/central-server/src/apiV2/surveyResponses/saveResponsesToDatabase.js
```diff
@@ -66,7 +66,10 @@ function buildResponseRecord(user, entitiesByCode, body) {
     approval_status: approvalStatus,
   } = body;
 
-  const timezoneName = getTimezoneNameFromTimestamp(timestamp);
+  const additionalConfigs = {};
+  if (timestamp) {
+    additionalConfigs.timezone = getTimezoneNameFromTimestamp(timestamp);
+  }
```
rohan-bes, line 72: Is this change so that it works even if we can't get the timezone name from the timestamp?

packages/meditrak-app-server/src/routes/sync/PushChangesRoute/types.ts
```diff
@@ -1 +1 @@
-export type SurveyResponseObject = Record<string, unknown>;
+export type SurveyResponseObject = Record<string, any>;
```
rohan-bes, line 1: Would love if we could still get things working with `unknown` rather than `any`. `any` is kind of a bad practice in typescript as it allows for skipping validation entirely... Happy to do a bit of a pair on it if you want?

Also, quite OCD, but can you add the copyright to the top of the file 😅

rohan-bes (approved):
Looks great! Thanks @billli0 

A couple of more minor comments, happy to chat about them.

Once this PR is merged, could you put up a PR for the original `rn-556-finish-push-changes-route` branch and I'll review that?

packages/central-server/src/apiV2/surveyResponses/saveResponsesToDatabase.js
billli0: Indeed, `timezone` needs to be passed in from meditrak-app. Thanks for catching that.

PR #4207, opened 2022-10-05
RN-592: Add Multiline To FreeText Question Field

chris-pollard:
### Changes:

- Add multiline === true as prop to the FreeTextQuestion component.


rohan-bes (approved):
Nice one @chris-pollard

PR #4212, opened 2022-10-10
RN-645: Add internal data tables for fetching entities and entity relations

rohan-bes:
### Issue RN-645:

We're adding 2 new data tables in the PR.
`entities` - fetches metadata for entities (useful for getting entity names from codes for instance)
`entity_relations` - fetches relations within an entity hierarchy (useful for entity aggregation)

Also adding the `Parameters` route for fetching the parameters required by a given data-table. It's not entirely needed, but there's been some rebasing shenanigans going on and I'd rather not have to create a separate PR for it 👍

### DIFF
--- a/packages/data-table-server/src/__tests__/dataTableService/internal/AnalyticsDataTableService.test.ts
+++ b/packages/data-table-server/src/__tests__/dataTableService/internal/AnalyticsDataTableService.test.ts
@@ -3,9 +3,12 @@
  * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
  */
 
+import MockDate from 'mockdate';
 import { AccessPolicy } from '@tupaia/access-policy';
 import { TupaiaApiClient } from '@tupaia/api-client';
 import { DataTableServiceBuilder } from '../../../dataTableService';
+
+const CURRENT_DATE_STUB = '2020-12-31';
 
 const TEST_ANALYTICS = [
   { period: '2020-01-01', organisationUnit: 'TO', dataElement: 'PSSS_AFR_Cases', value: 7 },
@@ -21,7 +24,7 @@ const fetchFakeAnalytics = (
   {
     organisationUnitCodes,
     startDate = '2020-01-01',
-    endDate = '2020-12-31',
+    endDate = CURRENT_DATE_STUB,
   }: { organisationUnitCodes: string[]; startDate?: string; endDate?: string },
 ) => {
   return {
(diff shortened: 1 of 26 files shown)


packages/database/src/migrations/20221006004520-AddEntityRelationsDataTable-modifies-data.js
```diff
@@ -0,0 +1,38 @@
...
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  return insertObject(db, 'data_table', {
+    id: generateId(),
+    code: 'entity_relations',
+    description: 'Fetches data for entity relations',
+    type: 'internal',
+    permission_groups: '{*}',
```
rohan-bes, line 25: Safe to have public permissions for these data-tables, since the entity-api does permissions checking itself

packages/data-table-server/src/__tests__/dataTableService/internal/EntityRelationsDataTableService.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        config: { innerType: { required: true, type: 'string' }, required: true, type: 'array' },
+        name: 'entityCodes',
+      },
+      { config: { type: 'string', required: true }, name: 'ancestorType' },
+      { config: { type: 'string', required: true }, name: 'descendantType' },
+    ]);
+  });
+
+  describe('fetchData', () => {
+    // TODO: Implement these tests when RN-685 is done
```
rohan-bes, line 64: These tests really would benefit from a mock entity-api to work nicely. That currently sits in the `report-server`, but as of RN-685 it's moving to `api-client` at which point it'll be easy to write these tests

packages/data-table-server/src/dataTableService/internal/EntitiesDataTableService.ts
```diff
@@ -0,0 +1,85 @@
...
+    includeDescendants: boolean;
+  }) {
+    const { hierarchy, entityCodes, filter, fields, includeDescendants } = params;
+
+    const entities = await this.apiClient.entity.getEntities(hierarchy, entityCodes, {
+      fields,
+      filter,
+    });
+
+    if (!includeDescendants) {
```
rohan-bes, line 49: Thoughts on this `includeDescendants` flag? I wanted there to be an easy way for users to pull all the entities in a hierarchy, but you could argue we should make a separate `entity_descendants` data-table for that?

billli0 (approved):
Code looks very clean!

packages/data-table-server/src/dataTableService/internal/EntitiesDataTableService.ts
rohan-bes: I think we definitely want to sometimes not pull descendants. For instance, if we have a list of entity codes and we want to get their names, we want to use this data-table to fetch the names from the codes.

But we can always change this later if we don't like how it behaves 👍

packages/data-table-server/src/__tests__/dataTableService/internal/AnalyticsDataTableService.test.ts
```diff
@@ -120,53 +123,76 @@ describe('AnalyticsDataTableService', () => {
     });
   });
 
-  it('can fetch data from Aggregator.fetchAnalytics()', async () => {
+  it('getParameters', () => {
     const analyticsDataTableService = createDataTableService(
       analyticsDataTable,
       {} as TupaiaApiClient,
     );
```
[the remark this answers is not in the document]
rohan-bes: Good call, have cleaned this up since rebasing off `dev` 👍

PR #4213, opened 2022-10-10
RN-685: Add MockTupaiaApiClient for use in unit tests

rohan-bes:
### Issue RN-685:

As more and more packages switch to using the `TupaiaApiClient` for interacting with external packages, it would be nice if we could provide mocks of those services so that it makes writing unit tests for code that depends on those services easier.

We've actually been sprouting out little mocks here and there in various packages already, so this PR is to create the `MockTupaiaApiClient` and then move the existing mocks over to it. I think the main one is in `report-server` for the `entity-api`


packages/api-client/src/__tests__/mockConnections/MockAuthApi.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
rohan-bes, line 1: Had to add this to the top of all the mock apis since they don't implement many of the methods

packages/api-client/src/__tests__/mockConnections/MockEntityApi.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    const { fields, filter } = queryOptions;
+
+    const filteredDescendants = filter
+      ? descendantEntities.filter(entity => entity.type === filter.type)
+      : descendantEntities;
+
+    return fields ? filteredDescendants.map(e => pick(e, fields)) : filteredDescendants;
+  }
```
rohan-bes, line 47: All this logic is copied from `mockEntityApi` in `report-server`

packages/api-client/src/__tests__/mockConnections/MockTupaiaApiClient.ts
```diff
@@ -0,0 +1,34 @@
...
+  public readonly central: CentralApiInterface;
+  public readonly auth: AuthApiInterface;
+  public readonly report: ReportApiInterface;
+
+  public constructor({
+    auth = new MockAuthApi(),
+    central = new MockCentralApi(),
+    entity = new MockEntityApi(),
+    report = new MockReportApi(),
+  }) {
```
rohan-bes, line 28: Clients can override these mock instances with their own custom implementations if they wish

packages/api-client/src/connections/AuthApi.ts
```diff
@@ -52,3 +53,5 @@ export class AuthApi extends BaseApi {
     return { accessToken, refreshToken, accessPolicy, email, user };
   }
 }
+
+export interface AuthApiInterface extends PublicInterface<AuthApi> {}
```
rohan-bes, line 57: So I appreciate what I'm doing here is a little bit unfashionable. 
Typically in this scenario you:
1. Define an interface with the public methods you want it to have
2. Define a concrete implementation of the interface
3. Define a mock implementation of the interface

The downside to that pattern is that you have to remember to update the interface whenever you want to add methods. It's also a bit more boilerplate-y.

Instead what we're doing here is inferring an interface from the public methods of the concreate implementation, which means we can just freely add methods to the concrete implementation and the interface will automatically update. But if this feels too awkward then I'm happy to switch to the more standard way of doing things 👍

packages/api-client/src/connections/types.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+// Using a mapped type will strip out private properties of a class
+export type PublicInterface<T> = {
```
rohan-bes, line 7: Was a little surprised this was required, but it does sort of make sense.

packages/report-server/src/__tests__/reportBuilder/context/buildContext.test.ts
```diff
@@ -44,14 +43,12 @@ describe('buildContext', () => {
...
-
   const reqContext: ReqContext = {
     hierarchy: HIERARCHY,
     permissionGroup: 'Public',
-    services: {
-      entity: apiMock,
-    } as ReqContext['services'],
+    services: new MockTupaiaApiClient({
+      entity: new MockEntityApi(ENTITIES, RELATIONS),
+    }),
```
rohan-bes, line 51: I'm pretty happy with how elegant this pattern turned out 😄 and not having to do explicit type casting anymore is a nice bonus imo!

packages/api-client/src/index.ts
```diff
@@ -13,3 +13,5 @@ export { AuthHandler } from './types';
 export * from './auth';
 
 export * from './constants';
+
+export * from './__tests__/mockConnections';
```
rohan-bes, line 17: Tried a little bit to see if it's possible to [just export these when in `development` mode](https://webpack.js.org/guides/package-exports/#optimizations). But it was proving a bit tricky with how we resolve local packages in yarn. If someone knows how to do this I think it'd be nice, but seems not a big deal to me?

packages/api-client/src/connections/AuthApi.ts
IgorNadj: No this is great, I like it!

packages/api-client/src/index.ts
IgorNadj: We shouldn't ever be putting __tests__ into the dist folder 😄 

MockApiClient isn't actually a test, it's a utility that can be used for testing, like a dry run flag in a lib to use an analogy, it's a real part of the lib.

IgorNadj (approved):
Looks great, just one comment for where MockApiClient ends up, pre-approving

packages/api-client/src/index.ts
rohan-bes: Oh so you think I should move it into somewhere else in the `src` directory? I guess I had felt since they're just for testing they should be under the `__tests__` directory.

But I could see them fitting under `src/connections/mocks`, seems appropriate 👍

PR #4214, opened 2022-10-11
RN-628: Convert 'fetch' to be a viz-builder transform step

rohan-bes:
### Issue RN-628:

Pretty major overhaul here of how we do data-fetching in the viz-builder.

So currently, the 'fetch' step of the viz-builder is its own process. We perform a fetch, then go on to transform the data from that fetch. With the new system, we will perform fetches as part of the transformation process. This allows us to do the following:
1. Perform multiple independent fetches
2. Use results from a previous fetch/transformation process to control a later fetch (eg. fetch for entities that had results in the first fetch)
3. Fetch from multiple different sources and combine the data (once data-tables comes in this will be very common)

These are really powerful capabilities, which will allow us to build vizes more easily, and effectively.

It also allows us to remove some of the existing functionality in the future (ie. aggregator, context dependent functions in report-server) which will really help simplify things.

#### Example
old:
```
{
  fetch: {
    dataElements: ['BCD1']
  },
  transform: ["keyValueByDataElement"]
} 
```
new:
```
{
  transform: [
    {
      transform: "fetchData",
      parameters: {
        dataElements: ['BCD1']
      }
    },
    "keyValueByDataElement"
  ]
} 
```

#### How to review:

This PR roughly breaks up into 3 parts, and if the review prefers I'm happy to break them out explicitly, but going commit by commit also works.
1. https://github.com/beyondessential/tupaia/pull/4214/commits/f6342ae3da8fa12cd8765b2af3c6da9cb96f3a46 - Convert the 'fetch' logic to be a transform step called 'fetchData'. Most of the logic changes are here
2. https://github.com/beyondessential/tupaia/pull/4214/commits/847cfa511bedece1d9d85b172b980acdfc61ad0a - Rework the viz-builder UI. Mainly just deleting lots of code we no longer need
3. https://github.com/beyondessential/tupaia/pull/4214/commits/d3388b196c086e2b3819e04b0072662f51fd6503 - Migrations to convert existing reports to the new format. Turned out to be simpler than anticipated

Aside from that it's just adding/removing/re-writing tests, which makes up the bulk of the deltas of this PR, so please don't be too scared 🙏


packages/admin-panel-server/src/routes/index.ts
```diff
@@ -8,7 +8,6 @@ export * from './dashboardVisualisations';
 export * from './mapOverlayVisualisations';
 export * from './FetchHierarchyEntitiesRoute';
 export * from './FetchReportPreviewDataRoute';
-export * from './FetchAggregationOptionsRoute';
```
rohan-bes, line 11: Removed all the 'FetchAggregationOptions' stuff as there's no longer a UI for the fetch config. This is a bit of a step backwards in terms of viz-builder functionality, which is sad, but I think things will return better than before soon enough when we convert the transform steps to having a complete UI 👍

packages/admin-panel-server/src/viz-builder/utils.ts
```diff
@@ -12,7 +12,6 @@ export const extractDataFromReport = (report: Report) => {
     return { customReport: config.customReport };
   }
 
-  const { fetch, transform } = config;
-  const { aggregations, ...restOfFetch } = fetch;
-  return { fetch: restOfFetch, aggregate: aggregations, transform };
```
rohan-bes, line 17: Really glad that some of these messy abstractions are being un-done here 👍

packages/database/src/migrations/20221005045601-MoveFetchIntoTransformInReports-modifies-data.js
```diff
@@ -0,0 +1,76 @@
...
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+const convertToNewConfig = config => {
+  const { fetch, transform, ...restOfConfig } = config;
+  if (!fetch) {
+    return config;
```
rohan-bes, line 20: Some reports are no 'customReports' that don't have a `fetch` or `transform` step, so just ignore those

packages/expression-parser/src/expression-parser/ExpressionParser.js
```diff
@@ -29,14 +29,17 @@ import { customNamespaces } from './customNamespaces';
  */
 
 const ADDITIONAL_ALPHA_CHARS = ['@'];
+const CUSTOM_CONFIG = {
+  matrix: 'Array', // Have mathjs treat array-like objects as plain arrays rather than mathjs Matrix
+};
```
rohan-bes, line 34: I've been looking for this option for **AGES!** Basically by default `mathjs` will treat arrays as `Matrix` type, eg.
` typeof ([1, 2]) => Matrix`

This makes things really painful when combining with our own custom functions since we constantly have to cast matrixes to arrays. Now it will just treat matrixes as arrays automatically!

packages/report-server/src/reportBuilder/transform/functions/fetchData/query/buildOrganisationUnitParams.ts
```diff
@@ -0,0 +1,66 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
```
rohan-bes, line 4: Mostly copied from the previous `packages/report-server/src/reportBuilder/query/buildOrganisationUnitParams.ts`

packages/report-server/src/reportBuilder/transform/functions/fetchData/query/QueryBuilder.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
```
rohan-bes, line 4: Mostly copied from the previous `packages/report-server/src/reportBuilder/query/QueryBuilder.ts`

packages/report-server/src/reportBuilder/reportBuilder.ts
```diff
@@ -52,22 +46,18 @@ export class ReportBuilder {
...
 
-    const context = await buildContext(this.config.transform, this.reqContext, data, query);
+    const context = await buildContext(this.config.transform, this.reqContext);
     const transform = buildTransform(this.config.transform, context);
-    const transformedData = transform(data.results);
+    const transformedData = await transform(data);
 
-    const outputContext = { ...this.config.fetch };
-    const output = buildOutput(this.config.output, outputContext, aggregator);
+    const outputContext = {}; // TODO: Fix output context
```
rohan-bes, line 59: Working on this now!

packages/report-server/src/__tests__/reportBuilder/context/buildContext.test.ts
```diff
@@ -157,11 +168,14 @@ describe('buildContext', () => {
       ];
       const data = { results: analytics };
 
-      const context = await buildContext(transform, reqContext, data);
-      const expectedContext = {
+      const context = await buildContext(transform, reqContext);
+      expect(context).toEqual(expect.objectContaining({ dependencies: ['orgUnits'] }));
```
billli0, line 172: I like the idea of adding the dependency keys into context!

packages/report-server/src/__tests__/reportBuilder/transform/fetchData/fetchData.test.ts
```diff
@@ -0,0 +1,523 @@
...
+      const rows = [{ date: '2018-01-01' }, { date: '2020-02-03' }];
+      const transform = buildTestTransform(
+        [
+          {
+            transform: 'fetchData',
+            parameters: {
+              dataElements,
+              organisationUnits,
+              startDate,
+              endDate,
```
billli0, line 191: We are finally having first step forward to be able to manage request query!

billli0 (approved):
These are massive pieces of great work @rohan-bes ! I followed the commit to review and it helps a lot. 

And we finally can manage request query! So excited!

packages/report-server/src/reportBuilder/reportBuilder.ts
billli0: Ah would be nice to finish this up.

billli0 (changes requested):
Request to add output context.

rohan-bes:
@billli0 I think that TODO was from an earlier commit, the output context stuff should be working 👍

packages/report-server/src/reportBuilder/transform/functions/fetchData/query/QueryBuilder.ts
```diff
@@ -0,0 +1,45 @@
...
+  private parseConfig() {
+    const parsedConfig = { ...this.config };
+    Object.entries(parsedConfig).forEach(([field, value]) => {
+      if (typeof value === 'string') {
+        parsedConfig[field as keyof FetchConfig] = this.parser.evaluate(value);
+      } else {
+        parsedConfig[field as keyof FetchConfig] = value as any;
+      }
+    });
+    return parsedConfig as ParsedFetchConfig;
```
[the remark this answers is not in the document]
rohan-bes: Hmmm it's not a bad idea. But I worry it might be hard to avoid it being too strict... Maybe lets play with it in practice and if it turns out users could do with more error handling for incorrect data formats we can add a validation check here?

billli0 (approved):

PR #4215, opened 2022-10-11
RN-677 Fix DHIS Server validation

IgorNadj:
RN-677

### DIFF
--- a/packages/central-server/src/dataAccessors/editSurvey.js
+++ b/packages/central-server/src/dataAccessors/editSurvey.js
@@ -111,6 +111,9 @@ class SurveyEditor {
         serviceType:
           this.updatedFieldsByResource[RESOURCE_TYPES.DATA_GROUP].service_type ||
           this.dataGroup.service_type,
+        dhisInstanceCode:
+          this.updatedFieldsByResource[RESOURCE_TYPES.DATA_GROUP]?.config?.dhisInstanceCode ||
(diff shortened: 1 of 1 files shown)


packages/central-server/src/dataAccessors/editSurvey.js
```diff
@@ -111,6 +111,9 @@ class SurveyEditor {
         serviceType:
           this.updatedFieldsByResource[RESOURCE_TYPES.DATA_GROUP].service_type ||
           this.dataGroup.service_type,
+        dhisInstanceCode:
+          this.updatedFieldsByResource[RESOURCE_TYPES.DATA_GROUP]?.config?.dhisInstanceCode ||
+          this.dataGroup.config?.dhisInstanceCode,
```
rohan-bes, line 116: I feel like there can be a lot of boilerplate around some of this code huh? It's really easy to miss out on things. Eg. the fact you need to add an entry in the `newRecordValidationRules` to add a create endpoint for a resource...

Would be a nice thing for us to cleanup some day

rohan-bes (approved):
LGTM

packages/central-server/src/dataAccessors/editSurvey.js
IgorNadj: Absolutely, it's all 🍝

PR #4218, opened 2022-10-13
RN-363 Map Overlay Group Relation Id to Code

chris-pollard:
### Changes:

- Update admin panel front end to view child code and to edit and create using code.
- Removed type and id from editing and creation due to dependent child code field. Added grouping in Autocomplete component.
- Updated Create, GET and Update api functions in the central server to return child code based on child_id.
---

### Screenshots:
Screen Shot 2022-10-13 at 11 39 28 am


packages/ui-components/src/components/Inputs/Autocomplete.js
```diff
@@ -68,6 +69,7 @@ export const Autocomplete = ({
     value={value}
     disabled={disabled}
     onChange={onChange}
+    groupBy={option => option[optionGroupKey]}
```
chris-pollard, line 72: Added 'groupBy' prop input. If 'optionGroupKey' is undefined, the drop down will not be grouped.

packages/admin-panel/src/editor/Editor.js
```diff
@@ -35,9 +35,9 @@ export const Editor = ({ fields, recordData, onEditField }) => {
     <div>
       {fields
         .filter(({ show = true, editConfig = {} }) => {
-          const { visibilityCriteria } = editConfig;
+          const { visibilityCriteria, hideForEdit } = editConfig;
 
-          if (!show) {
+          if (!show || hideForEdit) {
```
chris-pollard, line 40: Added condition for a new property in 'editConfig' object. As 'show' property applies across the whole field object, 'hideForEdit' can only apply for the edit modal. Thus we can show the 'child_id' in the admin panel table, but not in the edit modal.

packages/central-server/src/apiV2/GETMapOverlayGroupRelationChildren.js
```diff
@@ -0,0 +1,86 @@
...
+      .map(o => {
+        return { childId: o.id, childCode: o.code, childType: 'Map Overlay' };
+      })
+      .concat(
+        mapOverlayGroupRecords
+          .filter(o => o.code !== 'Root') // exclude Root group from children list
+          .map(o => {
+            return { childId: o.id, childCode: o.code, childType: 'Map Overlay Group' };
+          }),
+      );
```
chris-pollard, line 44: I created this new endpoint due to the fact that all map overlay group relation child options are not easily retrieved with mapOverlayGroupRelation, as they are sourced from two separate tables and cannot be joined with map_overlay_group_relation, since not all children options would be already recorded in map_overlay_group_relation. I'm sure there are probably more logical strategies for this, but this is what I've come up with so far.

rohan-bes (changes requested):
Great stuff here @chris-pollard! Love all the unit tests and storybook stuff.

I have a feeling we might wanna take a look at how we're handling the map overlay relation joins here. Part of me feels that we should either:
- Switch to using `codes` all the way down to the DB level
- Keep using `ids` but do the joins at a lower level than the `route`

I'm a bit wary of putting too much logic into the CRUD routes in central-server as usually they're quite simple, and I think it makes things easier to manage that way.

Wanna pair on this one? My thoughts aren't fully clarified so I wouldn't mind talking things through 😄

PR #4221, opened 2022-10-16
RN-692: Apply permissions checking on Analytics and Events data-tables

rohan-bes:
### Issue RN-692:

Bit of a whoopsie by me, I forgot to pass the `accessPolicy` in to data-broker when instantiating it in the `analytics` and `events` data-tables. This means that currently it's not doing any permissions checking.

Also took the opportunity here to convert `createDataTableService()` into the `DataTableServiceBuilder`. I'm pretty happy with this refactor, as there had been a growing issue where different data-table services required different dependencies (eg. `AnalyticsDataTableService` requires `apiClient` and `accessPolicy`, `SqlDataTableService` requires `models` etc.) The creator function was starting to get rather awkward, and each data-table service was required to take the dependencies of all others in their constructor which was pretty ugly. 

Now each data-table service can just specify which dependencies it needs and then the `DataTableServiceBuilder` will work out the rest. 🎉


packages/data-table-server/src/dataTableService/DataTableService.ts
```diff
@@ -3,28 +3,28 @@
...
-import { TupaiaApiClient } from '@tupaia/api-client';
 import { yup } from '@tupaia/utils';
 
 export abstract class DataTableService<
+  Context extends Record<string, unknown> = Record<string, unknown>,
   ParamsSchema extends yup.AnySchema = yup.AnySchema,
   ConfigSchema extends yup.AnySchema = yup.AnySchema,
   RecordSchema = unknown
 > {
+  protected readonly ctx: Context;
```
rohan-bes, line 14: This always surprises me, and seems like a bug, but eslint complains if you call a field of a class `context` with the following [error](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/static-property-placement.md).

I guess I don't mind too much, but it is always a little surprising. Perhaps we look into this? Perhaps we choose a different term other than `context` to refer to arbitrary state?

packages/data-table-server/src/dataTableService/internal/EventsDataTableService.ts
```diff
@@ -77,7 +84,12 @@ export class EventsDataTableService extends DataTableService<
       aggregations,
     } = params;
 
-    const aggregator = new Aggregator(new DataBroker({ services: this.apiClient }));
+    const aggregator = new Aggregator(
+      new DataBroker({
+        services: this.ctx.apiClient,
+        accessPolicy: this.ctx.accessPolicy,
```
rohan-bes, line 90: Passing in the `accessPolicy` is the only real change here, everything else is just refactoring

chris-pollard (approved):
LGTM and nice tests! Only question I have is around the naming of 'DataTableServiceBuilderForService'. Took me a little reading to understand why the name, I think it's a little confusing even though it describes it accurately. Maybe 'DataTableServiceBuilderForInternalService'? hmm... not much better. Or since the class is not exported, what about 'InternalServiceBuilder'?

rohan-bes:
> Only question I have is around the naming of 'DataTableServiceBuilderForService'. Took me a little reading to understand why the name, I think it's a little confusing even though it describes it accurately. Maybe 'DataTableServiceBuilderForInternalService'? hmm... not much better. Or since the class is not exported, what about 'InternalServiceBuilder'?

Good point @chris-pollard, I also felt the name was pretty awkward. Actually it's not just for internal data-tables, but any data-table service (there are only internal data-table services in existence right now so I can see how that's a bit confusing).

I'll have a bit of a brainstorm on a better name... `DataTableServiceBuilderForType`? But yeah the goal was to keep it as an internal class so that clients don't actually interact with it directly.

PR #4224, opened 2022-10-17
[no-issue]: Fix broken powered by Tupaia link in LESMIS

rohan-bes:
www.info.tupaia.org doesn't work anymore


billli0 (approved):
Thanks for the fix 🙏

PR #4226, opened 2022-10-17
[no-issue]: Fix upserting dynamic columns

rohan-bes:
### Issue https://beyondessential.slack.com/archives/C01MAA3NKM1/p1665977231627499:

This issue is a bit rare, but when we're using dynamic columns to override an existing column, the old values from the existing column are kept.

### DIFF
--- a/packages/report-server/src/reportBuilder/transform/functions/insertColumns.ts
+++ b/packages/report-server/src/reportBuilder/transform/functions/insertColumns.ts
@@ -6,7 +6,7 @@
 import { yup } from '@tupaia/utils';
 
 import { Context } from '../../context';
-import { FieldValue } from '../../types';
+import { FieldValue, Row } from '../../types';
 import { TransformParser } from '../parser';
 import { buildWhere } from './where';
 import { mapStringToStringValidator } from './transformValidators';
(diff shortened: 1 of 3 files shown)


packages/report-server/src/reportBuilder/transform/functions/insertColumns.ts
```diff
@@ -36,9 +38,7 @@ const insertColumns = (table: TransformTable, params: InsertColumnsParams, conte
       const columnName = parser.evaluate(key);
       const columnValue = parser.evaluate(expression);
       if (!newColumns[columnName]) {
-        newColumns[columnName] = table.hasColumn(columnName)
-          ? table.getColumnValues(columnName) // Upserting a column, so fill with current column values
-          : new Array(table.length()).fill(undefined); // Creating a new column, so fill with undefined
+        newColumns[columnName] = new Array(table.length()).fill(undefined);
```
rohan-bes, line 41: Since we now re-insert the skipped rows, there's no need for the conditional upsert logic, just overwrite everything and let the skipped rows re-insert themselves if they exist.

packages/report-server/src/reportBuilder/transform/functions/insertColumns.ts
```diff
@@ -51,7 +51,14 @@ const insertColumns = (table: TransformTable, params: InsertColumnsParams, conte
     values,
   }));
 
-  return table.upsertColumns(columnUpserts);
+  // Drop, then re-insert the original skipped rows
+  const rowsToDrop = Object.keys(skippedRows).map(rowIndexString => parseInt(rowIndexString));
```
billli0, line 55:
```suggestion
  const rowsToDrop = Object.keys(skippedRows);
```
Should keys of skippedRows are number?

billli0 (approved):
@rohan-bes Thanks for adding the unit test and the fix! Just got a minor suggestion, but can pre approve.

packages/report-server/src/reportBuilder/transform/functions/insertColumns.ts
rohan-bes: Unfortunately even though we set them as numbers, they come out as strings. That's JS for you 🤷

PR #4229, opened 2022-10-18
MAUI-856: Viz Builder Access Separated From BES Admin

chris-pollard:
### Changes:

- Replace BES Admin checks in front end with 'Viz Builder User' permission
- Replace BES Admin check in back end (admin-panel-server) with 'Viz Builder User' permission
- Update all visualisation routes to only accept Viz Builder or BES Admin, and to check against specific permissions.


packages/admin-panel/src/VizBuilderApp/App.js
```diff
@@ -20,13 +20,13 @@ const Container = styled.main`
 `;
 
 export const App = ({ Navbar, Footer }) => {
-  const { data, isLoading: isUserLoading, isBESAdmin } = useUser();
-
+  const { data, isLoading: isUserLoading, isVizBuilderUser } = useUser();
+  console.log('isVizBuilderUser in App.js: ', isVizBuilderUser);
```
rohan-bes, line 24: 👀

packages/admin-panel/src/pages/resources/ResourcePage.js
```diff
@@ -33,6 +33,7 @@ export const ResourcePage = ({
   defaultSorting,
   displayUsedBy,
 }) => {
+  console.log('LinksComponent', LinksComponent);
```
rohan-bes, line 36: 👀

packages/admin-panel-server/src/routes/UserRoute.ts
```diff
@@ -6,15 +6,23 @@
...
+};
+
 export class UserRoute extends Route {
   public async buildResponse() {
     const user = await this.req.ctx.services.central.getUser();
-    return { ...camelcaseKeys(user), isBESAdmin: isBESAdmin(user.accessPolicy) };
+    return {
+      ...camelcaseKeys(user),
+      isBESAdmin: isBESAdmin(user.accessPolicy),
+      isVizBuilderUser: isVizBuilderUser(user.accessPolicy),
```
rohan-bes, line 25:
```suggestion
      isBESAdmin: isBESAdmin(user.accessPolicy),
      isVizBuilderUser: isVizBuilderUser(user.accessPolicy),
```
I know this pattern pre-dates your change, but can we get away with not adding these keys on the backend and just working them out in the front-end? It feels quite ugly for an endpoint to be passing around derivable data like this.

packages/admin-panel/src/App.js
```diff
@@ -39,7 +39,11 @@ export const App = ({ user, isBESAdmin }) => {
               <Switch>
                 {route.tabs.map(tab => (
                   <Route key={`${route.to}-${tab.to}`} path={`${route.to}${tab.to}`} exact>
-                    <tab.component getHeaderEl={getHeaderEl} isBESAdmin={isBESAdmin} />
+                    <tab.component
+                      getHeaderEl={getHeaderEl}
+                      isBESAdmin={isBESAdmin}
+                      isVizBuilderUser={isVizBuilderUser}
```
rohan-bes, line 45: Likewise here, can we just allow whichever pages/routes that need to know check if the user is BESAdmin/VizVuilderUser just work that out from the user's access policy rather than us pass this in to every component?

rohan-bes (changes requested):
Thanks for this @chris-pollard!

Just a few console logs to cleanup, and a request to take a crack to cleaning up the whole `isBESAdmin/isVizBuilderUser` pattern. Lemme know if you feel it's too much 👍

chris-pollard:
@rohan-bes see the changes, you should find the pattern removed from server and cleaned up on front end. Cheers!

packages/admin-panel/src/pages/resources/MapOverlaysPage.js
```diff
@@ -138,7 +140,7 @@ const renderNewMapOverlayVizButton = () => (
   </StyledLink>
 );
 
-export const MapOverlaysPage = ({ getHeaderEl, isBESAdmin, ...props }) => {
+const MapOverlaysPageComponent = ({ getHeaderEl, isVizBuilderUser, ...props }) => {
```
rohan-bes, line 143: Can you use `useUser()` here instead of passing in `isVizBuilderUser` via the state props? I'd love if we could remove the whole `isVizBuilderUser` from the redux state entirely 👍

packages/admin-panel/src/authentication/reducer.js
```diff
@@ -42,8 +42,8 @@ const logoutStateUpdater = (payload, currentState) => ({
   emailAddress: currentState.emailAddress, // But, remember user's email address
 });
 
-const isBESAdmin = policy => {
-  return new AccessPolicy(policy).allowsSome(undefined, BES_ADMIN_PERMISSION_GROUP);
+export const reduceIsVizBuilderUser = policy => {
```
rohan-bes, line 45: Part of me feels this function should just be a generic util, rather than within this reducer file? Especially if we're gonna remove the whole `isVizBuilderUser` from the redux state entirely 👍

rohan-bes (changes requested):
Looks great! Thanks so much @chris-pollard! These changes will really help with the stuff I'm doing for getting viz-builder in the Admin Panel.

Just a request to remove the `isVizBuilderUser` from the redux state entirely and just rely on `react-query`?

chris-pollard:
@rohan-bes added some commits to remove the redux pattern in the front end. It does seem a little awkward to be importing from the VizBuilderApp folder, but I guess ok for now? Seems like we should just move all queries up into the main folder eventually

rohan-bes (approved):
> It does seem a little awkward to be importing from the VizBuilderApp folder, but I guess ok for now? Seems like we should just move all queries up into the main folder eventually

Yup, I agree, it's a little awkward but I think it's fine 👍 

Thanks so much for this @chris-pollard, great stuff

packages/lesmis/src/routes/AdminPanelApp.js
```diff
@@ -181,12 +184,10 @@ const AdminPanelApp = ({ user, isBESAdmin }) => {
...
       <LesmisAdminRoute path={`${path}/viz-builder`} isBESAdmin>
-        <VizBuilderProviders>
-          <VizBuilderApp
-            basePath={adminUrl}
-            Navbar={({ user: vizBuilderUser }) => <AdminPanelNavbar user={vizBuilderUser} />}
-          />
-        </VizBuilderProviders>
+        <VizBuilderApp
+          basePath={adminUrl}
+          Navbar={({ user: vizBuilderUser }) => <AdminPanelNavbar user={vizBuilderUser} />}
```
rohan-bes, line 189: I think we'll need to be mindful of this for LESMIS. It's the correct change, but we may need to give the LESMIS Admin users `VIZ_BUILDER_USER` access. Can you add a merge task around that @chris-pollard?

rohan-bes (approved):
Nice stuff @chris-pollard! Couple small things, mainly that we should note that if we want to release this to LESMIS (which I think we should) then we should update the LESMIS Admin users

packages/lesmis/src/routes/AdminPanelApp.js
chris-pollard: Yep! @rohan-bes  Where would I record a merge task? Is this our prod merge (task in release notes) or merging into external LESMIS repo?

rohan-bes: Yeah just in the release notes should be okay. Just flag that we need to release this change to LESMIS.

@IgorNadj how does adding the steps to do a LESMIS release to the release process slab doc sound to you? It'd be nice if this was a job that the release manager should be responsible for

IgorNadj: Ah yeah it would be nice, but a LESMIS release would require comms and planning and potentially take weeks, so we can't really put it into a process. 

When we do our next LESMIS release in a couple months we will have to go back through all previous releases and look at the release notes, so if we highlight things that will affect LESMIS in the Release notes that would be helpful. E.g. this one could be "Viz Builder Access Separated From BES Admin (admin-panel and LESMIS)" or something like that.

PR #4230, opened 2022-10-20
RN-581: Fix changes to the entity hierarchy not appearing in Tupaia

rohan-bes:
### Issue RN-581:

There were a few bugs here causing this behaviour:
1. The `cacheDependencies` were incorrectly set on the `AncestorDescendantRelationType` rather than the `AncestorDescendantRelationModel`, so it was never clearing its cache on changes to the entity relations
2. The `cacheDepedencies` were ignoring the `entity` table, which means if entities using canonical relations changed, the cache wouldn't clear
3. There is a pretty common/easy timing issue to occur, where even if the cache clears correctly, if a user then navigates to Tupaia before the hierarchy is rebuilt (which may take a few minutes) then the old hierarchy will be recached and still stay even after the new hierarchy has been built

To avoid these issues I switched to having the `EntityHierarchyCacher` manually flag the `ancestor_descendant_relation` table as changed once the hierarchies have been rebuild. This needs to be a manual task, as we don't want to add the automatic `notification()` trigger to the `ancestor_descendant_relation` table as it slows down the rebuilt process significantly (and would be incredibly noisy).

I consider this solution somewhat of a hack. Really, what this issue highlight is the following:
1. We probably shouldn't be doing caching on the models, it's hard to manage cache invalidation and it just a bit unexpected
2. We really need some kind of 'event queue' concept in Tupaia for handling side effects explicitly. As it stands, it can be really tricky to ensure the change handlers behave in a coordinated fashion
3. It would be nice to centralise the entity hierarchy fetching/caching/rebuilding in the `entity-server`. For now it just does fetching, and caching too I suppose but that's also done in other places. Moving `web-config-server` over to using the `entity-server` would help here...

What does the reviewer think? Is this too hacky? Is there are better option?


packages/database/src/changeHandlers/entityHierarchyCacher/EntityHierarchyCacher.js
```diff
@@ -60,6 +62,17 @@ export class EntityHierarchyCacher extends ChangeHandler {
     const start = Date.now();
     const subtreeRebuilder = new EntityHierarchySubtreeRebuilder(transactingModels);
     await subtreeRebuilder.rebuildSubtrees(rebuildJobs);
+
+    // explicitly flag ancestor_descendant_relation as changed so that model level caches are cleared
+    transactingModels.database.markRecordsAsChanged(
+      TYPES.ANCESTOR_DESCENDANT_RELATION,
+      rebuildJobs.map(({ hierarchyId, rootEntityId }) => ({
```
rohan-bes, line 69: Decided to send out a change notification per subtree rebuild. Not required, but opens the door for some nice ability to only clear requests from the cache that would have changed (at the moment the whole cache is blown up whenever an entity relation changes)

packages/database/src/TupaiaDatabase.js
```diff
@@ -181,7 +183,7 @@ export class TupaiaDatabase {
 
   wrapInTransaction(wrappedFunction) {
     return this.connection.transaction(transaction =>
-      wrappedFunction(new TupaiaDatabase(transaction)),
+      wrappedFunction(new TupaiaDatabase(transaction, this.changeChannel)),
```
rohan-bes, line 186: @IgorNadj had to update the code a little to get the unit tests to work. They were hanging because now we are instantiating a `changeChannel` connection within a transaction, which was never being cleaned up. I switched it to pass through the parent connection's `changeChannel` though there may possibly still be issues with this.
1. It's possible for the `changeChannel` to still be `null` when we create the transaction, in which case the transaction will still create it's own if it needs one. This seems pretty unlikely though since we create the `changeChannel` most of the time...
2. The `changeChannel` and the transaction have separate database connections. This is actually always the case generally it seems. However, this does mean that notifications will be sent before the transaction completes... Honestly it seems a bit odd to me that the two have separate connections, but I feel it's too much of a refactor to change it right now...

IgorNadj (approved):
LGTM

PR #4235, opened 2022-10-21
RN-699: Enforce permissions checking on both Data Elements and Entities in data-broker

rohan-bes:
### Issue RN-699:

So it turns out that `data-broker` has just been checking if the user has access to the requested data-elements permission groups anywhere, instead of checking that they have access to the permission groups in the entities they're requesting data for. This was never really the intended behaviour, so it's possible this got lost in merge conflicts or something...

Anyway, this PR is to enforce the stricter permissions check against data-elements and entities.


packages/access-policy/src/AccessPolicy.js
```diff
@@ -0,0 +1,132 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+export class AccessPolicy {
```
rohan-bes, line 6: This is an absolutely superficial change. `AccessPolicy` has always been one of the very rare cases where we keep a class in a file named something other than the class name. I find myself constantly forgetting this and struggling to find it when searching in vscode, so I moved it. Hope the reviewer agrees and sorry for doing it here but I didn't feel it warranted a PR of its own 😅

packages/access-policy/src/index.js
```diff
@@ -3,130 +3,4 @@
...
-   *
-   * @returns entities[] The entity objects
-   */
-  getEntitiesAllowed(permissionGroup) {
-    const allEntityCodes = Object.keys(this.policy);
-    if (!permissionGroup) return allEntityCodes;
-    return allEntityCodes.filter(e => this.allows(e, permissionGroup));
-  }
-}
+export { AccessPolicy } from './AccessPolicy';
```
rohan-bes, line 6: Moved to `AccessPolicy.js`

packages/data-broker/src/DataBroker.js
```diff
@@ -85,33 +88,66 @@ export class DataBroker {
...
-      ) {
-        continue;
+
+    const getDataElementsWithMissingPermissions = permissions =>
+      dataElements
+        .filter(element => element.permission_groups.length > 0)
+        .filter(element => !element.permission_groups.some(group => permissions.includes(group)))
+        .map(element => element.code);
+
+    if (countryCodes === null) {
```
rohan-bes, line 112: It's possible for countryCode to be null when fetching for `sync_groups` I believe. At the moment sync groups don't do permissions checks, but they could... so leaving this here for now 👍

packages/data-broker/src/DataBroker.ts
```diff
@@ -144,33 +153,70 @@ export class DataBroker {
...
-      ) {
-        continue;
+
+    const getDataElementsWithMissingPermissions = (permissions: string[]) =>
+      (dataElements as DataElement[])
+        .filter(element => element.permission_groups.length > 0)
+        .filter(element => !element.permission_groups.some(group => permissions.includes(group)))
+        .map(element => element.code);
+
+    if (countryCodes === null) {
```
IgorNadj, line 178: I'd love to drop support for no country codes if possible, I can't wrap my head around it. In fact I don't know if it's ever done?

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.ts
```diff
@@ -329,6 +330,88 @@ describe('DataBroker', () => {
...
+              type: 'dataElement',
+            },
+            { organisationUnitCodes: ['TO'] },
+          ),
+        ).toBeRejectedWith(
+          `Missing permissions to the following data elements:\nTO: RESTRICTED_01`,
+        );
+      });
+
+      it("throws an error if any of entities in fetch the user doesn't have required permissions for", async () => {
```
IgorNadj, line 374: Beautiful tests

IgorNadj (approved):
Looks great nice one! Thanks

packages/data-broker/src/DataBroker.ts
IgorNadj: Ah just saw your other comment, all good.

PR #4247, opened 2022-10-27
RN-701 Update terms and conditions URL

avaek:
Issue:[ RN-701](https://linear.app/bes/issue/RN-701/fix-terms-and-conditions-link-in-meditrak-app)

### DIFF
--- a/packages/meditrak-app/app/user/CreateUserPage.js
+++ b/packages/meditrak-app/app/user/CreateUserPage.js
@@ -202,7 +202,7 @@ const fields = [
         I agree to the{' '}
         <Text
           style={localStyles.termsLink}
-          onPress={() => Linking.openURL('http://beyondessential.com.au/terms-and-conditions/')}
+          onPress={() => Linking.openURL('https://beyondessential.com.au/terms-and-conditions/')}
         >
           terms and conditions
         </Text>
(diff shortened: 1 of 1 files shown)


packages/meditrak-app/app/user/CreateUserPage.js
```diff
@@ -202,7 +202,7 @@ const fields = [
         I agree to the{' '}
         <Text
           style={localStyles.termsLink}
-          onPress={() => Linking.openURL('http://beyondessential.com.au/terms-and-privacy/')}
+          onPress={() => Linking.openURL('http://beyondessential.com.au/terms-and-conditions/')}
```
rohan-bes, line 205:
```suggestion
          onPress={() => Linking.openURL('https://beyondessential.com.au/terms-and-conditions/')}
```
Tiny and probably doesn't matter since I think it'll always get switched to https, but might be nice?

rohan-bes (approved):
Nice one @avaek. Small comment but approving all the same

packages/meditrak-app/app/user/CreateUserPage.js
avaek: Thanks for the suggestion @rohan-bes! I wasn't sure which to go for and ended up just copying the format of the previous link but happy to implement your suggestion

PR #4248, opened 2022-10-28
RN-702: Re-instate excluding columns when using the 'exclude' merge function

rohan-bes:
### Issue RN-702:

Just re-introducing some behaviour that got removed as part of RN-175.

Also scrapped the `where` clause support on `mergeRows`, as it's never been used and makes the code more complex than needed

### DIFF
--- a/packages/report-server/src/__tests__/reportBuilder/transform/parser/parser.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/transform/parser/parser.test.ts
@@ -121,31 +121,6 @@ describe('parser', () => {
   });
 
   describe('in transforms', () => {
-    it('mergeRows supports parser lookups on where', () => {
-      const transform = buildTransform([
-        {
-          transform: 'mergeRows',
-          using: {
-            organisationUnit: 'exclude',
-            period: 'exclude',
-            '*': 'sum',
-          },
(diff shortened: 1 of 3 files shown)


packages/report-server/src/__tests__/reportBuilder/transform/parser/parser.test.ts
```diff
@@ -121,31 +121,6 @@ describe('parser', () => {
   });
 
   describe('in transforms', () => {
-    it('mergeRows supports parser lookups on where', () => {
```
IgorNadj, line 124: Do we need a new test for mergeStrategy = exclude?

rohan-bes: Ah good call! I'll re-add them, thanks 👍

rohan-bes: Ah wait sorry, we do actually have the test for `exclude` removing the columns from the table

IgorNadj: Ah cool

IgorNadj (approved):

PR #4251, opened 2022-11-02
RN-441 Add ids to admin panel

IgorNadj:

### DIFF
--- a/packages/admin-panel/src/editor/BulkEditButton.js
+++ b/packages/admin-panel/src/editor/BulkEditButton.js
@@ -12,7 +12,10 @@ import { openBulkEditModal } from './actions';
 
 export const BulkEditButtonComponent = ({ dispatch, value: recordId, actionConfig, row }) => {
   return (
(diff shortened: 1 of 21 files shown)


packages/admin-panel/src/editor/BulkEditButton.js
```diff
@@ -12,7 +12,10 @@ import { openBulkEditModal } from './actions';
 
 export const BulkEditButtonComponent = ({ dispatch, value: recordId, actionConfig, row }) => {
   return (
-    <IconButton onClick={() => dispatch(openBulkEditModal(actionConfig, recordId, row))}>
+    <IconButton
+      className="bulk-edit-button"
```
rohan-bes, line 16: Out of curiosity, why do we use `className` on `IconButton` as opposed to `id`?

rohan-bes (approved):
Noice! 👍

packages/admin-panel/src/editor/BulkEditButton.js
IgorNadj: Good catch, that was a mistake. Thanks

PR #4254, opened 2022-11-04
Add unfpa population tile

billli0:
### Issue #:
MAUI-1045

### Changes:
- new tile for UNFPA population


chris-pollard:
@billli0 , I was wondering why the migration was needed, because this config can be added in the project page in admin panel. Also, should projectDashboardHeader: 'Regional' is removed in new config, but should it be?

billli0:
@chris-pollard Oh nice catch thanks! Yeah we can edit configs in admin panel too. My point for migration is it would automate things, plus it is just copy and paste.

chris-pollard (approved):
LGTM!

PR #4257, opened 2022-11-07
RN-708: Fix Kobo Sync Groups being stuck in a 'Syncing' state

rohan-bes:
### Issue RN-708:

Flag all kobo sync groups as being in an IDLE sync state upon server startup

### DIFF
--- a/packages/central-server/src/kobo/startSyncWithKoBo.js
+++ b/packages/central-server/src/kobo/startSyncWithKoBo.js
@@ -170,6 +170,10 @@ export async function startSyncWithKoBo(models) {
     const koboDataServiceSyncGroups = await models.dataServiceSyncGroup.find({
       service_type: SERVICE_TYPE,
     });
+
+    // Put all kobo sync groups in an idle state, just in case they're stuck in an 'syncing' state due to a server shutdown
+    await Promise.all(koboDataServiceSyncGroups.map(async syncGroup => syncGroup.setSyncIdle()));
+
     koboDataServiceSyncGroups.forEach(dssg =>
       setInterval(() => syncWithKoBo(models, dataBroker, dssg.code), PERIOD_BETWEEN_SYNCS),
     );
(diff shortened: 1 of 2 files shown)


packages/database/src/modelClasses/DataServiceSyncGroup.js
```diff
@@ -22,6 +22,10 @@ const syncStatuses = {
 class DataServiceSyncGroupType extends DatabaseType {
   static databaseType = TYPES.DATA_SERVICE_SYNC_GROUP;
 
+  async setSyncIdle() {
+    return this.model.update({ id: this.id }, { sync_status: syncStatuses.idle });
+  }
```
rohan-bes, line 27: Technically this is doing the exact same thing as `setSyncCompletedSuccessfully()`, however I kind of like having two separate methods as it's a bit of an implementation detail that they both just set the state to `IDLE`. Also it makes the code read a lot nicer 😄

packages/central-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -170,6 +170,10 @@ export async function startSyncWithKoBo(models) {
     const koboDataServiceSyncGroups = await models.dataServiceSyncGroup.find({
       service_type: SERVICE_TYPE,
     });
+
+    // Put all kobo sync groups in an idle state, just in case they're stuck in an 'syncing' state due to a server shutdown
+    await Promise.all(koboDataServiceSyncGroups.map(async syncGroup => syncGroup.setSyncIdle()));
```
rohan-bes, line 175: It does feel a touch arbitrary/hacky doing this. However, this code is really designed to run once, upon server startup, which feels like the appropriate place to force the state to be `IDLE`.

Given that the state is just for application logic, perhaps there's a bit of a code smell here that we ought to manage the sync group state within memory rather than the DB? So that it wouldn't persist between server startups... But feels a bit much for right now

IgorNadj (approved):

PR #4259, opened 2022-11-08
RN-648: SQL Data Table Service

rohan-bes:
### Issue RN-648:

Introducing a new `SqlDataTableService` which allows us query an `externalDatabaseConnection` with user defined SQL.


packages/data-table-server/src/dataTableService/userDefined/SqlDataTableService.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { yup } from '@tupaia/utils';
+import { DataTableServerModelRegistry } from '../../types';
+import { UserDefinedDataTableService } from './UserDefinedDataTableService';
+
+const configSchema = yup.object().shape({
+  externalDatabaseConnectionCode: yup.string().required(),
```
rohan-bes, line 11: Chose to go with `code` over `id`. Arguably `id` is better since it's more stable, but I doubt we're going to change these codes often/ever and `code` is a lot more readable 👍

packages/data-table-server/src/dataTableService/DataTableService.ts
```diff
@@ -45,7 +45,7 @@ export abstract class DataTableService<
    */
```
IgorNadj, line 45: Can we add a class comment on DataTableService explaining what it is please? It only took me a few mins to grok it but would be helpful! 🙏

packages/data-table-server/src/dataTableService/DataTableServiceBuilder.ts
```diff
@@ -11,6 +11,7 @@ import {
   EntityRelationsDataTableService,
   EventsDataTableService,
 } from './internal';
+import { SqlDataTableService } from './userDefined';
 
 /**
  * Generic builder class that allows us to configure the context for a specific DataTableService
```
IgorNadj, line 17: DataTableServiceBuilder could just be a fn? I don't see the need for OOP here?

packages/data-table-server/src/dataTableService/DataTableServiceBuilder.ts
```diff
@@ -51,7 +52,9 @@ const internalDataTableServiceBuilders = {
   entity_relations: () => new DataTableServiceBuilderForType(EntityRelationsDataTableService),
 };
 
-const userDefinedDataTableServiceBuilders = {};
+const userDefinedDataTableServiceBuilders = {
```
IgorNadj, line 55: Since these all have the same builders, why the differentiation between internal and user defined? Logically they are the same right? It just so happens that some of them target tupaia data and some of them target other data.

Could we achieve the same effect but putting them all in one services folder and then maybe prefixing the tupaia ones with TupaiaX?

packages/data-table-server/src/dataTableService/userDefined/UserDefinedDataTableService.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { yup } from '@tupaia/utils';
```
IgorNadj, line 6: See comment above about what the difference is? From a quick scan it looks like UserDefinedDataTables have specific shaped params, but nothing here makes me think this shouldn't just be on DataTable.

packages/data-table-server/src/dataTableService/utils/dataTableParamsToYupSchema.ts
```diff
@@ -0,0 +1,53 @@
+/**
```
IgorNadj, line 1: Probs better as a chat but... Biao and I were talking about yup today. With the new @tupaia/types package, we want to generate as much as possible to avoid staleness etc. 

He's having a look at generating json-schema from the typescript type definitions, which would then be passed into a json-schema validator like https://github.com/ajv-validator/ajv. This would mean we would get validation for free across all packages. Let's have a look at moving over to this new way?

packages/data-table-server/src/models/ExternalDatabaseConnection.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import {
+  ExternalDatabaseConnectionModel as BaseExternalDatabaseConnectionModel,
+  ExternalDatabaseConnectionType as BaseExternalDatabaseConnectionType,
+} from '@tupaia/database';
+import { Model } from '@tupaia/server-boilerplate';
+
+export type ExternalDatabaseConnectionFields = Readonly<{
```
IgorNadj, line 12: Import from @tupaia/types, merged into dev

IgorNadj (changes requested):
Looking good! Some comments mostly around user-defined/internal and types/validation.

packages/data-table-server/src/dataTableService/DataTableServiceBuilder.ts
rohan-bes: That is true. Tbh, the main reason I went with the builder pattern here is to get the generics working how I wanted it to.

With the current implementation, you get some pretty neat behaviour that ensures that the `context` that gets passed in meets the requirements of the `serviceType`, ie.
```
new DataTableServiceBuilder()
  .setServiceType('entities')
  .setContext({ models: tupaiaModels }) // ts error: models is not assignable to parameter of type EntitiesDataTableContext
  .build();
```

I've tried my hand at getting this working in a function, but couldn't quite work it out. Did you have any suggestions on how to get this working, I'd be down to pair on it for a bit 😄 

If it's impossible tho, I'd prefer to keep the functionality as is, since the type validation on the context object is really clean atm 👍

packages/data-table-server/src/dataTableService/userDefined/UserDefinedDataTableService.ts
rohan-bes: So the general idea is that with `UserDefinedDataTables`, the user gets to control which parameter the data table takes. This is that part of the UI where the user creates the parameters. 

So in the backend, we basically have to read the config to work out what parameters the user has defined, and then create a yup schema from that config so that when we request data for that data table we can validate it has been requested with the correct parameters.

Whereas with the non-user defined data-tables, the parameters are hard coded. I did actually consider moving those hard-coded parameters into the config, and then the code could be the same with the user-defined data-tables? The main drawback I found there is that we couldn't use quite as precise validators, but perhaps it's worth it, thoughts?

packages/data-table-server/src/dataTableService/DataTableServiceBuilder.ts
IgorNadj: Hmm yeah I see what you mean. My first thought would be to just have the same context passed in to each service.

But if we want each service to have a subset of the context, I think something like this would work? 
https://github.com/beyondessential/tupaia/commit/a9b0f8612ee68e1de5e3c75ae2a25ec7fb1b04c4

packages/data-table-server/src/dataTableService/userDefined/UserDefinedDataTableService.ts
IgorNadj: I see, I think I see it like configuring each DataTable is just a form with a few fields, similar to a Slack bot integration, or a VSCode plugin settings screen, or a chrome extension settings screen etc. It just so happens to be that the SqlDataTable has a field which is an array.

Like GoogleSheetsDataTable will need sheetId. 
```
{
  name: 'myGoogleSheetsDataTable',
  params: {
    sheetId: '123'
  }
}
```
so the UI for this would be 
```
Name: [       ]
Params:
  Sheet Id: [      ]
```

And SqlDataTable will need connectionId and sqlParameters (optional).
```
{
  name: 'mySqlDataTable',
  params: {
    connectionId: '1',
    sqlParams: {
      dataElements: ['ABC123', 'XYZ456']
    }
  }
}
```
and the UI for this would be:
```
Name: [       ]
Params:
  Connection: [      v]
  SQL Params (optional):
    1 Key [      ] Value [      ] Type [     v] (+)(-) 
    1 Key [      ] Value [      ] Type [     v] (+)(-) 
    1 Key [      ] Value [      ] Type [     v] (+)(-) 
```

Not sure if I'm off the mark but yeah just how the mental image is in my head.

rohan-bes:
@IgorNadj I reworked the whole Internal/UserDefined data-tables structure as per our discussion.

packages/data-table-server/src/dataTableService/DataTableServiceBuilder.ts
IgorNadj: Discussed, builder is cool 😎

IgorNadj (approved):
Love it, thanks so much for having the chat with me and taking me through it, happy to get how it all fits in together now!

packages/database/src/migrations/20230119053753-ConvertInternalDataTablesToTheirOwnTypes-modifies-schema.js
```diff
@@ -0,0 +1,44 @@
...
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = async function (db) {
+  // Copy sync_cursor over to data_service_sync_group
+  const internalDataTables = (await db.runSql("SELECT * FROM data_table WHERE type = 'internal'"))
+    .rows;
+  for (let i = 0; i < internalDataTables; i++) {
```
IgorNadj, line 21: Missing .length

rohan-bes: Ah! Great catch! 🙏

packages/database/src/migrations/20230119053753-ConvertInternalDataTablesToTheirOwnTypes-modifies-schema.js
```diff
@@ -0,0 +1,44 @@
...
+    const dataTable = internalDataTables[i];
+    await db.runSql(
```
IgorNadj, line 23: These weren't working for me 😢 

Enum juggling is weird, this worked for me:
```
exports.up = async function () {
  const db = new TupaiaDatabase();

  const internalDataTables = await db.executeSql(
    "SELECT * FROM data_table WHERE type = 'internal'",
  );

  const newEnum = internalDataTables.map(dt => dt.code);
  await db.executeSql(
    `
      ALTER TYPE data_table_type RENAME TO data_table_type_temp$;
      CREATE TYPE data_table_type AS ENUM(${arrayToDbString(newEnum)});
      ALTER TABLE data_table DROP COLUMN type;
      ALTER TABLE data_table ADD COLUMN type data_table_type;
      DROP TYPE data_table_type_temp$;
    `,
    [],
  );

  for (let i = 0; i < internalDataTables.length; i++) {
    const dataTable = internalDataTables[i];
    await db.executeSql(`UPDATE data_table SET type = ? WHERE code = ?;`, [
      dataTable.code,
      dataTable.code,
    ]);
  }
};
```

IgorNadj: And then re-add sql into the enum...

rohan-bes: Ah thanks so much for this! 🙏  

My local db isn't running migrations atm and getting the latest dump is taking some time. So I've just been blindly writing SQL which of course is a reliable method of development 😉

PR #4263, opened 2022-11-14
RN-710 Add superset instances

IgorNadj:
### Changes:

- Remove insecure flag, not needed any more, servers have valid certs now
- Swap out package `needle` for standard `node-fetch`
- Add new superset instances

## Manual Release Tasks
- [x] Add ENV vars


rohan-bes (approved):
Noice!

PR #4271, opened 2022-11-25
Update lesmis on premise

billli0:
Note: We cannot fully install dependencies because it took a bit of time to install a dependency due to low network speed, and it will break the pipe. Please see other notes in the changes.

### DIFF
--- a/packages/devops/scripts/deployment-vm-app/install.sh
+++ b/packages/devops/scripts/deployment-vm-app/install.sh
@@ -32,6 +32,9 @@ git reset --hard origin/${GIT_BRANCH}
 # Yarn install
 yarn install --frozen-lockfile
 
+# TODO: 1. Wrong password in lastpass to login lesmis@tupaia.org
+#       2. It will pull wrong key for REACT_APP_LESMIS_API_URL in lesmis .env
+#       3. change all service api urls to localhost as they interact locally
 # Fetch env vars
 echo "Note: if lastpass fails, check email account $LASTPASS_EMAIL for a verification check"
 LASTPASS_EMAIL=$LASTPASS_EMAIL LASTPASS_PASSWORD=$LASTPASS_PASSWORD yarn download-env-vars $DEPLOYMENT_NAME
--- a/packages/devops/scripts/deployment-vm-app/install.sh
+++ b/packages/devops/scripts/deployment-vm-app/install.sh
@@ -32,9 +32,6 @@ git reset --hard origin/${GIT_BRANCH}
 # Yarn install
 yarn install --frozen-lockfile
 
-# TODO: 1. Wrong password in lastpass to login lesmis@tupaia.org
-#       2. It will pull wrong key for REACT_APP_LESMIS_API_URL in lesmis .env
-#       3. change all service api urls to localhost as they interact locally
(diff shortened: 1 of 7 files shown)


packages/devops/scripts/deployment-vm-app/install.sh
```diff
@@ -32,6 +32,9 @@ git reset --hard origin/${GIT_BRANCH}
 # Yarn install
 yarn install --frozen-lockfile
 
+# TODO: 1. Wrong password in lastpass to login lesmis@tupaia.org
```
IgorNadj, line 35: Discussed

packages/devops/scripts/deployment-vm-app/install.sh
```diff
@@ -32,6 +32,9 @@ git reset --hard origin/${GIT_BRANCH}
...
+#       2. It will pull wrong key for REACT_APP_LESMIS_API_URL in lesmis .env
+#       3. change all service api urls to localhost as they interact locally
```
IgorNadj, line 37: Good idea, let's leave it for future when we can do it for both our infra and the LESMIS infra at the same time

packages/devops/scripts/manual/vm-app/deploy.sh
```diff
@@ -5,17 +5,20 @@ cd "$SCRIPT_DIR"
 
 APP_SERVER_HOST=${1?Argument 1 - APP_SERVER_HOST required.}
 APP_SERVER_SSH_KEY=${2?Argument 2 - APP_SERVER_SSH_KEY required.}
+APP_SERVER_PORT=${3?Argument 3 - APP_SERVER_PORT required.}
+
+# TODO: manually add ssh key or change to use password
```
IgorNadj, line 10: Added as manual step in documentation

packages/devops/scripts/manual/vm-db/deploy.sh
```diff
@@ -5,14 +5,17 @@ cd "$SCRIPT_DIR"
 
 DB_SERVER_HOST=${1?Argument 1 - DB_SERVER_HOST required.}
 DB_SERVER_SSH_KEY=${2?Argument 2 - DB_SERVER_SSH_KEY required.}
+APP_SERVER_PORT=${3?Argument 3 - APP_SERVER_PORT required.}
+
+# TODO: manually add ssh key or change to use password
```
IgorNadj, line 10: Same as above

IgorNadj:
> it took a bit of time to install a dependency due to low network speed, and it will break the pipe
They will get a fiber connection soon, should be fine

IgorNadj (approved):

packages/devops/scripts/deployment-vm-db/setup.sh
```diff
@@ -6,6 +6,7 @@ SCRIPT_DIR=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 cd "$SCRIPT_DIR"
 
 # install postgres 13
+# do not trigger if no psql
```
IgorNadj, line 9: Discussed, lets get rid the if check, it is safe to re-run these commands every time

PR #4272, opened 2022-11-25
RN-712: Fix broken pre-aggregator permissions

rohan-bes:
### Issue RN-712:

Grant pre-aggregators BES Admin permissions so that they can fetch required data


billli0 (approved):
Thank you for helping to fix this @rohan-bes

PR #4274, opened 2022-12-01
Background img cover the entire area of <div> and clips the excess

billli0:
### Issue #:
https://linear.app/bes/issue/MAUI-1335#comment-6cbe0091


chris-pollard (approved):
LGTM!

PR #4275, opened 2022-12-04
OSC-10: Admin panel entity hierarchy - Backend

kael89:
### Issue #:
[OS-10](https://linear.app/bes/team/OSC/active) - just the backend in this PR

### Changes:
* Added a `GET /entityHierarchyTree` endpoint in `admin-panel-server`
* Automatically stringify `columns`,  `sort` in `central-server` requests done with `api-client`
* Add support for `<`, `<=`, `>`, `>=` comparators in `entity-server`
* Add support for filtering on `ancestor_descendant_relation.generational_distance` for entities fetched with `entity-server`
* Add `GET /entityRelations` endpoint in `central-server` (basic version, needs more work)

### DIFF
--- a/packages/admin-panel-server/src/routes/FetchEntityHierarchyTreeRoute.ts
+++ b/packages/admin-panel-server/src/routes/FetchEntityHierarchyTreeRoute.ts
@@ -1,0 +1,87 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { Request } from 'express';
+
+import { Route } from '@tupaia/server-boilerplate';
+import { groupBy } from 'lodash';
+import { getSortByKey } from '@tupaia/utils';
+
+interface EntityNode {
+  code: string;
+  name: string;
+  hierarchyName: string;
+}
+
+export type FetchEntityHierarchyTreeRequest = Request<
+  Record<string, never> | { hierarchyName: string; entityCode: string },
+  (EntityNode & {
+    children: EntityNode[];
+  })[],
+  Record<string, never>,
+  Record<string, never>
+>;
+
+export class FetchEntityHierarchyTreeRoute extends Route<FetchEntityHierarchyTreeRequest> {
+  public async buildResponse() {
+    const { hierarchyName, entityCode } = this.req.params;
+
+    return hierarchyName && entityCode
+      ? this.buildHierarchyEntityTree(hierarchyName, entityCode)
+      : this.buildRootTree();
+  }
+
+  private async buildRootTree() {
+    const { central: centralApi } = this.req.ctx.services;
+
+    const rootEntities: Record<string, string>[] = await centralApi.fetchResources('entities', {
+      filter: { type: 'project' },
+      columns: ['id', 'code', 'name'],
+      sort: ['name'],
+    });
+
+    const childRelations = await centralApi.fetchResources('entityRelations', {
+      filter: { parent_id: rootEntities.map(({ id }) => id) },
+      columns: ['parent_id', 'entity.code', 'entity.name'],
+      sort: ['entity.name'],
+    });
+    const childRelationsByParent = groupBy(childRelations, 'parent_id');
+
+    return rootEntities.map(({ id, code, name }) => ({
+      code,
+      name,
+      hierarchyName: code,
+      children: (childRelationsByParent[id] || []).map(child => ({
+        code: child['entity.code'],
+        name: child['entity.name'],
+        hierarchyName: code,
+      })),
+    }));
+  }
+
+  private async buildHierarchyEntityTree(hierarchyName: string, entityCode: string) {
+    const { entity: entityApi } = this.req.ctx.services;
+
+    const descendants = await entityApi.getDescendantsOfEntity(hierarchyName, entityCode, {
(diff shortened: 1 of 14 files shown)


packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -4,28 +4,33 @@
  */
 
 import { QueryConjunctions } from '@tupaia/server-boilerplate';
+import { getSortByKey } from '@tupaia/utils';
 
-import { Writable, ObjectLikeKeys, Flatten } from '../../../types';
-import { EntityFilter, EntityFields } from '../../../models';
+import { Writable, NumericKeys, ObjectLikeKeys, Flatten } from '../../../types';
```
kael89, line 9: @rohan-bes  I had to add/refactor a bunch of code because `generational_distance` is the first non-string field here...

I could be a bit lazy and define `generational_distance` as a string - the code wouldn't complain :wink: but it felt wrong...

That being said, I believe that simplifying the types in this file would make it more manageable. They are very cool, but they go a long way to ensure that things are DRY.

As an example, instead of inferring the nested JSON fields and their types from `EntityFields`, they could be explicitly defined as

```ts
interface FilterableFields {
  code: 'name';
  // ...etc
  atttributes_type: 'string';
  generational_distance: number;
}
```

Tests could be used to assert that at least the types used are compatible with the DB schema.

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -58,35 +68,56 @@ const operatorToSqlComparator = {
...
-const convertValueToAdvancedCriteria = (operator: Operator, value: string | string[]) => {
+const formatValue = <T extends keyof EntityFilterQuery>(field: T, value: string) =>
+  isNumericField(field) ? parseFloat(value) : value;
+
+const convertValueToAdvancedCriteria = (
+  field: keyof EntityFilterQuery,
+  operator: Operator,
+  value: string,
+) => {
+  const formattedValue = value.includes(MULTIPLE_VALUES_DELIMITER)
```
kael89, line 98: @rohan-bes  I had to move value parsing from `extractFilterFromQuery()` to this method. The reason is that here the field is `keyof EntityFilterQuery`, e.g. `attributes_type`, and not yet formatted to `attributes->>type`. We generally use `_` to define the filterable fields and their types.

This change has a bit of a risk, but luckily `entity-server` has an amazing test suite, which includes some filter tests too! I can also do some manual testing for this one

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -129,14 +154,11 @@ export const extractFilterFromQuery = (
   }
 
   const filterClauses = queryFilter.split(CLAUSE_DELIMITER).map(toFilterClause);
-  const filter: Writable<NotNullValues<EntityFilter>> = {};
+  const filter: Writable<NotNullValues<Record<string, unknown>>> = {};
```
kael89, line 157: @rohan-bes I had to make this type simpler because TS complained about the following in L161:

```ts
filter[formattedField] = convertValueToAdvancedCriteria(field, operator, value);
```

I think this is because we don't know if the converted value (`string | number`) can be assigned to the selected `field`, which would accept only `string` or only `number`. In the past everything was a `string`, so that was fine.

Perhaps more elaborate types could take care of this, but I wouldn't bother :stuck_out_tongue:

packages/utils/src/object.js
```diff
@@ -33,7 +33,7 @@ export const getKeysSortedByValues = (object, options = {}) => {
  *
  * @param {string} key
  * @param {{ ascending: boolean }} [options]
- * @returns { (a: Object<string, any>, b: Object<string, any> ) => number }
+ * @returns { (a: any, b: any ) => number }
```
kael89, line 36: Types other than `Object<string, any>` also have sortable properties, e.g. array length, string length

packages/admin-panel-server/src/routes/FetchEntityHierarchyTreeRoute.ts
```diff
@@ -0,0 +1,87 @@
...
+      columns: ['id', 'code', 'name'],
+      sort: ['name'],
+    });
+
+    const childRelations = await centralApi.fetchResources('entityRelations', {
+      filter: { parent_id: rootEntities.map(({ id }) => id) },
+      columns: ['parent_id', 'entity.code', 'entity.name'],
+      sort: ['entity.name'],
+    });
+    const childRelationsByParent = groupBy(childRelations, 'parent_id');
```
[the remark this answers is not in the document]
edmofro: This is a little complicated and I might agree with moving the responsibility to the entity-server, but I suspect, having worked with Flic on UI designs, in the end we won't need it. Instead, I think we'll have an initial step where you select the project you're working within, and users are never expected to view or manipulate more than one project hierarchy at a time (which I think reflects reality)

edmofro (commented):
All of the code looks great @kael89, from that perspective this is approved! However I think it's worth waiting for Flic's UI designs before merging, as that may impact the back end (the bones of this will surely stay though)

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
kael89: In Grug we trust 😄

PR #4278, opened 2022-12-05
RN-205: Build With Test Data w/out Orgunit or Hierarchy

chris-pollard:
### Changes:

- Fix bug that stops test data being previewed due to unselected org unit or hierarchy. The user should be able to submit test data and preview aggregations and transforms without selecting org unit or hierarchy.


billli0 (approved):
@chris-pollard Looks great! Thanks for fixing it and adding the unit tests.

billli0:
One question though, if we don't specify the org unit and hierarchy, can we perform any org unit related transform?

PR #4279, opened 2022-12-05
hotfix: Add 'Entities' section to LESMIS Admin Panel

rohan-bes:


IgorNadj (approved):

IgorNadj:
Can change base to lesmis-on-premise if you like

IgorNadj:
Not sure what that means sorry, but it's ok, this will make its way into their deployment

PR #4281, opened 2022-12-06
RN-678: Fix LESMIS dashboard date picker

chris-pollard:
### Changes:

- Update DashboardReport component to account for different date param inputs.

### DIFF
--- a/packages/lesmis/src/components/DashboardReport.js
+++ b/packages/lesmis/src/components/DashboardReport.js
@@ -14,7 +14,7 @@ import { Chart, ListVisual } from './Visuals';
 import * as COLORS from '../constants';
 import { useDashboardReportDataWithConfig } from '../api/queries';
 import { FlexEnd } from './Layout';
-import { I18n, useIsFavouriteDashboardSelected, useUrlParams, useUrlSearchParam } from '../utils';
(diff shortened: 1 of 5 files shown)


billli0 (changes requested):
@chris-pollard This logic looks solid! Thanks for fixing it. 

Seems the date logic have been updated, could you also apply that to `DashboardReportModal` (expand page when you click 'see more'), and `DashboardReportPage` (page for PDF exporting)? 

These two components also have usage of `startDate` and `endDate`. The new logic can be reused and applied to them.

chris-pollard:
> 

Hey @billli0 actually had a look through the components, and as the modal and export both use search params and the history api, it works perfectly. So no need for that conditional logic in the modal or export.

billli0:
Good point @chris-pollard, seems like we don't need to do that for `DashboardReportPage`. However on `DashboardReportModal`:
https://github.com/beyondessential/tupaia/blob/c64af87a907df19cb875bfbf0d0ea5da9b8508b8/packages/lesmis/src/components/DashboardReportModal.js#L134

It fetches `startDate` and `endDate` from search params, but the `year` has been neglected, which causes that if 2018 is selected, then click `see more`, the date picker is not well addressed and it will show default period.

chris-pollard:
> 

Good point, @billli0 I'll look at that.

billli0 (approved):

PR #4299, opened 2023-01-12
RN-721 Superset filter by org unit

IgorNadj:
RN-721


rohan-bes (approved):
Nice one! Simple fix 😄

PR #4301, opened 2023-01-16
RN-736: Fixed LESMIS Admin Panel requiring BES Admin permissions

rohan-bes:
### Issue RN-736:

So looks like there was still code hanging around that enforced `BES Admin` permissions on the LESMIS Admin Panel. I've removed this, so that the front-end just requires `LESMIS Admin` permissions. However, the backend requires `Tupaia Admin Panel` permissions since we're talking to the `admin-panel-server`. Honestly, that seems fair enough to me, and I'd prefer not to change the server code.

So in short, to access the LESMIS Admin Panel, the user must have both: `LESMIS Admin`, and `Tupaia Admin Panel` permissions


billli0:
Thanks @rohan-bes  for fixing that, the logic looks solid.

Could you also change it in `DashboardItemsPage.js` and `MapOverlaysPage.js` in `package/lesmis`? It is required to edit viz through viz builder. I guess it can be always shown as the permission is already checked.

chris-pollard:
> 

@billli0 do you think this change will conflict with https://linear.app/bes/issue/MDEV-23/shift-viz-builder-access-from-bes-admin-to-viz-builder-user#comment-85e19e62 ?

billli0:
> > 
> 
> @billli0 do you think this change will conflict with https://linear.app/bes/issue/MDEV-23/shift-viz-builder-access-from-bes-admin-to-viz-builder-user#comment-85e19e62 ?

@chris-pollard Probably a question to @rohan-bes, but I would say yes if `Tupaia Admin Panel` permission group will be modified.

rohan-bes:
Good catch @billli0!

@chris-pollard are you happy to fixup the permissions of those two pages as part of your changes? I think as of our most recent decision we are scrapping the permissions check all together

billli0 (approved):

PR #4302, opened 2023-01-16
Add ajv validator

billli0:
Example
```
    import { MeditrakSurveyResponseSchema } from '@tupaia/types';
    import type { MeditrakSurveyResponse } from '@tupaia/types';
    import { getAjv } from '@tupaia/tsutils';

    const ajv = getAjv(models);
    const surveyResponseValidate = ajv.compile(MeditrakSurveyResponseSchema);
    const validatedSurveyResponseObject = (await surveyResponseValidate(
      surveyResponseObject,
    )) as MeditrakSurveyResponse;

    return validatedSurveyResponseObject;
```

It might be Typescript challenge to avoid using `as` to assign type to object, I tried but I've failed the mission :)


packages/tsutils/src/validation/ajv/getAjv.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { constructRecordExistsWithId } from '@tupaia/utils';
+import Ajv from 'ajv';
+import addFormats from 'ajv-formats';
+
+import { IdFormat } from './customFormats';
+
+type DatabaseModels = { entity: any; question: any; optionSet: any };
```
billli0, line 12: I guess we are fine to leave it as any? As DatabaseModelType only exists in database package.

packages/types/src/schemas.ts
```diff
@@ -0,0 +1,3280 @@
+export const PeriodTypeSchema = {
```
billli0, line 1: As long as EntityType is imported from '../../models' in types, all types in models will be converted too. Can't find a solution to avoid that.

packages/tsutils/src/validation/ajv/getAjv.ts
IgorNadj: Yeah sounds good, can deal with it later 👍

IgorNadj (approved):
LGTM!

PR #4307, opened 2023-01-17
MDEV-1: Add trap entity type to pacmossi project

chris-pollard:
### Changes:

Migrations:
- Add enum type 'trap' and add to entity model.
- Add 'trap' to canonical hierarchy in entity_hierarchy table.

### DIFF
--- a/packages/database/src/modelClasses/Entity.js
+++ b/packages/database/src/modelClasses/Entity.js
@@ -38,7 +38,6 @@ const LOCAL_GOVERNMENT = 'local_government';
 const MEDICAL_AREA = 'medical_area';
 const NURSING_ZONE = 'nursing_zone';
 const FETP_GRADUATE = 'fetp_graduate';
-const TRAP = 'trap';
 
 // Note: if a new type is not included in `ORG_UNIT_ENTITY_TYPES`, but data is to be stored against
 // it on DHIS2, a corresponding tracked entity type must be created in DHIS2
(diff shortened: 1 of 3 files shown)


packages/database/src/modelClasses/Entity.js
```diff
@@ -66,6 +67,7 @@ const ENTITY_TYPES = {
   MEDICAL_AREA,
   NURSING_ZONE,
   FETP_GRADUATE,
+  TRAP,
```
billli0, line 70: Adding new entity type here is not required now. See note: https://github.com/beyondessential/tupaia/blob/867d568f61759c3c524ff039c026f467d6cc6895/packages/database/src/modelClasses/Entity.js#L13

billli0 (changes requested):
@chris-pollard, there are some updates on adding entity type. PR: https://github.com/beyondessential/tupaia/pull/4200

With that, adding new type can be done by executing `alter type` query without any code changes, if new type is successfully added, it will be on `Entity Type` section in admin panel. 

Minor changes like adding canonical types on `entity_hierarchy` table, feel free to do it via DBeaver or Postico. 

Better approach would be to peer with a dev and do it on dev first, but it is up to you if you want to merge this PR to insert new type without any more work.

packages/database/src/modelClasses/Entity.js
chris-pollard: Ah thanks for pointing that out @billli0 . Will update the PR. I might keep in the canonical type for this one anyway.

chris-pollard:
Ok, that should be good now, thanks! @billli0

billli0 (approved):

chris-pollard:
Added in production db manually

PR #4314, opened 2023-01-19
OSC-10 Create Decision Tree component

kael89:
### Issue #:
OS-10

Note: this PR is against `dev` since everything is new functionality inside `ui-components`. This means it can be safely merged without testing if you want to.

### Changes:
I created a reusable **decision tree** component. This will be used for visualising entity hierarchies in the admin panel for now, but it should be reusable for other types of hierarchies too (e.g. dashboard items, map overlays).

* Each hierarchy layer renders in a column
* The top level (root) column specifies the hierarchy we are in
* Every other child column renders recursively
* If data changes in an ancestor column (e.g. a new hierarchy is selected), all expanded descendant columns are unmounted
* The component supporters filtering in each column

The data fetching mechanism is through a `fetchData` function that is passed as a prop, and shared between the components.

```ts
interface Node {
  id: string;
  name: string;
}

type Fetcher = (
  rootNode: Node,
  node: Node,
) => {
  id: string;
  name: string;
  children: Node[];
}[];

```

```js
const fetchData = async (rootNode, node) => getEntityDescendants(rootNode?.code, node?.code);

<DecisionTree fetchData={fetchData} />
```

See storybook for usage examples of the component 

### Screenshots:
I attached a demo video to the ticket


edmofro:
> Note: this PR is against dev since everything is new functionality inside ui-components. This means it can be safely merged without testing if you want to.

Love this!!

packages/ui-components/src/components/DecisionTree/index.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+export * from './DecisionTree';
```
kael89, line 6: Another name I considered is `HorizontalTree`

packages/ui-components/src/components/DecisionTree/Column.js
```diff
@@ -0,0 +1,143 @@
...
+import styled from 'styled-components';
+import IconButton from '@material-ui/core/IconButton';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
+import ChevronRightIcon from '@material-ui/icons/ChevronRight';
+import { ColumnFilter, useFilter } from './ColumnFilter';
+import { FetchLoader } from '../FetchLoader';
+
+const CLASS_NAME = 'TupaiaDecisionTreeColumn';
```
kael89, line 17: Inspired by material-ui classes

packages/ui-components/src/components/DecisionTree/Column.js
```diff
@@ -0,0 +1,143 @@
...
+export const Column = ({ data, isLoading, error, showExpandIcon, onSelect, onExpand }) => {
+  const { updateFilter, applyFilter, checkMatchesFilter } = useFilter();
+  const [selected, setSelected] = useState(undefined);
+
+  const filteredData = applyFilter(data, { field: 'name' });
+  const isError = !!error;
+
+  useEffect(() => {
+    setSelected(undefined);
+  }, [data]);
```
kael89, line 46: This means that we will the selected item when `data` changes , e.g. when a different entity is selected in an ancestor column

packages/ui-components/stories/decisionTree.stories.js
```diff
@@ -0,0 +1,110 @@
+/**
```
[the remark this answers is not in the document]
kael89: I loved using storybook, it makes iterative component development soooo much better - and encourages you to create reusable components!!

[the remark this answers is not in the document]
kael89: The only thing I am not sure of is whether you want to keep all you reusable components under that same package, or whether you want to break it into smaller packages eventually

packages/ui-components/src/components/DecisionTree/Column.js
```diff
@@ -0,0 +1,143 @@
...
+  return (
+    <>
+      <Container>
+        <ColumnFilter onChange={updateFilter} disabled={isLoading || isError} />
+        <ContentContainer className={`${CLASS_NAME}-content`}>
+          <FetchLoader isLoading={isLoading} isError={isError} error={error}>
+            <List>
+              {filteredData?.length === 0 && (
+                <ListItem>
+                  <ListItemText primary="No data found" color="primary" />
```
kael89, line 86: I chose this empty state message but should confirm with the designer

packages/ui-components/src/components/DecisionTree/ColumnFilter.js
```diff
@@ -0,0 +1,58 @@
...
+  padding: 20px 10px;
+  background: ${({ theme }) => theme.palette.grey['200']};
+  border-bottom: 1px solid ${({ theme }) => theme.palette.grey['400']};
+
+  .MuiFormControl-root {
+    margin: auto;
+  }
+`;
+
+export const useFilter = () => {
```
kael89, line 22: The filter is case insensitive

packages/ui-components/src/components/DecisionTree/DecisionTree.js
```diff
@@ -0,0 +1,102 @@
...
+          error={error}
+          onSelect={() => {
+            setSelectedRoot(undefined);
+            clearChildren();
+          }}
+          onExpand={rootNode => {
+            setSelectedRoot(rootNode);
+            fetchChildren(rootNode, rootNode);
+          }}
+          showExpandIcon={false}
```
kael89, line 80: Because the root component doesn't show expand icons, when we click on any of its items (hierarchies) we both select and expand it

packages/ui-components/src/hooks/useData.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { useCallback, useState } from 'react';
+
+export const useData = fetcher => {
```
kael89, line 8: This hooks is similar to the ones in react query, if built right this can be reused in many places

packages/ui-components/stories/story-utils/api/data.js
```diff
@@ -0,0 +1,153 @@
...
+  { id: id(), code: 'AU_Queensland_Brisbane', name: 'Brisbane' },
+  { id: id(), code: 'AU_Queensland_Cairns', name: 'Cairns' },
+  { id: id(), code: 'AU_Tasmania', name: 'Tasmania' },
+  { id: id(), code: 'AU_Victoria', name: 'Victoria' },
+  { id: id(), code: 'AU_Victoria_Ballarat', name: 'Ballarat' },
+  { id: id(), code: 'AU_Victoria_Melbourne', name: 'Melbourne' },
+  { id: id(), code: 'AU_Western_Australia', name: 'Western Australia' },
+
+  // Austria
+  { id: id(), code: 'AT', name: 'Austria' },
```
kael89, line 36: I thought a bit about using `faker` for all this data, but I realised that it returns random cities for a selected country.

Seeing Athens inside Austria (this time - could be Barcelona next time) would be confusing. We wouldn't be able to tell if this works or not :smile: 

This also tests the column overflow logic as the list of European countries is long

packages/ui-components/src/components/DecisionTree/index.js
rohan-bes: I think I actually prefer `HorizontalTree`. `DecisionTree` feels as though it implies some kind of meaning about the data within the tree (ie. it's used for making decisions, eg. 'if I perform action A, event B will happen')

rohan-bes (approved):
Looks great! Thanks so much for this @kael89 🎉

packages/ui-components/src/components/DecisionTree/index.js
kael89: Happy to rename it!

kael89:
@rohan-bes I renamed the component to `HorizontalTree` and implemented some changes on the selection based on an updated version of the mockups.

I will merge this in `dev` as soon as you re-approve, since it doesn't have any breaking changes

rohan-bes (approved):
Looks great, thanks for the name change @kael89

PR #4319, opened 2023-01-24
[no-issue]: Added data-table-api to the tupaia api-client

rohan-bes:
Just adding the `dataTableApi` to the `TupaiaApiClient` to allow for using it from other micro-services.


IgorNadj (approved):
👍

PR #4338, opened 2023-02-08
RN-733 Add preview to datatable

biaoli0:
### Issue #:
RN-733 Add preview to datatable

### Manual step
Change following constraint in `data-table` table: 
- `type` added NOT NULL
- `permission_groups` from default `{}` to no default

With that the type and schema for these two columns generated in `type` package will not contain `undefined` type.  

### Changes:
See commits. Suggest to follow commits history to review.


packages/data-table-server/src/routes/FetchPreviewData.ts
```diff
@@ -0,0 +1,39 @@
...
+      if (typeof row === 'object' && row) {
+        columnArray = new Set([...columnArray, ...Object.keys(row).map(key => key)]);
+      }
+    }
+
+    return {
+      rows: data.splice(0, LIMIT),
+      columns: Array.from(columnArray),
+      total: data.length,
+      limit: LIMIT,
```
biaoli0, line 36: It is devastating to pass options to `knex` to add limit and pagination. 

I think it should be fine to do it here. It could take ~30 secs to fetch all records though.

IgorNadj:
>Manual step
>Change following constraint in data-table table:
>
>type added NOT NULL
>permission_groups from default {} to no default
>With that the type and schema for these two columns generated in type package will not contain undefined type.

Do this with a migration?

packages/admin-panel-server/src/routes/FetchDataTablePreviewDataRoute.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { Request } from 'express';
+
+import { Route } from '@tupaia/server-boilerplate';
+
+export type FetchDataTablePreviewDataRequest = Request<
+  Record<string, never>,
+  Record<string, unknown>,
+  {
+    previewConfig: Record<string, unknown>;
```
IgorNadj, line 15: We should be able to type previewConfig. Add a type in @tupaia/types/src/types/requests/data-table-server/ that can be used by: admin-panel-server (here), data-table-server, and api-client.

packages/data-table-server/src/app/createApp.ts
```diff
@@ -22,6 +29,11 @@ export function createApp(database = new TupaiaDatabase()) {
       attachDataTableToContext,
       handleWith(FetchDataRoute),
     )
+    .post<FetchPreviewDataRequest>(
+      'dataTable/fetchPreviewData',
+      attachDataTableFromPreviewToContext,
```
IgorNadj, line 34: Why is this a middleware? Seems like it's just used once in this route. The other middleware is used multiple times.

biaoli0:
> > Manual step
> > Change following constraint in data-table table:
> > type added NOT NULL
> > permission_groups from default {} to no default
> > With that the type and schema for these two columns generated in type package will not contain undefined type.
> 
> Do this with a migration?

I would love to, but the type package is built before the migration run, unless the migration is one release prior to this PR.

packages/data-table-server/src/middleware/attachDataTableFromPreviewToContext.ts
```diff
@@ -0,0 +1,67 @@
...
+  next: NextFunction,
+) => {
+  try {
+    const { accessPolicy, models, ctx, body } = req;
+    const dataTableFields = validateDataTableFields(body);
+
+    const dataTable = await models.dataTable.generateInstance(dataTableFields);
+    const { code: dataTableCode } = dataTable;
+    const { permission_groups: permissionGroups } = dataTable;
+    if (
```
IgorNadj, line 42: Can DRY a lot of this with the existing attachDataTableToContext function. I suggest extracting this out to `validatePermissions(dataTable, permissionGroups)` and `getDataTableService(dataTable)`

packages/data-table-server/src/routes/FetchPreviewData.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+import { Request } from 'express';
+
+import { Route } from '@tupaia/server-boilerplate';
+import { DataTableType } from '../models';
+
+export type FetchPreviewDataRequest = Request<
+  Record<string, unknown>,
```
IgorNadj, line 12: Same as comment on FetchDataTablePreviewDataRequest above

packages/data-table-server/src/routes/FetchPreviewData.ts
```diff
@@ -10,7 +10,7 @@ import { DataTableType } from '../models';
 
 export type FetchPreviewDataRequest = Request<
   Record<string, unknown>,
-  { data: unknown[] },
+  { rows: unknown[]; columns: unknown[] },
```
IgorNadj, line 13: Does `columns` mean column headers?

packages/admin-panel/src/dataTables/query/useDataTablePreview.js
```diff
@@ -0,0 +1,95 @@
...
+        }
+      }
+    ],
+    "entityCode": "AU_SA%",
+    "permission_groups": [
+      "*"
+    ]
+  }
+*/
+const pullValidConfig = previewConfig => {
```
IgorNadj, line 53: I'm assuming this is a difference between the frontend and the backend? If so one of these should change so there is consistency 😄

packages/data-table-server/src/routes/FetchPreviewData.ts
IgorNadj: Oh yep so the idea was to prevent the 30 sec wait. The limit on the way back after the SQL query has completed doesn't do much.

Can we implement this in SqlDataTableService by wrapping the sql query:
```
const limit = 1000;
const wrappedResultQuery = 'SELECT * FROM ${inputSql} LIMIT ${limit}';
const wrappedCountQuery = 'SELECT count(*) FROM ${inputSql}';
``` 

The Postgres Query Plan Builder should be smart enough to optimise such these queries to be super quick.

IgorNadj (changes requested):
Huge PR! Looks great structurally, just a couple small requests to DRY stuff and a comment on the limit/count stuff.

packages/admin-panel-server/examples.http
```diff
@@ -174,3 +174,31 @@ Authorization: {{authorization}}
...
+    },
+    "additionalParameters": [
+      {
+        "name": "entityCode",
+        "config": {
+          "type": "string"
+        }
+      }
+    ],
+    "entityCode": "AU_SA%",
```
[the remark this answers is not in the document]
biaoli0: Good question. That is how to run query using `knex.js`, the value of the params are on root. I did not think too much but just follow the rules. I agree to wrap it inside a property, or wrap it in each `additionalParameter`.

packages/data-table-server/src/routes/FetchPreviewData.ts
biaoli0: Great example, thanks a lot!

packages/admin-panel-server/examples.http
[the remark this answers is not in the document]
biaoli0: Have added `runtimeParameters` to wrap this.

packages/data-table-server/src/app/createApp.ts
biaoli0: I am thinking to keep the logic consistent to the other middleware. Does it look confusing?

packages/data-table-server/src/routes/FetchPreviewData.ts
biaoli0: Yep.

IgorNadj (approved):

PR #4339, opened 2023-02-10
RN-758: Do not send email if SMTP is not set

biaoli0:
### Issue #:
RN-758, LESMIS don't want to set email server. It throws error if it is not set.

### DIFF
--- a/packages/central-server/src/utilities/sendEmail.js
+++ b/packages/central-server/src/utilities/sendEmail.js
@@ -10,6 +10,10 @@ const DEFAULT_SIGN_OFF = 'Cheers,\n\nThe Tupaia Team';
 export const sendEmail = async (to, subject, text, attachments, signOff = DEFAULT_SIGN_OFF) => {
   const { SMTP_HOST, SMTP_USER, SMTP_PASSWORD, SITE_EMAIL_ADDRESS } = process.env;
 
+  if ([SMTP_HOST, SMTP_USER, SMTP_PASSWORD, SITE_EMAIL_ADDRESS].includes(undefined)) {
+    return {};
+  }
+
   const transporter = nodemailer.createTransport({
     port: 465,
     host: SMTP_HOST,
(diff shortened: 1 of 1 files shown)


packages/central-server/src/utilities/sendEmail.js
```diff
@@ -10,6 +10,10 @@ const DEFAULT_SIGN_OFF = 'Cheers,\n\nThe Tupaia Team';
 export const sendEmail = async (to, subject, text, attachments, signOff = DEFAULT_SIGN_OFF) => {
   const { SMTP_HOST, SMTP_USER, SMTP_PASSWORD, SITE_EMAIL_ADDRESS } = process.env;
 
+  if ([SMTP_HOST, SMTP_USER, SMTP_PASSWORD, SITE_EMAIL_ADDRESS].includes(undefined)) {
+    return {};
+  }
```
rohan-bes, line 15:
```suggestion
  if (!SMTP_HOST || !SMTP_USER || !SMTP_PASSWORD || !SITE_EMAIL_ADDRESS) {
    return;
  }
```

Pretty minor suggestion here. I guess I find it a bit more readable this way? But no pressure.

Also for the return value, I'm almost certain we actually never use the return value of this function, so I think it's safe to just return nothing. 👍

rohan-bes (approved):
Minor suggestions, but aside from that looks good. This one will require a LESMIS release, yeah? If so we should flag that in the release PR

packages/central-server/src/utilities/sendEmail.js
biaoli0: Yep good suggestion, will apply that. This function expects to return a value, so might just keep the `return {};`

biaoli0:
> Minor suggestions, but aside from that looks good. This one will require a LESMIS release, yeah? If so we should flag that in the release PR

We don't have a proper LESMIS release process. Currently we will just rebase branch `lesmis-to-premise` to dev to inherit all new commits. Not sure if a flag can help.

packages/central-server/src/utilities/sendEmail.js
rohan-bes: I think the return value is just a `promise` to be `awaited`, which this will return regardless of whether we return a value or not due to it being an `async` function. But yes, I'm just being pedantic, I'm fine with whatever 👍

biaoli0: Oh not at all, thanks for sharing! I did try to use `return;`, but Eslint gets suspicious to say that this function should either return value or not. 🫠

rohan-bes: Ah, thanks for trying 🙏 I daren't question Eslint's guidance 😁

PR #4342, opened 2023-02-13
RN-743 Fix maximum call stack size in viz builder

biaoli0:
Spreading a long results to parameters list will cause oversize error, since the maximum size is 255

### Changes:

- Example

---


rohan-bes (approved):
Awesome find @biaoli0! This error is such a code-smell huh? I feel it's always a silly bug like this causing it 👍

biaoli0:
> Awesome find @biaoli0! This error is such a code-smell huh? I feel it's always a silly bug like this causing it 👍

Yes it's always a bug like this. Surprisingly `report-server` error log can locate where the error is in `data-broker`, while previously it was a bulk of meaningless errors. 

Perhaps it is related to Kostas' data-broker ts PR? If so we are blessed!

PR #4346, opened 2023-02-15
RN-762: Lesmis Admin Panel log-in defaults to Not Authorised view

chris-pollard:
### Changes:

- Update the Lesmis Admin Route to use isLesmisAdmin instead of isBESAdmin.


rohan-bes (commented):
@chris-pollard I think these changes are very similar to the changes in https://github.com/beyondessential/tupaia/pull/4301

I think with your approach we're relying on the `isLesmisAdmin` flag coming from the backend, whereas with my approach we are determining it on the front end. I'm pretty happy with either approach, but shall we merge these two tickets?

chris-pollard:
Cancelling as #4301 already covers this bug.

PR #4347, opened 2023-02-16
MDEV-105 superset pullMetadata

IgorNadj:
### Changes:

- Before only dhis and tupaia data services supported pullMetadata, the rest would throw an error. This fix changes it so that instead of throwing an error, these other services return no metadata.


chris-pollard (approved):
LGTM

PR #4350, opened 2023-02-20
Add data element data group metadata

biaoli0:
### Issue #:
RN-768
RN-769
RN-770
RN-771
### Changes
Add following records to data table: 
 

 tupaia_data_group_metadata | Fetches metadata for tupaia data group | {} | {*} | data_group_metadata
 
 dhis_data_group_metadata | Fetches metadata for DHIS2 data group | {} | {*} | data_group_metadata

 tupaia_data_elements_metadata | Fetches metadata for tupaia data elements | {} | {*} | data_elements_metadata

 dhis_data_elements_metadata | Fetches metadata for DHIS2 data elements | {} | {*} | data_elements_metadata


packages/data-table-server/src/dataTableService/services/DataGroupMetaDataDataTableService.ts
```diff
@@ -0,0 +1,84 @@
...
+  dataElements: { code: string; name: string; text: string; options: Record<string, unknown>[] }[];
+};
+
+type DataGroupMetadata = {
+  dataGroupCode: string;
+  dataGroupName: string;
+  dataElementCode: string;
+  dataElementName: string;
+  [key: string]: unknown;
+};
```
biaoli0, line 35: Do a bit data massage here to merge data group metadata with data element metadata, I find it handy to return an array instead of an object for viz-builder.

packages/data-table-server/examples.http
```diff
@@ -71,6 +71,44 @@ Authorization: {{authorization}}
...
+{
+    "dataGroupCode": "SPF"
+}
+
+
+### Fetch data from dhis_data_elements_metadata data-table
+POST http://{{host}}/{{version}}/dataTable/dhis_data_elements_metadata/fetchData HTTP/1.1
+content-type: {{contentType}}
+Authorization: {{authorization}}
+
```
biaoli0, line 98: I am actually on the fence between `data_elements_metadata` and `data_element_metadata` without the `s`...

rohan-bes (changes requested):
This all looks fantastic, nice work @biaoli0! Only change requested is to add some unit tests for `DataElementMetadataDataTableService` and `DataGroupMetadataDataTableService`

biaoli0:
> This all looks fantastic, nice work @biaoli0! Only change requested is to add some unit tests for `DataElementMetadataDataTableService` and `DataGroupMetadataDataTableService`

Oh I totally forgot, thanks for reminding!

packages/database/src/migrations/20230219232553-AddMetadataInDataTable-modifies-data.js
```diff
@@ -0,0 +1,43 @@
...
+
+exports.up = async function (db) {
+  const query = newTypes
+    .map(t => {
+      const tupaiaMetadataDescription = `Fetches metadata for tupaia ${t.split('_').join(' ')}`;
+      const dhisMetadataDescription = `Fetches metadata for DHIS2 ${t.split('_').join(' ')}`;
+      const metadataCode = `${t}_metadata`;
+      return `
+      INSERT INTO data_table VALUES ('${generateId()}','tupaia_${metadataCode}', '${tupaiaMetadataDescription}', '{}', '{"*"}', '${metadataCode}');
+      INSERT INTO data_table VALUES ('${generateId()}','dhis_${metadataCode}', '${dhisMetadataDescription}', '{}', '{"*"}', '${metadataCode}');
```
rohan-bes, line 29: What's the purpose of separating Tupaia and DHIS2 metadata data-tables here? Won't they behave the exact same, since we just pass the data-element codes to data-broker (which supports Tupaia and DHIS2)?

biaoli0: Yeah they are both supported, and we could use the same endpoint for them. However data broker does not accept data elements from different data service. I find it might be easier for users to avoid this exception.

rohan-bes: I think data-broker throws a fairly clear error message about this if it happens right? (And if not we could just improve the error messaging).
I think users should learn that they can only pull from one service type at a time, and then they can just do multiple pulls from the same data-table if they need data from multiple services 👍

biaoli0: Yeah fair point👍. Will apply.

biaoli0:
Heya @rohan-bes , are you able to re-review this PR? 

I've added the unit tests, and change to use `data_element_metadata` from `data_elements_metadata` as data table code. It is good to have a second look.

rohan-bes (approved):
Looks great! Thanks @biaoli0 🎉

PR #4353, opened 2023-02-20
OSC-10: Add Entity Relations Page in admin panel

kael89:
### Issue #:

OSC-10

### Changes:

This PR adds a new Entity Relations page in admin panel, which includes the entity hierarchy visualiser.

I added a `TreeResourcePage` component, drawing inspiration from the existing `ResourcePage`. That component uses [Horizontal Tree](https://github.com/beyondessential/tupaia/pull/4314) under the hood. Usage example:

```js
  <TreeResourcePage
    title="Entity Relations"
    endpoint="entityHierarchyTree"
    nodeKey="code"
    getHeaderEl={getHeaderEl}
  />
```

For now this component supports only a  handful of props required for this feature, but it can be extended in the future to be more customisable.

### Screenshots:

100% zoom on my laptop screen:
Screenshot from 2023-02-20 22-31-27

90% zoom so that you can see the whole component:
Screenshot from 2023-02-20 22-32-29

### Remaining scope

The only thing left is adding permission checks for the `entityRelations` endpoint in `central-server` (this was introduced in  a previous PR)


packages/admin-panel-server/src/app/createApp.ts
```diff
@@ -11,7 +11,7 @@ import {
 
 import { AdminPanelSessionModel } from '../models';
 import { hasTupaiaAdminPanelAccess } from '../utils';
-import { upload } from '../middleware';
+import { upload, verifyBESAdminAccess } from '../middleware';
```
kael89, line 14: I saw that this util was removed in https://github.com/beyondessential/tupaia/pull/4229. I re-instated it because if I'm not mistaken we need it for the two `entityHierarchyTree` endpoints that I have added in `admin-panel-server` for this feature.

Let me know if auth should be handled in a another way now @rohan-bes

kael89: By the way I didn't get any merge conflicts when I merged `dev` in the base branch, although in `dev` this util has been removed and my branch uses it :confused: 

I'm not quite sure why this happened...

packages/admin-panel-server/src/routes/FetchEntityHierarchyTreeRoute.ts
```diff
@@ -44,19 +38,16 @@ export class FetchEntityHierarchyTreeRoute extends Route<FetchEntityHierarchyTre
...
-      sort: ['entity.name'],
+      columns: ['parent_id', 'child_id'],
     });
     const childRelationsByParent = groupBy(childRelations, 'parent_id');
 
     return rootEntities.map(({ id, code, name }) => ({
+      id,
       code,
       name,
-      hierarchyName: code,
```
kael89, line 55: `hierarchyName` is not required eventually, neither are all other children props except for `id`

packages/admin-panel/src/pages/resources/EntityRelationsPage.js
```diff
@@ -0,0 +1,21 @@
...
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { TreeResourcePage } from './TreeResourcePage';
+
+export const EntityRelationsPage = ({ getHeaderEl }) => (
+  <TreeResourcePage
+    title="Entity Relations"
+    endpoint="entityHierarchyTree"
+    nodeKey="code"
```
kael89, line 14: In the tradition of `ResoucePage`, I expressed aspects of this resource's config as separate, simple props.

* `endpoint` is the base resource URL that will be used
* `nodeKey` is the field that serves as the identifier for the selected tree and node in the resource URL  (default: `id`)

packages/admin-panel/src/pages/resources/TreeResourcePage.js
```diff
@@ -0,0 +1,64 @@
...
+import { useApi } from '../../utilities/ApiProvider';
+
+const Container = styled(PageBody)`
+  overflow: auto;
+`;
+
+const StyledHorizontalTree = styled(HorizontalTree)`
+  margin-top: 40px;
+  margin-bottom: 40px;
+  max-height: 870px;
```
kael89, line 23: This height was chosen so that 20 items will be shown without having to use the component's scrollbar

packages/central-server/src/apiV2/entityRelations/GETEntityRelations.js
```diff
@@ -6,12 +6,6 @@
 import { GETHandler } from '../GETHandler';
 
 export class GETEntityRelations extends GETHandler {
-  // TODO we should allow a table to be joined multiple times (here for child_id and parent_id)
-  customJoinConditions = {
-    entity: ['entity.id', 'entity_relation.child_id'],
-    entity_hierarchy: ['entity_hierarchy.id', 'entity_relation.entity_hierarchy_id'],
-  };
-
   // TODO add permissions - see GetDashboardRelations
```
kael89, line 9: This TODO will be addressed in the next (and seemingly final) PR for this feature

packages/ui-components/src/components/HorizontalTree/HorizontalTree.js
```diff
@@ -36,7 +36,7 @@ const DescendantsContainer = styled.div`
   border-radius: 5px 5px 0px 0px;
 `;
 
-export const HorizontalTree = ({ fetchData }) => {
+export const HorizontalTree = ({ fetchData, className }) => {
```
kael89, line 39: This allows us to use `styled(HorizontalTree)`

packages/admin-panel/src/pages/resources/EntityRelationsPage.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { TreeResourcePage } from './TreeResourcePage';
+
+export const EntityRelationsPage = ({ getHeaderEl }) => (
+  <TreeResourcePage
+    title="Entity Relations"
+    endpoint="entityHierarchyTree"
```
kael89, line 13: I'd like another opinion on this endpoint's name. I introduced it in a previous PR but the fact that it doesn't match the resource page ("Entity Relations") is kind of a smell.

Other names that could be considered:
1. `entityHierarchy`
2. `entityRelationsTree`

I am thinking that perhaps "entity hierarchy" is a tree of "entity relation" nodes.

I'm leaning towards `entityRelationsTree`. It may be nice to start a convention for naming tree-compatible endpoints using `{nodeType}Tree`.

packages/admin-panel-server/src/routes/FetchEntityHierarchyTreeRoute.ts
```diff
@@ -9,16 +9,10 @@ import { Route } from '@tupaia/server-boilerplate';
 import { groupBy } from 'lodash';
 import { getSortByKey } from '@tupaia/utils';
 
-interface EntityNode {
-  code: string;
-  name: string;
-  hierarchyName: string;
-}
-
 export type FetchEntityHierarchyTreeRequest = Request<
```
rohan-bes, line 12: I know it's a little late in the game, and that we've discussed how annoying it is when people make re-design comments too late, but could I make a suggestion here? 😅 

I guess I feel that it's really a front-end concern that the data is structured as a tree. So when it comes to the data formatting I feel it makes sense if that's done on the front-end.

This leaves the logic of dispatching the request to the appropriate micro-service depending on if this is a root level request or a branch level request (ie. if `hierarchyName` and `entityCode` are present). I was thinking it might make sense for the `HorizontalTree` component to actually have 2 parameters; `rootEndpoint` and `branchEndpoint` (or maybe `drillEndpoint`). That way the frontend itself can make the appropriate request.

So there'd be no more need for this `FetchEntityHierarchyTreeRoute`. We may however need to add a `EntityDescendantsRoute` to the `admin-panel-server` that ferries the request to the `entity-api`.

Thoughts?

---

Lastly, as a side note, I wouldn't mind if we added a `/hierarchies` endpoint to the `entity-server` which returned all the entity hierarchies that the user has access to. It feels a little funny that we need to request from the `central-server` to determine which hierarchies exist, and then fetch relations via the `entity-server`

packages/admin-panel/src/pages/resources/EntityRelationsPage.js
rohan-bes: I like the page being called `EntityHierarchiesPage`? I guess the semantics in my mind are:
- entity hierarchy: the hierarchically structured collection of entities
- entity relations: given relations between entities

So if the page were a flat list of relations, I'd argue we call it `EntityRelationsPage`. But since it's a hierarchical view, `EntityHierarchiesPage` seems more apt

rohan-bes (changes requested):
Bit of a request to change how we manage the endpoints for the HorizontalTree, lemme know your thoughts 👍 

But it's all looking great! Thanks so much @kael89 !

packages/admin-panel/src/pages/resources/EntityRelationsPage.js
kael89: I agree, I think we discussed naming the page `Entity Relations` early on but `Entity HIerarchies` seems more appropriate.

packages/admin-panel-server/src/routes/FetchEntityHierarchyTreeRoute.ts
```diff
@@ -66,21 +57,19 @@ export class FetchEntityHierarchyTreeRoute extends Route<FetchEntityHierarchyTre
 
     const descendants = await entityApi.getDescendantsOfEntity(hierarchyName, entityCode, {
       filter: { generational_distance: { comparator: '<=', comparisonValue: 2 } },
```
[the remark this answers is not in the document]
kael89: Yeap, I added it + integration tests in a [previous PR](https://github.com/beyondessential/tupaia/pull/4275)

packages/admin-panel-server/src/routes/FetchEntityHierarchyTreeRoute.ts
kael89: About the last side note: I do agree, adding this endpoint would be nice. If you don't mind, I will create a new ticket and start working on it so that it can be merged separately and faster

kael89: About `HorizontalTree` having separate props for root and branch data: I think that's a very good call, since each callback is called with different input depending on the level (no input for root, `rootNode` and `node` for branch).

Will change :+1:

kael89: Finally, about your main comment: what you say makes sense. In general, the question "where should this live - frontend or backend"? is not a trivial one to answer. It depends on many factors:

* Do you have a dedicated backend for your frontend? (Tupaia generally follows this pattern)
* Do you think you may reuse this API? Would it be a candidate for a public REST API e.g.?
* Any complexity/performance/security concerns that would make moving the logic in the backend more sensible?

...and others.

In this case, I guess that if you don't expect to reuse the tree API elsewhere (e.g. Tupaia frontend hierarchies), it may be better to construct it in the frontend.

In the current implementation, I was influenced by the existing pattern in `admin-panel` where each resource has a dedicated endpoint in the backend. However, we don't have to follow this pattern here. We can define the data fetchers in `admin-panel`, and pass them as props in `TreeResourcePage`:


```js
// @admin-panel/src/pages/resources/TreeResourcePage.js

// Note: this tree can be used for other resources: map overlays, dashboard charts etc
export const TreeResourcePage = ({ fetchRootData, branchFetcher }) =>
  <HorizontalTree
    fetchRootData={fetchRootData}
    fetchBranchData={fetchBranchData}
  />
```

```js
// @admin-panel/src/api/entityHierarchy.js

export const fetcHierarchiesWithChildren = () => {};
export const fetchEntityChildrenAndGrandchildren = (rootNode, node) => {};
```

```js
// @admin-panel/src/pages/resources/EntityRelationsPage.js

import {
  fetcHierarchiesWithChildren,
  fetchEntityChildrenAndGrandchildren
} from '@admin-panel/src/api/entityHierarchy.js';

export const EntityHierarchiesPage = () => (
  <TreeResourcePage
    title="Entity Relations"
    fetchRootData={fetcHierarchiesWithChildren}
    fetchBranchData={fetchEntityChildrenAndGrandchildren}
  />
);
```

kael89:
@rohan-bes I have implemented all requested changes and merged the latest dev. Ready to re-review!

packages/ui-components/stories/story-utils/api/entity.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { getSortByKey } from '@tupaia/utils';
+import { ENTITIES, ENTITY_HIERARCHY_RELATIONS } from './data';
+
+const getEntity = code => {
```
kael89, line 9: See diff with `getEntityDescendants.js`

packages/admin-panel/src/routes.js
```diff
@@ -200,11 +200,12 @@ export const ROUTES = [
         to: '/entityTypes',
         component: EntityTypesPage,
       },
-      {
-        label: 'Entity Relations',
-        to: '/entityRelations',
-        component: EntityRelationsPage,
-      },
+      // TODO OSC-10 Uncomment to enabled Entity Hierarchies
```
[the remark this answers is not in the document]
kael89: Demo video with this block uncommented: https://linear.app/bes/issue/OSC-17#comment-c9129f72

packages/server-boilerplate/src/utils/useForwardUnhandledRequests.ts
```diff
@@ -28,9 +32,11 @@ export const useForwardUnhandledRequests = (
   target: string,
   prefix?: string,
   attachSession: Middleware = defaultAttachSession,
+  router?: ProxyOptions['router'],
```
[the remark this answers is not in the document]
rohan-bes: I actually have a plan of how we could rework how we use this middleware.
Ideally, we'd do it a bit more elegantly using the inbuilt `app.use()` function of express, and do it in the `ApiBuilder`.

I'll create a PR separate to this one and tag you in it for reference 👍

rohan-bes (approved):
Great stuff @kael89 🙏 

I'll make the changes from my comments myself, and then do a quick regression test with the feature flag disabled. If that goes well I'll merge it and we'll create a ticket for feature testing as a separate process 👍

packages/admin-panel/src/routes.js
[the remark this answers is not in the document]
rohan-bes: So the feature flags implementation relies on env vars. I think I forgot about the differences between browser and node env var management at the time of implementation.

It's a bit awkward, so I'll just go with a variable (the comment version was causing linting errors)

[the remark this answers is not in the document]
kael89: Nice! Feature flags are awesome - especially for changes like this PR, where there is no refactoring involved.

rohan-bes:
@kael89 This feature is shipping in the next release! Thanks so much for this, it looks great!

PR #4354, opened 2023-02-21
RN-781: Update fetchData config migration with correct property names

chris-pollard:
### Changes:

- Change 'dataElements' to 'dataElementCodes'
- Change 'dataGroups' to 'dataGroupCode'
- Change any aggregation element that is a string to an object.


rohan-bes (approved):
Thanks so much for this! Champion @chris-pollard!

PR #4355, opened 2023-02-21
RN-782: dataTableCode Migration Checks Against Wrong Data Group Key

chris-pollard:
### Changes:

- Look for 'dataGroupCode' instead of 'dataGroup' in parameters to determine 'events' data table.


rohan-bes (approved):
TYVM! 🙏

PR #4357, opened 2023-02-22
RN-787 Fix dhis data element metadata

biaoli0:
### Issue #:
RN-787

### DIFF
--- a/packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
+++ b/packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
@@ -6,7 +6,7 @@
 import groupBy from 'lodash.groupby';
 
 import type { DhisApi } from '@tupaia/dhis-api';
-import { DataElementMetadata, DataElementModel, DhisMetadataObject } from '../../../types';
+import { DataElementMetadata, DataElementModel } from '../../../types';
 import { DataElement } from '../types';
 import { DhisTranslator } from '../translators';
 import type { PullMetadataOptions as BasePullMetadataOptions } from '../../Service';
@@ -38,7 +38,7 @@ export class DataElementsMetadataPuller {
     options: PullDataElementsOptions,
   ) => {
     const dataSourcesByDhisType = this.groupDataSourcesByDhisDataType(dataSources);
-    let metadata: DhisMetadataObject | DataElementMetadata[] = [];
+    let metadata: DataElementMetadata[] = [];
 
     for (const entry of Object.entries(dataSourcesByDhisType)) {
       const [dhisDataType, groupedDataSources] = entry;
--- a/packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
+++ b/packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
@@ -6,7 +6,7 @@
 import groupBy from 'lodash.groupby';
 
 import type { DhisApi } from '@tupaia/dhis-api';
-import { DataElementModel } from '../../../types';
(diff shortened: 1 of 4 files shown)


packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
```diff
@@ -38,25 +38,26 @@ export class DataElementsMetadataPuller {
...
     for (const entry of Object.entries(dataSourcesByDhisType)) {
       const [dhisDataType, groupedDataSources] = entry;
       const dataElementCodes = groupedDataSources.map(({ dataElementCode }) => dataElementCode);
       if (dhisDataType === this.dataSourceModel.getDhisDataTypes().INDICATOR) {
         const indicators = await api.fetchIndicators({ dataElementCodes });
-        metadata.push(
-          ...this.translator.translateInboundIndicators(indicators, groupedDataSources),
-        );
+        const results = this.translator.translateInboundIndicators(indicators, groupedDataSources);
+        metadata = metadata.concat(results);
```
chris-pollard, line 49: @biaoli0 since the 'metadata' variable is mutable, for this assignment, could you just do:
`metadata = results`
since we know it will be an empty array?

packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
```diff
@@ -38,25 +38,26 @@ export class DataElementsMetadataPuller {
     options: PullDataElementsOptions,
   ) => {
     const dataSourcesByDhisType = this.groupDataSourcesByDhisDataType(dataSources);
-    const metadata = [];
+    let metadata: DhisMetadataObject | DataElementMetadata[] = [];
```
chris-pollard, line 41: Why is the metadata defined as an object or an array here?

packages/data-broker/src/services/dhis/translators/DhisTranslator.ts
```diff
@@ -207,19 +207,33 @@ export class DhisTranslator {
     return translateElementKeysInEventAnalytics(eventAnalytics, dataElementToSourceCode);
   };
 
+  // Override metadata code by data element code
   public translateInboundDataElements = (
-    dataElements: DhisMetadataObject[],
-    dataSources: DataElement[],
+    metadata: DhisMetadataObject[],
+    dataElements: DataElement[],
```
chris-pollard, line 213: The naming of these parameters is a little confusing, since the first argument is data elements and the second is data sources.

chris-pollard (approved):
LG. Couple of questions but overall makes sense.

packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
biaoli0: This is within the for loop, which makes it an accumulated results.

packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
biaoli0: Yeah you are right. It should be DataElementMetadata[]. Thanks!

packages/data-broker/src/services/dhis/translators/DhisTranslator.ts
biaoli0: Yeah it was. That's one of the reasons why I want to rename it properly. Another reason is we don't have `dataSources` anymore.

packages/data-broker/src/services/dhis/pullers/DataElementsMetadataPuller.ts
chris-pollard: ah gotcha!

PR #4359, opened 2023-02-24
RN-783 Remove dataElementCodeToName alias

biaoli0:


packages/database/src/migrations/20230222233428-RewriteVizsUsedDataElementCodeToName-modifies-schema.js
```diff
@@ -0,0 +1,127 @@
...
+
+    let newTransform = transform
+      .map(t => {
+        if (JSON.stringify(t).includes('dataElementCodeToName(')) {
+          return [
+            fetchMetaData,
+            excludeUnusedMetadata,
+            renameDataElementName,
+            t,
+            removeDataElementMetadata,
```
biaoli0, line 74: The idea is to add `fetchMetaData`, `excludeUnusedMetadata`, `renameDataElementName` in front of every transform that has `dataElementCodeToName`, then remove it afterward.

I know this is annoying, but if the metadata is added at the beginning of the transforms, it will crash most of the reports because it introduces a new column.

chris-pollard (approved):
Looks good, nice one!

PR #4362, opened 2023-02-27
Update Migration: Convert Data Group Code To String

chris-pollard:
### Changes:

- Change the transform conversion to ensure that dataGroupCode is a string and not an array.


rohan-bes (approved):
Nice fix @chris-pollard!

PR #4365, opened 2023-02-28
[no-issue]: Fix test database setup script

kael89:
I was working on something and bumped into this. Postgis was not present, which means the `entity` table could not be created.

This happened in a machine with no tupaia DB present. My original goal was to run integration tests for `entity-server`.

I suspect that this is an issue only if `CREATE EXTENSION postgis` hasn't run again for `tupaia_test` in that particular host. I ran the setup script again by commenting out the line I added, and this time it succeeded 🤷 

I can't understand how this works since the setup script drops the db before re-creating it. It's as if it persists the fact that Postgis was once enabled for `tupaia_test`.


rohan-bes (approved):
Thanks @kael89

packages/database/scripts/setupTestDatabase.sh
```diff
@@ -14,6 +14,7 @@ fi
 
 PGPASSWORD=$DB_PG_PASSWORD psql -h $DB_URL -p $DB_PORT -U $DB_PG_USER -c "DROP DATABASE IF EXISTS $DB_NAME"
 PGPASSWORD=$DB_PG_PASSWORD psql -h $DB_URL -p $DB_PORT -U $DB_PG_USER -c "CREATE DATABASE $DB_NAME WITH OWNER $DB_USER"
+PGPASSWORD=$DB_PG_PASSWORD psql -h $DB_URL -p $DB_PORT -U $DB_PG_USER -c "CREATE EXTENSION postgis"
```
rohan-bes, line 17: I think we need to do a check to see if postgis already exists. Should be doable using something like:
```
SELECT * FROM pg_extension WHERE extname = 'postgis';
```

rohan-bes (changes requested):

rohan-bes (approved):

PR #4367, opened 2023-03-02
Fix data table content overflow

chris-pollard:
### Issue NO-ISSUE:

The problem is that when there are lots of columns of data, the preview component flows way over and also breaks the json editor when toggled.

### Changes:

- Add min-width property equal to 0, which fixes data table content overflow problem.


IgorNadj (approved):
Thanks for that, much better now!

PR #4368, opened 2023-03-03
Trial Data Grid For Viz Builder

chris-pollard:
### Issue #RN-801:

Currently the Viz Builder preview panel for viewing data uses the same react-table code  / material-ui table component as other views in the admin panel. Not only are these components a little old, the look and feel is not appropriate for data viewing. The boxes are too big and there are no built in sorting/filtering abilities.

Introducing the MUI X Data Grid from Material-Ui. Marketed as performant advanced components, Mui x components are meant to be a little more feature rich and advanced technically than the regular mui components. Some of the components are premium but there is a base model of the data grid which is MIT License and available to use.

It is quite sophisticated and easy to use. The only real props it requires is row and column data, but there is an impressive number of other inputs in the API, including sort and filter configuration.

The features which could benefit viz builder the most are:
- sorting out of the box
- filter on keyword, out of the box
- pagination and items per page, out of the box
- Three size and display settings, which make big data sets very managable.
- Horizontal and Vertical scrolling is built in, and is beautiful. 

One downside is that the free version doesn't include user column resizing, but since the size of the cells and data is much smaller and managable, it's not a huge loss. In fact its not a loss at all because you can't resize currently!

### Changes:

- Update all frontend packages to be compatible with react and react-dom 17.
- Add Data Grid and storybook in ui-components
- Add Data Grid to Viz Builder
---

### Screenshots:
Screen Shot 2023-03-03 at 1 56 22 pm


chris-pollard:
Cancelled, this was for demo purposes only

PR #4370, opened 2023-03-07
RN-756 Set default value for measureLevel

biaoli0:
### Issue #:
RN-756


packages/web-frontend/src/selectors/measureSelectors.js
```diff
@@ -248,7 +248,7 @@ export const selectAreMeasuresOnTheSameEntityLevel = createSelector(
     }
 
     const measureLevelsFromMeasures = Object.values(measureInfo)
-      .map(({ measureLevel }) => measureLevel)
+      .map(({ measureLevel = [] }) => measureLevel)
```
biaoli0, line 251: I cannot find a way to reproduce the error, but here is the only place to read `measureLevel` from an object.

chris-pollard (approved):
Nice spot Biao!

PR #4371, opened 2023-03-08
Convert external database connections to drop list

biaoli0:
### Issue #:
https://linear.app/bes/issue/RN-729#comment-8ca52151

### Changes:

- Example

---


packages/central-server/src/apiV2/index.js
```diff
@@ -230,6 +233,7 @@ apiV2.get('/dataServiceSyncGroups/:recordId?', useRouteHandler(GETSyncGroups));
 apiV2.get('/dataServiceSyncGroups/:recordId/logs', useRouteHandler(GETSyncGroupLogs));
 apiV2.get('/dataServiceSyncGroups/:recordId/logs/count', useRouteHandler(GETSyncGroupLogsCount));
 apiV2.get('/dataElementDataServices/:recordId?', useRouteHandler(BESAdminGETHandler));
+apiV2.get('/externalDatabaseConnections', useRouteHandler(GETExternalDatabaseConnections));
```
IgorNadj, line 236: This route is already handled by the line below, `:recordId?` means optional, so if you don't pass it in it returns a list.

biaoli0: I didn't know that. Thanks!

IgorNadj (approved):

PR #4374, opened 2023-03-09
RN-773 Rewrite viz using entity relations

IgorNadj:
RN-773

This PR contains a few cleanups and the removal of the insertNumberOfFacilitiesColumn transform which is the previous method of accessing entity relations.


packages/entity-server/src/routes/hierarchy/middleware/attachEntityContext.ts
```diff
@@ -98,22 +100,15 @@ export const attachSingleEntityContext = async (
 };
 
 export const attachMultiEntityContext = async (
-  req: Request<
-    { hierarchyName: string },
-    any,
-    { entities: string[] | undefined },
-    { filter?: string }
-  > & {
+  req: Request<{ hierarchyName: string }, any, { entities: string[] }, { filter?: string }> & {
```
IgorNadj, line 103: `entities` can technically be undefined, but not after the validate step which is the first line of this function.

This is a classic typing + validation question, where we can receive a request of any shape, but after validation it is a well defined shape. 

My thinking on this is to move the validation out as early as possible, so the route handler has typings of the well defined shape. We should be able to pass a schema into the server-boilerplate use/get/post etc methods as an optional param which simply calls `await ajvValidate(schema, req.body);`. And gradually these schemas and typings can be moved to @tupaia/types. The schemas are auto generated from the request types, and the types can be reused in api-client as well. @rohan-bes thoughts? 

I didn't in this case as I thought it would be going too far for a _tidy as you go_ in a PR.

packages/entity-server/src/routes/hierarchy/types.ts
```diff
@@ -20,7 +20,21 @@ export interface MultiEntityRequestParams {
...
+  entities: string[];
+};
+
+export const MultiEntityRequestBodySchema = {
+  type: 'object',
+  properties: {
+    entities: {
+      type: 'array',
+      items: {
+        type: 'string',
```
IgorNadj, line 32: Had a hard to debug error come out because the entities array was provided, but the value was null instead of a string. Swapped to json schema to improve error messaging.

chris-pollard (approved):
Curious how you would get the facility count in new world. Fetch entity hierarchy table with descendant type and count all rows with values?

IgorNadj:
> Curious how you would get the facility count in new world. Fetch entity hierarchy table with descendant type and count all rows with values?

Yep exactly that. I've got the conversion in a comment in the card.

chris-pollard (commented):
@IgorNadj another question, where are the vizes rewritten? I was expecting a migration.

chris-pollard:
> @IgorNadj another question, where are the vizes rewritten? I was expecting a migration.

Sorry @IgorNadj realised that you're tracking this in the card.

PR #4375, opened 2023-03-09
RN-688: Remove Output Context

chris-pollard:
### Changes:

- Remove output context, which only impacted RawDataExport type output. 
- Remove outputContext reference in tests

### DIFF
--- a/packages/report-server/src/__tests__/reportBuilder/output/rawDataExport.test.ts
+++ b/packages/report-server/src/__tests__/reportBuilder/output/rawDataExport.test.ts
@@ -27,8 +27,7 @@ describe('rawDataExport', () => {
     });
   });
 
-  // legacy test from outputContext for raw data table output. Due to this being deprecated, now the output will behave exactly the same as rows and columns.
-  it('Does not add the rest of data elements that has no value into columns', async () => {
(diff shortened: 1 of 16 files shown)


packages/report-server/src/__tests__/reportBuilder/output/rawDataExport.test.ts
```diff
@@ -27,7 +27,8 @@ describe('rawDataExport', () => {
     });
   });
 
-  it('Add the rest of data elements that has no value into columns', async () => {
+  // legacy test from outputContext for raw data table output. Due to this being deprecated, now the output will behave exactly the same as rows and columns.
```
IgorNadj, line 30: What makes this test legacy?

chris-pollard: @IgorNadj this is legacy because previously with output context, even if not all data elements had values, all data elements from data group would become columns. Where as now, it doesn't care about data group or any extra context, the only columns that get printed are those literally in the report following transforms.

IgorNadj: Discussed, test covers how old behaviour changed to new

IgorNadj (approved):

PR #4376, opened 2023-03-09
Set lookup even with no row

biaoli0:
### Issue #:
Fix RN-783 testing issues, see issue links on commit messages.

### DIFF
--- a/packages/data-table-server/src/dataTableService/services/DataElementMetadataDataTableService.ts
+++ b/packages/data-table-server/src/dataTableService/services/DataElementMetadataDataTableService.ts
@@ -10,7 +10,7 @@ import { yup } from '@tupaia/utils';
 import { DataTableService } from '../DataTableService';
 
 const requiredParamsSchema = yup.object().shape({
-  dataElementCodes: yup.array().of(yup.string().required()).required(),
+  dataElementCodes: yup.array().of(yup.string().required()).default([]),
 });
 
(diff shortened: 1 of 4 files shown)


chris-pollard (approved):

packages/data-table-server/src/dataTableService/services/DataElementMetadataDataTableService.ts
```diff
@@ -10,7 +10,7 @@ import { yup } from '@tupaia/utils';
 import { DataTableService } from '../DataTableService';
 
 const requiredParamsSchema = yup.object().shape({
-  dataElementCodes: yup.array().of(yup.string().required()).required(),
+  dataElementCodes: yup.array().of(yup.string().required()).default([]),
```
IgorNadj, line 13: Test fail shows this is actually meant to be required?

IgorNadj:
Is Issue 3 fixed by either of these commits?

packages/data-table-server/src/dataTableService/services/DataElementMetadataDataTableService.ts
biaoli0: I suggest we remove the unit test and make it accept dataElementCodes as undefined.

IgorNadj: Yep sounds good

PR #4382, opened 2023-03-14
OSC-19: Add GET entityRelations endpoint in central-server

kael89:
### Issue #:
https://linear.app/bes/issue/OSC-19

This is a prerequisite for the admin-panel entity visualiser (OSC-10)


packages/central-server/src/apiV2/entityRelations/assertEntityRelationPermissions.js
```diff
@@ -0,0 +1,54 @@
...
+
+  if (!hasBESAdminAccess(accessPolicy)) {
+    // Our permissions check logic is simplified based on the following assumptions:
+    // 1. Each child is associated to a non-empty subset of its parent's countries
+    // 2. Projects cannot be children of other entities
+    //
+    // All valid relations should satisfy the above invariants.
+    // Note that invalid relations can expose parents that the user doesn't have permissions to,
+    // e.g. entityInCountryA -> countryB, where the user only has permissions to countryB
+    dbConditions['child.country_code'] = mergeFilter(
```
kael89, line 36: I don't think that `mergeFilter` is really required here,, since

* There is no `child` table in the DB currently
* Users can filter based on columns from other tables, but I don't think they can use aliases for them

However, it's cheap to use and it will be more robust in case any of the above changes in the future

packages/central-server/src/apiV2/entityRelations/assertEntityRelationPermissions.js
```diff
@@ -0,0 +1,54 @@
...
+
+  return true;
+};
+
+export const createEntityRelationDbFilter = (accessPolicy, criteria, options) => {
+  const dbConditions = { ...criteria };
+  const dbOptions = { ...options };
+
+  if (!hasBESAdminAccess(accessPolicy)) {
+    // Our permissions check logic is simplified based on the following assumptions:
```
kael89, line 29: Hopefully this makes sense? The alternative would be to use a quite convoluted RAW criteria logic similar to the one in [assertDashboardRelationsPermissions.js#L141](https://github.com/beyondessential/tupaia/blob/dev/packages/central-server/src/apiV2/dashboardRelations/assertDashboardRelationsPermissions.js#L141), that checks permissions to the associated country (non-project) / any associated coumntry (project).

What's more, this should be done for both `child_id` and `parent_id`.

rohan-bes (approved):
Nice one @kael89! Looks good to me, just a question about changing the dhis2instance endpoint permissions

packages/central-server/src/apiV2/index.js
```diff
@@ -220,12 +221,13 @@ apiV2.get('/dataTables/:recordId?', useRouteHandler(GETDataTables));
...
 apiV2.get('/entities/:recordId?', useRouteHandler(GETEntities));
 apiV2.get('/entities/:parentRecordId/surveyResponses', useRouteHandler(GETSurveyResponses));
+apiV2.get('/entityRelations/:recordId?', useRouteHandler(GETEntityRelations));
 apiV2.get('/countries/:recordId?', useRouteHandler(GETCountries));
 apiV2.get('/clinics/:recordId?', useRouteHandler(GETClinics));
 apiV2.get('/facilities/:recordId?', useRouteHandler(GETClinics));
 apiV2.get('/geographicalAreas/:recordId?', useRouteHandler(GETGeographicalAreas));
 apiV2.get('/reports/:recordId?', useRouteHandler(GETReports));
-apiV2.get('/dhisInstances/:recordId?', useRouteHandler(TupaiaAdminGETHandler));
+apiV2.get('/dhisInstances/:recordId?', useRouteHandler(BESAdminGETHandler));
```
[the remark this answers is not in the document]
kael89: Oh man I have no clue how this ended up here!! 🙃 

Great catch, will revert

PR #4387, opened 2023-03-17
RN-802 Default parameters on preview

biaoli0:
### Issue #:
RN-802
 
### Changes:
1. Add a couple of Fields for builtInParams, see: packages/admin-panel/src/dataTables/components/PreviewFilters/filters/FilterTypeOptions.js
2. Rename variables and functions from *Parameters (e.g. useParameters) to *Params to be consistent with data-table-server
3. Refactor Autocomplete to ReduxAutocomplete

### DIFF
--- a/packages/admin-panel/src/autocomplete/Autocomplete.js
+++ b/packages/admin-panel/src/autocomplete/Autocomplete.js
@@ -1,18 +1,14 @@
-/**
- * Tupaia MediTrak
- * Copyright (c) 2018 Beyond Essential Systems Pty Ltd
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import throttle from 'lodash.throttle';
-import { connect } from 'react-redux';
-import MuiChip from '@material-ui/core/Chip';
 import { createFilterOptions } from '@material-ui/lab/Autocomplete';
 import styled from 'styled-components';
-import { Autocomplete as AutocompleteBase } from '@tupaia/ui-components';
-import { getAutocompleteState } from './selectors';
-import { changeSelection, changeSearchTerm, clearState } from './actions';
+import MuiChip from '@material-ui/core/Chip';
+import { Autocomplete as UIAutocomplete } from '@tupaia/ui-components';
 
 const Chip = styled(MuiChip)`
   &:first-child {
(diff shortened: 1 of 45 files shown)


packages/admin-panel/src/autocomplete/Autocomplete.js
```diff
@@ -1,198 +1,133 @@
...
+        renderTags: (values, getTagProps) =>
+          values.map((option, index) => (
+            <Chip
+              color="primary"
+              label={optionLabelKey ? option[optionLabelKey] : option}
+              {...getTagProps({ index })}
+            />
+          )),
+      }
+    : {};
```
biaoli0, line 76: These parts can be reused for multiple selections.

packages/admin-panel/src/autocomplete/Autocomplete.js
```diff
@@ -1,198 +1,133 @@
...
-    return Array.isArray(placeholder) ? placeholder.join(', ') : placeholder;
-  }
-  return 'Start typing to search';
-};
-
-const filter = createFilterOptions();
-
-const AutocompleteComponent = React.memo(
-  ({
+export const Autocomplete = props => {
```
biaoli0, line 19: This components will accept options from props. This brings some flexibilities to fetch options either from Redux dispatch or use query.

packages/admin-panel/src/autocomplete/Autocomplete.js
```diff
@@ -1,198 +1,133 @@
...
-import throttle from 'lodash.throttle';
-import { connect } from 'react-redux';
-import MuiChip from '@material-ui/core/Chip';
 import { createFilterOptions } from '@material-ui/lab/Autocomplete';
 import styled from 'styled-components';
-import { Autocomplete as AutocompleteBase } from '@tupaia/ui-components';
-import { getAutocompleteState } from './selectors';
-import { changeSelection, changeSearchTerm, clearState } from './actions';
+import MuiChip from '@material-ui/core/Chip';
+import { Autocomplete as UIAutocomplete } from '@tupaia/ui-components';
```
chris-pollard, line 11: I wonder if by convention, the name should be UiAutocomplete. Having the first word in caps doesn't seem to fall in line with the way we name react components

chris-pollard (approved):
Nice refactoring work!

packages/admin-panel/src/autocomplete/Autocomplete.js
biaoli0: We will do that when it comes to abbreviation, e.g. PDFExportPage

PR #4392, opened 2023-03-21
RN-810 Convert RH orgUnitCodeToName Reports

chris-pollard:
### Changes:

- Migration to replace orgUnitCodeToName function with entities data table fetch.


IgorNadj (approved):
Nice one, looks great.

PR #4394, opened 2023-03-22
NONE Fix wrong import

kael89:
Fixes a wrong import that I introduced in https://github.com/beyondessential/tupaia/pull/438


rohan-bes (approved):
Okay, this time approving with the correct github user 😅

PR #4395, opened 2023-03-22
Prevent fetchData error when entityCodes undefined

IgorNadj:


packages/database/src/migrations/20230314042110-MigrateInsertNumberOfFacilitiesColumn-modifies-data.js
```diff
@@ -144,7 +144,7 @@ const updateStandardReports = async db => {
             transform: 'fetchData',
             dataTableCode: 'entity_relations',
             parameters: {
-              entityCodes: '=@all.organisationUnit',
+              entityCodes: '=@all.organisationUnit or []',
```
IgorNadj, line 147: The other option would have been to change the behaviour of the @all selector to return [] instead of undefined when there are no matching rows. I went with this option for quickness.

biaoli0 (approved):

PR #4399, opened 2023-03-24
RN-826 Add exitOnNoData flag to fetchData

IgorNadj:
RN-826

### DIFF
--- a/packages/report-server/src/reportBuilder/transform/functions/fetchData/fetchData.ts
+++ b/packages/report-server/src/reportBuilder/transform/functions/fetchData/fetchData.ts
@@ -9,11 +9,13 @@ import { Row } from '../../../types';
 import { createJoin } from './createJoin';
 import { TransformTable } from '../../table';
 import { TransformParser } from '../../parser';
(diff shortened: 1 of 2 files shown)


biaoli0:
@IgorNadj Nice one! This will solve the most of the cases. 
However there is one scenario that wouldn't be handled, which is `fetchData -> filter data (rows become empty []) -> fetch metadata`. Maybe we should force `fetchData` to be straight after a fetchData?

biaoli0:
> 

I am aware that we have this kind of transforms but this is a rare case. I think we can leave it as tech debt and let the fix in.

biaoli0 (approved):

IgorNadj:
> @IgorNadj Nice one! This will solve most of the cases. However there is one scenario that wouldn't be handled, which is `fetchData -> filter data (rows become empty []) -> fetch metadata`. Maybe we should force `fetchData` to be straight after a fetchData?

Yeah good point. Unfortunately I don't think we can force fetchData -> fetchData, sometimes you have to do transforms before fetching again.

PR #4400, opened 2023-03-27
Fix request schema

IgorNadj:
(no issue)

--

Fixes `Internal server error: API error 500: Internal server error: must NOT have additional properties` when fetching preview data

### DIFF
--- a/packages/types/src/types/requests/data-table-server/DataTablePreviewRequest.ts
+++ b/packages/types/src/types/requests/data-table-server/DataTablePreviewRequest.ts
@@ -10,5 +10,8 @@
 import { DataTable } from '../../models';
 
 export interface DataTablePreviewRequest extends DataTable {
-  runtimeParams: Record<string, unknown>;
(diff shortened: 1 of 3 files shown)


packages/types/src/types/requests/data-table-server/DataTablePreviewRequest.ts
```diff
@@ -10,5 +10,8 @@
 import { DataTable } from '../../models';
 
 export interface DataTablePreviewRequest extends DataTable {
-  runtimeParams: Record<string, unknown>;
+  /**
+   * @additionalProperties true
+   */
+  runtimeParams?: Record<string, unknown>;
```
IgorNadj, line 16: The ts to json schema generator puts this strictly as {} if you don't specify that it allows additional properties. 

Also, I found that it is not always present, so I set it as optional.

chris-pollard (approved):
LGTM

PR #4404, opened 2023-03-27
RN-828: Fix error in Admin Panel when viewing User Permissions

rohan-bes:
### Issue RN-828:

This bug was being caused by us trying to create a `WHERE` clause with too many parameters in it (since every entity that the user has any access to was being listed)

This change should be fine, as we currently only allow UserAccess and UserEntityPermissions at the country level.


chris-pollard (approved):
LGTM

PR #4406, opened 2023-03-28
Fixup broken date offset vizes

rohan-bes:
Now we use expressions to calculate date offsets. A bit more technical, but vastly more powerful, so worth it imo


chris-pollard (approved):
LGTM

PR #4408, opened 2023-03-28
RN-156: Added startDate and endDate selectors to the viz-builder

rohan-bes:
### Issue: RN-156

Needed to do this to test the changes in https://github.com/beyondessential/tupaia/pull/4406

Such a simple feature, and it makes the viz-building experience so much better. So I feel it's worth it

### DIFF
--- a/packages/admin-panel-server/src/routes/FetchReportPreviewDataRoute.ts
+++ b/packages/admin-panel-server/src/routes/FetchReportPreviewDataRoute.ts
@@ -29,8 +29,8 @@ export type FetchReportPreviewDataRequest = Request<
   {
     entityCode: string;
     hierarchy: string;
-    startDate: string;
-    endDate: string;
+    startDate?: string;
+    endDate?: string;
     permissionGroup?: string;
     previewMode?: PreviewMode;
     dashboardItemOrMapOverlay: DashboardItemOrMapOverlayParam;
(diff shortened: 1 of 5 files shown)


packages/admin-panel/src/VizBuilderApp/components/PreviewOptions.js
```diff
@@ -151,6 +173,24 @@ export const PreviewOptions = () => {
     setLocation(value.code);
   };
 
+  const handleChangeStartDate = date => {
+    if (!date) {
```
biaoli0, line 177: I suggest to set value when it is null. User case is user wants to see all data from all period, they will delete the previous date input.

packages/admin-panel/src/VizBuilderApp/components/PreviewOptions.js
```diff
@@ -151,6 +173,24 @@ export const PreviewOptions = () => {
...
+      return;
+    }
+
+    setSelectedStartDate(date.toISOString());
+    setStartDate(date.toISOString());
+  };
+
+  const handleChangeEndDate = date => {
+    if (!date) {
```
biaoli0, line 186: same here

biaoli0 (changes requested):
@rohan-bes Thanks for adding the date selector in viz builder! It will make everyone's life easier!

Just request a minor change to handle a special user case.

packages/admin-panel-server/src/routes/FetchReportPreviewDataRoute.ts
```diff
@@ -29,6 +29,8 @@ export type FetchReportPreviewDataRequest = Request<
   {
     entityCode: string;
     hierarchy: string;
+    startDate: string;
+    endDate: string;
```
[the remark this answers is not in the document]
rohan-bes: Ah I see, will make it optional I think 👍

biaoli0 (approved):
LGTM!

PR #4409, opened 2023-03-28
Mdev 116 add cicd check for out of date

biaoli0:
### Issue #:
- Mdev-116

### Changes:
- Use `git diff` comparing to `master` branch to validate the created date of new migrations.
- Add a new validation docker file. It looks like we don't need it at this point of view but it is better to have it.

 
- Example
https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/bea8efae-98f3-4374-9245-8f263cfe3d39?component=step_Validate_and_test_Validate_Validate_new_migrations


packages/devops/ci/validation.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+WORKDIR /tupaia
+
+# copy everything else from the repo
+COPY . ./
+
+# run yarn without building, so we can cache node_modules without code changes invalidating this layer
+RUN SKIP_BUILD_INTERNAL_DEPENDENCIES=true yarn install --frozen-lockfile
+
+#
+RUN yarn workspace @tupaia/utils build
```
rohan-bes, line 22: How come we are building tupaia and the utils package here?

packages/devops/scripts/ci/validateNewMigrations.sh
```diff
@@ -0,0 +1,77 @@
...
+    done <<< "$new_migration_names_in_string"
+
+    if [[ "$errors" != "" ]]; then
+        echo $errors;
+        exit 1;
+    fi
+}
+
+current_branch_name=$(get_branch_name)
+origin_branch_name="master"
```
rohan-bes, line 63: Great solve for comparing against master with this one! I had been struggling with how to fix this since I figured we'd need to compare against the last green build commit, but just using master should work pretty well 👍 

The only issue I see with it is that there is a chance this could fail to catch a new commit coming in to master which has migrations that expired. However, in practice that's pretty unlikely, so I think it's fine.

rohan-bes (approved):
Thanks for picking this one up @biaoli0! Great fix for a really unexpected issue.

Just a question about why we are building the code in the dockerfile, is it for the other validation tests?

packages/devops/ci/validation.Dockerfile
biaoli0: Yeah need this because `/scripts/node/validateTests.js` use `utils` package. I will add a comment.

packages/devops/scripts/ci/validateNewMigrations.sh
biaoli0: Good thinking! Yeah it is a rare case.

PR #4410, opened 2023-03-29
OSC-20 Create hierarchies endpoint

kael89:
[OSC-20](https://linear.app/bes/issue/OSC-20)

Inspiration for this work is based on this comment's [side note](https://github.com/beyondessential/tupaia/pull/4353#discussion_r1112488760).

@rohan-bes I had to do some refactoring in `entity-server` to accommodate the new endpoint... Since this package is mainly your brain child, let me know if you have any opinions on the approach 🙂

I tried to maintain feature parity with the existing `/hierarchy` endpoints for the most part. However, I didn't bother working on aspects that I thought were not worth the time at this stage. I will note any differences or inconsistencies as PR comments.


packages/entity-server/src/routes/hierarchies/HierarchiesRoute.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    const entities = await this.req.models.entity.find(
+      {
+        type: 'project',
+        [QueryConjunctions.RAW]: {
+          // Pulls permission_group/country_code pairs from the project
+          // Returns any project where we have access to at least one of those pairs
+          sql: `(
+            SELECT COUNT(*) > 0 FROM
```
kael89, line 43: Same logic as https://github.com/beyondessential/tupaia/blob/dev/packages/central-server/src/apiV2/GETProjects.js#L68

packages/entity-server/src/routes/hierarchies/types.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { Flattable, Flattened } from '../../types';
+
+export type HierarchyFields = Readonly<{
+  id: string;
```
kael89, line 9: I skipped all entity fields that seem to be empty for project entities.

Also left out:
* `parent_id`
* `type`
* `bounds` - to keep things simple. Also, not sure if we want to include geographical data in the hierarchy resource anyway

packages/entity-server/src/routes/utils/ResponseObjectBuilder.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { Writable } from '../../types';
+
+export class ResponseObjectBuilder<T> {
```
kael89, line 8: I extracted this because I thought I'd reuse it in the new endpoint.

Turns out I didn't have too, but I think it still makes sense to be a generic class.

packages/server-boilerplate/src/models/types.ts
```diff
@@ -46,7 +46,12 @@ export enum QueryConjunctions {
  * }
  */
 type ConjunctionCriteria<T> = {
-  [field in QueryConjunctions]?: DbFilterCriteria<T>;
```
kael89, line 49: This type was wrong for `_raw_` specifically, fixed

packages/entity-server/src/routes/hierarchies/HierarchiesRoute.ts
```diff
@@ -0,0 +1,68 @@
...
+            WHERE country_code IN
+            (
+              SELECT TRIM('"' FROM JSON_ARRAY_ELEMENTS(?::JSON->permission_group)::TEXT)
+            )
+          )`,
+          parameters: [JSON.stringify(countryCodesByPermissionGroup)],
+        },
+      },
+      {
+        sort: ['name'],
```
kael89, line 61: This is inconsistent with the existing endpoints where there is no sorting.

It seems a reasonable default until sorting is supported as a query param:
* If hierarchies are visualised, we usually want them sorted by name
* If they are not, we usually don't care about sorting

packages/entity-server/src/routes/hierarchies/HierarchiesRoute.ts
```diff
@@ -0,0 +1,68 @@
...
+  field?: string;
+}
+
+export interface HierarchyRequest extends Request<ReqParams, ResBody, ReqBody, ReqQuery> {
+  ctx: HierarchyContext;
+}
+
+export class HierarchyRoute extends Route<HierarchyRequest> {
+  public async buildResponse() {
+    const { field, fields } = this.req.ctx;
```
kael89, line 27: Inconsistency with existing endpoint: `filter` is not supported.

To do this the "right way", extracting and refactor existing filtering functionality to generalise it may be required. I felt that it's not worth the effort at this stage

packages/entity-server/src/app/createApp.ts
```diff
@@ -38,28 +40,31 @@ export function createApp(db = new TupaiaDatabase()) {
     new MicroServiceApiBuilder(db, 'entity')
       .useBasicBearerAuth()
 
+      // Hierarchy routes
+      .get<HierarchyRequest>('hierarchies', attachHierarchyContext, handleWith(HierarchyRoute))
```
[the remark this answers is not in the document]
rohan-bes: Ah that's a very good point @kael89 

It is indeed awkward for the Entity/MultiEntity routes. I'd be okay with changing them to what you suggested 👍 

Though I feel with the other routes it's not so bad.
`/hierarchy/explore/TO/descendants` returns the descendants of Tonga, within Explore. I feel that our current ordering makes sense, it's kind of like a nested directory?

packages/entity-server/src/routes/hierarchies/HierarchiesRoute.ts
rohan-bes: Hmmm this code is pretty complex, it'd be nice to encapsulate it in one spot.

I'd suggest the `ProjectModel` in the `@tupaia/database` package? Some kind of `getAccessibleProjects(accessPolicy)` method. Then we can re-use in central-server and entity-server 👍

Alternatively, you could actually call out to the `central-api` to get the projects here? But I think I prefer that a little less...

packages/entity-server/src/routes/hierarchies/HierarchiesRoute.ts
rohan-bes: A `sortBy` param is definitely a nice to have one day! Happy with this sort though 👍

packages/entity-server/src/routes/hierarchies/types.ts
rohan-bes: Might've been fine leaving `type`, but all good calls 👍

rohan-bes (changes requested):
Great refactors! Thanks @kael89. Things are looking tidier and this route feels like it should've always been there 😁 

Just a comment about keeping things DRY with the permissions logic around projects, but all good with all the rest

packages/entity-server/src/routes/hierarchies/HierarchiesRoute.ts
kael89: This is done now. However, I didn't refactor the mentioned code in `central-server` as it would require re-testing etc. I'd suggest for it to be done in a separate PR

kael89:
@rohan-bes I've pushed the requested changes, I think this should be ready to merge?

Although the build failed initially, then @edmofro restarted it and got green, then I merged the latest `dev` and it's red again... We should investigate if it introduces flakiness for whatever reason.

I believe the previous failure was because of db connection timeouts, is this still the case?

rohan-bes:
Sorry @kael89 for taking so long on this one. I'm totally happy with the code 👍 

Just debugging the timeouts in the tests... it seems really weird. Leave it to me tho, I think without access to codeship it's gonna be impossible for you to debug.

kael89:
> Sorry @kael89 for taking so long on this one. I'm totally happy with the code +1
> 
> Just debugging the timeouts in the tests... it seems really weird. Leave it to me tho, I think without access to codeship it's gonna be impossible for you to debug.

Thanks for looking into this @rohan-bes ! It seems like a tricky issue to investigate...

Since this is a blocker for the entity visualiser work, perhaps we could consider skipping the tests for now? Or try removing parts of them and see if the builds pass.

Let me know if there is any information I could use to debug this locally. I think I mainly followed the existing approach in other `report-server` integration tests.

Perhaps we leave some resources open? Integration tests run serially on a per-package level, but in parallel in the CI/CD level. Perhaps that may be relted?

rohan-bes:
@kael89 Thanks for changing up the test parallelisation. I'll admit I've been a little stubborn with this one. I couldn't see a clear reason why the tests started timing out just due to your changes, so I've been trying to troubleshoot exactly what's been going wrong.

Since we're on the cusp of switching to Github Actions for CI/CD (which scales resources dynamically based on the amount of parallelisation) I'm happy with your change and we can look at performance again once we've switched 👍

edmofro:
@rohan-bes they also started timing out when Waiheke split two packages out of ui-components (maps and charts), so I think we were just running right up against a resource threshold in codeship. GHA has incredible parallelisation - looking forward to that shift 🚀

rohan-bes (approved):

PR #4413, opened 2023-03-30
Fix validation too strict on fetchDataGroup result

biaoli0:
### Issue #:
When doing raw data download, it throws yup validation error with DHIS data group result because validation is too strict.

This needs to be a hot fix. Slack thread: https://beyondessential.slack.com/archives/C04QFLD87E3/p1680130358072369


packages/report-server/src/reportBuilder/output/functions/rawDataExport/rawDataExportBuilder.ts
```diff
@@ -50,7 +50,7 @@ export class RawDataExportBuilder {
           const { dataElements } = await this.aggregator.fetchDataGroup(surveyCode);
           return dataElements.map(dataElement => ({
             key: dataElement.code,
-            title: dataElement.text,
+            title: dataElement.text || dataElement.name || '',
           }));
```
biaoli0, line 54: This part of the code will be removed anyway in `data-tables-ph1`, so I think it is fine to go with the quickest way.

chris-pollard (approved):
LG!

PR #4415, opened 2023-04-02
WAITP-1120 bar chart alignment fix for time series

acdunham:
### Changes:

This bug was a result of [a known bug with Recharts](https://github.com/recharts/recharts/issues/2711) where bar graphs with timestamps as the x-axis values cause overlapping of the bars on either side.

Until a fix has been implemented by Recharts, I have added a workaround with setting the bar sizes for time scale charts (when not expanded, as this already seems to have been accounted for when expanded) and changed the values used to calculate x-axis padding on preview charts.

It should not affect bar graphs with large datasets, as this also seems to have been accounted for previously, so it's mainly on smaller datasets on a chart, e.g. [this one here](https://tupaia.org/pacmossi/SB?overlay=PacMosssi_Mosquito_Species_District)

### Screenshots:

Before:
image

After: 
image

### DIFF
--- a/packages/ui-components/src/components/Chart/BarChart.js
+++ b/packages/ui-components/src/components/Chart/BarChart.js
@@ -8,6 +8,7 @@ import PropTypes from 'prop-types';
 import { Bar, LabelList } from 'recharts';
 import { formatDataValueByType } from '@tupaia/utils';
 import { BLUE, CHART_TYPES } from './constants';
+import { getIsTimeSeries } from './utils';
 
 export const BarChart = ({
   color = BLUE,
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/components/Chart/BarChart.js
```diff
@@ -39,6 +40,13 @@ export const BarChart = ({
         return 1;
       }
     }
+    /*
+      There is a known bug with Recharts where if a bar graph is a time scale, it overflows the bars over the edge of the axes.
```
edmofro, line 44: Really clear note, thanks @acdunham! Could you please also include a link to the issue in the recharts repo?

acdunham: Added :)

tcaiger:
@acdunham Just noting that we like to add a link a couple of things in the PR description above. 
- a reference to the ticket number
- a short description of the changes

Here's an example from Rohan.
https://github.com/beyondessential/tupaia/pull/4404

Also if there are visual front end changes then we add screenshot(s) to the ticket to help the reviewer and in case there is any feedback from the product owner or someone like that

tcaiger (approved):
Looks good to me. Great work!

PR #4416, opened 2023-04-03
WAITM-1120: Bar chart alignment is not as expected

acdunham:
### Changes:

This bug was a result of [a known bug with Recharts](https://github.com/recharts/recharts/issues/2711) where bar graphs with timestamps as the x-axis values cause overlapping of the bars on either side.

Until a fix has been implemented by Recharts, I have added a workaround with setting the bar sizes for time scale charts (when not expanded, as this already seems to have been accounted for when expanded) and changed the values used to calculate x-axis padding on preview charts.

It should not affect bar graphs with large datasets, as this also seems to have been accounted for previously, so it's mainly on smaller datasets on a chart, e.g. [this one here](https://tupaia.org/pacmossi/SB?overlay=PacMosssi_Mosquito_Species_District)

### Screenshots:

Before:
image

After: 
image


tcaiger (approved):
LGTM

PR #4417, opened 2023-04-03
WAITP-1123 Fix table sorting on columns with dates

acdunham:
### Changes:

- Fixed table sorting on tables with date columns. Previously was occasionally handling some dates incorrectly if the data was not sorted by date initially. Fixed by sorting by date initially as well.

### Example

Go [to this link](https://dev.tupaia.org/psss/PW/Syndromic%20Surveillance?overlay=PSSS_PW_AFR_Syndrome_Heat_Map&overlayPeriod=DEFAULT_PERIOD&report=PSSS_Country_AFR_Case_Numbers_With_Percent_Sites_Reporting&reportPeriod=13th_Jul_2020-18th_Jul_2021). If you sort by date descending and then ascending again, 23 November 2020 ends up at the bottom, which is incorrect.

---

### Screenshots:

Before:
image

After:
image


tcaiger (approved):
Good find. It seems like it should already be sorting by date by default if the date is the first column but it is better to have this as a default for all time series data and obviously fixes the bug so I'm happy with it.

PR #4418, opened 2023-04-03
No issue: error log maintenance

biaoli0:


packages/data-api/src/validators.ts
```diff
@@ -3,23 +3,19 @@
...
+import { takesDateForm, takesIdForm, yup, yupUtils } from '@tupaia/utils';
 
 export const eventOptionsValidator = yup.object().shape({
   organisationUnitCodes: yup.array().of(yup.string().required()).strict().required(),
   startDate: yup
     .string()
-    .test(
-      yupUtils.yupTestAny([hasNoContent, takesDateForm], 'startDate should be in ISO 8601 format'),
-    ),
+    .test(yupUtils.yupTest(takesDateForm, 'startDate should be in ISO 8601 format')),
```
biaoli0, line 12: In `yupUtils.yupTestAny`, if one of functions testing passes(`hasNoContent` and `takesDateForm`), it will pass the validation. Even though the input is valid, it will still log the validation error during the validation tests run, especially for `hasNoContent`. The reason is that `hasNoContent` extends `LoggedError` and it will always write the log if it is invalid.

I think it should be fine not to use `hasNoContent` here, because `yup` will handle the same validation well if `.required()` is not attached, plus the startDate and endDate seems to always be defined (see translateOptionsForApi()).

scripts/bash/backendStartDev.sh
```diff
@@ -16,7 +16,7 @@ type_script=false
 inspect_port=${1}
 
 # Start server command for JS
-start_server="nodemon -w src --exec \"babel-node src --inspect=${inspect_port} --config-file '../../babel.config.json'\""
+start_server="nodemon -w src --exec \"babel-node src --inspect=${inspect_port} --stack-trace-limit=100 --stack-size=1024 --config-file '../../babel.config.json'\""
```
biaoli0, line 19: The default number of stack trace is not large enough to provide informative logs as it contains some unnecessary track in node_modules (path to yup modules for example). 

We don't have to worry about the long and duplicated error stack in logs, because long error stack will be trimmed and pass into other servers that are called from.

packages/data-api/src/validators.ts
biaoli0: Looks like some unit tests are failed. 😀

biaoli0: put this on hold.

biaoli0: I can't think of any good way to disable the winston log in ValidationError when using yupTestAny and our custom validation functions at the same time. Will revert the changes.

PR #4420, opened 2023-04-04
WAITP-1145 create commands for reloading on dev

acdunham:
### Changes:
- Created command for `ui-components` `start-dev` that watches changes
- Created command for `web-frontend` `start-servers` that just starts the servers, so that when running the frontend you can run the servers separately if you wish.
- Created command for `web-frontend`  `start-ui-components` that runs the `start-dev` command in the `ui-components` package
- Created command for `web-frontend` `start-frontend` that runs and watches the `web-frontend` and `ui-components` packages. This means when you update anything in `ui-components/src`, it will trigger a reload of `web-frontend` so you don't have to rebuild anything manually.


chris-pollard (approved):
Looks Good To Me (LGTM!). This is going to help a lot, thanks!

packages/web-frontend/package.json
```diff
@@ -18,13 +18,16 @@
...
     "start-central-server": "yarn workspace @tupaia/central-server start-dev",
     "start-entity-server": "yarn workspace @tupaia/entity-server start-dev",
     "start-report-server": "yarn workspace @tupaia/report-server start-dev",
     "start-web-config-server": "yarn workspace @tupaia/web-config-server start-dev",
+    "start-servers": "npm-run-all -c -l -p start-central-server start-entity-server start-report-server start-web-config-server",
     "storybook": "start-storybook -s public -p 6007",
-    "test": "node scripts/test.js --env=jsdom"
+    "test": "node scripts/test.js --env=jsdom",
+    "start-ui-components": "yarn workspace @tupaia/ui-components start-dev",
+    "start-frontend": "npm-run-all -c -l -p start-components start-dev"
```
tcaiger, line 30:
```suggestion
    "start-frontend": "npm-run-all -c -l -p start-ui-components start-dev"
```
I think this should be start-ui-components right?

tcaiger (changes requested):
Amazing work getting the watch to work in ui-components. I tried to test it locally and the ui-components dev build seems to work well but I'm getting an error when I run the yarn build in ui-components and also for some reason the my server scripts are throwing errors too when I try to start them. Do you get these errors locally too? I think that is why the codeship build is failing too.

packages/ui-components/package.json
```diff
@@ -19,7 +19,8 @@
     "storybook": "start-storybook -s public -p 6006",
     "test": "yarn package:test --env=jsdom",
     "test:coverage": "yarn test --coverage",
-    "test:watch": "yarn test --watch"
+    "test:watch": "yarn test --watch",
+    "start-dev": "babel src --out-dir dist -w"
```
IgorNadj, line 23: At the risk of having too many cooks in the kitchen (too many reviewers), can we follow the convention we have and call this `build-dev:watch`?

IgorNadj: Also, this seems like a great candidate for a reusable script. See `package:build:js`, it is written once in the root package.json and then called by many packages. This also means we don't need to have babel as a dependency, it uses the standard package build script.

IgorNadj (changes requested):

packages/ui-components/package.json
acdunham: Thanks @IgorNadj, I have pushed those changes.

IgorNadj (approved):
Thanks!

packages/ui-components/package.json
IgorNadj: Much appreciated 🙏

tcaiger (approved):
Thanks again!

PR #4422, opened 2023-04-04
Waitp 1121 loading indicator not visible

acdunham:
### Changes:
- Shows loading indicator when changing dates on a visualisation. Just uses existing `LoadingIndicator` components, which I am assuming is correct? Not sure what it used to look like (ticket implies it used to have the loader there before). There is info in the card about how to replicate the issue.


tcaiger:
@acdunham would you mind resolving these merge conflicts before I review. Thanks

packages/web-frontend/src/containers/EnlargedDialog/EnlargedDialogContent.js
```diff
@@ -128,9 +128,10 @@ export class EnlargedDialogContent extends PureComponent {
   }
 
   renderBody() {
-    const { viewContent, errorMessage } = this.props;
+    const { viewContent, errorMessage, isLoading } = this.props;
     const noData = viewContent.data && viewContent.data.length === 0;
 
+    if (isLoading) return <LoadingIndicator />;
```
tcaiger, line 134: Nice find. I'm not sure what it did before either but I think this is a good choice of component to use. I know there is no design but I think the transparent white background is probably overkill so could we have just the loading spinner with no background?

acdunham: @tcaiger Just updated this for you

tcaiger: Nice - thanks

tcaiger (approved):

PR #4423, opened 2023-04-04
Mdev-89 warn user if a question code is existed

biaoli0:
### Issue #:
mdev-89

### Change 
- Add a flag `confirmRequest` from frontend to backend to handle the warning status 
- If an existing question code is found, throw a warning to request confirm, if client wants to import it anyway, overwrite the existing question.

### DIFF
--- a/packages/utils/src/validation/ObjectValidator.js
+++ b/packages/utils/src/validation/ObjectValidator.js
@@ -68,7 +68,7 @@ async function runValidators(validators, object, fieldKey, constructError) {
       await validators[j](object[fieldKey], object, fieldKey);
     } catch (error) {
       throw constructError
-        ? constructError(error.message, fieldKey)
+        ? constructError(error.message, fieldKey, { ...error?.extraFields })
         : new ValidationError(
             `Invalid content for field "${fieldKey}" causing message "${error.message}"`,
           );
--- a/packages/admin-panel/src/importExport/ImportModal.js
+++ b/packages/admin-panel/src/importExport/ImportModal.js
@@ -27,6 +27,7 @@ const STATUS = {
   TIMEOUT: 'timeout',
   SUCCESS: 'success',
   ERROR: 'error',
+  WARNING: 'warning',
 };
 
 const defaultFinishedMessage = () => <span>Your import has been successfully processed.</span>;
@@ -50,6 +51,7 @@ export const ImportModalComponent = React.memo(
     const [status, setStatus] = useState(STATUS.IDLE);
     const [finishedMessage, setFinishedMessage] = useState(null);
     const [errorMessage, setErrorMessage] = useState(null);
+    const [warningMessage, setWarningMessage] = useState(null);
     const [isOpen, setIsOpen] = useState(false);
     const [values, setValues] = useState({});
     const [files, setFiles] = useState([]);
@@ -67,6 +69,7 @@ export const ImportModalComponent = React.memo(
     const handleDismiss = () => {
       setStatus(STATUS.IDLE);
       setErrorMessage(null);
+      setWarningMessage(null);
       setFinishedMessage(null);
       setFiles([]);
       setFileName(noFileMessage);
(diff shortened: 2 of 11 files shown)


packages/central-server/src/apiV2/import/importSurveys/importSurveys.js
```diff
@@ -242,9 +248,12 @@ export async function importSurveys(req, res) {
         let hasSeenDateOfDataQuestion = false;
         const questionCodes = []; // An array to hold all qustion codes, allowing duplicate checking
         const configImporter = new ConfigImporter(transactingModels, rows);
-        const objectValidator = new ObjectValidator(constructQuestionValidators(transactingModels));
+        const objectValidator = new ObjectValidator(
+          constructQuestionValidators(transactingModels, isNewSurvey, confirmRequest, res),
+        );
         let hasPrimaryEntityQuestion = false;
         for (let rowIndex = 0; rowIndex < rows.length; rowIndex++) {
+          const excelRowNumber = rowIndex + 2; // +2 to make up for header and 0 index
```
biaoli0, line 256: The excelRowNumber should be defined in front of where it got called.

biaoli0:
@IgorNadj @rohan-bes I am not pretty sure if this is the best approach to add a flag to handle the warning status, let me know if there is a better way 🙏

IgorNadj:
> @IgorNadj @rohan-bes I am not pretty sure if this is the best approach to add a flag to handle the warning status, let me know if there is a better way 🙏

I like this generally speaking, I would just change some of the terminology. In our case the user will see a warning and will have to confirm it, but the server doesn't know about users, it's best to think server first in this case to come up with the terminology.

Basically what is happening is there are two levels of validation, one that runs normally and one that is stricter. In the stricter mode, it will say the request is invalid if question codes exist. In the loose mode, it will allow the import to continue.

So I would change the flag to strictValidationMode: boolean, which is true by default. The request can then contain a flag strictValidationMode=false if needed.

Then the validation error that gets thrown will need to have a flag on it too. In `constructQuestionValidators` you should be able to throw a subclass of Error which is then caught in the importSurveys try/catch and responds with `{ isStrictValidationError: true, error: 'A question with the same code already exists...' }`, which will then show up to the user as "Warning: A question with the same code..." (import failed in strict mode, try again in not strict mode?).

rohan-bes:
> @IgorNadj @rohan-bes I am not pretty sure if this is the best approach to add a flag to handle the warning status, let me know if there is a better way 🙏

Yeah I'm pretty happy with this approach as well @biaoli0, though I do like @IgorNadj's suggestions to make this a bit more generic. It's possible there will be other 'strict' errors that we will want to get confirmation from the user on.

Really though, this is another case of how we'd really like to have some kind of event based architecture for processing user data submissions. Ideally an import file / survey response goes into a processing queue, and if something goes wrong during that processing we alert the user and they either tell the system to carry on, or cancel and they can rectify the error.

It's obviously out of the scope for this PR, but I do think it's worth us thinking about when and how we could switch to such an architecture.

rohan-bes (changes requested):
As discussed, might cleaner using a exception handling approach to this one. But overall design is solid 👍

biaoli0:
Changes:
1. use flag `isStrictValidationModeError` and `strictValidationMode`. If error contains `isStrictValidationModeError`, set as a warning.
2. Add new error type `StrictValidationModeError`

biaoli0:
@rohan-bes Would you like to have a second look?

packages/utils/src/validation/ObjectValidator.js
```diff
@@ -68,7 +68,7 @@ async function runValidators(validators, object, fieldKey, constructError) {
       await validators[j](object[fieldKey], object, fieldKey);
     } catch (error) {
       throw constructError
-        ? constructError(error.message, fieldKey)
+        ? constructError(error.message, fieldKey, { ...error?.extraFields })
```
biaoli0, line 71: Spend a bit more time to figure out how it works here. Using `constructError` only for `constructImportValidationError` is confusing. It could be nicer to catch the error along with where it happens...

biaoli0:
@rohan-bes The unit tests are fixed. Good to have a second look. Thanks Rohan!

IgorNadj:
Squashed commits in preparation for rebase

IgorNadj:
Rebased on dev

PR #4424, opened 2023-04-04
RN-832: Raw data exports treat numbers as text

chris-pollard:
### Changes:

- Add 'number' check in the useTableExport hook function.

### DIFF
--- a/packages/ui-components/src/components/DataTable/useDataTableExport.js
+++ b/packages/ui-components/src/components/DataTable/useDataTableExport.js
@@ -28,7 +28,9 @@ export const useDataTableExport = (columns, data, title, startDate, endDate) => 
         ? tableData.map(row =>
             tableColumns.map(col => {
               const value = row.values[col.id];
-              return isNaN(value) ? value : Number(value);
+              const num = parseFloat(value);
+              // if it's a number, return in number format
+              return isNaN(num) ? value : num;
             }),
           )
         : [['There is no available data for the selected time period.']];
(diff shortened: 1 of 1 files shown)


packages/ui-components/src/components/DataTable/useDataTableExport.js
```diff
@@ -25,7 +25,12 @@ export const useDataTableExport = (columns, data, title, startDate, endDate) =>
     );
     const body =
       tableData.length > 0
-        ? tableData.map(row => tableColumns.map(col => row.values[col.id]))
+        ? tableData.map(row =>
+            tableColumns.map(col => {
+              const value = row.values[col.id];
+              return isNaN(value) ? value : Number(value);
```
biaoli0, line 31: `isNaN()` has a surprising behaviour. The logic provided here will convert empty string '', false, null..etc to 0 because `isNaN` use `Number()` to validate the type and Number(false) = 0, so isNaN(false) = false.. 

Here I suggest we can do something like:
```
function convertToNumber(str) {
  // Try to parse the string as a floating-point number
  const num = parseFloat(str);

  // Check if the parsed value is a number (not NaN)
  if (!isNaN(num)) {
    return num;
  }

  // If the parsed value is NaN, return the original string
  return str;
}

biaoli0 (changes requested):

packages/ui-components/src/components/DataTable/useDataTableExport.js
chris-pollard: Good point, @biaoli0 , thanks for the suggestion

biaoli0 (approved):

PR #4427, opened 2023-04-05
Add options back to metadata

biaoli0:
### Issue #:
Issue found in regression, https://beyondessential.slack.com/archives/C01MAA3NKM1/p1680664696247249

Basically options are missing so legacy report can't translate value to option name.


IgorNadj (approved):
Failing test but otherwise looks good. I'm not sure how category options are part of metadata but I assume they are, rather than their own concept.

biaoli0:
> Failing test but otherwise looks good. I'm not sure how category options are part of metadata but I assume they are, rather than their own concept.

Yeah, they are slightly different, but it makes sense to combine them as they are not big.

PR #4428, opened 2023-04-05
WAITP-1144 verify email view also loads explore

acdunham:
### Changes:
- Update verify email view to load the explore view behind the modal as well as showing the landing modal. Should work for both success and error cases of verifying emails.


packages/web-frontend/src/sagas/watchUser/watchSetVerifyEmailToken.js
```diff
@@ -4,6 +4,7 @@ import {
   openEmailVerifiedPage,
   setOverlayComponent,
```
tcaiger, line 5:
```suggestion

```
Don't need this import anymore

PR #4430, opened 2023-04-06
Convert week to date for period in psss

biaoli0:
### Issue #:

https://beyondessential.slack.com/archives/C01MAA3NKM1/p1680744749341159
report server doesn't support `period` param anymore after data table phase 1 

link to the original function: https://github.com/beyondessential/tupaia/blob/6bf947a131c2b00d9ea0c5c0a2df51a0bb854a5b/packages/report-server/src/reportBuilder/transform/functions/fetchData/query/buildPeriodParams.ts#L30
### Changes:

- Example

---

### DIFF
--- a/packages/psss-server/src/connections/ReportConnection.ts
+++ b/packages/psss-server/src/connections/ReportConnection.ts
@@ -38,13 +38,13 @@ export class ReportConnection extends ApiConnection {
       throw new Error('No organisationUnitCodes provided');
     }
 
-    const [startWeek, endWeek] = periods;
-    const [startDate] = convertPeriodStringToDateRange(startWeek);
-    const [, endDate] = convertPeriodStringToDateRange(endWeek);
-
     if (orgUnitCodes.length === 0) {
       return buildEmptyReport(periods);
     }
+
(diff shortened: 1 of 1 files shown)


packages/psss-server/src/connections/ReportConnection.ts
```diff
@@ -37,13 +37,19 @@ export class ReportConnection extends ApiConnection {
...
       throw new Error('No organisationUnitCodes provided');
     }
+
+    const [startWeek, endWeek] = periods;
+    const [startDate] = convertPeriodStringToDateRange(startWeek);
+    const [, endDate] = convertPeriodStringToDateRange(endWeek);
+
     if (orgUnitCodes.length === 0) {
       return buildEmptyReport(periods);
     }
```
chris-pollard, line 47: A very minor edit, but I think the conversion of periods variable to startDate and endDate should come after checking that there are no orgUnitCodes, as there isn't any point running that code if orgUnitCodes length is 0.

chris-pollard (changes requested):

chris-pollard (approved):
LG!

PR #4431, opened 2023-04-06
Run Postgres using docker compose in development

amenasse:
Avoid having to install specific versions of Postgres locally. Following starts postgres 13 with the postgis extension installed:

    $ docker-compose up 

Port can be overriden:

     $ PG_PORT=8765 docker-compose up

The database is stored as a docker volume.

```
$ docker volume ls

DRIVER    VOLUME NAME
local          tupaia_db
 ```
 To clean it up

    $ docker volume rm tuapai_db

Main motivation for using a volume is not to pollute the working dir with the db files. I  guess .gitignore + .dockerignore  works too, so happy to change this to a bind mount inside the working dir if preferred  :shrug:  

The `refreshDatabase.js` script has been updated to respect `DB_URL` and `DB_PORT` variables so it can connect to the compose db. There is a --no-gis (or --gis=false) option so that the postgis extension install isn't attempted. This is need as extension installation will fail on the postgis image, and isn't possible using the postgres image (at least without some hacks)

   $ yarn refresh-database --no-gis  dump.sql

I guess if approved,  I should document this somewhere.. is [tupaia-mono-repo-setup](https://beyond-essential.slab.com/posts/tupaia-monorepo-setup-v5egpdpq) the right place ??

### DIFF
--- a/packages/database/scripts/refreshDatabase.js
+++ b/packages/database/scripts/refreshDatabase.js
@@ -62,8 +62,9 @@ class RefreshDatabaseScript extends Script {
     const dbArgs = { user: dbPgUser, port: dbPort, host: dbHost };
     this.execDbCommand(`DROP DATABASE IF EXISTS ${dbName}`, dbArgs);
     this.execDbCommand(`CREATE DATABASE ${dbName} WITH OWNER ${dbUser}`, dbArgs);
-    // Loading GIS extension is optional - it's already loaded when using
-    // postgis/postgis Docker container
+    // Loading GIS extension is optional. Its already loaded in the
+    // postgis/postgis Docker image. Loading an extension thats already loaded
+    // fails.
     if (this.args.gis) {
       this.execDbCommand('CREATE EXTENSION postgis', dbArgs);
     }
--- a/packages/database/scripts/refreshDatabase.js
+++ b/packages/database/scripts/refreshDatabase.js
@@ -10,7 +10,7 @@ require('dotenv/config');
 const fs = require('fs');
 const path = require('path');
 
-const { requireEnv, Script } = require('@tupaia/utils');
+const { requireEnv, Script, getEnvVarOrDefault } = require('@tupaia/utils');
 
 class RefreshDatabaseScript extends Script {
   config = {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/database/scripts/refreshDatabase.js [hunk 1]
```diff
@@ -50,18 +59,23 @@ class RefreshDatabaseScript extends Script {
...
-    this.execDbCommand(`ALTER USER ${dbUser} WITH SUPERUSER`, { user: dbPgUser });
-    this.exec(`psql -U ${dbUser} -d ${dbName} -f "${dumpPath}"`);
-    this.execDbCommand(`ALTER USER ${dbUser} WITH NOSUPERUSER`, { user: dbPgUser });
+    const dbArgs = { user: dbPgUser, port: dbPort, host: dbHost };
+    this.execDbCommand(`DROP DATABASE IF EXISTS ${dbName}`, dbArgs);
+    this.execDbCommand(`CREATE DATABASE ${dbName} WITH OWNER ${dbUser}`, dbArgs);
+    // Loading GIS extension is optional - it's already loaded when using
+    // postgis/postgis Docker container
+    if (this.args.gis) {
+      this.execDbCommand('CREATE EXTENSION postgis', dbArgs);
```
IgorNadj, line 68: IIRC this command does nothing if the extension already exists, so it should be ok to run always.

IgorNadj:
Happy to approve, however, I'm not sure anyone would want to do this? Most of our devs use Mac/Windows and containerised Postgres has a big performance hit.

amenasse:
> Happy to approve, however, I'm not sure anyone would want to do this? Most of our devs use Mac/Windows and containerised Postgres has a big performance hit.

Ah right, I  didn't realise performance would be an issue on those platforms. The motivation was primarily so I could avoid   to polluting my system with old Postgres versions and I figured it was worth sharing to make developer setup a bit easier

 For Windows,  the [instructions](https://beyond-essential.slab.com/posts/tupaia-monorepo-setup-v5egpdpq#h4qjy-%F0%9F%93%A4-windows) recommend running Postgres under WSL. I don't imagine running it under Docker with the WSL backend  would be much worse.. but i've never tried.  

The focus at the moment is getting the ECS stuff going, so happy to leave this for now.

IgorNadj (approved):

IgorNadj:
The failing build is because the branch name ends with "db". If you recreate this PR with branch name ant/docker-compose-db2 it should work.

PR #4432, opened 2023-04-06
Allow to set startDate and endDate to be undefined in viz builder

biaoli0:
### Changes:

- Example

---


rohan-bes (approved):
Woops, nice fix!

PR #4435, opened 2023-04-11
WAITM-336: Handle 404 urls

tcaiger:
### Changes:

- Add new 404  modal page template. The design has the 404 page as a full page but we don't have full pages anywhere so I thought it was more consistent to make it a modal
- Add handling for when no project can be found (eg. http://localhost:8088/asdf)
-  Add handling for when no org unit can be found to show the 404 page (eg. http://localhost:8088/fanafana/asdf)
- Add handling for when the dashboard in the url can not be found in the list of requested dashboards to show the 404 page in the case that the user already has access to all the entities in the project (eg. http://localhost:8088/fanafana/TO/asdf).

I experimented with handling the dashboard not found flow in a more direct way by trying to determine if a user has permission to a given dashboard but it quickly got quite difficult so I went for the more practical approach described above.

---

### Screenshots:
Screen Shot 2023-04-11 at 1 20 15 PM


packages/web-frontend/src/containers/OverlayDiv/components/NotFoundPage.js
```diff
@@ -0,0 +1,57 @@
...
+`;
+
+const StyledButton = styled(PrimaryButton)`
+  width: 300px;
+`;
+
+const NotFoundPageComponent = ({ goHome }) => {
+  return (
+    <Container>
+      <Image src="/images/404.svg" />
```
acdunham, line 42: Strictly speaking, having text in images isn't recommended, mainly because any user using assistive technologies won't be able to interpret it. There are a couple of options here:

- Instead of using an image, make this into a heading (probably h2) and style to look like the image
- If you need it to be an image, adding alt text to the image to describe what it is will help the screenreader read out what it says. 

My inclination would be to style a heading to look like the image, purely because the image itself is just text, so there is nothing else to display except '404'.

packages/web-frontend/src/containers/OverlayDiv/components/NotFoundPage.js
```diff
@@ -0,0 +1,57 @@
...
+
+const Container = styled.div`
+  height: 500px;
+  width: 800px;
+  max-width: 100%;
+  max-height: 100%;
+  text-align: center;
+  padding-top: 20%;
+  padding-bottom: 30px;
+`;
```
acdunham, line 20: Potentially using `rems`/`ems`/percentages here instead of `px` would be better for future-proofing things

packages/web-frontend/src/containers/OverlayDiv/components/NotFoundPage.js
```diff
@@ -0,0 +1,57 @@
...
+  max-width: 100%;
+  margin-bottom: 30px;
+`;
+
+const Text = styled(Typography)`
+  font-size: 18px;
+  line-height: 21px;
+  color: white;
+  margin-bottom: 30px;
+`;
```
acdunham, line 33: Same as above, if possible, using `rem` or `em` would be ideal

packages/web-frontend/src/containers/OverlayDiv/components/NotFoundPage.js
```diff
@@ -0,0 +1,57 @@
...
+const Image = styled.img`
+  width: 300px;
+  max-width: 100%;
+  margin-bottom: 30px;
+`;
+
+const Text = styled(Typography)`
+  font-size: 18px;
+  line-height: 21px;
+  color: white;
```
acdunham, line 31: What are your thoughts on using a theme color for text color, on the off chance that the theme color changes (e.g. `palette.text.secondary` etc etc). It's not a random hex code so should be pretty straightforward.

acdunham (changes requested):
Just added a few comments/changes for you to implement. Nice way of doing this though!

packages/web-frontend/src/containers/OverlayDiv/components/NotFoundPage.js
tcaiger: 👍

packages/web-frontend/src/containers/OverlayDiv/components/NotFoundPage.js
tcaiger: Yes I agree - good suggesion

packages/web-frontend/src/containers/OverlayDiv/components/NotFoundPage.js
tcaiger: Yes good point. I think that alt text should achieve the same goal for a screen reader right (and I should have added that in the first place obviously) without having to change the design. I don't think that I can make it look exactly like the image with just css as I think it's a different font. Plus using alt text I can put a more human centric text such as page not found.

acdunham:
@tcaiger just one final question - with the modal, what should happen if the user clicks outside the modal/clicks the 'x' close button? Should probably redirect back to home do you think?

tcaiger:
> @tcaiger just a couple of final questions
> 
> 1. with the modal, what should happen if the user clicks outside the modal/clicks the 'x' close button? Should probably redirect back to home do you think?
> 2. In the case where you do something like http://localhost:8088/fanafana/TO/asdf - since that's not a valid project to request access to, is there some way to know that and handle that as a 404? Not totally sure what should happen here, might be worth asking Felicity?

1. I thought about that but I think that it would be a bit of an unexpected behaviour as everywhere else where we close a modal with the x button it simply closes the modal.
2. I am handling that case (as far as I know) in `packages/web-frontend/src/sagas/handlers/fetchCountryAccessDataIfRequired.js`. Did it not work when you tested it? I had to handle it in a round about way which is what I was referring to in my comment in the description but it should be handled.

acdunham:
> > @tcaiger just a couple of final questions
> > 
> > 1. with the modal, what should happen if the user clicks outside the modal/clicks the 'x' close button? Should probably redirect back to home do you think?
> > 2. In the case where you do something like http://localhost:8088/fanafana/TO/asdf - since that's not a valid project to request access to, is there some way to know that and handle that as a 404? Not totally sure what should happen here, might be worth asking Felicity?
> 
> 1. I thought about that but I think that it would be a bit of an unexpected behaviour as everywhere else where we close a modal with the x button it simply closes the modal.
> 2. I am handling that case (as far as I know) in `packages/web-frontend/src/sagas/handlers/fetchCountryAccessDataIfRequired.js`. Did it not work when you tested it? I had to handle it in a round about way which is what I was referring to in my comment in the description but it should be handled.

Ah yes I see - it didn't do it for me the first time but it does not - not sure if it cached for me the first time or something!

acdunham (approved):

PR #4436, opened 2023-04-11
RN-851 Knex to handle numeric and bigInt type

biaoli0:
### Issue #:
RN-851 

Checked that type `timestamp without timezone`, `Integer` work fine.

### DIFF
--- a/packages/database/src/modelClasses/ExternalDatabaseConnection.js
+++ b/packages/database/src/modelClasses/ExternalDatabaseConnection.js
@@ -5,12 +5,17 @@
 
 import { getEnvVarOrDefault, requireEnv } from '@tupaia/utils';
 import knex from 'knex';
+import { types as pgTypes } from 'pg';
 
 import { DatabaseModel } from '../DatabaseModel';
 import { DatabaseType } from '../DatabaseType';
 import { TYPES } from '../types';
 
 const EXT_DB_CONNECTION_ENV_VAR_PREFIX = 'EXT_DB';
+
+pgTypes.setTypeParser(pgTypes.builtins.NUMERIC, value => {
+  return parseFloat(value);
+});
 
 export class ExternalDatabaseConnectionType extends DatabaseType {
   static databaseType = TYPES.EXTERNAL_DATABASE_CONNECTION;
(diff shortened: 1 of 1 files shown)


packages/database/src/modelClasses/ExternalDatabaseConnection.js
```diff
@@ -5,13 +5,17 @@
...
+import { types as pgTypes } from 'pg';
 
 import { DatabaseModel } from '../DatabaseModel';
 import { DatabaseType } from '../DatabaseType';
 import { TYPES } from '../types';
 
 const EXT_DB_CONNECTION_ENV_VAR_PREFIX = 'EXT_DB';
 
+pgTypes.setTypeParser(pgTypes.builtins.NUMERIC, parseFloat);
+pgTypes.setTypeParser(20, parseInt); // bigInt type to Integer
```
rohan-bes, line 17: Nice one @biaoli0! Any idea why this issue is occurring here, but not for `TupaiaDatabase`? Pretty sure we pass numeric data fine there...

Also, I saw this code in `TupaiaDatabase` that seems to be used to get around timezone issues. Might be of interest:
```
// turn off parsing of timestamp (not timestamptz), so that it stays as a sort of "universal time"
// string, independent of timezones, rather than being converted to local time
pgTypes.setTypeParser(pgTypes.builtins.TIMESTAMP, val => val);
```

rohan-bes (commented):
Does Boolean work with this btw?

packages/database/src/modelClasses/ExternalDatabaseConnection.js
biaoli0: Yeah I can't tell but it is weird. Can we take it as a tech debt? I will create a ticket and can take a deep look during the upcoming sprint.

biaoli0:
> Does Boolean work with this btw?

Yes Boolean works.

packages/database/src/modelClasses/ExternalDatabaseConnection.js
rohan-bes: Agreed, can u add a ticket into data-tables phase 3 about it 👍

rohan-bes (approved):

packages/database/src/modelClasses/ExternalDatabaseConnection.js
biaoli0: As discussed with Edwin, there is no type issue before because Tupaia considers all these types as a string. Since `pgTypes.setTypeParser` will setup globally, it currently introduces a bug for many places. I will work on a hotfix about it.

PR #4441, opened 2023-04-11
RN-855: Allow using non-explore enitites for organisationUnit param default

rohan-bes:
### Issue RN-855:
Allow using any entities when using the 'organisationUnit' param type, regardless of hierarchy. Arguably a bit of a loss of functionality since filtering by the hierarchy was kind of cool, but I think this is more easy to work with.

Also, for a data set like LESMIS where all the data is in the same project, needing to specify a hierarchy is kind of pointless. So it's nice not to have the organisationUnits parameter dependant on a hierarchy parameter


packages/admin-panel/src/VizBuilderApp/api/queries/useEntities.js
```diff
@@ -0,0 +1,25 @@
...
+      const endpoint = stringifyQuery(undefined, `entities`, {
+        columns: JSON.stringify(['name', 'code']),
+        filter: JSON.stringify({
+          name: { comparator: 'ilike', comparisonValue: `%${search}%`, castAs: 'text' },
+        }),
+      });
+      return get(endpoint);
+    },
+    {
+      ...DEFAULT_REACT_QUERY_OPTIONS,
```
biaoli0, line 23: Can you add a debounce here? I haven't tried but I guess it would be a bit lagging when adding words.

packages/admin-panel/src/dataTables/components/PreviewFilters/filters/OrganisationUnitCodesField.js
```diff
@@ -3,25 +3,19 @@
...
+import React, { useState } from 'react';
 
 import PropTypes from 'prop-types';
 
 import { ParameterType } from '../../editing';
-import { useLocations } from '../../../../VizBuilderApp/api';
+import { useEntities } from '../../../../VizBuilderApp/api';
 import { Autocomplete } from '../../../../autocomplete';
 
-export const OrganisationUnitCodesField = ({ name, onChange, runtimeParams }) => {
```
biaoli0, line 14: can you also remove `runtimeParams` on where it is called? It is at `PreviewFilters.js`.

biaoli0 (changes requested):
Nice one, small changes requested, overall looks good!

biaoli0:
Do we also want to do it for the hierarchy and entity selection on viz builder?

rohan-bes:
> Do we also want to do it for the hierarchy and entity selection on viz builder?

Good thinking @biaoli0. I think it should be okay though. Those params are meant to replicate Tupaia front-end, and the locations availabole are determined by the selected project there. So having it filter works fine and is quite convenient

packages/admin-panel/src/VizBuilderApp/api/queries/useEntities.js
rohan-bes: Ah great catch @biaoli0! I was caching the query and not refetching based on the search term. I think it should be fixed now 👍

biaoli0 (approved):
LGTM! A reminder to change the base back to dev.

PR #4445, opened 2023-04-13
WAITP-1139: Fix password reset form

tcaiger:
### Changes:

- Don't let the user close the password reset form without submitting it

---

### Screenshots:
No visible ui changes


packages/web-frontend/src/containers/UserBar/index.js
```diff
@@ -94,13 +94,13 @@ export class UserBar extends Component {
...
+    // Don't let the user close a dialog if it is a password reset form as the one time login token
+    // is used to load the form and so they need to do the password reset now */
     return (
       <Dialog
         open={isDialogVisible}
         style={{ zIndex: 1301 }}
         PaperProps={{ style: { ...styles.dialogBody, width, maxWidth: width } }}
-        {/* Don't let the user close a dialog if it is a password reset form as the one time login
-        token is used to load the form and so they need to do the password reset now*/}
         onClose={dialogPage === DIALOG_PAGE_RESET_PASSWORD ? null : () => onCloseUserDialog()}
```
acdunham, line 104: Just one small thing - generally having inline functions (like `() => onCloseUserDialog()` isn't advised, unless you have to pass something into the function being called. This is because whenever the view is rerendered (e.g. when state or props change) the function is recreated, which has a (minor) performance impact. I'd probably either have a method called `handleCloseUserDialog` within the class component, and if `dialogPage === DIALOG_PAGE_RESET_PASSWORD` then return early, else call `onCloseUserDialog` or just remove the arrow function part, to make it `dialogPage === DIALOG_PAGE_RESET_PASSWORD ? null : onCloseUserDialog`. Either is technically fine, I guess you could argue that that ternary might be doing too much, but you could also argue that it isn't!

packages/web-frontend/src/containers/UserBar/index.js
```diff
@@ -94,13 +94,13 @@ export class UserBar extends Component {
...
     const width = this.getWidth();
+    // Don't let the user close a dialog if it is a password reset form as the one time login token
+    // is used to load the form and so they need to do the password reset now */
     return (
       <Dialog
         open={isDialogVisible}
         style={{ zIndex: 1301 }}
         PaperProps={{ style: { ...styles.dialogBody, width, maxWidth: width } }}
-        {/* Don't let the user close a dialog if it is a password reset form as the one time login
-        token is used to load the form and so they need to do the password reset now*/}
```
acdunham: Nice commenting

acdunham (changes requested):
Looks good, just one minor change.

packages/web-frontend/src/containers/UserBar/index.js
tcaiger: woops, yes that doesn't need to be an inline function. I think your first recommendation is best so I'll go with that.

acdunham (approved):

PR #4449, opened 2023-04-13
WAITP-92: Fetch projects after one time login 

tcaiger:
### Changes:

- We weren't fetching a fresh list of projects after the user was logged in with a one time login token and so they only had access to the public projects

---


rohan-bes (approved):
Nice! Thanks @tcaiger

PR #4455, opened 2023-04-14
Waitp 1142 upload image file input

acdunham:
### Changes:
- Change `ProfileImageField` to be more generic. Now named `ImageUploadField`, and anywhere that imports this has been updated accordingly.
- Make `ImageUploadField` allow max width and height of an image to be set for validation purposes. If the image is outside of these bounds, then it won't upload, and will show an error message.
- Update project editor and creator fields to be of type `image` to make it use the `ImageUploadField`
- Update input field factory to allow `image` field types
- Update `CreateProject` API logic to upload `image_url` and `logo_url` base64 encoded images to S3 whenever these are set, and use the returned URL as the value of these fields when creating the project. Creates a standardised image name from the project code and the image type.
- Update `EditProject` API logic to upload `image_url` and `logo_url` base64 encoded images to S3 whenever these are set, and use the returned URL as the value of these fields when creating the project. Creates a standardised image name from the project code and the image type.
- Update `S3Client` to allow overriding of images by passing in a param `allowOverrides` that defaults to `false`. If it is `true`, overriding of the image by the same name is allowed.

### Screenshots:
Old:
image

New:
image


packages/admin-panel/src/editor/Editor.js
```diff
@@ -7,18 +7,32 @@ import React from 'react';
...
+    font-size: 0.9375rem;
+    text-transform: initial;
+    color: ${props => props.theme.palette.text.secondary};
+  }
+`;
 
 export const Editor = ({ fields, recordData, onEditField }) => {
-  const onInputChange = (inputKey, inputValue, editConfig = {}) => {
-    const { setFieldsOnChange } = editConfig;
+  const onInputChange = async (inputKey, inputValue, editConfig = {}) => {
```
rohan-bes, line 22: I don't love the switching to `async` here. It probably would be fine, but any idea if we could re-work the `createBase64Image` function to be synchronous using `FileReaderSync`? It seems broadly supported across browers, though it does require being in worker threads... No biggie but just wanted to suggest it

rohan-bes (approved):
Awesome stuff @acdunham, thanks for this 🙏 

A handful of general comments but nothing too major, so pre-approving.

rohan-bes (approved):
Great stuff! Thanks 👍 

If you were so inclined, I noticed that `GETProjects.js` isn't inside the `project/` directory. Totally optional, but if you wanted to move it in there as part of this PR that would be very cool 😎

packages/central-server/src/tests/apiV2/projects/GETProjects.test.js
```diff
@@ -0,0 +1,171 @@
+/**
```
rohan-bes, line 1: Thanks for adding tests! 🙏

rohan-bes (approved):
Thanks for that cleanup!

PR #4456, opened 2023-04-14
RN-864: Add 'Delete' button to Data-Tables page in the Admin Panel

rohan-bes:
### Issue RN-864:

Added missing 'Delete' button. Also chose to tighten up permissions on creating data-tables, more explanation in PR comments.


packages/central-server/src/apiV2/dataTables/CreateDataTables.js
rohan-bes, line 1: Chose to delete this, and instead use BES Admin permissions requirements for creating data-tables.

Did this because:
- Currently we don't have a way of ensuring that a user has access to the External Database Connection they're creating/editing their data-table with. 
- We could add this permissions check, but I figured that since it's really only going to be internal users creating data-tables for the forseable future, using BES Admin for everything data-tables is sufficient. 
- If reviewer feels differently I'm happy to introduce appropriate logic for ensuring that a user doesn't create/edit data-tables which use external database connections that they don't have access to.

chris-pollard (approved):
Nice one @rohan-bes !

PR #4458, opened 2023-04-16
WAITP-1141 Update autocomplete fields to run more efficiently

EMcQ-BES:
### Changes:

- Remove unnecessary `distinct` from Autocomplete api calls
  - See comment on linear ticket for explanation of why that distinct is unnecessary
- Run initial search api call on opening autocomplete field rather than on mount
- Debounce text entry in autocomplete field

---


tcaiger (commented):
Good find on the distinct select and the lazy load changes look good. I think that we should also add debouncing though because if the user can crash the database by opening 10 modals then it's even more likely that the user could crash the database by hitting 10 key strokes in an autocomplete field, and that would stack up the requests much quicker.

tcaiger (approved):
Nice

PR #4459, opened 2023-04-16
WAITP-1122: Update users import

tcaiger:
### Changes:

- Include the ability to import with no gender OR have an 'unknown' option
- Include the ability to import as 'verified' 

---


packages/central-server/src/apiV2/import/importUsers.js
```diff
@@ -90,10 +91,11 @@ const FIELD_VALIDATORS = {
   first_name: [hasContent],
   last_name: [],
   email: [hasContent],
-  gender: [hasContent, constructIsOneOf(['f', 'm'])],
+  gender: [constructIsEmptyOr(constructIsOneOf(['f', 'm', 'unknown']))],
```
rohan-bes, line 94: Never really knew why we track gender honestly... But good change! 👍

packages/central-server/src/apiV2/import/importUsers.js
```diff
@@ -90,10 +91,11 @@ const FIELD_VALIDATORS = {
...
   employer: [hasContent],
   position: [hasContent],
   mobile_number: [],
   password: [hasContent],
   permission_group: [hasContent],
+  verified_email: [constructIsEmptyOr(constructIsOneOf(['unverified', 'verified', 'new_user']))],
```
rohan-bes, line 100:
```suggestion
  import { UserAccountSchema } from '@tupaia/types';
  
  ...
  
  verified_email: [constructIsEmptyOr(constructIsOneOf(UserAccountSchema.verified_email.enum))],
```

This is fairly new functionality, but now we have a package called `@tupaia/types` which we use to maintain a source of truth about our models. It's possible we could replace all these validators in central-server with ones that use `ajv` and `@tupaia/types` but for now I think we can at least rely on them for enums and the sort.

@IgorNadj I feel you've got a stronger grasp on using `@tupaia/types` that I do, does what I'm suggesting here seem appropriate?

rohan-bes (changes requested):
Just a suggestion to use `@tupaia/types` over re-defining the enum options for verified_email

packages/central-server/src/apiV2/import/importUsers.js
IgorNadj: As much as I'd love to use @tupaia/types here, I would't want to burden Tom with being the first to use it in central-server :) 

We still have a little work to do in finding a nice clean and consistent way to introduce type validation into central-server. Until then I think sticking with the old school custom validators is ok. Thanks for checking though!

IgorNadj: Ah yep, here's an example of using an enum:
https://github.com/beyondessential/tupaia/blob/dev/packages/central-server/src/apiV2/utilities/constructNewRecordValidationRules.js#L292

tcaiger: Nice, this is great. I have added the type for verified email but the type for gender is just a string so I've left that one as is.

rohan-bes (approved):
Awesome! Thanks @tcaiger

PR #4462, opened 2023-04-17
RN-848: Improve Option Set Import Error Message

chris-pollard:
### Changes:

- Small lint fix
- Add custom error message for when option set cannot be found in database, but throw database message if not specific case.

---

### Screenshots:
Screen Shot 2023-04-17 at 12 42 55 pm


packages/central-server/src/apiV2/import/importSurveys/importSurveys.js
```diff
@@ -417,11 +422,23 @@ export async function importSurveys(req, res) {
...
+      new ImportValidationError(message, excelRowNumber, field, tabName);
+    try {
+      const optionSet = await models.optionSet.findOne({ name });
+      return optionSet.id;
+    } catch (error) {
+      console.log(error.message);
+      const message =
+        error.message === "Cannot read property 'id' of null"
+          ? 'The option set listed cannot be found in the database'
+          : error.message;
```
biaoli0, line 438: @chris-pollard Do you want to locate this error where trying to read 'id' from a null record? Maybe we could throw a better generic error message from there?

biaoli0 (changes requested):
Thanks @chris-pollard for some linting fix along with the PR. I raise a comment about adding a generic error message, let me know what you think.

packages/central-server/src/apiV2/import/importSurveys/importSurveys.js
biaoli0: My bad. The `Cannot read property 'id' of null` error is not from database model, instead it is from reading id from null on line 432.
We could throw an error by checking if return result is null.

chris-pollard: Hey @biaoli0 fixed it up, thanks

packages/central-server/src/apiV2/import/importSurveys/importSurveys.js
```diff
@@ -417,10 +422,14 @@ export async function importSurveys(req, res) {
   respond(res, { message: 'Imported surveys' });
 }
 
-async function processOptionSetName(models, name) {
+async function processOptionSetName(models, name, excelRowNumber, tabName) {
   // TODO: Figure out why undefined value is returning all results.
   if (name) {
     const optionSet = await models.optionSet.findOne({ name });
+    if (optionSet !== null) {
```
biaoli0, line 429: Should it be:
```suggestion
    if (!optionSet) {
```

chris-pollard: Thanks @biaoli0 I pushed a commit.

biaoli0 (approved):
LGTM

PR #4465, opened 2023-04-17
RN-875: Fix booleans not showing up in data-table preview

rohan-bes:
### Issue RN-875:
Moved cell content formatting to DataTableCellComponent
- Added 'data-type-x' class name so that we can used css to colour different data types differently
- Added colour coding for number, boolean, and null values

### DIFF
--- a/packages/ui-components/src/components/DataTable/StyledTable.js
+++ b/packages/ui-components/src/components/DataTable/StyledTable.js
@@ -58,4 +58,17 @@ export const StyledTable = styled(MuiTable)`
       border-right: none;
     }
   }
+
+  td.data-type-number {
+    color: #0f0f87;
+  }
+
+  td.data-type-boolean {
+    color: #630000;
+  }
+
+  td.data-type-null {
+    color: #8f8e8e;
+    font-style: italic;
+  }
 `;
--- a/packages/admin-panel/src/utilities/getColumns.js
+++ b/packages/admin-panel/src/utilities/getColumns.js
@@ -3,18 +3,6 @@
  * Copyright (c) 2023 Beyond Essential Systems Pty Ltd
  */
 
-const convertValueToPrimitive = val => {
-  if (val === null) return val;
-  switch (typeof val) {
-    case 'object':
-      return JSON.stringify(val);
-    case 'function':
-      return '[Function]';
-    default:
-      return val;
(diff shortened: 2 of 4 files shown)


packages/ui-components/src/components/DataTable/StyledTable.js
```diff
@@ -58,4 +58,17 @@ export const StyledTable = styled(MuiTable)`
...
+  }
+
+  td.data-type-boolean {
+    color: #630000;
+  }
+
+  td.data-type-null {
+    color: #8f8e8e;
+    font-style: italic;
+  }
```
acdunham, line 73: This is cool. One thing you might want to consider as an alternative is to use styled-components to apply the colors. For example:

- `DataTableCell` would make a styled component that extends the `TableCell` from MUI
- The new styled component would be used like so:
``` <StyledTableCell {...props}  dataType={getDataType(value)}>```
- the styled component would apply the styles according to `props.dataType`, for example:
````
   color: ${({dataType})=> {
        if(dataType === "number") return "#0f0f87";
        else if (dataType === "boolean") return "#630000";
        else if (dataType === "null") return "#8f8e8e";
   }};
   font-style: ${({dataType}) => dataType === "null" ? "italic" : "initial"};
````
Either one is fine and works, of course. I might argue that having a styled component contained within `DataTableCell` fits with the pattern that the `ui-components` package uses, but that might also be an argument against it.

acdunham (approved):

packages/ui-components/src/components/DataTable/StyledTable.js
rohan-bes: Oh thanks for suggesting this, I prefer it!

Out of curiosity, what's the argument against using `styled-components` within `DataTableCell.js`? I didn't create the pattern, but it always seemed solid enough to me.

rohan-bes: Okay, I tried this out with the `styled-components` method. Unfortunately due to how the css is defined the `td` definition from the parent `StyledTable` component overrides the child's color.

I could move all the css around a bit, but I don't really wanna do that. I'm pretty happy with either option, so I might just stick with the original. Thanks all the same!

acdunham (approved):

PR #4476, opened 2023-04-18
WAITP-1134 project modal covers reset password screen

acdunham:
### Changes:

- Hid the project modal when is mobile and is password reset screen
- Updated the yarn command for starting in mobile so that it works correctly


packages/web-frontend/src/sagas/handlers/handleUserPage.js
```diff
@@ -12,9 +12,12 @@ import {
   URL_COMPONENTS,
   VERIFY_EMAIL_PREFIX,
 } from '../../historyNavigation/constants';
+import { isMobile } from '../../utils';
 
 export function* handleUserPage(userPage, initialComponents) {
-  yield put(setOverlayComponent(LANDING));
+  // Don't show the overlay on mobile if the user is on the password reset page, because the overlay component appears over the top of the password reset page
+  const shouldDisplayOverlay = !isMobile() || userPage !== PASSWORD_RESET_PREFIX;
+  if (shouldDisplayOverlay) yield put(setOverlayComponent(LANDING));
```
tcaiger, line 20: This is a good solution. I think the ideal flow would be that the user is taken to the Landing screen after the password reset form is done. Is that easily possible to add in?

acdunham: I have made changes to the flow so that:
- The project modal doesn't appear when reset password and/or change password modals are open
- On closing the reset password/change password modals, if a project is not currently selected, 'go home'
- Does this for both mobile and web, because on web you also would have a situation of potentially not having a project selected if you dismiss the reset password modal AND the project modal

packages/web-frontend/package.json
```diff
@@ -17,7 +17,7 @@
     "lint:fix": "yarn lint --fix",
     "start": "PORT=8088 yarn package:start:react-scripts",
     "start-dev": "yarn start",
-    "start-mobile": "REACT_APP_APP_TYPE=mobile PORT=8089 PORT=8088 package:start:react-scripts",
+    "start-mobile": "REACT_APP_APP_TYPE=mobile PORT=8089 PORT=8088 yarn package:start:react-scripts",
```
tcaiger, line 20: thanks

tcaiger (approved):
Great - thanks for making the extra changes.

PR #4478, opened 2023-04-19
RN-870 Migrate validation service to GHA

biaoli0:
### Issue #:
RN-870

### Changes:
1. Upgrade yarn from 3.2.1 to 3.2.4, as GHA is not compatible with v.3.2.1 for module caching 
2. Migrate validation service to GHA 

CI: 
https://github.com/beyondessential/tupaia/actions/runs/4738148278


.github/workflows/validation.yaml
```diff
@@ -0,0 +1,52 @@
+name: Validation
+on:
+  push:
+    branches:
+      - '*'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
```
rohan-bes, line 10: This cancels the running job if a new one comes up?

.github/workflows/validation.yaml
```diff
@@ -0,0 +1,52 @@
+name: Validation
+on:
+  push:
+    branches:
+      - '*'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
```
rohan-bes, line 9: When is the `ref` undefined?

.github/workflows/validation.yaml
```diff
@@ -0,0 +1,52 @@
...
+  validate_tests:
+    name: Validate unit tests
+    strategy:
+      matrix:
+        node:
+          - '14.19.3'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
```
rohan-bes, line 43: Oh this sets up node? Cool 😎

rohan-bes (approved):
Looks great! Thanks @biaoli0 

I just had questions really about how things work, but I've enjoyed playing around with GHA!

.github/workflows/validation.yaml
biaoli0: This is a very good question, it is copy and paste from Tamanu workflow and I didn't look into that deep. 

`github.ref` would be the branch name for Tupaia in most cases, unless someone uses `git tag` and leave `undefined` message, `github.ref` could be undefined here.

.github/workflows/validation.yaml
biaoli0: Yeah GHA use this kind of format to call actions. You can create a new action or call a third party action from marketplace too.

.github/workflows/validation.yaml
biaoli0: This will cancel the running jobs if a new commit pushed to the same branch.

PR #4479, opened 2023-04-20
Remove build non internal dependencies in ci for testing service

biaoli0:
### Issue #:
It is not necessary to run `yarn build:non-internal-dependencies` in testing service because non internal packages are not relied on others anymore.  

1. This will save us ~6 mins to build testing service
2. Easier to migrate to GHA 

### Changes:
  Remove report-server dependency in admin-panel-server (tiny)


rohan-bes (approved):
I love this change 😻 Thank you SO MUCH!

PR #4480, opened 2023-04-20
WAITP-1186 add example.http to dhis-api

acdunham:
### Changes:
- Added an `example.http` file to `dhis-api` package so that the endpoints can be tested with a rest client.


tcaiger (approved):
Thanks for adding this!

PR #4482, opened 2023-04-21
WAITP-1183 create hexcode preview input

acdunham:
### Changes:
- Added new field input to `ui-components` for hexcode input, which is a text field with preview of the input color. Only shows preview color if is a valid input.
- Implemented to allow configuration of field type `hexcode` in admin panel

### Screenshots:
image


packages/ui-components/src/components/Inputs/HexcodeField.js
```diff
@@ -0,0 +1,84 @@
...
+const HexcodeFieldContainer = styled.div`
+  display: flex;
+  align-items: flex-end;
+  margin-top: 1em;
+`;
+
+const HexcodePreview = styled.div`
+  background-color: ${({ backgroundColor }) => backgroundColor};
+  height: 100%;
+  width: 100%;
```
tcaiger, line 20:
```suggestion
  height: 100%;
  width: 100%;
```
Currently these are getting over written below. Are they meant to be max width?

packages/ui-components/src/components/Inputs/HexcodeField.js
```diff
@@ -0,0 +1,84 @@
...
+  border-radius: 50%;
+  width: 3em;
+  height: 3em;
+  margin-left: 1em;
+  margin-bottom: 0.2em;
+  border: ${({ theme }) => `1px solid ${theme.palette.common.black}`};
+`;
+
+const HexInput = styled(TextField)`
+  max-width: 18em;
```
tcaiger, line 30:
```suggestion
  max-width: 18em;
```
This seems like a sensible value but we probably don't want to hard code this right? Better to leave it to the parent to determine.

packages/ui-components/src/components/Inputs/HexcodeField.js
```diff
@@ -0,0 +1,84 @@
...
+  margin-bottom: 0;
+`;
+
+const HEXCODE_PATTERN = /^#([0-9A-F]{3}){1,2}$/i;
```
tcaiger, line 34: Nice

packages/ui-components/src/components/Inputs/HexcodeField.js
```diff
@@ -0,0 +1,84 @@
...
+          inputProps={{
+            'aria-describedby': helperText ? `${id}-helper-text` : null,
+            pattern: HEXCODE_PATTERN,
+          }}
+        />
+        {getIsValidHexCode() && (
+          <HexcodePreview backgroundColor={value} aria-label={`Colour preview for ${label}`} />
+        )}
+      </HexcodeFieldContainer>
+      {helperText && <FormHelperText id={`${id}-helper-text`}>{helperText}</FormHelperText>}
```
tcaiger, line 63: Out of interest, what is the id for? Does that help people with screen readers?

tcaiger (changes requested):
Looks great and very speedy. I left a few small questions and comments.

packages/ui-components/src/components/Inputs/HexcodeField.js
acdunham: Yep, so in the case of helper text, this is considered to be a additional description for the input. In order to link that to the input so the screen reader can announce to the user that it has an extra description, we put a unique id onto the helper text and give the input an `aria-describedby` attribute of the helper text's id.

tcaiger (approved):

PR #4486, opened 2023-04-23
Waitp 1192 create tooltip in admin panel

acdunham:
### Changes:
- Created a new `InputLabel` component that allows a tooltip to be set as well as label text. Also allows setting of different elements, in case of being wrapped in an MUI form element that includes `label` tag, or using as `legend` etc. 
- Updated all inputs in `ui-components` to use the new `InputLabel` as the label component


packages/ui-components/src/components/Inputs/Autocomplete.js
```diff
@@ -145,4 +148,5 @@ Autocomplete.defaultProps = {
   inputRef: null,
   name: null,
   defaultValue: null,
+  tooltip: '',
```
tcaiger, line 151: Any reason not to make it null?

acdunham: Purely because it's a string type, so could be either really!

packages/ui-components/src/components/Inputs/TextField.js
```diff
@@ -128,3 +129,22 @@ export const TextField = styled(BaseTextField)`
     padding: 10px;
   }
 `;
+
+export const TextField = ({ label, tooltip, ...props }) => (
+  <StyledTextField
+    fullWidth
+    {...props}
+    variant="outlined"
+    label={<InputLabel label={label} tooltip={tooltip} as="span" />}
```
tcaiger, line 138: Very nice way of extending the form fields

packages/ui-components/src/components/Inputs/InputLabel.js
```diff
@@ -0,0 +1,65 @@
...
+import { Tooltip } from '../Tooltip';
+
+/** Styled label for inputs. Handles tooltips for labels if present. */
+
+const LabelWrapper = styled.span`
+  display: flex;
+  align-items: center;
+`;
+
+const TooltipWrapper = styled.span`
```
tcaiger, line 19: I know that I said not to worry about matching the design for the field grouping but since this is a new component I think we should make it look like the design as much as possible (sorry for the inconsistency).

tcaiger (changes requested):
Nice work! Great component design with the labels. I left a request about styling. Also there is some work to plumb the tooltip in the admin panel which just involves passing the tooltip prop through a few places. It could be done here or in a seperate PR - whichever you prefer.

packages/ui-components/src/components/Inputs/InputLabel.js
acdunham: Sure - the only thing is this label component gets used by all the inputs, so if I style it to match the design, it will change all the labels everywhere - is that okay?

tcaiger (approved):
Awesome - looks so good now!

PR #4488, opened 2023-04-24
WAITP-1161 add option to hide null value from legend

acdunham:
### Changes:
- Added support for `hideNullFromLegend` to `spectrum` and `shaded-spectrum` type overlays. 
- If set to be `true`, the `No data` option will not show up in the legend. 
- If there is no other data at all, the legend won't display at all.

Note: setting `hideNullFromLegend` does not exclude `null` values from appearing on the map altogether. In order to do that, there is already another setting called `hideByDefault` where this can be applied. The reason for this is that hiding the value from the legend does not always mean you want to hide from the map, and vice-versa, so it seemed best to keep these decoupled. There is also an existing setting called `hideFromLegend` but this hides the entire legend for the overlay, and I didn't think it would be a good option to reformat that to be an object like `hideByDefault`, since that would require checking the types of values in order to be backwards compatible. 

### Example
- Update config for an overlay as explained above, and the legend option should no longer be visible.


packages/ui-components/src/components/Map/Legend/Legend.js
```diff
@@ -78,11 +78,16 @@ export const Legend = React.memo(
...
-          // Spetrum legend has values = []
-          (values.length === 0 || values.filter(value => !value?.hideFromLegend).length > 0),
+        ({ type, hideFromLegend, values = [], min, max, noDataColour }) => {
+          // if type is radius or popup-only, don't create a legend
+          if ([MEASURE_TYPE_RADIUS, MEASURE_TYPE_POPUP_ONLY].includes(type)) return false;
+          // if hideFromLegend is true, don't create a legend
+          if (hideFromLegend === true) return false;
+          // if type is spectrum or shaded-spectrum, only create a legend if min and max are set OR noDataColour is set. If noDataColour is not set, that means hideNullFromLegend has been set as true in the map overlay config. Spectrum legends 'values' property will always be []
+          if ([MEASURE_TYPE_SHADED_SPECTRUM, MEASURE_TYPE_SPECTRUM].includes(type))
+            return noDataColour ? true : !(min === null || max === null);
```
EMcQ-BES, line 88: Should probably check for `undefined` here too?

Can't do a regular falsey check since `min = 0` is a valid value

packages/ui-components/src/components/Map/Legend/Legend.js
```diff
@@ -78,11 +78,16 @@ export const Legend = React.memo(
...
-        ({ type, hideFromLegend, values = [] }) =>
-          ![MEASURE_TYPE_RADIUS, MEASURE_TYPE_POPUP_ONLY].includes(type) &&
-          hideFromLegend !== true &&
-          // Spetrum legend has values = []
-          (values.length === 0 || values.filter(value => !value?.hideFromLegend).length > 0),
+        ({ type, hideFromLegend, values = [], min, max, noDataColour }) => {
+          // if type is radius or popup-only, don't create a legend
+          if ([MEASURE_TYPE_RADIUS, MEASURE_TYPE_POPUP_ONLY].includes(type)) return false;
+          // if hideFromLegend is true, don't create a legend
+          if (hideFromLegend === true) return false;
```
EMcQ-BES, line 85:
```suggestion
          if (hideFromLegend) return false;
```

I have a personal vendetta against checking boolean values like this 😤

EMcQ-BES (changes requested):
Basic structure looks good, couple small comments

EMcQ-BES (approved):

PR #4489, opened 2023-04-24
WAITP-1184 field grouping support for admin panel

acdunham:
### Changes:
- Created a `ui-component` called `InputGroup` that displays a title, description, and an array of fields
- Updated `admin-panel` edit files to handle groups of fields. To maintain backwards compatibility, this is done by passing in an optional array of `sections` to the `ResourcePage` component. Each item in the array will have an array of field names, which must be defined as fields are currently, as well as an optional title and description. The `Editor` will group the fields accordingly. This seemed the simplest way to handle these without needing to refactor everything else? 

Note: if sections are defined together with fields that don't belong to a section, the sections will appear first, and then the standalone fields. 
Also note that designs are not set in stone, so some of the styling may change.

### Screenshots:
image

### DIFF
--- a/packages/admin-panel/src/editor/Editor.js
+++ b/packages/admin-panel/src/editor/Editor.js
@@ -8,6 +8,7 @@ import PropTypes from 'prop-types';
 import { InputGroup } from '@tupaia/ui-components';
 import { InputField } from '../widgets';
 import { checkVisibilityCriteriaAreMet, labelToId } from '../utilities';
+import { SECTION_FIELD_TYPE } from './constants';
 
 export const Editor = ({ fields, recordData, onEditField }) => {
   const onInputChange = (inputKey, inputValue, editConfig = {}) => {
@@ -70,7 +71,7 @@ export const Editor = ({ fields, recordData, onEditField }) => {
 
   // Get the form fields and sections that are visible from the fields prop, handling nested sections
   const visibleFormItems = filterVisibleFields(fields).reduce((result, field) => {
-    if (field.type === 'section') {
+    if (field.type === SECTION_FIELD_TYPE) {
       const visibleSubfields = filterVisibleFields(field.fields);
       if (!visibleSubfields.length) return result;
       return [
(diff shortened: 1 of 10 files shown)


packages/admin-panel/src/editor/Editor.js
```diff
@@ -5,10 +5,11 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
+import { InputGroup } from '@tupaia/ui-components';
 import { InputField } from '../widgets';
 import { checkVisibilityCriteriaAreMet, labelToId } from '../utilities';
 
-export const Editor = ({ fields, recordData, onEditField }) => {
+export const Editor = ({ fields, recordData, onEditField, sections }) => {
```
EMcQ-BES, line 12: I think we could probably simplify this a fair bit by just defining sections within the `fields` array? That way this whole function could be simplified to just checking for `type: 'subsection'` and mapping those to `InputGroup`, and it'd still maintain backwards compatibility.

```
fields = [
  {
    type: 'subsection', // Or whatever language you think suits
    title: 'foo',
    description: 'bar',
    fields: [
      {
        // In theory we could eventually support nested sections but that's def beyond the scope of this ticket
        Header: 'Name',
        source: 'name',
      }
    ]
  }
]
```

This way the structure of the data matches the way it's being displayed which simplifies the dev process. It also means we don't have to update both arrays when adding/moving a field, which should completely remove the chance of data mismatches.

EMcQ-BES (approved):
Lookin much cleaner 👍 Can probably clean up those extra uses of the magic string since you already have the constant

Is there anywhere we want to actually add some of these sections at the moment? Would be nice to actually put some in place for testing.

acdunham:
> Lookin much cleaner 👍 Can probably clean up those extra uses of the magic string since you already have the constant
> 
> Is there anywhere we want to actually add some of these sections at the moment? Would be nice to actually put some in place for testing.

Not as yet, because they are for the new custom landing pages. However, the `ui-components` `storybook` has examples from a UI perspective under `InputGroup`

EMcQ-BES:
> Not as yet, because they are for the new custom landing pages. However, the ui-components storybook has examples from a UI perspective under InputGroup

Yep, I saw that, and the screenshots look good 👍

Just wanted to make sure we have an upcoming use case for this so we can test the nesting logic properly. Since we can test  that portion of code along side the landing page work we can probably just run a quick regression against this ticket for now 🙏

PR #4492, opened 2023-04-26
WAITP-1156 handle permissions errors

acdunham:
### Changes:
- If a permissions error is caught, changed to throw the error so it can be handled accordingly in the frontend. Prior to this the frontend didn't know an error was caught, and so the screen would be blank as the FE was trying to access data that was undefined.


packages/web-config-server/src/apiV1/measureData.js
```diff
@@ -12,6 +12,7 @@ import { DataAggregatingRouteHandler } from './DataAggregatingRouteHandler';
 import { MapOverlayPermissionsChecker } from './permissions';
 import { DATA_SOURCE_TYPES } from './dataBuilders/dataSourceTypes';
 import { reportServer } from './dataBuilders';
+import { log } from 'console';
```
EMcQ-BES, line 15:
```suggestion
```

EMcQ-BES (approved):
Nice simple fix 👍 

Looks like codeship is failing, but it looks like it's just some timeout flakey tests so I've restarted the build. Please double check that succeeds

PR #4494, opened 2023-04-27
Rn 832 release fix

chris-pollard:
Issue [RN-832](https://linear.app/bes/issue/RN-832):
Changes:

- A fix for the [RN-832](https://linear.app/bes/issue/RN-832) card, which avoids parsing any string with a percentage symbol and number.
- Also includes an addition to the web-frontend package.json for including data-table-server in the start-servers command.


chris-pollard:
Cancelled due to inclusion of unnecessary commits.

PR #4496, opened 2023-04-27
RN-832: Avoid parsing percentage string

chris-pollard:
### Changes:

- Avoid parsing percentage string for excel download.


biaoli0 (approved):
LGTM!

PR #4500, opened 2023-04-27
WAITP-1170 Fix entity attribute filter on mobile

EMcQ-BES:
### Changes:

- Run entity filtering using `Array.filter` rather than `Realm.Results.filtered`


packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -27,7 +27,7 @@ const NO_RESULTS_SECTION = { title: 'No results', data: [] };
 export class EntityList extends PureComponent {
   constructor(props) {
     super(props);
-    this.allEntitiesQuery = fetchEntities(
```
EMcQ-BES, line 30: Wanted to get rid of the word `query` since it _isn't_ a realm query

packages/meditrak-app/app/entityMenu/helpers.js
```diff
@@ -34,7 +34,7 @@ export const getEntityAttributeChecker = (state, questionId) => {
   }
 
   return entity =>
-    Object.entries(JSON.parse(attributes)).every(([key, config]) => {
```
EMcQ-BES, line 37: Because this is running after `Realm` converts it to a value object it is already parsed into a JSON object

packages/meditrak-app/app/entityMenu/helpers.js
```diff
@@ -78,6 +78,10 @@ export const getRecentEntities = (database, state, baseFilters) => {
   return sortedRecentEntities;
 };
 
+/*
+ * This returns a js array, NOT a Realm.Results object
+ * The attributes filter uses a [].filter which converts the results to value iterables
```
EMcQ-BES, line 83: I originally tried to fix this while maintaining the `Realm.Results` object but that was much more difficult.

`Realm` treats the `attributes` field as a string so we can't compare against the fields within that object. I wasn't able to find a JSON parse within the `Realm` query language so running through `filtered` doesn't work.

packages/meditrak-app/app/entityMenu/EntityList.js
```diff
@@ -188,7 +182,7 @@ export class EntityList extends PureComponent {
     const { isOpen, numberToShow } = this.state;
 
     // if base query is empty, the survey is probably misconfigured
```
chris-pollard, line 184: Change 'query' to 'fetch' ?
` // if base fetch is empty, the survey is probably misconfigured`

chris-pollard (approved):
Nice one!

PR #4501, opened 2023-04-28
Fix CI by reverting change

biaoli0:
Broken build: https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/b482cb2e-633c-4653-98f1-8604c1f24897?component=step_Validate_and_test_Test_Test_batch_1_Test_central-server

Not sure why the build failed on stopping in the middle of creating test database in some specific packages.

This reverts part of commit 989b9230a21df47f18c89deaef6dbc9ef66461f3. The rest should be fine to change.


IgorNadj (approved):
Approving because it is safe to revert this. 

Discussed in slack, Biao is looking into what the actual error is locally since the logs aren't loading on Codeship.

PR #4503, opened 2023-04-28
WAITP-1194 fix type of response expected for events

acdunham:
### Changes:
- Updated the type that is expected from the event check query to be number, not string
- Added tests for `checkIsEventBased` in `SurveyResponse`


IgorNadj:
Closing in favour of #4511

PR #4507, opened 2023-04-28
Hotfix: Pg types parser errors

biaoli0:
### Issue #:
Since we added `pg.setTypeParser()` for parsing bigInt type to integer ([PR](https://github.com/beyondessential/tupaia/pull/4436#discussion_r1179883239)), a bug created in sync queue because originally we treated bigInt type data as string. 

Summary:
1. There is only one place in Tupaia database use bigInt, which is `access_token_expiry`, they got used to compare with Date.now(), this is fine as string and integer types can be used to do this comparsion.
2. Have check all sql queries in Tupaia to see if any mathematic functions would be broken, like count, sum, min, max.. 
3. most count are captured and parsed to Int in the codebase except one.
4. coconuts, pigs got changed from string to int, they are fine for displaying purpose on frontend.

- Example

---

### DIFF
--- a/packages/central-server/src/database/models/SurveyResponse.js
+++ b/packages/central-server/src/database/models/SurveyResponse.js
@@ -99,7 +99,7 @@ export class SurveyResponseModel extends MaterializedViewLogDatabaseModel {
       [surveyResponseId],
     );
     if (result.length === 0) return false;
-    return result[0].count === '1';
+    return parseInt(result[0].count) === 1;
   }
 
   approvalStatusTypes = SURVEY_RESPONSE_APPROVAL_STATUS;
--- a/packages/database/src/TupaiaDatabase.js
+++ b/packages/database/src/TupaiaDatabase.js
@@ -67,6 +67,7 @@ const HANDLER_DEBOUNCE_DURATION = 250;
 // string, independent of timezones, rather than being converted to local time
 pgTypes.setTypeParser(pgTypes.builtins.TIMESTAMP, val => val);
 pgTypes.setTypeParser(pgTypes.builtins.NUMERIC, parseFloat);
+pgTypes.setTypeParser(20, parseInt); // bigInt type to Integer
 
 export class TupaiaDatabase {
   /**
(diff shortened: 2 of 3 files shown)


packages/database/src/TupaiaDatabase.js
```diff
@@ -62,9 +62,12 @@ const COMPARATORS = {
 // keeping all the tests passing
 const HANDLER_DEBOUNCE_DURATION = 250;
 
+// Danger: pgTypes is a global variable, so it will be shared between all instances of TupaiaDatabase
 // turn off parsing of timestamp (not timestamptz), so that it stays as a sort of "universal time"
 // string, independent of timezones, rather than being converted to local time
 pgTypes.setTypeParser(pgTypes.builtins.TIMESTAMP, val => val);
+pgTypes.setTypeParser(pgTypes.builtins.NUMERIC, parseFloat);
+pgTypes.setTypeParser(20, parseInt); // bigInt type to Integer
```
biaoli0, line 70: Move configs in one central place.

packages/central-server/src/database/models/SurveyResponse.js
```diff
@@ -99,7 +99,7 @@ export class SurveyResponseModel extends MaterializedViewLogDatabaseModel {
       [surveyResponseId],
     );
     if (result.length === 0) return false;
-    return result[0].count === '1';
+    return parseInt(result[0].count) === 1;
```
edmofro, line 102: @biaoli0 I like the extra safety here in case we pull out that lower level parsing in future
But, my preference is always for "make things blow up in an obvious way, so you notice and fix it quickly" rather than "have graceful handling that lets unintended changes slip through unnoticed"
@acdunham when you combine this into your one, can you add an assertion that the value within `count` is an integer, as expected? That way we will notice in future with a big fat error log if we break it.

packages/database/src/TupaiaDatabase.js
edmofro: Please include this (and the matching deletions in the next file) in the combined PR @acdunham

IgorNadj:
Closing in favour of #4511

PR #4512, opened 2023-05-01
MDEV-101: Fix table props in admin panel

tcaiger:
### Changes:

- A recent PR for MDEV broke the expandable tables being incorrectly setting a table prop

---


packages/admin-panel/src/table/DataFetchingTable.js
```diff
@@ -239,7 +239,7 @@ DataFetchingTableComponent.defaultProps = {
   numberOfPages: 0,
   nestingLevel: 0,
   deleteConfig: {},
-  TableComponent: null,
+  TableComponent: undefined,
```
tcaiger, line 242: should be undefined instead of null so that it doesn't over write the material ui default component.

acdunham (approved):

PR #4515, opened 2023-05-01
WAITP-1151: Custom landing pages login flow

tcaiger:
### Changes:

- Add mock Landing Page template (most of this doesn't need review) at top level of app
  - Included an additional OverlayDiv in the landing page to re-use login form etc
- Refactored the project card logic into a re-usable ProjectCardList component
- Added utils for navigating to login and logout routes in redux for ease of integration with the existing app
- Hide View Projects from LoginForm and Request Dialogs when it is a custom landing page

I still have to update the login button to stop it re-directing to the explore project. I didn't do that in this PR as it is a bit tricky.

### DIFF
--- a/packages/web-frontend/src/containers/OverlayDiv/components/RequestProjectAccessDialog.js
+++ b/packages/web-frontend/src/containers/OverlayDiv/components/RequestProjectAccessDialog.js
@@ -27,6 +27,7 @@ import {
 import { LANDING } from '../constants';
 import { TUPAIA_LIGHT_LOGO_SRC } from '../../../constants';
 import { BLUE, WHITE, GREY } from '../../../styles';
+import { useCustomLandingPages } from '../../../screens/LandingPage/useCustomLandingPages';
 
 const leftPadding = '40px';
 
@@ -196,6 +197,7 @@ export const RequestProjectAccessComponent = React.memo(
     success,
     errorMessage,
   }) => {
+    const { isCustomLandingPage } = useCustomLandingPages();
     const requestedCountries = countries.filter(c => c.accessRequests.includes(project.code));
     const availableCountries = countries.filter(c => !c.accessRequests.includes(project.code));
 
@@ -204,15 +206,42 @@ export const RequestProjectAccessComponent = React.memo(
       hideForm = true;
 
     const modalMessage = success ? (
-      <SuccessMessage projectName={project.name} onBackToProjects={onBackToProjects} />
+      <>
+        <Alert severity="success">
+          Thank you for your access request to {project.name}. We will review your application and
+          respond by email shortly.
+        </Alert>
+        <Note>
+          Note: This can take some time to process, as requests require formal permission to be
+          granted.
+        </Note>
+        {!isCustomLandingPage && (
+          <BackButton onClick={onBackToProjects}>Back to Projects</BackButton>
+        )}
+      </>
     ) : (
-      <RequestPendingMessage
-        requestedCountries={requestedCountries}
-        availableCountries={availableCountries}
-        setRequestingAdditionalCountryAccess={setRequestingAdditionalCountryAccess}
-        handleRequest={onRequestProjectAdditionalAccess}
-        onBackToProjects={onBackToProjects}
-      />
+      <>
+        <p>
+          <b>You have already requested access to this project</b>
+        </p>
+        <RequestedProjectCountryAccessList
+          requestedCountries={requestedCountries}
+          availableCountries={availableCountries}
+          handleRequest={onRequestProjectAdditionalAccess}
+        />
+        <p>
+          This can take some time to process, as requests require formal permission to be granted.
+        </p>
+        <p>
+          {`If you have any questions, please email: `}
+          <a style={styles.contactLink} href="mailto:admin@tupaia.org">
+            admin@tupaia.org
+          </a>
+        </p>
+        {!isCustomLandingPage && (
+          <BackButton onClick={onBackToProjects}>Back to Projects</BackButton>
+        )}
+      </>
     );
 
     return (
@@ -225,10 +254,12 @@ export const RequestProjectAccessComponent = React.memo(
               world
             </TagLine>
           </div>
-          <ExploreButton onClick={onBackToProjects} variant="outlined">
-            <ExploreIcon />
-            &nbsp; View other projects
-          </ExploreButton>
+          {!isCustomLandingPage && (
+            <ExploreButton onClick={onBackToProjects} variant="outlined">
+              <ExploreIcon />
+              &nbsp; View other projects
+            </ExploreButton>
+          )}
         </Header>
         <Title>Requesting Project Access</Title>
         <HeroImage src={project.imageUrl}>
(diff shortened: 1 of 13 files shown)


packages/web-frontend/src/containers/OverlayDiv/components/RequestProjectAccessDialog.js
```diff
@@ -280,45 +311,6 @@ export const RequestProjectAccessComponent = React.memo(
   },
 );
 
-export const SuccessMessage = ({ projectName, onBackToProjects }) => (
```
tcaiger, line 283: Moved these messages inline

packages/web-frontend/src/screens/LandingPage/useAuth.js
```diff
@@ -0,0 +1,12 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { useSelector } from 'react-redux';
```
tcaiger, line 6: This util might be unnecessary but i imagine that it might end up having more business logic in it over time

packages/web-frontend/src/screens/LandingPage/useNavigation.js
```diff
@@ -0,0 +1,32 @@
...
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { useDispatch } from 'react-redux';
+import { setProject, setRequestingAccess } from '../../projects/actions';
+import { REQUEST_PROJECT_ACCESS, LANDING } from '../../containers/OverlayDiv/constants';
+import { attemptUserLogout, setOverlayComponent } from '../../actions';
+
+export const useNavigation = () => {
```
tcaiger, line 11: I'm doing my best to work with the existing redux navigation while still making it easy to refactor later which is why these are in their seperate util file. Once we refactor to proper routes none of these utils will be necessary.

packages/web-frontend/src/screens/desktop/RootScreen/MainPage.js
```diff
@@ -47,26 +49,33 @@ const MapContainer = styled.div`
...
-      <LoadingScreen isLoading={isLoading} />
-    </Container>
-    <MapContainer $rightOffset={sidePanelWidth}>
-      <Map />
-    </MapContainer>
-  </>
-);
+const MainPage = ({ enlargedDialogIsVisible, isLoading, sidePanelWidth }) => {
+  const { isCustomLandingPage } = useCustomLandingPages();
+  if (isCustomLandingPage) {
```
tcaiger, line 54: I couldn't get the routes in App.js to work with the custom landing page routes as they wouldn't integrate and sync properly with the way we do redux routing so I had to do this slightly clunky way of adding the route which is basically how we handle routes at the moment in the app.

packages/web-frontend/src/containers/UserBar/index.js
```diff
@@ -92,14 +92,16 @@ export class UserBar extends Component {
...
     const width = this.getWidth();
+    // Don't let the user close a dialog if it is a password reset form as the one time login token
+    // is used to load the form and so they need to do the password reset now */
     return (
       <Dialog
         open={isDialogVisible}
         style={{ zIndex: 1301 }}
         PaperProps={{ style: { ...styles.dialogBody, width, maxWidth: width } }}
-        onClose={() => onCloseUserDialog()}
+        onClose={dialogPage === DIALOG_PAGE_RESET_PASSWORD ? null : onCloseUserDialog}
```
acdunham, line 104: Yay for removing unnecessary inline functions! 🎉

packages/web-frontend/src/screens/LandingPage/LandingPage.js
```diff
@@ -0,0 +1,135 @@
...
+  background-size: cover;
+  background-position: center;
+`;
+
+const Container = styled(MuiContainer)`
+  position: relative;
+  min-height: calc(100vh - 75px);
+  background-color: #bfbfbf;
+  padding-left: 50px;
+  padding-right: 50px;
```
acdunham, line 29: Very minor, but would it be possible use `rem`/`em`/`percentages` instead of `px` values, so that will help us down the line with responsive styling?

packages/web-frontend/src/screens/LandingPage/LandingPage.js
```diff
@@ -0,0 +1,135 @@
...
+  padding-top: 5%;
+`;
+
+const Projects = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr;
+  column-gap: 50px;
```
acdunham, line 36: Same as above, perhaps using responsive units of some kind would be cool

packages/web-frontend/src/screens/LandingPage/LandingPage.js
```diff
@@ -0,0 +1,135 @@
...
+  const { projects } = useCustomLandingPages();
+  const { isUserLoggedIn, currentUserUsername } = useAuth();
+
+  return (
+    <>
+      <NavBar>
+        {currentUserUsername && (
+          <Box mr={3}>
+            {currentUserUsername}
+            <span style={{ marginLeft: 30 }}>|</span>
```
acdunham, line 92: A couple of things here:
- Personally I hate inline styling, and it doesn't really fit the pattern we use elsewhere so having a styled-component is preferable here I think
- Text values should always be wrapped in a text element, so perhaps making a styled component for the text would solve both problems
- Since the `|` isn't a value that we need to read, it's purely visual, I'd suggest doing it via styling (borders, or pseudo elements if needed), just so that the screen reader wouldn't try to read this out.

packages/web-frontend/src/screens/LandingPage/LandingPage.js
```diff
@@ -0,0 +1,135 @@
...
+              accessType="pending"
+              action={navigateToRequestProjectAccess}
+              actionText="Approval in progress"
+            />
+            <ProjectCardList
+              projects={projects}
+              accessType={false}
+              action={isUserLoggedIn ? navigateToRequestProjectAccess : navigateToLogin}
+              actionText={isUserLoggedIn ? 'Request access' : 'Log in'}
+            />
```
acdunham, line 125: I love the reusable component for `ProjectCardList`. One thought, which may or may not be useful, is that it looks like wherever `ProjectCardList` gets used it displays all three access types, with the same text, is that right? If so, would it make sense to make `ProjectCardList` just accept `projects` and potentially `actions` as an object keyed by the different types (or some similar way of passing in actions), and then `ProjectCardList` could loop through the projects and display accordingly?

packages/web-frontend/src/screens/LandingPage/useNavigation.js
acdunham: Nice work with these, seems like a good way to make a simple use of what we already have.

acdunham (commented):
Nice work, just a few suggested changes from me, but overall looks good!

PR #4517, opened 2023-05-02
RN-801: Add Data Grid to Viz Builder

chris-pollard:
### Changes:

- Example

---


acdunham (commented):
Looks good to me. Probably a good idea to have a decent amount of regression testing on the off chance something changed unintentionally.

tcaiger (approved):
Looks good to me. I agree with Alex that it will need a pretty good regression test.

PR #4521, opened 2023-05-02
Waitp 1185 landing pages tab

acdunham:
### Changes:
- Added route and tab to landing pages view
- Updated table columns on landing pages view to allow sorting of order, custom headers in columns, and a clickable link to the custom link

### Screenshots:
With fake data:
Screen Shot 2023-05-02 at 4 03 15 pm

### DIFF
--- a/packages/admin-panel/src/pages/resources/CustomLandingPagesPage.js
+++ b/packages/admin-panel/src/pages/resources/CustomLandingPagesPage.js
@@ -13,260 +13,242 @@ const URL_PREFIX = window.location.origin.replace('admin', 'www');
 
 const DISPLAY_URL_PREFIX = `${URL_PREFIX.replace(new RegExp('(https://)|(http://)'), '')}/`;
 
-// All the fields for create/edit of a custom landing page
-const FIELDS = [
-  {
-    type: SECTION_FIELD_TYPE,
-    fields: [
-      {
-        Header: 'Name',
-        source: 'name',
-        // this is to denote this field as a column in the table
-        column: {
-          // sets what order to put the column in the table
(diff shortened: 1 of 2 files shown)


packages/admin-panel/src/pages/resources/CustomLandingPagesPage.js
```diff
@@ -8,9 +9,9 @@ import { prettyArray } from '../../utilities';
...
 // the URL prefix to display in the url_segment field
-const URL_PREFIX = `${window.location.origin
-  .replace('admin', 'www')
-  .replace(new RegExp('(https://)|(http://)'), '')}/`;
+const URL_PREFIX = window.location.origin.replace('admin', 'www');
+
+const DISPLAY_URL_PREFIX = `${URL_PREFIX.replace(new RegExp('(https://)|(http://)'), '')}/`;
 
 // All the fields for create/edit of a custom landing page
 const FIELDS = [
```
IgorNadj, line 17: Suggestion:
FIELDS lists fields flat { name: ..., code: ... }

EDIT_FIELDS has nesting and plucks from FIELDS e.g. [type: section, fields: [FIELDS.name, FIELDS.code]] or default [...Object.values(FIELDS)]

COLUMNS plucks from FIELDS e.g. [FIELDS.name, FIELDS.code] or default [...FIELDS]

Decorating FIELDS inline to reorder columns etc. is a great idea, my suggestion would be to keep it outside because in the future we will get FIELDS straight from the db and largely autogenerate the admin panel.

IgorNadj (approved):
One suggestion, otherwise looks great thanks!

PR #4526, opened 2023-05-03
Waitp 1189a landing pages UI template part 1

acdunham:
### Changes:
- refactored top bar and logo components to be reusable across landing pages and main views
- Updated user menu and user bar to decouple user menu and make it standalone
- Separated out the single and multi project landing page views
- Refactored project card list to handle the project filtering by status


packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/ProjectCard.js
```diff
@@ -130,11 +134,15 @@ export const ProjectCard = ({
         <Button
```
tcaiger, line 134: This suggestion is optional - just if you are interested. 
It looks like all the logic is in the button and it would be nice if the ProjectCard was just a display component that wasn't aware of the access type etc. There are a couple of ways this would be made more streamlined but one method I think would work well is to use composed components which would invert some of the control to the ProjectCardList which already knows about the access types. You could export a `ProjectLockButton`, `ProjectPendingButton` and a `ProjectSelect` button and then use it in the ProjectCardList. 
eg.

```
import { ProjectCard, onSelectProject } from '../ProjectCard';

<ProjectCard Button={<ProjectButton action={onSelectProject} />} />

File: packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/ProjectCardList.js [hunk 2]
```diff
@@ -5,26 +5,51 @@
...
+    [PENDING_PROJECT_ACCESS_TYPE]: 'Approval in progress',
+    [ALLOWED_PROJECT_ACCESS_TYPE]: 'View project',
+    [DENIED_PROJECT_ACCESS_TYPE]: isUserLoggedIn ? 'Request access' : 'Log in',
+  };
+  const sortedProjects = [
+    ALLOWED_PROJECT_ACCESS_TYPE,
+    PENDING_PROJECT_ACCESS_TYPE,
+    DENIED_PROJECT_ACCESS_TYPE,
+  ].reduce((result, accessType) => {
+    const action = actions[accessType];
```
tcaiger, line 27: One minor maintenance concern is that the actions have to have these 3 keys otherwise it will break. Maybe we could add prop types for the actions and/or hardcode the actions into 3 props to reflect the project types.

packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/ProjectCardList.js
```diff
@@ -5,26 +5,51 @@
...
+        })),
+    ];
+  }, []);
+  return sortedProjects.map(project => (
+    <ProjectCard
+      key={project.name}
+      projectAction={() => project.action(project)}
+      actionText={project.actionText}
+      accessType={project.accessType}
+      {...project}
```
tcaiger, line 52: There is a convention that Edwin likes where we list out the object properties so that it's more transparent about what data is getting passed in. There are exceptions where we don't do it if we are extending a component and would have to list lots of props just to add one new one but here it seems like a good idea.
```suggestion
 name={project.name}
 etc
```

packages/web-frontend/src/constants/constants.js
```diff
@@ -36,3 +36,7 @@ export const LOGIN_TYPES = {
   TOKEN: 'token',
   MANUAL: 'manual',
 };
+
+export const PENDING_PROJECT_ACCESS_TYPE = 'PENDING';
```
tcaiger, line 40: Good idea adding these as conts. What do you think about making them a set. eg.
```suggestion
export const PROJECT_ACCESS_TYPES = {
  PENDING: 'PENDING',
  DENIED: 'DENIED',
  ALLOWED: 'ALLOWED'
```

packages/web-frontend/src/containers/TopBar/index.js
```diff
@@ -6,32 +6,49 @@
...
-    display: 'flex',
-    zIndex: 1000,
-    position: 'relative',
-    padding: '0 10px',
-    borderBottom: '1px solid rgba(151, 151, 151, 0.3)',
-  },
-};
+// Both min height and height must be specified due to bugs in Firefox flexbox
+// that means that topbar height will be ignored even if using flex-basis.
+const TopBarWrapper = styled.div`
```
tcaiger, line 18: Thank you for converting to styled-components 🙏

packages/web-frontend/src/screens/desktop/RootScreen/MainPage.js
```diff
@@ -51,28 +51,34 @@ const MapContainer = styled.div`
 
 const MainPage = ({ enlargedDialogIsVisible, isLoading, sidePanelWidth }) => {
   const { isCustomLandingPage } = useCustomLandingPages();
-  if (isCustomLandingPage) {
```
tcaiger, line 54: If it's ok I would prefer to have the landing page as it's own template with it's own EnvBanner and TopBar as it will save a bunch of conditional statements on isCustomLandingPage. I always find conditionals are hard to read and don't scale well.

tcaiger (changes requested):
Great refactoring - so much better and it's really good to be re-using the Topbar component. I left a few comments. Let me know if you have any questions or by all means push back on anything.

packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/ProjectCardList.js
```diff
@@ -4,27 +4,123 @@
...
+    {text}
+  </Button>
+);
+
+export const ProjectCardList = ({ projects, actions, isUserLoggedIn }) => {
+  const actionTexts = {
+    [PROJECT_ACCESS_TYPES.PENDING]: 'Approval in progress',
+    [PROJECT_ACCESS_TYPES.ALLOWED]: 'View project',
+    [PROJECT_ACCESS_TYPES.DENIED]: isUserLoggedIn ? 'Request access' : 'Log in',
+  };
```
tcaiger, line 57:
```suggestion
  
```

packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/ProjectCardList.js
```diff
@@ -4,27 +4,123 @@
...
+    <LockIcon />
+    {text}
+  </Button>
+);
+
+// eslint-disable-next-line react/prop-types
+const ProjectPendingButton = ({ action, text }) => (
+  <StyledPendingButton onClick={action} variant="contained">
+    <AlarmIcon />
+    {text}
```
tcaiger, line 42:
```suggestion
Approval in progress
```

packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/ProjectCardList.js
```diff
@@ -4,27 +4,123 @@
...
+// eslint-disable-next-line react/prop-types
+const ProjectDeniedButton = ({ action, text }) => (
+  <Button onClick={action} color="primary" variant="outlined">
+    <LockIcon />
+    {text}
+  </Button>
+);
+
+// eslint-disable-next-line react/prop-types
+const ProjectPendingButton = ({ action, text }) => (
```
tcaiger, line 39:
```suggestion
const ProjectPendingButton = ({ action }) => (
```

packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/ProjectCardList.js
```diff
@@ -4,27 +4,123 @@
...
+    <AlarmIcon />
+    {text}
+  </StyledPendingButton>
+);
+// eslint-disable-next-line react/prop-types
+const ProjectAllowedButton = ({ action, text }) => (
+  <Button onClick={action} variant="contained" color="primary">
+    {text}
+  </Button>
+);
```
tcaiger, line 50:
```suggestion
const ProjectAllowedButton = ({ action }) => (
  <Button onClick={action} variant="contained" color="primary">
    Approval in progress
  </Button>
);
```

tcaiger (approved):
Great work. Thanks for all the changes!

PR #4528, opened 2023-05-04
Docker lastpass

amenasse:
A Dockerfile for downloading environment config from LastPass and writing to a directory. 

Attempting multiple logins to LastPass from the same IP in succession often results in LastPass blocking the IP for some time. To avoid getting blocked, a  "non essential" container can be run in the ECS Task to fetch configuration from LastPass prior to the application containers starting. The configuration is written to a Docker volume which can be mounted into each application container.

The container uses a script slightly modified from [scripts/bash/downloadEnvironmentVariables.sh](https://github.com/beyondessential/tupaia/blob/c4f45f37516904843d278047b8908f4c0134aec6/scripts/bash/downloadEnvironmentVariables.sh)

The following changes have been made to the original script:

- By not being in the ../../scripts/bash directory, modifications to this script don't invalidate the Docker cache prior to the application packages build. This was done to speed up development  of the script.

- Containers running in the ECS task will read env files from a common volume mount. To support this a destination directory needs to be provided as the first argument.

- The fallback to 'dev' environments has been removed, if the specified deployment name is missing from the LastPass vault the script will error.


packages/devops/downloadenv.Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Fetch environment from LastPass and write to /env/<package>/.env
+# Should be run by the tupaia ECS task prior to starting the service containers.
+FROM alpine:3.17.3
+RUN apk --no-cache add bash && \
+    apk --no-cache add lastpass-cli
```
passcod, line 5:
```suggestion
RUN apk --no-cache add bash lastpass-cli
```

IgorNadj (approved):

scripts/docker/downloadEnvironmentVariables.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash -e
+
+# Fetch environment from LastPass, write to ${ENV_DEST}/${DEPLOYMENT_NAME}/.env
+# Is modified from ../../scripts/bash/downloadEnvironmentVariables.sh
```
IgorNadj, line 4: Minor: downloadEnvironmentVariables.sh will be deleted once we swap out to this new infra. So not a biggie.

packages/devops/downloadenv.Dockerfile
```diff
@@ -0,0 +1,13 @@
+# Fetch environment from LastPass and write to /env/<package>/.env
```
IgorNadj, line 1: packages/devops/docker if its not too annoying with layer invalidation

amenasse: Done in 697d87be786a16dfacaa14e14d27f3510a187940
Added packages/devops/docker to .dockergnore so invalidation won't be a problem

PR #4531, opened 2023-05-04
Ant frontend container

amenasse:
This is based off the branch from #4528 , that should be reviewed and merged first.

Introduces a frontend.Dockerfile which uses the existing nginx configuration to serve the frontend services. The current Dockerfile has been updated to not build the frontend packages.

In hindsight this could probably be combined into a single Dockerfile with different stages for frontend and backend services.
https://docs.docker.com/build/building/multi-stage/#stop-at-a-specific-build-stage

A script is run at boot to copy the .env files mounted at /env into the web root of each package.. this will be fetched by the app on startup (not yet implemented)


Dockerfile
```diff
@@ -50,14 +50,17 @@ COPY ./tsconfig* babel.config.json tsconfig-js.json jest.config-ts.json .eslintr
 # Build and install internal dependencies
```
IgorNadj, line 50: Need to consider index.html as well

scripts/docker/nginx/create-env.sh
```diff
@@ -0,0 +1,41 @@
...
+
+    echo "}" >> "$env_js"
+}
+
+
+for p in "$tupaia_package_dir"/*; do
+    package=$(basename "$p")
+    if [ -f "$tupaia_env_dir/$package/.env" ]; then
+        echo  "creating $p/.env-config.js"
+        env-js "$tupaia_env_dir/$package/.env" "$p/.env-config.js"
```
IgorNadj, line 39: Either output .env-config.js into /build so it can be served alongside everything else, or we need to update nginx config to serve it.

scripts/docker/nginx/create-env.sh
```diff
@@ -0,0 +1,41 @@
...
+    fi
+done
```
IgorNadj, line 41: Discussed
`const baseUrl = process.env.REACT_APP_CONFIG_SERVER_BASE_URL || window.env.REACT_APP_CONFIG_SERVER_BASE_URL || '...'`

scripts/docker/nginx/create-env.sh
amenasse: Resolved in f815dcc5e3459af14a91ca0fd8224e1077805a1a

PR #4545, opened 2023-05-08
WAITP-1148: Integrate landing page route with backend

tcaiger:
### Changes:

- fetch landing page data using redux saga
  - I was hoping to side step sagas but it ended up being easier to follow the existing pattern for now. This can be refactored to react-query in the re-write
- tweak landing page data in front end to match backend


packages/web-frontend/src/screens/LandingPage/LandingPage.js
```diff
@@ -23,7 +23,7 @@ const Wrapper = styled.div`
   position: relative;
   background-size: cover;
   background-position: center;
-  background-color: #000000;
+  background-color: #262834;
```
acdunham, line 26: Maybe just check this with Felicity - one of the tickets said somewhere the default background was black?

acdunham (approved):
Just one small question I added around the default colour (super minor), otherwise looks good to me from a FE perspective. I'll leave to Ethan to review those small backend changes

packages/web-frontend/src/screens/LandingPage/LandingPage.js
tcaiger: Oh yes you are right. I found black quite jarring when I was testing so went for this which is used in the loader. I've added a note for Flic in the ticket to check.

PR #4547, opened 2023-05-09
WAITP-1142 connect file upload to landing pages

acdunham:
### Changes:
- Created handlers for creating and editing landing pages to allow for image upload handling
- Updated S3Client util to handle different image types because the previous way would not correctly upload svg images. 
- Added validation S3Client util to throw an error if the file is not an image type
- Fixed casing error in validator util: `isUrl` became `isURL`

### DIFF
--- a/packages/admin-panel/src/pages/resources/CustomLandingPagesPage.js
+++ b/packages/admin-panel/src/pages/resources/CustomLandingPagesPage.js
@@ -285,6 +285,13 @@ export const CustomLandingPagesPage = ({ getHeaderEl }) => {
       columns={COLUMNS}
       getHeaderEl={getHeaderEl}
       createConfig={CREATE_CONFIG}
+      onProcessDataForSave={(editedFields, recordData) => {
+        // If the landing page is being edited, and the url_segment field is not being edited, then include the existing url_segment in the edited fields so that it can be used for generating landing page image names.
+        if (recordData.url_segment && !editedFields.url_segment) {
+          return { ...editedFields, url_segment: recordData.url_segment };
+        }
+        return editedFields;
+      }}
     />
   );
 };
--- a/packages/central-server/src/apiV2/landingPages/CreateLandingPage.js
+++ b/packages/central-server/src/apiV2/landingPages/CreateLandingPage.js
@@ -1,0 +1,55 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+/* eslint-disable camelcase */
+import AWS from 'aws-sdk';
+import { S3Client } from '@tupaia/utils';
+import { BESAdminCreateHandler } from '../CreateHandler';
+import { getStandardisedImageName } from '../../utilities';
+
+export class CreateLandingPage extends BESAdminCreateHandler {
+  async uploadImage(encodedImage, landingPageUrlSegment, type) {
+    const s3Client = new S3Client(new AWS.S3());
+    // Upload the image with a standardised file name and upload to s3.
+    const imagePath = await s3Client.uploadImage(
+      encodedImage,
+      getStandardisedImageName(landingPageUrlSegment, type),
+      true,
+    );
+    return imagePath;
+  }
+
+  async createRecord() {
+    const { url_segment, image_url, logo_url } = this.newRecordData;
+
+    await this.models.wrapInTransaction(async transactingModels => {
+      // Add the project, and then upload the images afterward, so that if an error is caught when creating the record, the images aren't uploaded unnecessarily
+      const newLandingPage = await transactingModels.landingPage.create({
+        ...this.newRecordData,
+        image_url: '',
+        logo_url: '',
+      });
+      await this.insertImagePaths(
+        transactingModels,
+        newLandingPage.id,
+        url_segment,
+        image_url,
+        logo_url,
+      );
+
+      return newLandingPage;
+    });
+  }
+
+  async insertImagePaths(models, landingPageId, landingPageUrlSegment, imageUrl, logoUrl) {
+    // image_url and logo_url are currently required fields, so the validator will error before this point if either of these is falsey.
+    const updates = {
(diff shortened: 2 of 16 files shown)


packages/admin-panel/src/pages/resources/CustomLandingPagesPage.js
```diff
@@ -285,6 +285,13 @@ export const CustomLandingPagesPage = ({ getHeaderEl }) => {
       columns={COLUMNS}
       getHeaderEl={getHeaderEl}
       createConfig={CREATE_CONFIG}
+      onProcessDataForSave={(editedFields, recordData) => {
+        // If the landing page is being edited, and the url_segment field is not being edited, then include the existing url_segment in the edited fields so that it can be used for generating landing page image names.
```
IgorNadj, line 289: I see the intent but the server should be able to pull this up from the existing record, I'd say doing it server side will be clearer than in the frontend. Also api-first thinking means that in the future if another client wants to call this same API the behaviour will not need to be reimplemented.

packages/central-server/src/apiV2/landingPages/CreateLandingPage.js
```diff
@@ -0,0 +1,55 @@
...
+      true,
+    );
+    return imagePath;
+  }
+
+  async createRecord() {
+    const { url_segment, image_url, logo_url } = this.newRecordData;
+
+    await this.models.wrapInTransaction(async transactingModels => {
+      // Add the project, and then upload the images afterward, so that if an error is caught when creating the record, the images aren't uploaded unnecessarily
```
IgorNadj, line 28: Great comment!

packages/central-server/src/apiV2/landingPages/EditLandingPage.js
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import AWS from 'aws-sdk';
+import { S3Client } from '@tupaia/utils';
+import { BESAdminEditHandler } from '../EditHandler';
+import { getStandardisedImageName } from '../../utilities';
+
+export class EditLandingPage extends BESAdminEditHandler {
+  async uploadImage(encodedImage = '', landingPageUrlSegment, type) {
+    // If the image is not a base64 encoded image, then it is already a url to an image in S3, so we can just return that url as it is. This is on the off chance that the image is being uploaded as a url instead of an encoded image, or it is not being updated at all.
```
IgorNadj, line 12: I've seen a couple PRs with this logic. Not sure why we didn't just pick one method and stick with it, and why we support two? (The api receiving a url or a base64 encoded image)

packages/central-server/src/apiV2/landingPages/EditLandingPage.js
```diff
@@ -0,0 +1,38 @@
...
+    // Upload the image with a standardised file name and upload to s3
+    const imagePath = await s3Client.uploadImage(
+      encodedImage,
+      getStandardisedImageName(landingPageUrlSegment, type),
+      true,
+    );
+    return imagePath;
+  }
+
+  // Before updating the landingPage, if the image_url and logo_url have changed, we need to upload the new images to S3 and update the image_url and logo_url fields with the new urls. This also will handle deleting of the image_url and logo_url fields, as the uploadImage function will return the original value if the encoded image is null or not a base64 encoded image.
```
IgorNadj, line 24: I'm assuming the user can delete the image for the record in the frontend somehow, and this will pass up an empty string as the base64 encoded image? 

If so a simpler comment might be:
```
- If image_url/logo_url are undefined we are not editing them
- If image_url/logo_url are empty string, we are deleting 
```

I would also probably just create a convenience function deleteImage that you can call in the second case, just makes readability a tiny bit better.

packages/central-server/src/apiV2/landingPages/CreateLandingPage.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+/* eslint-disable camelcase */
+import AWS from 'aws-sdk';
+import { S3Client } from '@tupaia/utils';
+import { BESAdminCreateHandler } from '../CreateHandler';
+import { getStandardisedImageName } from '../../utilities';
+
+export class CreateLandingPage extends BESAdminCreateHandler {
+  async uploadImage(encodedImage, landingPageUrlSegment, type) {
```
IgorNadj, line 13: Opportunity to DRY uploadImage(), maybe a `apiV2/landingPages/uploadImage.js`

packages/utils/src/s3/S3Client.js
```diff
@@ -105,18 +105,29 @@ export class S3Client {
    * @param {*} [allowOverwrite]
    */
   async uploadImage(base64EncodedImage, fileId, allowOverwrite = false) {
-    const buffer = Buffer.from(
-      base64EncodedImage.replace(/^data:image\/\w+;base64,/, ''),
-      'base64',
-    );
+    const imageTypes = ['png', 'jpeg', 'jpg', 'gif', 'svg+xml'];
```
IgorNadj, line 108: We're slowly moving things from utils to tsutils. New stuff goes into tsutils but also big changes to existing files I think we should move over. Totally ok if you don't want to or don't have time for this one, just thought I'd mention in case you want to :)

packages/central-server/src/tests/apiV2/landingPages/EditLandingPage.test.js
```diff
@@ -0,0 +1,143 @@
+/**
```
IgorNadj, line 1: Great tests, thanks!

IgorNadj (changes requested):
Couple small comments, otherwise looks great! Thanks!

acdunham:
Merging this in at the request of @tcaiger so that we can start testing the epic-landing-pages branch in it's entirety, and if there are any further changes I will make another PR.

IgorNadj (commented):
Changes look great, thanks for that!

PR #4549, opened 2023-05-09
[no-issue]: Got setupTestDatabase script working on RDS

rohan-bes:
Because we can't set SUPERUSER on RDS, needed to modify the script a little to handle that.


chris-pollard (approved):
Looking forward to having a test database in the RDS!

PR #4554, opened 2023-05-11
WAITP-1148: landing page loader

tcaiger:
### Changes:
- Update the app wide loading spinner to keep loading after the app has mounted and then turn it off after the landing page data fetch is done


packages/web-frontend/src/utils/request.js
```diff
@@ -164,7 +164,7 @@ export const download = async (resourceUrl, errorFunction, options, fileName) =>
   }
 };
 
-export const getAbsoluteApiRequestUri = (resourceUrl) => {
+export const getAbsoluteApiRequestUri = resourceUrl => {
```
tcaiger, line 167: this is just formatting

packages/web-frontend/public/index.html
```diff
@@ -150,6 +159,10 @@
     <noscript>
       <p style="color: white">You need to enable JavaScript to run this app.</p>
     </noscript>
-    <div id="root"><div class="spinner"></div></div>
+    <!-- App wide loading spinner.
+      It is placed in the index.html file so that it displays while the react app is mounting and it
+      is outside of the root so that it still displays while we fetch essential initial app data (eg. landing pages)-->
+    <div id="tupaia-spinner"></div>
```
tcaiger, line 165: this feels a bit heavy handed but it was the only way that I could get the loader to continue seamlessly from the first render of the page right though until the landing page data is fetched.

packages/web-frontend/src/screens/desktop/RootScreen/MainPage.js
```diff
@@ -50,8 +50,21 @@ const MapContainer = styled.div`
...
 
 const MainPage = ({ enlargedDialogIsVisible, isLoading, sidePanelWidth }) => {
-  const { isCustomLandingPage } = useCustomLandingPages();
-  if (isCustomLandingPage) return <LandingPage />;
+  const { isCustomLandingPageLoading, isCustomLandingPage } = useCustomLandingPages();
+
+  // Once the custom landing pages are loaded, hide the tupaia app loader in index.html
+  useEffect(() => {
+    if (isCustomLandingPageLoading === false) {
+      const el = document.getElementById('tupaia-spinner');
```
tcaiger, line 58: the loading element is outside of the react app so we can't use a ref like we normally would

packages/web-frontend/src/projects/sagas.js
```diff
@@ -24,14 +25,26 @@ export function* fetchProjectData() {
...
+      type: CUSTOM_LANDING_PAGE_LOADING,
+      isLoading: false,
+    });
   } catch (error) {
+    yield put({
+      type: CUSTOM_LANDING_PAGE_LOADING,
+      isLoading: false,
+    });
     // eslint-disable-next-line no-console
     console.error(error);
```
acdunham, line 49: Potentially if you wanted to be extra fancy you could use a `finally` instead of repeating the same call twice, but this is very minor

packages/web-frontend/src/screens/LandingPage/useCustomLandingPages.js
```diff
@@ -2,6 +2,7 @@
  * Tupaia
  *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
+import { useEffect } from 'react';
```
acdunham, line 5: Does this end up getting used?

acdunham (approved):
A couple of minor things, but happy to pre-approve

packages/web-frontend/src/screens/LandingPage/useCustomLandingPages.js
tcaiger: No. I initially had the useEffect in the hook but moved it out to the main page to make it easier to find.

PR #4555, opened 2023-05-11
Waitp 1219 UI charts components package

acdunham:
### Changes:
- Created a standalone package for `ui-chart-components`
- Updated everywhere that used charts from `ui-components` to use new package
- Added chart utils for `utils` package, because these are used across a few packages, to avoid circular dependencies


packages/admin-panel/src/VizBuilderApp/components/PreviewSection.js
```diff
@@ -7,8 +7,8 @@ import { useParams } from 'react-router-dom';
 import styled from 'styled-components';
 import MuiTab from '@material-ui/core/Tab';
 import MuiTabs from '@material-ui/core/Tabs';
-import { Chart, FlexSpaceBetween, FetchLoader, DataTable, JsonEditor } from '@tupaia/ui-components';
-
+import { FlexSpaceBetween, FetchLoader, DataTable, JsonEditor } from '@tupaia/ui-components';
+import { Chart } from '@tupaia/ui-chart-components';
```
tcaiger, line 11: this makes me so happy

packages/ui-chart-components/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Story Book
+
+Storybook is an open source tool for developing UI components.
+[https://github.com/storybookjs/storybook](https://github.com/storybookjs/storybook)
+
+## Material UI
+
+The components are mostly built on top of components from the Material UI library. https://material-ui.com.
+
```
tcaiger, line 23:
```suggestion
## ReCharts

The components are mostly built on top of components from the ReCharts library. https://recharts.org

```

tcaiger, line 23:
```suggestion
## ReCharts

The components are mostly built on top of components from the ReCharts library. https://recharts.org

```

packages/utils/src/period/chartGranularities.js
```diff
@@ -1,15 +1,14 @@
 /**
  * Tupaia Web
```
tcaiger, line 2: Can we please name this file `periodGranularities.js` (since this will end up being the source of truth for all that stuff eventually) and rename the existing `periodGranularities` file to `getDateRangeForGranularity` (since that is all the file exports)

packages/utils/src/chart.js
```diff
@@ -0,0 +1,33 @@
+export const CHART_TYPES = {
+  AREA: 'area',
+  BAR: 'bar',
+  COMPOSED: 'composed',
+  LINE: 'line',
+  PIE: 'pie',
+  GAUGE: 'gauge',
+};
+
+export const getIsChartData = ({ chartType, data }) => {
```
tcaiger, line 10: Can we please move  `CHART_TYPES ` and `getIsChartData ` to the charts package.

Overall the move of the utils from chart to utils looks great. I think these 2 should belong in charts though.

tcaiger (changes requested):
This is looking soo good. I'm really glad we are doing it. I just left a couple of requested changes to move a couple of files around but apart from that it is good to go.

tcaiger (approved):
Ka pai!

PR #4557, opened 2023-05-11
Waitp 1188 multi project landing page styling

acdunham:
### Changes:
- Restricted the height setting of the wrapper to single project view so that multi projects on mobile didn't overlap the footer
- Updated media queries to handle landscape mobile
- Updated names of fields to reflect casing of API return value


tcaiger (approved):
Good pick ups before testing!

PR #4561, opened 2023-05-12
Merge tag 'v2023-19' into lesmis-on-premise

rohan-bes:


IgorNadj (approved):

rohan-bes:
@IgorNadj tests failed due to migrations being old. I don't think there's an easy way to manage this since we haven't merged to the branch in over 4 months (would have to rename a bunch of migrations, and deal with ordering issues). The tests themselves passed so I think it's safe to merge 👍

PR #4562, opened 2023-05-12
WAITP-1261: UI map components package

tcaiger:
### Changes:
- Created a standalone package for ui-map-components
- Updated everywhere that used maps from ui-components to use new package


acdunham (approved):
So good!

PR #4563, opened 2023-05-12
Waitp 1257 UI components ts setup

acdunham:
### Changes:
- Updated the build for `ui-components` to use the root package `ts` build scripts.
- Updated FE packages that watch `ui-components` and `ui-chart-components` to just use the `-w` flag, since the separate build script is unnecessary and added some errors when the build was triggered on file change


tcaiger (approved):
Looks good to me. I tested it locally and it all works well.

PR #4565, opened 2023-05-14
Waitp 1191 image upload caching

acdunham:
### Changes:
- Added timestamp query string to URL of image in `ImageUpload` component to prevent browser caching incorrectly
- Added timestamp query string to URL of image in landing page background and logo images to prevent browser caching incorrectly
Note: this is in response to an issue identified during testing. There is a slight performance impact by doing this but this should be minor.

### DIFF
--- a/packages/ui-components/src/components/Inputs/ImageUploadField.js
+++ b/packages/ui-components/src/components/Inputs/ImageUploadField.js
@@ -126,6 +126,13 @@ export const ImageUploadField = React.memo(
       if (validated) onChange(image);
     };
 
+    const timestamp = Date.now();
+
+    const decachedImage =
+      encodedImage && !encodedImage.includes('base64')
+        ? `${encodedImage}?t=${timestamp}`
+        : encodedImage;
+
     return (
       <Label as="label" htmlFor={name}>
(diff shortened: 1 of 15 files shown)


packages/ui-components/src/components/Inputs/ImageUploadField.js
```diff
@@ -79,7 +79,7 @@ const LabelWrapper = styled(Box)`
 export const ImageUploadField = React.memo(
   ({
     name,
-    encodedImage,
+    imageSrc,
```
tcaiger, line 82: this is a good rename. I was confused when I was reading encodedImage earlier today since it's a url sometimes

packages/utils/src/s3/S3Client.js
```diff
@@ -98,6 +98,25 @@ export class S3Client {
     return this.uploadPrivateFile(s3FilePath, fileStream);
   }
 
+  async deleteFile(filePath) {
```
tcaiger, line 101: Nice - adding this as a util

tcaiger (approved):
Great - this feels much more right to me!

PR #4566, opened 2023-05-15
MDEV-138: Delete Palau's Household Entity

chris-pollard:
### Changes:

- Delete a household record from entity table.


chris-pollard:
Closed, no longer required as migration.

PR #4567, opened 2023-05-15
MDEV-48 pre fill survey

IgorNadj:
### Changes:

My approach with this one was to split importSurveys into two files: one for handling survey changes and one for importing the survey questions file. Doing this lets us pass data to the survey-level one, and pass a file into the survey questions one.

There was some prep work necessary before this, e.g. limiting import to one survey at a time, and by code instead of by name. 

There are a few commits relating to file upload. We submit a request containing both a file and data to the backend. Previously the data was on the query string, I moved this to the request body itself. Doing so required updates to the frontend to form the request, and a new middleware in the backend to parse it, before being handed off to our CRUD handlers in a generic way.

The behaviour of actually creating/editing a survey and the survey questions file should be unchanged, the changes were mostly one level up from that, to the wrapping functions. The actual clean up of the contents of the SurveyEditor and importQuestions files I will leave for a future PR.

#### Screenshots:
https://linear.app/bes/issue/MDEV-48#comment-c47ceb28


packages/central-server/src/apiV2/surveys/README.md
rohan-bes, line 1: Amazing, love the docs ❤️

packages/central-server/src/tests/scaffolding.test.js
```diff
@@ -43,6 +40,6 @@ before(async () => {
 after(async () => {
   const models = getModels();
   SendEmail.sendEmail.restore();
-  await clearTestData(models.database, testStartTime);
+  await clearAllTestData(models.database);
```
rohan-bes, line 43: What happened here? Are there just too many different types of records being created for us to clean up the old way?

This `clearTestData()` function has usually worked surprisingly well, though it ought to be a tad brittle. Wondering about what issues it may have.

rohan-bes (approved):
Awesome stuff @IgorNadj! This just makes so much more sense than the previous system.

The only real question would be about what's the reason for switching from sheets having the survey name to the survey code? Is it just because they are unique?

Pre-approving 👍

packages/central-server/src/apiV2/export/exportSurveys/SurveyExporter.js
```diff
@@ -74,7 +74,7 @@ export class SurveyExporter {
           const metadataRow = await this.buildMetadataRow(currentSurvey);
           if (metadataRow) rowsForExport.push(metadataRow);
 
-          const sheetName = currentSurvey.name.substring(0, 31);
+          const sheetName = currentSurvey.code.substring(0, 31);
```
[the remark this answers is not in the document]
IgorNadj: Ah yep, this makes it consistent with survey responses and lets us uniquely identify surveys when importing. 

I think it was an important commit at a time when I was breaking down the problem into manageable mental chunks. Now that I understand it fully I can see that it's not necessary. Will revert.

packages/central-server/src/tests/scaffolding.test.js
IgorNadj: Ah yep. BTW I started adding second lines to commit messages with notes. This one is "clearnAllTestData simply truncates all tables now that we use tupaia_test instead of the main tupaia db". 

However that doesn't tell the full story... The reason this was necessary here is because the importer creates a bunch of different records across a few different models. The tests would need some heavy mocking to make sure every model creates records with test ids. I think in this case specifically something like survey_screen was being created without a test id.

The alternative of just truncating all tables seemed much simpler, and is equivalent in terms of behaviour, thanks to your recent migration to using a separate db!

packages/central-server/src/apiV2/surveys/README.md
```diff
@@ -0,0 +1,38 @@
...
+The /surveys routes represent a somewhat custom CRUD endpoint on the Survey resource.
+
+The Survey _resource_ model also includes some child _data_ models:
+- survey_screen
+- survey_screen_component
+- question
+
+Currently, properties of the survey data model are managed normally, and the child data models are managed by:
+- GET: property `surveyQuestions` returns a string value e.g. '16 Questions'
+- PUT: property `surveyQuestions` can be a base64 encoded spreadsheet
```
IgorNadj, line 12: Note to self: no longer base64 encoded

packages/central-server/src/apiV2/export/exportSurveys/SurveyExporter.js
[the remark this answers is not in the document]
IgorNadj: Reverted

packages/central-server/src/apiV2/surveys/CreateSurvey.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+export class CreateSurvey extends CreateHandler {
+  permissionsFilteredInternally = true;
+
+  // TODO: merge with SurveyEditor
+  // async assertUserHasAccess() {
+  //   const surveyChecker = accessPolicy =>
+  //     assertSurveyEditPermissions(accessPolicy, this.models, this.recordId);
+  //   await this.assertPermissions(assertAnyPermissions([assertBESAdminAccess, surveyChecker]));
+  // }
```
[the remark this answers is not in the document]
IgorNadj: Tidied up

packages/central-server/src/apiV2/surveys/EditSurvey.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+export class EditSurvey extends EditHandler {
+  permissionsFilteredInternally = true;
+
+  // TODO: merge with SurveyEditor
+  // async assertUserHasAccess() {
+  //   const surveyChecker = accessPolicy =>
+  //     assertSurveyEditPermissions(accessPolicy, this.models, this.recordId);
+  //   await this.assertPermissions(assertAnyPermissions([assertBESAdminAccess, surveyChecker]));
+  // }
```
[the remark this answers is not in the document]
IgorNadj: Tidied up

packages/central-server/src/apiV2/surveys/README.md
IgorNadj: Fixed

IgorNadj:
@rohan-bes a bunch of (very legitimate) issues came out testing. I've pushed those up and while the PR is fundamentally the same, the new commits would benefit from new 👀 , thanks!

rohan-bes (approved):
Looks good!

PR #4572, opened 2023-05-16
RN-917: Fix KoBo sync error when no data is returned

rohan-bes:
### Issue RN-917:

Pretty simple issue in the end. At some point `DataBroker.pull` started returning `undefined` if no data has been found.

No exactly sure why that changed, and arguably it should always return an array, but I don't wanna deal with that now.


chris-pollard (approved):
Nice!

PR #4578, opened 2023-05-18
Waitp 1260 ts UI components1

acdunham:
### Changes:
- Convert all general (non-complex, ie. not in own folder) UI components to Typescript
- Add any additional setup for build
- Get storybook and tests working
Note: this is part 1 or 3 PRs for the same task, so there may be some overlap for these 3 PRs


.eslintrc
```diff
@@ -151,37 +181,62 @@
...
+      "files": [
+        "packages/ui-components/**"
+      ],
       "rules": {
-        "import/no-extraneous-dependencies": "off"
+        "import/no-extraneous-dependencies": "off",
+        "react/prop-types": "off",
+        "import/no-unresolved": 0, // Turn off linting error for imports
+        "import/extensions": 0, // Turn off linting error for imports
+        "react/require-default-props": "off"
```
tcaiger, line 214: I assume these are the only changes and the rest is just auto formatting, right?

packages/ui-components/src/components/ActionsMenu.tsx
```diff
@@ -64,8 +63,14 @@ const StyledMenuIcon = styled(MoreVertIcon)`
...
 
-export const ActionsMenu = ({ options }) => {
-  const [anchorEl, setAnchorEl] = React.useState(null);
+interface Option {
+  label: string;
+  action: () => void;
+  style?: CSSProperties;
+}
+
+export const ActionsMenu = ({ options }: { options: Option[] }) => {
```
tcaiger, line 72: I had been adding `React.FC` but I just looked it up a bit more and I agree that you are right to not include it.

tcaiger: FYI - I've added a section to the code style guide for typescript [here](https://beyond-essential.slab.com/posts/web-frontend-rewrite-03l47wk9#hkjg6-typescript)

packages/ui-components/src/components/Button.tsx
```diff
@@ -208,19 +205,19 @@ export const GreyOutlinedButton = styled(OutlinedButton)`
   }
 `;
 
-const StyledLink = styled(MuiLink)`
+const StyledLink = styled(MuiLink)<OverrideableComponentProps<MuiLinkProps>>`
```
tcaiger, line 208: Does OverrideableComponentProps need to be imported?

packages/ui-components/src/components/Card.tsx
```diff
@@ -25,37 +24,39 @@ const StyledDiv = styled.div`
...
   line-height: 1.125rem;
   font-weight: 400;
   color: ${props => props.theme.palette.text.secondary};
 `;
 
-export const CardHeader = ({ title, label, color }) => (
+interface CardHeaderProps {
+  title: string | ReactNode;
+  label?: string | ReactNode;
+  color?:
```
tcaiger, line 43: Can this be something like TypographyTypeMap.colors from material-ui?

packages/ui-components/src/components/CardTabs.tsx
```diff
@@ -0,0 +1,118 @@
...
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import React, {
+  useState,
+  createContext,
+  useContext,
+  useCallback,
+  ReactNode,
+  ElementType,
```
tcaiger, line 12:
```suggestion

```

packages/ui-components/src/components/Dialog.tsx
```diff
@@ -59,7 +58,26 @@ const CloseButton = styled(IconButton)`
   color: ${props => props.theme.palette.text.primary};
 `;
 
-export const DialogHeader = ({ title, onClose, color, children }) => (
+interface DialogHeaderProps {
+  title: string;
+  onClose: () => void;
+  color?:
```
tcaiger, line 64: Hopefully this can be imported from material-ui TypographyTypeMap? Otherwise we should create a shared type for it somewhere.

.eslintrc
acdunham: I believe so, yes!

packages/ui-components/src/components/FavouriteButton.tsx
```diff
@@ -7,9 +7,19 @@ import FavoriteBorderIcon from '@material-ui/icons/FavoriteBorder';
 import FavoriteIcon from '@material-ui/icons/Favorite';
 import IconButton from '@material-ui/core/IconButton';
 
-import PropTypes from 'prop-types';
+interface FavouriteButtonProps {
+  isFavourite?: boolean;
+  onChange: () => void;
+  color?: 'primary' | 'secondary' | 'inherit' | 'default';
```
tcaiger, line 13: Can these be pulled from material ui? It seems so since they are implicitly used in the IconButton.tsx component below this?

packages/ui-components/src/components/Button.tsx
acdunham: So it is, but this is one of those ones where it might have been lost in translation with my PR splitting - potentially is done elsewhere/will be a part of the overall branch that we can do a PR for if not?

packages/ui-components/src/components/Tabs.tsx
```diff
@@ -17,10 +17,10 @@ export const StyledTabs = styled(MuiTabs)`
   }
 `;
 
-export const Tabs = props => {
+export const Tabs = (props: TabsProps) => {
   const [value, setValue] = useState(0);
   const handleChange = useCallback(
-    (event, newValue) => {
+    (event: any, newValue: any) => {
```
tcaiger, line 23: Any reason that you didn't use React.MouseEvent?

packages/ui-components/src/components/Toolbar.tsx
```diff
@@ -22,21 +21,17 @@ const ToolbarWrapper = styled.div`
   }
 `;
 
-export const BaseToolbar = ({ children, maxWidth }) => (
+interface BaseToolbarProps {
+  maxWidth?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | false | undefined;
```
tcaiger, line 25: can we use ContainerTypeMap.maxWidth? and maybe the default should be false instead of undefined?

packages/ui-components/tsconfig-build.json
```diff
@@ -3,5 +3,12 @@
...
-  }
+  },
+  "include": [
+    "src/**/*"
+  ],
+  "exclude": [
+    "src/__tests__",
+    "src/stories/**/*"
+  ]
 }
```
tcaiger, line 14: complaining about needing an extra new line

packages/ui-components/tsconfig.json
```diff
@@ -6,12 +6,17 @@
...
   "include": [
-    "src/**/*"
-  ],
-  "exclude": [
-    "src/__tests__",
-    "src/stories/**/*"
+    "src/**/*",
+    "jest.config.ts",
+    "helpers/**/*"
   ]
```
tcaiger, line 21: here too

tcaiger (changes requested):
Looks really good. I just left a few suggestions to see if we can use some more of the material-ui props. If they don't work for some reason than just reply and let me know and don't worry about it.

packages/ui-components/src/components/Button.tsx
tcaiger: Yes i can see it in the other pr

tcaiger (approved):
🥳

PR #4579, opened 2023-05-18
Waitp 1260 ts UI components3

acdunham:
### Changes:
- Convert Icons and Inputs components to Typescript
- Add any additional setup for build
- Get storybook and tests working
Note: this is part 3 of 3 PRs for the same task, so there may be some overlap for these 3 PRs


packages/ui-components/src/components/Inputs/AsyncAutocomplete.tsx
```diff
@@ -3,17 +3,19 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
-import React, { useState, useEffect, useCallback } from 'react';
+import React, { useState, useEffect } from 'react';
 import throttle from 'lodash.throttle';
```
tcaiger, line 7: this might be another one that only happens for me but the linter is telling me to install @types/lodash.throttle

tcaiger (approved):
Just double check that lodash error but otherwise it all looks good to me.

PR #4580, opened 2023-05-19
Waitp 1189 landing page name length

acdunham:
### Changes:
- Add validation to make sure landing page name length is no more than 40 characters
- Update styling to allow for 40 character landing page name lengths in top bar

### Screenshots:
With shorter name at mobile size:
Screenshot 2023-05-19 at 12 37 28 pm

With long name at mobile size: text wraps and top bar expands (not much way around making the bar expand, because 40 characters is long enough to require it)
Screenshot 2023-05-19 at 12 38 33 pm


packages/web-frontend/src/containers/TopBarLogo/index.js
```diff
@@ -29,8 +29,13 @@ const LogoWrapper = styled.div`
 `;
 
 const LogoImage = styled.img`
-  max-height: 80%;
+  max-height: 40px; // 80% of top bar height
```
tcaiger, line 32: I'm not too concerned but could we use calc based on the top bar height variable for these heights?

tcaiger (approved):
😄

PR #4583, opened 2023-05-19
RN-922: Fix PNG exports failing on LESMIS

rohan-bes:
### Issue RN-922:


packages/lesmis/src/utils/export/useExportToImage.js
```diff
@@ -24,12 +24,12 @@ export const useExportToImage = (filename, format) => {
   const [isExporting, setIsExporting] = useState(false);
   const [isExportLoading, setIsExportLoading] = useState(false);
   const sanitisedFileName = toFilename(filename, true);
-  const node = exportRef.current;
 
   const exportToImg = async () => {
     setIsExporting(true);
     setIsExportLoading(true);
 
+    const node = exportRef.current;
```
rohan-bes, line 32: Pretty simple issue in the end. We set the ref on the component correctly, but since we were taking it's value outside of the function scope, it was always set to null.

IgorNadj (approved):

PR #4585, opened 2023-05-21
Move utcMoment and PERIOD_TYPES to tsutils

kael89:
1. Change imports of `utcMoment` to`@tupaia/tsutils` instead of `@tupaia/utils`. The util had been copied there in a previous PR
2. Moved `PERIOD_TYPES` to `@tuapia/tsutils` and changed imports to point there

I won't delete their definitions in `@tupaia/utils` yet, since other modules in that package depend on them

### DIFF
--- a/packages/lesmis/src/api/queries/useVitalsData.js
+++ b/packages/lesmis/src/api/queries/useVitalsData.js
@@ -4,7 +4,7 @@
  *
  */
 import { useQuery } from 'react-query';
-import { utcMoment } from '@tupaia/utils';
+import { utcMoment } from '@tupaia/tsutils';
 import { post } from '../api';
 import { useProjectEntitiesData } from './useEntitiesData';
 import { useEntityData } from './useEntityData';
--- a/packages/lesmis/src/api/queries/useVitalsData.js
+++ b/packages/lesmis/src/api/queries/useVitalsData.js
@@ -4,7 +4,7 @@
  *
  */
 import { useQuery } from 'react-query';
-import { utcMoment } from '@tupaia/tsutils';
+import { utcMoment } from '@tupaia/utils';
 import { post } from '../api';
 import { useProjectEntitiesData } from './useEntitiesData';
 import { useEntityData } from './useEntityData';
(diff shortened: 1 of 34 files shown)


packages/tsutils/src/period/period.ts
```diff
@@ -0,0 +1,21 @@
...
+const DAY = 'DAY';
+const WEEK = 'WEEK';
+const MONTH = 'MONTH';
+const YEAR = 'YEAR';
+const QUARTER = 'QUARTER';
+
+/**
+ * Available period formats for aggregation server data
+ */
+export const PERIOD_TYPES = {
```
IgorNadj, line 15: This should be enum?

IgorNadj (approved):
LGTM, one question

packages/tsutils/src/period/period.ts
kael89: Good point! I left this as is to make the migration easier and a bit safer.

If you don't mind, I can change it to an `enum` in a separate PR, after all the depending utils have moved under `tsutils` too

kael89:
@IgorNadj any idea why the build failed? :thinking:

rohan-bes (approved):
Nice one, thanks @kael89

kael89:
@IgorNadj @rohan-bes builds pass now, I'll leave merging to you

kael89:
> > @IgorNadj any idea why the build failed? 🤔
> 
> ```
> ➤ YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.
> ```

I forgot to run `yarn`

packages/lesmis/src/api/queries/useVitalsData.js
```diff
@@ -4,7 +4,7 @@
  *
  */
 import { useQuery } from 'react-query';
-import { utcMoment } from '@tupaia/tsutils';
+import { utcMoment } from '@tupaia/utils';
```
kael89: Using `tsutils` in `lesmis` caused builds to fail.

I had a look and `tsutils` is not used in any frontend package. This does ring a bell about not being able to use  any tupaia TS package in the frontend currently.

This can be investigated and fixed at a later stage, skipping for now

PR #4587, opened 2023-05-22
MDEV-151: Add a facility to the UNFPA hierarchy

chris-pollard:
### Changes:

- Add a new record in the entity_hierarchy table with 'Saladak Dispensary' as child of Pohnpei District in UNFPA project.


chris-pollard:
Avoiding these types of migrations, chosen Over The Shoulder instead.

PR #4589, opened 2023-05-23
RN-835 Enforce up to date types

IgorNadj:
### Changes:

- Runs against the test db. If a migration is added and types are not re-generated the build fails.

Failing [build](https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/2fddae2a-2b61-4a4d-8d9b-f65cb34e9af2?component=step_Validate_and_test_Db_Validation_Validate_types) example:
Screen Shot 2023-05-25 at 3 58 00 pm

Passing [build](https://app.codeship.com/projects/70159bc0-0dac-0138-fdcb-260b82737f4e/builds/2935a99d-74c0-4320-97d3-28330bb6c7d0?component=step_Validate_and_test_Validate_Validate_branch_name) example:
Screen Shot 2023-05-25 at 4 46 46 pm


packages/devops/ci/testing.env.encrypted
```diff
@@ -1,2 +1,2 @@
...
\ No newline at end of file
+xnpR6UV/QIrB5eKb7b0329V76cFEnWFSc8M9eEOPaYwdhW1TF20VKtu58Kj1LMzuaXGuZDr/jjIRBcLyfwkmS9kZSPC8+bEvf3aqrW1PSYLmqZI8jYSgxrcsvP/6/YoO9JEsxg0pu/WqkiXasPuY7DFXtVU3E9aBH4iFHpMV4FUIq2wEaQdJDsOkqIkpP2jSsgaWn6iKd7kViVAhGM930pNECmSeStYJ20TJS/7KPumHO5Jg5KbWIMtRWonPAMclI2/Z001cy0uIvy5zHRQ6gpS9eVtsv4m7mVwWVauBF5ZAPezNh7lzpLM/UxGNakO10z0pr7BZ6s6Bg8a7+mHvAv17QZgGOwaPmEZ92Y0zMALdwNYckJ9qnRaI0inOszWK6h1tIrIDvZM2kR+S63qO25bZgn4jajXmFeTJtHkPda9H2eh7Shd8wkatkfaQD+/Bngf3DYxfkUENC1z+xvrsss/gqEPD+lZpUhLUa9PbqqDuMPLxp2k/9DOn+qYXQVZB7bAZVpd+Kv8iyggyLl6uYln6iRJJJEIsHsS41ReHrYnOBDg3DXEAPHBGpnP8/gljtMkBA/H2aF3JZdItPU3auJAdDO4r0bTPK3iHNXftgsEew3I2+lNGKSKC65rfQ36yatXLoR06y+hz9/+HkZETn9L+oQoLTDXb91a/4dn5stRFYuhTNdJ6FpvhRGLnpVHEBOtWhdNjNnPYcs1UEUWeOjPkfthmV2O+NeLN+UUObA0D3qKNs3A/4M3DEwHOAWe5+0xC/jhCgGNBHUwOfmXxM07de5wYty2ieHkzHsZ6kZ9IEkT43Pck+hh7wJ5lLc3xP5L235Tx1pPBDN3uSESBH7U=
```
IgorNadj, line 2: Changed db name from `tupaia` to `tupaia_test`. Not sure how this was working before... e.g. entity-server's tests explicitly run against a database named `tupaia_test`.

packages/types/package.json
```diff
@@ -22,7 +22,8 @@
     "test": "echo \"No tests specified\" && exit 0",
     "generate": "yarn generate:models && yarn generate:schemas && yarn build",
     "generate:models": "ts-node generate-models.ts",
-    "generate:schemas": "ts-node generate-schemas.ts"
+    "generate:schemas": "ts-node generate-schemas.ts",
+    "assert-no-changes": "yarn generate:models --failOnChanges"
```
rohan-bes, line 26: Is there a reason we're only running this assertion against the models? Might be good to check if someone has changed any of the schema types and forgotten to update the schemas too

rohan-bes (approved):
Great stuff, thanks @IgorNadj.

Requesting changes to also validate if the schemas have changed, but pre-approving 👍

packages/types/package.json
IgorNadj: Yeah good point, I did consider this and I'm not 100% sure it's needed. Schemas are auto generated from models, so we would catch someone who ran generate:schemas instead of just generate, but that seems a little unlikely. Also the way I was trying to get it going, by doing a git diff was driving me nuts with how difficult it is to develop codeship steps locally 😢 Are you happy for us to go with this for now which should catch almost all cases?

packages/devops/ci/testing.env.encrypted
IgorNadj: Oh yeah I mean I'm not sure how it was creating the db with name tupaia_test before.

https://github.com/beyondessential/tupaia/blob/dev/packages/database/scripts/setupTestDatabase.sh#L16
Reading the script it should have been creating a db called tupaia not tupaia_test. Oh well, it's not important, just a curiosity.

packages/types/package.json
rohan-bes: > Schemas are auto generated from models

Schemas also come from raw types right? Eg. `@tupaia/types/src/types/request/**` So it's possible someone updated those but never re-ran `generate:schemas`

Instead of a git diff, is there a reason we can't use a similar `if (failOnChanges) read existing file and compare to new text` kind of logic?

All good if I'm missing something here, happy to just have the models validation for now and we can add schema validation if/when it becomes an issue and after we've switched to GHA 👍

IgorNadj: Yes correct. I was being lazy, added the check on schemas as well.

PR #4593, opened 2023-05-25
WAITP-1258: ui-chart-components ts filename and config

tcaiger:
### Changes:

- Renaming the files to make the upcoming PRs easier to read
- TS config setup


IgorNadj (approved):
Nice one!

PR #4598, opened 2023-05-28
[no-issue]: Move period type utils to tsutils

kael89:
Moving a few more of those across

### DIFF
--- a/packages/tsutils/src/period/period.ts
+++ b/packages/tsutils/src/period/period.ts
@@ -3,6 +3,20 @@
  * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
+import { reduceToDictionary } from '@tupaia/utils';
+import get from 'lodash.get';
+
+type PeriodType = keyof typeof PERIOD_TYPES;
+
+interface PeriodTypeConfig {
+  format: string;
+  length: number;
+  granularityOrder: number;
+  displayFormat: string;
+  momentShorthand: string;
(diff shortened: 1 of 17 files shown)


packages/tsutils/src/period/period.ts
```diff
@@ -19,3 +34,111 @@ export const PERIOD_TYPES = {
   QUARTER, // e.g. '2018Q1'
   YEAR, // e.g. '2018'
 } as const;
+
+const NON_NUMERIC_PERIOD_TYPES = [WEEK, QUARTER];
```
kael89, line 38: Everything here (except for types) is copied over from `period.js` in `@tupaia/utils`

kael89:
@rohan-bes now that you are in the review flow, you may also want to have a look at this one too 🙃 

All good if you are already aware of it - it's not urgent! Just a bump in case it got lost in a pile of emails and notifications 😄

rohan-bes (approved):
Awesome stuff, thanks @kael89 💜 

Just a heads up, that we're planning on converting `@tupaia/utils` into a TS/JS hybrid package soon. At that point we're gonna combine `@tupaia/utils` and `@tupaia/tsutils`. It should make these kinds of changes heaps easier, since you don't need to update imports, just convert the file to TS 😁

kael89:
> 

Ah amazing, that makes a lot of sense! I'll hold off on further conversions for now then.

Happy to help convert the rest of utils to TS after that happens 🙂

PR #4600, opened 2023-05-29
Waitp 1259 map components to ts

acdunham:
### Changes:
- Converted all components and utils in `ui-map-components` to Typescript
- Updated associated linting rules
- Updated associated scripts


packages/ui-components/src/components/ReferenceTooltip.tsx
```diff
@@ -46,13 +46,13 @@ const StyledToolTip = withStyles(theme => ({
   },
 }))(Tooltip);
 
-interface Reference {
+export interface ReferenceProps {
   text?: string;
```
acdunham, line 50: I renamed and exposed this for use in `ui-map-components`

packages/web-frontend/package.json
```diff
@@ -22,14 +22,14 @@
     "start-central-server": "yarn workspace @tupaia/central-server start-dev",
     "start-entity-server": "yarn workspace @tupaia/entity-server start-dev",
     "start-report-server": "yarn workspace @tupaia/report-server start-dev",
-    "start-web-config-server": "yarn workspace @tupaia/web-config-server start-dev",
-    "start-servers": "npm-run-all -c -l -p start-central-server start-entity-server start-report-server start-web-config-server",
+    "start-web-config-server": "yarn workspace @tupaia/web-config-server start-dev", 
+    "start-servers": "npm-run-all -c -l -p start-central-server start-entity-server start-report-server start-web-config-server start-data-table-server",
     "start-data-table-server": "yarn workspace @tupaia/data-table-server start-dev",
```
acdunham, line 27: Added this in so the visualisations using this worked locally for testing

packages/ui-map-components/src/components/Markers/disasterIcons/Cyclone.tsx
```diff
@@ -2,9 +2,9 @@
  * Tupaia
  *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
-import React from 'react';
```
tcaiger, line 5: We are not supporting the Disasters feature moving forward and that is the only use case for the disaster icons so let's remove all of them and any references to them in this package. thanks

packages/ui-map-components/src/types/leaflet-config.d.ts
```diff
@@ -0,0 +1,19 @@
+/*
```
tcaiger, line 1: This is nice - seems like the correct way to set out the custom types

tcaiger (changes requested):
Looks really good. I can't find anything that I would improve. It would be great to remove those disaster icons and references to them but apart from that it's good to go.

tcaiger (approved):

PR #4602, opened 2023-05-29
WAITP-1255: Tupaia web server report and dashboard routes

EMcQ-BES:
### Changes:

- Adding `reports` route to `tupaia-web-server`
  - Basic forwarder route
- Adding `dashboards` route
  - Fetch dashboards and items based on the entity hierarchy of the project
- Adding `TempLogout` route
  - Currently the frontend uses a `GET` for logout which doesn't play nice with the `server-boilerplate` logout functionality

### DIFF
--- a/packages/tupaia-web-server/src/routes/ReportRoute.ts
+++ b/packages/tupaia-web-server/src/routes/ReportRoute.ts
@@ -1,0 +1,30 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ *
+ */
+
+import { Request } from 'express';
+import { Route } from '@tupaia/server-boilerplate';
+
+export type ReportRequest = Request<
+  { reportCode: string },
+  any,
+  any,
+  any
+>;
+
+export class ReportRoute extends Route<ReportRequest> {
+  public async buildResponse() {
+    const { query, ctx } = this.req;
+    const { reportCode } = this.req.params;
+    const { legacy } = query;
+
+    // Legacy data builders are handled through the web config server still
+    if (legacy === 'true') {
+      return ctx.services.webConfig.fetchReport(reportCode, query);
+    }
+
+    return ctx.services.report.fetchReport(reportCode, query);
+  }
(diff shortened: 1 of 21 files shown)


packages/tupaia-web-server/src/routes/ReportRoute.ts
```diff
@@ -0,0 +1,30 @@
...
+  any,
+  any,
+  any
+>;
+
+export class ReportRoute extends Route<ReportRequest> {
+  public async buildResponse() {
+    const { query, ctx } = this.req;
+    const { reportCode } = this.req.params;
+    const { legacy } = query;
```
rohan-bes, line 21: Out of curiosity, are we initially leaving the `tupaia-web <-> tupaia-web-server` interface as is? Then perhaps reworking at a later point?

It always struck me as a little odd that the front-end told the backend if the report were legacy or not...

rohan-bes (approved):
Looks good! Nice work @EMcQ-BES 

Just a couple of cleanup suggestions but pre-approving

packages/tupaia-web-server/src/routes/ReportRoute.ts
EMcQ-BES: Currently I'm just making it backwards compatible, because that's the easiest way to test backend before we've updated the frontend. We can definitely make adjustments to the API during this project, but I reckon better to strive for a "cleaned up" version of the current API rather than redesigning too much.

It'd be a little annoying to work out if the report is legacy in `tupaia-web-server`, since as an orchestration server it doesn't have direct DB access, so would need to do a couple of api calls to figure it out. And I think there's not currently a rule that says that a `legacy_report` and a `report` can't have the same `code`. So we can assume but not guarantee (or more realistically we'd just add that rule)

Splitting the difference, since the frontend already knows whether or not it's legacy maybe we could split into separate routes `/api/report/:code` and `/api/legacyReport/:code`

rohan-bes: I like having 2 routes. 👍

PR #4604, opened 2023-05-30
WAITP-1266 Retain dates when exporting to excel

acdunham:
### Changes:
- Updated `useDataTableExport` to handle string types (where `isNaN` returns true) by returning values as they are. This will include percentages and date strings.
- Otherwise return booleans as is, then turn anything else into a number


packages/ui-components/src/components/DataTable/useDataTableExport.js
```diff
@@ -28,9 +28,18 @@ export const useDataTableExport = (columns, data, title, startDate, endDate) =>
...
-              const num = valueIsPercentageString ? value : parseFloat(value);
-              // if it's a number, return in number format
+
+              // Attempt to convert to a moment object if is a string that can't be converted to a number (isNaN). If it's a date string, return in date format, otherwise continue
+              // This check will handle percentage strings as well as date strings, and any other non-numeric strings
+              if (typeof value === 'string' && isNaN(value)) {
+                const momentObj = displayStringToMoment(value);
+                // if it's a date string, return in date format
+                if (momentObj.isValid()) return momentObj.format('DD/MM/YYYY');
+                return value;
```
EMcQ-BES, line 38: I think this will break regular number fields

`displayStringToMoment` attempts to parse the string through a number of standard formats, but that includes things like `YYYY`, so if I enter _any_ four digit number it'll get parsed and then turned into `01/01/1234`?

(I don't recall if `YYYY` will successfully parse shorter numbers, but I think it does. So the number `1` get turned into `01/01/0001`)

acdunham: Ordinarily yes it probably would break it, but because it's inside this check: `if (typeof value === 'string' && isNaN(value))` normal numbers will never get there. It can only be strings that aren't immediately recognisable as a number, e.g. `11/2/10`

EMcQ-BES: Oh yup! My brain glossed over the `isNaN`, that makes sense

EMcQ-BES (approved):
LGTM

acdunham:
@EMcQ-BES or @tcaiger, would you be able rereview this? I added a check for format to fix comments from Andrew, so that MMM YYYY dates remain handled as is.

packages/ui-components/src/components/DataTable/useDataTableExport.js
```diff
@@ -28,9 +32,21 @@ export const useDataTableExport = (columns, data, title, startDate, endDate) =>
         ? tableData.map(row =>
             tableColumns.map(col => {
               const value = row.values[col.id];
-              const valueIsPercentageString = value.includes('%');
-              const num = valueIsPercentageString ? value : parseFloat(value);
```
EMcQ-BES, line 32: Going back in time a little. What happens if we just swap this `parseFloat(value)` to `Number(value)`?

acdunham: So I think if you do `Number(true)` then it becomes `1` but if you do `parseFloat(true)` then it becomes `NaN`, which is what we want I think...

EMcQ-BES: That's true. We'd want to avoid specifically cases where the value is a boolean or an empty string. But it'd reject percentage strings as well as almost all our date formats (except `YYYY`, which I think we can just let parse to number). Am I missing any other edge cases?

```
const num = (value && typeof value !== 'boolean') ? Number(value) : value;
```

acdunham: So I have a check for `typeof value === 'string' and isNaN(value)` which will catch any strings that aren't considered stringified numbers, e.g. percentages and date strings. And then it handles dates within that, but converting to a moment object, and if it is a date that is valid and is a 'D*' format (ie.g. anything with a day in it) then it converts to 'DD/MM/YYYY'.
Any other values (e.g. booleans, strings that are stringified numbers, and numbers) go to the parseFloat. So yeah, we could do that as well at this same stage. But in order to format dates appropriately then we need to do the rest of the steps above as well, unless Regina/Georgia make the decision that not all dates need to be converted, and that as long as they don't parse to numbers then that's fine (I have asked on the ticket)

EMcQ-BES: Yeah, looking at the code you've written it looks like it'd work. It just feels like a bit of code smell.

Personally I'd push back that we shouldn't be reformatting the dates at all. It's a raw data export. It should be the raw data formatting.

acdunham: Yeah makes sense - let me wait on the answer on whether we can keep the formatting as is and go from there

EMcQ-BES (approved):
I think that's good. Can't think of any edge cases. Let's see if our testers can find any 😅

PR #4606, opened 2023-05-30
WAITP-1199 Setup API abstraction

acdunham:
### Changes:
- Setup basic API files for `tupaia-web`
- Install `axios`

Note: No actual API calls in there yet, these will come when we do the associated functionality I believe

### DIFF
--- a/packages/tupaia-web/src/api/api.ts
+++ b/packages/tupaia-web/src/api/api.ts
@@ -31,6 +31,7 @@ const getRequestOptions = (options: RequestParametersWithMethod) => {
   };
 };
 
+// Todo: Move api request util to ui-components and allow for mapping to backend request type safety
 const request = async (endpoint: string, options: RequestParametersWithMethod) => {
   const requestOptions = getRequestOptions(options);
 
--- a/packages/tupaia-web/src/api/api.ts
+++ b/packages/tupaia-web/src/api/api.ts
@@ -1,0 +1,71 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ *
+ */
+import axios from 'axios';
+import FetchError from './fetchError';
+import { getApiUrl } from '../utils';
+
+// withCredentials needs to be set for cookies to save @see https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials
+axios.defaults.withCredentials = true;
+
+const timeout = 45 * 1000; // 45 seconds
+
+type RequestParameters = Record<string, any> & {
+  params?: Record<string, any>;
+};
+
+type RequestParametersWithMethod = RequestParameters & {
+  method: 'get' | 'post' | 'put' | 'delete';
+};
+const getRequestOptions = (options: RequestParametersWithMethod) => {
+  const locale = window.location.pathname.split('/')[1];
+  return {
+    ...options,
+    timeout,
+    params: {
+      ...options.params,
+      locale,
+    },
+  };
+};
+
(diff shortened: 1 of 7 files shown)


packages/tupaia-web/src/utils/getApiUrl.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+export const getApiUrl = () => {
+  return process.env.REACT_APP_CONFIG_SERVER_BASE_URL || 'http://localhost:8080/api/v1/';
```
rohan-bes, line 7:
```suggestion
  return process.env.REACT_APP_TUPAIA_WEB_API_URL || 'http://localhost:8100/api/v1/';
```
I think that's the standard .env var pattern now 👍

packages/tupaia-web/src/api/api.ts
rohan-bes, line 1: Sorry if this holds things up a bit, but @tcaiger and I were discussing if we'd like to abstract out this interface. We use it in every front-end (except for admin-panel I think) and it'd be nice to make things DRY-er.

Also, I see a neat potential for converting the `request` function into a generic function which could allow for some really nice mapping between backend and frontend type interfaces.

@acdunham would it be okay if you @tcaiger and myself had a chat on this one later this week?

tcaiger: ps. I'd be keen to get @EMcQ-BES in on this too since it is the. design of the interface between the front end and backend types.

packages/tupaia-web/src/api/mutations/index.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ *
+ */
+
+export * from './useLogin';
+export * from './useLogout';
+export * from './useRegisterUser';
+export * from './useUpdateFavouriteDashboardItem';
```
tcaiger, line 10:
```suggestion

```
Let's add these as reqiured

packages/tupaia-web/src/utils/getApiUrl.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+export const getApiUrl = () => {
```
tcaiger, line 6: I think that in lesmis there was a bit more going on to get the api url but here it is very simple and also probably only ever used in the api function so can we please move this function in to the api.ts file.

packages/tupaia-web/src/api/api.ts
```diff
@@ -0,0 +1,71 @@
...
+    ...options,
+    timeout,
+    params: {
+      ...options.params,
+      locale,
+    },
+  };
+};
+
+const request = async (endpoint: string, options: RequestParametersWithMethod) => {
```
tcaiger, line 34:
```suggestion
// Todo: Move api request util to ui-components and allow for mapping to backend request type safety
const request = async (endpoint: string, options: RequestParametersWithMethod) => {
```

Can we add a comment like the above so that it's obvious that this is still a work in progress. thanks

tcaiger (changes requested):
Looks good. Just left a few small suggestions relating to these changes and then we will pick up the extra work that Rohan suggested in a new pr later.

packages/tupaia-web/src/api/api.ts
rohan-bes: Thanks guys 👍 Set up a meeting for this next week when we'll all be in Waiheke

PR #4607, opened 2023-05-30
RN-929 File Question

IgorNadj:


packages/meditrak-app/app/assessment/specificQuestions/FileQuestion.js
```diff
@@ -0,0 +1,114 @@
...
+    marginHorizontal: BUTTON_SPACING,
+  },
+});
+
+export const FileQuestion = ({ answer, extraProps, onChangeAnswer }) => {
+  const filename = answer ? answer.substring(answer.lastIndexOf('/') + 1) : null;
+
+  const handleOpenFilePicker = async () => {
+    try {
+      // We copy the file to document storage (NSDocumentDirectory/getFilesDir on ios/android), which is our application sandboxed file storage.
```
chris-pollard, line 65: Interesting. So file must be copied in case the original file is not accessible at time of sync.

packages/meditrak-app/app/assessment/specificQuestions/FileQuestion.js
```diff
@@ -0,0 +1,114 @@
...
+      // The file may not be available at sync time otherwise.
+      // This can fail if there is no disk space.
+      // Additionally, a large backlog of files to be synced up will take up a lot of disk space on the device.
+      // Finally, we must delete files after we sync them up, or the app will just take up disk space unnecessarily.
+      // See https://github.com/rnmods/react-native-document-picker/tree/v8.2.1
+      const filePickerResponse = await DocumentPicker.pickSingle({
+        copyTo: 'documentDirectory'
+      });
+
+      // If we previously picked a file (stored a copy in app documents) and now picked a new one, we need to delete the old copy
```
chris-pollard, line 75: So the copy happens during the process, not after submission. Therefore, we need to keep the app documents storage clean as we go? e.g. if someone chooses, scraps that, then chooses another.

packages/meditrak-app/app/assessment/specificQuestions/FileQuestion.js
```diff
@@ -0,0 +1,114 @@
...
+      if (answer) {
+        try {
+          await RNFS.unlink(answer);
+        } catch (e) {
+          console.warn(`Failed to unlink previously picked file: ${answer}`);
```
chris-pollard, line 80: Is this a warning for the client log, or the user? If for user, maybe could be a bit clearer in plain terms?

chris-pollard (approved):
Looks good to me!

packages/meditrak-app/app/assessment/specificQuestions/FileQuestion.js
IgorNadj: Yep that's right.

packages/meditrak-app/app/assessment/specificQuestions/FileQuestion.js
IgorNadj: Yeah reading through how files are handled on ios/android you can pick from many places, e.g. google drive. So copying means we download the file to documents for working with.

packages/meditrak-app/app/assessment/specificQuestions/FileQuestion.js
IgorNadj: Currently this doesn't go anywhere, just the terminal while you're developing. I did consider sending this up to bugsnag (our error logging service), but I couldn't see any way to do that easily so I left it. It should be very unlikely anyway.

PR #4608, opened 2023-06-01
Waitp 1202 setup routes

acdunham:
### Changes:
- Setup basic routing for `tupaia-web`, for user pages, project, and landing page, including handling the cases of similar routes (projects and landing pages)
- Also include redirect to default project on first landing of application if none is provided

Note: This is just basic logic, there will need to be more logic added in around modals and so on when these are implemented.


packages/tupaia-web/src/Router.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { Navigate, createBrowserRouter } from 'react-router-dom';
+import { LoginForm, PasswordResetForm, RequestAccessForm, Base } from './pages';
+import { RegisterForm } from './pages/RegisterForm';
+import { DEFAULT_URL } from './constants';
+
+const Routes = [
+  {
+    path: '/',
+    element: <Navigate to={`/${DEFAULT_URL}`} replace />, // This is to redirect from no URL to the default project URL. Some logic will also have to be added in here to render the projects modal in this case as well
```
tcaiger, line 14: great comments

packages/tupaia-web/src/Router.tsx
```diff
@@ -0,0 +1,40 @@
...
+  },
+  {
+    // This is the base view for anything with a code in the URL. This will render a landing page if the URL has only one segment, and if it has more than 1 segment it will render the MapView
+    path: '/:code/:entityCode?/*',
+    element: <Base />,
+  },
+  // The below user pages will actually be modals, which will be done when each view is created. There is an example at: https://github.com/remix-run/react-router/tree/dev/examples/modal
+  {
+    path: '/register',
```
tcaiger, line 23: I reckon we should put these specific hard coded urls above the Base url as they should always exist. We should also add this list of urls to a black list of codes that can't be used for project landing pages. Would you mind updating the constructNewRecordValidationRules.js with the extra validation rule.

packages/tupaia-web/src/Router.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { DEFAULT_URL } from './constants';
+
+const Routes = [
+  {
+    path: '/',
+    element: <Navigate to={`/${DEFAULT_URL}`} replace />, // This is to redirect from no URL to the default project URL. Some logic will also have to be added in here to render the projects modal in this case as well
+  },
+  {
+    // This is the base view for anything with a code in the URL. This will render a landing page if the URL has only one segment, and if it has more than 1 segment it will render the MapView
+    path: '/:code/:entityCode?/*',
```
tcaiger, line 18:
```suggestion
    path: '/:code/:entityCode?/:dashboardCode?',
```
Any reason not to include the dashboard code param?

packages/tupaia-web/src/pages/Base.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const Base = () => {
+  const { pathname } = useLocation();
+
+  // This is to handle trailing '/' in a pathname
+  const finalURLChar = pathname[pathname.length - 1];
+  const finalCharSubtractor = finalURLChar === '/' ? 1 : 0;
+
+  // If a pathname has only 1 segment, treat as a landing page, otherwise treat as a project
+  const isLandingPage = pathname.split('/').length - 1 - finalCharSubtractor === 1;
```
tcaiger, line 17:
```suggestion
  const { entityCode } = useParams();

  const isLandingPage = !entityCode;
```
What about just using the named param to check if it's a landing page?

packages/tupaia-web/src/Router.tsx
```diff
@@ -0,0 +1,40 @@
...
+    element: <RegisterForm />,
+  },
+  {
+    path: '/login',
+    element: <LoginForm />,
+  },
+  {
+    path: '/reset-password',
+    element: <PasswordResetForm />,
+  },
```
tcaiger, line 33: Can we also add one for verify-emai

packages/tupaia-web/src/pages/Base.tsx
tcaiger: Actually I think we can do away with Base altogether and just have this check in the Router

packages/tupaia-web/src/Router.tsx
```diff
@@ -0,0 +1,40 @@
...
+const Routes = [
+  {
+    path: '/',
+    element: <Navigate to={`/${DEFAULT_URL}`} replace />, // This is to redirect from no URL to the default project URL. Some logic will also have to be added in here to render the projects modal in this case as well
+  },
+  {
+    // This is the base view for anything with a code in the URL. This will render a landing page if the URL has only one segment, and if it has more than 1 segment it will render the MapView
+    path: '/:code/:entityCode?/*',
+    element: <Base />,
+  },
```
tcaiger, line 20:
```suggestion
    {
    // Render a landing page if the URL has only one segment
    path: '/:code',
    element: <LandingPage />,
  },
  {
    path: '/:code/:entityCode/:dashboardCode?',
    element: <MainPage />,
  },
```
What do you think about setting the route like this instead of the base route to make it more transparent. And any reason not to include the dashboardCode param?

Also I'm open to suggestions on renaming the `pages` folder to views or routes if you think that makes more sense?

tcaiger (changes requested):
Great work, it's so exciting to be using real routes! I left a couple of suggestions, let me know what you think. thanks

packages/tupaia-web/src/Router.tsx
```diff
@@ -0,0 +1,41 @@
...
+  PasswordResetForm,
+  Project,
+  RegisterForm,
+  RequestAccessForm,
+  VerifyEmailForm,
+} from './pages';
+import { DEFAULT_URL } from './constants';
+
+/**
+ * This Router is using [version 6.3]{@link https://reactrouter.com/en/v6.3.0}, as later versions are not supported by our TS setup. See [this issue here]{@link https://github.com/remix-run/react-router/discussions/8364}
```
tcaiger, line 19: thanks for the comment

tcaiger (approved):
Looks good - great work!

PR #4609, opened 2023-06-01
MDEV-6 Add defaulting behaviour to newly created data elements

EMcQ-BES:
### Changes:

- When importing a survey creates a data element, give that newly created data element the survey's permission group as its default permission group


tcaiger (approved):
LGTM

PR #4610, opened 2023-06-01
MDEV-5 Swap exceptions check from every to some

EMcQ-BES:
### Changes:

- When checking if a user has the permission groups listed in the "exceptions" for a project excluded entity types, check for any of those permissions instead of all

### DIFF
--- a/packages/web-config-server/src/apiV1/RouteHandler.js
+++ b/packages/web-config-server/src/apiV1/RouteHandler.js
@@ -99,7 +99,7 @@ export class RouteHandler {
         const userPermissionGroups = await this.req.accessPolicy.getPermissionGroups(
           allCountryCodes,
         );
-        const userHasAccessToExcludedTypes = permissionGroups.every(permissionGroup =>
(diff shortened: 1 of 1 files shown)


packages/web-config-server/src/apiV1/RouteHandler.js
```diff
@@ -99,7 +99,7 @@ export class RouteHandler {
         const userPermissionGroups = await this.req.accessPolicy.getPermissionGroups(
           allCountryCodes,
         );
-        const userHasAccessToExcludedTypes = permissionGroups.every(permissionGroup =>
+        const userHasAccessToExcludedTypes = permissionGroups.some(permissionGroup =>
```
rohan-bes, line 102: Oh gosh, I forgot this functionality even existed. It's used by one project. I would be partial to us to scrapping it (not now though). I'll check with Georgia on the use case for it, and create a ticket to clean up if needed.

rohan-bes (approved):
Good change!

PR #4616, opened 2023-06-06
WAITP-1205: Login Modal

tcaiger:
### Changes:

- Add react-query and react-hook-form
- Add Login Modal component
- Add route for Login Modal

---

### Screenshots:
image

image


packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -15,30 +15,35 @@ interface ModalProps {
 }
 
 const Wrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
   text-align: center;
   overflow-x: hidden;
-  padding: 2em;
```
tcaiger, line 20: I changed these to rems as I found that the ems were a bit hard to maintain here since the font size in context could be coming from a few different places, the rems still fulfills the same accessibility requirement.

packages/tupaia-web/src/layout/AuthModal.tsx
```diff
@@ -4,97 +4,64 @@
...
-  // This modal will always be open in this case, because these are only visible on certain routes
+export const AuthModal = ({ children, title, subtitle, onClose }: AuthModalProps) => {
   return (
-    <Modal isOpen={true} onClose={onCloseModal}>
+    <Modal isOpen={true} onClose={onClose}>
       <Logo src={TUPAIA_LIGHT_LOGO_SRC} alt="Tupaia Logo" />
       <Title variant="h2">{title}</Title>
       {subtitle && <Subtitle variant="h3">{subtitle}</Subtitle>}
       {children}
-      {primaryButton && (
```
tcaiger, line 92: There are a few small variations in the auth modals in the designs such as the forgot password link so I refactored this to leave the button placements up the children prop.

packages/tupaia-web/src/api/api.ts
```diff
@@ -6,7 +6,8 @@
 import axios from 'axios';
 import FetchError from './fetchError';
 
-export const API_URL = process.env.REACT_APP_TUPAIA_WEB_API_URL || 'http://localhost:8100/api/v1/';
+export const API_URL =
+  import.meta.env.REACT_APP_TUPAIA_WEB_API_URL || 'http://localhost:8100/api/v1/';
```
acdunham, line 11: I think potentially this should be port 8000, as this is the `web-config-server`, until we switch over to the new server

packages/tupaia-web/src/api/mutations/useLogin.ts
```diff
@@ -0,0 +1,35 @@
...
+          emailAddress: email,
+          password,
+          deviceName: window.navigator.userAgent,
+        },
+      });
+    },
+    {
+      onSuccess: () => {
+        queryClient.invalidateQueries();
+        navigate(-1);
```
acdunham, line 31: So I think technically, if someone went straight to `/login` and then it navigated back to `-1`, it means that the user would be taken to whether they were before coming to Tupaia. So potentially we might need to check whether there is another route in the history, and if not, go to the default project?

packages/tupaia-web/src/layout/UserMenu/PopoverMenu.tsx
```diff
@@ -47,7 +49,20 @@ export const PopoverMenu = ({
...
+            <MenuItem href="/login" onCloseMenu={onCloseMenu}>
+              Log in
+            </MenuItem>
+            <MenuItem href="/register" onCloseMenu={onCloseMenu}>
+              Register
+            </MenuItem>
+          </>
+        )}
+        {children}
+      </MenuList>
```
acdunham, line 65: Are these going to be in the popover menu? Or still in the top bar?

packages/tupaia-web/src/pages/Login.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const {
+    handleSubmit,
+    register,
+    formState: { errors },
+  } = useForm();
+  const { mutate: login, isLoading, isError, error } = useLogin();
+  const navigate = useNavigate();
+
+  const onClose = () => {
+    navigate(-1);
```
acdunham, line 51: I wonder if we need to do a check here as well, to make sure navigating doesn't take the user back to another site or an empty tab, if they go directly to `/login`?

packages/tupaia-web/src/theme/constants.ts
```diff
@@ -6,7 +6,3 @@
 /**
  * Theme constants that don't require being in the entire theme object, ie. things that won't change if a custom theme was applied
  */
-
-export const FONT_SIZES = {
-  viewTitle: '2em',
-};
```
acdunham, line 12: What are your thoughts on keeping font sizes standardised somewhere? So we don't end up hard coding them in different places?

acdunham (changes requested):
A few questions, but otherwise looks good!

packages/tupaia-web/src/api/mutations/useLogin.ts
tcaiger: Very good point. Thanks for thinking about that. I have add a util for safely navigating back and used it in the auth modal for closing and also in the login on success.

packages/tupaia-web/src/theme/constants.ts
tcaiger: Yes good question. I guess I prefer to have a reusable component with styles encapsulated rather than having reusable styles separately. I find it quite hard to restyle and refactor components that are using global style settings. I also think that this particular heading is only going to be used in the auth modal so the styles for it can live in there. I do think that it would be good to have some reusable typography styles or components for typography styles that will be used widely. The one that comes to mind is the paragraph style. I'm not sure the best way to reuse that. Maybe the material-ui theme or maybe a specific <Text> component?

```
    color: white;
    font-size: 0.875rem;
    line-height: 1.3;

File: packages/tupaia-web/src/Routes.tsx [hunk 8]
```diff
@@ -33,11 +33,14 @@ export const Routes = () => {
...
         <Route path="verify-email" element={<VerifyEmailForm />} />
+        <Route path="/login" element={<Login />} />
         <Route path="/:landingPageUrlSegment" element={<LandingPage />} />
         {/** Because react-router v 6.3 doesn't support optional url segments, we need to handle dashboardCode with a splat/catch-all instead */}
         <Route path="/:projectCode/:entityCode/*" element={<Project />} />
       </RouterRoutes>
 
+      {/* The `backgroundLocation` state is the location that we were at when one of the gallery links was clicked. If it's there, use it as the location for
+      the <Routes> so we show the gallery in the background, behind the modal. See react router docs [here]{@Link https://github.com/remix-run/react-router/tree/dev/examples/modal} */}
       {state?.backgroundLocation && (
```
acdunham, line 44: Super duper pedantic, but maybe just update the comment to reflect our use case, as opposed to the example one?

packages/tupaia-web/src/utils/useNavigateBack.ts
```diff
@@ -0,0 +1,21 @@
...
+  return function navigateBack() {
+    /* The location.key property is a unique string associated with this location. On the initial location, this will be the string default. On all subsequent locations, this string will be a unique identifier.
+    { @link https://github.com/remix-run/history/blob/main/docs/api-reference.md#locationkey } */
+    if (key === 'default') {
+      navigate(DEFAULT_URL);
+    } else {
+      navigate(-1);
+    }
+  };
+};
```
acdunham, line 21: Nice way to make this reusable everywhere

packages/tupaia-web/src/layout/UserMenu/PopoverMenu.tsx
tcaiger: I have moved this to the top bar

acdunham (changes requested):

packages/tupaia-web/src/layout/TopBar/TopBar.tsx
```diff
@@ -43,11 +44,48 @@ const Header = styled.header<{
...
+
 export const TopBar = () => {
   // When handing custom landing pages, pass the primary and secondary colors to the Header component
   return (
     <Header>
       <Logo />
+      <Inner>
+        <SearchBar />
+        <RegisterButton to="/register">Register</RegisterButton>
+        <LoginButton to="/login">Log in</LoginButton>
```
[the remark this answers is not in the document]
tcaiger: ok I have removed them

acdunham (approved):

PR #4618, opened 2023-06-06
Hotfix 2023-23-1 Prevent analytics rebuild on migration

IgorNadj:


IgorNadj:
Have tested, correctly prevents the log$_question table from populating when the migration runs.

EMcQ-BES (approved):

PR #4623, opened 2023-06-08
WAITP-1216 Add project permissions to landing page projects in response

acdunham:
### Changes:
- Exposed function from `projects` API in `web-config-server` to reuse this in landing pages
- Updated landing page logic to fetch the projects and use the same logic as `projects` API to return the array of projects with the landing page


EMcQ-BES (approved):
Looks good :+1:

PR #4624, opened 2023-06-08
Hotfix: Release 2023-23-2

IgorNadj:


IgorNadj:
Tested in a deployment. Took the same amount of time to rebuild analytics table as baseline. No additional entries present in the mv queues.

PR #4625, opened 2023-06-08
MDEV-152 Update map overlay translation schema to mark values as an array

EMcQ-BES:
### Changes:

- Treat `values` as an array of objects, not an object on its own

The `lesmis` translation handling was being told to interpret this field as an object, so it got converted to an object with numeric keys `{ "0": { X, Y } }`. Since the front end was expecting an array, it would try to run an `Array.map` against this, and crash


EMcQ-BES:
Typo'd the ticket number of the commit message 🤦 Reminder to myself to adjust that when I squash and merge

rohan-bes (approved):
Ah nice! I looked into this issue originally and got very confused as to how the data was being converted from an array in the database source to an object. Never crossed my mind it could be the translation layer.

PR #4626, opened 2023-06-08
MDEV-154 Add migration for creating new fiji asset management entity types

EMcQ-BES:
### Changes:

- Add migration for new entity types
  - Following established pattern in other `AddEntityType...` migrations


rohan-bes:
Happy for these statements to just be run against prod, consider this OTS'd 👍

PR #4627, opened 2023-06-08
Waitp 1216 projects default dashboard response

acdunham:
### Changes:
- return `entity_hierarchy_id` with `getAllProjectDetails` response
- return default dashboard code with projects response in `web-config-server` (uses `dashboardGroupName` to find default dashboard if possible, otherwise returns the first dashboard code. If no dashboards can be found it returns 'General' - this is because this is what web-frontend does I believe, so I kept the same logic)


tcaiger (approved):
LGTM

PR #4629, opened 2023-06-12
WAITP-1253 Tupaia web server cleanup

EMcQ-BES:
### Changes:

- `SessionSwitchingAuthHandler`
  - AuthHandlers are run _prior_ to the `attachSession` function, so don't expect a `session` to exist on the req
  - Move class into `server-boilerplate`
    - Since we need to know more about the definition of a `session` and those types exist in `server-boilerplate`
    - This seemed cleaner to me rather than either re-defining `Session` types in `api-client`
- Update `useForwardUnhandledRequests` to allow specifying an `AuthHandler`
  - This was to allow session switching to happen on our forwarded requests
- Small cleanup - Update `DashboardsRoute` to use `entity.code` instead of `entity_id`


rohan-bes (commented):
Yup, happy with all this. Nice stuff 👍

rohan-bes (approved):

PR #4630, opened 2023-06-12
Waitp 1215 projects modal

acdunham:
### Changes:
- Added ability to have modal routes using search params
- Updated `web-config-server` to rename `defaultDashboard` as `defaultDashboardCode` to match pattern of other properties in projects response
- Added projects modal
- Updated `isUserLoggedIn` to come from `useUser` hook
- Added `LegacyProjectCard` component
- Updated anywhere that used router link buttons to use `RouterButton` component
- Added `View projects` link to user menu when in default view


packages/tupaia-web/src/layout/AuthModal.tsx
```diff
@@ -27,6 +27,13 @@ const Subtitle = styled(Typography)`
   margin-top: 1rem;
 `;
 
+const Wrapper = styled.div`
+  width: 660px;
```
tcaiger, line 31: Should be add `max-width: 100%` and should the widths be in rems?

tcaiger: let's used styled-components and pass through the className prop

packages/tupaia-web/src/pages/ModalRoute.tsx
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * This is the wrapper to handle any search param routes that should be modals
+ */
+export const ModalRoute = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const modal = searchParams.get('modal') as typeof MODAL_TYPES[keyof typeof MODAL_TYPES];
+  // If no modal param or invalid modal param, return null
+  if (!modal || !Object.values(MODAL_TYPES).includes(modal)) return null;
+
+  const modalViews = {
```
tcaiger, line 21: I don't feel too strongly but have a slight preference for putting the modalViews const outside the component body. Just above the component in this file. I think it's a bit easier to reference that way.

tcaiger (approved):
Looks great!

PR #4631, opened 2023-06-12
WAITP-1203 Add map packages

acdunham:
### Changes:
- Installed latest version of `@mapbox/polyline`
- Added `ui-components` and `ui-map-components` to project dependencies
- Added basic map and tilepicker to main view
- Arranged components in correct order with map so that controls would appear over the top of the map

Note: This is the only map-based packages used in `web-frontend` so I only added those. It seems like `ui-map-components` should be compatible with the new version of `@mapbox/polyline`, since it didn't add much to the `yarn.lock` file


packages/tupaia-web/src/layout/MapLayout/MapLayout.tsx
```diff
@@ -3,41 +3,32 @@
...
 const MapLegendWrapper = styled.div`
   display: flex;
   flex-direction: row;
   align-items: flex-end;
   justify-content: center;
+  position: absolute;
+  background-color: grey;
+  width: 300px;
+  height: 50px;
+  bottom: 1em;
```
tcaiger, line 29: Do you mind if we use rem to make it more stable?

packages/tupaia-web/src/layout/MapLayout/MapLayout.tsx
```diff
@@ -3,41 +3,32 @@
...
+  left: calc(50% - 150px);
```
tcaiger, line 30:
```suggestion
  left: 50%;
  transform: translateX(-50%);
```
What do you think about this way of centering it?

packages/tupaia-web/src/layout/MapLayout/MapLayout.tsx
```diff
@@ -53,21 +44,36 @@ const MapOverlaySelector = styled.div`
...
+  };
   return (
     <Wrapper>
-      <MapControlsContainer>
-        <MapOverlaySelector />
-        <MapLegendWrapper>{/** This is where the map legend would go */}</MapLegendWrapper>
-      </MapControlsContainer>
-      <Map />
-      {/** This is where the tilepicker would go */}
+      {/** order here matters - Map needs to be first so any controls can go over the top of it */}
```
tcaiger, line 70: I'm not sure if this comment still makes sense?
Overall this approach is much better. Thanks for updating it.

tcaiger (approved):
Looks great. I left a couple of very minor suggestions which are optional.

packages/tupaia-web/src/layout/MapLayout/MapLayout.tsx
acdunham: Usually absolutely, but as this is just a palceholder and will definitely change, I reckon we just update it when the time comes

PR #4632, opened 2023-06-12
RN-930: Added GET /downloadFiles route to central-server and web-config-server

rohan-bes:
### Issue RN-930:

In order for the document download viz type to work, we need a route that allows us to download files from S3 (so that we can user the server's AWS credentials to download private files).

### DIFF
--- a/packages/central-server/src/apiV2/DownloadFiles.js
+++ b/packages/central-server/src/apiV2/DownloadFiles.js
@@ -1,0 +1,54 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import {
+  S3,
+  S3Client,
+  ValidationError,
+  respondWithDownload,
+  writeStreamToFile,
+} from '@tupaia/utils';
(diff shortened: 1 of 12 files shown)


packages/central-server/src/apiV2/DownloadFiles.js
```diff
@@ -0,0 +1,54 @@
...
+  S3Client,
+  ValidationError,
+  respondWithDownload,
+  writeStreamToFile,
+} from '@tupaia/utils';
+import { RouteHandler } from './RouteHandler';
+import { getTempDirectory } from '../utilities';
+import { zipMultipleFiles } from './utilities';
+
+export class DownloadFiles extends RouteHandler {
```
rohan-bes, line 17: Couldn't quite decide about whether to call this `DownloadFiles` or perhaps another name like `DownloadResources`.

`DownloadFiles` feels a bit generic (for instance we already have the `export/download` route), but there isn't really a term in our system to describe this (and I don't think there needs to be for now).

chris-pollard (approved):
LGTM!

rohan-bes:
@tcaiger This PR touches web-config-server

PR #4633, opened 2023-06-12
Fixup frontend epic CI/CD

EMcQ-BES:
Fix a few technical issues breaking the codeship build

- Fix the build by removing an unused include
- Swap `ui-chart-components` to run tests as typescript
  - Change the single test file to a `.ts` file
- Update `tupaia-web` with a stub test file, so running `test` against it actually does something
- Regenerate `types`

This leaves a couple of broken unit test issues
- One in `ui-chart-components`
- One in `web-frontend`

I don't know _quite_ enough about the rewrite to solve these particular issues, so I'm going to leave those for one of the frontend devs.


tcaiger (approved):
thanks

PR #4634, opened 2023-06-12
WAITP-1206: Register form

tcaiger:
### Changes:

- Add Register Form
- Move User Routes to Modal component with modal query string param
- Add Email Verification Flow

---

### Screenshots:

Register Modal
image

Registration Success Screen
image

Resend Email Verification Modal
image


packages/tupaia-web/src/pages/Register/Register.tsx
```diff
@@ -0,0 +1,174 @@
...
+    >
+      {isSuccess ? (
+        <SignupComplete />
+      ) : (
+        <>
+          {isError && <Typography color="error">{error.message}</Typography>}
+          <StyledForm onSubmit={handleSubmit(onSubmit as SubmitHandler<any>)} noValidate>
+            <TextField
+              name="firstName"
+              label="First name *"
```
acdunham, line 75: What do you think about a way to handle the `*` in the label dynamically, when `required` is true? 
Also might be worth passing the `required` prop to the `TextField` (or does the `inputRef` apply this to the input as well?)

packages/tupaia-web/src/pages/Register/Register.tsx
```diff
@@ -0,0 +1,174 @@
...
+              helperText={errors?.firstName && errors?.firstName.message}
+              inputRef={register({
+                required: 'Required',
+              })}
+            />
+            <TextField
+              name="lastName"
+              label="Last name *"
+              error={!!errors?.lastName}
+              helperText={errors?.firstNalastNameme && errors?.lastName.message}
```
acdunham, line 86: Looks like a typo here

packages/tupaia-web/src/pages/Register/Register.tsx
```diff
@@ -0,0 +1,174 @@
...
+              })}
+            />
+            <StyledCheckbox
+              name="hasAgreed"
+              color="primary"
+              label="I agree to the terms and conditions"
+              error={!!errors.hasAgreed}
+              helperText={errors?.hasAgreed?.message}
+              inputRef={register({
+                required: 'Required',
```
acdunham, line 157: I wonder if we can DRY some of this up, since we are repeating required settings a lot in here. Not major though.

packages/tupaia-web/src/pages/VerifyEmailResend.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const CheckEmailMessage = styled.p`
+  text-align: center;
+  padding: 0 15px;
+`;
+
+const StyledForm = styled.form`
+  margin-top: 1rem;
+  width: 21rem;
+  max-width: 100%;
```
acdunham, line 26: I wonder if we can do the width here as a percentage as well, since the modal body has a set width?

packages/tupaia-web/src/pages/VerifyEmailResend.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { SubmitHandler, useForm } from 'react-hook-form';
+import { useResendVerificationEmail } from '../api/mutations';
+
+const ModalBody = styled(AuthModalBody)`
+  width: 53rem;
+`;
+
+const CheckEmailMessage = styled.p`
+  text-align: center;
+  padding: 0 15px;
```
acdunham, line 20: Can we make this use rems?

acdunham (changes requested):

packages/tupaia-web/src/pages/EmailVerification.tsx
```diff
@@ -0,0 +1,27 @@
...
+`;
+
+export const EmailVerification = () => {
+  const { isSuccess, isError } = useEmailVerification();
+
+  return (
+    <>
+      {isSuccess && <Success>Your e-mail was successfully verified</Success>}
+      {isError && <Error>Your email address could not be verified</Error>}
+    </>
```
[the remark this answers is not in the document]
tcaiger: Yes I agree that's a bit better

packages/tupaia-web/src/pages/Register/Register.tsx
```diff
@@ -0,0 +1,174 @@
...
+import { MODAL_ROUTES } from '../../constants';
+
+const ModalBody = styled(AuthModalBody)`
+  width: 53rem;
+`;
+
+const StyledForm = styled.form`
+  margin-top: 1rem;
+  width: 42rem;
+  max-width: 100%;
```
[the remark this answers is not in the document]
tcaiger: I guess I want it to not change until it has to so percentage wouldn't do the desired thing.

packages/tupaia-web/src/pages/VerifyEmailResend.tsx
tcaiger: Same as before this is intentionally set to a width that is less than the outer width to match the design as an alternative to adding padding as I find that it's a nice way to make all the widths work nicely responsively.

packages/tupaia-web/src/pages/Register/Register.tsx
tcaiger: I ended up adding a form component with context and moving some of this repeated code to the TextField component. thanks for the suggestion.

packages/tupaia-web/src/pages/Register/SignUpComplete.tsx
```diff
@@ -0,0 +1,57 @@
...
+        &nbsp;and
+        <FormLink
+          href="https://play.google.com/store/apps/details?id=com.tupaiameditrak&hl=en"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          &nbsp;Android&nbsp;.
+        </FormLink>
+      </Typography>
+      <RouterButton to={MODAL_ROUTES.VERIFY_EMAIL_RESEND}>Re-send verification email</RouterButton>
```
acdunham, line 54: Does this need to have `modal={MODAL_ROUTES.VERIFY_EMAIL_RESEND`?

packages/tupaia-web/src/utils/useModal.ts
```diff
@@ -0,0 +1,19 @@
...
+  const { hash, ...location } = useLocation();
+
+  function navigateToModal(hashKey: string) {
+    navigate({ ...location, hash: hashKey });
+  }
+  function closeModal() {
+    navigate(location);
+  }
+  return { hash: hash.substring(1), closeModal, navigateToModal };
+};
```
acdunham, line 19: Nice, I like this

acdunham (approved):
Just a couple very minor comments/suggestions, but otherwise happy to approve :) Nice work, it looks really good

PR #4638, opened 2023-06-13
[no-issue]: Add generational_distance filter in entity-server

kael89:
### Issue #:

This is required for OSC-17. It's used in https://github.com/beyondessential/tupaia/pull/4353, but I'm looking to get this merged separately to avoid further conflicts.

This PR is almost the same as the already approved https://github.com/beyondessential/tupaia/pull/4352 (reviewed by @edmofro).  Changes are:

* Removed any functionality that is implemented in `dev` now (`central-server: GetEntityRelations`)
* Removed `admin-panel-server: FetchEntityHierarchyTreeRoute` as it won't be needed
* Conflicts with `dev` resolved

After that there should be one more PR and OSC-17 should be ready for testing!

---

**Testing notes:** this PR does touch filtering in `entity-server`, so it may need to be QAed. At the same time, `entity-serfver` has great test coverage. I'll leave the testing strategy to you - happy to help with QA if there are specific things I could test locally @rohan-bes

### DIFF
--- a/packages/entity-server/src/routes/hierarchy/middleware/filter.ts
+++ b/packages/entity-server/src/routes/hierarchy/middleware/filter.ts
@@ -5,8 +5,9 @@
 
 import { QueryConjunctions } from '@tupaia/server-boilerplate';
 import { ObjectLikeKeys, Flatten } from '@tupaia/types';
-import { Writable } from '../../../types';
-import { EntityFilter, EntityFields } from '../../../models';
+import { getSortByKey } from '@tupaia/utils';
+import { Writable, NumericKeys } from '../../../types';
+import { EntityFilter, EntityQueryFields } from '../../../models';
 
 const CLAUSE_DELIMITER = ';';
 const NESTED_FIELD_DELIMITER = '_';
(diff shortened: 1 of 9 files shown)


packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -58,35 +68,56 @@ const operatorToSqlComparator = {
...
-    comparisonValue: formatComparisonValue(value, operator),
+    comparisonValue: formatComparisonValue(formattedValue, operator),
   };
 };
 
 const toFilterClause = (queryClause: string) => {
-  const operator = filterOperators.find(o => queryClause.includes(o));
+  const [operator] = filterOperators
+    .filter(o => queryClause.includes(o))
+    .sort(getSortByKey('length', { ascending: false }));
```
edmofro, line 120: Clever! Smells a little too clever, but it does work every time :-)

edmofro (approved):
lgtm @kael89! 

I think we should have a quick manual regression test of entity filtering before merging - @rohan-bes can you suggest where we should do this?

rohan-bes:
> I think we should have a quick manual regression test of entity filtering before merging

I don't think we actually use this much in Tupaia. LESMIS does its filtering on the frontend.

I do know that the `dataSourceEntityFilter` aggregation config option uses it, of which I could only find one viz that actually uses that anymore: `ncd_ws_text_percentage_primary_schools_assessed_using_sns`

Alternatively, Waiheke are just about to switch `tupaia-web` over to using the entity-server for search, so we will get a nice proper test at that point 👍

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
rohan-bes: Okay honestly I am confused 😂 what does this do??

rohan-bes (approved):
Thanks @kael89, this is great!

packages/entity-server/src/types.ts
```diff
@@ -18,6 +18,16 @@ export interface EntityServerModelRegistry extends ModelRegistry {
...
+// Extracts keys that have object-like values from type T
+export type ObjectLikeKeys<T> = {
+  [K in keyof T]: T[K] extends Record<string, unknown> ? K : never;
+}[keyof T];
+
+// Extracts keys that have numeric values from type T
+export type NumericKeys<T> = {
+  [K in keyof T]: T[K] extends number ? K : never;
+}[keyof T];
+
```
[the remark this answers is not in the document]
kael89: Ah coolz, will import from there then

packages/entity-server/src/routes/hierarchy/middleware/filter.ts
kael89: It expands the match to use the longest operator found, e.g. `>=` instead of just `>`. The first `filter` pass will match both of them, so then we need to decide which one to use.

Alternatively we could change the order of `filterOperators` so that the longest operators are listed first, then use the first matched operator. Or we could sort that array by desc length. But I think that this would be a bit less implicit than what we do here...

I will add a comment

kael89:
@rohan-bes @edmofro I addressed all comments - I'll leaving merging to you :raised_hands:

rohan-bes:
Tested manually on a feature instance and all seemed good.

PR #4640, opened 2023-06-14
Fix chart tests

acdunham:
### Changes:
- Updated tests to be correct types in `ui-chart-components`
- Updated `$all` to be optional in `parseChartConfig`, because the method itself handles when this is unset
- Updated `parseChartConfig` to not try and spread `name` string


tcaiger (approved):
LGTM. thanks for doing this, it is totally my bad for not including it in my original PR 🙏

PR #4642, opened 2023-06-15
Waitp 1208 reset password form

acdunham:
### Changes:
- Added a 'forgot password' modal route and view
- Added reset password view and logic
- Updated url to be added to email in reset password logic to handle the modal routes with hash and search params
- Updated `closeModal` logic to handle removal of password reset token
- Updated `isLoggedIn` logic to use `email` instead of `name` to determine if user is logged in or not
- Added 'Change password' to user menu for default views when user is logged in

Note: There was no design for change password form, so I went with the existing styles of the other modals we have to keep it streamlined with the auth modals, instead of restyling to be like the current one we have

### Screenshots:

Without token
Screenshot 2023-06-15 at 12 01 40 pm

With token
Screenshot 2023-06-15 at 12 02 43 pm
 
Error message
Screenshot 2023-06-15 at 12 04 26 pm

Success message
Screenshot 2023-06-15 at 12 04 48 pm

Mobile
Screenshot 2023-06-15 at 12 03 30 pm


packages/central-server/src/apiV2/requestPasswordReset.js
```diff
@@ -28,7 +28,7 @@ export const requestPasswordReset = async (req, res) => {
     user_id: user.id,
   });
 
-  const passwordResetUrl = `${process.env.TUPAIA_FRONT_END_URL}/reset-password?passwordResetToken={token}`;
+  const passwordResetUrl = `${process.env.TUPAIA_FRONT_END_URL}?passwordResetToken={token}#reset-password`;
   const resetUrl = passwordResetUrl.replace('{token}', token);
```
acdunham, line 32: Search params have to be after the hash so that the token isn't assumed by the browser to be part of the hash

packages/tupaia-web/src/api/queries/useUser.ts
```diff
@@ -12,6 +12,6 @@ export const useUser = () => {
 
   return {
     ...userResponse,
-    isLoggedIn: data?.name !== undefined,
+    isLoggedIn: data?.email ? true : false, // when the name is public
```
tcaiger, line 15: Oh yes the reason I changed this to name!== undefined is because in the new server it simply returns {} if the user is logged out. I think what you've done is ideal anyway since it will work with both the old and the new server.

packages/tupaia-web/src/pages/ForgotPassword.tsx
```diff
@@ -0,0 +1,87 @@
...
+  text-transform: none;
+  font-size: 1rem;
+  width: 22rem;
+  max-width: 100%;
+  margin-left: 0 !important;
+  padding: 0.375rem 1rem; // to match the height of the primary button
+  border-color: ${({ theme }) => theme.palette.text.secondary};
+  ${AuthModalButton} + & {
+    margin-top: 1.3rem;
+  }
```
tcaiger, line 35:
```suggestion

    margin-top: 1.3rem;
  
```
Since it's only used in one place that style can be simplified

packages/tupaia-web/src/api/mutations/useRequestResetPassword.ts
```diff
@@ -0,0 +1,29 @@
...
+        data: {
+          emailAddress,
+        },
+      });
+    },
+    {
+      onSuccess: () => {
+        queryClient.invalidateQueries();
+      },
+    },
```
tcaiger, line 27:
```suggestion
   
```
The invalidate queries cleares all the react query caches. We need it when we login and logout as the user's permissions determine which data they get from the api but we don't need it here as the user stays logged out

packages/central-server/src/apiV2/requestPasswordReset.js
tcaiger: This will also be used in lesmis so will break if we change it. How about we add a redirect in the front end routes to redirect from /reset-password?passwordResetToken={token}` to ?passwordResetToken={token}#reset-password`?

packages/tupaia-web/src/api/mutations/useResetPassword.ts
```diff
@@ -0,0 +1,38 @@
...
+
+type ResetPasswordParams = {
+  oldPassword: string;
+  newPassword: string;
+  passwordConfirm: string;
+};
+export const useResetPassword = () => {
+  const queryClient = useQueryClient();
+  const [urlParams] = useSearchParams();
+  const oneTimeLoginToken = urlParams.get(PASSWORD_RESET_TOKEN_PARAM);
```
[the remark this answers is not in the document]
acdunham: Yeah I saw that old flow, but figured it would be better to have 1 less step. It did work for me - what's happening for you?

packages/tupaia-web/src/utils/useModal.ts
```diff
@@ -3,18 +3,26 @@
...
   const { hash, ...location } = useLocation();
+  const [urlParams] = useSearchParams();
 
   function navigateToModal(hashKey: `${MODAL_ROUTES}`) {
     navigate({ ...location, hash: hashKey });
   }
   function closeModal() {
-    navigate(location);
+    // Remove the password reset token from the url, if it is set
+    if (urlParams.get(PASSWORD_RESET_TOKEN_PARAM)) {
```
tcaiger, line 19: It would be nice to keep the specific url params out of the util if possible to prevent it getting messy. Can remove this from here and add the `urlParams.delete(PASSWORD_RESET_TOKEN_PARAM)` to the `useResetPassword` `onSuccess` handler. That would also clear the url params after the form submits instead of when the modal is closed which is more appropriate any way.

tcaiger (changes requested):
Nice work. Just left a couple of suggestions. Thanks

tcaiger (approved):
Looks great

PR #4647, opened 2023-06-16
WAITP-1287: Dashboard layout

tcaiger:
### Changes:

- Dashboard layout
- Static Map component
- Entity Photo component

---

### Screenshots:

Dashboard Static Map
image

Expanded Dashboard
image

Dashboard with Photo
image

Dashboard Photo Modal (Very basic but same as existing)
image


packages/tupaia-web/src/api/queries/useEntity.ts
```diff
@@ -0,0 +1,30 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { useQuery } from 'react-query';
+import { sleep } from '@tupaia/utils';
+
+const response = {
```
tcaiger, line 9: this is just mock data for testing

packages/tupaia-web/src/layout/Sidebar/Breadcrumbs.tsx
```diff
@@ -0,0 +1,16 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import styled from 'styled-components';
+
+export const Breadcrumbs = styled.div`
```
tcaiger, line 8: This is just temporary. The breadcrumbs will be added in waitp-1212

packages/tupaia-web/src/layout/Sidebar/Sidebar.tsx
```diff
@@ -5,41 +5,122 @@
...
+}>`
+  display: grid;
+  background-color: ${PANEL_GREY};
+  grid-template-columns: repeat(auto-fill, minmax(${MIN_SIDEBAR_WIDTH}px, auto));
+  column-gap: 0.5rem;
+  row-gap: 0.5rem;
+  padding: ${({ $isExpanded }) => ($isExpanded ? '0 0.5rem 0.5rem' : '0 0 0.5rem 0')};
+`;
+
+const Chart = styled.div`
```
tcaiger, line 81: The charts are just temporary for testing the layout

packages/tupaia-web/src/layout/Sidebar/Sidebar.tsx
```diff
@@ -5,41 +5,122 @@
...
+  position: relative;
+  background-color: ${TRANSPARENT_BLACK};
+  // Use padding to maintain aspect ratio
+  padding-bottom: 80%;
 `;
 
 export const Sidebar = () => {
   const [isExpanded, setIsExpanded] = useState(false);
-
+  const { data: entityData } = useEntity('fiji');
```
acdunham, line 90: Do we need to swap out `fiji` for the entity in the URL params?

packages/tupaia-web/src/layout/Sidebar/Sidebar.tsx
```diff
@@ -5,41 +5,122 @@
...
       <ExpandButton setIsExpanded={toggleExpanded} isExpanded={isExpanded} />
+      <ScrollBody>
+        <Breadcrumbs />
+        {bounds ? (
+          <StaticMap polygonBounds={bounds} />
+        ) : (
+          <Photo title={entityData?.name} photoUrl={entityData?.photoUrl} />
+        )}
+        <TitleBar>
+          <Title variant="h5">Northern</Title>
```
acdunham, line 107: Wondering if this should be an `h3` or `h4` since `h5` skips a hierarchy level I think?

packages/tupaia-web/src/layout/Sidebar/StaticMap.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const longs = bounds.map(latLong => latLong[1]);
+  const minLat = Math.min(...lats);
+  const maxLat = Math.max(...lats);
+  const minLong = Math.min(...longs);
+  const maxLong = Math.max(...longs);
+
+  const latitude = minLat + (maxLat - minLat) / 2;
+
+  let longitude = minLong + (maxLong - minLong) / 2;
+
```
acdunham, line 27: This is pedantic, but I wonder if we can DRY this up with a reusable function, since these essentially do the same thing? Could accept an array of coords, make it calculate the min, max, and actual value, and then the longitude could do something beyond that with the rest of the logic, if necessary.

packages/tupaia-web/src/layout/Sidebar/StaticMap.tsx
```diff
@@ -0,0 +1,73 @@
...
+    [polygonBounds[0][0], polygonBounds[0][1]],
+    [polygonBounds[0][0], polygonBounds[1][1]],
+    [polygonBounds[1][0], polygonBounds[1][1]],
+  ];
+
+  const encodedPolyline = encodeURIComponent(
+    polyline.encode(polygonPoints as Array<[number, number]>),
+  );
+
+  const showBox = polygonBounds !== DEFAULT_BOUNDS;
```
acdunham, line 50: I have a feeling using `!==` with arrays won't work, and would potentially always be `false`... Might need to use `.every()` or compare the stringified arrays here

acdunham (changes requested):
Just a couple of questions. I also get an error about types for `@mapbox/polyline`, so may need to install a `@types` package

packages/tupaia-web/src/api/queries/useEntity.ts
```diff
@@ -0,0 +1,30 @@
...
+    'https://tupaia.s3-ap-southeast-2.amazonaws.com/uploads/images/1499489588784_647646.png',
+  countryHierarchy: [],
+};
+
+export const useEntity = (entityCode: string) => {
+  return useQuery(['entity', entityCode], async () => {
+    await sleep(1000);
+    return response;
+  });
+};
```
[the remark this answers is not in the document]
tcaiger: I'm not sure yet. Let's see when the actual api is ready.

packages/tupaia-web/src/layout/Sidebar/Sidebar.tsx
tcaiger: yes good point

packages/tupaia-web/src/layout/Sidebar/StaticMap.tsx
tcaiger: Yes you are absolutely right. I copied this from the old site and didn't think about it that much and so after your review comment I was confused about why it works on the old site. I think it's because the polygonBounds 
 are set using the actual DEFAULT_BOUNDS const upstream so the equality does actually work which is just lucky and pretty dodgy. I ended up doing the array comparison just by using JSON.stringify which works well for small arrays as long as there are no nested functions which fits this use case.

acdunham (approved):
Looks good, thanks!

PR #4648, opened 2023-06-16
WAITP-1207 Add getProject route

acdunham:
### Changes:
- Added `fetchProjects` to web config in  `api-client`
- Added `/project/:projectCode` route to `tupaia-web-server`


EMcQ-BES (approved):
👍

PR #4653, opened 2023-06-19
Waitp 1207 request access form

acdunham:
### Changes:
- Created a request project access view 
- Created a request country access view 
- Handled logic where url search params for modal views need to be removed on close of modal
- Handled logic when countries have already been requested
- Added a CheckboxList component for an array of checkboxes for a field
- Added an Alert component to match from old `web-frontend`

### Screenshots

Full access already
Screenshot 2023-06-19 at 12 22 19 pm

Request access list
Screenshot 2023-06-19 at 12 23 43 pm

Already request access to some countries
Screenshot 2023-06-19 at 12 25 19 pm
60a9e822984)


packages/tupaia-web/src/api/queries/useCountryAccessList.ts
```diff
@@ -0,0 +1,230 @@
...
+    accessRequests: [],
+  },
+  {
+    id: '5e38df0b61f76a6bcf212630',
+    name: 'Wallis and Futuna',
+    hasAccess: false,
+    accessRequests: [],
+  },
+];
+
```
acdunham, line 216: This is temporary data until the working entities endpoint has been merged in

packages/web-config-server/examples.http
```diff
@@ -35,4 +35,3 @@ content-type: {{contentType}}
 ### Fetch Flutracking LGA % participants with fever and cough map overlay
 GET {{host}}/api/v1/measureData?endDate=2021-11-14&mapOverlayCode=AU_FLUTRACKING_LGA_Fever_And_Cough&organisationUnitCode=AU&projectCode=covidau&shouldShowAllParentCountryResults=true&startDate=2021-11-08
 content-type: {{contentType}}
-
```
acdunham, line 38: This is just formatting

packages/tupaia-web/src/utils/useModal.ts
```diff
@@ -13,8 +14,11 @@ export const useModal = () => {
   function navigateToModal(hashKey: `${MODAL_ROUTES}`) {
     navigate({ ...location, hash: hashKey });
   }
-  function closeModal() {
```
tcaiger, line 16: I think there are scenarios where the modal needs to be closed but the url query string params need to be retained.  Would it be possible to keep the close modal as it was and do the removing of params in the onSuccess handler of the specific form that you are trying to clear the query string param for?

packages/tupaia-web/src/pages/RequestProjectAccess/RequestedCountries.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
+import { CountryAccessListItem } from '../../types';
+import { Link, Typography } from '@material-ui/core';
+import styled from 'styled-components';
+import { List } from '@material-ui/core';
+import { AuthModalButton, RouterButton } from '../../components';
+import { MODAL_ROUTES, PROJECT_PARAM } from '../../constants';
+import { TextButton } from '@tupaia/ui-components';
```
tcaiger, line 12:
```suggestion

```
This looks unused

packages/tupaia-web/src/utils/useModal.ts
acdunham: So this is for when you close the modal by clicking out of the modal or clicking outside of it. And this one will only remove any params you pass in to be removed, e.g. the password reset token or the `project=` param

packages/tupaia-web/src/api/mutations/useRequestCountryAccess.ts
```diff
@@ -0,0 +1,30 @@
...
+  message?: string;
+};
+export const useRequestCountryAccess = () => {
+  const [urlParams] = useSearchParams();
+  const projectCode = urlParams.get(PROJECT_PARAM);
+  return useMutation<any, Error, RequestCountryAccessParams, unknown>(
+    ({ entityIds, message }: RequestCountryAccessParams) => {
+      return post('requestCountryAccess', {
+        data: {
+          entityIds: Array.isArray(entityIds) ? entityIds : [entityIds], // Ensure entityIds is an array, as when there is only one option in a checkbox list, react-hook-form formats this as a single value string
```
tcaiger, line 23: Nicely handled

packages/tupaia-web/src/components/CheckboxField.tsx
```diff
@@ -31,7 +42,7 @@ export const CheckboxField = ({ name, label, required, options = {}, ...props }:
       required={required}
       color="primary"
       error={!!errors[name]}
-      helperText={errors[name]?.message}
+      helperText={isArray ? '' : errors[name]?.message}
```
tcaiger, line 45: This is an ok solution but I wonder if we would do it without needing to make the CheckboxField aware of the other CheckboxList field. Perhaps we can just set helperText={null} in the CheckboxList field instead?

packages/tupaia-web/src/components/CheckboxList.tsx
```diff
@@ -0,0 +1,56 @@
...
+  name: string;
+  required?: boolean;
+}
+
+export const CheckboxList = ({ options, legend, name, required }: CheckboxListProps) => {
+  const { errors = {} } = useFormContext();
+
+  return (
+    <Fieldset>
+      <Legend>{legend}</Legend>
```
tcaiger, line 42: Great to see the correct semantic html

packages/tupaia-web/src/components/RouterButton.tsx
```diff
@@ -14,13 +15,20 @@ interface RouterLinkProps extends Record<string, any> {
...
+export const RouterButton = ({
+  to,
+  modal,
+  removeParams,
+  children,
+  ...props
+}: RouterButtonProps) => {
   const location = useLocation();
-  const link = modal ? { ...location, hash: modal } : to;
+  const link = modal ? { ...location, hash: modal, search: useRemoveParams(removeParams) } : to;
```
tcaiger, line 31: I'm not quite sure I understand what is happening here. Do you want the modal links to never have query string params? Why is that?

packages/tupaia-web/src/constants/constants.ts
```diff
@@ -36,4 +37,5 @@ export const FORM_FIELD_VALIDATION = {
   },
 };
 
+export const PROJECT_PARAM = 'project';
```
tcaiger, line 40: Good idea adding this as a const

packages/tupaia-web/src/layout/UserMenu/UserMenu.tsx
```diff
@@ -85,12 +85,21 @@ export const UserMenu = () => {
...
     : [VisitMainSite];
 
-  const baseMenuItems = isLoggedIn ? [ViewProjects, ChangePassword, Logout] : [ViewProjects];
+  const baseMenuItems = isLoggedIn
+    ? [
+        ViewProjects,
+        ChangePassword,
+        <BaseMenuItem key="request-country-access" modal={MODAL_ROUTES.REQUEST_COUNTRY_ACCESS}>
+          Request country access
+        </BaseMenuItem>,
```
tcaiger, line 100: I think you accidentally removed the logout button

packages/tupaia-web/src/components/RouterButton.tsx
acdunham: So I made a way to remove specified params, so `removeParams` is an array, and `useRemoveParams` removes any params in that array, otherwise returns as is. There are a few modals, e.g. OneTimeLogin and RequestProjectAccess that have search params as well as hash, and when they close, we want them to remove the associated search params

packages/tupaia-web/src/pages/ModalRoutes.tsx
```diff
@@ -40,8 +47,12 @@ export const ModalRoutes = () => {
   if (!modal || !Object.values(MODAL_ROUTES).includes(modal)) return null;
 
   const ModalView = modalViews[modal];
+
+  const onCloseModal = () => {
+    closeModal(modalParams[modal as keyof typeof modalParams]);
```
tcaiger, line 52: Ah I see you are clearing the params onclose even if the form is not submitted. This actually makes quite a lot of sense now that I see it in action.

packages/tupaia-web/src/utils/useModal.ts
tcaiger: Now that I've read the rest of the code I can see how this is working now. It only removes the passed in params so should work for all cases. You can disregard the comment above 🙏

packages/tupaia-web/src/utils/useRemoveParams.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+export const useRemoveParams = (paramsToRemove?: string[]) => {
```
tcaiger, line 6:
```suggestion
export const removeUrlSearchParams = (paramsToRemove?: string[]) => {
```
The `use` prefix is usually only used for custom react hooks which is a function that has a hook inside of it. Since there isn't one in this util it's probably best not to use the prefix. Also I always get mixed up between url params and query string params so it would be good to be explicit with the name of the util.

packages/tupaia-web/src/pages/RequestProjectAccess/ProjectHero.tsx
```diff
@@ -0,0 +1,69 @@
...
+  project?: SingleProject;
+}
+
+export const ProjectHero = ({ project }: ProjectHeroProps) => {
+  return (
+    <FullWidth>
+      <HeadingWrapper>
+        <Heading>Requesting Project Access</Heading>
+      </HeadingWrapper>
+      <HeroImage src={project?.imageUrl} aria-hidden={true}>
```
tcaiger, line 64: Need to resolve the ts error about src being sometimes undefined

packages/tupaia-web/src/pages/RequestCountryAccess.tsx
```diff
@@ -0,0 +1,117 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
+import { SubmitHandler, useForm } from 'react-hook-form';
+import {
+  AuthModalButton,
+  CheckboxField,
```
tcaiger, line 9:
```suggestion

```

packages/tupaia-web/src/components/Alert.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * Tupaia
+ * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
+import MuiAlert, { AlertProps } from '@material-ui/lab/Alert';
+import styled from 'styled-components';
+import { CheckCircle, Warning } from '@material-ui/icons';
+
+const StyledSmallAlert = styled(MuiAlert)`
```
tcaiger, line 11: Can we use the Alert component that's in ui-components. It doesn't matter if it's not 100% the same as the alert in the old site as the alerts are pretty inconstant any way. Or we could extend the ui-components ones if need be. also I would prefer to just use the material-ui theme for the colors. We can set alert colors in the theme if need be but I'm sure the error and success ones are already pretty good.

tcaiger (changes requested):
Well done this is a huge amount of work. I left a few bits and bobs to look at.

tcaiger (approved):
🎉

PR #4655, opened 2023-06-19
WAITP-1287: Dashboard dropdown

tcaiger:
### Changes:

- Tweak DashboardRoute to return correct dashboards and to return response in camel case
- Add Dashboards query
- Add dashboard menu component which creates a menu of links to specific dashboards

---

### Screenshots:
image


packages/tupaia-web/src/layout/Sidebar/DashboardMenu.tsx
```diff
@@ -0,0 +1,83 @@
...
+import styled from 'styled-components';
+import { PANEL_GREY } from '../../constants';
+import { useDashboards } from '../../api/queries';
+import { DashboardCode } from '../../types';
+
+const MenuButton = styled(ButtonBase)`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background: ${PANEL_GREY};
```
acdunham, line 19: What do you think of putting this grey in the theme? I'm guessing that if we ever themed the app for a particular organisation then this colour may potentially be a part of that?

packages/tupaia-web-server/src/routes/DashboardsRoute.ts
```diff
@@ -4,28 +4,40 @@
...
-  any,
-  any,
-  any
->;
+export type DashboardsRequest = Request<any, any, any, any>;
 
 export class DashboardsRoute extends Route<DashboardsRequest> {
   public async buildResponse() {
     const { query, ctx } = this.req;
     const { organisationUnitCode, projectCode } = query;
```
acdunham, line 15: We were thinking about removing `organisationUnitCode` in favour of `entityCode`, right? What do you think of doing the web-server side of it now, so that eventually it just slots right into the entity server as and when changes are made? I know it's not strictly part of this ticket, but something to consider I guess?

acdunham (approved):
This is looking really good. I had a couple of questions, but happy to pre-approve.

packages/tupaia-web-server/src/routes/DashboardsRoute.ts
```diff
@@ -4,28 +4,40 @@
...
+      })
+    )[0];
+    const baseEntity = await ctx.services.entity.getEntity(
+      project['entity_hierarchy.name'],
+      organisationUnitCode,
+    );
+
+    const dashboards = await ctx.services.central.fetchResources('dashboards', {
+      filter: { root_entity_code: baseEntity.code },
+    });
```
EMcQ-BES, line 30: A dashboard attached to an entity can also appear on any of that entity's children (depending on the existence of dashboard items which are usually specific to entity types) which is why I was checking against ancestors.

This route handler needs another fix-up pass anyway so I'm happy to pick that up in a separate PR

packages/tupaia-web-server/src/routes/DashboardsRoute.ts
EMcQ-BES: @acdunham The back end needs a bit of a clean up anyway so I reckon if the front end looks good let's not let the back end hold it up

packages/tupaia-web/src/layout/Sidebar/DashboardMenu.tsx
tcaiger: yes that's a good idea. We will have to come up with a scheme to make all these theme colors make sense at some point. Perhaps it will need to evolve over time. I've added a panel section to the theme and used that through out the side bar for now.

PR #4659, opened 2023-06-20
WAITP-1205 Show login form on first load of projects view

acdunham:
### Changes:
- If user is not logged in when main view first loads, and the project is private, show the login screen
- If the user is logged in and goes to `/` route, direct to projects modal 
- If the user is not logged in and goes to `/` route, direct to login modal
- If the user is requesting access to a project by going directly to a URL, and they are not logged in, direct to login, and then redirect back to the request access modal. This is because the country access list is inaccurate if the user is not logged in. The user also would get bounced if they tried to request access when not logged in. As part of this, added support in login logic for `backgroundLocation` in location state, to allow for redirect routes


packages/tupaia-web/src/Routes.tsx
```diff
@@ -7,23 +7,42 @@ import { Navigate, Route, Routes as RouterRoutes, useLocation } from 'react-rout
...
- * **/
+ *
+ * * */
 
 export const Routes = () => {
-  let location = useLocation();
+  const location = useLocation();
+  const { isLoggedIn, isLoading } = useUser();
+  // if the user is loading, show a loading screen, so that we can handle redirects when this is done
+  if (isLoading)
```
tcaiger, line 24: It's a bit of a shame that nothing really gets rendered if the user request is loading. While that is pretty unavoidable for the "/" route, it could be avoided for other routes. Could we move this is loading check into the redirect somehow. Maybe something like this

````
const HomeRedirect = () => {
  const { isLoading, isLoggedIn } = useUser();

  if (isLoading) {
    return <LoadingScreen isLoading />;
  }
  return (
    <Navigate
      to={`${DEFAULT_URL}#${isLoggedIn ? MODAL_ROUTES.PROJECTS : MODAL_ROUTES.LOGIN}`}
      replace
    />
  );
};

<Route path="/" element={<HomeRedirect />} />

File: packages/tupaia-web/src/api/mutations/useLogin.ts [hunk 2]
```diff
@@ -28,7 +32,12 @@ export const useLogin = () => {
     {
       onSuccess: () => {
         queryClient.invalidateQueries();
-        closeModal();
+        // if the user was redirected to the login page, redirect them back to the page they were on
+        if (location.state?.backgroundLocation)
```
tcaiger, line 36:
```suggestion
        if (location.state?.referrer)
```

Nice - this is a good idea and we shouldn't need it much since the hashes can just be appended to urls but I imagine there will be a few uses for it. Could we rename `backgroundLocation` to `referrer` as I think it describes our scenario more accurately?

packages/tupaia-web/src/pages/Project.tsx
```diff
@@ -17,11 +22,22 @@ const Container = styled.div`
...
 export const Project = () => {
-  // Use these to fetch the project and any other entity info you might need
-  // const { projectCode, entityCode, '*': dashboardCode } = useParams();
+  const { projectCode } = useParams();
+  const { data: project, isLoading } = useProject(projectCode);
+  const { isLoggedIn } = useUser();
+  const { navigateToModal } = useModal();
 
+  // check if the user is logged in. If they are not logged in and the project requires login, direct first to the login modal
+  useEffect(() => {
```
tcaiger, line 31: This looks good. We might have to do something similar for the entity requests and the dashboard requests so it might make sense to make something generic and I'm not sure if it's best placed in the query hook or in the view code. I think this is good for now but let's keep it in mind as we add those other use cases.

packages/tupaia-web/src/pages/Project.tsx
```diff
@@ -17,11 +22,22 @@ const Container = styled.div`
...
+    const checkLogin = () => {
+      if (isLoggedIn || isLoading || (project && project.hasAccess)) return;
+      navigateToModal(MODAL_ROUTES.LOGIN);
```
tcaiger, line 34: What do you think about adding navigateToLogin to the useModal util?

packages/tupaia-web/src/pages/RequestProjectAccess/RequestProjectAccess.tsx
```diff
@@ -26,14 +28,38 @@ const ModalBody = styled.div`
...
+    const checkLogin = () => {
+      if (isLoadingUser || isLoggedIn || isFetching) return;
+      navigate(
+        {
+          ...location,
+          hash: MODAL_ROUTES.LOGIN,
+        },
+        {
+          state: {
+            backgroundLocation: location,
```
tcaiger, line 50:
```suggestion
            referrer: location,
```

tcaiger (changes requested):
Looks good. Sorry I'm being a bit pedantic on these ones since they're pretty core to the app.

tcaiger (approved):

PR #4663, opened 2023-06-21
WAITP-1256 Dashboard route updates

EMcQ-BES:
### Changes:

- Split out a `legacyDashboardReport` route
- Add a `getAncestors` style function to `entity-server`
  - Follows the same pattern as `getDescendants`
- Update the `dashboards` route to fetch the full `dashboard_items` list, instead of just `dashboard_relations`
- Update `dashboards` route to use url param for `entityCode`


packages/tupaia-web-server/src/routes/DashboardsRoute.ts
```diff
@@ -21,22 +26,62 @@ export class DashboardsRoute extends Route<DashboardsRequest> {
...
       project['entity_hierarchy.name'],
       entityCode,
+      {},
+      true,
     );
+    const rootEntity = entities.find((e: Entity) => e.code === entityCode);
 
     const dashboards = await ctx.services.central.fetchResources('dashboards', {
-      filter: { root_entity_code: baseEntity.code },
+      filter: { root_entity_code: entities.map((e: Entity) => e.code) },
```
rohan-bes, line 39: Appreciate you getting all this going using the central-api. If it ever feels too complex or that we're not keeping things DRY enough, I think we can consider giving `tupaia-web-server` access to some of the models here (eg. dashboard, dashboardItem, etc.). Being able to use of some the utility functions like `getDashboards` could be useful.

The big pro of just using the central-api is that we keep permissions centralised which is really nice. So let's see how things go

rohan-bes (approved):
Looks good. Just a request for some unit tests and an example in the `entity-server` for the new route 👍

packages/tupaia-web-server/src/routes/DashboardsRoute.ts
EMcQ-BES: Yeah I've considered it but I also feel like that sorta betrays the core tenant of an orchestration server. I think generally it's nicer to extend functionality of our micro-services (which is why in a case like this I extended the entity server) and keep the orchestration servers using minimal DB fetches

packages/entity-server/src/routes/hierarchy/EntityAncestorsRoute.ts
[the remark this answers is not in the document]
EMcQ-BES: Updated to add both 👍

PR #4670, opened 2023-06-23
Vite hot reloading

acdunham:
### Changes:
- change `.js` react files to `.jsx` extensions in `ui-components` because `vite` doesn't like `jsx` in `js`
- Update `vite` config to handle hot reloading for `tupaia-web` and any internal component libraries
- Fixed issue in `Toolbar.tsx` in `ui-components` that was trying to use `useRouteMatch` in from `react-router-dom`, but passing in the `baseRoute` from the component that consumes it
- Updated `storybook` in `ui-components` to work
- Removed now unnecessary scripts from `package.json` in `tupaia-web`


tcaiger (approved):
Thanks for doing this. And great brainwave on the refactor of the toolbar. I think that is the way to go for the other react router components.

packages/ui-components/.storybook/webpack.config.js
```diff
@@ -36,6 +36,9 @@ module.exports = async ({ config }) => {
     ...config.resolve.alias,
     yargs: path.resolve(__dirname, 'moduleMock.js'),
     child_process: path.resolve(__dirname, 'moduleMock.js'),
+    '@aws-sdk/credential-providers': path.resolve(__dirname, 'moduleMock.js'),
```
IgorNadj, line 39: Is this the same issue as RN-950?

acdunham: I think so! It just started so I think it must be

packages/tupaia-web/vite.config.ts
```diff
@@ -40,6 +41,19 @@ export default defineConfig(({ command }) => {
         ...baseConfig.define,
         global: {},
       },
+      resolve: {
```
IgorNadj, line 44: I'm not sure how maintainable this is. We're basically re-specifying all dependencies of a frontend here?

The nice thing about the old way was that we don't have to do that, it reads the dependencies from package.json and generates the list of dirs to watch like that.

tcaiger: This is just for hot reload so it's probably not a fair comparison to the old build tooling since hot reload didn't work with ui-components, it was just live reload and it was not very reliable. It's only for other frontend internal dependencies that are in the mono-repo so it will only ever be these 3 dependencies so I think it is pretty minor. It will be the exact same for the other front end apps when we move them to vite so we should probably make a shared vite config for the mono-repo in the same way that we do for babel and eslint and that should help ensure that the list of aliases in this file are easier to maintain.

IgorNadj: I see, sounds good, thanks for the info 🙏

PR #4672, opened 2023-06-23
WAITP-1211: Entity Dropdown Menu

tcaiger:
### Changes:

- Add entities fetch
- Add recursive entity menu
- Add placeholder search bar

---

### Screenshots:
image

image


packages/tupaia-web/src/features/EntitySearch/SearchBar.tsx
```diff
@@ -0,0 +1,55 @@
...
+  onFocusChange: (isFocused: boolean) => void;
+}
+
+export const SearchBar = ({ value = '', onChange, onFocusChange }: SearchBarProps) => {
+  const onChangeInputValue = (event: React.ChangeEvent<HTMLInputElement>) => {
+    onChange(event.target.value);
+  };
+
+  return (
+    <SearchInput value={value} onChange={onChangeInputValue} onFocus={() => onFocusChange(true)} />
```
acdunham, line 53: Not sure, but would it be useful to have an `onBlur` event listener here too?

acdunham (changes requested):
Looks good! Just a few comments from me

packages/tupaia-web/src/features/EntitySearch/EntityMenu.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const link = { ...location, pathname: `/${projectCode}/${entity.code}` };
+
+  return (
+    <div>
+      <FlexRow>
+        <MenuLink to={link}>
+          {entity.name} {entity.type === 'facility' && <HospitalIcon />}
+        </MenuLink>
+        <IconButton onClick={onExpand} disabled={parentIsLoading || !nextChildren}>
```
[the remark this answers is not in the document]
tcaiger: I thought about this and player around with it a bit. The button is really helpful to keep the spacing from glitching when the Icon button appears. The button is semantically disabled which helps.

packages/tupaia-web/src/features/EntitySearch/SearchBar.tsx
tcaiger: I initially had the onBlur event closing the entity menu but then I realised that the menu kept closing whenever I clicked on it and realised the blur was the problem so I don't think we want the blur for that interaction. Is there any thing else that you think should happen when you blur?

packages/tupaia-web/src/features/EntitySearch/EntityMenu.tsx
```diff
@@ -100,12 +99,21 @@ const EntityMenuItem = ({
         <MenuLink to={link}>
           {entity.name} {entity.type === 'facility' && <HospitalIcon />}
         </MenuLink>
-        <IconButton onClick={onExpand} disabled={parentIsLoading || !nextChildren}>
+        <IconButton
+          onClick={toggleExpanded}
+          disabled={parentIsLoading || !nextChildren}
+          aria-controls="entity-expand-button"
```
acdunham, line 105: Nice, thank you! Can we just add an `aria-label` of something like `toggle menu for ....` ?

packages/tupaia-web/src/features/EntitySearch/EntityMenu.tsx
```diff
@@ -100,12 +99,21 @@ const EntityMenuItem = ({
...
           <ExpandIcon />
         </IconButton>
       </FlexRow>
       {isExpanded && (
-        <EntityMenu children={nextChildren} projectCode={projectCode} isLoading={isLoading} />
+        <EntityMenu
+          children={nextChildren}
+          projectCode={projectCode}
+          isLoading={isLoading}
+          aria-expanded={isExpanded}
```
acdunham, line 115: Just checking - does this get carried down to the menu element?

acdunham (approved):
Looks good. Had a quick question on the aria stuff, but pre-approving.

packages/tupaia-web/src/features/EntitySearch/EntityMenu.tsx
tcaiger: ah very good point. I have moved it to the List element and it's always true there since the menu list only shows when expanded.

PR #4673, opened 2023-06-25
WAITP-1211: Dashboard Navigation

tcaiger:
### Changes:

- Add a redirect for routes that don't have a dashboard name included
- Handle when the dashboard name from the url is not in the dashboards data. This can happen for permission reasons or an incorrect url


packages/tupaia-web-server/src/routes/DashboardsRoute.ts
```diff
@@ -17,9 +17,10 @@ export class DashboardsRoute extends Route<DashboardsRequest> {
     const project = (
       await ctx.services.central.fetchResources('projects', {
         filter: { code: projectCode },
-        columns: JSON.stringify(['entity.code', 'entity_hierarchy.name']),
```
tcaiger, line 20: this is temporary code that Ethan is refactoring in another PR so I wouldn't be too worried about it

packages/tupaia-web/src/constants/constants.ts
```diff
@@ -3,9 +3,6 @@
  * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
-export const DEFAULT_PROJECT_CODE = 'explore';
-export const DEFAULT_ENTITY_CODE = 'explore';
-
```
tcaiger, line 8: these consts are only used in the default url so I'm just inlining them below as I think the concept of default project and entity doesn't make sense outside of the default url

packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -99,15 +99,27 @@ const Chart = styled.div`
...
+
+  let activeDashboard = null;
+
+  if (dashboards.length > 0) {
+    activeDashboard =
+      dashboards.find(dashboard => dashboard.name === dashboardName) || dashboards[0];
+  }
+
+  return { dashboards, activeDashboard };
+};
```
acdunham, line 114: Nice, I like this hook

acdunham (approved):
Nice, this is great. Thanks for doing this!

PR #4677, opened 2023-06-26
WAITP-1286 Map Overlays Route

EMcQ-BES:
### Changes:

- Add `MapOverlaysRoute` to `tupaia-web-server`
- Split out `legacyMapOverlayReport` route

The way we build nested map overlay groups is kind of weird and backwards honestly. All the access handling is attached to the `map_overlay` itself. The groups are just a series of relations, and we assume that you can access any group the map_overlay is in, and build up the hierarchy backwards from there.

I've got a working implementation here by fetching the overlay group relations in layers. This implementation probably uses more requests than entirely necessary. It might be worth while to add some functionality to the central server to fetch a full map overlay hierarchy


packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
```diff
@@ -0,0 +1,172 @@
...
+import { Route } from '@tupaia/server-boilerplate';
+import { MapOverlay, MapOverlayGroup, MapOverlayGroupRelation } from '@tupaia/types';
+import groupBy from 'lodash.groupby';
+import keyBy from 'lodash.keyby';
+
+// TODO: WAITP-1278 split request types to types package
+// (And actually define it)
+export type MapOverlaysRequest = Request<any, any, any, any>;
+
+// TODO: Can these be moved into types?
```
rohan-bes, line 16: Yup, I reckon so 👍

packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
```diff
@@ -0,0 +1,172 @@
...
+interface TranslatedMapOverlayGroup {
+  name: string;
+  children: OverlayChild[];
+}
+type OverlayChild = TranslatedMapOverlayGroup | TranslatedMapOverlay;
+
+export class MapOverlaysRoute extends Route<MapOverlaysRequest> {
+  public async buildResponse() {
+    const { params, ctx } = this.req;
+    const { projectCode, entityCode } = params;
```
rohan-bes, line 37: You make a great point about how we do the map overlay nesting backwards. It definitely makes this kind of route pretty confusing. And it's inconsistent with dashboard relations.

We took a stab at refactoring this [here](https://linear.app/bes/issue/RN-839/manage-permissions-and-projects-for-map-overlays-via-map-overlay) but at the time refactor didn't seem worth the effort (and honestly I just lost heart and got a bit distracted with more pressing matters).

It would be nice to finish the job really. Maybe once the epic branch gets merged we can look at picking it up in a maintenance sprint.

rohan-bes (approved):
Nice stuff, looks good 👍

PR #4683, opened 2023-06-26
RN-932: Fix KoBo sync

rohan-bes:
### Issue RN-932:

Bit of an unexpected bug here. Because we were spreading the sync group data source Db objects we were losing the `databaseType` getter.

Added a check for that since it's a bit of a hard bug to track down.

### DIFF
--- a/packages/central-server/src/kobo/startSyncWithKoBo.js
+++ b/packages/central-server/src/kobo/startSyncWithKoBo.js
@@ -122,7 +122,7 @@ export async function syncWithKoBo(models, dataBroker, syncGroupCode) {
 
     // Pull data from KoBo
     const koboData =
-      (await dataBroker.pullSyncGroupResults(syncGroupCode, {
+      (await dataBroker.pullSyncGroupResults([syncGroupCode], {
         startSubmissionTime: dataServiceSyncGroup.sync_cursor,
       })) || {};
 
--- a/packages/central-server/src/kobo/startSyncWithKoBo.js
+++ b/packages/central-server/src/kobo/startSyncWithKoBo.js
@@ -52,50 +52,48 @@ const writeKoboDataToTupaia = async (transactingModels, koboData, syncGroupCode)
 
   let newSyncTime = dataServiceSyncGroup.sync_cursor;
   let numberOfSurveyResponsesCreated = 0;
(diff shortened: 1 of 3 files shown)


packages/central-server/src/kobo/startSyncWithKoBo.js
```diff
@@ -52,50 +52,48 @@ const writeKoboDataToTupaia = async (transactingModels, koboData, syncGroupCode)
...
-          Object.entries(responseData.dataValues).map(([questionCode, answer]) => ({
-            id: generateId(),
-            type: questionByCode[questionCode].type,
-            survey_response_id: surveyResponse.id,
-            question_id: questionByCode[questionCode].id,
-            text: answer,
-          })),
-        );
+  for (const [surveyCode, responses] of Object.entries(koboData)) {
+    const survey = await transactingModels.survey.findOne({ code: surveyCode });
```
rohan-bes, line 56: Mostly just whitespace changes here, took out the top level of the loop

IgorNadj: Ah yikes what an ugly bug, good spotting and solution. 
I haven't looked too much into it but is this a type issue? The fact that that typescript didn't complain when the getter disappeared?

IgorNadj:
Ah yikes what an ugly bug, good spotting and solution.
I haven't looked too much into it but is this a type issue? The fact that that typescript didn't complain when the getter disappeared?

IgorNadj (approved):

rohan-bes:
> I haven't looked too much into it but is this a type issue? The fact that that typescript didn't complain when the getter disappeared?

@IgorNadj thanks for mentioning this. I had missed that we had a `ts-ignore` comment that avoided this typescript error appearing. Decided to remove the manual check and just allow ts to enforce it instead 👍

IgorNadj:
> > I haven't looked too much into it but is this a type issue? The fact that that typescript didn't complain when the getter disappeared?
> 
> @IgorNadj thanks for mentioning this. I had missed that we had a `ts-ignore` comment that avoided this typescript error appearing. Decided to remove the manual check and just allow ts to enforce it instead 👍

Good catch. I had a look through the history and it was something I introduced, sorry about that :( Not sure why it was like that, will be more careful in the future.

PR #4684, opened 2023-06-27
WAITP-1209: Map entities part 1

tcaiger:
### Changes:

- Example

---


packages/tupaia-web/src/features/EntitySearch/EntitySearch.tsx
```diff
@@ -33,33 +30,49 @@ const ResultsWrapper = styled.div`
...
   const { projectCode } = useParams();
   const { data: project } = useProject(projectCode!);
   const { data: entity, isLoading } = useEntities(projectCode!, project?.entityCode, {
     enabled: !!project?.entityCode,
   });
   const [searchValue, setSearchValue] = useState('');
-  const [isSearchFocused, setIsSearchFocused] = useState(false);
+  const [isOpen, setIsOpen] = useState(false);
+
+  const onClose = () => {
```
tcaiger, line 65: the mobile menu is not in scope for this ticket but I wanted to get it working a bit for testing mobile

packages/tupaia-web/src/features/Map/PolygonLayer.tsx
```diff
@@ -0,0 +1,9 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+// @ts-ignore
+export const PolygonLayer = ({ entity }) => {
+  return null;
```
tcaiger, line 8: This will be completed in part 2 of this ticket

packages/tupaia-web/src/api/queries/useEntities.ts
```diff
@@ -2,16 +2,27 @@
...
  */
-
-import { useQuery } from 'react-query';
+import { AxiosRequestConfig } from 'axios';
+import { useQuery, QueryObserverOptions, QueryOptions } from 'react-query';
+import { Entity } from '@tupaia/types';
 import { get } from '../api';
 
-export const useEntities = (projectCode: string, entityCode?: string, options?: any) => {
+type EntityResponse = Entity & { children?: Entity[] };
```
acdunham, line 10: This feels like a type that might be useful down the track, so could go in the `types` file. Not a biggie, just a thought

packages/tupaia-web/src/features/EntitySearch/EntitySearch.tsx
acdunham: That makes sense. When the time comes to do that, one thing to watch is whether `isMobile` gets called heaps of times or not when it is used like `if(isMobile())` or whether we need to do something to stop it re-rendering the component. I'm not sure if it will be an issue, but I was thinking about it when working on the chart components `isMobile` function

packages/tupaia-web/src/features/Map/Map.tsx
```diff
@@ -63,16 +66,24 @@ const TilePickerWrapper = styled.div`
   }
 `;
 
+const ENTITY_FIELDS = ['parent_code', 'code', 'name', 'type', 'bounds', 'region'];
```
acdunham, line 69: Technically, this doesn't need to be it's own variable because it's only used in one place, but I also understand that it feels tidier. Just a thought.

packages/tupaia-web/src/views/ProjectPage/ProjectPage.tsx
```diff
@@ -1,19 +1,44 @@
...
+import { EnvBanner } from '@tupaia/ui-components';
+import { TopBar } from '../../layout';
+import { MOBILE_BREAKPOINT } from '../../constants';
 import { MobileTabLayout } from './MobileTabLayout';
+import { DesktopLayout } from './DesktopLayout';
 
 /**
- * This is the layout for the project/* view. This contains the map and the sidebar, as well as any overlays that are not auth overlays (i.e. not needed in landing pages)
+ * This is the layout for the entire app, which contains the top bar and the main content. This is used to wrap the entire app content
  */
```
acdunham, line 16: I'm slightly tempted to rename this file to something else, but I'm also not sure. Just because `ProjectPage` implies the `projectCode` route, since that's the sort of pattern we have going, and it's not strictly that route, if that makes sense. But I am also on the fence about this one.

acdunham (approved):
This looks good. Pre-approving it, but added a couple of questions. I also wonder if we need to do any layout stuff to the landing pages for the top bar etc? Or is that to come?

packages/tupaia-web/src/api/queries/useEntities.ts
tcaiger: Well Ethan has a ticket to generate all the response types for the orchestration server so I'm hoping this one can just be swapped out when that is ready.

packages/tupaia-web/src/features/EntitySearch/EntitySearch.tsx
tcaiger: Good point - let's do some testing when we do the mobile menu ticket

packages/tupaia-web/src/features/Map/Map.tsx
tcaiger: yes that's right I just pulled it out for readability. I feel like it might end up being a shared const soon or I might create a couple of versions of the useEntities hook.

packages/tupaia-web/src/views/ProjectPage/ProjectPage.tsx
tcaiger: I know what you mean. I actually just added back the MainLayout and wrapped that around the routes at the top level and so now I think it makes a bit more sense to be called project page again.

PR #4685, opened 2023-06-28
Waitp 1224 overlay selector desktop

acdunham:
### Changes:
- Updated `DesktopMapOverlaySelector` to be fully functional
- Added `MapOverlayList` for use across mobile and desktop overlay selectors
- Added new `useMapOverlays` hook to `utils` that will also return useful map overlay associated info
---

### Screenshots:
No overlays available
Screenshot 2023-06-28 at 2 44 23 pm

Overlays available and selected
Screenshot 2023-06-28 at 2 45 21 pm


packages/tupaia-web/src/api/queries/useFetchMapOverlays.ts
```diff
@@ -0,0 +1,20 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { useQuery } from 'react-query';
+import { get } from '../api';
+import { EntityCode, ProjectCode } from '../../types';
+
+export const useFetchMapOverlays = (projectCode?: ProjectCode, entityCode?: EntityCode) => {
```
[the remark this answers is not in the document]
tcaiger: I think it would be good to just move all the code from packages/tupaia-web/src/utils/useMapOverlays.ts into this file since it is just derived api data with the exception of `updateSelectedMapOverlay`. You could then put updateSelectedMapOverlay inline in  MapOverlayList, make it a stand alone util or keep it with useMapOverlays if you really want.

packages/tupaia-web/src/api/queries/useMapOverlayData.ts
```diff
@@ -0,0 +1,28 @@
...
+export const useMapOverlayData = (
+  projectCode?: ProjectCode,
+  entityCode?: EntityCode,
+  mapOverlayCode?: SingleMapOverlayItem['code'],
+) => {
+  return useQuery(
+    ['mapOverlayData', projectCode, entityCode, mapOverlayCode],
+    async () => {
+      return get(
+        `measureData?mapOverlayCode=${mapOverlayCode}&organisationUnitCode=${entityCode}&projectCode=${projectCode}&shouldShowAllParentCountryResults=${
```
tcaiger, line 19: it would be great to use the axios params field here for readability and ease of maintenance. 
Also Ethan has re-done the endpoints for the map overlays so this should use `legacyMapOverlayReport` or `report` depending on whether or not it is legacy.

packages/tupaia-web/src/features/Map/MapLegend/DesktopMapLegend.tsx
```diff
@@ -13,19 +13,24 @@ const Wrapper = styled.div`
...
-  border-radius: 5px;
+  width: 100%;
+  padding: 1rem;
   @media screen and (max-width: ${MOBILE_BREAKPOINT}) {
     display: none;
   }
 `;
 
+const Legend = styled.div`
+  height: 50px;
```
tcaiger, line 24: probably should be rems, right?

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -5,23 +5,212 @@
...
 import React from 'react';
 import styled from 'styled-components';
+import { Accordion, Typography, AccordionSummary, AccordionDetails } from '@material-ui/core';
+import { ExpandMore, Layers } from '@material-ui/icons';
+import { Skeleton as MuiSkeleton } from '@material-ui/lab';
 import { MOBILE_BREAKPOINT } from '../../../constants';
+import { Entity } from '../../../types';
+import { useMapOverlayData } from '../../../api/queries';
+import { useParams } from 'react-router';
+import { periodToMoment } from '@tupaia/utils';
```
tcaiger, line 15: just check order of imports here

packages/tupaia-web/src/utils/useMapOverlays.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+import { ChangeEvent } from 'react';
+import { useSearchParams, useParams } from 'react-router-dom';
+import { UseQueryResult } from 'react-query';
+import { useFetchMapOverlays } from '../api/queries';
+import { URL_SEARCH_PARAMS } from '../constants';
+import { MapOverlayGroup, SingleMapOverlayItem } from '../types';
+
+const flattenMapOverlays = (mapOverlayGroups: MapOverlayGroup[] = []): SingleMapOverlayItem[] => {
```
tcaiger, line 13: Could we make this `mapOverlaysByCode` and have it generate a dictionary of the map overlays like below. Then we could just lookup the result below instead of finding it. eg `mapOverlaysByCode[selectedOverlayCode]`
``` 
{ 
  123: {...},
  128: {...}
}

tcaiger (changes requested):
looks good. I left a couple of suggestions. Happy to discuss anything if you like.

packages/tupaia-web/src/api/queries/useMapOverlayData.ts
acdunham: Is that endpoint for fetching the measure data, or for something else?

packages/tupaia-web/src/features/Map/MapLegend/DesktopMapLegend.tsx
acdunham: It should, but this is just a placeholder so I don't think it's important at this stage

packages/tupaia-web/src/api/queries/useMapOverlayData.ts
tcaiger: yes that's right. for measure data

packages/tupaia-web/src/features/Map/MapLegend/DesktopMapLegend.tsx
tcaiger: oh true. I just see px and I jump 😆

packages/tupaia-web/src/api/queries/useMapOverlayData.ts
acdunham: I'm currently getting an error using this, and it seems to be different to `measureData` which is for map overlays (is that right?) and reports which I was using for dashboard report data. I'll add this in as part of part 2

tcaiger (approved):

packages/tupaia-web/src/api/queries/useMapOverlayData.ts
tcaiger: oh yeah it looks like there is a bug in the endpoint so yes that's leave this as is for now

PR #4686, opened 2023-06-28
WAITP-1222: Marker layer

tcaiger:
### Changes:

- Restructure the map controls to stop interferring with the map
- Fix the useOnChangeMapOverlay util
- Add MarkerLayer

---

### Screenshots:

image

image


packages/tupaia-web/src/api/queries/index.ts
```diff
@@ -14,5 +14,5 @@ export { useEntities, useEntitiesWithLocation } from './useEntities';
 export { useEmailVerification } from './useEmailVerification';
 export { useDashboards } from './useDashboards';
 export { useReport } from './useReport';
+export { useMapOverlayReport } from './useMapOverlayReport';
```
tcaiger, line 17: renamed report to make it a bit more consistent with the endpoint

packages/tupaia-web/src/features/Map/Map.tsx
```diff
@@ -99,23 +102,25 @@ export const Map = () => {
...
-              tileSets={TILE_SETS}
-              activeTileSet={activeTileSet}
-              onChange={onTileSetChange}
-            />
-          </TilePickerWrapper>
-        </MapControlWrapper>
         <MapWatermark />
       </StyledMap>
+      {/* Map Controls need to be outside the map so that the mouse events on controls don't inter wit the map */}
+      <MapControlWrapper>
```
tcaiger, line 111: restructure the controls to be outside the map to fix those scroll bugs that were happenning.

packages/tupaia-web/src/features/Map/MarkerLayer/MarkerLayer.tsx
```diff
@@ -0,0 +1,101 @@
...
+        },
+      },
+    },
+    { enabled: !!measureLevel },
+  );
+};
+
+export const MarkerLayer = () => {
+  // set the map default overlay if there isn't one selected
+  useDefaultMapOverlay();
```
tcaiger, line 67: This effect could go anywhere and I'm not really sure where it should go. I put it here so that it's close to where the report data is used.

packages/ui-map-components/src/components/Markers/MeasureMarker.tsx
```diff
@@ -11,7 +11,7 @@ import { MarkerProps } from '../../types';
 export const MeasureMarker = React.memo((props: MarkerProps) => {
   const { icon, radius = 0 } = props;
 
-  if (radius && parseInt(String(radius), 10) === 0) {
+  if (radius !== undefined && parseInt(String(radius), 10) === 0) {
```
tcaiger, line 14: If radius was 0 this was not doing what we expect

packages/tupaia-web/src/api/queries/useEntities.ts
```diff
@@ -7,21 +7,25 @@ import { useQuery, QueryObserverOptions } from 'react-query';
...
+    enabled = enabled && queryOptions.enabled;
+  }
 
   return useQuery(
-    ['entities', projectCode, entityCode, axiosConfig],
-    (): Promise<EntityResponse[]> =>
-      get(`entities/${projectCode}/${entityCode}`, {
+    ['entities', projectCode, entityCode, axiosConfig, queryOptions],
+    async (): Promise<EntitiesResponse> => {
+      return get(`entities/${projectCode}/${entityCode}`, {
```
acdunham, line 27: Quick question - is there a particular reason for adding `async` and `return` here?

packages/tupaia-web/src/api/queries/useEntities.ts
```diff
@@ -34,29 +38,40 @@ export const useEntities = (
...
+  axiosConfig?: AxiosRequestConfig,
+  queryOptions?: QueryObserverOptions,
+) =>
+  useEntities(
+    projectCode,
+    entityCode,
+    {
+      params: {
+        ...{ ...axiosConfig?.params },
+        fields: [
```
acdunham, line 61: This is nice. Can you think of any use case where we may want to overwrite those fields?

packages/tupaia-web/src/api/queries/useMapOverlayReport.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  return {
+    measureCode,
+    measureLevel,
+    measureOptions,
+    serieses: measureOptions,
+    measureData: measureDataResponse,
+  };
+};
+
```
acdunham, line 37: I don't see anything wrong with this, but I guess it opens up the question, should we be doing this on the frontend, or should the endpoint be returning data the correct way? Or is this because dashboard reports and map overlay reports are both using the `report` endpoint?

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -49,6 +49,7 @@ const Heading = styled(Typography).attrs({
 
 const Container = styled(MaxHeightContainer)`
   border-radius: 0 0 5px 5px;
+  pointer-events: auto;
```
acdunham, line 52: This is super pedantic, but I wonder if `pointer-events: auto` should go in `MaxHeightContainer` since it's used as the base for both `Wrapper` and `Container` (assuming it isn't used anywhere else in the file)

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -49,6 +49,7 @@ const Heading = styled(Typography).attrs({
 
 const Container = styled(MaxHeightContainer)`
   border-radius: 0 0 5px 5px;
```
acdunham, line 51: I wonder if we should save this border radius value somewhere (just the `5px`) since it gets used a lot. I know not strictly part of this PR, but just something I thought of

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapOverlayList.tsx
```diff
@@ -86,27 +88,45 @@ const MapOverlayAccordion = ({ mapOverlayGroup }: { mapOverlayGroup: MapOverlayG
...
+  const [urlSearchParams, setUrlParams] = useSearchParams();
+  const onChangeMapOverlay = (e: ChangeEvent<HTMLInputElement>) => {
+    urlSearchParams.set(URL_SEARCH_PARAMS.MAP_OVERLAY, e.target.value);
+    // when overlay changes, reset period to default
+    urlSearchParams.set(URL_SEARCH_PARAMS.MAP_OVERLAY_PERIOD, DEFAULT_PERIOD_PARAM_STRING);
+    setUrlParams(urlSearchParams);
+  };
+
+  return { onChangeMapOverlay };
+};
```
acdunham, line 114: I'm guessing this was moved to this file to be more contained in where it's used? In that case, I'm on the fence about whether it should be a hook or just a function inside the component

packages/tupaia-web/src/features/Map/MarkerLayer/MarkerLayer.tsx
acdunham: I agree it could go anywhere. I almost feel like it should be built into the links, but then it's possible a link could be navigated to in some way without a default overlay code being in it (e.g directly entering the url). Is `useDefaultMapOverlay` used anywhere else? Could it possibly be contained in the map component?

packages/tupaia-web/src/features/Map/MarkerLayer/MarkerLayer.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const { data: mapOverlayData } = useMapOverlayReport(projectCode, entityCode, selectedOverlay);
+  const { data: entity } = useEntity(entityCode);
+
+  if (!entitiesData || !mapOverlayData || !entity) {
+    return null;
+  }
+
+  // todo: move to mapOverlays route
+  const displayOnLevel = mapOverlayData.serieses.find((series: any) => series.displayOnLevel);
+  if (displayOnLevel && camelCase(entity.type!) !== camelCase(displayOnLevel.displayOnLevel)) {
```
acdunham, line 82: This is super pedantic, but I personally don't love it when variables are named the same as a property inside them.

packages/tupaia-web/src/features/Map/MarkerLayer/MarkerLayer.tsx
```diff
@@ -0,0 +1,101 @@
...
+    return null;
+  }
+
+  const processedMeasureData = processMeasureData({
+    entitiesData,
+    measureData: mapOverlayData.measureData,
+    serieses: mapOverlayData.serieses,
+    hiddenValues: {},
```
acdunham, line 90: Do we need to pass in something for `hiddenValues` from the fetched data at all?

packages/tupaia-web/src/features/Map/MarkerLayer/processMeasureData.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+import { calculateRadiusScaleFactor, getMeasureDisplayInfo } from '@tupaia/ui-map-components';
+import { Entity } from '@tupaia/types';
+
+interface processMeasureData {
+  measureData: any;
+  entitiesData: any;
+  serieses: any;
+  hiddenValues: any;
+}
```
acdunham, line 13: I wonder if we should add types for these?

packages/tupaia-web/src/features/Map/MarkerLayer/useDefaultMapOverlay.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!selectedMapOverlay || !isValidMapOverlayId) {
+      urlSearchParams.set(URL_SEARCH_PARAMS.MAP_OVERLAY, defaultMapOverlayId);
+    }
+
+    if (!selectedMapOverlayPeriod) {
+      urlSearchParams.set(URL_SEARCH_PARAMS.MAP_OVERLAY_PERIOD, DEFAULT_PERIOD_PARAM_STRING);
+    }
+
+    setUrlParams(urlSearchParams);
+  }, [JSON.stringify(mapOverlayGroups)]);
```
acdunham, line 41: I know it's probably pretty self explanatory what's going on here, but I wonder if we should perhaps add some more comments to this and anything else surrounding the map overlays in here, just for the sake of any new devs that come into the codebase, and because the map stuff can be quite complex

acdunham (changes requested):
This is a massive amount of effort, nice work! I have a few questions but in general it's looking good, and awesome to be hitting this point in the project!

packages/tupaia-web/src/api/queries/useEntities.ts
tcaiger: just debugging

packages/tupaia-web/src/api/queries/useEntities.ts
tcaiger: No I don't think so and I kind of don't want to because the less variations of params there are the more caching there is

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
tcaiger: I actually had the pointer events set on the wrapper initially but it was blocking map interactions when the menu was closed so I moved it down. I've added a comment.

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapOverlayList.tsx
tcaiger: yes it was because I think it will only be used in the one place and yes I think that you are right to move it in to the component.

packages/tupaia-web/src/features/Map/MarkerLayer/MarkerLayer.tsx
tcaiger: Yes agreed. I've changed it and added a comment.

packages/tupaia-web/src/features/Map/MarkerLayer/MarkerLayer.tsx
tcaiger: The hidden values need to be implemented as part of the legend component as that is where they are set. I've added a comment.

packages/tupaia-web/src/api/queries/useMapOverlayReport.ts
tcaiger: Yes good question. I just had a chat to Ethan about this and he said that this could be done on the backend but it would be messy because we use the report route is used fro both the map overlays and the dashboard items.

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
tcaiger: hmmm we could do but I wonder if it would actually be helpful. I don't think that is something that would change in a theme for example.

packages/tupaia-web/src/features/Map/MarkerLayer/processMeasureData.ts
tcaiger: yes agreed. I'm just working on exporting the types that I need out of ui-map-components

packages/tupaia-web/src/features/Map/useDefaultMapOverlay.ts
```diff
@@ -0,0 +1,45 @@
...
+import { useProject } from '../../api/queries';
+import {
+  DEFAULT_MAP_OVERLAY_ID,
+  DEFAULT_PERIOD_PARAM_STRING,
+  URL_SEARCH_PARAMS,
+} from '../../constants';
+import { MapOverlayGroup, ProjectCode, EntityCode } from '../../types';
+
+// When the map overlay groups change, update the default map overlay
+export const useDefaultMapOverlay = (
```
tcaiger, line 17: I ended up refactoring this a little to try and make it clearer and moved it to the map component so it's easier to find.

packages/tupaia-web/src/features/Map/MarkerLayer/processMeasureData.ts
```diff
@@ -2,29 +2,35 @@
...
+  Series,
+} from '@tupaia/ui-map-components';
 import { Entity } from '@tupaia/types';
 
-interface processMeasureData {
-  measureData: any;
-  entitiesData: any;
-  serieses: any;
-  hiddenValues: any;
+interface processMeasureDataProps {
```
acdunham, line 14: /
```suggestion
interface ProcessMeasureDataProps {
```

packages/tupaia-web/src/features/Map/MarkerLayer/processMeasureData.ts
```diff
@@ -2,29 +2,35 @@
...
+  serieses: Series[];
+  hiddenValues: LegendProps['hiddenValues'];
 }
 export const processMeasureData = ({
   measureData,
   entitiesData,
   serieses,
   hiddenValues,
-}: processMeasureData) => {
+}: processMeasureDataProps) => {
```
acdunham, line 25:
```suggestion
}: ProcessMeasureDataProps) => {
```

acdunham (approved):

acdunham:
Nice work, thanks for doing all the types in this too!

PR #4688, opened 2023-06-28
WAITP-1293 Update filtering on entities route

EMcQ-BES:
### Changes:

- `EntitiesRoute` now takes `projectCode` and does a lookup for `hierarchyName`
- Allow overwriting the default `generational_distance` filter easier


tcaiger (approved):
thanks

PR #4689, opened 2023-06-28
WAITP-1268: Update map overlay route

tcaiger:
### Changes:
- Pass mapOverlay param to query string params instead of report report route param in legacy map overlay report route 
- Handle when there are no map overlays for a project and entity on the map overlays route


EMcQ-BES (approved):
👍  Whoops, Ethan should look closer at how legacy routes work 😅

PR #4690, opened 2023-06-29
WAITP-1298 Add EntityAncestors route

EMcQ-BES:
### Changes:

- Add `entityAncestors` route to `tupaia-web-server`
- Returns an array of entities in ascending hierarchy order `[parent, grandparent, great-grandparent]`
  - This is the default return from `entity-server` so it's mostly just forwarding the request


tcaiger (approved):
LGTM

PR #4691, opened 2023-06-29
RN-934 duplicate filenames

IgorNadj:
### Issue #: RN-934

Screen recording: https://linear.app/bes/issue/RN-934#comment-8fd8b1c7

### DIFF
--- a/packages/central-server/src/apiV2/DownloadFiles.js
+++ b/packages/central-server/src/apiV2/DownloadFiles.js
@@ -9,6 +9,8 @@ import {
   ValidationError,
   respondWithDownload,
   writeStreamToFile,
+  getUniqueFileNameParts,
+  getDeDuplicatedFileName,
 } from '@tupaia/utils';
 import { RouteHandler } from './RouteHandler';
 import { getTempDirectory } from '../utilities';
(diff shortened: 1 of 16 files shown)


packages/web-frontend/src/components/View/DownloadFilesVisual.js
```diff
@@ -19,22 +20,34 @@ const StyledDownloadFilesVisual = styled(BaseDownloadFilesVisual)`
...
-      const fileName = files.length === 1 ? getFilenameFromPath(files[0]) : getZipFileName();
-      await download(`downloadFiles?files=${filesString}`, undefined, {}, fileName);
+      if (uniqueFileNames.length === 1) {
+        const { fileName } = getUniqueFileNameParts(uniqueFileNames[0]);
+        await download(
+          `downloadFiles?uniqueFileNames=${uniqueFileNames[0]}`,
+          undefined,
+          {},
+          fileName,
+        );
```
rohan-bes, line 35: Ideally we wouldn't have to do this, and we could just rely on the filename that came back from the server. I'd kind of prefer that since it means this viz is agnostic to the whole `uniqueFilename` concept. It does come back in the `content-disposition` header I think. But at the time I didn't see an easy way to access it.

Can I ask you to take a quick look to see if it's possible?

Also, if it is possible, then I think we can do the same for the zip, if we pass in the `zipFileName` query param

packages/utils/src/__tests__/filesystem.test.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import { getDeDuplicatedFileName, getUniqueFileNameParts } from '../filesystem';
+
+describe('filesystem', () => {
```
rohan-bes, line 8: I ❤️ tests

packages/ui-components/src/components/DownloadFilesVisual/DownloadFilesVisual.js
```diff
@@ -52,40 +52,48 @@ export const DownloadFilesVisual = ({
...
-  const fileNames = data.map(({ name }) => name);
-  const noneSelected = Object.fromEntries(fileNames.map(fileName => [fileName, false]));
+
+  // This mapping does nothing, just commenting some typing for future conversion to ts
+  const options = data.map(({ uniqueFileName, label }) => ({
+    uniqueFileName, // string e.g. 5da02ed278d10e8695530688_report.pdf
+    label, // string e.g. 'Instruction Manual' or 'report.pdf'
+  }));
+
+  console.log('options', options);
```
rohan-bes, line 62:
```suggestion
```

packages/meditrak-app/.env.example
```diff
@@ -1,4 +1,11 @@
+# E.g. dev
 BETA_BRANCH=
+
+# For Local dev can specify local central server
+# e.g. http://localhost:8090/v2
+# Tip: Android Emulator uses 10.0.2.2 for host -> http://10.0.2.2:8090/v2
+CENTRAL_API_URL=
```
rohan-bes, line 7: Oh this handy 👍

packages/meditrak-app/app/Meditrak.js
```diff
@@ -44,7 +44,7 @@ class MeditrakContainer extends React.Component {
   renderBetaBanner() {
     return (
       <View style={localStyles.betaBanner} pointerEvents="none">
-        <Text style={localStyles.betaBannerText}>{betaBranch.toUpperCase()}</Text>
+        <Text style={localStyles.betaBannerText}>{betaBranch.toUpperCase() || centralApiUrl}</Text>
```
rohan-bes, line 47:
```suggestion
        <Text style={localStyles.betaBannerText}>{centralApiUrl || betaBranch.toUpperCase()}</Text>
```
To consider, I'm not sure honestly. But part of me feels the banner is to help us know which server we're pointing to (so that we don't submit data un-intentionally).

Perhaps even:
```suggestion
        <Text style={localStyles.betaBannerText}>{`${betaBranch.toUpperCase()}${centralApiUrl ? ` -> ${centralApiUrl}` : ''}</Text>
```

All very optional btw, just some thoughts I had.

rohan-bes (approved):
Looks good! Just a request to see if we can control the download file names based on what the server returns, but if that's not possible then happy with all this 👍

packages/central-server/src/apiV2/DownloadFiles.js
```diff
@@ -20,34 +22,44 @@ export class DownloadFiles extends RouteHandler {
...
+    );
 
-    // Unique directory to avoid clashes when doing multiple downloads simultaneously
-    const tempDir = getTempDirectory(`downloads/${userId}/${Date.now()}`);
+    const tempDir = getTempDirectory(`downloads/`);
 
-    const writeFileToTempDir = async (fileName, file) => {
-      const filePath = `${tempDir}/${fileName}`;
+    const writeFileToTempDir = async (uniqueFilename, file) => {
+      const filePath = `${tempDir}/${uniqueFilename}`;
```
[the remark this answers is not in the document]
IgorNadj: I like it, it's a specific name for a specific concept :) If it was just fileName it would be super easy to get it mixed up.

packages/web-frontend/src/components/View/DownloadFilesVisual.js
IgorNadj: Yeah that confused me a lot when I came across it. I decided to leave it as is to keep changes to the frontend small for the rewrite. Definitely something that would be nice to clean up in the future though.

packages/meditrak-app/app/Meditrak.js
IgorNadj: Ah yeah you only really need to use one or the other. BETA_BRANCH is nice for non-technical people to get apk builds up, CENTRAL_API_URL is only something we would use on local. So the banner shows for both cases. Here's my local:
Screen Shot 2023-07-04 at 1 30 29 pm

Is it confusing having both? Maybe needs a little more comments?

rohan-bes: Oh I see. I think it's okay then, the comment you left in .env.example is clear enough

packages/web-frontend/src/components/View/DownloadFilesVisual.js
rohan-bes: I suspect we'd use a different method (and probably different library) to do this post frontend re-write, so yeah happy to hold off til then as well

PR #4692, opened 2023-06-29
WAITP-1293: Refactor tupaia-web for flat entities data

tcaiger:
### Changes:
The context of this refactor is that in [waitp-1293-entity-route-better-filtering](https://github.com/beyondessential/tupaia/tree/waitp-1293-entity-route-better-filtering) the response from the entities api endpoint was refactored to return a flat list of entities and so the front end has to be refactored to match.

- Entity Menu
- Entity Polygons

### DIFF
--- a/packages/tupaia-web/src/api/queries/useEntity.ts
+++ b/packages/tupaia-web/src/api/queries/useEntity.ts
@@ -5,7 +5,7 @@
 
 import { useQuery } from 'react-query';
 import { useParams } from 'react-router-dom';
-import { Entity } from '../../types';
+import { EntityResponse } from '../../types';
 import { get } from '../api';
 import { DEFAULT_BOUNDS } from '../../constants';
 
@@ -15,7 +15,7 @@ export const useEntity = (entityCode?: string) => {
 
   return useQuery(
     ['entities', projectCode, entityCode],
-    async (): Promise<Entity> => {
+    async (): Promise<EntityResponse> => {
       const entities = await get(`entities/${projectCode}/${entityCode}`, {
         params: {
           includeRoot: true,
--- a/packages/tupaia-web/src/api/queries/useEntity.ts
+++ b/packages/tupaia-web/src/api/queries/useEntity.ts
@@ -4,34 +4,32 @@
  */
 
 import { useQuery } from 'react-query';
-import { sleep } from '@tupaia/utils';
-
-const response = {
-  type: 'Country',
-  organisationUnitCode: 'TO',
-  countryCode: 'TO',
-  name: 'Tonga',
-  location: {
-    type: 'area',
-    point: null,
-    bounds: [
-      [-24.1625706, 180.604802],
-      [-15.3655722, 186.4704542],
-    ],
-    region: [],
-  },
-  photoUrl:
-    'https://tupaia.s3-ap-southeast-2.amazonaws.com/uploads/images/1499489588784_647646.png',
-  countryHierarchy: [],
-};
+import { useParams } from 'react-router-dom';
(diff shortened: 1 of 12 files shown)


packages/tupaia-web/src/api/queries/useEntity.ts
```diff
@@ -4,34 +4,32 @@
...
-  countryHierarchy: [],
-};
+import { useParams } from 'react-router-dom';
+import { Entity } from '../../types';
+import { get } from '../api';
+import { DEFAULT_BOUNDS } from '../../constants';
 
 export const useEntity = (entityCode?: string) => {
+  //  Todo: use entity endpoint when it's done and remove project code
+  const { projectCode } = useParams();
```
tcaiger, line 14: There is a ticket coming soon for a single entity route but this is useful for testing.

packages/tupaia-web/src/features/MapOverlays/PolygonLayer.tsx
```diff
@@ -23,22 +24,30 @@ const ChildEntities = ({ entities }: { entities: EntityResponse['children'] }) =
   );
 };
 
-const SiblingEntities = ({ entity }: { entity: EntityResponse }) => {
+const SiblingEntities = ({
+  parentEntityCode,
+  activeEntityCode,
+}: {
+  parentEntityCode: Entity['code'];
+  activeEntityCode: Entity['code'];
```
acdunham, line 32: I believe we have a helper type of `EntityCode` which you could also use here if you wanted, but it's not imperative.

I also tend to like to go by the rule of 'if a component has more than 1 prop, create an interface for it' so if you wanted to, you could add one here.

packages/tupaia-web/src/features/MapOverlays/PolygonLayer.tsx
```diff
@@ -63,16 +72,32 @@ const ActiveEntity = ({ entity }: { entity: EntityResponse }) => {
...
 
-export const PolygonLayer = ({ entityData }: { entityData: EntityResponse }) => {
-  if (!entityData) {
+export const PolygonLayer = ({
+  entities,
+  entityCode,
+}: {
+  entities: EntityResponse[];
+  entityCode: string;
+}) => {
```
acdunham, line 81: Same here, you could maybe add an interface for this one

packages/tupaia-web/src/features/MapOverlays/PolygonLayer.tsx
```diff
@@ -23,22 +24,30 @@ const ChildEntities = ({ entities }: { entities: EntityResponse['children'] }) =
...
   );
 
-  if (isLoading || !siblingEntities) {
+  if (isLoading || !siblingEntities || siblingEntities.length === 0) {
     return null;
   }
 
-  const children = siblingEntities?.children?.filter(e => e.code !== entity.code) || [];
+  const children = siblingEntities.filter(
+    (entity: EntityResponse) => entity.code !== activeEntityCode,
```
acdunham, line 45: I wonder if the types need to be added here, or if having them in the `useEntitiesWithLocation` would solve any TS errors with that? Might save us writing it out each time

acdunham (approved):
Looks good!  A couple of questions/comments, but otherwise pre-approving

packages/tupaia-web/src/features/MapOverlays/PolygonLayer.tsx
tcaiger: oh yes you are correct. I didn't need them here. I must have added them in the wrong order. thanks

PR #4693, opened 2023-06-29
Waitp 1252 datepicker

acdunham:
### Changes:
- Add `DateRangePicker` component to `tupaia-web`
- Style to make it fit in with `tupaia-web` styles
- Add `stories` for `tupaia-web`
- Add `GlobalStyle` to `tupaia-web` for higher level styles
- Update `ui-components` `DateRangePicker` to reset dates on cancel, and change `Label` to be text, not a button as it doesn't do anything

#### Day
Screenshot 2023-06-30 at 10 17 21 am

#### Single day
Screenshot 2023-06-30 at 10 17 32 am

#### Week
Screenshot 2023-06-30 at 10 18 02 am

#### Single week
Screenshot 2023-06-30 at 10 18 12 am

#### Month
Screenshot 2023-06-30 at 10 18 20 am

#### Single Month
Screenshot 2023-06-30 at 10 18 30 am

#### Quarter
Screenshot 2023-06-30 at 10 18 41 am

#### Single Quarter
Screenshot 2023-06-30 at 10 19 01 am

#### Year
Screenshot 2023-06-30 at 10 19 14 am

#### Single year
Screenshot 2023-06-30 at 10 19 26 am


packages/ui-components/src/components/DateRangePicker/DateRangePicker.tsx
```diff
@@ -55,11 +56,22 @@ const Button = styled(MuiButton)`
...
-  pointer-events: none;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: ${props => props.theme.palette.text.secondary};
+  font-size: 1rem;
+  background-color: white;
+  border-left-width: 2px;
+  cursor: auto;
+  margin-left: 0 !important;
```
tcaiger, line 71: Since this is a ui-component it would be nice to not use !important if possible. What is the important over writing? Can we do this with some mui- selectors instead?

packages/tupaia-web/src/theme/GlobalStyle.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Theme } from '@material-ui/core';
+import { createGlobalStyle } from 'styled-components';
+
+/**
+ * These are global styles used in the app. They are overrides that might not otherwise be possible to control at a component level, e.g. the datepicker dialog, as this actually appears outside of the component in the DOM
+ */
+export const GlobalStyle = createGlobalStyle<{
+  theme: Theme;
+}>` 
+  #date-picker-dialog {
```
tcaiger, line 15: I would prefer to keep these styles in the component file of DateRangePicker. I assume that you added them here as you weren't able to style the modal in the DateRangePicker. It is actually possible using a a round about way which is to make a styled Paper component and pass that into the Mui Dialog component. Have a look at tupaia-web -> Modal.tsx for an example. We will also have to update the ui-components DatePicker Dialog to pass through the Paper component but I think that's still a good approach.

tcaiger (changes requested):
Overall this looks great. All the styling and customisations look really good. I'm pleased that we went with the css approach over making a new copy of this. I just have one real requested change to replace the GlobalStyles with a custom Paper component instead. Let me know if you have any questions.

tcaiger (approved):
Great changes!

PR #4694, opened 2023-06-29
Waitp 1247 map overlays date picker

acdunham:
### Changes:
- Implemented date picker for map overlay selector
- Updated `useMapOverlayData` to query with start and end dates, and handle legacy versus modern overlays
- Implemented date picker for charts
- Rearranged enlarged dashboard report to only have 1 modal that finds the data for the selected report
- Added a util `useDateRanges` that handles the dates in the URL string and charts and maps

Note: some of these changes are from `WAITP-1252`, which I merged into this one so it would work, with the `DateRangePicker`


packages/tupaia-web/src/api/queries/useDashboards.ts
```diff
@@ -4,15 +4,31 @@
...
     (): Promise<DashboardsResponse[]> =>
       get('dashboards', {
         params: { entityCode, projectCode },
       }),
     { enabled: !!entityCode && !!projectCode },
   );
+
+  let activeDashboard = null;
+
+  if (dashboards.length > 0) {
```
tcaiger, line 26:
```suggestion
  if (dashboards.length > 0 && dashboard) {
```

tcaiger: There might be places that we just want the dashboards but there isn't a dashboard name in the url yet

packages/tupaia-web/src/api/queries/useMapOverlays.ts
```diff
@@ -62,6 +64,8 @@ export const useMapOverlays = (): UseMapOverlaysResult => {
 
   const updateSelectedMapOverlay = (e: ChangeEvent<HTMLInputElement>) => {
```
tcaiger, line 65: This is a bit pedantic and academic but I reckon that the updateSelectedMapOverlay should probably be moved out of this hook and into a util as well to keep the data fetching a bit cleaner.

packages/tupaia-web/src/features/DashboardItem/EnlargedDashboardItem.tsx
```diff
@@ -23,28 +30,114 @@ const Container = styled.div`
...
+  variant: 'h2',
+})`
+  font-size: 1.25rem;
+  font-weight: ${({ theme }) => theme.typography.fontWeightMedium};
+  text-align: center;
+  margin: 0.3rem 0 1rem 0;
+  line-height: 1.4;
+`;
+
+const TitleWrapper = styled.div`
```
tcaiger, line 48: What do you think about making a little util component for this called FlexRow or something?

packages/types/src/schemas/schemas.ts
```diff
@@ -9381,7 +9381,7 @@ export const MultiValueViewConfigSchema = {
 	]
 } 
 
-export const DashboardItemConfigSchema = {
+export const DashboardReportConfigSchema = {
```
tcaiger, line 9384: Did these changes come from the dev branch? If not, won't they break code in other packages?

packages/tupaia-web/src/utils/useDateRanges.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+  // We rely on dates being rounded in state for range formats
+  const momentEndDate = moment(endDate, urlFormat);
+  return roundStartEndDates(granularity, momentStartDate, momentEndDate);
+};
+
+/**
+ * This handles the logic for setting and getting the dates for the dates with map overlays and reports, including updating the URLSearchParams
+ */
+export const useDateRanges = (
```
tcaiger, line 78: This hook is looking great. To me it's a much more easy to understand place to keep the logic for the date pickers. The code is really dense but there is no easy way to avoid that, right. I assume most of it was copied across from web-frontend which is cool. I was just about to ask if you could copy across and update the dateRangePicker tests but then I noticed that there are some in ui-components for the dateRangePicker so I think that should be fine.

tcaiger (changes requested):
Looking really good. The date range pickers are very complex so it's a great milestone to have this done. I left a couple of questions which are just due diligence.

packages/types/src/schemas/schemas.ts
acdunham: That was a mistake from me 🤦

tcaiger (approved):
Nice! Ka Pai!

PR #4695, opened 2023-07-02
[no-issue]: Fix Storybook babel config

kael89:
Storybook in `ui-components` is currently broken, it fails with errors like this:

> ERROR in /home/kostas/projects/beyondessential/tupaia/node_modules/@aws-sdk/smithy-client/dist-es/default-error-handler.js 11:21
> Module parse failed: Unexpected token (11:21)
> File was processed with these loaders:
>  * ../../node_modules/babel-loader/lib/index.js
> You may need an additional loader to handle the result of these loaders.
> |   const statusCode = $metadata.httpStatusCode ? $metadata.httpStatusCode + "" : undefined;
> |   const response = new exceptionCtor({
> >     name: parsedBody?.code || parsedBody?.Code || errorCode || statusCode || "UnknownError",
> |     $fault: "client",
> |     $metadata
>  @ /home/kostas/projects/beyondessential/tupaia/node_modules/@aws-sdk/smithy-client/dist-es/index.js 7:0-40 7:0-40
> [... ]

I think https://github.com/beyondessential/tupaia/pull/4601 broke it, because the new `aws` packages use `?.` that is not supported by the existing Storybook babel config.


packages/ui-components/.babelrc.js
```diff
@@ -1,42 +0,0 @@
-/*
- * Tupaia
- * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
- */
-
-const isTest = process.env.NODE_ENV === 'test';
```
kael89, line 6: It seems that this was not used... I deleted it to avoid confusion

edmofro: Oh, thanks!

edmofro (approved):
lgtm, thanks for taking care of a chore @kael89!

edmofro:
Ah, unfortunately this is failing with errors like the following
image

Are you sure the .babelrc wasn't being used?

kael89:
> Ah, unfortunately this is failing with errors like the following image
> 
> Are you sure the .babelrc wasn't being used?

I guess it was used in tests! I'll reinstate

PR #4696, opened 2023-07-03
Re-write: Fix admin panel tabs

tcaiger:
### Changes:

- Pass in correct base url to Toolbar component

### DIFF
--- a/packages/admin-panel/src/App.js
+++ b/packages/admin-panel/src/App.js
@@ -4,9 +4,9 @@
  */
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Route, Switch, Redirect, useRouteMatch} from 'react-router-dom';
+import { Route, Switch, Redirect } from 'react-router-dom';
 import { connect } from 'react-redux';
-import { TabsToolbar } from '@tupaia/ui-components'; 
+import { TabsToolbar } from '@tupaia/ui-components';
 import { Navbar, Footer } from './widgets';
 import { ROUTES } from './routes';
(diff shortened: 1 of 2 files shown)


packages/admin-panel/src/App.js
```diff
@@ -17,7 +17,6 @@ import { labelToId } from './utilities';
 
 export const App = ({ user }) => {
   const headerEl = React.useRef(null);
-  const match = useRouteMatch()
```
tcaiger, line 20: the match is reading the context of where the hook is placed so the top level App component is always matching `/`

packages/admin-panel/src/App.js
```diff
@@ -39,24 +38,32 @@ export const App = ({ user }) => {
...
-                    <tab.component getHeaderEl={getHeaderEl} />
-                  </Route>
-                ))}
-                <Redirect to={route.to} />
-              </Switch>
-            </Route>
+            <Route
+              key={route.to}
+              path={route.to}
+              render={({ match }) => {
```
tcaiger, line 44: this is the correct match as it is in the context that the tabs are rendered.

acdunham (approved):
Looks good to me! Thanks for doing this. Interesting about this one - I tested this when I made this change to the `TabsToolbar` and it worked, but this makes sense. Does it need updating in LESMIS as well?

tcaiger:
> 

Just noting that the tabs did render ok but when clicking on them they errored as the tab link was not working.

PR #4698, opened 2023-07-03
Split DataBroker.pull API per data source

kael89:
## Motivation

I've wanted to make this change for some time now... We're talking about years :smile: 

When we were writing `data-broker`, @edmofro had warned me that we shouldn't use the same API for different data sources (e.g. `pull()` for both analytics and events). Unfortunately, I didn't heed his warning back then - we were moving fast and all that. It is time to set things right!

This PR splits `DataBroker.pull()` in one API per pulled data source type:

* `pullAnalytics()`
* `pullEvents()`
* `pullSyncGroupResults`

Benefits:
* Fix an over-abstraction. Get rid of `this.mergers`, `this.fetchers` etc
* More sensible types
* Easier to see what's happening at the function level
* Can still reuse functionality when needed, as an implementation detail

## Testing

The package has good coverage and most parts of the refactoring are mechanical, however it may be wise to do a smoke test with a few vizes using various data source types, even if post-merge.

A change like this is probably not a top priority, so it's only worth the hassle if it's relatively easy to review and test.

## Next steps

Next steps could be:
* Split `DataBroker.push()`, `DataBroker.delete()`, `DataBroker.pullMetadata()` APIs
* Split Service-level implementations of those APIs (`DhisService.pull()` etc)

Good news is that those changes can happen gradually, in small and safe PRs :slightly_smiling_face:

### DIFF
--- a/packages/aggregator/src/Aggregator.js
+++ b/packages/aggregator/src/Aggregator.js
@@ -3,6 +3,7 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  */
 
+import { toArray } from '@tupaia/utils';
 import {
   adjustOptionsToAggregationList,
   aggregateAnalytics,
@@ -59,8 +60,6 @@ export class Aggregator {
     }, analytics);
 
   async fetchAnalytics(codeInput, fetchOptions, aggregationOptions = {}) {
-    const code = Array.isArray(codeInput) ? codeInput : [codeInput];
-    const dataSourceSpec = { code, type: this.dataSourceTypes.DATA_ELEMENT };
     const [adjustedFetchOptions, adjustedAggregationOptions] = await adjustOptionsToAggregationList(
       this.context,
       fetchOptions,
@@ -79,7 +78,8 @@ export class Aggregator {
       };
     }
 
-    const { results, metadata } = await this.dataBroker.pull(dataSourceSpec, adjustedFetchOptions);
+    const codes = toArray(codeInput);
+    const { results, metadata } = await this.dataBroker.pullAnalytics(codes, adjustedFetchOptions);
     const rawAnalytics = results.reduce((array, { analytics }) => array.concat(analytics), []);
 
     return {
@@ -108,7 +108,6 @@ export class Aggregator {
    * @returns
    */
   async fetchEvents(code, fetchOptions, aggregationOptions = {}) {
-    const dataSourceSpec = { code, type: this.dataSourceTypes.DATA_GROUP };
     const [adjustedFetchOptions, adjustedAggregationOptions] = await adjustOptionsToAggregationList(
       this.context,
       fetchOptions,
@@ -120,7 +119,7 @@ export class Aggregator {
       return [];
     }
 
-    const events = await this.dataBroker.pull(dataSourceSpec, adjustedFetchOptions);
+    const events = await this.dataBroker.pullEvents([code], adjustedFetchOptions);
 
     return this.processEvents(events, adjustedAggregationOptions);
   }
--- a/packages/aggregator/src/__tests__/Aggregator/Aggregator.test.js
+++ b/packages/aggregator/src/__tests__/Aggregator/Aggregator.test.js
@@ -25,7 +25,8 @@ const { DATA_ELEMENT, DATA_GROUP } = DATA_SOURCE_TYPES;
 const dataBroker = createJestMockInstance('@tupaia/data-broker', 'DataBroker', {
   context: {},
   getDataSourceTypes: () => DATA_SOURCE_TYPES,
-  pull: ({ type }) => RESPONSE_BY_SOURCE_TYPE[type],
+  pullAnalytics: async () => RESPONSE_BY_SOURCE_TYPE[DATA_ELEMENT],
+  pullEvents: async () => RESPONSE_BY_SOURCE_TYPE[DATA_GROUP],
 });
 
 let aggregator;
@@ -56,17 +57,6 @@ describe('Aggregator', () => {
   });
 
   describe('fetchAnalytics()', () => {
-    const assertDataBrokerPullIsInvokedCorrectly = ({ codeInput }, additionalOptions) => {
-      expect(dataBroker.pull).toHaveBeenCalledOnceWith(
-        { code: codeInput, type: DATA_ELEMENT },
-        { ...fetchOptions, ...additionalOptions },
-      );
-    };
-
-    const assertDataBrokerPullIsNotInvoked = () => {
-      expect(dataBroker.pull).toHaveBeenCalledTimes(0);
-    };
-
     it('`aggregationOptions` parameter is optional', async () => {
       const assertErrorIsNotThrown = async emptyAggregationOptions =>
         expect(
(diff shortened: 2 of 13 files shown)


packages/aggregator/src/Aggregator.js
```diff
@@ -120,7 +119,7 @@ export class Aggregator {
       return [];
     }
 
-    const events = await this.dataBroker.pull(dataSourceSpec, adjustedFetchOptions);
+    const events = await this.dataBroker.pullEvents([code], adjustedFetchOptions);
```
kael89, line 122: I've had a look around and `code` is always a single string

packages/aggregator/src/__tests__/Aggregator/Aggregator.test.js
```diff
@@ -56,17 +57,6 @@ describe('Aggregator', () => {
...
   describe('fetchAnalytics()', () => {
-    const assertDataBrokerPullIsInvokedCorrectly = ({ codeInput }, additionalOptions) => {
-      expect(dataBroker.pull).toHaveBeenCalledOnceWith(
-        { code: codeInput, type: DATA_ELEMENT },
-        { ...fetchOptions, ...additionalOptions },
-      );
-    };
-
-    const assertDataBrokerPullIsNotInvoked = () => {
-      expect(dataBroker.pull).toHaveBeenCalledTimes(0);
```
kael89, line 67: Better to use one-liner assertions inline

packages/data-broker/src/DataBroker/DataBroker.ts
```diff
@@ -323,24 +301,15 @@ export class DataBroker {
...
-    options: Record<string, unknown>,
-  ): Promise<EventResults>;
-  public async pull(
-    dataSourceSpec: DataSourceSpec<'syncGroup'>,
-    options: Record<string, unknown>,
-  ): Promise<SyncGroupResults>;
-  public async pull(dataSourceSpec: DataSourceSpec, options: Record<string, unknown> = {}) {
-    const dataSources = await this.fetchDataSources(dataSourceSpec);
-    const { type } = dataSourceSpec;
+  public async pullAnalytics(
```
kael89, line 304: This is how I split the `pull` method into one per data source type:

* Copy-pasta the whole thing :spaghetti: 
* Replace `dataSourceSpec` with a plain array of codes.
* Rename `dataSource` to `dataElement`/`dataGroup`/`syncGroup`
* Use specific APIs for fetching and merging

packages/data-broker/src/DataBroker/DataBroker.ts
```diff
@@ -323,24 +301,15 @@ export class DataBroker {
...
+  ): Promise<AnalyticResults> {
+    const dataElements = await fetchDataElements(this.models, dataElementCodes);
     const validatedOptions = setOrganisationUnitCodes(options);
 
-    const pulls = await this.getPulls(dataSources, validatedOptions.organisationUnitCodes);
+    const pulls = await this.getPulls(dataElements, validatedOptions.organisationUnitCodes);
+    const type = this.getDataSourceTypes().DATA_ELEMENT;
     const nestedResults = await Promise.all(
       pulls.map(({ dataSources: dataSourcesForThisPull, serviceType, dataServiceMapping }) => {
         return this.pullForServiceAndType(
```
kael89, line 315: `pullForServiceAndType` should also be split - inline the data source specific logic, and perhaps used a shared function for shared parts of the logic.

I'll leave this for a future PR to keep the PR simple

packages/data-broker/src/DataBroker/DataBroker.ts
```diff
@@ -352,12 +321,61 @@ export class DataBroker {
...
+          dataSourcesForThisPull,
+          validatedOptions,
+          type,
+          serviceType,
+          dataServiceMapping,
+        );
+      }),
+    );
+
+    return (nestedResults as EventResults[]).flat();
```
kael89, line 349: Merging events is simple, we don't need a separate function

packages/data-broker/src/DataBroker/DataBroker.ts
```diff
@@ -352,12 +321,61 @@ export class DataBroker {
...
+          dataSourcesForThisPull,
+          validatedOptions,
+          type,
+          serviceType,
+          dataServiceMapping,
+        );
+      }),
+    );
+
+    return (nestedResults as SyncGroupResults[]).reduce(
```
kael89, line 373: Merging sync groups results is also simple, it's just merging hashmaps

packages/data-broker/src/DataBroker/DataBroker.ts
```diff
@@ -383,56 +401,6 @@ export class DataBroker {
     });
   };
 
-  private mergeAnalytics = (
```
kael89, line 386: Analytics merging is kind of complicated. Here I just extracted the existing `DataBroker.mergeAnalytics` method into a pure function.

`DataBroker` has a lot of responsibilities and more have been added over the years... Good old separate methods/classes per concern can help tame complexity

packages/data-broker/src/DataBroker/mergeAnalytics.ts
```diff
@@ -0,0 +1,59 @@
...
+  results: {
+    analytics: Analytic[];
+    numAggregationsProcessed: number;
+  }[];
+  metadata: {
+    dataElementCodeToName: Record<string, string>;
+  };
+}
+
+export const mergeAnalytics = (
```
kael89, line 18: This is just the existing `DataBroker.mergeAnalytics()` extracted in a pure function

packages/data-broker/src/__tests__/DataBroker/DataBroker.test.ts
```diff
@@ -282,98 +294,38 @@ describe('DataBroker', () => {
   });
 
   describe('pull()', () => {
-    describe('input validation', () => {
```
kael89, line 285: This is the responsibility of a different API now (`fetchAnalytics()` etc), test cases have moved there

edmofro (approved):
Such a great refactor, it becomes so obvious that the bit of duplication this adds is worth the increased clarity now you've done it. Coincidentally I watched [this talk](https://www.youtube.com/watch?v=8bZh5LMaSmE) yesterday and although it's not exactly the same pattern, the principles are totally resonant. Including the temporary increased overall complexity that we have while things like `fetchDataSources` are still around, which I would love to see go, but must remain until the refactor is complete. Great move keeping this contained though, this is about the right size PR for me to pick up and review same day - I think any larger and it may have sat for quite a while.

I have read this through closely and honestly don't have any changes to suggest, it unfolds naturally. Great work @kael89, your subsequent years of programming are showing!

Have you tested this locally? If so, I'm happy to merge this into dev today or tomorrow, as we're at the start of a release cycle so this will undergo passive regression testing until our next release in a fortnight.

kael89:
> Such a great refactor, it becomes so obvious that the bit of duplication this adds is worth the increased clarity now you've done it. Coincidentally I watched [this talk](https://www.youtube.com/watch?v=8bZh5LMaSmE) yesterday and although it's not exactly the same pattern, the principles are totally resonant. Including the temporary increased overall complexity that we have while things like `fetchDataSources` are still around, which I would love to see go, but must remain until the refactor is complete. Great move keeping this contained though, this is about the right size PR for me to pick up and review same day - I think any larger and it may have sat for quite a while.
> 
> I have read this through closely and honestly don't have any changes to suggest, it unfolds naturally. Great work @kael89, your subsequent years of programming are showing!
> 
> Have you tested this locally? If so, I'm happy to merge this into dev today or tomorrow, as we're at the start of a release cycle so this will undergo passive regression testing until our next release in a fortnight.

Thanks for sharing your thoughts and the talk by Sandi Metz! Her article https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction is great too.

I haven't tested this locally end-to-end. I'll try to do so tonight and let you know of the results. It will be amazing to merge this tomorrow if all goes well! :rocket:

kael89:
@edmofro I did some testing locally:

* Vizes using analytics from a single service (tupaia or indicator) :heavy_check_mark: 
* Vizes using analytics from multiple service (tupaia & indicator) :heavy_check_mark: 
* Vizes using events from a single service (tupaia or indicator) :heavy_check_mark: 

So that leaves out:
* Vizes using events from multiple "pulls". I'm not sure if this is tupaia & DHIS, or just different DHIS instances :question: 
* Sync groups :question: 

In general, testing DHIS/Kobo vizes end-to-end without integration with these systems is hard if not pointless: I'd have to fake the APIs. To compensate for that, I added a couple of tests for sync group fetching.

Given the nature of the refactoring and the package's test coverage, this seems relatively safe to merge but up to you.

coverage after

If you are concerned I volunteer to do some testing comparing a feature branch against another one, however I'd need someone to spin up those and give me access.

edmofro:
That's a decent lot of testing, and anything else should be caught by either a) its use in dev over the next fortnight, or b) regression & e2e testing during the next release. 
cc- @avaek highlighting that this is coming in as a solidly unit tested, but partially manual tested change. There are no differences in behaviour, it is purely a refactor.

PR #4701, opened 2023-07-04
Update central server map overlays filter to account for empty string

EMcQ-BES:
### Changes:

- Special case filter for when we have no permission groups on a map overlay
- Increase the limits on the fetch from `tupaia-web-server -> central`, with full permissions it's possible to have >100 map overlays return at once

### DIFF
--- a/packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
+++ b/packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
@@ -16,6 +16,7 @@ export type MapOverlaysRequest = Request<any, any, any, any>;
 // TODO: Can these be moved into types?
 const ROOT_MAP_OVERLAY_CODE = 'Root';
 const MAP_OVERLAY_CHILD_TYPE = 'mapOverlay';
+const PAGE_SIZE = 200;
 
 // We return a simplified version of data to the frontend
 interface TranslatedMapOverlay {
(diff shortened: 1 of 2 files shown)


packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
```diff
@@ -57,6 +58,7 @@ export class MapOverlaysRoute extends Route<MapOverlaysRequest> {
           comparisonValue: [projectCode],
         },
       },
+      pageSize: PAGE_SIZE,
```
IgorNadj, line 61: Should pageSize be sent from the client instead? Or is the client not pagination aware when it should be?

I'm not super across this epic branch but seems like we will have the same issue with all REST endpoints, and it might be better to do a generic solution than a specific one.

EMcQ-BES: Yeah usually I just pass through the user's query params to this main fetch, but in a case like this where we're doing multiple fetches for relationship entries, the later fetches may have more entries than this initial one. So we can't take `pageSize` from the user for all of these.

I've updated to pass `pageSize: 'ALL'` which will remove the `100` restriction that the `central-server` uses, but also the allow the user to override that with their query

IgorNadj: Ah ok sounds good!

IgorNadj (approved):

PR #4707, opened 2023-07-06
Waitp 1231 base matrix component

acdunham:
### Changes:
- Created reusable matrix component in `ui-components`
- Added stories
- Exposed matrix-related types from `types` package

Note: column groups are not implemented as yet, that will be in a separate PR

#### Basic matrix
Screenshot 2023-07-06 at 4 36 56 pm

#### Nested with dots
Screenshot 2023-07-06 at 4 37 47 pm

#### Nested with basic data
Screenshot 2023-07-06 at 4 38 39 pm

#### Modal with data, shown on click of cell
Screenshot 2023-07-06 at 4 38 09 pm


packages/ui-components/src/components/Matrix/Matrix.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { MatrixColumnType, MatrixRowType } from '../../types';
+import { hexToRgba } from './utils';
+import { ACTION_TYPES, MatrixContext, MatrixDispatchContext, matrixReducer } from './MatrixContext';
+import { MatrixNavButtons } from './MatrixNavButtons';
+import { MatrixRow } from './MatrixRow';
+import { EnlargedMatrixCell } from './EnlargedMatrixCell';
+
+const MatrixTable = styled.table`
+  border-collapse: collapse;
+  border: 1px solid ${({ theme }) => hexToRgba(theme.palette.text.primary, 0.2)};
```
tcaiger, line 19: I'm keen to remove the `hexToRgba` function in favour of just using hex colors with opacity. So for example we should just be able to do `${theme.palette.text.primary}33` to get the same result, right?.
https://www.w3schools.com/cssref/css_colors_legal.php

packages/ui-components/src/components/Matrix/Matrix.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <MatrixContext.Provider
+        value={{
+          presentationOptions,
+          columns,
+          rows,
+          startColumn,
+          maxColumns,
+          expandedRows,
+          enlargedCell,
+        }}
```
tcaiger, line 79: Do you think there is any need to wrap this context value in a useMemo to optimise performance?

packages/ui-components/src/components/Matrix/MatrixCell.tsx
```diff
@@ -0,0 +1,97 @@
...
+export const Dot = styled.div<{ $color?: string }>`
+  width: 2rem;
+  height: 2rem;
+  border-radius: 50%;
+  background-color: ${({ $color }) => $color};
+  border: 0.375rem solid
+    ${({ theme }) =>
+      theme.palette.background.default === 'transparent'
+        ? 'transparent'
+        : hexToRgba(theme.palette.background.default, 0.8)};
```
tcaiger, line 22:
```suggestion
        : `${theme.palette.background.default}cc`;
```

packages/ui-components/src/components/Matrix/Matrix.tsx
```diff
@@ -0,0 +1,96 @@
...
+    },
+  );
+  const tableEl = useRef<HTMLTableElement | null>(null);
+
+  useEffect(() => {
+    // Update the number of columns per page when the table is mounted and data comes in, so that we can determine when to show the nav buttons
+    const updateMaxColumns = () => {
+      if (!tableEl || !tableEl?.current || !tableEl?.current?.offsetWidth) return;
+      const { offsetWidth } = tableEl?.current;
+      // 200px is the max width of a column that we want to show
```
tcaiger, line 57: nice comment

packages/ui-components/src/components/Matrix/MatrixContext.ts
```diff
@@ -0,0 +1,89 @@
+import { Dispatch, createContext } from 'react';
+import { MatrixColumnType, MatrixRowType } from '../../types';
+import { PresentationOptions } from '@tupaia/types';
+
+type RowTitle = MatrixRowType['title'];
+
+const defaultContextValue = {
```
tcaiger, line 7: This is really nice. I think that useReducer and context are a great choice here and it organises all these fiddly interactions really nicely.

packages/ui-components/src/components/Matrix/index.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2022 Beyond Essential Systems Pty Ltd
+ */
+
+export { Matrix } from './Matrix';
+export * from './utils';
```
tcaiger, line 7: Do the utils need to be exported? I can't see them used anywhere else.

packages/ui-components/src/components/Matrix/utils.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+};
+
+export function getIsUsingDots(presentationOptions: PresentationOptions = {}) {
+  return Object.keys(presentationOptions).length > 0;
+}
+
+export function checkIfApplyDotStyle(presentationOptions: PresentationOptions = {}) {
+  return presentationOptions?.applyLocation?.columnIndexes;
+}
```
tcaiger, line 92:
```suggestion

```
I don't think this is used anywhere right?

tcaiger (changes requested):
Looks so good. Nice work handling all that complexity. I just have one requested change about removing the rgb util

packages/ui-components/src/components/Matrix/index.ts
acdunham: I mainly did this in case of using them in `tupaia-web` but I can remove until that comes up

tcaiger (approved):
Nice!

PR #4709, opened 2023-07-07
RN-950: Configure cypress babel-preprocessor to parse node_modules

rohan-bes:
### Issue RN-950:

The issue here is that the [default cypress webpack preprocessor](https://github.com/cypress-io/cypress/tree/develop/npm/webpack-preprocessor) config only supports ES2015 and doesn't parse files under `/node_modules/`.

This meant that dependencies which used modern js (eg. optional chaining) caused webpack build errors. 

Extended the config here to parse depedencies and to stub out unsupported peer depedencies.

### DIFF
--- a/packages/e2e/cypress/plugins/index.js
+++ b/packages/e2e/cypress/plugins/index.js
@@ -4,6 +4,7 @@
  */
 
 const cypressDotenv = require('cypress-dotenv');
+const webpackPreprocessor = require('@cypress/webpack-preprocessor');
 const fs = require('fs');
 
 module.exports = (on, config) => {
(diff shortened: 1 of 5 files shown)


rohan-bes:
@IgorNadj I chose to leave `requireCyEnv` in the `@tupaia/e2e` package since it wasn't needed anywhere else.

As this change stands, there's actually no build issues regardless of changing the webpack config, because none of the packages being used cause errors. But we should be safe to add `@tupaia/utils` or any other package back as a dependency and things should still run smoothly

IgorNadj:
Discussed in a huddle, cypress should support polymorphic (CJS+ESM) packages out of the box, lets try an upgrade. Cypress is 6 major versions behind, about 2 years.

If the upgrade doesn't work or complicates things too much, the changes in this PR to customise cypress' webpack to parse ESM dist files into commonjs is odd but ok, leave a big comment explaining it.

Storybook has the same problem. 1 major version behind, about 2 years, upgrade should fix it.

PR #4711, opened 2023-07-07
WAITP-1289 Entity Search Route

EMcQ-BES:
### Changes:

- Add `entitySearch` routes to `tupaia-web-server` and `entity-server`
- Add unit tests for `entity-server`
- Allow using raw SQL sort queries in database `find` functions


packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
```diff
@@ -0,0 +1,63 @@
...
+    const { hierarchyId, fields, field, filter } = this.req.ctx;
+    const { searchString: rawString } = this.req.params;
+    const { pageSize, page } = this.req.query;
+    const searchString = rawString.toLowerCase();
+
+    const entities = await this.req.models.entity.find(
+      {
+        ...filter,
+        [RAW]: {
+          // Running this RAW is much faster for larger hierarchies
```
EMcQ-BES, line 36: Anecdotally, about 6x quicker than an array filter

packages/tupaia-web-server/src/routes/EntitySearchRoute.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const project = (
+      await ctx.services.central.fetchResources('projects', {
+        filter: { code: projectCode },
+        columns: ['entity_hierarchy.name'],
+      })
+    )[0];
+    const entityHierarchyName = project['entity_hierarchy.name'];
+
+    return ctx.services.entity.entitySearch(entityHierarchyName, searchString, {
```
rohan-bes, line 26:
```suggestion
    return ctx.services.entity.entitySearch(projectCode, searchString, {
```
You're all good using the project code here. For some time now we've committed to there being a one to one mapping between projects and entity hierarchies, and the project code and the hierarchy name are always the same. It's an assumption made in various parts of the code.

packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const { RAW } = QUERY_CONJUNCTIONS;
+
+export type EntitySearchRequest = MultiEntityRequest<
+  MultiEntityRequestParams & { searchString: string },
+  EntityResponse[],
+  MultiEntityRequestBody,
+  EntityRequestQuery & { pageSize?: number; page?: number }
+>;
+export class EntitySearchRoute extends Route<EntitySearchRequest> {
```
rohan-bes, line 25: I was really wanting this to be something we could support using just the existing routes. We already have the ability to filter results by a search term, so if we added the ability to paginate and sort across all routes then we'd almost be there.

But I think the custom sort that we do here is quite hard to support. Even if we broke the request from the `tupaia-web-server` into multiple api calls, getting the pagination working across those multiple calls would prove difficult.

So happy with this being added, and it'll probably get nice re-use on other platforms too 👍

packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
rohan-bes: Oooh interesting. I wonder if this is more performant than the existing hierarchy functions on `EntityModel` (`getDescendantsOfEntities`,`getAncestorsOfEntities`,etc.) I think those use a join on the `ancestor_descendant_relation` table

rohan-bes (approved):
Looks great! Thanks for the tests 👍 

Could I ask you add an example to the `example.http` file? But pre-approving all the same

packages/tupaia-web-server/src/routes/EntitySearchRoute.ts
EMcQ-BES: Oh cool. I run this same code against a few routes so I'll rip them all out. 👍 Big fan of deleting code

packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
EMcQ-BES: This is almost exactly the thought process I went through too. Tried supporting through existing routes with multiple calls, found that difficult for pagination, figured this route would be useful for other platforms 😅

packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
EMcQ-BES: It'd require some testing but academically the answer is "probably not". In general postgres can optimise a join query better than a subquery.

packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
rohan-bes: Thanks for thinking that through, it's always nice to try and keep the api lean. 👍 

I think if we returned a bit of metadata with the request, we could look at supporting this with the existing routes.
eg. something like:
```
/hierarchy/explore/explore/descendants?field=name&filter=name=@cat&page=0&pageSize=10&sort=name asc
=>
{
  page: 0,
  totalPages: 10,
  data: [
    "Yacata Nursing Station",
    "Lemakot Catholic Miss",
    ...
  ]
}
```

Honestly, that makes for a more powerful api generally I think, and would be a nice change to make at some point. But for now I'm happy with going with what we've got 👍

PR #4713, opened 2023-07-07
Fix meditrak api url resolution

IgorNadj:
### Issue #: no issue

This fixes a problem where an empty CENTRAL_API_URL env var was causing the BASE_URL to resolve to '';

### DIFF
--- a/packages/meditrak-app/app/api/TupaiaApi.js
+++ b/packages/meditrak-app/app/api/TupaiaApi.js
@@ -22,7 +22,7 @@ const SOCIAL_FEED_ENDPOINT = 'socialFeed';
 const CURRENT_USER_REWARDS_ENDPOINT = 'me/rewards';
 const DEV_BASE_URL = 'https://dev-api.tupaia.org/v2';
 const PRODUCTION_BASE_URL = `https://${isBeta ? `${betaBranch}-` : ''}api.tupaia.org/v2`;
-export const BASE_URL = centralApiUrl ?? (__DEV__ ? DEV_BASE_URL : PRODUCTION_BASE_URL);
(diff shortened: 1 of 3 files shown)


packages/meditrak-app/app/api/TupaiaApi.js
```diff
@@ -22,7 +22,7 @@ const SOCIAL_FEED_ENDPOINT = 'socialFeed';
 const CURRENT_USER_REWARDS_ENDPOINT = 'me/rewards';
 const DEV_BASE_URL = 'https://dev-api.tupaia.org/v2';
 const PRODUCTION_BASE_URL = `https://${isBeta ? `${betaBranch}-` : ''}api.tupaia.org/v2`;
-export const BASE_URL = centralApiUrl ?? (__DEV__ ? DEV_BASE_URL : PRODUCTION_BASE_URL);
+export const BASE_URL = __DEV__ ? centralApiUrl || DEV_BASE_URL : PRODUCTION_BASE_URL;
```
IgorNadj, line 25: TIL `a ?? b` only returns a if a is `null` or `undefined`, not falsy like most other conditionals in javascript. 

```
'' ? '' : 'x'   --> 'x'
'' || 'x'       --> 'x'
'' ?? 'x'       --> ''
```

I've also moved centralApiUrl to behind the __DEV__ check so it can only be used in local development.

EMcQ-BES (approved):
LGTM

packages/meditrak-app/app/api/TupaiaApi.js
EMcQ-BES: Ah, nullish vs falsey, a classic gotcha

PR #4715, opened 2023-07-09
added country column to admin-panel survey responses page

WDuncker:
### Changes:
- added custom join condition to GETSurveyResponses to join country table.
- updated front end to add country column to the admin panel survey responses page

### DIFF
--- a/packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
+++ b/packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
@@ -20,6 +20,10 @@ import { getQueryOptionsForColumns } from '../GETHandler/helpers';
 export class GETSurveyResponses extends GETHandler {
   permissionsFilteredInternally = true;
 
+  customJoinConditions = {
+    country: ['country.code', 'entity.country_code'],
+  };
+
   async findSingleRecord(surveyResponseId, options) {
     const surveyResponse = await super.findSingleRecord(surveyResponseId, options);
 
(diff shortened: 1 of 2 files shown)


acdunham (approved):
The frontend looks good to me. Well done. The codeship build is failing because the branch name is too long (we have a limit of 46 characters.) So what you might need to do is create a new branch off `dev` with a name under 46 characters long, and then cherrypick this commit into that branch. Then you can just link me to the PR for that new branch and I can approve it :)

acdunham (changes requested):
I just also ran it locally, and I get this error:
image

packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
```diff
@@ -20,6 +20,10 @@ import { getQueryOptionsForColumns } from '../GETHandler/helpers';
 export class GETSurveyResponses extends GETHandler {
   permissionsFilteredInternally = true;
 
+  customJoinConditions = {
+    country: ['country.code', 'entity.country_code'],
```
EMcQ-BES, line 24: I think there's a possible edge case here. Can you test what happens if you request the `country.name` column, but don't request any `entity.{field}` columns?

I think you can just comment out the entity name section of that admin panel page to test it.

acdunham: Could this be related to this error?
image

EMcQ-BES: Ah yep. @acdunham I think this'll be related to your error message. We're relying on having already joined to the entity table. So if we don't request an entity column it'll fail to join. But if we request a country column _before_ an entity column it'll generate the joins backwards and also fail to join.

PR #4716, opened 2023-07-10
rename types and add to exports

acdunham:
### Changes:
- renamed general types files in `ui-components` and `ui-map-components` packages so that they can be imported from other packages
- Added `types` to exports in main `index` file
- Fixed build errors in all FE packages


EMcQ-BES (approved):
LGTM

PR #4717, opened 2023-07-10
Waitp 1301 add country to admin panel srp

WDuncker:
### Changes:

-Joined country and entity tables to survey response table in GETSurveyResponses.js
-Added to getQueryOptionsForColumns in GETHandler/helpers.js to allow for handling of through joins.
-Added test to GETSurveyResponses
-Added Country column to admin panel survey responses page in front end
-Updated customJoinConditions format everywhere they appear in the central-server


packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
```diff
@@ -20,6 +20,11 @@ import { getQueryOptionsForColumns } from '../GETHandler/helpers';
 export class GETSurveyResponses extends GETHandler {
   permissionsFilteredInternally = true;
 
+  customJoinConditions = {
+    entity: ['entity.id', 'survey_response.entity_id'],
+    country: ['country.code', 'entity.country_code'],
+  };
+
   async findSingleRecord(surveyResponseId, options) {
```
acdunham, line 28: @EMcQ-BES can you cast your eyes over this one?

acdunham: I'm still getting this error:
image

packages/central-server/src/tests/apiV2/GETSurveyResponses.test.js
```diff
@@ -122,7 +124,7 @@ describe('Permissions checker for GETSurveyResponses', async () => {
     app.revokeAccess();
   });
 
-  describe('GET /surveyResponses/:id', async () => {
+  describe.only('GET /surveyResponses/:id', async () => {
```
EMcQ-BES, line 127:
```suggestion
  describe('GET /surveyResponses/:id', async () => {
```

packages/central-server/src/tests/apiV2/GETSurveyResponses.test.js
```diff
@@ -152,7 +154,7 @@ describe('Permissions checker for GETSurveyResponses', async () => {
     });
   });
 
-  describe('GET /surveyResponses', async () => {
+  describe.only('GET /surveyResponses', async () => {
```
EMcQ-BES, line 157:
```suggestion
  describe('GET /surveyResponses', async () => {
```

packages/central-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -86,19 +86,51 @@ const getForeignKeyColumnName = foreignTable => {
...
+  return multiJoin;
+};
+
 export const getQueryOptionsForColumns = (
   columnNames,
   baseRecordType,
   customJoinConditions = {},
   joinType = null,
 ) => {
+  console.log(baseRecordType);
```
EMcQ-BES, line 124:
```suggestion
```

packages/central-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -86,19 +86,51 @@ const getForeignKeyColumnName = foreignTable => {
...
   if (columnNames.some(c => c.startsWith('_'))) {
     throw new ValidationError(
       'No columns start with "_", and conjunction operators are reserved for internal use only',
     );
   }
   const multiJoin = [];
   const recordTypesInQuery = new Set([baseRecordType]);
+  console.log(recordTypesInQuery.size);
+  console.log(baseRecordType);
```
EMcQ-BES, line 133:
```suggestion
```

packages/central-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -107,15 +139,24 @@ export const getQueryOptionsForColumns = (
...
+          );
+        }
       }
     });
+  console.log(recordTypesInQuery.size);
   // Ensure every join table is added to the sort, so that queries are predictable during pagination
+  console.log([...recordTypesInQuery]);
   const sort = [...recordTypesInQuery].map(recordType => `${recordType}.id`);
+  console.log(sort);
+  console.log(multiJoin);
```
EMcQ-BES, line 160:
```suggestion
  // Ensure every join table is added to the sort, so that queries are predictable during pagination
  const sort = [...recordTypesInQuery].map(recordType => `${recordType}.id`);
```

packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
```diff
@@ -73,6 +77,8 @@ export class GETSurveyResponses extends GETHandler {
       this.defaultJoinType,
     );
 
+    // manipulate resulting multi-join with country code specific requirements
+
```
EMcQ-BES, line 81:
```suggestion
```

packages/central-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -86,19 +86,51 @@ const getForeignKeyColumnName = foreignTable => {
...
+  multiJoin,
+  baseRecordType,
+  customJoinConditions,
+  joinType,
+) => {
+  if (!recordTypesInQuery.has(recordType)) {
+    const join = customJoinConditions[recordType];
+    const joinCondition = join.through
+      ? [join.foreignTable, join.foreignKey]
+      : [`${recordType}.id`, `${baseRecordType}.${getForeignKeyColumnName(recordType)}`];
```
EMcQ-BES, line 101: This is going to ignore the old format `customJoinConditions`

If `join.through` doesn't exist currently we generate a default `joinCondition`, but for the old format, we would want `joinCondition = join`

EMcQ-BES (changes requested):
Two things to fix up

- You've included some debugging code, `console.log` / `describe.only`
- The `joinCondition` logic isn't doing _quite_ what you think it is

packages/central-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -107,15 +139,24 @@ export const getQueryOptionsForColumns = (
...
       const recordType = resourceToRecordType(resource);
       if (!recordTypesInQuery.has(recordType)) {
-        const joinCondition = customJoinConditions[recordType] || [
-          `${recordType}.id`,
-          `${baseRecordType}.${getForeignKeyColumnName(recordType)}`,
-        ];
-        multiJoin.push({ joinWith: recordType, joinCondition, joinType });
-        recordTypesInQuery.add(recordType);
+        const join = customJoinConditions[recordType];
+        if (join) {
```
EMcQ-BES, line 143: Currently you're checking if a `customJoinCondition` exists before running the `addJoin` function. This means it won't generate default joins

I think we can remove all these pre-checks and _just_ run the `addJoin` function, since we run similar checks within that function

EMcQ-BES (approved):
Looks good to me 👍

packages/central-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -86,6 +86,35 @@ const getForeignKeyColumnName = foreignTable => {
...
+  customJoinConditions,
+  joinType,
+) => {
+  if (!recordTypesInQuery.has(recordType)) {
+    const join = customJoinConditions[recordType];
+    const joinCondition = join
+      ? [join.foreignTable, join.foreignKey]
+      : [`${recordType}.id`, `${baseRecordType}.${getForeignKeyColumnName(recordType)}`];
+    if (join?.through) {
+      addJoin(
```
edmofro, line 103: Nice recursive function!

packages/central-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -86,6 +86,35 @@ const getForeignKeyColumnName = foreignTable => {
   return exceptions[foreignTable] || `${foreignTable}_id`;
 };
 
+const addJoin = (
+  recordType,
+  recordTypesInQuery,
+  multiJoin,
+  baseRecordType,
+  customJoinConditions,
+  joinType,
```
edmofro, line 95: Although js will allow it, it's an anti-pattern to mutate (change) function parameters. This is a really good learning opportunity - here you are passing in `multiJoin` and expecting the function to add more joins to it, but you should _never do this_, or you end up in all sorts of trouble. Instead, you should make a _deep_ clone of the object, mutate that, and return it for the caller to use instead of its original object. If you allow functions to mutate things inside them, all of a sudden you lose predictability, and can't trust what invisible side effects might be going on. 
Can you please share this learning with Chao too, it's not obvious when you first start js but important to avoid getting in a pickle.

packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
```diff
@@ -20,6 +20,18 @@ import { getQueryOptionsForColumns } from '../GETHandler/helpers';
 export class GETSurveyResponses extends GETHandler {
   permissionsFilteredInternally = true;
 
+  customJoinConditions = {
+    country: {
+      through: 'entity',
+      foreignKey: 'entity.country_code',
+      foreignTable: 'country.code',
```
edmofro, line 27: I really like `through`! I think the other two fields could do with clearer names though. When we're going `through` another table, there are up to four columns we have to deal with:
- The column we launch off from the base table this endpoint represents 
- The column that the base table column connects to in the through table
- The column in the through table that we connect further on with
- The column in the final join table that we connect to

I wouldn't be able to distinguish which column either `foreignKey` or `foreignTable` relates to, or what would happen if I needed to specify rather than use the default for all four columns. Perhaps have a chat to @EMcQ-BES about alternative names that make this all clear - remember that code is read a lot more than it is written, so we take time now to optimise for people to understand it easily later on!

edmofro (changes requested):
Sorry to push back on an already-approved PR, but I took a look seeing as it's your first PR @WDuncker and have some suggestions that I feel are important to hold up our high code quality bar in tupaia. Thanks guys!

packages/central-server/src/apiV2/GETHandler/helpers.js
```diff
@@ -86,6 +86,35 @@ const getForeignKeyColumnName = foreignTable => {
   return exceptions[foreignTable] || `${foreignTable}_id`;
 };
 
+const addJoin = (
```
edmofro, line 89: This function on its own would be great to cover with unit tests, to make sure the inputs and outputs are expected in the different scenarios. It is quite dense and specific, and has well defined inputs and outputs, so is a great candidate for testing

packages/central-server/src/apiV2/surveyResponses/GETSurveyResponses.js
EMcQ-BES: @edmofro Chatting with @WDuncker about this. I figure we can't _quite_ call them something like e.g. `localTableKey` and `foreignTableKey` because when connecting with a `through` table the first key (`entity.country_code` in this case) technically _isn't_ local to the original table. That variable is playing a bit of double duty here.

Wim floated `nearTableKey` and `farTableKey` which I reckon describes the two pieces of behaviour pretty well. But I figure it might be worth getting a final opinion from you as a more external viewpoint

edmofro: That double duty is a bit what I'm worried about @EMcQ-BES. We don't currently have a requirement where you have to specify non-default columns on all four possible places (the local table from, the through table to, the through table from, and final table to), but what if you did? It feels a bit messy/unpredictable that when you define `farTableKey` you don't know whether it'll be used in the through or the final table.

Perhaps we need `through` to take an object, which has `localTableKey` and `foreignTableKey` as well as `table`?

EMcQ-BES: If you need all four columns to be custom you'd just add another `customJoin`. They chain together in the recursive function. The `farTableKey` is always against the table we're joining to with that definition.

```
Example GETDashboards.js

customJoins: {
  entity: {
    nearTableKey: "dashboard.root_entity_code",
    farTableKey: "entity.code",
  },
  country: {
    through: "entity",
    nearTableKey: "entity.country_code",
    farTableKey: "country.code",
  },
}
```

I feel like I might be missing something in your question. Perhaps we can chat about this tomorrow

edmofro: Right, this makes sense! In that case, I think the names are as good as you'd get. Thanks for bearing with me :-)

packages/central-server/src/tests/apiV2/helpers.test.js
```diff
@@ -0,0 +1,189 @@
...
+import { ValidationError } from '@tupaia/utils';
+import { getQueryOptionsForColumns } from '../../apiV2/GETHandler/helpers';
+
+const baseRecordType1 = 'survey_response';
+const baseRecordType2 = 'project';
+const columnNames1 = ['survey_response.id', 'survey.name'];
+const columnNames2 = ['survey_response.id', 'survey.name', 'user_account.first_name'];
+const columnNames3 = ['survey_response.id', 'country.name'];
+const columnNames4 = ['survey_response.id', 'country.name', 'disaster.type'];
+const columnNames5 = ['_survey_response.id', 'survey.name'];
```
edmofro, line 16:
```suggestion
const columnNames5 = ['survey_response.id', 'survey.name'];
```

edmofro: Ah I see, this is testing a specific case. Because the data here is not huge, I would actually recommend inlining it so that within each test you can see what is going on and why

packages/central-server/src/tests/apiV2/helpers.test.js
```diff
@@ -0,0 +1,189 @@
...
+  entity: ['entity.id', 'project.entity_id'],
+};
+const joinType = null;
+const err = new ValidationError(
+  'Error: No columns start with "_", and conjunction operators are reserved for internal use only',
+);
+const err2 = new ValidationError(
+  'nearTableKey must refer to a column on the table you wish to join through',
+);
+const err3 = new ValidationError('Incorrect format for customJoinConditions: entity');
```
edmofro, line 55: Rather than constructing the full exact error, you can use jest's `toThrow` and do a partial match on the error text

edmofro (changes requested):
Thanks @WDuncker, those tests are great! Just asking for a bit of refactor on them to make them easier to read and a little more robust. After that, I have some changes I'd like to suggest to `addJoin`, but let's tweak the tests first - having tests in place against the next function up the chain (`getQueryOptionsForColumns`) is awesome because it means we can make changes to the implementation _and the interface_ of `addJoin` and the tests will check we've kept the behaviour the same :-)

(wduncker@LAPTOP-AFVENG10 C:Userswimdu.wslconfig)
```diff
@@ -0,0 +1,2 @@
+[wsl2]
+memory=10GB
```
edmofro, line 2: This looks like an accidental commit. Can you delete, commit that, then reinstate and add to the .gitignore?

packages/tupaia-web/.eslintrc.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-env node */
+module.exports = {
+  extends: ['../../.eslintrc-ts-frontend.json'],
+};
```
edmofro, line 4: This has also snuck in. In this case I don't think it needs to be .gitignored, but just deleted on this branch.

edmofro (changes requested):
Looking much nicer to my eyes! There are a couple of extra files that have snuck in here. It might be worth doing a rebase and forced push to avoid the eslint one causing conflicts with the real .eslintrc

edmofro (approved):
lgtm, thanks for your patience!

PR #4718, opened 2023-07-10
Fix issue with parsing chart config

acdunham:
### Changes:
- Update default value of `chartConfig` in `parseChartConfig` to fix issue with chart columns


tcaiger (approved):
great find

PR #4720, opened 2023-07-11
Waitp 1316 Chart table

acdunham:
### Changes:
- Updated `Chart` component to have toggle buttons and table option when enlarged
- Removed forced heights of responsive `recharts` containers, to let it be handled by `recharts`
---

### Screenshots:
Screenshot 2023-07-11 at 2 09 49 pm


packages/tupaia-web/src/api/queries/useDashboards.ts
```diff
@@ -23,12 +23,11 @@ export const useDashboards = (
...
   let activeDashboard = null;
 
-  if (dashboards.length > 0 && dashboardName) {
+  if (data?.length > 0 && dashboardName) {
     activeDashboard =
-      dashboards.find(
-        (dashboard: DashboardsResponse) => dashboard.dashboardName === dashboardName,
-      ) || dashboards[0];
+      data?.find((dashboard: DashboardsResponse) => dashboard.dashboardName === dashboardName) ||
+      data[0];
```
acdunham, line 29: This was because I was getting some weird TS errors, and this fixed it

packages/tupaia-web/src/features/Chart.tsx
```diff
@@ -3,51 +3,128 @@
...
+    display: none;
+  }
+`;
+
+const TabsGroup = styled(Tabs)`
+  border: 1px solid
+    ${({ theme }) => {
+      let hexString = theme.palette.text.primary.replace('#', '');
+      hexString = hexString.length === 3 ? hexString.replace(/(.)/g, '$1$1') : hexString;
+      return `#${hexString}33`;
```
tcaiger, line 47: This seems a bit clunky to me and maybe a bit fragile too as the primary text color could be 3 digit hex, 6 digit hex or also rgba which is what it is actually by default in the [default theme](https://v4.mui.com/customization/default-theme/#default-theme). Maybe we could use lighten instead?

packages/tupaia-web/src/features/Chart.tsx
```diff
@@ -3,51 +3,128 @@
...
-  $isEnlarged: boolean;
-  $hasData: boolean;
-}>`
+const Wrapper = styled.div`
   display: flex;
   position: relative;
   align-content: stretch;
   -webkit-box-align: stretch;
   align-items: stretch;
-  height: ${({ $isEnlarged, $hasData }) => {
```
tcaiger, line 19: this line is fine- just noticed that webkit selector above which is now redundant.

tcaiger (approved):
Nice work. I left a tiny suggestion about the color on the tabs but that's an optional change

PR #4723, opened 2023-07-11
RN-903 QR code Admin Panel

IgorNadj:


packages/admin-panel/src/pages/resources/EntitiesPage.js
```diff
@@ -98,10 +57,13 @@ const FIELDS = [
     },
   },
   {
-    Header: 'QR Code',
+    Header: 'QR',
     source: 'id',
-    type: 'logs',
-    Cell: QRCodeButton,
+    type: 'qrCode',
```
rohan-bes, line 62: Good idea to go with this approach since it's consistent with our other existing modals, but isn't it such an awkward pattern? The logic ends up spread over a bunch of locations, and the modal exists on every page. Maybe we should think about refactoring this part of the Admin Panel otherwise it could grow out of hand a bit...

packages/admin-panel/src/qrCode/QrCode.js
```diff
@@ -0,0 +1,98 @@
...
+        height: SCALED_CODE_SIZE,
+      });
+      const img = new Image();
+      const codeX = (SCALED_CANVAS_SIZE - SCALED_CODE_SIZE) / 2;
+      const codeY = 65 * DOWNLOAD_IMAGE_SCALE; // Make sure to avoid the QR code "quiet zone"
+      img.onload = () => ctx.drawImage(img, codeX, codeY);
+      img.src = qrDataUrl;
+    };
+    run();
+  }, []);
```
rohan-bes, line 81: At least this section, but potentially the whole component might wanna live in `@tupaia/ui-components`? We will definitely want to re-use it for at least the QR Code viz

packages/admin-panel/src/table/columnTypes/QrCodeButton.js
```diff
@@ -0,0 +1,49 @@
...
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+import { openQrCodeModal } from '../../qrCode';
+import { IconButton } from '../../widgets';
+
+const LittleQrCodeIcon = () => (
```
rohan-bes, line 12: MUI has a few options for QR Code Icons if liked any of them? https://mui.com/material-ui/material-icons/?query=qr+code

packages/admin-panel/src/qrCode/QrCode.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export const QrCode = ({ qrCodeContents, humanReadableId }) => {
+  // Draw into a canvas so that we can download it as an image
+  const canvasRef = useRef(null);
+
+  const handleDownload = () => {
+    const canvas = canvasRef.current;
+    const dataUri = canvas.toDataURL('image/jpeg', 1.0);
+    downloadImage(dataUri, `${humanReadableId}.jpeg`);
```
rohan-bes, line 44:
```suggestion
    downloadJs(dataUri, `${humanReadableId}.jpeg`);
```
Looks like we use [downloadJs](https://www.npmjs.com/package/downloadjs) for image downloads in other places

rohan-bes (changes requested):
Nice stuff! Just a few requests for changes.

Also, would you be open to rebasing this (and the commits you need from the epic branch) on to dev? I was thinking of doing that with the QR Code in Meditrak stuff and that way the two parts can be tested and merged independently.

packages/admin-panel/src/pages/resources/EntitiesPage.js
IgorNadj: Yeah it's not ideal but we don't really have the time to do the refactor in this project, maybe the next one.

packages/admin-panel/src/qrCode/QrCode.js
IgorNadj: Yeah this is just working on this card in isolation, I'm happy for this to be pulled out into a library in a future PR, should be easy enough.

packages/admin-panel/src/pages/resources/EntitiesPage.js
rohan-bes: Oh yup, agreed 👍

IgorNadj:
Rebased on dev (no changes)

packages/admin-panel/src/table/columnTypes/QrCodeButton.js
IgorNadj: Good spotting, unfortunately not in the version of mui we're on. Happy for it to stay custom for now?

packages/admin-panel/src/qrCode/QrCode.js
IgorNadj: Nice catch, updated

packages/admin-panel/src/table/columnTypes/QrCodeButton.js
rohan-bes: Sure ting 👍

rohan-bes (approved):
Looks great!

PR #4724, opened 2023-07-11
WAITP-1223: Tupaia web map legend

tcaiger:
### Changes:

- Desktop map legend
- Mobile map legend
- Setup hidden values to handle clicking on the map legend

---

### Screenshots:

image

image

image

image

image

image


packages/tupaia-web/package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "yarn package:lint",
     "lint:fix": "yarn lint --fix",
     "preview": "vite preview",
-    "start-dev": "REACT_APP_DEPLOYMENT_NAME='dev' && vite",
+    "start-dev": "vite",
```
tcaiger, line 18: Setting this variable shows the EnvBanner on local dev which we don't need

packages/tupaia-web/src/features/Map/utils/useMapOverlayReport.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { useParams } from 'react-router';
+import {
+  useMapOverlays,
+  useMapOverlayReport as useMapOverlayReportQuery,
+} from '../../../api/queries';
+import { useDateRanges } from '../../../utils';
+import { URL_SEARCH_PARAMS } from '../../../constants';
+
+export const useMapOverlayReport = () => {
```
tcaiger, line 14: To be honest I keep going back and forth on the use of url params in the query hooks. I added this util as a middle ground so that it keeps the query hooks in the api directory simple but also makes the use of map overlay report data more convenient.

acdunham: I think thats a nice idea, especially for ones like this that pass in more than 1 or 2 params

acdunham (approved):
This is looking really good. I have a couple of questions about possible ways to do some things, but they are optional to carry out.

PR #4726, opened 2023-07-12
WAITP-1269: Fix front end tests config

tcaiger:
### Changes:

- Updates to moment-timezone meant that the timezone wasn't getting set correctly at the start of period granularity tests. Fixed by setting the timezone through jest config instead.

### DIFF
--- a/packages/web-frontend/src/utils/formatDateForApi.js
+++ b/packages/web-frontend/src/utils/formatDateForApi.js
@@ -5,13 +5,12 @@
  * found in the LICENSE file in the root directory of this source tree.
  */
 
-import moment from 'moment-timezone';
+import moment from 'moment';
 
 const DATE_FORMAT = 'YYYY-MM-DD';
 
-export const formatDateForApi = (date, timezone) => {
+export const formatDateForApi = date => {
   if (!date) return undefined;
   const dateAsMoment = moment(date);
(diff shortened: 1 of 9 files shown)


packages/web-frontend/src/utils/formatDateForApi.js
```diff
@@ -5,13 +5,12 @@
  * found in the LICENSE file in the root directory of this source tree.
  */
 
-import moment from 'moment-timezone';
+import moment from 'moment';
 
 const DATE_FORMAT = 'YYYY-MM-DD';
 
-export const formatDateForApi = (date, timezone) => {
```
tcaiger, line 12: The timezone prop was never being used and removing support for it means that we can completely remove moment-timezone as a dependency from web-frontend.

acdunham: Nice one. There is actually a function in `utils` called `formatDateForApi` which looks like it does the same thing. Might not be worth swapping it out tho, since we are going to eventually get rid of `web-frontend`

acdunham (approved):
Nice one, thanks for doing this!

packages/web-frontend/src/utils/formatDateForApi.js
tcaiger: Good spotting but yeah probably not worth it at this point in web-frontend.

PR #4727, opened 2023-07-12
WAITP-1290: Fix Chart pdf export styles

tcaiger:
### Changes:

- Assume isMobile is false when exporting so that the exports don't include mobile styles


acdunham (approved):
LGTM - thanks for fixing this!

PR #4731, opened 2023-07-13
WAITP-1278 Split out req/res types for tupaia-web-server

EMcQ-BES:
### Changes:

- Split out type definitions for `Request<>` types in `tupaia-web-server`
- Use the newly split out types


IgorNadj (approved):
Nice one, looks great!

PR #4742, opened 2023-07-17
Waitp 1314 trim input fields

chaoxinnhin:
### Issue #:
waitp-1314-trim-input-fields
### Changes:
Added .trim() to the actions.js in the admin-panel folder, in the tables folder. The method was added in line 83 under the filterObject function.


tcaiger (approved):
Thanks

PR #4743, opened 2023-07-17
RN-815: Added support for QR Code scanning for Entity questions in Meditrak App

rohan-bes:
### Issue RN-815:

Screenshots: https://linear.app/bes/issue/RN-815#comment-36b14387

Add a new component `QrCodeScanner` for using QR code data to answer a question

Added that component to the `EntityQuestion` if the `entity.allowScanQrCode` config option is set to `true`.


package.json
```diff
@@ -45,7 +45,8 @@
     "package:eject": "cd $INIT_CWD && react-scripts eject"
   },
   "resolutions": {
-    "@typescript-eslint/eslint-plugin": "^4.18.0"
+    "@typescript-eslint/eslint-plugin": "^4.18.0",
+    "react-native-qrcode-scanner/react-native-permissions": "^3.6.1"
```
rohan-bes, line 49: This one took a while to work out. So the version of `react-native-permissions` in `react-native-qrcode-scanner` is no longer supported https://github.com/moaazsidat/react-native-qrcode-scanner/issues/403

This seems to fix things simply enough, though it would be nice if we could add comments to a `package.json` file explaining why

IgorNadj: Hmm, that repo is archived. Should we jump to another one? Looks like plain old react native camera supports QR code reading: https://react-native-camera.github.io/react-native-camera/docs/rncamera#onbarcoderead and https://react-native-camera.github.io/react-native-camera/docs/rncamera#ongooglevisionbarcodesdetected

IgorNadj: ...which is also archived. The replacement repo is https://github.com/mrousavy/react-native-vision-camera

IgorNadj: Discussed, it's completely understandable that a package we want to add is archived, the rest of our packages are ~7 years old. Happy to leave swapping this out for later.

packages/meditrak-app/app/widgets/Button.js
```diff
@@ -29,12 +42,17 @@ export function Button(props) {
       analyticsLabel={title}
       onPress={onPress}
     >
-      <Text style={[localStyles.text, textStyle]}>{title}</Text>
+      {Label}
```
IgorNadj, line 45: Seems like a change that could affect a bunch of things. Code wise it looks ok, but just wondering if it's best to do now?

IgorNadj (approved):
LGTM, one minor question

packages/meditrak-app/app/widgets/Button.js
rohan-bes: Good thinking. I did a bit of a snoop around during development and I think it's safe. I figure that meditrak will be getting a bit of testing over the next few days so I reckon the QAs will catch any bugs. Happy to see how it goes and fix/rollback if there are issues

PR #4744, opened 2023-07-17
[no-issue]: Fix missing android:exported="true" in AndroidManifest.xml

rohan-bes:
https://stackoverflow.com/questions/70333565/targeting-s-version-31-and-above-requires-that-an-explicit-value-for-android

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ dockercfg
 packages/.vscode/settings.json
 **/testRequests/*.http
 *:Zone.Identifier
+.gradle
 
 # yarn
 .yarn/*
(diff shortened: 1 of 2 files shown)


rohan-bes:
@IgorNadj I ran into this issue trying to run android locally on my device after the upgrade to API 31. I'm not sure if it'll affect appcenter, if so we may need to hotfix the release. Otherwise I'm happy it just going in to dev.

.gitignore
```diff
@@ -25,6 +25,7 @@ dockercfg
 packages/.vscode/settings.json
 **/testRequests/*.http
 *:Zone.Identifier
+.gradle
```
IgorNadj, line 28: Is this intentional?

IgorNadj (approved):

.gitignore
rohan-bes: Yeah. I did a first time build of gradle from the root directory using: `yarn workspace @tupaia/meditrak-app run-android` and it installed it there.

I figured it's fine to have this here and it'll keep things clean if others do what I did 👍

PR #4746, opened 2023-07-18
Waitp 1235 matrix drilldown

acdunham:
### Changes:
- Swapped `useEntity` hook over to single entity endpoint, now that this is in play
- Enabled matrix row links in `ui-components`
- Added drilldown functionality to matrix in `tupaia-web`


packages/tupaia-web/src/api/queries/useEntity.ts
```diff
@@ -14,16 +14,14 @@ export const useEntity = (entityCode?: string) => {
   const { projectCode } = useParams();
 
   return useQuery(
-    ['entities', projectCode, entityCode],
+    ['entity', projectCode, entityCode],
     async (): Promise<EntityResponse> => {
-      const entities = await get(`entities/${projectCode}/${entityCode}`, {
+      const entity = await get(`entity/${projectCode}/${entityCode}`, {
```
acdunham, line 19: I just changed this over here because it was a small change

packages/tupaia-web/src/api/queries/useReport.ts
```diff
@@ -9,7 +9,7 @@ import { get } from '../api';
 import { DashboardItemType, EntityCode, ProjectCode, DashboardsResponse } from '../../types';
 import { formatDateForApi, getBrowserTimeZone } from '@tupaia/utils';
 
-type QueryParams = {
+type QueryParams = Record<string, unknown> & {
```
acdunham, line 12: This is to enable drilldown query params

packages/tupaia-web/src/features/DashboardItem/DashboardItemContent.tsx
```diff
@@ -101,6 +101,8 @@ export const DashboardItemContent = ({
 
   const DisplayComponent = DisplayComponents[type as keyof typeof DisplayComponents] || null;
 
+  if (!DisplayComponent) return null;
```
acdunham, line 104: This is to stop any errors for other viz types that haven't been made yet

packages/tupaia-web/src/utils/useDateRanges.ts
```diff
@@ -112,7 +112,6 @@ export const useDateRanges = (
 
   const { startDate: urlStartDate, endDate: urlEndDate } = convertUrlPeriodStringToDateRange(
     currentPeriodString,
-    periodGranularity,
```
tcaiger, line 115: Why do we no longer need periodGranularity? Won't it always now default to day?

packages/tupaia-web/src/features/DashboardItem/EnlargedDashboardItem.tsx
```diff
@@ -5,15 +5,17 @@
...
+import { FlexColumn, IconButton } from '@tupaia/ui-components';
 import { URL_SEARCH_PARAMS } from '../../constants';
 import { useDashboards } from '../../api/queries';
 import { DashboardItemContent } from './DashboardItemContent';
 import { useDateRanges } from '../../utils';
 import { useReport } from '../../api/queries/useReport';
 import { DateRangePicker, Modal } from '../../components';
+import { DashboardItemType } from '../../types';
 
 const Wrapper = styled.div<{
```
tcaiger, line 20: I'm not sure exactly where it needs to go but can we please add a min-width for the enlarged dashboard items somewhere to match the OG site. When the Matrix only has one row or is loading it shrinks down.

packages/tupaia-web/src/features/DashboardItem/EnlargedDashboardItem.tsx
```diff
@@ -131,6 +196,7 @@ export const EnlargedDashboardItem = () => {
               />
             )}
           </TitleWrapper>
+
```
tcaiger, line 199:
```suggestion
```

packages/tupaia-web/src/features/DashboardItem/EnlargedDashboardItem.tsx
```diff
@@ -84,25 +103,50 @@ export const EnlargedDashboardItem = () => {
...
 
   if (!reportCode || (!isLoadingDashboards && !currentDashboardItem)) return null;
 
-  // // On close, remove the report search param from the url
+  // // On close, remove the report search params from the url
   const handleCloseModal = () => {
     urlSearchParams.delete(URL_SEARCH_PARAMS.REPORT);
     urlSearchParams.delete(URL_SEARCH_PARAMS.REPORT_PERIOD);
+    urlSearchParams.delete(URL_SEARCH_PARAMS.REPORT_DRILLDOWN_ID);
     setUrlSearchParams(urlSearchParams.toString());
```
tcaiger, line 150: I don't think `toString()` is necessary, right?

packages/tupaia-web/src/utils/useDateRanges.ts
acdunham: We do use the period granularity for the display, but when converting to and from the URL search parameters the current way is to use an actual full date range (e.g. 01_April_2023-30_April_2023) as opposed to the display equivalent (i.e.g Apr_2023) because the query won't work with the display version

packages/tupaia-web/src/features/Matrix.tsx
```diff
@@ -125,19 +148,22 @@ interface MatrixProps {
...
 
   const placeholderImage = getPlaceholderImage(config);
-  // in the dashboard, show a placeholder image
-  if (!isEnlarged) return <img src={placeholderImage} alt="Matrix Placeholder" />;
 
-  const parsedRows = parseRows(rows, undefined, searchFilter);
+  const { periodGranularity, drillDown } = config;
+
+  const baseDrillDownLink = getBaseDrilldownLink(drillDown);
+  const parsedRows = parseRows(rows, undefined, searchFilter, drillDown, baseDrillDownLink);
```
tcaiger, line 161: sheesh there is a lot going on in this component now but I think that it all fits in pretty nicely.

packages/tupaia-web/src/api/queries/useReport.ts
```diff
@@ -20,7 +20,16 @@ type QueryParams = {
...
-  const { dashboardCode, projectCode, entityCode, itemCode, startDate, endDate, legacy } = params;
+  const {
+    dashboardCode,
+    projectCode,
+    entityCode,
+    itemCode,
+    startDate,
+    endDate,
+    legacy,
+    ...rest
```
tcaiger, line 31: So rest is just always `reportDrillDownId` right? Assuming so, can we explicitly add that to the list?

acdunham: In the URL it is `reportDrilldownId` but in reality when we use `parameterLink` from the parent it could be any property name, e.g. `eventId` etc

tcaiger: Oh now I see what's going on. That's the key field from the parent data. Ok please ignore my question.

packages/tupaia-web/src/features/DashboardItem/EnlargedDashboardItem.tsx
```diff
@@ -112,11 +156,32 @@ export const EnlargedDashboardItem = () => {
 
   const { type } = currentDashboardItem || {};
 
+  // If the report is a drilldown, we want to show a back button to the parent dashboard item
+  const getBackLink = () => {
```
tcaiger, line 160: I was trying to think of some way to split out the drill down code as it makes this component quite confusing and it's in danger of becoming a scary component to touch (the OG component in the OG site is terrifying). One quick win would be to make a BackButton component in this file and move this back link logic into it.

tcaiger (changes requested):
Nice work. It's a great milestone to have this done. I think this is the most complicated piece of ui in the whole site.

packages/ui-components/src/components/Matrix/MatrixRow.tsx
```diff
@@ -139,6 +153,7 @@ export const MatrixRow = ({ row, parents = [] }: MatrixRowProps) => {
             rowTitle={row.title}
             colKey={key}
             isCategory={isCategory}
+            link={link}
```
tcaiger, line 156: Sorry one more comment. I think this makes every Cell a clickable link, right? Only the first column should be clickable.

tcaiger (approved):
LGTM

PR #4752, opened 2023-07-19
No issue: Fix navigation bugs

acdunham:
### Changes:
- Handle when default overlay for a project is not a valid option (select first available overlay if possible, or else use default overlay code)
- Fix error when modal closes when a hash is in the URL before the default overlay gets added to the URL
- Redirect back to projects modal when being directed to login from there
- Open projects modal when going directly to default project, on first load only
- Fix errors that cause the dashboard to crash the app when the user attempts to go directly to a URL that they don't have permissions for (Note: other login handling around this will be a separate PR as depends on some backend changes)
- Show a loader on the projects modal when fetching projects to give the user some visual feedback on what's happening


packages/tupaia-web/src/layout/MainLayout.tsx
```diff
@@ -31,7 +31,24 @@ const Container = styled.div`
...
+      (location.pathname === DEFAULT_URL || location.pathname === DEFAULT_PROJECT_ENTITY) &&
+      !location.hash
+    ) {
+      navigate({
+        ...location,
+        hash: MODAL_ROUTES.PROJECTS,
+      });
+    }
+  }, []);
+};
```
acdunham, line 49: I was on the fence about whether this should be inside the component or outside, but went with outside because it makes the component less busy to read

packages/tupaia-web/src/layout/ProjectCardList/LegacyProjectCard.tsx
```diff
@@ -105,13 +106,15 @@ const StyledPendingButton = styled(RouterButton).attrs({
...
   children,
+  routerState,
 }: {
-  url: string;
+  to: To;
   children: ReactNode;
+  routerState?: Record<string, unknown> | null;
 }) => (
-  <RouterButton to={url} color="primary" variant="outlined">
+  <RouterButton to={to} color="primary" variant="outlined" routerState={routerState}>
```
acdunham, line 117: There was a weird issue where if we used a URL `string` instead of a `To` object, the location updated multiple times and removed the router `state`, so I changed it to be a `To` object

packages/tupaia-web/src/features/Map/utils/useDefaultMapOverlay.ts
```diff
@@ -32,14 +34,27 @@ export const useDefaultMapOverlay = (
     const isValidMapOverlayId = !!mapOverlaysByCode[selectedMapOverlay!];
 
     if (!selectedMapOverlay || !isValidMapOverlayId) {
-      const defaultMapOverlayId = project.defaultMeasure || DEFAULT_MAP_OVERLAY_ID;
+      const { defaultMeasure } = project;
+      const overlayCodes = mapOverlaysByCode ? Object.keys(mapOverlaysByCode) : [];
+      // If there are no map overlays, the default overlay is the DEFAULT_MAP_OVERLAY_ID
+      // however if there are map overlays, use the first available overlay
+      let defaultMapOverlayId = overlayCodes.length > 0 ? overlayCodes[0] : DEFAULT_MAP_OVERLAY_ID;
```
tcaiger, line 41: Thanks for picking this up - looks good!

tcaiger: Oh actually I think it's still not quite right. A default map overlay is getting assigned when there is no map overlay data and I don't think that ever makes sense. What about this?

```
useEffect(() => {
    const isValidMapOverlayId = !!mapOverlaysByCode[selectedMapOverlay!];
    const overlayCodes = mapOverlaysByCode ? Object.keys(mapOverlaysByCode) : [];

    if (!project || overlayCodes.length === 0) {
      return;
    }

    if (!selectedMapOverlay || !isValidMapOverlayId) {
      const { defaultMeasure } = project;

      if (mapOverlaysByCode[defaultMeasure]) {
        urlSearchParams.set(URL_SEARCH_PARAMS.MAP_OVERLAY, defaultMeasure);
      }

      if (mapOverlaysByCode[DEFAULT_MAP_OVERLAY_ID]) {
        urlSearchParams.set(URL_SEARCH_PARAMS.MAP_OVERLAY, DEFAULT_MAP_OVERLAY_ID);
      }
    }

    if (!selectedMapOverlayPeriod) {
      urlSearchParams.set(URL_SEARCH_PARAMS.MAP_OVERLAY_PERIOD, DEFAULT_PERIOD_PARAM_STRING);
    }

    // we have to navigate using navigate instead of setUrlParams because setUrlParams seems to override any existing params and existing hash
    navigate({
      ...location,
      search: urlSearchParams.toString(),
    });
  }, [JSON.stringify(mapOverlaysByCode), project, selectedMapOverlay]);

File: packages/tupaia-web/src/layout/MainLayout.tsx (hunk 1 above)
Comment (tcaiger) [reply packages/tupaia-web/src/layout/MainLayout.tsx:49]: I think this is not quite right because it will only work if the user refreshed on to the home page and in the case of logging in they have moved there so I think this should be moved to the login success handler instead. eg.

```
if (location.state?.referrer) {
    navigate(location.state.referrer, {
      state: null,
    });
  } else if (
    location.pathname === DEFAULT_URL ||
    location.pathname === DEFAULT_PROJECT_ENTITY
  ) {
    navigateToModal(MODAL_ROUTES.PROJECTS);
  } else {
    closeModal();
 }
}

tcaiger (changes requested):
Thanks for making all these changes. I left a couple of suggestions relating to the url behaviour. Let me know if you have any questions.

packages/tupaia-web/src/layout/MainLayout.tsx
acdunham: How about when the user is already logged in? Then it won't happen?

tcaiger (approved):
Great changes! Thanks

PR #4760, opened 2023-07-21
Waitp 1299 link to project new tab

chaoxinnhin:
### Issue #:
WAITP-1299 Custom landing page: Open link to project in new tab
### Changes:

- Added target to BaseLink component in tupaia-web/src/layout/ProjectCardList/ProjectCard.tsx line 133,

- Added target to ActionLink component in tupaia-web/src/views/LandingPage/SingleProjectLandingPage.tsx line 104

### Screenshots:
image

### DIFF
--- a/packages/tupaia-web/src/components/RouterButton.tsx
+++ b/packages/tupaia-web/src/components/RouterButton.tsx
@@ -25,6 +25,7 @@ export const RouterButton = ({
   modal,
   searchParamsToRemove,
   children,
+  url,
   ...props
 }: RouterButtonProps) => {
   const location = useLocation();
@@ -44,7 +45,7 @@ export const RouterLink = ({ to, modal, children, ...props }: RouterButtonProps)
   const link = modal ? { ...location, hash: modal } : to;
 
   return (
-    <Link to={link} {...props}>
+    <Link to={link} {...props} target='_newtab'>
       {children}
     </Link>
   );
--- a/packages/tupaia-web/src/components/RouterButton.tsx
+++ b/packages/tupaia-web/src/components/RouterButton.tsx
@@ -45,7 +45,7 @@ export const RouterLink = ({ to, modal, children, ...props }: RouterButtonProps)
   const link = modal ? { ...location, hash: modal } : to;
 
   return (
-    <Link to={link} {...props} target='_newtab'>
+    <Link to={link} {...props}>
       {children}
     </Link>
   );
--- a/packages/tupaia-web/src/components/RouterButton.tsx
+++ b/packages/tupaia-web/src/components/RouterButton.tsx
@@ -25,7 +25,6 @@ export const RouterButton = ({
   modal,
   searchParamsToRemove,
   children,
-  url,
   ...props
 }: RouterButtonProps) => {
   const location = useLocation();
--- a/packages/tupaia-web/src/layout/ProjectCardList/ProjectCard.tsx
+++ b/packages/tupaia-web/src/layout/ProjectCardList/ProjectCard.tsx
@@ -13,7 +13,6 @@ import { MODAL_ROUTES } from '../../constants';
 import { RouterButton } from '../../components';
 import { useLandingPage } from '../../api/queries';
 import { useParams } from 'react-router';
-import { LandingPage } from '../../views';
 
 const Card = styled.div`
   display: flex;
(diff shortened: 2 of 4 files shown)


packages/tupaia-web/src/components/RouterButton.tsx
```diff
@@ -44,7 +45,7 @@ export const RouterLink = ({ to, modal, children, ...props }: RouterButtonProps)
   const link = modal ? { ...location, hash: modal } : to;
 
   return (
-    <Link to={link} {...props}>
+    <Link to={link} {...props} target='_newtab'>
```
acdunham, line 48: I probably wouldn't put this here, because this means anywhere that uses `RouterLink` will open everything in another tab

packages/tupaia-web/src/views/LandingPage/SingleProjectLandingPage.tsx
```diff
@@ -100,6 +101,7 @@ export function SingleProjectLandingPage({
       {accessType && (
         <ActionLink
           variant="contained"
+          target= {`${useLandingPage()}`}
```
acdunham, line 104: I'm not sure what this is doing - `useLandingPage` should be returning an object of data, loading state etc. Instead, it would be more ideal if we could just set `target='_blank'` on the link

packages/tupaia-web/src/layout/ProjectCardList/ProjectCard.tsx
```diff
@@ -127,8 +128,9 @@ export const ProjectPendingLink = () => (
     Approval in progress
   </OutlineLink>
 );
+
 export const ProjectAllowedLink = ({ url }: LinkProps) => (
-  <BaseLink to={url}>View project</BaseLink>
+  <BaseLink to={url} target={`${useLandingPage(url)}`}>View project</BaseLink>
```
acdunham, line 133: One thing to remember is that these cards and links will be reused by the Projects modal later on, and we don't want them to open in a new tab from there. Ideally the `ProjectAllowedLink` would take a param that flags whether it should open in a new tab or not, and apply `target='_blank'` when that condition is met

acdunham (changes requested):
This is close, but not quite right for our use case. I have added some comments :)

packages/tupaia-web/src/layout/ProjectCardList/ProjectCard.tsx
```diff
@@ -127,9 +129,18 @@ export const ProjectPendingLink = () => (
     Approval in progress
   </OutlineLink>
 );
-export const ProjectAllowedLink = ({ url }: LinkProps) => (
-  <BaseLink to={url}>View project</BaseLink>
-);
+
+export const ProjectAllowedLink = ({ url }: LinkProps) => {
```
acdunham, line 133: Sorry, this was my bad, I should have mentioned, you can just get `ProjectAllowedLink` to accept another param, as well as `url`, to be like `isLandingPage` or something like that and set the target based on that value. Then in the landing pages, where this is used, pass in `isLandingPage` as `true`

acdunham (changes requested):
Nice work. I just have one teeny change requested, but other than that it looks good

packages/tupaia-web/src/layout/ProjectCardList/ProjectCard.tsx
```diff
@@ -127,8 +128,11 @@ export const ProjectPendingLink = () => (
     Approval in progress
   </OutlineLink>
 );
-export const ProjectAllowedLink = ({ url }: LinkProps) => (
-  <BaseLink to={url}>View project</BaseLink>
+
+export const ProjectAllowedLink = ({ url, isLandingPage }: LinkProps) => (
+  <BaseLink to={url} target={isLandingPage}>
```
acdunham, line 133: So with this one, this is on the right track. But you can even make `isLandingPage` just a `boolean` and when it is true, `target` will be `_blank`, otherwise it is `_self`. In this case, passing in a prop for `target` works but the naming convention isn't quite right because it implies a boolean but is a string. Personally I'd prefer the approach of just passing in `true/false` for `isLandingPage` and the `ProjectAllowedLink` then handling the `target` from that.

packages/tupaia-web/src/views/LandingPage/MultiProjectLandingPage.tsx
```diff
@@ -74,6 +75,9 @@ export function MultiProjectLandingPage({
   projects,
   includeNameInHeader,
 }: MultiProjectLandingPageProps) {
+  const { landingPageUrlSegment } = useParams();
+  const { isLandingPage } = useLandingPage(landingPageUrlSegment);
```
acdunham, line 79: You don't really need this check in here, because in this view it will always be a landing page, and on the projects modal it will never be a landing page

acdunham (changes requested):
Sorry to be a pain, but there are a couple improvements I think we could make here. Technically what you've done works, but I think we can make it a bit clearer.

acdunham (approved):
🎉

PR #4761, opened 2023-07-21
Waitp 1320 organisation name style

chaoxinnhin:
### Issue #:
WAITP-1320: Custom Landing Page: Update organisation name style
### Changes:

-  Changed font-size to 1.25rem in tupaia-web/src/layout/TopBar/Logo.tsx on line 41, 48,
-  Changed Name variant="h1" to Name variant="h6"  in tupaia-web/src/layout/TopBar/Logo.tsx on line 82

### Screenshots:
image


packages/tupaia-web/src/layout/TopBar/Logo.tsx
```diff
@@ -79,7 +79,7 @@ export const Logo = ({ logoSrc, displayName = false, name = '' }: LogoProps) =>
         {/** If a custom landing page has set to display the name in the header, display it here */}
         {displayName && (
           <NameWrapper>
-            <Name variant="h1">{name}</Name>
+            <Name variant="h6">{name}</Name>
```
acdunham, line 82: The way that heading levels work is that they are by hierarchy, and not by size. They naturally have different sizes, but it's not considered good practice to use a particular heading level for styling purposes. You should always have the correct order, and not have, for example, `h6` without an `h1`-`h5` ahead of it. In this case, we want it to be `h1` because it is the page title. 
Instead, just setting the font size on the `Name` component would be better.

acdunham (changes requested):
I think maybe this branch was made from your other branch, which means if we were to merge this in, it would bring these changes with it. You'll likely need to make a new branch off dev and do your changes off that so we don't have that issue. I've also left a comment on the font-size change :)

PR #4762, opened 2023-07-22
WAITP-1334: fix tupaia-web fe build

tcaiger:
### Changes:

The tupaia-web build is failing because there is some webpack config coming through one of our dependency imports. The dependency is ace-builds which is used in the JSON editor and SQLEditor components which are only used in the admin panel. I'm  fixing it here by moving those components into the admin-panel package. The change has the added benefits of reducing the dist bundle size of tupaia-web by about 1.5MB and removing some complex code from ui-components that doesn't need to be in a shared package.


packages/admin-panel/package.json
```diff
@@ -59,7 +62,9 @@
     "prop-types": "^15.6.0",
     "query-string": "^5.1.1",
     "react": "^16.13.1",
+    "react-ace": "^10.1.0",
     "react-autobind": "^1.0.6",
+    "react-beautiful-dnd": "^13.1.0",
```
tcaiger, line 67: these are all moved from ui-components

acdunham (approved):
Thanks so much for doing this!

PR #4765, opened 2023-07-23
Waitp 1335 social feed update img upload

WDuncker:
### Changes:

- updated image upload functionality to match base64 upload on projects page.
- Built CreateFeedItems handler in central server feedItems endpoint.
- Built EditFeedItems handler in central server feedItems endpoint.
- Updated routes in index file to use new handlers.

---


EMcQ-BES:
@WDuncker Looks like you've got some file permission chaff clogging up this PR. Another of the lovely Linux/Windows issues.

Try running this command so that git ignores filemode changes in future (You'll still need to do some manual clean up on this PR):

```
git config core.filemode false
```

packages/central-server/src/apiV2/feedItems/EditFeedItems.js
```diff
@@ -0,0 +1,50 @@
...
+        image: existingImage,
+        body: existingBody,
+        link: existingLink,
+      },
+    } = await this.getFields();
+
+    if (base64Image !== undefined) {
+      updatedFields.template_variables = {
+        title: updatedTitle || existingTitle,
+        image: await uploadImage(base64Image, existingImage),
```
EMcQ-BES, line 43: The params here don't quite match the function.

```
export const uploadImage = async (
  encodedImage,
  uniqueId,
  imageSuffix,
  useTimestamp = false,
  existingImagePath,
)
```

From glancing at that function I think all your images are going to be uploaded with the name `undefined_undefined` which is probably not what we want. Let's add `uniqueId` and `imageSuffix` values to both uses of `uploadImage`

packages/central-server/src/apiV2/feedItems/CreateFeedItems.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    await this.insertImagePath(this.models, image, newFeedItem);
+  }
+
+  async insertImagePath(models, base64Image, newFeedItem) {
+    const update = {
+      template_variables: {
+        ...newFeedItem.template_variables,
+        image: await uploadImage(base64Image),
```
EMcQ-BES, line 49: See my other comment about `uniqueId`/`imageSuffix`

packages/admin-panel/src/editor/Editor.js
```diff
@@ -26,19 +25,14 @@ export const Editor = ({ fields, recordData, onEditField, onSetFormFile }) => {
   }
 
   const onInputChange = async (inputKey, inputValue, editConfig = {}) => {
-    const { setFieldsOnChange, type } = editConfig;
-    let updatedValue = inputValue;
-    if (type === 'image' && inputValue) {
-      // If the input is a file, we need to convert the file to a base64 encoded image.
-      updatedValue = await createBase64Image(inputValue);
-    }
+    const { setFieldsOnChange } = editConfig;
```
EMcQ-BES, line 28: Uh... Idk what's going on here, 😅 @acdunham can you confirm these changes are good?

packages/central-server/src/apiV2/feedItems/CreateFeedItems.js
```diff
@@ -0,0 +1,54 @@
...
+      permission_group_id,
+      creation_date,
+      template_variables: { title, image, body, link },
+      type,
+      geographical_area_id,
+      user_id,
+      record_id,
+    } = this.newRecordData;
+
+    const newFeedItem = await this.models.feedItem.create({
```
EMcQ-BES, line 26: Can you take a look into wrapping this in a transaction, similar to the way it happens in the projects handler? It'd be good here since we don't know if the image upload will succeed, so we'd prefer not to have a half baked feed item pointing to a broken image.

EMcQ-BES (changes requested):
A couple small fixes but overall structure looking good 👍

packages/admin-panel/src/editor/Editor.js
acdunham: They are all good :) Although... @WDuncker did you have any changes with `ui-components` to go with these FE changes? i think the `ui-components` package should now be handling the conversion to bas64 data image?

packages/central-server/src/apiV2/feedItems/CreateFeedItems.js
```diff
@@ -0,0 +1,56 @@
...
+      creation_date,
+      template_variables: { title, image, body, link },
+      type,
+      geographical_area_id,
+      user_id,
+      record_id,
+    } = this.newRecordData;
+
+    await this.models.wrapInTransaction(async () => {
+      const newFeedItem = await this.models.feedItem.create({
```
EMcQ-BES, line 27: Hmm. I'm not 100% sure what the behaviour would be here, but you're not using `wrapInTransaction` in our standard way (using `transactingModels`). I'm not sure if using `this.models` inside the `wrapInTransaction` block is going to correctly run as a db transaction 🤔

```suggestion
    await this.models.wrapInTransaction(async transactingModels => {
      const newFeedItem = await transactingModels.feedItem.create({
```

packages/central-server/src/apiV2/feedItems/CreateFeedItems.js
```diff
@@ -0,0 +1,56 @@
...
+          body,
+          link,
+        },
+        type,
+        geographical_area_id,
+        user_id,
+        record_id,
+      });
+
+      await this.insertImagePath(this.models, image, newFeedItem);
```
EMcQ-BES, line 43:
```suggestion
      await this.insertImagePath(transactingModels, image, newFeedItem);
```

EMcQ-BES (changes requested):
An improvement, just some final tweaks

EMcQ-BES (approved):
Awesome, I reckon this is looking good 👍

WDuncker:
In the uploadImage function for the editFeedItems handler I have se the add timestamp to true. This is important only for the edit handler as without it the image field is still rendering the old image rather than the new one unless the page is manually refreshed (perhaps a caching issue). With the add timestamp set to true, the page automatically re-renders the image field with the new image.

PR #4769, opened 2023-07-24
No-issue: Fix fe ts lint config

tcaiger:
### Changes:

Create a new frontend config for typescript packages and apply it to the new typescript front end packages.

### DIFF
--- a/.eslintrc-ts-frontend.json
+++ b/.eslintrc-ts-frontend.json
@@ -1,0 +1,12 @@
+{
+  "extends": ["./.eslintrc-ts.json"],
+  "env": {
+    "browser": true
+  },
+  "rules": {
+    "jsx-a11y/href-no-hash": "off",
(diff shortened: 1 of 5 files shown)


.eslintrc-ts-frontend.json
```diff
@@ -0,0 +1,12 @@
...
+  "env": {
+    "browser": true
+  },
+  "rules": {
+    "jsx-a11y/href-no-hash": "off",
+    "react/destructuring-assignment": "off",
+    "react/no-unescaped-entities": "off",
+    "import/extensions": "off"
+  }
+}
```
acdunham, line 12: I like this reusable setting. Just want to check, are we going to miss out on some useful things by not using the `react-app` lint settings?

acdunham (approved):
This looks good, but I just have a question on one of the files

PR #4787, opened 2023-07-26
WAITP-1247 Fixup auth on export route

EMcQ-BES:
### Changes:

- Forward authorization headers if we're using them, rather than always assume we're using a session


tcaiger (approved):
LGTM thanks

PR #4788, opened 2023-07-26
Waitp 1237 export dashboard item

acdunham:
### Changes:
- Reorganised enlarged dashboard item to be multiple smaller components
- Added context for exporting dashboard item to pass state around the enlarged dashboard item components
- Enabled exporting for single dashboard items (charts and matrices)
- Updated types to include export settings in presentation options

Note: Merged in [the BE PR](https://github.com/beyondessential/tupaia/pull/4787) to test matrix exports, so those changes are in this PR but aren't really part of it

---

### Screenshots:
Matrix export view
Screenshot 2023-07-27 at 8 21 22 am

PNG chart export6 preview
Screenshot 2023-07-27 at 8 22 24 am


packages/web-config-server/src/export/ExportSurveyDataHandler.js
```diff
@@ -51,6 +53,10 @@ export class ExportSurveyDataHandler extends RouteHandler {
       queryParameters,
       sessionCookieName,
       sessionCookie,
+      true,
+      {
+        authorization: authHeader,
+      }
     );
```
acdunham, line 60: This change is actually from the BE PR

packages/web-config-server/src/export/exportChartHandler.js
```diff
@@ -52,6 +54,10 @@ export const exportChartHandler = async (req, res) => {
     queryParameters,
     sessionCookieName,
     sessionCookie,
+    true,
+    {
+      authorization: authHeader,
+    }
   );
```
acdunham, line 61: This change is actually from the BE PR

packages/web-config-server/src/export/requestFromTupaiaConfigServer.js
```diff
@@ -18,10 +18,12 @@ export const requestFromTupaiaConfigServer = async (
   sessionCookieName,
   sessionCookie,
   shouldRetryOnFail = true,
+  additionalHeaders = {},
 ) => {
   try {
     const headers = {
       cookie: `${sessionCookieName}=${sessionCookie}`,
+      ...additionalHeaders
     };
```
acdunham, line 27: This change is actually from the BE PR

packages/tupaia-web/src/features/EnlargedDashboardItem/utils/ExportContext.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { Dispatch, createContext } from 'react';
```
tcaiger, line 6: I have a new found fondness for reducers after seeing these context reducers coming out so nicely.

packages/tupaia-web/src/api/queries/useExportToExcel.ts
```diff
@@ -0,0 +1,38 @@
...
+    endDate?: string;
+    dataElementHeader?: string;
+    legacy?: boolean;
+  },
+  enabled?: boolean,
+) => {
+  const timeZone = getBrowserTimeZone();
+  const { projectCode, entityCode, itemCode, startDate, endDate } = params;
+  return useQuery(
+    ['export/chart', projectCode, entityCode, itemCode, startDate, endDate],
```
tcaiger, line 26:
```suggestion
    ['export', 'chart', projectCode, entityCode, itemCode, startDate, endDate],
```

packages/tupaia-web/src/features/EnlargedDashboardItem/EnlargedDashboardItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+const Wrapper = styled.div<{
+  $hasBigData?: boolean;
+  $applyWidth?: boolean;
+}>`
+  max-width: 100%;
+  min-width: ${({ $hasBigData, $applyWidth }) => ($applyWidth && $hasBigData ? '90vw' : 'auto')};
+  width: ${({ $hasBigData, $applyWidth }) => {
+    if (!$applyWidth) return 'auto';
+    return $hasBigData ? '90%' : '48rem';
+  }};
```
tcaiger, line 33:
```suggestion
  min-width: ${({ $hasBigData }) => ($hasBigData ? '90vw' : 'auto')};
  width: ${({ $hasBigData }) => {
    if (!$applyWidth) return 'auto';
    return $hasBigData ? '90%' : '48rem';
  }};
```
This can be simplified a bit now right? And is there any way of getting rid of applyWidth altogether?

acdunham: We can, but we would have to have the data download modal also as 48rem, is that okay?

packages/tupaia-web/src/features/EnlargedDashboardItem/utils/useExportDashboardItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+        exportWithLabels,
+        exportWithTable,
+        exportWithTableDisabled,
+      },
+    } as ViewContent,
+    exportTitle,
+  );
+
+  // use the useExportToExcel hook to export the matrix dashboard item. We disable the fetch initially so that we can only do the query when we want to export
+  const { error: excelError, refetch: callExportToExcel, isFetching } = useExportToExcel(
```
tcaiger, line 112: I think useExportToExcel should be a [mutation](https://tanstack.com/query/v3/docs/react/guides/mutations) since the api request gets triggered by a user click (I think it counts as a server-side-effect in react-query vernacular). Them we wouldn't have to use the refetch function to make the request which doesn't quite seem right.

packages/tupaia-web/src/features/EnlargedDashboardItem/utils/useExportDashboardItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+    }
+  };
+
+  const EXPORT_FUNCTIONS = {
+    [EXPORT_FORMATS.PNG]: exportToPNG,
+    [EXPORT_FORMATS.XLSX]: type === 'matrix' ? exportMatrixToExcel : doExport,
+  };
+
+  // set export error message if an error is caught from the useExportToExcel hook
+  useEffect(() => {
```
tcaiger, line 161: Once the useExport hook is refactored to a mutation there should be a nicer way to do these side effects such as the onError handler or you can use a try catch block if that suits better.

tcaiger (changes requested):
Such good work - so much good design and code in here. I just have the one request to convert useExport to be a mutation. Thanks

packages/tupaia-web/src/features/EnlargedDashboardItem/EnlargedDashboardItem.tsx
tcaiger: I vote yes

packages/tupaia-web/src/features/EnlargedDashboardItem/utils/useExportDashboardItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+    currentDashboardItem,
+  });
+
+  const EXPORT_FUNCTIONS = {
+    [EXPORT_FORMATS.PNG]: exportToPNG,
+    [EXPORT_FORMATS.XLSX]: type === 'matrix' ? () => exportToExcel(excelParams) : doExport,
+  };
+
+  // set export error message as null and send ga event when the export is successful/starting
+  useEffect(() => {
```
tcaiger, line 175: This side effect can just be moved into `handleStartExport` right?

packages/tupaia-web/src/features/EnlargedDashboardItem/utils/useExportDashboardItem.tsx
```diff
@@ -0,0 +1,185 @@
...
+    if (exportError || !isExportMode) return;
+    gaEvent('Export', file, isExporting ? 'Attempt' : 'Success');
+  }, [isExporting]);
+
+  // reset the export state when the current dashboard item changes
+  useEffect(() => {
```
tcaiger, line 181: I am on the fence about this one but I was reading the useEffect docs recently and it recommends a different approach for [resetting state based on props](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes). I'd be interested to hear which you think is best?

acdunham: I'm not sure about that one. I don't really like the idea of having to maintain our own state with previous and current props. It seems like reinventing the wheel, when that's what `useEffect` watches anyway. What do you think?

packages/tupaia-web/src/features/EnlargedDashboardItem/EnlargedDashboardItem.tsx
acdunham: Done

packages/tupaia-web/src/features/EnlargedDashboardItem/utils/useExportDashboardItem.tsx
tcaiger: Yes I agree with you that it seems quite clunky to keep the prev props in state so let's stay with the useEffect 👍

tcaiger (approved):
🎉

PR #4791, opened 2023-07-27
RN-957 package resolution

IgorNadj:


IgorNadj:
Will include this in a larger PR for RN-955

PR #4795, opened 2023-07-27
MDEV-174: Add fillRows transform to the viz-builder

rohan-bes:
### Issue MDEV-174:

Added a new transform to fill in rows in the table where we want there to be data for given columns.


package.json
```diff
@@ -38,8 +38,8 @@
     "package:build:react-scripts": "cd $INIT_CWD && SKIP_PREFLIGHT_CHECK=true GENERATE_SOURCEMAP=false react-scripts build",
     "package:build:types": "cd $INIT_CWD && tsc",
     "package:lint": "cd $INIT_CWD && eslint .",
-    "package:test": "cd $INIT_CWD && jest",
-    "package:test:withdb": "cd $INIT_CWD && yarn workspace @tupaia/database check-test-database-exists && DB_NAME=tupaia_test jest",
+    "package:test": "cd $INIT_CWD && jest --maxWorkers=50%",
+    "package:test:withdb": "cd $INIT_CWD && yarn workspace @tupaia/database check-test-database-exists && DB_NAME=tupaia_test jest --maxWorkers=50%",
```
rohan-bes, line 42: Found that setting `maxWorkers=50%` decently improved test run time and prevented my machine from slowing down during them. This seems like a common phenomenon: https://blog.developer.adobe.com/improve-jest-runner-performance-a8f56708ba94

packages/expression-parser/src/expression-parser/customFunctions.js
```diff
@@ -32,6 +33,8 @@ const translate = (value, translations) => {
 
 const date = (...argumentList) => new Date(...argumentList);
 
+const dateToString = date => fns.format(date, 'yyyy-MM-dd');
```
rohan-bes, line 36: There's probably a greater conversion about standardising date strings in tupaia to be had. But for now adding this utility function I think will help

packages/report-server/src/__tests__/reportBuilder/transform/fillRows.test.ts
```diff
@@ -0,0 +1,282 @@
...
+        },
+      },
+    ]);
+    expect(await transform(TransformTable.fromRows(FILLABLE_ANALYTICS))).toStrictEqual(
+      TransformTable.fromRows([
+        { period: '20191231', organisationUnit: 'org unit 1', BCD1: 0 },
+        { period: '20200101', organisationUnit: 'TO', BCD1: 1 },
+        { period: '20200102', organisationUnit: 'TO', BCD1: 2 },
+        { period: '20200103', organisationUnit: 'TO', BCD1: 3 },
+        { period: '20200104', organisationUnit: 'org unit 3', BCD1: 3 }, // Indexing is a little awkward due to not knowing rows are being filled
```
rohan-bes, line 275: Don't love that the new index doesn't get determined dynamically. But it's consistent with `insertRows` and the lookups in the viz-builder don't get used too heavily and are in need of a re-write anyway.

packages/report-server/src/reportBuilder/transform/functions/fillRows/fillRows.ts
```diff
@@ -0,0 +1,76 @@
...
+import { FieldValue, Row } from '../../../types';
+import { Context } from '../../../context';
+import { TransformParser } from '../../parser';
+import { TransformTable } from '../../table';
+import { checkColumnsExistInTable, expressionOrArrayValidator } from '../utils';
+import { generateRowInserts } from './generateRowInserts';
+
+type FillRowsParams = {
+  requiredColumnValues: { column: string; values: string | FieldValue[] }[];
+  missingRowValues: Row;
```
rohan-bes, line 17: Arguably this field should be called `columns` instead of `missingRowValues` to be consistent with `insertRows`.
However, I think that's pretty confusing in this context, and I don't really feel it's too intuitive in `insertRows` either...

I've got Nikhita and Gagan having a play with this transform and I'll get their feedback on the naming of the fields.

package.json
```diff
@@ -38,8 +38,8 @@
     "package:build:react-scripts": "cd $INIT_CWD && SKIP_PREFLIGHT_CHECK=true GENERATE_SOURCEMAP=false react-scripts build",
     "package:build:types": "cd $INIT_CWD && tsc",
     "package:lint": "cd $INIT_CWD && eslint .",
-    "package:test": "cd $INIT_CWD && jest",
-    "package:test:withdb": "cd $INIT_CWD && yarn workspace @tupaia/database check-test-database-exists && DB_NAME=tupaia_test jest",
+    "package:test": "cd $INIT_CWD && jest --maxWorkers=50%",
```
rohan-bes, line 41: Set maxWorkers=50% here since it lead to a decent improvement in test performance in my local. These results seem pretty common: https://blog.developer.adobe.com/improve-jest-runner-performance-a8f56708ba94

packages/report-server/src/reportBuilder/transform/functions/fillRows/ExpectedRows.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (this.counts[i] >= this.maxCounts[i]) {
+        this.counts[i] = 0;
+        this.counts[i - 1] += 1;
+      } else {
+        break;
+      }
+    }
+
+    return row;
+  }
```
edmofro, line 69: This is super smart, and absolutely does my head in 😅 
I find it hard to follow because it is so clever. 

Any appetite for changing to a mechanism that pregenerates all the expected rows into an object (i.e. actually spits out all the permutations), and then just iterates through that when calling `next()`?

edmofro (changes requested):
This is really nice stuff!

packages/report-server/src/reportBuilder/transform/functions/fillRows/ExpectedRows.ts
edmofro: Ok I've had a go at this, and yeah, it's very hard to keep it sane without using recursion 😂 
All good as is! The complexity is well-contained

rohan-bes: Thanks for giving it a crack 🙏 yeah I really did try and make this cleaner, but it was proving hard. And recursion really blows out the performance.

I'll add some comments tho which I hope will improve things. It's not the most complex thing conceptually, but I think the naming and lack of comments makes it harder to understand.

packages/report-server/src/reportBuilder/transform/functions/utils/rowValuesKey.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { Row } from '../../../types';
+
+const divider = '___';
+
+export const rowValuesKey = (row: Row, columnsOfInterest?: string[]) => {
```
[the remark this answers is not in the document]
rohan-bes: Yeah good call 👍

packages/report-server/src/reportBuilder/transform/functions/fillRows/ExpectedRows.ts
rohan-bes: Btw, could be a fun code change if we're looking for fresh interview questions 😉

edmofro (approved):
Muuuch easier to follow, thanks for taking the time to improve the naming/structure/comments 🙏

package.json
IgorNadj: Weird, good catch. I get about a 15% speedup too on my M1 mac

PR #4799, opened 2023-07-28
Add conditional style to StyledTable.tsx to render white text if in d…

WDuncker:
…ark theme

### Changes:

-Added conditional statement to StyledTable.tsx in ui-components to render blue text in light theme and white text in dark theme.

---

### Note:
-Unable to replicate the bug due to potential permissions issue. Can you check this against the table shown on the issue card in Linear please?


tcaiger (approved):
This looks good to me 🎉

PR #4802, opened 2023-07-30
Waitp 1273 update project cards

chaoxinnhin:
### Issue #:
WAITP-1273: Projects Modal: Update project cards
### Changes:

- Changed all px above 5px to rem in tupaia-web/src/views/ProjectsModal.tsx, tupaia-web/src/layout/ProjectCardList/ProjectCard.tsx and tupaia-web/src/layout/ProjectCardList/LegacyProjectCard.tsx
- Added a StyledRouterButton in tupaia-web/src/layout/ProjectCardList/LegacyProjectCard.tsx, on line 98 to change to look of the LegacyProjectAllowedLink
- Created and added ProjectsTitle in tupaia-web/src/views/ProjectsModal.tsx on line 75 and line 117
- Added text-transform: none in tupaia-web/src/views/ProjectsModal.tsx on line 66 and text-align in line 65
- Increased the ProjectsGrid gap in tupaia-web/src/views/ProjectsModal.tsx on line 45
- Added text-align: left in tupaia-web/src/views/ProjectsModal.tsx on line 33
- Created and added line in tupaia-web/src/views/ProjectsModal.tsx on line 69 and line 116
- Increased the width and height in tupaia-web/src/views/ProjectsModal.tsx on line 60 and 61
- Increased padding in tupaia-web/src/components/Modal.tsx on line 24
- Added webkit scroll styling in tupaia-web/src/components/Modal.tsx on line 25 and 26
- Changed right to right: 0rem on tupaia-web/src/components/Modal.tsx on line 40
- Added padding left and padding right in tupaia-web/src/components/Modal.tsx on line 45 and 46
 
### Screenshots:
image


packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -22,6 +22,9 @@ const Wrapper = styled.div`
   text-align: center;
   overflow-x: hidden;
   padding: 2rem;
+  ::-webkit-scrollbar {
```
tcaiger, line 25: I think that we should keep the scroll bar visible even though it looks a bit ugly as it helps users see that the modal is scrolable.

packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -34,12 +37,13 @@ const CloseButton = styled(IconButton)`
   min-width: initial;
   position: absolute;
   top: 0.1rem;
-  right: 0.1rem;
+  right: 0rem;
```
tcaiger, line 40: Any reason to move the close button? I think that i's closer to the design with the 0.1rem to the left isn't it?

packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -34,12 +37,13 @@ const CloseButton = styled(IconButton)`
...
 `;
 
 const Paper = styled(MuiPaper)`
-  background-color: ${({ theme }) => theme.palette.background.default};
-  padding: 0;
+  background-color: #202124;
+  padding-left: 3.125rem;
+  padding-right: 3.125rem;
```
tcaiger, line 46: If we make these changes on here on the Modal component then they will apply everywhere that uses the modal component which includes things like the LoginModal, RegisterModal etc. I think that is correct for the background color but not for the padding. But the good news is you can just make the padding change in the ProjectsModal component instead.

packages/tupaia-web/src/layout/ProjectCardList/LegacyProjectCard.tsx
```diff
@@ -14,8 +14,8 @@ const Card = styled.div`
   display: grid;
```
tcaiger, line 14: I'm pretty confident that you can just delete this whole file now since it's not being used anywhere. (Also need to delete the export of it which is in ProjectCard/index.ts)

packages/tupaia-web/src/views/ProjectsModal.tsx
```diff
@@ -54,19 +56,33 @@ const ExploreButton = styled(RouterButton).attrs({
...
 })`
   margin-bottom: 16px;
-  width: 250px;
-  height: 50px;
+  margin-left: 0.4rem;
+  width: 200px;
+  height: 40px;
   border-radius: 3px;
-  font-size: 13px;
+  font-size: 14px;
```
tcaiger, line 63: We should probably use Rem here too

packages/web-frontend/src/containers/OverlayDiv/components/ProjectPage/LegacyProjectCard.js
```diff
@@ -117,7 +117,7 @@ export const ProjectPendingButton = ({ onClick }) => (
 );
 // eslint-disable-next-line react/prop-types
 export const ProjectAllowedButton = ({ onClick }) => (
-  <Button onClick={onClick} variant="contained" color="primary">
+  <Button onClick={onClick} variant="contained" color="secondary">
```
tcaiger, line 120: Is this change intentional? We probably don't need it right?

packages/tupaia-web/src/views/ProjectsModal.tsx
tcaiger: Also I think there are a few tiny design tweaks to make:
- the font-weight of the Projects text
- the styles of the intro text
- the Explore Tupaia.org button outline color and font-weight

packages/tupaia-web/src/views/ProjectsModal.tsx
```diff
@@ -29,19 +28,22 @@ const Wrapper = styled.div`
...
-  width: 54rem;
+  width: 65rem;
   max-width: 100%;
+  text-align: left;
 `;
 
 const TagLine = styled.p`
-  margin: 0.625rem 0 2.5rem;
+  margin: 0.5rem 0.4rem 1.5rem;
+  width: 40%;
```
tcaiger, line 38: I think it would be better to do `max-width: 26rem` because we don't want the TagLine to be 40% on mobile width screens.

packages/tupaia-web/src/views/ProjectsModal.tsx
```diff
@@ -29,19 +28,22 @@ const Wrapper = styled.div`
...
+  font-weight: 300;
 `;
 
 const ProjectsGrid = styled.div`
   display: grid;
   grid-template-columns: repeat(3, 1fr);
-  gap: 1rem;
-  margin: 1.5rem 0;
+  gap: 2rem;
+  margin: 1.4rem 0;
```
tcaiger, line 46: I think that there needs to be more responsive styles added since the grid doesn't space out properly on tablet size devices. Maybe it should stack to 2 columns at some size?

tcaiger (changes requested):
Nice work! Well done on getting everything updated and working quickly. There are just some small UI changes that I've requested. Thanks

packages/tupaia-web/src/components/Modal.tsx
chaoxinnhin: Yeah for whatever reason, it didn't look the way I wanted it to but I think its because of the removal of scrollbar that made it look janky. But I've changed it back to the original position.

packages/tupaia-web/src/views/ProjectsModal.tsx
```diff
@@ -45,6 +46,10 @@ const ProjectsGrid = styled.div`
   gap: 2rem;
   margin: 1.4rem 0;
 
+  @media (max-width: 1024px) {
```
tcaiger, line 49: Nice

tcaiger (approved):
Nice work!

PR #4804, opened 2023-07-31
WAITP-1286 Fixup overlay sort

EMcQ-BES:
### Changes:

- Sort by `sort_order`, then by `name`
- Update dev dependencies to use `lodash.sortby`
- Add `sortOrder` to types

### DIFF
--- a/packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
+++ b/packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
@@ -13,6 +13,7 @@ import {
 } from '@tupaia/types';
 import groupBy from 'lodash.groupby';
 import keyBy from 'lodash.keyby';
+import sortBy from 'lodash.sortby';
 
 export type MapOverlaysRequest = Request<
   TupaiaWebMapOverlaysRequest.Params,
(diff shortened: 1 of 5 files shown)


packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
```diff
@@ -133,21 +129,29 @@ export class MapOverlaysRoute extends Route<MapOverlaysRequest> {
               code: overlay.code,
               reportCode: overlay.report_code,
               legacy: overlay.legacy,
+              sortOrder: relation.sort_order,
```
EMcQ-BES, line 132: For the sake of the two layered sort, it's easiest if we have access to both `sort_order` and `name` on a single object. But usually `sort_order` is attached to the `relation` entry, while `name` is attached to the `map_overlay[_group]`

So we just temporarily attach `sortOrder` up until we sort the children below, then we discard it so it's not returned to the frontend

packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
```diff
@@ -133,21 +129,29 @@ export class MapOverlaysRoute extends Route<MapOverlaysRequest> {
...
+            sortOrder: relation.sort_order,
+          };
         },
       );
       // Translate Map Overlay Group
       return {
         name: parentEntry.name,
-        children: nestedChildren,
+        children: sortBy(nestedChildren, ['sortOrder', 'name']).map((child: OverlayChild) => {
+          // We only needed the sortOrder for sorting, strip it before we return
```
tcaiger, line 151: thanks for the comment - very helpful

tcaiger (approved):
Nice - thanks

PR #4805, opened 2023-07-31
WAITP-1364: tupaia-web request types

tcaiger:
### Changes:

- Update tupaia web with server response types


packages/tupaia-web/src/types/mapOverlay.d.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export type MapOverlayGroup = {
+  name: MapOverlay['name'];
+  children: SingleMapOverlayItem[] | MapOverlayGroup[];
+};
+
+// Todo: use TupaiaWebMapOverlaysRequest
+// The types from the server and the types defined above need to be merged which will require backend
+// changes in MapOverlayRoute.ts
```
EMcQ-BES, line 27: Do we have a ticket to put this task against? Wanna make sure it doesn't get lost

EMcQ-BES (approved):
The type usage all looks good to me, but I'm not particularly familiar with the surrounding code so I feel like I can't entirely predict issues that might arise 🤔

packages/tupaia-web/src/types/mapOverlay.d.ts
tcaiger: Good point - I've just added one to the refinement queue

PR #4806, opened 2023-07-31
WAITP-1365 Maintain status codes in API connections

EMcQ-BES:
### Changes:

- When an API connection throws an error, don't just discard that error code


tcaiger (approved):
LGTM

PR #4807, opened 2023-07-31
WAITP-1289 Allow requesting qualified name from entity server

EMcQ-BES:
### Changes:

- Allow requesting a qualified name from entity server as an extended field
- Request qualified name by default in `tupaia-web-server` search route


tcaiger (approved):
LGTM

PR #4809, opened 2023-08-01
Set a default y axis formatter

acdunham:
### Changes:
- If `valueType` is `number`, set a default `valueFormat` if not already set. This will be `0.00` when the max value is less than 5, so as not to end up with duplicated tick values from rounding


tcaiger (approved):
Nice work - thanks for picking this up!

PR #4815, opened 2023-08-02
Waitp 1316 map overlay table flippa

WDuncker:
### Changes:

-Built Map table modal to display data from selected Map Overlay in table form

---


packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -59,7 +60,7 @@ export const Modal = ({ children, isOpen, onClose }: ModalProps) => {
       onClose={onClose}
       PaperComponent={Paper}
       fullScreen={fullScreen}
-      disablePortal
+      disablePortal={disablePortal}
```
acdunham, line 63: Is there a particular reason we would like to set this as false, ever? The main reason we keep it as try is so it stays within the relevant components and we can control styles.

packages/tupaia-web/src/features/Map/Map.tsx
```diff
@@ -110,20 +110,21 @@ export const Map = () => {
   const onTileSetChange = (tileSetKey: string) => {
     setActiveTileSet(TILE_SETS.find(({ key }) => key === tileSetKey) as typeof TILE_SETS[0]);
   };
+  const [isOpen, setIsOpen] = useState(false);
```
acdunham, line 113: I think we could probably be more descriptive with our naming here, to indicate what `isOpen` refers to

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -138,56 +148,65 @@ interface DesktopMapOverlaySelectorProps {
...
+    const { data: mapOverlayData } = useMapOverlayReport();
+    const handleOpen = () => {
+      setIsOpen(true);
+    };
+    return (
+      <Wrapper>
+        <Header>
+          <Heading>Map Overlays</Heading>
+          <Tooltip arrow interactive placement="top" title="Generate Report">
+            <TableAssignmentIcon onClick={() => handleOpen()} />
```
acdunham, line 172: Since this is acting like a button, this should probably be wrapped in a `button` element. In this case, then you can set `title` on the `button` to be `Generate report` and not need to create a `Tooltip`

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -138,56 +148,65 @@ interface DesktopMapOverlaySelectorProps {
...
-            onChange={toggleOverlayLibrary}
-            square
-          >
-            <OverlayLibraryHeader
-              expandIcon={<ArrowDropDown />}
-              aria-controls="overlay-library-content"
-              id="overlay-library-header"
+  setIsOpen,
+}: DesktopMapOverlaySelectorProps) =>
+  // { setIsOpen }: any,
```
acdunham, line 159: Just need to remove this commented out line

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -73,11 +74,14 @@ const useEntitiesByMeasureLevel = (measureLevel?: string) => {
   );
 };
 
-export const DataVisualsLayer = ({
-  hiddenValues,
-}: {
-  hiddenValues: LegendProps['hiddenValues'];
-}) => {
+export const DataVisualsLayer = (
+  { isOpen, setIsOpen }: any,
```
acdunham, line 78: I don't think we need `DataVisualsLayer` to know anything about the `MapTableModal`, I think all the logic in it should be contained within `MapTableModal` (see other comments for more on this)

packages/tupaia-web/src/features/Map/MapTableModal/MapTableModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+  processedMeasureData,
+  setIsOpen,
+}: any) => {
+  const handleCloseModal = () => {
+    setIsOpen(false);
+  };
+
+  const { projectCode, entityCode } = useParams();
+  const { data = [] } = useEntityAncestors(projectCode, entityCode);
+  const entityArray = data.reverse();
```
acdunham, line 53: Can you refresh my memory on why we need to reverse this?

packages/tupaia-web/src/features/Map/MapTableModal/MapTableModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const MapTableModal = ({
+  serieses,
+  selectedOverlay,
+  processedMeasureData,
+  setIsOpen,
+}: any) => {
+  const handleCloseModal = () => {
+    setIsOpen(false);
+  };
+
```
acdunham, line 50: I think that fetching the measure data and serieses etc, and processing them should be contained in this file. I know it sounds not right because we already do it in `DataVisualsLayer` but `react-query` is pretty smart, so if you query the same thing in to files, it only does the query once, so there won't be any unnecessary network requests.
So my recommendation would be to keep all the logic that the map table needs within this file, and make `DataVisualsLayer` totally unaware of the `MapTableModal`. Then, instead of rendering `MapTableModal` inside `DataVisualsLayer`, render it inside `DesktopMapOverlaySelector`. Then you won't need to pass around as many props, and can have the open/close logic in `DesktopMapOverlaySelector` as well.

packages/ui-map-components/src/components/Table/getMapTableData.tsx
```diff
@@ -26,9 +26,9 @@ const processColumns = (serieses: Series[]) => {
       Header: 'Name',
       accessor: 'name',
       // eslint-disable-next-line react/prop-types
-      Cell: ({ value }: { value: string | number | boolean | undefined }) => (
-        <FirstColumnCell>{String(value)}</FirstColumnCell>
-      ),
+      Cell: ({ value }: { value: string | number | boolean | undefined }) => {
```
acdunham, line 29: Just need to replace the curly braces here with regular parentheses, otherwise nothing is getting returned in this `Cell` function

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -26,6 +26,7 @@ import {
 import { useMapOverlayReport } from '../utils';
 import { EntityCode } from '../../../types';
 import { processMeasureData } from './processMeasureData';
+import { MapTableModal } from '../MapTableModal/MapTableModal';
```
acdunham, line 29: Ideally imports like this would be like `import { MapTableModal } from './MapTableModal'` because we would export anything from an index file. However I think we don't need `MapTableModal` to be a folder in this case, because there is only one file, so it can just sit as a single file within `Map` folder, or, even better, inside `MapOverlaySelector` since it relates to that.

acdunham (changes requested):
Nice work. I have a few changes requested, mainly around the structuring of the files. Let me know if you have any questions.

packages/tupaia-web/src/features/Map/MapTableModal/MapTableModal.tsx
WDuncker: When `useEntityAncestors` creates it's array, it lists the members starting with deepest level at 0. Meaning that as you go deeper into the entity hierarchy (for example to village or region level, the actual country name gets further and further down the array. But if I reverse it, the country level remains at position 1. I'm creating this array as the map overlay table title should read <map overlay name>, <country>.

packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -53,17 +53,13 @@ export const Modal = ({ children, isOpen, onClose }: ModalProps) => {
...
-      onClose={onClose}
-      PaperComponent={Paper}
-      fullScreen={fullScreen}
-      disablePortal
-    >
-      <CloseButton onClick={onClose} color="default">
-        <CloseIcon />
-      </CloseButton>
-      <Wrapper id="overlay-wrapper">{children}</Wrapper>
+    <Dialog open={isOpen} onClose={onClose} PaperComponent={Paper} fullScreen={fullScreen}>
```
acdunham, line 56: Just need to reinstate `disablePortal` as true always

packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -53,17 +53,13 @@ export const Modal = ({ children, isOpen, onClose }: ModalProps) => {
...
+      <Wrapper id="overlay-wrapper">
+        <CloseButton onClick={onClose} color="default">
```
acdunham, line 58: Looks like the CloseButton has been moved to inside `Wrapper` - any particular reason?

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -146,47 +164,59 @@ export const DesktopMapOverlaySelector = ({
...
-              aria-controls="overlay-library-content"
-              id="overlay-library-header"
+    <>
+      {mapModalOpen ? <MapTableModal setIsOpen={setMapModalOpen} /> : null}
+      <Wrapper>
+        <Header>
+          <Heading>Map Overlays</Heading>
+          <MapTableButton>
+            <Tooltip arrow interactive placement="top" title="Generate Report">
+              <TableAssignmentIcon onClick={() => handleOpen()} />
```
acdunham, line 180: This doesn't need to be an arrow function. When we have arrow functions inline, it technically has a slight performance impact, because it causes re-rendering of the component.

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { useParams } from 'react-router';
+import { useEntityAncestors, useMapOverlays, useEntitiesWithLocation } from '../../../api/queries';
+import { useMapOverlayReport } from '../utils';
+import { processMeasureData } from '../MapOverlays/processMeasureData';
+
+const Wrapper = styled(FlexColumn)`
+  justify-content: flex-start;
+`;
+
+const Title = styled.div.attrs({
```
acdunham, line 21: In this case, because this is an `h2`, we should use the `Typography` component instead of a `div`. Still keep all the rest, but we just want to avoid using divs anywhere that has a better semantic alternative.

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+const TitleWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
+  margin-bottom: 3rem;
+  position: relative;
+  border-bottom: 1px solid rgba(255, 255, 255, 0.12);
+`;
+
+const useEntitiesByMeasureLevel = (measureLevel?: string) => {
```
acdunham, line 43: I'm guessing this is copied over from `DataVisualsLayer`? I think instead of copying it completely, we could move it to the `utils` folder, and import it wherever we use it, ie. in these two files.

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const { projectCode, entityCode } = useParams();
+  const { selectedOverlay } = useMapOverlays(projectCode, entityCode);
+  const { data: entitiesData } = useEntitiesByMeasureLevel(selectedOverlay?.measureLevel);
+  const { data: mapOverlayData } = useMapOverlayReport();
+
+  if (!entitiesData || !mapOverlayData) {
+    return null;
+  }
+
+  const processedMeasureData = processMeasureData({
```
acdunham, line 81: If you wanted to, to make some things reusable between `DataVisualsLayer` and `MapTableModal` you could turn some of it into a hook, like combine `useEntitiesByMeasureLevel` and `processedMeasureData` and handle the fetching etc of all the data. Then it could just return `processedMeasureData` and `serieses`, and anything else we directly need. This could go in  `Map/utils` folder, since it will only ever be used inside the `Map` folder

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+          <TitleWrapper>
+            <Title>{titleText}</Title>
+            <IconButton>
+              <DownloadIcon />
+            </IconButton>
+          </TitleWrapper>
+          <MapTable
+            serieses={serieses}
+            measureData={processedMeasureData as MeasureData[]}
+          ></MapTable>
```
acdunham, line 110: This can be self-closing, since there are no children inside `MapTable`

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+    return null;
+  }
+
+  const { serieses } = mapOverlayData;
+  const { data = [] } = useEntityAncestors(projectCode, entityCode);
+  const entityArray = data.reverse();
+  const titleText = `${selectedOverlay.name}, ${entityArray[1].name}`;
+
+  return (
+    <>
```
acdunham, line 98: You probably don't need this fragment, since there is only one direct child in this, the `Modal`

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -48,45 +48,46 @@ export const DataVisualsLayer = ({
...
 
   return (
-    <LayerGroup>
-      {processedMeasureData.map(measure => {
-        const { region, organisationUnitCode: entity, color, name } = measure;
-        if (region) {
+    <>
+      <LayerGroup>
+        {processedMeasureData.map(measure => {
+          if (measure.region) {
```
acdunham, line 54: Could be nice to go back to the destructuring of before. So this file should be unchanged (except for any possible reusable hooks that get used instead)

packages/ui-map-components/src/components/Table/getMapTableData.tsx
```diff
@@ -26,9 +26,9 @@ const processColumns = (serieses: Series[]) => {
       Header: 'Name',
       accessor: 'name',
       // eslint-disable-next-line react/prop-types
-      Cell: ({ value }: { value: string | number | boolean | undefined }) => (
-        <FirstColumnCell>{String(value)}</FirstColumnCell>
-      ),
+      Cell: (value: { value: string | number | boolean | undefined }) => {
+        <FirstColumnCell>{String(value)}</FirstColumnCell>;
+      },
```
acdunham, line 31: This one needs to be reverted back to regular parentheses, otherwise nothing gets returned here

acdunham (changes requested):
Looks good. Just a few more changes from my end, mainly around some structure we can improve.

packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -12,6 +12,7 @@ interface ModalProps {
   children?: ReactNode;
   onClose: () => void;
   isOpen: boolean;
+  disablePortal: true;
```
acdunham, line 15: So `disablePortal` shouldn't be a prop, it should just always be set on the `Dialog` like so:
image

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -4,18 +4,27 @@
...
 import { Entity } from '../../../types';
 import { useMapOverlays } from '../../../api/queries';
-import { useActiveMapOverlayReport } from '../utils';
+import { useMapOverlayReport } from '../utils';
 import { MapOverlayList } from './MapOverlayList';
 import { MapOverlaySelectorTitle } from './MapOverlaySelectorTitle';
 import { MapOverlayDatePicker } from './MapOverlayDatePicker';
+import MuiIconButton from '@material-ui/core/IconButton';
+import { Tooltip } from '@tupaia/ui-components';
+import { MapTableModal } from './MapTableModal';
```
acdunham, line 22: Just need to reorder these imports

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -4,18 +4,27 @@
  */
 
 import React from 'react';
+import { useState } from 'react';
 import styled from 'styled-components';
 import { useParams } from 'react-router';
 import { Accordion, Typography, AccordionSummary, AccordionDetails } from '@material-ui/core';
-import { ArrowDropDown, Layers } from '@material-ui/icons';
+import { ArrowDropDown, Layers, Assignment } from '@material-ui/icons';
```
acdunham, line 11:
```suggestion
import { ArrowDropDown, Layers, Assignment, IconButton } from '@material-ui/icons';
```

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -4,18 +4,27 @@
...
 import { periodToMoment } from '@tupaia/utils';
 import { MOBILE_BREAKPOINT } from '../../../constants';
 import { Entity } from '../../../types';
 import { useMapOverlays } from '../../../api/queries';
-import { useActiveMapOverlayReport } from '../utils';
+import { useMapOverlayReport } from '../utils';
 import { MapOverlayList } from './MapOverlayList';
 import { MapOverlaySelectorTitle } from './MapOverlaySelectorTitle';
 import { MapOverlayDatePicker } from './MapOverlayDatePicker';
+import MuiIconButton from '@material-ui/core/IconButton';
```
acdunham, line 20:
```suggestion
```

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -4,18 +4,27 @@
...
+import { Tooltip } from '@tupaia/ui-components';
+import { MapTableModal } from './MapTableModal';
+
+const MapTableButton = styled(MuiIconButton)`
```
acdunham, line 24:
```suggestion
const MapTableButton = styled(IconButton)`
```

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -4,18 +4,27 @@
...
+  margin: -10px -10px -10px 0;
+  padding: 15px 10px 15px 18px;
+`;
```
acdunham, line 27: These need to all be in `rems`

packages/ui-map-components/src/components/Table/getMapTableData.tsx
```diff
@@ -26,7 +26,7 @@ const processColumns = (serieses: Series[]) => {
       Header: 'Name',
       accessor: 'name',
       // eslint-disable-next-line react/prop-types
-      Cell: ({ value }: { value: string | number | boolean | undefined }) => (
+      Cell: (value: { value: string | number | boolean | undefined }) => (
```
acdunham, line 29:
```suggestion
      Cell: ({ value } : { value: string | number | boolean | undefined }) => (
```

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -146,47 +164,59 @@ export const DesktopMapOverlaySelector = ({
...
-            expanded={overlayLibraryOpen}
-            onChange={toggleOverlayLibrary}
-            square
-          >
-            <OverlayLibraryHeader
-              expandIcon={<ArrowDropDown />}
-              aria-controls="overlay-library-content"
-              id="overlay-library-header"
+    <>
+      {mapModalOpen ? <MapTableModal setMapModalOpen={setMapModalOpen} /> : null}
```
acdunham, line 174: I think it might be nice to pass in `closeModal={() => setModalOpen(false)}` instead of `setMapModalOpen`

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -146,47 +164,59 @@ export const DesktopMapOverlaySelector = ({
 }: DesktopMapOverlaySelectorProps) => {
   const { projectCode, entityCode } = useParams();
   const { hasMapOverlays } = useMapOverlays(projectCode, entityCode);
-  const { period } = useActiveMapOverlayReport();
-
+  const { data: mapOverlayData } = useMapOverlayReport();
```
acdunham, line 167: Just checking if it's intentional that this was changed from `const { period } = useActiveMapOverlayReport();` to this?

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+import DownloadIcon from '@material-ui/icons/GetApp';
+import MuiIconButton from '@material-ui/core/IconButton';
+import { FlexColumn } from '@tupaia/ui-components';
+import { Typography } from '@material-ui/core';
+import { MapTable } from '@tupaia/ui-map-components';
+import { Modal } from '../../../components';
+import { useEntityAncestors, useMapOverlays } from '../../../api/queries';
+import { useMapOverlayData } from '../utils';
+
+interface entityObject {
```
acdunham, line 17: `interfaces` are usually capitalized, but I think in this case it might not be necessary, because we have an `Entity` type you can import

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  const { projectCode, entityCode } = useParams();
+  const { selectedOverlay } = useMapOverlays(projectCode, entityCode);
+  const { serieses, processedMeasureData } = useMapOverlayData(projectCode, entityCode);
+  const { data = [] } = useEntityAncestors(projectCode, entityCode);
+  const entityObject = data?.find((obj: entityObject) => obj.type === 'country');
+  const titleText = `${selectedOverlay.name}, ${entityObject.name}`;
+  return (
+    <>
```
acdunham, line 62:
```suggestion
```

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <TitleWrapper>
+            <Title>{titleText}</Title>
+            <IconButton>
+              <DownloadIcon />
+            </IconButton>
+          </TitleWrapper>
+          <MapTable serieses={serieses!} measureData={processedMeasureData!} />
+        </Wrapper>
+      </Modal>
+    </>
```
acdunham, line 74:
```suggestion
```

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -24,26 +27,44 @@ const ShadedPolygon = styled(Polygon)`
...
+    projectCode,
+    entityCode,
+    hiddenValues,
+  );
+  const popUpSerieses = mapOverlayData?.serieses;
 
-  if (!measureData || !serieses || !entities) {
+  // Don't show the marker layer if the entity type doesn't match the measure level
+  const firstSeries = mapOverlayData?.serieses?.find((series: any) => series.displayOnLevel);
+  if (firstSeries && camelCase(entity?.type!) !== camelCase(firstSeries.displayOnLevel)) {
```
acdunham, line 63: Can I just check where this came from? Is it a bug you found?

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -24,26 +27,44 @@ const ShadedPolygon = styled(Polygon)`
   }
 `;
 
+const useNavigateToEntity = () => {
```
acdunham, line 30: There is a util we already have called `useEntityLink` you might be able to use instead of this

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -24,26 +27,44 @@ const ShadedPolygon = styled(Polygon)`
...
 
-  const processedMeasureData = processMeasureData({
-    entitiesData: entities!,
-    measureData,
-    serieses,
-    hiddenValues,
-  });
-
-  if (!processedMeasureData) {
+  if (!processedMeasureData || !mapOverlayData?.serieses) {
```
acdunham, line 67:
```suggestion
  if (!processedMeasureData || !serieses) {
```

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -24,26 +27,44 @@ const ShadedPolygon = styled(Polygon)`
...
-  const { serieses, measureData, entities } = useActiveMapOverlayReport();
+  const { projectCode, entityCode } = useParams();
+  const { data: mapOverlayData } = useMapOverlayReport();
+  const { data: entity } = useEntity(projectCode, entityCode);
+  const { serieses, processedMeasureData } = useMapOverlayData(
+    projectCode,
+    entityCode,
+    hiddenValues,
+  );
+  const popUpSerieses = mapOverlayData?.serieses;
```
acdunham, line 59: Should we just be using the `serieses` from `useMapOverlayData`?

packages/tupaia-web/src/features/Map/utils/useMapOverlayData.ts
```diff
@@ -0,0 +1,75 @@
...
+import { useMapOverlayReport } from '../../../api/queries';
+import { EntityCode, ProjectCode } from '../../../types';
+
+export const useMapOverlayData = (
+  projectCode?: ProjectCode,
+  entityCode?: EntityCode,
+  hiddenValues?: LegendProps['hiddenValues'],
+): { serieses: Series[] | undefined; processedMeasureData: MeasureData[] | undefined } => {
+  const useEntitiesByMeasureLevel = (measureLevel?: string) => {
+    const getSnakeCase = (measureLevel?: string) => {
```
acdunham, line 19: Just for the sake of keeping this tidy, I think `getSnakeCase` can move outside of the hook. It can stay in this file but just as a standalone function

packages/tupaia-web/src/features/Map/utils/useMapOverlayData.ts
```diff
@@ -0,0 +1,75 @@
...
+import { useMapOverlays } from '../../../api/queries';
+import { useMapOverlayReport } from '../../../api/queries';
+import { EntityCode, ProjectCode } from '../../../types';
+
+export const useMapOverlayData = (
+  projectCode?: ProjectCode,
+  entityCode?: EntityCode,
+  hiddenValues?: LegendProps['hiddenValues'],
+): { serieses: Series[] | undefined; processedMeasureData: MeasureData[] | undefined } => {
+  const useEntitiesByMeasureLevel = (measureLevel?: string) => {
```
acdunham, line 18: I reckon this should be a separate hook, instead of having hooks within hooks, for the sake of readability

packages/tupaia-web/src/features/Map/utils/useMapOverlayData.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  const { serieses } = mapOverlayData;
+
+  if (!processedMeasureData || !mapOverlayData) {
+    return {
+      serieses: undefined,
+      processedMeasureData: undefined,
+    };
+  }
```
acdunham, line 73: We could probably define a default value of this, instead of rewriting the same thing again. E.g. 
``` 
const defaultValue = {
      serieses: undefined,
      processedMeasureData: undefined,
}
```

acdunham (changes requested):
This is looking heaps better. I still have a few requested changes for you, and some questions

packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -57,12 +59,14 @@ export const Modal = ({ children, isOpen, onClose }: ModalProps) => {
       onClose={onClose}
       PaperComponent={Paper}
       fullScreen={fullScreen}
-      disablePortal
+      // disablePortal
```
acdunham, line 62:
```suggestion
      disablePortal
```

packages/tupaia-web/src/components/Modal.tsx
```diff
@@ -57,12 +59,14 @@ export const Modal = ({ children, isOpen, onClose }: ModalProps) => {
...
     >
-      <CloseButton onClick={onClose} color="default">
-        <CloseIcon />
-      </CloseButton>
-      <Wrapper id="overlay-wrapper">{children}</Wrapper>
+      <Wrapper id="overlay-wrapper">
+        <CloseButton color="default">
+          <CloseIcon onClick={onClose} />
```
acdunham, line 66: The `onClick` here needs to be on the `CloseButton` as opposed to the `CloseIcon`

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -146,47 +164,61 @@ export const DesktopMapOverlaySelector = ({
...
-              id="overlay-library-header"
+    <>
+      {mapModalOpen && <MapTableModal onClose={toggleMapTableModal} />}
+      <Wrapper>
+        <Header>
+          <Heading>Map Overlays</Heading>
+          {measureData && (
+            <MapTableButton>
+              <Tooltip arrow interactive placement="top" title="Generate Report">
+                <TableAssignmentIcon onClick={toggleMapTableModal} />
```
acdunham, line 182: the `onClick` should be on the `MapTableButton` instead of the `TableAssignmentIcon`

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+export const MapTableModal = ({ onClose }: any) => {
+  const { projectCode, entityCode } = useParams();
+  const { selectedOverlay } = useMapOverlays(projectCode, entityCode);
+  const { serieses, measureData } = useMapOverlayData();
+  const { data } = useEntityAncestors(projectCode, entityCode);
+  const countryObject = data?.find(entity => entity.type === 'country');
+  const titleText = `${selectedOverlay.name}, ${countryObject?.name}`;
+
+  // const handleCloseModal = () => {
+  //   setMapModalOpen(false);
```
acdunham, line 52: Just need to remove these commented out parts

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -30,26 +32,23 @@ export const DataVisualsLayer = ({
...
-
-  if (!processedMeasureData) {
+  if (!measureData || !serieses) {
     return null;
   }
 
   return (
     <LayerGroup>
-      {processedMeasureData.map(measure => {
+      {measureData.map((measure: any) => {
```
acdunham, line 51: Can we update this not use `any`?

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
acdunham: Bumping this question

packages/tupaia-web/src/features/Map/utils/useMapOverlayData.ts
```diff
@@ -49,7 +51,7 @@ const getRootEntity = (entity?: Entity) => {
   return parentCode;
 };
 
-export const useActiveMapOverlayReport = () => {
+export const useMapOverlayData = (hiddenValues?: LegendProps['hiddenValues']) => {
```
acdunham, line 54: I see that `hiddenValues` is a param for this. But am I right in saying nowhere passes in `hiddenValues`? If so, should somewhere be doing that?

packages/tupaia-web/src/features/Map/utils/useMapOverlayData.ts
```diff
@@ -70,11 +72,22 @@ export const useActiveMapOverlayReport = () => {
...
+    hiddenValues: hiddenValues ? hiddenValues : {},
+  }) as MeasureData[];
+
   return {
     ...data,
     serieses: data?.serieses,
-    measureData: data?.measureData,
-    entities,
+    measureData: processedMeasureData,
+    entitiesData,
```
acdunham, line 90: Can probably rename this back to `entities` to be more accurate

acdunham (changes requested):
Sorry about this, I have a few more questions /suggestions in the review

packages/tupaia-web/src/features/Map/MapOverlaySelector/DesktopMapOverlaySelector.tsx
```diff
@@ -3,19 +3,28 @@
...
-import { MapOverlaySelectorTitle } from './MapOverlaySelectorTitle';
 import { MapOverlayDatePicker } from './MapOverlayDatePicker';
+import { MapOverlaySelectorTitle } from './MapOverlaySelectorTitle';
+import { useMapOverlayData } from '../utils';
+import { Entity } from '../../../types';
+import { useMapOverlays } from '../../../api/queries';
+import { MOBILE_BREAKPOINT } from '../../../constants';
+
+const MapTableButton = styled(IconButton)`
+  margin: -0.625rem -0.625rem -0.625 0;
```
acdunham, line 24:
```suggestion
  margin: -0.625rem -0.625rem -0.625rem 0;
```

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
acdunham: @WDuncker just bumping this one 🙏

packages/tupaia-web/src/features/Map/utils/useMapOverlayData.ts
acdunham: @WDuncker just bumping this one 🙏

acdunham (commented):
This looks really good, thank you for your patience! I have added a couple of questions, if you could possibly look into them @WDuncker ?

acdunham (approved):
🎉 nice one!

PR #4820, opened 2023-08-03
Waitp-1366 permissions error modal

acdunham:
### Changes:
- Added a `SpinningLoader` utility component to `ui-components` and updated anywhere that used `CircularProgress` to use this instead.
- Set the request access modal to show when user goes to a project when they don't have access and are logged in
- Set the login modal to show when a permissions error is caught for project or entity and the user is not logged in
- Set the request country access modal to show when a user is denied access to an entity and they are logged in
- Show the loading screen when loading the user details, so the user gets loaded before projects etc.
- Show a loader on request country access modal when the project is still loading, to make the user experience smoother if the modal is being shown because of permissions errors.


packages/ui-components/src/components/SpinningLoader.tsx
```diff
@@ -0,0 +1,20 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
```
tcaiger, line 4: I think that this is a different component to the other loading spinners in ui-components but there are a few now and it would be easy for duplicates to creep in so would you mind putting them all in a folder.
- FullPageLoader
- LoadingContainer
- SpinningLoader

packages/tupaia-web/src/utils/useHandleNoAccessError.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { useLocation, useParams } from 'react-router';
+import { useModal } from '.';
+import { useUser } from '../api/queries';
+import { MODAL_ROUTES, URL_SEARCH_PARAMS } from '../constants';
+
+/**
+ * Returns a function that redirects the user to the login page or the request access modal(s). Needs to be a hook format because it uses hooks inside it.
+ */
+export const useHandleNoAccessError = (isProject: boolean) => {
```
tcaiger, line 14: I think I might have said that we could re-use the logic between project and entity permission error handling but I now that I see it I'm finding it quite complex and hard to follow. I think that it would be better to just split this into 2 utils, one for handling project permission error and another for handling entity permission errors.

packages/tupaia-web/src/api/queries/useEntity.ts
```diff
@@ -25,6 +27,13 @@ export const useEntity = (projectCode?: ProjectCode, entityCode?: EntityCode) =>
 
       return entity;
     },
-    { enabled: !!entityCode && !!projectCode },
+    {
+      enabled: !!entityCode && !!projectCode,
+      onError: (e: any) => {
+        if (e.code === 403) {
```
tcaiger, line 33: I think this turned out pretty nice. Very clear and clean.

tcaiger (changes requested):
Looks good overall, it would just be good to try to make the util for handling the errors a bit simpler to understand somehow. Splitting in 2 is probably the best but I'm open to other ideas.

packages/tupaia-web/src/utils/useHandleNoAccessError.ts
acdunham: I decided to go with the approach of handling the errors in their respective places, since that particular error handling won't get used outside of those files.

packages/tupaia-web/src/api/queries/useEntity.ts
```diff
@@ -25,6 +31,14 @@ export const useEntity = (projectCode?: ProjectCode, entityCode?: EntityCode) =>
...
       return entity;
     },
-    { enabled: !!entityCode && !!projectCode },
+    {
+      enabled: !!entityCode && !!projectCode,
+      onError: (e: any) => {
+        if(e.code !== 403) return;
+        if (!isLoggedIn) return navigateToLogin();
+        if (location.hash === `#${MODAL_ROUTES.REQUEST_PROJECT_ACCESS}`) return;
+        return navigateToModal(MODAL_ROUTES.REQUEST_COUNTRY_ACCESS);
```
tcaiger, line 40: muuuch clearer - thanks

tcaiger (approved):
Perfecto!

PR #4821, opened 2023-08-03
WAITP-1317 Apply frontendExcluded filter in entity routes

EMcQ-BES:
### Changes:

- Add more specific array handling to `entity-server` filter generation
  - Add unit tests
- Generate a filter from `project.config.frontendExcluded`
  - Apply this filter to multi entity routes in `tupaia-web-server`

I've left in the option to override this filter, and I haven't bothered adding anything to the single entity route. I figure these are entities that the user _does_ have permissions to, we just don't want them returned by default. If we want to override the default then that's an option in the route.


packages/entity-server/src/routes/hierarchy/middleware/filter.ts
```diff
@@ -73,6 +73,15 @@ const operatorToSqlComparator = {
   '>': '>' as const, // Greater than
   '>=': '>=' as const, // Greater than or equal
 };
+const operatorToSqlArrayComparator = {
+  '==': 'IN' as const, // Contained in array
```
EMcQ-BES, line 77: Technically this isn't used since the `convertValueToAdvancedCriteria` exits early for `==`, and the default behaviour is to use `IN` if we use an array value without a specific sql comparator.

I've left it here anyway so if we change the logic elsewhere it's specifically defined

packages/tupaia-web-server/src/routes/EntityAncestorsRoute.ts
```diff
@@ -23,10 +24,18 @@ export class EntityAncestorsRoute extends Route<EntityAncestorsRequest> {
...
+        columns: ['config'],
+      })
+    )[0];
+    const { config } = project;
+
     const entities = await ctx.services.entity.getAncestorsOfEntity(
       projectCode,
       rootEntityCode,
-      { fields: DEFAULT_FIELDS, ...restOfQuery },
+      { filter: generateFrontendExcludedFilter(config), fields: DEFAULT_FIELDS, ...restOfQuery },
```
EMcQ-BES, line 38: I'm not entirely sure if this is useful in the ancestors route, I _think_ we're usually excluding types without any children. But I've added it here for consistency

rohan-bes: I'd be interested to know if things works when we exclude here. I'd expect them to not work, but that's fine. It'd basically break the hierarchy. Keep it here though, that way the faulty config will fail noisily rather than silently 👍

rohan-bes (approved):
Great stuff, love the addition of handling array type filters!

PR #4825, opened 2023-08-03
Waitp 1222 point overlay fix

acdunham:
### Changes:
- Only keep previous data for map overlay when query is enabled. Fixes issue where going back to an entity where there were no map overlays would keep the previous entity's data


tcaiger (approved):
Great fix! thanks

PR #4827, opened 2023-08-04
WAITP-1365: Unit test for web server error forwarding

EMcQ-BES:
### Changes:

- Add unit test setup and data to `tupaia-web-server`
- Write an actual unit test
- Allow `MockEntityApiInterface` to process permission errors


packages/tupaia-web-server/src/__tests__/errors.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+describe('Error responses', () => {
+  let app: TestableServer;
+
+  beforeAll(async () => {
+    app = await setupTestApp();
+  });
+
+  describe('Microservice errors', () => {
+    it('Requesting an entity with no permissions', async () => {
```
IgorNadj, line 17: I like the idea behind this, but with these orchestration servers we don't really need to test routes that are forwarded, otherwise we are doubling up on testing, once in the orchestration server and once in the actual server.

If we want to add coverage here we should test that the api client mock has had a specific method called on the respective client, here it would be apiClientMock.entity.getEntity. Up to you if we want to do this, I'm happy enough not covering forwarded routes, we still have a lot of core business logic that doesn't have tests yet that might warrant the attention more.

packages/api-client/src/connections/mocks/MockEntityApi.ts
```diff
@@ -22,6 +25,13 @@ export class MockEntityApi implements EntityApiInterface {
     const { field, fields, filter } = queryOptions;
 
     let filteredEntities = foundEntities;
+    if (this.getAccessPolicy) {
```
IgorNadj, line 28: With the comment below about testing forwarded routes, we probably don't need the complexity here? Happy either way, just thinking no need to overcomplicate things if not necessary. Tests that need a permissions failure response could just mock it out specifically?

packages/tupaia-web-server/src/__tests__/testUtilities/setup.ts
```diff
@@ -0,0 +1,77 @@
...
+} from '@tupaia/database';
+
+import { TestModelRegistry } from './testModelRegistry';
+import { PROJECTS, ENTITIES, ENTITY_RELATIONS, getAccessPolicy } from './fixtures';
+import { createApp } from '../../app';
+
+// Don't generate the proxy middlewares while we're testing
+jest.mock('http-proxy-middleware');
+
+jest.mock('@tupaia/api-client', () => {
```
IgorNadj, line 25: Here

IgorNadj (changes requested):
Looks great thanks! One bit of feedback re testing.

packages/tupaia-web-server/src/__tests__/errors.test.ts
EMcQ-BES: Yes and no... Perhaps I need to rephrase this test case

Prior to recent changes, the orchestration server would consume any error from a microservice and return `500` regardless of what the original error was. This unit test exists to confirm that the server is now forwarding errors correctly. It's testing the behaviour of the orchestration server, not the microservice

packages/api-client/src/connections/mocks/MockEntityApi.ts
EMcQ-BES: > Tests that need a permissions failure response could just mock that response from api client specifically

Oh this is a great point I should definitely change it to just do that 😅 Got a bit too in my head about reusing the mock api client

packages/tupaia-web-server/src/__tests__/errors.test.ts
IgorNadj: I see. Yeah good thinking and thanks for adding that test!

I think keeping the MockTupaiaApiClient simple and happy path is a good idea. 

I think a simple jest mock like this would be a good way to test non-happy path stuff: https://github.com/beyondessential/tupaia/commit/3037becf5555106e3553b5fd028cd14e610c704d

From the docs there does seem to be a way of defining a global module mock and then overriding the implementation on a per-test basis, but I couldn't figure it out. The other way we've done it in the past to avoid global mocks is with a test helper like setupApiClientMock() which I left unimplemented in the example.

EMcQ-BES: Cheers @IgorNadj , I've implemented some of those suggestions (although I find splitting out the `handleServerError` function a bit unnecessary)

Also reverted the changes to the mock entity api

IgorNadj (approved):
Looks great thanks!

PR #4830, opened 2023-08-04
Waitp 1238 overlay export to excel

chaoxinnhin:
### Issue #:
WAITP-1238: Exports: Map overlay Export to Excel

### Changes:

- Added startDate and EndDate in tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx on line 96 and 97
- Added a { doExport } function on line 99 
- Added onClick handler for {doExport} on line 114

### Screenshots:
image


tcaiger (approved):
Good work figuring this out!

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapTableModal.tsx
```diff
@@ -101,7 +111,7 @@ export const MapTableModal = ({ setIsOpen }: any) => {
           <TitleWrapper>
             <Title>{titleText}</Title>
             <IconButton>
-              <DownloadIcon />
+              <DownloadIcon onClick={doExport} />
```
acdunham, line 114: Can I just chime in here with one request - can you possibly put the onClick handler onto the `IconButton` instead of the `DownloadIcon` so we are more semantic?

acdunham (changes requested):
Just one small change from me, but otherwise looks good.

acdunham (approved):
Thanks!

PR #4835, opened 2023-08-06
WAITP-1256: Fixup dashboard item sort

EMcQ-BES:
### Changes:

- Actually sort the dashboard items, not just the dashboards
- (Snuck in a separate small fix for exclusion filter)


tcaiger (approved):
LGTM

PR #4836, opened 2023-08-07
WAITP-1283 admin panel to vite

acdunham:
### Changes:
- Update all jsx files to have the correct file extension in `admin-panel`
- Move vite config and associated scripts to root folder for reuse
- Update admin panel scripts to use the common vite scripts
- Update module mock file. This is because it doesn't get transformed correctly in `module.exports` format, so when running `vite preview` there was an error `module is not defined`
- Updated `JSONEditor` to not use `webpack-resolver` as this was breaking the vite build


tcaiger (approved):
Nice work. I love that you have already setup the package.json and vite config at the root level.

PR #4838, opened 2023-08-08
WAITP-1392 PSSS to vite

acdunham:
### Changes:
- Converted all react component files to `.jsx` extension
- Changed over usage of `process.env` to `import.meta.env`
- Update build scripts to use new `vite` scripts
- Updated other config to make work with `vite`


tcaiger (approved):
Nicely done

PR #4840, opened 2023-08-09
WAITP-1215: Misc backend fixups

EMcQ-BES:
### Changes:

- Allow returning no data dashboard
- Move dashboard sort to handle `null` values the same as prior to the rewrite
- Return minimal version of the user data


tcaiger (approved):
Nice - thanks

PR #4841, opened 2023-08-09
WAITP-1215 projects navigation fixes

acdunham:
### Changes:
- Fix issue with dashboards not correctly reloading between entities and projects
- Fix issue with datepicker not correctly calculating the default start/end dates
- Fix issue with map overlays not always showing up at the country level
- redirect user to default dashboard if current dashboard name param is invalid
- Handle encoding of dashboard names


packages/tupaia-web/src/api/queries/useDashboards.ts
```diff
@@ -13,11 +13,12 @@ export const useDashboards = (
...
   dashboardName?: DashboardName,
 ) => {
-  const { data = [], isLoading } = useQuery(
+  const enabled = !!entityCode && !!projectCode;
+  const { data = [], isLoading, isError } = useQuery(
     ['dashboards', projectCode, entityCode],
     (): Promise<TupaiaWebDashboardsRequest.ResBody> =>
       get(`dashboards/${projectCode}/${entityCode}`),
-    { enabled: !!entityCode && !!projectCode },
+    { enabled, keepPreviousData: enabled },
```
acdunham, line 21: This is so that if the query is now disabled, the old data does not remain visible

packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -103,17 +104,34 @@ const DashboardImageContainer = styled.div`
...
+    if (
+      !dashboards ||
+      isLoadingProject ||
+      project?.code !== projectCode ||
+      dashboards?.find(dashboard => dashboard.name === dashboardName)
+    )
+      return;
+
+    navigate(defaultEntityLink);
+  }, [JSON.stringify(dashboards), dashboardName, projectCode, isLoadingProject]);
```
acdunham, line 133: I was on the fence about whether to put this here or as a redirect in the `Routes` file, but went with here to match the pattern with default map overlays in the Map file

packages/tupaia-web/src/features/Map/MapOverlays/PolygonNavigationLayer.tsx
```diff
@@ -63,20 +62,23 @@ export const PolygonNavigationLayer = () => {
...
 
   if (!entities || entities.length === 0) {
     return null;
   }
 
   const childEntities = entities.filter((entity: Entity) => entity.parentCode === entityCode);
-
   const showChildEntities =
-    !POLYGON_MEASURE_TYPES.includes(selectedOverlay?.displayType) && childEntities?.length > 0;
+    POLYGON_MEASURE_TYPES.includes(selectedOverlay?.displayType) && childEntities?.length > 0;
```
acdunham, line 74: I changed this as it seemed like the reverse of what the current web frontend is doing

packages/tupaia-web/src/features/Map/MapOverlays/PolygonNavigationLayer.tsx
```diff
@@ -63,20 +62,23 @@ export const PolygonNavigationLayer = () => {
...
+
+  const showActiveEntity =
+    activeEntity && activeEntity?.type !== selectedOverlay?.measureLevel?.toLowerCase();
```
acdunham, line 77: Added in logic to match web frontend

packages/tupaia-web/src/features/Map/MapOverlays/processMeasureData.ts
```diff
@@ -17,15 +17,16 @@ interface processMeasureDataProps {
...
   serieses: Series[];
   hiddenValues: LegendProps['hiddenValues'];
+  includeEntitiesWithoutCoordinates?: boolean; // this is for differentiating between the processing of data for the table and for the map
 }
 export const processMeasureData = ({
   measureData,
   entitiesData,
   serieses,
   hiddenValues,
+  includeEntitiesWithoutCoordinates,
```
acdunham, line 27: This is because the map overlay table doesn't want to drill down the same as a map overlay does

packages/tupaia-web/src/features/Map/utils/useMapOverlayData.ts
```diff
@@ -21,24 +17,25 @@ const useEntitiesByType = (
...
   return useEntitiesWithLocation(
     projectCode,
     entityCode,
     {
       params: {
         // Don't include the root entity in the list of entities for displaying data as the
-        // data visuals are for children of the root entity
-        includeRootEntity: false,
+        // data visuals are for children of the root entity, unless it is a root entity, ie. a country
+        includeRootEntity: snakeCaseEntityType === 'country',
```
acdunham, line 28: Setting this to `false` for country level data was resulting in 0 results, which meant there was no map overlay data appearing on the screen

packages/tupaia-web/src/features/Visuals/NoDataAtLevelDashboard.tsx
```diff
@@ -23,19 +24,18 @@ const Text = styled(Typography)`
...
-  const { type = '' } = entity;
+  const { type = 'country' } = entity || ({} as Entity);
   const displayType = type?.toLowerCase();
   const isProject = displayType === 'project';
   return (
     <>
       <Text>
-        You do not currently have access to view project data
-        {isProject ? 'at the project level view' : ` for the selected ${displayType}`}
+        No project data is currently available for
```
acdunham, line 33: Realised this component was wrong, so just fixing text content

packages/tupaia-web/src/api/queries/useDashboards.ts
tcaiger: thanks - I was perhaps too aggressive when I added keepPreviousData as true globally since it seems to be creating bugs. Let's keep an eye on it and if we have to keep adding exceptions we can turn off that default.

packages/tupaia-web/src/features/DashboardItem/DashboardItemContent.tsx
```diff
@@ -120,7 +120,7 @@ export const DashboardItemContent = ({
 
   if (!DisplayComponent) return null;
 
-  if (isLoading || !report)
+  if (isLoading || (!report && config?.type !== 'component' && !error))
```
tcaiger, line 123: Would you mind adding a comment here?

packages/tupaia-web/src/views/ProjectsModal.tsx
```diff
@@ -132,19 +130,25 @@ export const ProjectsModal = () => {
...
               actions={{
                 [PROJECT_ACCESS_TYPES.ALLOWED]: ({
                   project: { code, homeEntityCode, dashboardGroupName },
+                }: {
+                  project: SingleProject;
                 }) => (
                   <ProjectAllowedLink
                     url={`/${code}/${homeEntityCode}${
-                      dashboardGroupName ? `/${dashboardGroupName}` : ''
+                      dashboardGroupName ? `/${encodeURIComponent(dashboardGroupName)}` : ''
```
tcaiger, line 142: good catch. Is this going to be an issue in other places too? Such as MultiProjectLandingPage and SingleProjectLandingPage and useNavigateToEntity?

packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -103,17 +104,33 @@ const DashboardImageContainer = styled.div`
...
   const { data: entity } = useEntity(projectCode, entityCode);
   const bounds = entity?.bounds || DEFAULT_BOUNDS;
+  const defaultEntityLink = useEntityLink(entityCode);
 
   const toggleExpanded = () => {
     setIsExpanded(!isExpanded);
   };
 
+  // check for valid dashboard name, and if not valid and not still loading, redirect to default dashboard
+  const dashboardNotFound =
```
tcaiger, line 126: I think this is the best place for it but I was getting an error about max renders exceeded which was caused by one of the useEffect dependencies so I refactored it slightly. As far as I can tell it has the same outcome

packages/tupaia-web/src/features/Map/MapOverlays/PolygonNavigationLayer.tsx
tcaiger: hmmm I'm not sure about this. It's quite hard to follow in the OG codebase but what this is meant to do in this version is only show child polygons if the selected map overlay is not a polygon type because then there would be polygons displayed on top of polygons. Also if there is no selected overlay this should show the children but after the change it doesn't. Is this fixing a bug that was found in testing?

tcaiger:
@acdunham I have made a couple of changes and am going to merge this as it will be easier to keep track of the fixes that way. There is an outstanding question about the Map polygon children which we can discuss separately.

PR #4843, opened 2023-08-10
Waitp 1371 Misc: Copy over google analytics logic to tupaia-web

WDuncker:
### Changes:

- Added gaEvent functions to tupaia-web to match what was being recorded in web-frontend.

---


packages/tupaia-web/src/ModalRoutes.tsx
```diff
@@ -42,10 +42,15 @@ export const ModalRoutes = () => {
   const { hash, closeModal } = useModal();
 
   const modal = hash as typeof MODAL_ROUTES[keyof typeof MODAL_ROUTES];
+  useEffect(() => {
+    if (modal === null) {
+      gaEvent('User', 'Close Modal');
+    }
+    gaEvent('User', modal, 'Open Modal');
+  }, [modal]);
```
acdunham, line 50: I wonder if this will have the effect of running on mount as well? If so, we probably don't want that

packages/tupaia-web/src/api/mutations/useLogin.ts
```diff
@@ -21,6 +21,7 @@ export const useLogin = () => {
 
   const { closeModal, navigateToModal } = useModal();
 
+  gaEvent('User', 'Log in', 'Attempt');
```
acdunham, line 24: I think we should put this inside an `onMutate` event, similar to the `onSuccess`, because otherwise this will get called more often than a login attempt (as the mutation call isn't called just whenever the hook is used, like `useQuery` is)

packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -109,9 +110,15 @@ export const Dashboard = () => {
...
   const { data: entity } = useEntity(projectCode, entityCode);
   const bounds = entity?.bounds || DEFAULT_BOUNDS;
+  useEffect(() => {
+    gaEvent('Dashboard', 'Change Tab', activeDashboard?.name);
+  }, [activeDashboard]);
 
   const toggleExpanded = () => {
     setIsExpanded(!isExpanded);
+    if (isExpanded === true) {
+      gaEvent('Pages', 'Toggle Info Panel');
```
acdunham, line 120: I'm guessing we want this to be called regardless of the value of `isExpanded`? In the old one I think it's just every time it's expanded

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapOverlaySelector.tsx
```diff
@@ -6,12 +6,16 @@
...
 import { MobileMapOverlaySelector } from './MobileMapOverlaySelector';
+import { gaEvent } from '../../../utils';
 
 export const MapOverlaySelector = () => {
   const [overlayLibraryOpen, setOverlayLibraryOpen] = useState(false);
 
   const toggleOverlayLibrary = () => {
     setOverlayLibraryOpen(!overlayLibraryOpen);
+    if (overlayLibraryOpen === false) {
+      gaEvent('Dropdown', 'Close');
```
acdunham, line 17: This should be called with `Overlay` `Close` if it's meant to be the same as the old one

packages/tupaia-web/src/layout/UserMenu/UserMenu.tsx
```diff
@@ -43,6 +45,7 @@ export const UserMenu = () => {
 
   const onCloseMenu = () => {
     setMenuOpen(false);
+    gaEvent('Dropdown', 'Close');
```
acdunham, line 48: Is this a new one? I don't think it gets called in the old one?

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -30,26 +34,29 @@ export const DataVisualsLayer = ({
   hiddenValues: LegendProps['hiddenValues'];
 }) => {
   const navigateToEntity = useNavigateToEntity();
-  const { serieses, measureData, entities } = useActiveMapOverlayReport();
+  const { projectCode, entityCode } = useParams();
+  const { data: entity } = useEntity(projectCode, entityCode);
+  const { serieses, measureData, activeEntity, selectedOverlay } = useMapOverlayData(hiddenValues);
+  useEffect(() => {
+    gaEvent('Map Overlays', 'Change', selectedOverlay?.name);
```
acdunham, line 41: Potentially just add a check here to make sure `selectedOverlay` is set, and make the value you're watching `selectedOverlay?.name`

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -30,26 +34,29 @@ export const DataVisualsLayer = ({
...
+  }, [selectedOverlay]);
+  useEffect(() => {
+    gaEvent('Entity', 'Change', activeEntity?.name);
```
acdunham, line 44: Also maybe add a check here that `activeEntity` is set. Possibly also make the property you're watching `activeEntity?.name`

acdunham (changes requested):
Looks good so far, I just have a few changes to request. Also may need to look into the codeship build error as well.

Looks like some of the commit is unrelated as well, so may need to check through the PR and check the files being committed are correct

packages/tupaia-web/src/ModalRoutes.tsx
```diff
@@ -42,10 +42,15 @@ export const ModalRoutes = () => {
   const { hash, closeModal } = useModal();
 
   const modal = hash as typeof MODAL_ROUTES[keyof typeof MODAL_ROUTES];
+  useEffect(() => {
+    if (modal === null) {
+      gaEvent('User', 'Close Modal');
+    }
+    gaEvent('User', modal, 'Open Modal');
```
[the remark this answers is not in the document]
tcaiger: Yes they should be exactly the same

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapOverlaySelector.tsx
WDuncker: I've removed this one altogether. I got a bit confused with some of the naming conventions in the old one and thought dropdown overlays were called every time a dropdown menu was opened. Tom explained it to me and we don't think the dropdown menus need them at all.

packages/tupaia-web/src/features/EntitySearch/EntitySearch.tsx
```diff
@@ -58,16 +58,22 @@ export const EntitySearch = () => {
...
-    gaEvent('Search', 'Change Search');
-  }, [setSearchValue]);
-  useEffect(() => {
-    if (isOpen === true) {
-      gaEvent('Search', 'Open Search Bar');
+
+  const toggleExpandedSearchBar = () => {
+    setIsOpen(!isOpen);
+    gaEvent('Search', 'Toggle Expand');
+    if (isOpen === false) {
```
acdunham, line 68: Just wondering if we should have like `const updatedOpenValue =. !isOpen` and then refer to that when checking whether to set search value as `''`. Just because there is sometimes a split second before it's updated

acdunham (changes requested):
Just one tiny change in there but other than that it is good to go

packages/tupaia-web/src/features/EntitySearch/EntitySearch.tsx
```diff
@@ -63,7 +63,8 @@ export const EntitySearch = () => {
   };
 
   const toggleExpandedSearchBar = () => {
-    setIsOpen(!isOpen);
+    const updatedOpenValue = !isOpen;
+    setIsOpen(updatedOpenValue);
     gaEvent('Search', 'Toggle Expand');
     if (isOpen === false) {
```
acdunham, line 69: Just need to use `updatedOpenValue === false` in the check instead of `isOpen === false`

acdunham (changes requested):
Final change :)

PR #4846, opened 2023-08-10
WAITP-1286 Use direct db access for MapOverlayRoute

EMcQ-BES:
### Changes:

- Remove some debug logs
- Allow `TupaiaDatabase.findRecursiveTree` to search for multiple ids at once
- Add a recursive parent search to `MapOverlayGroupRelationModel`
- Allow orchestration servers to attach models to req in `server-boilerplate`
  - This isn't best practice, so generates a warning on startup
- Add model and modelRegistry types to `tupaia-web-server`
  - Attach those to the default request type
- Update `MapOverlaysRoute` to fetch relations from the db directly

### DIFF
--- a/packages/database/src/TupaiaDatabase.js
+++ b/packages/database/src/TupaiaDatabase.js
@@ -276,9 +276,9 @@ export class TupaiaDatabase {
     const sql = `
      with recursive findParents as (
        select * from ${recordType}
-         where ${idKey} = '${id}'
+         where ${idKey} in ('${Array.isArray(id) ? id.join("','") : id}')
        union
-         select ${recordType}.* from ${recordType}
+         select distinct(${recordType}.*) from ${recordType}
            join findParents on findParents.${parentIdKey} = ${recordType}.${idKey}
      )
 
--- a/packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
+++ b/packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
@@ -16,6 +16,7 @@ import {
 import { ModelRegistry, TupaiaDatabase } from '@tupaia/database';
 import { AccessPolicy } from '@tupaia/access-policy';
 import { UnauthenticatedError } from '@tupaia/utils';
+import winston from 'winston';
 
 import { handleWith, handleError, emptyMiddleware } from '../../utils';
 import { TestRoute } from '../../routes';
@@ -47,7 +48,11 @@ export class ApiBuilder {
   // We add handlers at the end so that middlewares and initial routes can be set up first
   private handlers: { add: () => void }[] = [];
 
-  public constructor(transactingConnection: TupaiaDatabase, apiName: string) {
+  public constructor(
+    transactingConnection: TupaiaDatabase,
+    apiName: string,
+    options: { attachModels: boolean } = { attachModels: false },
+  ) {
     this.database = transactingConnection;
     this.models = new ModelRegistry(this.database) as ServerBoilerplateModelRegistry;
     this.apiName = apiName;
@@ -76,6 +81,12 @@ export class ApiBuilder {
      * Add singletons to be attached to req for every route
      */
     this.app.use((req: Request, res: Response, next: NextFunction) => {
+      if (options.attachModels) {
+        winston.warn(
+          "Best practices say orchestrator servers shouldn't access the db directly, are you sure you need req.models?",
+        );
+        req.models = this.models;
+      }
       const context = {}; // context is shared between request and response
       req.ctx = context;
       res.ctx = context;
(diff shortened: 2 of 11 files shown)


packages/database/src/TupaiaDatabase.js
```diff
@@ -276,9 +276,9 @@ export class TupaiaDatabase {
     const sql = `
      with recursive findParents as (
        select * from ${recordType}
-         where ${idKey} = '${id}'
+         where ${idKey} in ('${Array.isArray(id) ? id.join("','") : id}')
        union
-         select ${recordType}.* from ${recordType}
+         select distinct(${recordType}.*) from ${recordType}
```
EMcQ-BES, line 281: Added a `distinct` here because searching multiple children can result in multiples of the same parent

packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
```diff
@@ -61,46 +60,10 @@ export class MapOverlaysRoute extends Route<MapOverlaysRequest> {
...
-        filter: {
-          child_type: 'mapOverlayGroup',
-          child_id: mapOverlayRelations.map(
-            (relation: MapOverlayGroupRelation) => relation.map_overlay_group_id,
-          ),
-        },
-        pageSize: DEFAULT_PAGE_SIZE,
-      },
+    // Breaking orchestration server convention and accessing the db directly
+    const mapOverlayRelations = await this.req.models.mapOverlayGroupRelation.findParentRelationTree(
```
EMcQ-BES, line 64: I've left the final fetch for the `mapOverlayGroups` themselves as a request to `central-server`. I'd like to get this tested with this first section in place before deciding whether to replace that request with a db call too. Trying to avoid _too_ much breaking of orchestration server practices (at least without explicit reason)

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -76,6 +81,12 @@ export class ApiBuilder {
      * Add singletons to be attached to req for every route
      */
     this.app.use((req: Request, res: Response, next: NextFunction) => {
+      if (options.attachModels) {
+        winston.warn(
+          "Best practices say orchestrator servers shouldn't access the db directly, are you sure you need req.models?",
+        );
+        req.models = this.models;
+      }
```
rohan-bes, line 89: I do really like this, I think it's great to include these kinds of conventions in the code itself.

I worry this one might get a tad noisy, but happy to see how it goes 👍

packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -47,7 +48,11 @@ export class ApiBuilder {
   // We add handlers at the end so that middlewares and initial routes can be set up first
   private handlers: { add: () => void }[] = [];
 
-  public constructor(transactingConnection: TupaiaDatabase, apiName: string) {
+  public constructor(
+    transactingConnection: TupaiaDatabase,
+    apiName: string,
+    options: { attachModels: boolean } = { attachModels: false },
```
rohan-bes, line 54: Instead of passing in an option, thoughts on adding a function on the builder instead?
```
public attachModels() {
  winston.warn("Best practices say orchestrator servers shouldn't access the db directly, are you sure you need req.models?");
  this.app.use((req: Request, res: Response, next: NextFunction) => {
    req.models = this.models;
  });
  return this;
}
``

File: packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts (hunk 2 above)
Comment (rohan-bes) [reply packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts:64]: Good call. Arguably if we were to make a rule about this we might want to say that it's the central-server that enforces permissions, so we can't fetch resources themselves directly from the db and return them to the user in an orchestration server.

File: packages/tupaia-web-server/src/models/MapOverlayGroupRelation.ts [hunk 5]
```diff
@@ -0,0 +1,27 @@
...
+} from '@tupaia/database';
+import { Model } from '@tupaia/server-boilerplate';
+
+type MapOverlayGroupRelationFields = Readonly<{
+  id: string;
+  map_overlay_group_id: string;
+  child_id: string;
+  child_type: 'mapOverlay' | 'mapOverlayGroup';
+  sort_order: number | null;
+}>;
```
rohan-bes, line 18: I think we can get this from the `@tupaia/types` package now. 
```suggestion
import { MapOverlayGroupRelation } from '@tupaia/types';
```

rohan-bes (approved):
Good stuff, thanks for putting in the effort to do everything the recommended way up until now. Just a couple of suggestions but nothing major, pre-approving 👍

PR #4848, opened 2023-08-10
RN-899: Qr Code Visuals in Tupaia 

chris-pollard:
### Changes:

- Move the canvas generation function from admin panel, and create shared components in ui-components package.
- Create components that handle single and multiple qr codes.
- Add QR code viz to `web-frontend` and `tupaia-web`


packages/ui-components/src/components/QrCode/QrCodeContainer.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const handleDownload = async () => {
+    // if(!downloadRef) {
+    //   return
+    // }
+    // await createQrCodeCanvas(downloadRef, humanReadableId, qrCodeContents)
+    // const canvas = downloadRef.current
+
+    // if(!canvas) {
+    //   return
+    // }
```
acdunham, line 42: Just need to remove these commented out parts

packages/ui-components/src/components/QrCode/QrCodeContainer.tsx
```diff
@@ -0,0 +1,67 @@
...
+    
+    const dataUri = await getCanvasUrlForDownload(humanReadableId, qrCodeContents)
+    console.log('data uri for canvas', dataUri)
```
acdunham, line 45: Just need to remove this console log

packages/ui-components/src/components/QrCode/QrCodeContainer.tsx
```diff
@@ -0,0 +1,67 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+
+import React, { useRef, MutableRefObject } from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import { Button } from '../Button';
+import download = require("downloadjs")
```
acdunham, line 10: I don't think we should be using an `import` and a `require` at the same time

packages/ui-components/src/components/QrCode/QrCodeContainer.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { QrCodeImage } from './QrCodeImage';
+import { createQrCodeCanvas, getCanvasUrlForDownload } from './createQrCodeCanvas';
+
+const Container = styled.div`
+  text-align: center;
+`;
+
+const ButtonContainer = styled.div`
+  margin-top: 30px;
```
acdunham, line 19: Can we make this `rem`s?

packages/ui-components/src/components/QrCode/QrCodeContainer.tsx
```diff
@@ -0,0 +1,67 @@
+/*
```
acdunham, line 1: Is this a copy of the one in `admin-panel`? If so, can we remove one of them? If this is only used in `admin-panel` I'd be in favour of removing this one.

packages/ui-components/src/components/QrCode/QrCodeContainer.tsx
```diff
@@ -0,0 +1,67 @@
...
+      <ButtonContainer>
+        <Button type="button" onClick={handleDownload}>
+          Download
+        </Button>
+      </ButtonContainer>
+    </Container>
+  );
+};
+
+QrCodeContainer.propTypes = {
```
acdunham, line 64: We don't need `prop-types` in `ui-components` anymore, as typescript types are superseding this.

packages/ui-components/src/components/QrCode/QrCodeImage.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+
+  return (
+    <Box style={{
+      display: 'flex',
+      justifyContent: 'center',
+      width,
+      margin
+    }}>
+      <StyledCanvas style={{width: '80%'}} ref={ref} />
```
acdunham, line 36: Can this `width` be applied to the `StyledCanvas` styled component? I don't really like the use of inline styling, especially when we already have a styled component we can use.

packages/ui-components/src/components/QrCode/QrCodeImage.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const QrCodeImage = ({ qrCodeContents, humanReadableId, width = undefined, margin = 'auto' }: QrCodeImageProps) => {
+  const ref = useRef<HTMLCanvasElement | null>(null)
+  useEffect(() => {
+    drawQrCodeCanvas(ref, humanReadableId, qrCodeContents);
+  }, [ref]);
+
+
+  return (
+    <Box style={{
+      display: 'flex',
```
acdunham, line 31: Can we make a styled component for this, instead of inline styles? This way it matches the pattern we have already

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+const FormContainer = styled.div`
+  display: flex;
+  justify-content: center;
+  padding: 0px 40px 30px 40px;
+`;
+
+const Error = styled.div`
+  color: ${props => props.theme.palette.error.main};
+  margin-top: 10px;
```
acdunham, line 33: Can we please make this `rems`?

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+  text-align: center;
+`;
+
+const StyledCanvas = styled.canvas`
+  outline: 1px solid #dedede;
+  width: 500px;
+  height: 500px;
```
acdunham, line 40: These should also be in `rems`

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+`;
+
+interface DownloadQrCodeVisualProps {
+  downloadImages: (qrCodeCanvasUrlsWithFileNames: {name: string, url: string | undefined}[]) => Promise<void>;
+  config?: object;
```
acdunham, line 45: I'm not sure that `object` is the best `type` here. Technically it works, but I think we can be more descriptive. If we already have a config type for qr code viz in the `types` package you can use this, and if we don't I think it would be worth making one

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  const noneSelected = Object.fromEntries(
+    options.map(({ value }) => [value, false]),
+  );
+  const [selectedQrCodes, setSelectedQrCodes] = useState(noneSelected);
+
+  const toggleSelectFile = (value: string) =>
+  setSelectedQrCodes({ ...selectedQrCodes, [value]: !selectedQrCodes[value] });
+
+  const [isDownloading, setIsDownloading] = useState(false);
```
acdunham, line 73: This is really pedantic, but can we put all the `useState` things at the top of the component, grouped together?

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+      .filter(({value, name}) => selectedItemValues.includes(value)
+    )
+      .map(async ({name, value}) => {
+      const url = await getCanvasUrlForDownload(name, value)
+      return {
+        url,
+        name
+      }
+    }));
+    console.log('qrCodeCanvasUrls',qrCodeCanvasUrlsWithFileNames)
```
acdunham, line 92: Just need to remove this console log

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+    await downloadImages(qrCodeCanvasUrlsWithFileNames);
+    setIsDownloading(false);
+  };
+
+  if (!isEnlarged) {
+    if(options.length > 1) {
+      return (
+        <Container className={className} maxWidth="sm">
```
acdunham, line 100: I'd be keen to separate these out into a multi qr code display and a single qr code display, for readability

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+          {
+            options.map(({ name, value }, index) => {
+              if(index > 5) {
+                return
+              }
+              return (
+                <Grid 
+                  container 
+                  style={{paddingBottom: 5}}
+                  alignContent="center"
```
acdunham, line 111: Can we get rid of any of the inline styling and put in a styled component?

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+    console.log('qrCodeCanvasUrls',qrCodeCanvasUrlsWithFileNames)
+    await downloadImages(qrCodeCanvasUrlsWithFileNames);
+    setIsDownloading(false);
+  };
+
+  if (!isEnlarged) {
+    if(options.length > 1) {
+      return (
+        <Container className={className} maxWidth="sm">
+          <p style={{color: 'white', textAlign: 'center'}}>QR Code</p>
```
acdunham, line 101: Can we use a styled component for this? Also, instead of setting text color to be white, can we use the theme primary text value?

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+                    onChange={() => toggleSelectFile(value)}
+                    value={value}
+                  />
+                }
+                label={name}
+              />
+            ))}
+          </FormGroup>
+        </FormControl>
+      </FormContainer>
```
acdunham, line 189: Do we already have a `CheckboxList` component we can use here?

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -0,0 +1,203 @@
...
+      <DialogActions>
+        <Button onClick={onClose}>Cancel</Button>
+        <Button
+          color="primary"
+          onClick={() => downloadSelectedQrCodes()}
```
acdunham, line 194: This doesn't need to be an inline function, and there are performance improvements when we don't have inline functions

packages/ui-components/src/components/QrCode/createQrCodeCanvas.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react'
+import { toDataURL } from 'qrcode';
+
+const CODE_SIZE = 320; // width/height at browser display size
+const CANVAS_SIZE = 395; // width/height at browser display size
+const DOWNLOAD_IMAGE_SCALE = 2; // render the canvas at double size to get a larger and clearer image download
+const SCALED_CODE_SIZE = CODE_SIZE * DOWNLOAD_IMAGE_SCALE;
+const SCALED_CANVAS_SIZE = CANVAS_SIZE * DOWNLOAD_IMAGE_SCALE;
```
acdunham, line 8: If any of these values are the same as the ones in admin panel, you could export them from a constants file inside this folder to reuse in admin panel, to save rewriting the same ones.

packages/ui-components/src/components/QrCode/createQrCodeCanvas.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    // Add qr code
+    // @ts-ignore
+    const qrDataUrl = await toDataURL(qrCodeContents, {
+      width: SCALED_CODE_SIZE,
+      height: SCALED_CODE_SIZE,
+    });
+    const img = new Image();
+    const codeX = (SCALED_CANVAS_SIZE - SCALED_CODE_SIZE) / 2;
+    const codeY = 65 * DOWNLOAD_IMAGE_SCALE; // Make sure to avoid the QR code "quiet zone"
```
acdunham, line 39: If any of this logic is the same anywhere else, I'd be keen to make utils for it in this folder, so we don't have repetition

packages/ui-components/src/components/QrCode/useQrCodeCanvas.tsx
```diff
@@ -0,0 +1,209 @@
...
+    // // img.onload = () => {
+    // //   ctx.drawImage(img, codeX, codeY)
+
+    // // };
+    // // @ts-ignore
+    // img.src = qrDataUrl;
+    // await img.decode();
+    // console.log('image ln47', img)
+    // ctx.drawImage(img, codeX, codeY)
+
```
acdunham, line 49: Can we remove these commented out bits?

packages/ui-components/src/components/QrCode/useQrCodeCanvas.tsx
```diff
@@ -0,0 +1,209 @@
...
+    const canvas: HTMLCanvasElement | null = await createQrCodeCanvas(humanReadableId, qrCodeContents)
+    if(!canvas) {
+      return
+    }
+    const url = canvas.toDataURL('image/jpeg', 1.0);
+    console.log('url in nested function',url)
```
acdunham, line 55: Just need to remove this console log

acdunham (changes requested):
Just a few changes in the comments from me. Also, as a general rule, can we please avoid putting any styling inline? It goes against our current pattern. I think there is also room for making some things more reusable as well.

packages/ui-components/src/components/QrCode/QrCodeImage.tsx
chris-pollard: This one is reliant on a prop from the component, so is there a way to do that?

acdunham: Yep, you would pass the styled component props of `width` and `margin` and then use those in tge styled component, similar to how you use themes in styled components.
E.g. 
``` width: ${(props) => props.width};```

File: packages/tupaia-web/src/features/DashboardItem/DashboardItem.tsx [hunk 23]
```diff
@@ -79,12 +79,14 @@ export const DashboardItem = ({ dashboardItem }: { dashboardItem: DashboardItemT
...
   } = config as DashboardItemConfig;
+
   const isExpandable = !!(
     periodGranularity ||
     type === 'chart' ||
     type === 'matrix' ||
     viewType === 'dataDownload' ||
-    viewType === 'filesDownload'
+    viewType === 'filesDownload' ||
+    (viewType === 'qrCodeVisual' && report && report.data && report?.data?.length > 1)
```
rohan-bes, line 89: Based on team tech design on this we're thinking of doing the following:
- Create an `ExpandableVisualWrapper` component that we explicitly wrap any visual that supports expanding
- Move the logic of rendering the `ExpandedVisualModal` into that wrapper
- Create a `useIsExpanded()` hook to check if the viz is expanded rather than passing around an `isExpanded` prop in `tupaia-web`

packages/lesmis/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "author": "Beyond Essential Systems <admin@tupaia.org> (https://beyondessential.com.au)",
   "main": "dist/index.js",
-  "scripts": { 
+  "scripts": {
```
rohan-bes, line 13: Ah thanks 🙏 I keep seeing this change haha

packages/ui-components/src/components/QrCode/utils/generateQrCodeDataUrl.ts
```diff
@@ -0,0 +1,54 @@
...
+  const codeX = TEXT_BOX_WIDTH;
+  const codeY = 0; // Make sure to avoid the QR code "quiet zone"
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  img.src = qrDataUrl;
+  await img.decode();
+  ctx.drawImage(img, codeX, codeY);
+
+  const url = canvas.toDataURL('image/jpeg', 1.0);
+  return url;
```
rohan-bes, line 53:
```suggestion
```
Thoughts on this method being called something like `drawQrCodeToCanvas` and just drawing, and then having the caller get the dataUrl from their canvas if they need? I think I find it a little confusing that we're generating a dataUrl as a means to draw the qr code in contexts where we  aren't interested in the dataUrl

packages/web-frontend/package.json
```diff
@@ -48,10 +48,11 @@
     "core-js": "^3.16.1",
     "dom-to-image": "2.6.0",
     "dotenv": "^8.2.0",
-    "downloadjs": "1.4.7",
+    "downloadjs": "^1.4.7",
     "email-validator": "^1.1.1",
     "history": "^4.7.2",
     "is-mobile": "^0.2.2",
+    "jszip": "^3.10.1",
```
rohan-bes, line 55:
```suggestion
```

rohan-bes (changes requested):
Thanks for helping out on this @acdunham 🙏 

Just a few changes requested

packages/ui-components/src/components/QrCode/QrCodeVisual/QrCodeVisual.tsx
```diff
@@ -0,0 +1,38 @@
...
+  onCloseModal,
+  isEnlarged,
+}: QrCodeVisualProps) => {
+  if (isEnlarged) return <DownloadMultiQrCodes data={options} onCancelDownload={onCloseModal} />;
+
+  return (
+    <Wrapper>
+      {options.length > 1 ? <MultiQrCodes data={options} /> : <SingleQrCode data={options} />}
+    </Wrapper>
+  );
```
rohan-bes, line 37: Been thinking a bit more about my suggestion to move the `isEnlarged` logic out of here and in to `tupaia-web`. Honestly, I think was being a bit too pedantic, I'm pretty happy with either approach 👍

packages/tupaia-web/src/features/Dashboard/ExportDashboard.tsx
```diff
@@ -82,7 +82,7 @@ export const ExportDashboard = ({ isOpen, onClose, dashboardItems }: ExportDashb
       projectCode,
       entityCode,
       dashboardName,
-      selectedDashboardItems: selectedItems.map(({ code }) => code),
+      selectedDashboardItems: selectedItems.map(({ code }) => code!),
```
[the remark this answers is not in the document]
acdunham: I think it was to handle a TS error that popped up which would break the build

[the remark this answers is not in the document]
rohan-bes: Ah I see now. Looks like we changed the `ListItemProps` so that `code` could be `undefined`.

This assertion would be safe, but maybe we could also try making the `CheckboxList` component be generic, so that it can handle any `ListItem` types as long as they extend `ListItemProps`? I might just do that in a separate PR to avoid holding this one up any longer 👍

rohan-bes (approved):
Thanks @acdunham

A couple of comments but I'll put up a PR about one of them and the other probably doesn't block this. Apologies for the delay on reviewing, and really do appreciate the effort you've put  into this 🙏 

Pre-approving

rohan-bes:
Ah sorry @acdunham I think the build failed due to using the `generateQrCodeCanvas` without it being exported and imported. I referenced it there since I think it's the more appropriate function to use, but I should have been more explicit that it'd require updating the imports.

PR #4850, opened 2023-08-11
WAITP-1215 project nav fixes

acdunham:
### Changes:
- Fixed up styling of no data at level dashboard
- Fixed infinite rerender when error with dashboards is caught
- Stopped fetching of incorrect entities when project code changes
- Fixed issue where multiple redirects to dashboards were happening (this and the above were caused by `keepPreviousData`)
- Update handling of active entities on polygon layers to match current web frontend

### DIFF
--- a/packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
+++ b/packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
@@ -12,14 +12,15 @@ import {
   MeasureMarker,
   LayerGroup,
   MeasurePopup,
-  Polygon,
+  BasePolygon,
   AreaTooltip,
   MeasureData,
 } from '@tupaia/ui-map-components';
 import { useEntity } from '../../../api/queries';
 import { useMapOverlayData, useNavigateToEntity } from '../utils';
+import { ActiveEntityPolygon } from './ActiveEntityPolygon';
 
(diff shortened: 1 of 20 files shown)


packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer.tsx
```diff
@@ -49,8 +50,11 @@ export const DataVisualsLayer = ({
   return (
     <LayerGroup>
       {measureData.map((measure: MeasureData) => {
-        const { region, organisationUnitCode: entity, color, name } = measure;
+        const { region, organisationUnitCode: entity, color, name, code } = measure;
         if (region) {
+          if (code === entityCode) {
+            return <ActiveEntityPolygon key={entity} entity={measure} />; // this is so that the polygon is displayed as the active entity, i.e correctlyt shaded etc.
+          }
```
acdunham, line 57: I was conflicted whether to display this in here or just make it return null so that the `PolygonNavigationLayer` handled it. But if we did that then would have to fetch the measure item to get the color anyway

tcaiger: Yes I know what you mean about not knowing whether to put things in here or PolygonNavigationLayer. I am open to suggestions about organising that better if possible but otherwise this seems good.

tcaiger (approved):
Nice

PR #4851, opened 2023-08-11
WAITP-1364 Add map_overlay.config to types package

EMcQ-BES:
### Changes:

- Add `map_overlay.config` definition to the types package
- Move existing enumerated string definitions into types packages
- Update `tupaia-web` to use the new type definitions


packages/tupaia-web/src/types/mapOverlay.d.ts
```diff
@@ -1,33 +0,0 @@
-/*
```
tcaiger, line 1: Great to be able to delete this

tcaiger: Do we need to delete the line from types/index.d.ts as well?

packages/ui-map-components/src/components/Markers/markerIcons.tsx
```diff
@@ -250,71 +250,71 @@ const FadedCircle = ({ color, scale = 1 }: IconProps) => {
 };
 
 const icons = {
-  pin: {
+  [IconKey.PIN]: {
```
tcaiger, line 253: Nice - this is a great improvement

tcaiger (approved):
Looks good to me. It's a huge improvement to be replacing all these random const with types from the types package. Just one question about the index file in tupaia-web/src/types

packages/tupaia-web/src/types/mapOverlay.d.ts
EMcQ-BES: Good catch, removed

PR #4853, opened 2023-08-11
RN-976: QR Code Container Square to Rectangle

chris-pollard:
### Issue RN-976:

- Create image from rectangular canvas, text on left, qr on right.
- Make QRCode images bigger in dashboard visual unenlarged.

Screen Shot 2023-08-11 at 2 08 40 pm


packages/ui-components/src/components/QrCode/QrCodeImage.tsx
```diff
@@ -18,13 +18,15 @@ interface QrCodeImageProps {
   humanReadableId: string;
   width?: number | undefined;
   margin?: number | string | undefined;
+  padding?: number | string | undefined;
```
acdunham, line 21: I wonder, since there are so many styles passed into this now, if it would be better to handle this with styled components in the respective parents? Either by creating a styled component of `QrCodeImage` or doing child styles on a wrapper around them. I don't really love the idea of passing in these things as props.

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -23,6 +22,10 @@ import { NoData } from '../NoData';
 import { QrCodeImage } from './QrCodeImage';
 import { getCanvasUrlForDownload } from './useQrCodeCanvas';
 
+const StyledContainer = styled(Container)`
+  padding-bottom: 1rem;
```
acdunham, line 26: Thanks for using `rems`!

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -35,6 +38,11 @@ const Error = styled.div`
   text-align: center;
 `;
 
+const StyledParagraph = styled.p`
+  color: white;
```
acdunham, line 42: I think we could use the `Typography` component from mui for this, and we would ideally use theme variables for colours as well (if not already done by Typography component).

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -92,58 +100,52 @@ export const QrCodeVisual = ({
   if (!isEnlarged) {
     if (options.length > 1) {
       return (
-        <Container className={className} maxWidth="sm">
-          <p style={{ color: 'white', textAlign: 'center' }}>QR Code</p>
+        <StyledContainer className={className} maxWidth="sm">
+          <StyledParagraph>QR Codes</StyledParagraph>
```
acdunham, line 104: Since this is acting as a title, we should use the correct semantic HTML. Likely an h3, but depends on where it sits on the screen. E.g if there is no h2 anywhere above it, it should be h2, otherwise h3, and so on.
Also, can we name these more descriptively? E.g. instead of `StyledParagraph` it would be `Title` or something?

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -92,58 +100,52 @@ export const QrCodeVisual = ({
...
           {options
-            .filter((option, index) => index < 6)
+            .filter((option, index) => index < 4)
```
acdunham, line 106: I wonder if we do a `slice` or something here instead of a `filter`?

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -92,58 +100,52 @@ export const QrCodeVisual = ({
...
-                    {name}
-                  </Typography>
-                </Grid>
+              <Grid container direction="column">
+                <QrCodeImage
+                  qrCodeContents={value}
+                  humanReadableId={name}
+                  width={200}
+                  margin="auto"
+                  padding="0.5rem 0"
```
acdunham, line 114: I definitely think it would be better to take care of the margin, width, and padding in here instead of passing them as props. Seems like an anti-pattern to pass them through

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -92,58 +100,52 @@ export const QrCodeVisual = ({
...
+        </StyledContainer>
       );
     }
     const option = options[0];
     return (
-      <Container>
-        <p style={{ color: 'white', textAlign: 'center' }}>QR Code</p>
+      <StyledContainer>
+        <StyledParagraph>QR Code</StyledParagraph>
         <Button style={{ position: 'absolute', top: '5px', right: '5px' }}>
```
acdunham, line 125: This might be in your other PR but I'd like to do away with inline styling if possible

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -92,58 +100,52 @@ export const QrCodeVisual = ({
...
-        </Container>
+        </StyledContainer>
       );
     }
     const option = options[0];
     return (
-      <Container>
-        <p style={{ color: 'white', textAlign: 'center' }}>QR Code</p>
+      <StyledContainer>
+        <StyledParagraph>QR Code</StyledParagraph>
```
acdunham, line 124: Should probably be a heading component, like the other one

packages/ui-components/src/components/QrCode/QrCodeVisual.tsx
```diff
@@ -92,58 +100,52 @@ export const QrCodeVisual = ({
...
             ))}
-        </Container>
+        </StyledContainer>
       );
     }
     const option = options[0];
     return (
-      <Container>
-        <p style={{ color: 'white', textAlign: 'center' }}>QR Code</p>
+      <StyledContainer>
```
acdunham, line 123: I think in both cases here there will be a title and a container. It might be a better idea then to create a `const` called something like `displayedQrCodes` and when there is more than 1, slice it to get the top 4, otherwise it would be the array with 1 item in there. Then loop through, and use a container around it. You can use `as` prop on the container to say if there is more than on it would be `Grid` otherwise `div`, e.g. `as={displayedQrCodes.length=== 1 ? 'div' : 'Grid'}`

packages/ui-components/src/components/QrCode/useQrCodeCanvas.tsx
```diff
@@ -66,34 +60,34 @@ export const drawQrCodeCanvas = async (
...
-    width: SCALED_CODE_SIZE,
-    height: SCALED_CODE_SIZE,
+    width: CANVAS_HEIGHT,
+    height: CANVAS_HEIGHT,
   });
   const img = new Image();
-  const codeX = (SCALED_CANVAS_SIZE - SCALED_CODE_SIZE) / 2;
-  const codeY = 65 * DOWNLOAD_IMAGE_SCALE; // Make sure to avoid the QR code "quiet zone"
+  const codeX = 900;
+  const codeY = 0; // Make sure to avoid the QR code "quiet zone"
```
acdunham, line 90: Seems like there are a few repeated numbers in here. Would be good to make them variables like `CANVAS_WIDTH` etc are, so that if we end up changing any, they are just changed in 1 place and we don't accidentally forget some.

acdunham (changes requested):
Good work. A few changes requested from me, mainly around naming conventions and DRY-ing some things up

acdunham (approved):

PR #4854, opened 2023-08-13
WAITP-1283: Optimise Admin Panel Build

tcaiger:
### Changes:

- Update admin-panel build for the app and library to be in series rather than parallel to spread the cpu and memory load
- Add manual chunks config so that the build gets split into smaller files which reduces resource usage on build but also improves front end performance on first load.

Testing
The test branch [waitp-1283-test-vite-build](https://github.com/beyondessential/tupaia/tree/waitp-1283-test-vite-build) is deployed with the same configuration as dev and the build works.

---

### Screenshots:
image


acdunham (approved):
Such a nice fix, thank you for doing this!

PR #4860, opened 2023-08-14
WAITP-1397: Tupaia Maps - Auto-zoom out

WDuncker:
### Changes:

- Updated useEntity query in tupaia-web to fetch point column. 
- Passed point and zoom from Map to leafletMap component.
- switched if statement priority from bounds to point so the map auto movement will always use setView instead of fitBounds if there is a point available.


tcaiger (approved):
This looks great to me. Thanks

PR #4861, opened 2023-08-14
WAITP-1315 query cancellation

acdunham:
### Changes:
- Added in automatic query cancellation by default for get requests on Tupaia Web. This can be opted-out of
- Fixed performance issue in Dashboards view where redirect logic was being called too many times
- Fixed performance issue where every child entity was calling dashboards endpoint. Instead let the default dashboard logic handle redirects, to save on unnecessary requests.


packages/tupaia-web/src/api/api.ts
```diff
@@ -57,8 +60,28 @@ const request = async (endpoint: string, options?: RequestParametersWithMethod)
...
+    method: 'get',
+    ...rest,
+    signal: controller?.signal,
+  }) as Promise<any> & {
+    cancel?: () => void;
+  };
+
+  if (cancelOnUnmount) {
+    promise.cancel = () => controller?.abort();
+  }
```
acdunham, line 81: For some reason, even though we are on v3.39 and cancellation when you pass in a signal was included form v3, I couldn't get cancellation automatically working unless I handled it using this old way. So possibly it is an issue that was fixed later on

tcaiger (approved):
The query cancellation is great. It's a very clean addition to just have it all contained in api.ts. I did some testing and I think it does make the front end less laggy. I think that the queries still load on the backend like Edwin says and so it still takes almost as long to get the actual data but at the very least it's a boost to perceived performance.

It's only loosely related, but can you please also update `retry to 0` in  `AppProviders` which I'd previously configured to 1. I think I prefer failed requests to just display an error on the front end rather than loading for longer as there is a lot of loading time already.

tcaiger:
ps. I had to add ` "@tupaia/types": "1.0.0",` to ui-map-components to get the tupaia-web build to work. I think that might be a bug was introduced elsewhere but could you also add that in here as a tidy up. Thanks

PR #4862, opened 2023-08-14
WAITP-1239: Fixes for static text visuals

acdunham:
### Changes:
- Fix differences in static text visuals
- Fix issue where url for download link is incorrect if there is no trailing `/`

### DIFF
--- a/packages/tupaia-web/src/utils/transformDownloadLink.ts
+++ b/packages/tupaia-web/src/utils/transformDownloadLink.ts
@@ -10,5 +10,6 @@ import { API_URL } from '../api';
 export const transformDownloadLink = (resourceUrl: string) => {
   const baseUrl = API_URL;
   if (resourceUrl.includes('http')) return resourceUrl;
-  return `${baseUrl}${resourceUrl}`;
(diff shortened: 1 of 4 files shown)


packages/tupaia-web/src/utils/transformDownloadLink.ts
```diff
@@ -10,5 +10,6 @@ import { API_URL } from '../api';
 export const transformDownloadLink = (resourceUrl: string) => {
   const baseUrl = API_URL;
   if (resourceUrl.includes('http')) return resourceUrl;
-  return `${baseUrl}${resourceUrl}`;
+  if (baseUrl.endsWith('/')) return `${baseUrl}${resourceUrl}`; // remove trailing slash from baseUrl if it exists
```
tcaiger, line 13: I think that [trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) would do the trick here, right?

acdunham: That would only work with whitespace right?

tcaiger: Oh yeah sorry - you are totally correct - I didn't read it correctly.

tcaiger (approved):
Thanks for the fiddly little fix ups

PR #4863, opened 2023-08-14
WAITP-1283: Fix broken server config

tcaiger:
### Changes:

- Vite builds to dist instead of build and we need to updated the server config to match

---


rohan-bes (approved):
Perfecto

PR #4864, opened 2023-08-14
Waitp 1419 remove unused entity bounds data

WDuncker:
### Changes:

- Added migration to remove all bounds records in the entity table where there is a point record


tcaiger (approved):
Great work. Looks good to me. Let's merge it in to the epic branch too and then it can be tested implicitly with the other changes to the bounds that you did.

PR #4868, opened 2023-08-14
WAITP-1210 Navigation: Entity navigation via search

WDuncker:
### Changes:

- Updated searchResults.tsx in tupaia-web to use qualified_name instead of name as this will show the ancestor entities as well as the target entity when searching.

### DIFF
--- a/packages/tupaia-web/src/features/EntitySearch/SearchResults.tsx
+++ b/packages/tupaia-web/src/features/EntitySearch/SearchResults.tsx
@@ -105,7 +105,7 @@ export const SearchResults = ({ searchValue, onClose }: SearchResultsProps) => {
   return (
     <Container>
       <ScrollBody>
-        {searchResults.map(({ code, qualified_name }) => {
+        {searchResults.map(({ code, qualifiedName }) => {
           return (
             <ResultLink
               button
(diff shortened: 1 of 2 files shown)


packages/tupaia-web/src/features/EntitySearch/SearchResults.tsx
```diff
@@ -105,15 +105,15 @@ export const SearchResults = ({ searchValue, onClose }: SearchResultsProps) => {
...
+        {searchResults.map(({ code, qualified_name }) => {
           return (
             <ResultLink
               button
               key={code}
               onClick={onClose}
               to={{ ...location, pathname: `/${projectCode}/${code}` }}
             >
-              {name}
+              {qualified_name}
```
tcaiger, line 116: ah yes this makes sense - this was added by Ethan in there server and we never updated the frontend. good find! This is a super pedantic request but might as well get it to be consistent. We have a convention of snake case on the server and camel case on the front end and Ethan was supposed to do the conversion in the server but could you add that.

To do so you need add use something called `camelcaseKeys` in the return line of `EntitySearchRoute.ts` in tupaia-web-server, which will do the conversion. See `EntityAncestorsRoute.ts` as an example of the same thing. 

Then you will need to update this to camel case. eg. `qualifiedName`

tcaiger (changes requested):
Looks good. Just a small requested change - thanks

tcaiger (approved):
perfecto

PR #4869, opened 2023-08-15
WAITP-1201: Navigation: Entity navigation via search (issue 3)

WDuncker:
### Changes:

- Updated expandSearchBar function in EntitySearch to an expanding only function rather than a toggle as this was causing the tab changing bug. Behaviour is now consistent with production.

### DIFF
--- a/packages/tupaia-web/src/features/EntitySearch/EntitySearch.tsx
+++ b/packages/tupaia-web/src/features/EntitySearch/EntitySearch.tsx
@@ -3,7 +3,7 @@
  *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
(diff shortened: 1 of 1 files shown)


packages/tupaia-web/src/features/EntitySearch/EntitySearch.tsx
```diff
@@ -3,7 +3,7 @@
  *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
```
acdunham, line 6: I think the build will complain about unused import for `useEffect`

acdunham (approved):
Nice catch! Just that unused import to fix, but otherwise pre-approving

PR #4872, opened 2023-08-15
Waitp 1408 update styling

chaoxinnhin:
### Issue #:
WAITP-1408: Update styling on Tupaia-Web to match design

### Changes:

- Updated Topbar and dashboard design, mini map colour, modal for dashboard items and popover menu

### Screenshots:
image
image
image
image
image


packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -29,7 +29,7 @@ const Panel = styled.div<{
   $isExpanded: boolean;
 }>`
   position: relative;
-  background-color: ${({ theme }) => theme.panel.background};
+  background-color: ${({ theme }) => theme.searchBar.background};
```
acdunham, line 32: This works, but to be really pedantic I would maybe make a theme variable called something more generic, like `backgrounds` that would have `primary` and `secondary` inside instead of using `searchBar` and then everywhere that refers to `searchBar.background` would instead use `backgrounds.primary` and for the lighter black it would refer to `backgrounds.secondary`. Then if we ever want to redo the theme it's clear what we are updating

packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -64,7 +64,7 @@ const TitleBar = styled.div`
   justify-content: space-between;
   align-items: center;
   padding: 1rem;
-  background-color: ${({ theme }) => theme.panel.background};
+  background-color: ${({ theme }) => theme.projectCard.background};
```
acdunham, line 67: Same with the above comment

packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -75,25 +75,29 @@ const ExportButton = styled(Button).attrs({
   variant: 'outlined',
 })`
   font-size: 0.6875rem;
+  margin: 0 1rem;
```
acdunham, line 78: Thanks for using `rems`!

packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -75,25 +75,29 @@ const ExportButton = styled(Button).attrs({
...
   font-size: 1.625rem;
   line-height: 1.4;
+  padding: 0 1rem;
 `;
 
 const DashboardItemsWrapper = styled.div<{
   $isExpanded: boolean;
 }>`
+  padding-left: ${({ $isExpanded }) => ($isExpanded ? '2rem' : '0')};
+  padding-right: ${({ $isExpanded }) => ($isExpanded ? '2rem' : '0')};
```
acdunham, line 93: You could probably simplify this to be `padding: ${({ $isExpanded }) => ($isExpanded ? '0 2rem' : '0')}`

packages/tupaia-web/src/features/Dashboard/DashboardMenu.tsx
```diff
@@ -4,21 +4,37 @@
...
 import { Dashboard } from '../../types';
 
 const MenuButton = styled(ButtonBase)`
   display: flex;
-  justify-content: space-between;
-  align-items: center;
-  background-color: ${({ theme }) => theme.panel.secondaryBackground};
+  justify-content: flex-start;
+  align-items: flex-end;
+  background-color: #202124;
```
acdunham, line 16: Should be using a theme variable here, since where possible we want to avoid hard coded colours, if they're colours we use in multiple places

packages/tupaia-web/src/features/Dashboard/DashboardMenu.tsx
```diff
@@ -61,11 +77,11 @@ export const DashboardMenu = ({
       {activeDashboard && (
         <MenuButton onClick={handleClickListItem} disabled={!hasMultipleDashboards}>
           {activeDashboard?.name}
-          {hasMultipleDashboards && <ArrowDropDownIcon />}
+          {hasMultipleDashboards && <KeyboardArrowDownIcon />}
         </MenuButton>
       )}
 
-      <Menu
+      <ItemButton
```
acdunham, line 84: Can we rename this to be more descriptive? `ItemButton` implies one single button, but this is for the whole menu

packages/tupaia-web/src/features/DashboardItem/ExpandItemButton.tsx
```diff
@@ -33,16 +33,15 @@ const ExpandableButton = styled(Button).attrs({
...
     top: 0;
     left: 0;
     opacity: 0;
     margin-top: 0;
     &:hover,
     &:focus-visible {
-      opacity: 0.7;
-      background-color: ${({ theme }) => theme.palette.common.black};
+      background-color: rgba(32, 33, 36, 0.6);
+      opacity: 1;
```
acdunham, line 44: If you have an `rgba` value then you don't need `opacity` set as well, since the `a` in `rgba` is the opacity of the colour

packages/tupaia-web/src/features/DashboardItem/ExpandItemButton.tsx
```diff
@@ -54,8 +53,8 @@ const ExpandButtonText = styled.span`
 const ZoomInIcon = styled(MuiZoomIcon)`
   margin-right: 1rem;
   @media screen and (min-width: ${MOBILE_BREAKPOINT}) {
-    width: 3rem;
-    height: 3rem;
+    width: 1.5rem;
+    height: 1.5rem;
```
acdunham, line 57: These look quite small to me. If you haven't already, can you send Flic a screenshot of before and after this one change, to make sure it's correct?

packages/tupaia-web/src/features/Map/Map.tsx
```diff
@@ -48,13 +48,14 @@ const StyledMap = styled(LeafletMap)`
     top: -50px;
     right: 3px;
     a {
-      background: rgba(43, 45, 56, 0.8);
+      background: #34353f;
```
acdunham, line 51: I'd be tempted to make this a theme variable as well

packages/tupaia-web/src/features/Map/Map.tsx
```diff
@@ -48,13 +48,14 @@ const StyledMap = styled(LeafletMap)`
...
       box-shadow: none;
       border: none;
       color: white;
 
       &:hover {
-        background: ${TRANSPARENT_BLACK};
+        background: ${({ theme }) =>
+          theme.palette.type === 'light' ? 'white' : 'rgba(43, 45, 56, 0.7)'};
```
acdunham, line 58: I think we don't need to worry about light theme right now. I wonder if we can make use of `lighten` from `material-ui/core` here, instead of introducing another color?

packages/tupaia-web/src/features/Visuals/View/DataDownload.tsx
```diff
@@ -16,6 +16,7 @@ import { URL_SEARCH_PARAMS } from '../../../constants';
 const ListItem = styled.li`
   text-align: center;
   font-size: 1.25rem;
+  background-color: red;
```
acdunham, line 19: I think this is a mistake?

packages/tupaia-web/src/layout/TopBar/TopBar.tsx
```diff
@@ -21,8 +21,7 @@ const Header = styled.header<{
   $primaryColor?: string | null;
   $secondaryColor?: string | null;
 }>`
-  background-color: ${({ $primaryColor, theme }) =>
-    $primaryColor || theme.palette.background.default};
+  background-color: #2e2f33;
```
acdunham, line 24: Here we do want to use `$primaryColor` if set, because it might come from custom landing pages. We can also use a theme variable for the background color if `$primaryColor` is not set. You could potentially change `theme.palette.background.default` to be this black, since it gets used a lot

packages/tupaia-web/src/layout/TopBar/TopBar.tsx
```diff
@@ -33,8 +32,7 @@ const Header = styled.header<{
   padding: 0 0.625em;
 
   > * {
-    background-color: ${({ $primaryColor, theme }) =>
-      $primaryColor || theme.palette.background.default};
+    background-color: ${({ theme }) => theme.projectCard.background};
```
acdunham, line 35: Same as above - need to keep structured as before, but you can update theme variables

packages/tupaia-web/src/layout/UserMenu/PopoverMenu.tsx
```diff
@@ -35,7 +35,11 @@ export const PopoverMenu = ({
 }: PopoverMenuProps) => {
   return (
     <Popover
-      PaperProps={{ style: { backgroundColor: primaryColor } }}
+      PaperProps={{
+        style: {
+          backgroundColor: '#2E2F33',
```
acdunham, line 40: Similarly, needs to be `primaryColor` but you can change the default primaryColor to be this black. The reason is that if we have a custom landing page that has, say, white set as the primary color, then that white won't get carried through if we set to be black all the time

packages/tupaia-web/src/layout/UserMenu/UserMenu.tsx
```diff
@@ -21,6 +21,7 @@ const UserMenuContainer = styled.div<{
 }>`
   display: flex;
   align-items: center;
+  border: 1px white;
```
acdunham, line 24: Can we use this logic here? `${({ secondaryColor, theme }) => secondaryColor || theme.palette.text.primary}`

packages/ui-components/src/components/ActionsMenu.tsx
```diff
@@ -21,6 +21,7 @@ const StyledMenu = styled(MuiMenu)`
     .MuiList-root {
       padding-top: 0.25rem;
       padding-bottom: 0.25rem;
+      background-color: white;
```
acdunham, line 24: Just be careful here, because this gets imported to multiple projects.

packages/ui-components/src/components/Inputs/MultiSelect.tsx
```diff
@@ -19,6 +19,7 @@ const StyledCheckbox = styled(Checkbox)`
 const MenuItem = styled(MuiMenuItem)`
   padding-top: 0;
   padding-bottom: 0;
+  background-color: white;
```
acdunham, line 22: Same here, this gets imported in multiple places and might have unintended consequences.

packages/ui-map-components/src/components/TilePicker/TileControl.tsx
```diff
@@ -18,8 +18,7 @@ const StyledButton = styled(Button)<{
   display: block;
   pointer-events: auto;
   box-shadow: none;
-  background: ${({ theme }) =>
-    theme.palette.type === 'light' ? 'rgba(255, 255, 255, 0.9)' : 'rgba(43, 45, 56, 0.9)'};
+  background: #34353f;
```
acdunham, line 21: If we set one of the mui theme variables to be this black then we should be able to refer to it here. The main reason being that themes can change

acdunham (changes requested):
This is looking really nice. I have a few requested changes, mainly around using theme variables to keep things consistent.

packages/ui-map-components/src/components/TilePicker/TileControl.tsx
```diff
@@ -18,8 +18,7 @@ const StyledButton = styled(Button)<{
   display: block;
   pointer-events: auto;
   box-shadow: none;
-  background: ${({ theme }) =>
-    theme.palette.type === 'light' ? 'rgba(255, 255, 255, 0.9)' : 'rgba(43, 45, 56, 0.9)'};
+  background: ${({theme}) => theme.navigationBtn.main};
```
acdunham, line 21: So we can't use non-mui theme variables (e.g. custom ones like this) in the ui-components (and similar) libraries, because not all the packages that consume this will have this theme structure. So perhaps you can just hardcode it in after all

acdunham (changes requested):
This is looking really nice! I have one small change requested. Also, you might want to ask Felicity if she has planned colour changes for the map overlay selector (attached), now that things aren't really blue anymore. If it's just colour changes to that then that would be in scope I think.

Screenshot 2023-08-17 at 2 35 44 pm

chaoxinnhin:
> This is looking really nice! I have one small change requested. Also, you might want to ask Felicity if she has planned colour changes for the map overlay selector (attached), now that things aren't really blue anymore. If it's just colour changes to that then that would be in scope I think.
> 
> Screenshot 2023-08-17 at 2 35 44 pm

Thanks Alex! I've spoken to Felicity and she is still working on that map overlay. I'm assuming that she's doing a whole new redesign for that as well

acdunham (approved):

PR #4874, opened 2023-08-16
WAITP-1283 fix build

acdunham:
### Changes:
- Fixed typo in chunking config
- Removed max memory change build script


edmofro (approved):
Thanks for fixing this!

PR #4876, opened 2023-08-16
RN-955 Remove dependency cycles

IgorNadj:
### Changes:
- Remove aggregator <-> data-broker circular dependency
- Remove indicators <-> data-broker circular dependency
- Build using yarn workspaces
- Enforce no circular deps in CI/CD

### Timings
I replaced the root `yarn build` command to use `yarn workspaces foreach` which checks for circular dependencies. This also runs in parallel, which gives us some speed improvements. `yarn build` on 2020 M1 mac:
- **Before**: 8m 45s
- **After**: 4m (2x faster)


package.json
```diff
@@ -18,7 +18,7 @@
     ]
   },
   "scripts": {
-    "build": "yarn build:internal-dependencies && yarn build:non-internal-dependencies",
+    "build": "yarn workspaces foreach -ptv --jobs unlimited --exclude tupaia run build",
```
IgorNadj, line 21: p: parallel
t: topological (leaf nodes first, then parents etc)
v: verbose, prefixes each line with the package name
--jobs unlimited: use all the cpus you have
--exclude tupaia: avoid re-running build in root package

IgorNadj:
The general idea with removing the dep cycles e.g. `data-broker -> indicators -> data-broker` was to consider the bigger package the one that should know about the smaller package, and the smaller package can be passed in a runtime DataBroker, without knowing anything about it. 

We discussed a potential DataBrokerInterface which could be used in indicators to guarantee the shape of DataBroker but it wasn't needed to untangle the cycles, and I think the necessity of this could be a question for whoever needs it in the future.

packages/indicators/src/utils.ts
```diff
@@ -24,6 +25,7 @@ export const analyticsToAnalyticClusters = (analytics: Analytic[]): AnalyticClus
   return Object.values(clusterMap);
 };
 
+// TODO: use the one in tsutils
```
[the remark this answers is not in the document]
rohan-bes: I think it's all good 👍

packages/data-broker/src/__tests__/__integration/indicators/indicators-integration.test.ts
rohan-bes, line 1: I quite like this move actually for the integration tests. It makes sense that data-broker would run the integration testing, since it's part of the data pipeline. And the data services can just unit test themselves 👍

rohan-bes (approved):
Great stuff, thanks @IgorNadj 🙏

IgorNadj:
Checked e2e test run, all looks good.

PR #4878, opened 2023-08-17
Waitp 1244 chart fixes

acdunham:
### Changes:
- Fix issue with data not correctly displaying on charts
- Fix dashboard sizing and chart sizing
- Fix axis labels being incorrectly calculated due to base font size being different to current frontend


tcaiger (approved):
Looks good to me

PR #4886, opened 2023-08-17
RN-977: Add option to download QR code in Meditrak

rohan-bes:
### Issue RN-977:

Thanks to the formatting of the multiple QR codes and single QR code now being the same, was able to combine all the logic into a single component.

Used `react-native-file-access` to allow writing the images to the user's 'Downloads' folder. Had to use a pretty old version to be compatible with our version of `react-native` but it seems to do the job.


packages/meditrak-app/app/assessment/QrCodeScreen.js
```diff
@@ -0,0 +1,162 @@
...
+  transparentButtonLabel: {
+    color: THEME_COLOR_ONE,
+  },
+});
+
+const mapStateToProps = () => ({});
+
+const mapDispatchToProps = dispatch => ({
+  displayDownloadMessage: message => dispatch(addMessage('qr_code_download', message)),
+});
```
rohan-bes, line 160: Had to use redux connect here to be able to display the download success message.

packages/meditrak-app/app/assessment/QrCodeScreen.js
```diff
@@ -0,0 +1,162 @@
...
+      <Heading
+        text={`Share QR ${formatPlural('Code', 'Codes', qrCodes.length)}`}
+        style={localStyles.heading}
+      />
+      <ScrollView>
+        {qrCodes.map(({ data, name }, index) => (
+          <View key={data} style={localStyles.qrCodeContainer}>
+            <QrCode
+              getRef={ref => {
+                qrCodeImgRefs.current[index] = ref;
```
chris-pollard, line 59: Looks good. The only query I have is probably more semantics. But when you create qrCodeImgRefs, this is technically one ref right, with the current value an array. So is it correct to be referring to the element values as refs?

chris-pollard (approved):

packages/meditrak-app/app/assessment/QrCodeScreen.js
rohan-bes: Ah good point, technically it should be `qrCodesRef` since it's one ref to many codes 👍 I'll rename

PR #4888, opened 2023-08-18
WAITP-1222: Fix map overlay reports config handling

tcaiger:
### Changes:

Web config server does some specific map overlay config handling that is needed for both legacy and modern reports. In the re-write we were going around that code for modern reports which is causing issues so this change puts it back.


packages/tupaia-web-server/src/routes/MapOverlaysRoute.ts
```diff
@@ -49,7 +51,7 @@ export class MapOverlaysRoute extends Route<MapOverlaysRequest> {
           comparisonValue: [projectCode],
         },
       },
-      pageSize,
+      pageSize: pageSize || DEFAULT_PAGE_SIZE,
```
tcaiger, line 54: this is not related to the core issue, but I noticed while testing that the route was only returning a sub set of map overlays so re-instated that default page size here and fixed it

EMcQ-BES (approved):
Looks good, but I think you might get a CI failure for not re-generating the types

PR #4889, opened 2023-08-18
Rn 899 qr code refactor

acdunham:
### Changes:
- refactor of dashboard items to a `context` model, so that props don't get passed around as much

### DIFF
--- a/packages/tupaia-web/src/features/DashboardItem/ExpandItemButton.tsx
+++ b/packages/tupaia-web/src/features/DashboardItem/ExpandItemButton.tsx
@@ -3,14 +3,14 @@
  * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
-import React from 'react';
+import React, { useContext } from 'react';
 import styled from 'styled-components';
 import MuiZoomIcon from '@material-ui/icons/ZoomIn';
 import { useSearchParams } from 'react-router-dom';
-import { ViewConfig } from '@tupaia/types';
 import { Button } from '@tupaia/ui-components';
 import { MOBILE_BREAKPOINT, URL_SEARCH_PARAMS } from '../../constants';
-import { DashboardItem } from '../../types';
+import { DashboardItemConfig, ViewReport } from '../../types';
+import { DashboardItemContext } from './DashboardItemContext';
 
 const ExpandableButton = styled(Button).attrs({
   variant: 'outlined',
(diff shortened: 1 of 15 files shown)


packages/ui-map-components/src/utils/markerColors.ts
```diff
@@ -7,10 +7,7 @@
 import moment from 'moment';
 import { blue, red, green } from '@material-ui/core/colors';
 import { ScaleType, MeasureColorScheme } from '@tupaia/types';
-import {
-  BREWER_PALETTE,
-  HEATMAP_UNKNOWN_COLOR,
-} from '../constants';
+import { BREWER_PALETTE, HEATMAP_UNKNOWN_COLOR } from '../constants';
```
acdunham, line 10: This is just formatting

packages/ui-map-components/package.json
```diff
@@ -1,49 +1,50 @@
...
-        "@material-ui/styles": "^4.9.10",
-        "@mui/types": "^7.2.4",
-        "@storybook/react": "^6.3.9",
-        "@types/jest": "^29.5.1",
-        "@types/leaflet": "^1.7.1",
-        "@types/react-dom": "^16.9.18",
-        "@types/styled-components": "^5.1.26",
-        "react-docgen-typescript-plugin": "^1.0.5"
-    }
+  "name": "@tupaia/ui-map-components",
```
acdunham, line 2: This is formatting + fixing a type error that broke the build

packages/tupaia-web/src/features/DashboardItem/ExpandItemButton.tsx
```diff
@@ -60,29 +60,49 @@ const ZoomInIcon = styled(MuiZoomIcon)`
...
+  };
+
+  if (!getIsExpandable()) return null;
+
+  const getText = () => {
+    if (viewType === 'dataDownload' || viewType === 'qrCodeVisual')
+      return 'Expand to download data';
+    return 'Expand chart';
+  };
+
```
rohan-bes, line 94: Glad to have this living here rather than where it was previously 👍

rohan-bes (approved):
Thanks for putting in the effort to explore and clean up this one @acdunham 🙏 

I really like the addition of the context to share the dashboard item state throughout the component tree. Much nicer than passing through props everywhere.

I took a stab at pushing this one a bit further to iron out the 'isExpandable' logic. I think there's potential for us to get to a really clean place, but the code diff was starting to get pretty large. Also, a tech design is probably needed to create some clearer concepts and for us to decide what the viz components should be responsible for.

Happy to take the wins you've done here and plan to take things further in the maintenance sprints.

PR #4893, opened 2023-08-21
RN-865: Trigger adding looks at allow list instead of block list

chris-pollard:
### Changes:
- Change the exclusion list to an allow list for tables that require triggers.

### DIFF
--- a/packages/database/src/runPostMigration.js
+++ b/packages/database/src/runPostMigration.js
@@ -6,21 +6,30 @@
 import { arrayToDbString } from './utilities';
 
 const TABLES_REQUIRING_TRIGGER_CREATION = [
-  'entity_hierarchy',
-  'analytics_refresher',
-  'survey_response',
-  'entity',
-  'country',
+  'access_request',
+  'ancestor_descendant_relation',
+  'answer',
   'clinic',
+  'country',
+  'data_element',
+  'entity',
+  'entity_hierarchy',
+  'entity_relation',
   'geographical_area',
+  'indicator',
+  'map_overlay',
+  'map_overlay_group',
+  'meditrak_device',
   'option',
   'option_set',
   'permission_group',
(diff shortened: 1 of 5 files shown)


packages/database/src/runPostMigration.js
```diff
@@ -5,24 +5,22 @@
...
+  'clinic',
+  'geographical_area',
+  'option',
+  'option_set',
+  'permission_group',
+  'question',
+  'survey',
+  'survey_group',
+  'survey_screen',
+  'survey_screen_component',
```
rohan-bes, line 23: I did a bit of a dig and found these tables all needed triggers:
```suggestion
  'access_request',
  'ancestor_descendant_relation',
  'answer',
  'clinic',
  'country',
  'data_element',
  'entity',
  'entity_hierarchy',
  'entity_relation',
  'geographical_area',
  'indicator',
  'map_overlay',
  'map_overlay_group',
  'meditrak_device',
  'option',
  'option_set',
  'permission_group',
  'question',
  'survey',
  'survey_group',
  'survey_response',
  'survey_screen',
  'survey_screen_component',
  'user_entity_permission',
```
I think the main ones missing were due to the `ChangeHandler` classes (eg. `AnalyticsRefresher` and `EntityHierarchyCacher`) and the model notifiers

rohan-bes (changes requested):
Think there's a few missing here. We're looking for anything that calls down to `database.addChangeHandlerForCollection()`

Technically every model is listening to itself, for the purposes of cache functions. I'll push a commit to change that so that we only listen if the model has 'caching enabled', thus reminding devs to update this list

rohan-bes (approved):
Think we're good to go now yeah. Cheers @chris-pollard

rohan-bes:
@chris-pollard I think this is still good to go

PR #4894, opened 2023-08-21
added information hover commit

chaoxinnhin:
### Issue #:
WAITP-1239: Misc. Visuals: Static text

### Changes:

- Hover effect on certain data

### Screenshots:

image

### DIFF
--- a/packages/tupaia-web/src/features/DashboardItem/DashboardInfoButton.tsx
+++ b/packages/tupaia-web/src/features/DashboardItem/DashboardInfoButton.tsx
@@ -1,0 +1,67 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
+import styled from 'styled-components';
+import { ViewConfig } from '@tupaia/types';
+import MuiInfoIcon from '@material-ui/icons/Info';
+import { MOBILE_BREAKPOINT } from '../../constants';
+
+const OpenDashboardInfoButton = styled.div`
+  text-transform: none;
+  font-weight: ${({ theme }) => theme.typography.fontWeightRegular};
+  cursor: pointer;
+  color: ${({ theme }) => theme.palette.common.white};
+  border-color: ${({ theme }) => theme.palette.common.white};
+  padding: 0.3rem;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction: column;
+  gap: 0.3rem;
+  margin-top: 1rem;
+  @media screen and (min-width: ${MOBILE_BREAKPOINT}) {
+    .MuiButton-label {
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
+      align-items: center;
+    }
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    background-color: ${({ theme }) => theme.palette.common.black};
+    border: none;
+    top: 0;
+    left: 0;
+    opacity: 0;
+    margin-top: 0;
+    &:hover,
+    &:focus-visible {
+      opacity: 0.9;
+      background-color: ${({ theme }) => theme.palette.common.black};
+    }
+  }
+`;
+
+const DashboardInfoButtonText = styled.span`
+  font-size: 1rem;
+`;
+
+interface DashboardInfoButtonProps {
+  viewType?: ViewConfig['viewType'];
+  infoText: string | undefined;
+}
+
+export const DashboardInfoButton = ({ viewType, infoText }: DashboardInfoButtonProps) => {
+  const content = <DashboardInfoButtonText>{infoText}</DashboardInfoButtonText>;
+
+  return (
+    <OpenDashboardInfoButton>
+      <MuiInfoIcon />
+      <DashboardInfoButtonText>{viewType ? null : content}</DashboardInfoButtonText>
+    </OpenDashboardInfoButton>
+  );
+};
--- a/packages/tupaia-web/src/features/DashboardItem/DashboardInfoHover.tsx
+++ b/packages/tupaia-web/src/features/DashboardItem/DashboardInfoHover.tsx
@@ -5,11 +5,10 @@
 
 import React from 'react';
 import styled from 'styled-components';
-import { ViewConfig } from '@tupaia/types';
 import MuiInfoIcon from '@material-ui/icons/Info';
 import { MOBILE_BREAKPOINT } from '../../constants';
 
-const OpenDashboardInfoButton = styled.div`
+const Wrapper = styled.div`
   text-transform: none;
   font-weight: ${({ theme }) => theme.typography.fontWeightRegular};
   cursor: pointer;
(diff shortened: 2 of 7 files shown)


packages/tupaia-web/src/features/DashboardItem/DashboardInfoButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
+import styled from 'styled-components';
+import { ViewConfig } from '@tupaia/types';
+import MuiInfoIcon from '@material-ui/icons/Info';
+import { MOBILE_BREAKPOINT } from '../../constants';
+
+const OpenDashboardInfoButton = styled.div`
```
acdunham, line 12: Can we rename this to just be `Wrapper` or something? It's not actually a button so we should be more descriptive with what it is

packages/tupaia-web/src/features/DashboardItem/DashboardInfoButton.tsx
```diff
@@ -0,0 +1,67 @@
+/**
```
acdunham, line 1: Let's rename this to be `DashboardInfoHover`, since it isn't a button

packages/tupaia-web/src/features/Map/utils/useHiddenMapValues.ts
```diff
@@ -26,4 +26,5 @@ export const useHiddenMapValues = (serieses: Series[] = []) => {
   };
 
   return { setValueHidden, hiddenValues };
+  console.log(hiddenValues);
```
acdunham, line 29: Just need to remove this console log

packages/tupaia-web/src/features/DashboardItem/DashboardInfoButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+  font-size: 1rem;
+`;
+
+interface DashboardInfoButtonProps {
+  viewType?: ViewConfig['viewType'];
+  infoText: string | undefined;
+}
+
+export const DashboardInfoButton = ({ viewType, infoText }: DashboardInfoButtonProps) => {
+  const content = <DashboardInfoButtonText>{infoText}</DashboardInfoButtonText>;
```
acdunham, line 59: Can we get this to return `null` if `infoText` is not set? It also looks like this is only visible for `SingleValue` view types, and only when `periodGranularity` is not set in the config. So perhaps instead of rendering this for all view types, just add to the single value one. Unless Andrew knows of any other types? 
If it's just the one type, then it doesn't even need to be a separate file, it can just be a styled component inside the SingleValue file

packages/tupaia-web/src/features/DashboardItem/DashboardInfoButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+const DashboardInfoButtonText = styled.span`
+  font-size: 1rem;
+`;
+
+interface DashboardInfoButtonProps {
+  viewType?: ViewConfig['viewType'];
+  infoText: string | undefined;
+}
+
+export const DashboardInfoButton = ({ viewType, infoText }: DashboardInfoButtonProps) => {
```
acdunham, line 58: Looks like we never pass in `viewType` so can probably remove anything about this one

packages/tupaia-web/src/features/Visuals/View/View.tsx
```diff
@@ -86,15 +87,18 @@ export const View = ({ report, config, isEnlarged }: ViewProps) => {
...
   if (!Component) return null;
 
+
   const formattedData = formatData(data, viewConfig);
-  return (
+  return ( 
+    <>
     <Component
       report={{
-        ...report,
```
acdunham, line 93: Is there a reason we removed this?

acdunham (changes requested):
Nice work finding where this one was. I have some suggestions for a few improvements, in the comments.

packages/tupaia-web/src/features/Visuals/View/View.tsx
chaoxinnhin: oops that that was by accident

packages/tupaia-web/src/features/Visuals/View/View.tsx
```diff
@@ -13,7 +13,7 @@ import { formatDataValueByType } from '@tupaia/utils';
 import { MultiValueRow } from './MultiValueRow';
 import { DataDownload } from './DataDownload';
 import { DownloadFiles } from './DownloadFiles';
-
+import { DashboardInfoHover } from '../../DashboardItem/DashboardInfoHover';
```
acdunham, line 16: I would add `DashboardInfoHover` as an export to `DashboardItem/index.ts` and then this import can become `...... from `../../DashboardItem`

packages/tupaia-web/src/features/DashboardItem/DashboardInfoHover.tsx
```diff
@@ -0,0 +1,63 @@
...
+const DashboardInfoHoverText = styled.span`
+  font-size: 1rem;
+`;
+
+interface DashboardInfoHoverProps {
+  infoText: string | undefined;
+}
+
+export const DashboardInfoHover = ({ infoText }: DashboardInfoHoverProps) => {
+  return (
```
acdunham, line 57: Are there any cases where the hover shouldn't show? E.g if there is an expand button already there, or if `infoText` is unset? We should handle these cases too

acdunham (changes requested):
This is looking better, but I have a couple of questions around how to handle when we don't want this hover to happen

packages/tupaia-web/src/features/DashboardItem/DashboardInfoHover.tsx
```diff
@@ -57,7 +58,7 @@ export const DashboardInfoHover = ({ infoText }: DashboardInfoHoverProps) => {
   return (
     <Wrapper>
       <MuiInfoIcon />
-      <DashboardInfoHoverText>{infoText}</DashboardInfoHoverText>
+      <DashboardInfoHoverText>{!infoText ? null : infoText}</DashboardInfoHoverText>
```
acdunham, line 61: So what I meant here was that `DashboardInfoHoverText` should return null if infoText is unset

acdunham (changes requested):
Almost! Just one last change

packages/tupaia-web/src/features/DashboardItem/DashboardInfoHover.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const DashboardInfoHoverText = styled(Typography)`
+  font-size: 1rem;
+`;
+
+interface DashboardInfoHoverProps {
+  infoText: string | undefined;
+}
+
+export const DashboardInfoHover = ({ infoText }: DashboardInfoHoverProps) => {
```
acdunham, line 57:
```suggestion
export const DashboardInfoHover = ({ infoText }: DashboardInfoHoverProps) => {
   if(!infoText) return null;
```

packages/tupaia-web/src/features/DashboardItem/DashboardInfoHover.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const content = <DashboardInfoHoverText>{infoText}</DashboardInfoHoverText>;
+  return (
+    <Wrapper>
+      <MuiInfoIcon />
+      {!infoText ? null : content}
```
acdunham, line 62:
```suggestion
      {content}
```

acdunham (changes requested):

acdunham (approved):
Thanks!

PR #4899, opened 2023-08-22
WAITP-1277 datatrak-web login form part 1

acdunham:
### Changes:
- Moved Login form to ui components
- Added view for login to datatrak-web
- Added view for verify email to datatrak-web
- Moved layout files to layout folder
- Added user menu to datatrak-web

---

### Screenshots:
Screenshot 2023-08-22 at 3 53 40 pm

Screenshot 2023-08-22 at 3 53 58 pm

Screenshot 2023-08-22 at 4 30 54 pm


packages/datatrak-web/src/theme/theme.ts
```diff
@@ -15,7 +15,14 @@ export const theme = createMuiTheme({
...
     text: {
       primary: '#333333', // dark text color
-      secondary: '#4A79EA', // purple text for secondary buttons etc
+      secondary: '#B8B8B8', // light grey text color
+    },
+  },
+  overrides: {
+    MuiButton: {
+      root: {
+        textTransform: 'none',
```
tcaiger, line 24: Nice use of the theme

packages/ui-components/src/features/HookForm/HookForm.tsx
```diff
@@ -0,0 +1,28 @@
...
+  onSubmit?: SubmitHandler<any>;
+  children: React.ReactNode;
+  className?: string;
+}
+
+/**
+ * A wrapper around react-hook-form's FormProvider and form element. Needed for custom form fields
+ * such as TextField and CheckboxField to be able to self register and validate with react-hook-form.
+ */
+export const HookForm = ({ formContext, onSubmit, children, className }: FormProps) => {
```
tcaiger, line 20: I don't feel too strongly so feel free to push back but it would be nice to not have to include Hook in the name. It is nice and extensible, if we added another form library then we could call it FormikForm etc but in reality I imagine we will only support one form type at any given time so could we just name these `Form` and `FormInput`?

packages/ui-components/src/features/Auth/LoginForm.tsx
```diff
@@ -0,0 +1,107 @@
...
+const StyledForm = styled(HookForm)`
+  margin-top: 1rem;
+  width: 22rem;
+  max-width: 100%;
+`;
+
+const LinkText = styled(Typography)`
+  font-weight: 400;
+  font-size: 11px;
+  line-height: 15px;
```
tcaiger, line 31: any reason not to use rems?

packages/ui-components/src/features/Auth/EmailVerificationDisplay.tsx
```diff
@@ -0,0 +1,33 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2021 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
+import styled from 'styled-components';
+import Typography from '@material-ui/core/Typography';
+
+export enum STATUS {
```
tcaiger, line 9: Since this is exported out of ui-components and will be imported into other contexts, could it be a more descriptive name such as EMAIL_VERIFICATION_STATUS? Or there could be a FORM_STATUS enum that is used in a bunch of places?

packages/ui-components/src/features/Auth/AuthViewButton.tsx
```diff
@@ -0,0 +1,16 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import styled from 'styled-components';
+import { Button } from '../../components';
+
+export const AuthViewButton = styled(Button)`
```
tcaiger, line 9: What do you think about calling this `SubmitButton` instead?

packages/ui-components/src/features/Auth/AuthViewButton.tsx
```diff
@@ -0,0 +1,16 @@
...
+  text-transform: none;
+  font-size: 1rem;
+  width: 22rem;
+  max-width: 100%;
+  margin-left: 0 !important;
```
tcaiger, line 14: Does it actually need important? Can we just use a couple of MUI classes instead?

packages/datatrak-web/src/views/LoginPage.tsx
```diff
@@ -0,0 +1,45 @@
...
+  message?: {
+    status?: STATUS | string;
+    text?: string;
+  };
+}) => {
+  const formContext = useForm();
+  const { mutate: login, isLoading, error } = useLogin();
+  return (
+    <Wrapper>
+      <LoginForm
```
tcaiger, line 34: This implementation of the LoginForm looks great!

packages/datatrak-web/src/views/LoginPage.tsx
```diff
@@ -0,0 +1,45 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
+import styled from 'styled-components';
+import { SubmitHandler, useForm } from 'react-hook-form';
+import { LoginForm, STATUS } from '@tupaia/ui-components';
```
tcaiger, line 8: I assume this is the status from the EmailVerification which doesn't make sense to use in the login form so another reason to make a generic FORM_STATUS enum.

packages/datatrak-web/src/constants/url.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+export const ROUTES = {
```
tcaiger, line 6: Good idea putting these in a const

packages/datatrak-web/src/api/api.ts
```diff
@@ -7,8 +7,7 @@ import axios from 'axios';
 import FetchError from './fetchError';
 
 // TODO: change this to be the correct default url when orchestration server is ready
-export const API_URL =
-  import.meta.env.REACT_APP_DATATRAK_WEB_API_URL || 'http://localhost:8000/api/v1';
+export const API_URL = import.meta.env.REACT_APP_DATATRAK_WEB_API_URL || 'http://localhost:8100/v1';
```
tcaiger, line 10:
```suggestion
export const API_URL = import.meta.env.REACT_APP_DATATRAK_WEB_API_URL || 'http://localhost:8110/v1';
```
This is not decided as 8110

packages/ui-components/src/features/Auth/AuthViewWrapper.tsx
```diff
@@ -0,0 +1,45 @@
...
+import styled from 'styled-components';
+import { Paper, Typography } from '@material-ui/core';
+
+const Wrapper = styled(Paper)`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  max-width: 100%;
+  box-shadow: none;
+  padding: 3rem 0;
```
tcaiger, line 16: I think there should be some left and right padding so that messages and error messages have some spacing on the sides

packages/datatrak-web/src/Routes.tsx
```diff
@@ -4,14 +4,33 @@
...
+
+/**
+ * If the user is logged in and tries to access the login page, redirect to the home page
+ */
+const LoginRedirect = () => {
+  const { isLoggedIn, isLoading, isFetched } = useUser();
+  if (isLoading || !isFetched) return <FullPageLoader />;
+  if (isLoggedIn) return <Navigate to="/" replace={true} />;
+  return <LoginPage />;
+};
```
tcaiger, line 33:
```suggestion
const LoginRedirect = ({ children ) => {
  const { isLoggedIn, isLoading, isFetched } = useUser();
  if (isLoading || !isFetched) return <FullPageLoader />;
  if (isLoggedIn) return <Navigate to="/" replace={true} />;
  return children;
};
```
Can we either rename this or make it accept children. I was looking around for the login route in the Routes below and didn't realise it was up here.

packages/datatrak-web/src/Routes.tsx
```diff
@@ -4,14 +4,33 @@
...
+
+/**
+ * If the user is not logged in, redirect to the login page
+ */
+const LandingPageRedirect = () => {
+  const { isLoggedIn, isLoading, isFetched } = useUser();
+  if (isLoading || !isFetched) return <FullPageLoader />;
+  if (!isLoggedIn) return <Navigate to="/login" replace={true} />;
+  return <LandingPage />;
+};
```
tcaiger, line 23: This needs to be more general to handle redirecting to the login page for all "private routes". [Here ](https://github.com/remix-run/react-router/blob/dev/examples/auth/src/App.tsx)is the example from the docs on that. It might be cleaner to make a PrivateRoute component or something if we have to call it in a few places. That is what we did in AdminPanel in `admin-panel/src/authentician/PrivateRoute`. Let me know if you want to pair on it?

tcaiger (changes requested):
Looking really good. Thanks for doing all the extra work on the header. Just have a few small suggestions and one main one about the private view handling.

packages/ui-components/src/features/Auth/LoginForm.tsx
acdunham: Ah my bad, I forgot!

packages/datatrak-web/src/Routes.tsx
```diff
@@ -23,10 +39,29 @@ export const Routes = () => {
   return (
     <RouterRoutes>
       <Route path="/" element={<MainPageLayout />}>
-        <Route index element={<LandingPage />} />
-        <Route path="survey" element={<SurveyPageLayout />}>
-          <Route index element={<SurveyPage />} />
-          <Route path="questions" element={<SurveyQuestionsPage />} />
+        <Route
```
tcaiger, line 42: My understanding is that the user has to be logged in to see any of the pages including the landing page, survey page, admin page etc In which case the PrivateRoute will need to be applied more broadly






[4:21](https://beyondessential.slack.com/archives/D05041PQ30X/p1692850869565859)
In which case the PrivateRoute will need to be applied more broadly

packages/datatrak-web/src/views/LoginPage.tsx
```diff
@@ -0,0 +1,45 @@
...
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
+import styled from 'styled-components';
+import { SubmitHandler, useForm } from 'react-hook-form';
+import { LoginForm, EMAIL_VERIFICATION_STATUS } from '@tupaia/ui-components';
+import { useLogin } from '../api/mutations';
+import { ROUTES } from '../constants';
+
+const Wrapper = styled.div`
```
tcaiger, line 12: This is pretty minor but is an easy change. The Box with the LoginForm is below center because this centering doesn't account for the header. It doesn't have to be pixel perfect but it does look a bit low to me. I guess the same will apply for the Register Modal and possibly others so it might make sense to make a change in ui-components?!

acdunham: Would you mind if I changed as part of the register form, since I made it a reusable wrapper there?

tcaiger (approved):
Ka pai

PR #4900, opened 2023-08-22
RN-899: Make the CheckboxList component generic

rohan-bes:
Making this component generic allows us to override the base `ListItem` type when using the `CheckboxList`.

Just to address this comment thread: https://github.com/beyondessential/tupaia/pull/4848/files#r1296553278

### DIFF
--- a/packages/tupaia-web/src/api/mutations/useExportDashboard.tsx
+++ b/packages/tupaia-web/src/api/mutations/useExportDashboard.tsx
@@ -4,7 +4,7 @@
  */
 import { useMutation } from 'react-query';
 import { stringifyQuery } from '@tupaia/utils';
-import { post } from '../api';
+import { API_URL, post } from '../api';
 import { DashboardItem, DashboardName, EntityCode, ProjectCode } from '../../types';
 
 type ExportDashboardBody = {
@@ -24,7 +24,7 @@ export const useExportDashboard = ({ onSuccess }: { onSuccess?: (data: Blob) => 
         selectedDashboardItems: selectedDashboardItems?.join(','),
       });
       // Auth cookies are saved against this domain. Pass this to server, so that when it pretends to be us, it can do the same.
-      const cookieDomain = new URL(import.meta.env.REACT_APP_TUPAIA_WEB_API_URL).hostname;
+      const cookieDomain = new URL(API_URL).hostname;
       return post('pdf', {
         responseType: 'blob',
         data: {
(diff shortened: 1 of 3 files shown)


packages/ui-components/src/components/CheckboxList.tsx
```diff
@@ -89,47 +89,45 @@ export interface ListItemProps extends Record<string, unknown> {
...
+
+  const intersection = (a: ListItem[], b: ListItem[]) => {
+    return a.filter(
+      item => b.findIndex(i => i[i.valueKey || 'code'] === item[item.valueKey || 'code']) !== -1,
+    );
+  };
+
+  const union = (a: ListItem[], b: ListItem[]) => {
+    return [...a, ...not(b, a)];
+  };
```
rohan-bes, line 117: Kind of a bummer that we need to move these function in to the component. We could make them generic themselves, but I didn't quite feel like writing so much boilerplate 😅

packages/tupaia-web/src/api/mutations/useExportDashboard.tsx
```diff
@@ -24,7 +24,7 @@ export const useExportDashboard = ({ onSuccess }: { onSuccess?: (data: Blob) =>
         selectedDashboardItems: selectedDashboardItems?.join(','),
       });
       // Auth cookies are saved against this domain. Pass this to server, so that when it pretends to be us, it can do the same.
-      const cookieDomain = new URL(import.meta.env.REACT_APP_TUPAIA_WEB_API_URL).hostname;
```
rohan-bes, line 27: Got caught out by not having this env var set during testing. So referenced the value in `api.ts`

packages/tupaia-web/src/features/Dashboard/ExportDashboard.tsx
```diff
@@ -50,14 +50,18 @@ const ButtonGroup = styled.div`
 interface ExportDashboardProps {
   isOpen: boolean;
   onClose: () => void;
-  dashboardItems?: DashboardItem[];
+  dashboardItems: DashboardItem[];
```
acdunham, line 53: My only comment here would be when the dashboard items are loading they may be undefined

acdunham (approved):
Thanks for doing this! I had one question around if `dashboardItems` is undefined (eg not loaded yet) but other than that pre-approving

packages/tupaia-web/src/features/Dashboard/ExportDashboard.tsx
rohan-bes: I added a default `dashboardItems = []` in the component args, so it should always have a value now. Seems like we were doing something similar before with defaulting to an empty list if the `dashboardItems` were `undefined` and I felt this was a little cleaner 👍

PR #4902, opened 2023-08-22
WAITP-1367 DataTrak Server

EMcQ-BES:
### Changes:

- Add a new orchestration server package for `datatrak-server`
  - Contains base `login/logout` routes and `getUser`
  - Forwards the rest to central server
  - I've added .env files to lastpass (based on `tupaia-web-server`, most of the fields aren't actually being used here yet)
- Collapse the `mkdir` commands in the `Dockerfile`
  - Dockerfiles can contain a max of `127` layers (commands to layers are vaguely one to one)
  - Reducing the number of commands fixes the build


packages/datatrak-server/.env.example
```diff
@@ -0,0 +1,12 @@
+DB_NAME=
```
tcaiger, line 1: Just double checking that you got around to adding in the env files in last pass? :)

packages/datatrak-server/src/models/DataTrakSession.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { SessionType, SessionModel } from '@tupaia/server-boilerplate';
+
+export class DataTrakSessionType extends SessionType {
+  public static databaseType = 'datatrak_session';
+}
+
+export class DataTrakSessionModel extends SessionModel {
```
tcaiger, line 12: I guess we will need to add the hardcoded public permissions that tupaia users have in to here somewhere as well. I'll make a note on the ticket for that one.

packages/datatrak-server/.env.example
EMcQ-BES: image

tcaiger: 🤦

tcaiger (approved):

PR #4906, opened 2023-08-24
WAITP-1210: Allow attaching filter context without entity context

EMcQ-BES:
### Changes:

- Allow calculating and attaching the `filter` context to `req` in `entity-server` without a specific entity context
  - This is only useful for the `entitySearch` route at the moment because searches happen within a hierarchy context, but may still require a filter applied
  - Fixes the issue with `frontendExcludedTypes` not being correctly applied to entity search


rohan-bes (approved):
Looks good to me 👍 Nice stuff

PR #4911, opened 2023-08-24
WAITP-1341: Remove disaster project from DB

tcaiger:
### Changes:

The disaster project is to be removed in the re-write version of tupaia web. This migration removes all the disaster data and visualisations from the DB and deletes any disaster related schema. Disaster related models and code are already deleted.


rohan-bes (changes requested):
So much red! 😁 

This is great! I think I found a few more things that could get removed:
- `DisasterResponse` page in `@tupaia/admin-panel`
- `Disaster` and `DisasterEvent` from `@tupaia/database/src/types`
- Remove `DISASTER` from the map overlay EntityLevel type in `@tupaia/types`
- The `ActiveDistasters` component in `@tupaia/web-frontend` and the type reference in `@tupaia/types`
- `DisasterLayer` component in `@tupaia/web-frontend` and the `DEFAULT_DISASTER_COLOR` in `tupaia/ui-map-components`
- Remove `disaster` from the entity_type enum in the db
- Remove the `disaster_type` and `disaster_event_type` enums from the db 

There's also a heap of code in `@tupaia/web-frontend` referring to disasters (sagas, redux, components, etc.) I figure we're hoping to avoid explicitly removing all that here and then just not supporting it in `@tupaia/tupaia-web`? I can see the benefit of that 👍 just hope that we're not putting `@tupaia/web-frontend` in a broken state

tcaiger:
> 

Hey Rohan, I did the first 3 bullet points from your list but haven't done the others as:
- we are not bothering with refactoring web-frontend as we go since it will just be switched out once testing is passed
- I'm pretty sure the enum types are a data task

Overall this PR will break Disasters on web-frontend but not break tupaia otherwise or create any crashes. However, I can just leave it here until testing is almost passed if you would rather not have broken disasters in the main trunk?

PR #4912, opened 2023-08-24
RN-974: Bump meditrak android api to level 33

rohan-bes:

### DIFF
--- a/packages/meditrak-app/package.json
+++ b/packages/meditrak-app/package.json
@@ -77,7 +77,7 @@
     "react-native-side-menu": "^1.1.3",
     "react-native-svg": "^12.1.0",
     "react-native-vector-icons": "^7.1.0",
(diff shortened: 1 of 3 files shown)


packages/meditrak-app/package.json
```diff
@@ -77,7 +77,7 @@
     "react-native-side-menu": "^1.1.3",
     "react-native-svg": "^12.1.0",
     "react-native-vector-icons": "^7.1.0",
-    "react-native-webview": "^11.0.0",
+    "react-native-webview": "^11.23.1",
```
rohan-bes, line 80: Had to set `react-native-webview` to `11.23.1` instead of latest because of the following issue after upgrading: https://github.com/react-native-webview/react-native-webview/issues/2959

I guess we can cross that bridge when we do the full upgrade

chris-pollard (approved):

PR #4915, opened 2023-08-25
WAITP-1222 map overlay fixes

acdunham:
### Changes:
- Fixed loading state on map overlay selector
- Fixed padding on modals
- Add in logic for the default tile in map tile picker

### DIFF
--- a/packages/tupaia-web/src/api/queries/useMapOverlays.ts
+++ b/packages/tupaia-web/src/api/queries/useMapOverlays.ts
@@ -48,7 +48,7 @@ const mapOverlayByCode = (
  */
 export const useMapOverlays = (projectCode?: ProjectCode, entityCode?: EntityCode) => {
   const [urlSearchParams] = useSearchParams();
-  const { data, isLoading, error } = useQuery(
+  const { data, isLoading, error, isFetched } = useQuery(
     ['mapOverlays', projectCode, entityCode],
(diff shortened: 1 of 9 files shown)


packages/tupaia-web/src/api/queries/useMapOverlays.ts
```diff
@@ -48,7 +48,7 @@ const mapOverlayByCode = (
  */
 export const useMapOverlays = (projectCode?: ProjectCode, entityCode?: EntityCode) => {
   const [urlSearchParams] = useSearchParams();
-  const { data, isLoading, error } = useQuery(
+  const { data, isLoading, error, isFetched } = useQuery(
```
tcaiger, line 51: I was looking like something like this a while ago and couldn't find it so good find

packages/tupaia-web/src/features/Map/Map.tsx
```diff
@@ -95,6 +101,25 @@ const MapControlColumn = styled.div`
   }
 `;
 
+/**
+ * Utility function to determine whether tileSet should default to satellite
+ * or to osm, based on page load time. This will only run when determining the
+ * initial state of the map.
+ */
+const getAutoTileset = () => {
```
tcaiger, line 109: Thanks for adding this back in

tcaiger (approved):
Nice one!

PR #4916, opened 2023-08-25
WAITP-1231: Matrix fixes

acdunham:
### Changes:
- Fixed error being caught for some matrix description enlarged cells
- Added message on mobile about matrix not being good to view at this screen size
- Fixed drilldown dashboard items appearing in side panel

### DIFF
--- a/packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
+++ b/packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
@@ -10,6 +10,7 @@ import { Typography, Button } from '@material-ui/core';
 import GetAppIcon from '@material-ui/icons/GetApp';
 import { DEFAULT_BOUNDS } from '@tupaia/ui-map-components';
 import { ErrorBoundary } from '@tupaia/ui-components';
+import { MatrixConfig } from '@tupaia/types';
 import { MOBILE_BREAKPOINT } from '../../constants';
 import { ExpandButton } from './ExpandButton';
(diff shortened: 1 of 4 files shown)


packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -148,6 +149,19 @@ export const Dashboard = () => {
...
+      const { config } = item as {
+        config: MatrixConfig;
+      };
+      if (config?.drillDown && config?.drillDown?.itemCode === code) return true;
+      return false;
+    });
+  };
+
+  const visibleDashboards =
+    activeDashboard?.items?.filter(item => !getIsDrilldown(item.code)) || [];
```
tcaiger, line 163:
```suggestion
const visibleDashboards = activeDashboard?.items?.reduce((items , item) => {
  const isDrillDown = activeDashboard?.items?.some(({ config }) =>
    config?.drillDown && config?.drillDown?.itemCode === item.code
  );
  return isDrillDown ? items : [...items, item];
}, []);
```
this is an optional suggestion and pretty subjective change so it's up to you but I like using reduce for this since the activeDashboard items are used in the filter and the some

tcaiger: ps. it would also be nice to add a short comment

tcaiger (approved):
Looks good  - thanks for all the fixes

PR #4917, opened 2023-08-25
WAITP-1327: Geolocate question type

WDuncker:
### Changes:

- Added geolocate question type component.
- Created modal to house react-leaflet map and functionality for geolocation.
- Created Component for LatLng fields to be housed in.


packages/datatrak-web/src/features/SurveyQuestion.tsx
```diff
@@ -0,0 +1,104 @@
+// @ts-nocheck
```
tcaiger, line 1: This looks like a old test file which can be deleted. The actual SurveyQuestion component that is used in the SurveyScreen is in the features/Survey folder

packages/datatrak-web/src/features/Questions/GeolocateQuestion/MapModal/Map.tsx
```diff
@@ -0,0 +1,105 @@
...
+  // round coordinates to 2 decimal places before setting them
+  const onUpdateCoordinates = ({ lat, lng }: LatLngLiteral) => {
+    setCoordinates({
+      lat: parseFloat(lat.toFixed(2)),
+      lng: parseFloat(lng.toFixed(2)),
+    });
+  };
+  return (
+    <MapContainer
+      center={[-20.53, 162.05]}
```
tcaiger, line 90: I'm not sure where these coords come from but this should probably be based on the DEFAULT_BOUNDS variable which is in ui-map-coordinates or at least a constant

packages/datatrak-web/src/features/Questions/GeolocateQuestion/LatLongFields.tsx
```diff
@@ -0,0 +1,61 @@
...
+      margin-right: 1.125rem;
+    }
+  }
+`;
+
+export const LatLongFields = ({ setGeolocation, geolocation, name }: any) => {
+  const handleChange = (e: ChangeEvent<{}>, field: string) => {
+    setGeolocation({
+      ...geolocation,
+      [field]: parseFloat((e.target as HTMLInputElement).value),
```
tcaiger, line 24:
```suggestion
      [field]: parseFloat((e.target as HTMLInputElement).value),
      accuracy: 20 // hard code the accuracy to 20 for browser based location accuracy
```

According to the ticket description the data needs to match the data of the meditrak-app geo locate question which includes an accuracy value. The browser doesn't give us one but in the meditrak app code it has a comment which says that the target accuracy should be 20 so that seems like as good of default value as any.

tcaiger (changes requested):
Great work Wim and Alex. Looking really good and a lot of complex code really nicely organised. Just a couple of tweaks from me.

tcaiger (approved):
Good to go!

PR #4919, opened 2023-08-27
WAITP-1210 Sort entity searches with higher hierarchy entities first

EMcQ-BES:
### Changes:

- Put entities that are higher in the hierarchy order higher in the search results

### DIFF
--- a/packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
+++ b/packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
@@ -35,14 +35,23 @@ export class EntitySearchRoute extends Route<EntitySearchRequest> {
         [RAW]: {
           // Running this RAW is much faster for larger hierarchies
           // Put names that begin with the search string first
-          // Then sort alphabetically within the two groups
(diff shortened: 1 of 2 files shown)


packages/entity-server/src/routes/hierarchy/EntitySearchRoute.ts
```diff
@@ -35,14 +35,23 @@ export class EntitySearchRoute extends Route<EntitySearchRequest> {
...
             AND name ILIKE :searchFilter
-            ORDER BY STARTS_WITH(LOWER(name), :searchString) DESC, name ASC`,
+            ORDER BY
+              STARTS_WITH(LOWER(name), :searchString) DESC,
+              (
+                SELECT MAX(generational_distance) FROM ancestor_descendant_relation
+                WHERE entity_hierarchy_id = :hierarchyId
+                AND descendant_id = entity.id
+                GROUP BY descendant_id
+              ) ASC,
```
rohan-bes, line 53: Oh wow! I never knew you could do a select within an `ORDER BY` clause.

rohan-bes (approved):
Great stuff 👍

PR #4922, opened 2023-08-28
WAITP-1243 Fixup missing dashboard items

EMcQ-BES:
### Changes:

- Remove the spammy warning that I added
- Override default page size for dashboard requests
  - Since we fetch data for all dashboards on an entity simultaneously, some of our entities have upwards of 20 different dashboards
  - This results in more than 100 items/relations across those dashboards

We used a similar solution for `mapOverlays` that wound up not working because of the sheer volume of values in that case. Map overlay relations are in a hierarchy structure, which means for cases with a lot of layers the numbers blow up exponentially. Those requests were going above 400 records at points, which caused us issues. This is less likely to cause issues in dashboards since they're flat structured. Experimenting on my local, the larger sets of dashboards were giving ranges of 150-200. So I reckon this simpler solution will work in this case (or at the very least is worth testing with)

Separately I've put together [a ticket](https://linear.app/bes/issue/WAITP-1426/allow-central-server-to-support-larger-get-requests) for supporting larger requests in `central-server`

### DIFF
--- a/packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
+++ b/packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
@@ -82,9 +82,6 @@ export class ApiBuilder {
      */
     this.app.use((req: Request, res: Response, next: NextFunction) => {
       if (options.attachModels) {
-        winston.warn(
-          "Best practices say orchestrator servers shouldn't access the db directly, are you sure you need req.models?",
-        );
         req.models = this.models;
       }
       const context = {}; // context is shared between request and response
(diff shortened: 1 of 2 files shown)


packages/server-boilerplate/src/orchestrator/api/ApiBuilder.ts
```diff
@@ -82,9 +82,6 @@ export class ApiBuilder {
      */
     this.app.use((req: Request, res: Response, next: NextFunction) => {
       if (options.attachModels) {
-        winston.warn(
-          "Best practices say orchestrator servers shouldn't access the db directly, are you sure you need req.models?",
-        );
```
EMcQ-BES, line 87: I thought this would only run on start up but it actually runs every request which is _way_ too spammy

acdunham (approved):
Looks good to me

PR #4923, opened 2023-08-28
Waitp 1417 top progress bar

chaoxinnhin:
### Changes:

- Added Progress bar logic and design

### Screenshots:
image


packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  background: ${({ theme }) => theme.palette.primary.main};
+  color: white;
+  border-radius: 3px;
+  text-transform: none;
+  font-size: 0.9375rem;
+  height: 2.7851875rem;
+  padding: 0.8125rem 1.5625rem;
+`;
+
+const CancelButton = styled(Button)`
```
acdunham, line 31: This might be a silly question, but are buttons part of this? I thought it was just the bar?

packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  display: flex;
+  justify-content: flex-start;
+  align-items: flex-start;
+  height: 0.75rem;
+  width: 100%;
+`;
+
+const ProgressButton = styled(Button)`
+  background: ${({ theme }) => theme.palette.primary.main};
+  color: white;
```
acdunham, line 23: What you can do with buttons is, instead of manually setting background and color, give the button a prop of `color='primary'`, and then Mui will set it to be the correct primary colours

packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  border-radius: 3px;
+  text-transform: none;
+  font-size: 0.9375rem;
```
acdunham, line 26: You can round these up if you like, no need to be super duper exact

packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  height: 2.7851875rem;
+  padding: 0.8125rem 1.5625rem;
+`;
+
+const CancelButton = styled(Button)`
+  background-color: white;
```
acdunham, line 32: Same with the colours here - this can be set by giving the button a prop of `variant='outlined'` for the outlined button

packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  currentSurveyQuestion: number;
+  totalNumberOfSurveyQuestions: number;
+}
+
+export const TopProgressBar = ({
+  currentSurveyQuestion,
+  totalNumberOfSurveyQuestions,
+}: ProgressPercentage) => {
+  const fraction = (currentSurveyQuestion / totalNumberOfSurveyQuestions) * 100;
+  console.log(fraction);
```
acdunham, line 56:
```suggestion

```

packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [progress, setProgress] = useState(fraction);
```
acdunham, line 57: I think this bar should be 'dumb' and not need to set the progress values. So there wouldn't be any buttons, just a bar with a dynamically filled amount based on the values that come in (currentSurveyQuestion and totalNumberOfSurveyQuestions)

packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -0,0 +1,92 @@
...
+      setProgress(0);
+    } else {
+      setProgress(progress - fraction);
+    }
+  };
+
+  return (
+    <>
+      <Wrapper>
+        <ProgressBar style={{ width: `${progress}%` }} />
```
acdunham, line 82: Instead of setting this with `style`, you can give `ProgressBar` a prop of `$progress={progress}` and then in the styled component set `width` like:
`width: ${({$progress}) => `${progress}%`};

packages/datatrak-web/src/theme/theme.ts
```diff
@@ -1,21 +1,28 @@
...
-      default: '#F9F9F9', // Off white background
-      paper: '#ffffff', // White background
-    },
-    text: {
-      primary: '#333333', // dark text color
-      secondary: '#4A79EA', // purple text for secondary buttons etc
+  },
+  {
+    progressBar: {
+      main: '#004167',
```
acdunham, line 25: Nice use of the theme

packages/datatrak-web/src/views/SurveyPage.tsx
```diff
@@ -3,6 +3,7 @@
  *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 import React from 'react';
+import { TopProgressBar } from '../components/TopProgressBar';
```
acdunham, line 6:
```suggestion

```

packages/datatrak-web/src/views/LandingPage.tsx
```diff
@@ -5,13 +5,15 @@
 
 import React from 'react';
 import { ButtonLink, PageContainer } from '../components';
+import { TopProgressBar } from '../components/TopProgressBar';
```
acdunham, line 8: Should export `TopProgressBar` from `components/index.ts` so that we can import this like `import { TopProgressBar } from '../components'`

acdunham (changes requested):
This is looking good. A few changes from me around the way we structure the component

packages/datatrak-web/src/components/TopProgressBar.tsx
chaoxinnhin: The buttons are there for testing purposes and styling. As Datatrak develops we can move the buttons to the appropriate files

packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -6,42 +6,50 @@ import ArrowBackIosRoundedIcon from '@material-ui/icons/ArrowBackIosRounded';
...
   height: 0.75rem;
-  width: 100%;
+  width: ${({ $progress }) => `${$progress}`}%;
 `;
 
-const ProgressButton = styled(Button)`
+/*
+Styling of buttons as the project progresses.
+
+const ProgressButton = styled(Button).attrs({
```
acdunham, line 26: Probably just need to remove these now they aren't used

acdunham (approved):
Looks really nice and clean now, good work. Just need to remove those unused variables which is likely breaking the build

acdunham (changes requested):
Thanks for doing those changes! Now that we have merged in the setup of the survey screens, can you just plumb in the progress bar to be on the survey screen?

packages/datatrak-web/src/views/SurveyPage.tsx
```diff
@@ -40,6 +48,10 @@ export const SurveyPage = () => {
     <SurveyContext>
       <Container>
         <Toolbar />
+        <TopProgressBar
+          currentSurveyQuestion={screenNumber}
+          totalNumberOfSurveyQuestions={numberOfScreens}
+        />
```
acdunham, line 54: If you add this to `components/SurveyScreen` instead of this file, you can use `numberOfScreens` from the `useSurveyForm` hook

acdunham (changes requested):

acdunham (approved):
Thanks for making those changes!

PR #4924, opened 2023-08-28
WAITP-1215 Filter out overlays with hideFromMenu set

EMcQ-BES:
### Changes:

- Filter overlays before the relations fetch
  - This way no relations will match and it won't get attached to the final return object


acdunham (approved):
Thanks for doing this!

PR #4927, opened 2023-08-29
hotfix: Fix error logging when importing surveys

rohan-bes:
### Issue: https://beyondessential.slack.com/archives/C01MAA3NKM1/p1693195752875529

Couple issues here to address:
- We always set a value for `generateQrCode` and `allowScanQrCode` when it'd be better for us to default them instead
- There are functions that throw errors that extend from `LoggedError` in the frontend. Figured it'd be easier just to just check if `winston` is available in that exception.


rohan-bes:
Build failed due to performance issues with vite

chris-pollard (approved):
LG

PR #4929, opened 2023-08-30
No-issue: fix lesmis build error

tcaiger:
### Changes:
Fixes this issue on the dev builds.
```
er-compatibility for more details.
\Tue Aug 29 00:43:59 UTC 2023 \✓ 11625 modules transformed.
\Tue Aug 29 00:44:00 UTC 2023 \Export "ExportModal" of module "../../node_modules/@tupaia/admin-panel/lib/importExport/ExportModal.js" was reexported through module "../../node_modules/@tupaia/admin-panel/lib/importExport/index.js" while both modules are dependencies of each otherand will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
\Tue Aug 29 00:44:00 UTC 2023 \Either change the import in "src/views/AdminPanel/components/getSurveyResponsesExportModal.jsx" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.

### DIFF
--- a/packages/lesmis/src/routes/PageRoutes.jsx
+++ b/packages/lesmis/src/routes/PageRoutes.jsx
@@ -3,10 +3,10 @@
  * Copyright (c) 2017 - 2020 Beyond Essential Systems Pty Ltd
  *
  */
-import React, { lazy, Suspense } from 'react';
+import React from 'react';
 import { Switch, Route, useRouteMatch } from 'react-router-dom';
 import { AdminPanelDataProviders } from '@tupaia/admin-panel';
-import { NavBar, Footer, FullPageLoader } from '../components';
+import { NavBar, Footer } from '../components';
 import { HomeView } from '../views/HomeView';
 import { ProfileView } from '../views/ProfileView';
 import { PageView, TwoColumnPageView } from '../views/PageView';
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/lesmis/src/routes/PageRoutes.jsx [hunk 1]
```diff
@@ -19,8 +19,8 @@ import { VerifyEmailView } from '../views/VerifyEmailView';
 import { ABOUT_PAGE, FQS_PAGE, CONTACT_PAGE } from '../constants';
 import { ExportView, PDF_DOWNLOAD_VIEW } from '../views/ExportView';
 import { getAdminApiUrl } from '../utils/getAdminApiUrl';
+import AdminPanel from './AdminPanelApp';
 
-const AdminPanel = lazy(() => import('./AdminPanelApp'));
```
tcaiger, line 23: Changing this from a lazy import to a standard es module import will increase the bundle size slightly on the main js bundle but will fix the circular dependency issue so is a trade off worth doing in my opinion. Also we have dramatically reduced the main bundle size already by using chunking in the js build so overall it's still better than it was.

packages/lesmis/src/routes/PageRoutes.jsx
```diff
@@ -19,8 +19,8 @@ import { VerifyEmailView } from '../views/VerifyEmailView';
 import { ABOUT_PAGE, FQS_PAGE, CONTACT_PAGE } from '../constants';
 import { ExportView, PDF_DOWNLOAD_VIEW } from '../views/ExportView';
 import { getAdminApiUrl } from '../utils/getAdminApiUrl';
+import AdminPanel from './AdminPanelApp';
```
rohan-bes, line 22: Never quite knew why this was a lazy import, did you?

rohan-bes (approved):
Thanks @tcaiger! 

Can you also just remove the import of `lazy`, but pre-approving 👍

PR #4930, opened 2023-08-31
WAITP-1345: core question types

acdunham:
### Changes:
- Created core question type components: Text, Radio, Binary, Date, DateTime, Instruction, Checkbox
- Updated types for request responses
- Added question numbers next to the question name, as per the design
- Updated the endpoint to parse any stringified JSON

---

### Screenshots:
Binary
Screenshot 2023-08-31 at 1 12 53 pm

Number
Screenshot 2023-08-31 at 1 13 10 pm

Radio
Screenshot 2023-08-31 at 2 59 49 pm

Date
Screenshot 2023-08-31 at 3 00 09 pm

Instruction
Screenshot 2023-08-31 at 3 01 17 pm

Checkbox
Screenshot 2023-08-31 at 3 03 17 pm

### DIFF
--- a/packages/datatrak-web/src/features/Questions/CheckboxQuestion.tsx
+++ b/packages/datatrak-web/src/features/Questions/CheckboxQuestion.tsx
@@ -19,7 +19,7 @@ export const CheckboxQuestion = ({
   id,
   label,
   name,
-  controllerProps: { value, onChange },
+  controllerProps: { value, onChange, ref },
 }: SurveyQuestionInputProps) => {
   return (
     <Checkbox
@@ -29,6 +29,7 @@ export const CheckboxQuestion = ({
       name={name!}
       checked={value === 'Yes'}
       onChange={e => onChange(e.target.checked ? 'Yes' : 'No')}
+      inputRef={ref}
     />
   );
 };
--- a/packages/datatrak-web/src/features/Questions/CheckboxQuestion.tsx
+++ b/packages/datatrak-web/src/features/Questions/CheckboxQuestion.tsx
@@ -4,8 +4,7 @@
  */
 
 import React from 'react';
-import styled from 'styled-components';
-import { useFormContext, Controller } from 'react-hook-form';
+import styled from 'styled-components'; 
 import { Checkbox as BaseCheckbox } from '@tupaia/ui-components';
 import { SurveyQuestionInputProps } from '../../types';
 
@@ -16,26 +15,21 @@ const Checkbox = styled(BaseCheckbox)`
     margin-right: 0.5rem;
   }
 `;
-export const CheckboxQuestion = ({ id, label, name }: SurveyQuestionInputProps) => {
-  const { control } = useFormContext();
-  // this is a controller component so that we can make the value save as 'No' when unchecking the checkbox
+export const CheckboxQuestion = ({
+  id,
+  label,
+  name,
+  ...controllerProps
+}: SurveyQuestionInputProps) => { 
   return (
-    <Controller
-      name={name}
-      control={control}
-      render={renderProps => {
-        return (
-          <Checkbox
-            value="Yes"
-            label={label}
-            id={id}
-            inputRef={renderProps.ref}
-            name={name}
-            checked={renderProps.value === 'Yes'}
-            onChange={e => renderProps.onChange(e.target.checked ? 'Yes' : 'No')}
-          />
-        );
-      }}
+    <Checkbox
+      {...controllerProps}
+      value="Yes"
+      label={label}
+      id={id}
+      name={name!}
+      checked={controllerProps.value === 'Yes'}
+      onChange={e => controllerProps.onChange(e.target.checked ? 'Yes' : 'No')}
     />
   );
 };
--- a/packages/datatrak-web/src/features/Questions/CheckboxQuestion.tsx
+++ b/packages/datatrak-web/src/features/Questions/CheckboxQuestion.tsx
@@ -19,17 +19,16 @@ export const CheckboxQuestion = ({
   id,
   label,
   name,
-  controllerProps,
+  controllerProps: { value, onChange },
 }: SurveyQuestionInputProps) => {
   return (
     <Checkbox
-      {...controllerProps}
       value="Yes"
       label={label}
       id={id}
       name={name!}
-      checked={controllerProps.value === 'Yes'}
-      onChange={e => controllerProps.onChange(e.target.checked ? 'Yes' : 'No')}
+      checked={value === 'Yes'}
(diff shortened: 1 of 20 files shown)


packages/datatrak-web/src/features/Survey/SurveyQuestion.tsx
```diff
@@ -24,51 +35,44 @@ const Placeholder = ({ name, type, id }) => {
...
-  return <FieldComponent inputRef={props.register()} {...props} />;
+  // Use a Controller so that the fields that require change handlers, values, etc work with react-hook-form, which is uncontrolled by default
+  return (
+    <Controller
+      name={name!}
+      control={control}
+      render={renderProps => (
+        <FieldComponent {...props} controllerProps={renderProps} name={name} type={type} />
+      )}
+      inputRef={register()}
```
tcaiger, line 75:
```suggestion

```
So I guess if we are making all the fields controlled inputs there is no need for the manual register stuff, right?

packages/datatrak-web/src/features/Survey/SurveyQuestion.tsx
```diff
@@ -24,51 +35,44 @@ const Placeholder = ({ name, type, id }) => {
...
 
-  return <FieldComponent inputRef={props.register()} {...props} />;
+  // Use a Controller so that the fields that require change handlers, values, etc work with react-hook-form, which is uncontrolled by default
+  return (
+    <Controller
+      name={name!}
+      control={control}
+      render={renderProps => (
+        <FieldComponent {...props} controllerProps={renderProps} name={name} type={type} />
+      )}
```
tcaiger, line 74:
```suggestion
      render={renderProps => {
       const onChange = e => {
          setFormData({ ...formData, [name]: e.target.value });
          renderProps.onChange(e);
        };
        return <FieldComponent {...props} controllerProps={{ ...renderProps, onChange }} name={name} type={type} />
      }}
```
Since we will need to support navigating between questions, we will also have to hook up the on change handler to keep the form state up to date - something like this. We could do it later but might as well include it now.

packages/datatrak-web/src/features/Questions/TextQuestion.tsx
```diff
@@ -0,0 +1,51 @@
...
+  type,
+  controllerProps,
+}: SurveyQuestionInputProps) => {
+  return (
+    <Label
+      label={label}
+      name={name!}
+      labelPlacement={'top'}
+      inputRef={inputRef}
+      {...controllerProps}
```
tcaiger, line 39: It's probably a pretty annoying request but can we explicitly add the controller props. It's something that we have discussed previously in dev chat and there was a consensus that it's more readable. There should only be four (onChange, onBlur, value, name, ref)

packages/datatrak-web/src/features/Questions/CheckboxQuestion.tsx
```diff
@@ -0,0 +1,35 @@
...
+`;
+export const CheckboxQuestion = ({
+  id,
+  label,
+  name,
+  controllerProps,
+}: SurveyQuestionInputProps) => {
+  return (
+    <Checkbox
+      {...controllerProps}
```
tcaiger, line 26: This is a good example of how explicitly writing out the controller props will make it simpler since 2 of the 4 props are overwritten anyway (name and onChange).

packages/datatrak-web/src/features/Questions/TextQuestion.tsx
tcaiger: Also I think it would be ideal to set the form state using the onBlur event for text inputs. eg. onBlur={onChange} so that we are not setting state unnecessarily on each keypress.

packages/datatrak-web/src/features/Survey/SurveyQuestion.tsx
```diff
@@ -6,6 +6,17 @@
 import React from 'react';
 import styled from 'styled-components';
 import { TextField } from '@tupaia/ui-components';
```
tcaiger, line 8:
```suggestion

```

packages/datatrak-web/src/features/Survey/SurveyQuestion.tsx
```diff
@@ -24,51 +35,44 @@ const Placeholder = ({ name, type, id }) => {
...
   }
 
-  return <FieldComponent inputRef={props.register()} {...props} />;
+  // Use a Controller so that the fields that require change handlers, values, etc work with react-hook-form, which is uncontrolled by default
+  return (
+    <Controller
+      name={name!}
+      control={control}
+      render={renderProps => (
+        <FieldComponent {...props} controllerProps={renderProps} name={name} type={type} />
```
tcaiger, line 73: I was thinking about whether it would be better to just spread the renderProps onto the component but I think it's quite nice how your got them here in the controllerProps as it is a nice way to keep track of them as the react-hook-form library props 👍

packages/datatrak-web/src/features/Survey/SurveyQuestion.tsx
```diff
@@ -24,51 +35,44 @@ const Placeholder = ({ name, type, id }) => {
...
-  register?: any;
-  label?: string;
-  code?: string;
-  text?: string;
-  options?: any;
-  config?: any;
-}
-export const SurveyQuestion = (props: SurveyQuestionProps) => {
-  const FieldComponent = QUESTION_TYPES[props.type];
+export const SurveyQuestion = ({ type, name, ...props }: SurveyQuestionFieldProps) => {
```
tcaiger, line 59: I'm on the fence about whether it's worth typing out these props as well but since this component is quite simple and they are more directly passed on then it's probably all good to just spread them as is.

tcaiger (changes requested):
Looks good. The Questions are all really nice, I just have a couple of requested changes about how the Questions are hooked up to the survey screen.

packages/datatrak-web/src/features/Questions/TextQuestion.tsx
acdunham: I can definitely explicitly add them. As for making it onBlur, because it's a controlled input, if we make it on blur, then the value the user types won't show up in the field until it's unfocussed, which I guess we don't want, right?

packages/datatrak-web/src/features/Survey/SurveyQuestion.tsx
acdunham: I'm pretty sure we don't need to add an onChange handler to set the form data, since they're now controlled inputs the value gets passed through with the controller and persists between screens

tcaiger (approved):

PR #4933, opened 2023-08-31
Fix error when answer is null

acdunham:
### Changes:
- Handle when values in arithmetic question are `null`, usually caused by selecting and unselecting an option in a radio question. If value is null, handle as if is undefined.
- Handle when value is set as a string but valueTranslation is not available, to prevent the app crashing


tcaiger (approved):
Classic javascript fix :)

acdunham:
@tcaiger can I trouble you to re-review this one? Have fixed a similar, related bug as part of this ticket

packages/meditrak-app/app/assessment/selectors.js
```diff
@@ -127,10 +127,12 @@ export const getArithmeticResult = (state, arithmeticQuestionId) => {
     if (valueTranslation[questionId] && valueTranslation[questionId][answer] !== undefined) {
       return valueTranslation[questionId][answer]; // return translated answer if there's any
     }
-    if (answer !== undefined) {
-      return answer; // return raw answer
+    // return raw answer if it's a number, else 0 (e.g. if no valueTranslation provided for the question and this specific answer when answer is non-numeric)
+    if (answer !== undefined && answer !== null) {
+      return isNaN(answer) ? 0 : answer; // return raw answer
```
EMcQ-BES, line 132: Do we protect from empty string somewhere?

EMcQ-BES (approved):
It'll work better for sure. What's the deal with empty strings though?

packages/meditrak-app/app/assessment/selectors.js
acdunham: Such a javascript issue to have 🙃 I have added a fix for that, thanks!

EMcQ-BES: Cheers, looks good 👍

PR #4938, opened 2023-09-01
WAITP-1224: Map overlay selector fixes

tcaiger:
### Changes:

- Enable reset to default button
- Tweak styles for disabled state, opening and closing, hover and action buttons
- Remove auto focus on modal close

---

### Screenshots:
image

image


packages/tupaia-web/src/features/Map/MapOverlaySelector/MapOverlayDatePicker.tsx
```diff
@@ -50,6 +51,7 @@ export const MapOverlayDatePicker = () => {
           maxDate={maxEndDate}
           granularity={periodGranularity}
           onSetDates={setDates}
+          onResetDate={onResetDate}
```
acdunham, line 54: Good catch, we must have forgotten that one

acdunham (approved):
Thanks!

PR #4941, opened 2023-09-03
WAITP-1221: Polygon overlay fixes

acdunham:
### Changes:
- Fix issue of zoom level being incorrect on selection of an entity
- Fix issue with polygon shades being wrong
- Fix issue with country outlines not showing on map
- Fix issue where child entities were not showing while map overlay report data was loading


tcaiger (approved):
Nice work and thanks for all the comments!

PR #4949, opened 2023-09-04
WAITP-1428: datatrak design updates

acdunham:
### Changes:
- Added single survey endpoint route for displaying survey name
- Updated backgrounds for views
- Added Button component for ability to add a tooltip to buttons
- Created SurveyToolbar component so that the survey form info can be accessed from the context
- Updated theme colours to match design

---

### Screenshots:
Screenshot 2023-09-04 at 5 07 42 pm 

Screenshot 2023-09-04 at 5 09 02 pm

Screenshot 2023-09-04 at 5 09 14 pm

Screenshot 2023-09-04 at 5 09 33 pm
 
Screenshot 2023-09-04 at 5 17 47 pm

Screenshot 2023-09-04 at 5 18 02 pm

Screenshot 2023-09-04 at 5 18 11 pm

Screenshot 2023-09-04 at 5 18 20 pm

Screenshot 2023-09-04 at 5 18 30 pm

Screenshot 2023-09-04 at 5 18 37 pm


packages/datatrak-web-server/src/app/createApp.ts
```diff
@@ -33,6 +35,7 @@ export function createApp() {
     .attachApiClientToContext(authHandlerProvider)
     .get<UserRequest>('getUser', handleWith(UserRoute))
     .get<SurveysRequest>('surveys', handleWith(SurveysRoute))
+    .get<SurveyRequest>('surveys/:surveyCode', handleWith(SurveyRoute))
```
tcaiger, line 38: Just in case you didn't know about it, there is a middleware at the bottom of this file which forwards all routes on to the central server if a route is not matched and it seems like the single survey could just be fetched from there. I think it's ok to keep the custom SurveyRoute that you've written since you've already done the work but it's something to keep in mind for future as it can save a bit of boilerplate code.

packages/datatrak-web-server/src/routes/SurveyRoute.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class SurveyRoute extends Route<SurveyRequest> {
+  public async buildResponse() {
+    const {
+      ctx,
+      query = {},
+      params: { surveyCode },
+    } = this.req;
+    const { fields } = query;
+    const surveys = await ctx.services.central.fetchResources('surveys', {
```
tcaiger, line 26: Any reason that you went for the filter over the url param? (eg. /surveys/:recordId)

packages/datatrak-web-server/src/app/createApp.ts
acdunham: I initially did use that, but because we want to fetch the survey by code, not id, it wasn't returning any results

packages/ui-components/src/components/Button.tsx
```diff
@@ -45,16 +45,27 @@ export interface ButtonProps extends MuiButtonProps {
...
-  isLoading = false,
-  loadingText = 'Loading',
-  disabled = false,
-  ...props
-}: OverrideableComponentProps<ButtonProps>) => (
-  <StyledButton variant="contained" color="primary" {...props} disabled={isLoading || disabled}>
-    {isLoading ? `${loadingText}...` : children}
-  </StyledButton>
+export const Button = forwardRef(
+  (props: OverrideableComponentProps<ButtonProps>, ref: Ref<any>) => {
```
tcaiger, line 49: thanks for fixing this up. That console error has been staring at me for a while now

packages/datatrak-web/src/layout/AuthPageLayout.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { PageContainer } from '../components';
+
+export const Background = styled(BaseBackground)`
+  h2 {
+    font-weight: ${({ theme }) => theme.typography.fontWeightBold};
+  }
+  h3 {
+    margin-top: 0.32rem;
+  }
+  .MuiFormControl-root {
```
tcaiger, line 18: I feel like these tweaks to the styles might belong in the ui-components forms. That would mean that they will also apply to tupaia-web but I think the plan is to keep datatrak and tupaia-web the same. Can you please check with Flic?

packages/datatrak-web/src/Routes.tsx
```diff
@@ -69,7 +69,7 @@ export const Routes = () => {
           <Route index element={<LandingPage />} />
         </Route>
         {/** Any views that should have the background image should go in here */}
-        <Route path="/" element={<BackgroundPageLayout />}>
+        <Route path="/" element={<AuthPageLayout />}>
           {/** Any public centred views should go in here */}
           <Route path="/" element={<CentredLayout />}>
```
tcaiger, line 74: The centred layout is not laying out quite right on shorter screens such as laptop. The paper should be vertically centred with padding around it and there should be a scrollbar inside the paper.

tcaiger (changes requested):
Looking very slick. Just a couple of minor suggestions, thanks

packages/datatrak-web/src/Routes.tsx
tcaiger: This still isn't quite right unfortunately. It has padding now but it still doesn't fit inside the screen (mine is 1200px high). I think that some where there needs to be a height: calc(100% - HEADER_HEIGHT) or something to make it fit.

Also it is very minor but in the design the Paper is flat with a border while in the app it has a box shadow (eg. Paper component needs variant=outlined prop)

image

acdunham: Sorry, my bad, fix has been pushed

tcaiger (approved):

PR #4953, opened 2023-09-04
Faster build step

IgorNadj:


.github/workflows/ci.yaml
```diff
@@ -81,6 +81,18 @@ jobs:
       - run: yarn build:internal-dependencies
       - run: ./packages/devops/scripts/ci/validateTypesAndDbSchemaInSync.sh
 
+  validate_lockfile:
```
IgorNadj, line 84: Yarn recommends to validate the lockfile is up to date explicitly in a separate validation CI step

.github/workflows/ci.yaml
```diff
@@ -11,6 +11,7 @@ concurrency:
 env:
   NODE_OPTIONS: '--max-old-space-size=6144'
   NODE_ENV: test
+  PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
```
IgorNadj, line 14: Seems to save 20 seconds

.github/workflows/ci.yaml
```diff
@@ -81,9 +82,60 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14.19.3
+          cache: yarn
+      - run: yarn set version 3.2.1
+      - run: SKIP_BUILD_INTERNAL_DEPENDENCIES=true yarn install --immutable
+
+  package_list:
```
IgorNadj, line 97: Not a core part of this PR but a nice little bit of automation so we don't have to maintain the package list.

.github/workflows/ci.yaml
```diff
@@ -81,9 +82,60 @@ jobs:
...
+    name: Build
+    needs:
+      package_list
+    runs-on: ubuntu-latest
+    env:
+      CI: false # react-scripts build treats warnings as errors if this is true
+
+    strategy:
+      fail-fast: false
+      matrix:
```
IgorNadj, line 117: The build step now builds each package in isolation leading to a speedup.

This is also needed for containerisation where each server might be its own container, we need to build that package in isolation.

This also improves our ideal of having each package declare its dependencies, for maintainability. Because only the single package and its deps are installed with `yarn workspaces focus`, it cannot rely on a module being present because a completely unrelated package installed it.

PR #4958, opened 2023-09-06
WAITP-1346: Survey contents menu

acdunham:
### Changes:
- Changed `SurveyContext` to use a reducer instead of state, now that we have more than 1 action
- Created survey contents menu and added to view

Note: I haven't included mobile styling here, as not sure of design yet
---

### Screenshots:
Closed  
Screenshot 2023-09-06 at 12 20 58 pm 

Open
Screenshot 2023-09-06 at 12 20 46 pm

### DIFF
--- a/packages/datatrak-web/src/components/Button.tsx
+++ b/packages/datatrak-web/src/components/Button.tsx
@@ -2,30 +2,54 @@
  * Tupaia
  *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
-import React, { ReactNode, Fragment } from 'react';
+import React, { ReactElement, ReactNode } from 'react';
 import { To, Link as RouterLink } from 'react-router-dom';
-import { Tooltip, Button as UIButton } from '@tupaia/ui-components';
+import { Tooltip as BaseTooltip, Button as UIButton } from '@tupaia/ui-components';
 import styled from 'styled-components';
 
+const TOOLTIP_COLOR = '#002d47';
 const StyledButton = styled(UIButton)`
   &.Mui-disabled {
     pointer-events: auto; // this is to allow the hover effect of a tooltip to work
   }
 `;
 
+const Tooltip = styled(BaseTooltip).attrs({
+  disablePortal: true,
+})`
+  .MuiTooltip-tooltip {
+    background-color: ${TOOLTIP_COLOR};
+    .MuiTooltip-arrow {
+      color: ${TOOLTIP_COLOR};
+    }
+  }
+`;
 interface ButtonProps extends Record<string, any> {
   tooltip?: string;
   children?: ReactNode;
   to?: To;
 }
-export const Button = ({ tooltip, children, to, ...restOfProps }: ButtonProps) => {
-  const Wrapper = tooltip ? Tooltip : Fragment;
 
+const ButtonWrapper = ({
+  children,
+  tooltip,
+}: {
+  children: ReactElement<any, any>;
+  tooltip?: ButtonProps['tooltip'];
+}) => {
+  if (!tooltip) return children;
   return (
-    <Wrapper title={tooltip!} arrow>
+    <Tooltip title={tooltip} arrow>
+      {children}
+    </Tooltip>
+  );
+};
+export const Button = ({ tooltip, children, to, ...restOfProps }: ButtonProps) => {
+  return (
+    <ButtonWrapper tooltip={tooltip}>
       <StyledButton component={to ? RouterLink : undefined} to={to} {...restOfProps}>
         {children}
       </StyledButton>
-    </Wrapper>
+    </ButtonWrapper>
   );
 };
--- a/packages/datatrak-web/src/features/Survey/SurveyContext.tsx
+++ b/packages/datatrak-web/src/features/Survey/SurveyContext.tsx
@@ -3,7 +3,7 @@
  *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
-import React, { Dispatch, createContext, useContext, useReducer, useState } from 'react';
+import React, { Dispatch, createContext, useContext, useReducer } from 'react';
 import { useParams } from 'react-router-dom';
 import { sortBy } from 'lodash';
 import { SurveyParams, SurveyScreenComponent } from '../../types';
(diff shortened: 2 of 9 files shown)


packages/datatrak-web/src/components/Button.tsx
```diff
@@ -2,30 +2,54 @@
...
+  children: ReactElement<any, any>;
+  tooltip?: ButtonProps['tooltip'];
+}) => {
+  if (!tooltip) return children;
+  return (
+    <Tooltip title={tooltip} arrow>
+      {children}
+    </Tooltip>
+  );
+};
```
acdunham, line 46: This was to stop an error in the console about `Fragment` not being allowed to have a `title` prop

packages/datatrak-web/src/features/Survey/SurveySideMenu/SurveySideMenu.tsx
```diff
@@ -0,0 +1,91 @@
...
+  width: 2rem;
+`;
+
+const SurveyScreenTitle = styled.span`
+  width: 100%;
+  font-weight: ${({ theme }) => theme.typography.fontWeightRegular};
+`;
+
+export const SurveySideMenu = () => {
+  const { sideMenuOpen, toggleSideMenu, surveyScreenComponents, screenNumber } = useSurveyForm();
```
tcaiger, line 71: I don't think the side menu open state needs to be in the context since the state could just be placed here and shared between the Drawer and  SideMenuButton. Would you mind taking it out of the context?

packages/datatrak-web/src/features/Survey/SurveyContext.tsx
```diff
@@ -21,20 +21,66 @@ const convertNumberToLetter = (number: number) => {
...
-const SurveyFormContext = createContext({} as SurveyFormContextType);
+const defaultContext = {
+  formData: {},
+  activeScreen: [],
+  isLast: false,
+  numberOfScreens: 0,
+  screenNumber: 1,
+  screenHeader: '',
+  displayQuestions: [],
+  sideMenuOpen: false,
```
tcaiger, line 43: I would have expected the sideMenu to default to open? It seems that way in the design

acdunham: When we went through the design with Flic the other day she said it defaulted to closed

packages/datatrak-web/src/features/Survey/SurveyContext.tsx
```diff
@@ -21,20 +21,66 @@ const convertNumberToLetter = (number: number) => {
...
+export const surveyReducer = (
+  state: SurveyFormContextType,
+  action: SurveyFormAction,
+): SurveyFormContextType => {
+  switch (action.type) {
+    case ACTION_TYPES.SET_FORM_DATA:
+      return {
+        ...state,
+        formData: action.payload as Record<string, any>,
+      };
```
tcaiger, line 70:
```suggestion
      return {
        ...state,
        formData: { ...state.formData, action.payload as Record<string, any> },
      };
```
What do you think about doing it like this so that when we call `setFormData` we just pass in the new form data and don't have to re-create the whole formData object everytime?

acdunham: Sure, can do

packages/datatrak-web/src/features/Survey/SurveyScreen.tsx
```diff
@@ -144,77 +153,79 @@ export const SurveyScreen = () => {
...
-                  Back
-                </Button>
-                <ButtonGroup>
-                  <Button variant="outlined" to={ROUTES.SURVEY_SELECT}>
-                    Cancel
+      <Wrapper>
+        <SurveySideMenu />
+        <Container $sideMenuOpen={sideMenuOpen}>
+          <Paper>
+            <FormProvider {...formContext}>
```
tcaiger, line 160: I think this is my bad from when I set this up but the FormProvider needs to go outside of the SurveySideMenu so that the Side menu can have access to form state

packages/datatrak-web/src/features/Survey/SurveyContext.tsx
acdunham: I actually initially planned to make it pass in a key and a value, but I decided against it for the occasions where we want to reset multiple things

packages/datatrak-web/src/features/Survey/SurveySideMenu/SurveySideMenu.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <>
+      <SideMenuButton />
+      <Drawer open={sideMenuOpen} onClose={toggleSideMenu}>
+        <SurveyMenuContent>
+          {Object.entries(surveyScreenComponents!).map(([key, screen]) => (
+            <SurveyMenuItem
+              key={screen[0].questionId}
+              to={`../${key}`}
```
tcaiger, line 80: The side menu needs to save the form state each time a user navigates to a new screen so this will have to be converted to a button with a onclick handler which gets the values from the form screen (using { getValues } = useFormContext()) and sets them in the overall form (using { setFormData } = useSurveyForm())

packages/datatrak-web/src/features/Survey/SurveyContext.tsx
tcaiger: ah ok - in that case ignore what I said :)

packages/datatrak-web/src/features/Survey/SurveyContext.tsx
tcaiger: Yes good decision I think we will be passing in a screen of values at a time

tcaiger (changes requested):
Looking really nice, just keen to include saving the form values on nav

packages/datatrak-web/src/features/Survey/SurveyContext.tsx
tcaiger: ps. this isn't really related but I noticed while testing that on the radio buttons, clicking on the text doesn't trigger the onClick event? If it's a quick fix, would you mind fixing that up to?

packages/datatrak-web/src/features/Survey/SurveySideMenu/SurveySideMenu.tsx
acdunham: I put it in the context because the screen component needs to know the side menu open status so that it can adjust the margin etc accordingly and shift to the side

tcaiger (approved):
Looks good. Thanks for the changes

PR #4959, opened 2023-09-06
Add setup-test command to tupaia-web-server

EMcQ-BES:
`tupaia-web-server` tests are failing in github actions because we don't correctly run the initial setup for the database, so the base `tupaia` user hasn't been created when it tries to run tests. Add the appropriate line to the `package.json`


acdunham (approved):
Thank you!

PR #4961, opened 2023-09-06
WAITP-1211: Exclude certain types from response

acdunham:
### Changes:
- Exclude certain types from entities response, as specified by `database`

### DIFF
--- a/packages/tupaia-web-server/src/utils/generateFrontendExcludedFilter.ts
+++ b/packages/tupaia-web-server/src/utils/generateFrontendExcludedFilter.ts
@@ -8,11 +8,14 @@ interface FrontEndExcludedConfig {
 }
 
 // In the db project.config.frontendExcluded is an array with one entry for some reason
-export function generateFrontendExcludedFilter({
-  frontendExcluded,
-}: {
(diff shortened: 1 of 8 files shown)


packages/tupaia-web-server/src/utils/generateFrontendExcludedFilter.ts
```diff
@@ -8,17 +8,25 @@ interface FrontEndExcludedConfig {
...
           comparisonValue: frontendExcluded[0].types,
         },
       }
-    : {};
+    : {
+        type: {
+          comparator: '!=',
+          comparisonValue: defaultFrontendExcludedFields,
+        },
+      };
```
EMcQ-BES, line 31: Should we default here? Or combine?

EMcQ-BES (approved):
Looks good. One question, which I'm pretty sure you're correct on, just want a quick confirmation.

packages/tupaia-web-server/src/utils/generateFrontendExcludedFilter.ts
acdunham: I think we default, seems like the `web-config-server` was either doing one or the other

PR #4965, opened 2023-09-07
Waitp 1427 survey submit success screen

chaoxinnhin:
### Issue #:
WAITP-1427 Survey submitted screen

### Changes:

- Changed the submit success page 
---

### Screenshots:
image


packages/datatrak-web/src/Routes.tsx
```diff
@@ -95,10 +95,12 @@ export const Routes = () => {
             </Route>
             <Route path={ROUTES.SURVEY} element={<SurveyPage />}>
               <Route index element={<SurveyStartRedirect />} />
-              <Route path={ROUTES.SURVEY_REVIEW} element={<SurveyReviewScreen />} />
               <Route path={ROUTES.SURVEY_SUCCESS} element={<SurveySuccessScreen />} />
-              <Route path={ROUTES.SURVEY_SCREEN} element={<SurveyScreen />} />
-              <Route path={ROUTES.QUESTIONS} element={<SurveyQuestionsPage />} />
+              <Route element={<ScrollableLayout />}>
+                <Route path={ROUTES.SURVEY_REVIEW} element={<SurveyReviewScreen />} />
+                <Route path={ROUTES.QUESTIONS} element={<SurveyQuestionsPage />} />
```
tcaiger, line 101: Can you please delete the SurveyQuestionsPage as I just added that for development and testing for Wim and Chao and it's not needed any more.

packages/datatrak-web/src/Routes.tsx
```diff
@@ -95,10 +95,12 @@ export const Routes = () => {
             </Route>
             <Route path={ROUTES.SURVEY} element={<SurveyPage />}>
               <Route index element={<SurveyStartRedirect />} />
-              <Route path={ROUTES.SURVEY_REVIEW} element={<SurveyReviewScreen />} />
               <Route path={ROUTES.SURVEY_SUCCESS} element={<SurveySuccessScreen />} />
-              <Route path={ROUTES.SURVEY_SCREEN} element={<SurveyScreen />} />
-              <Route path={ROUTES.QUESTIONS} element={<SurveyQuestionsPage />} />
+              <Route element={<ScrollableLayout />}>
```
tcaiger, line 99: I think that the layouts seem a bit too nested here and it is a bit hard to follow and maintain. Do you think that it's possible to remove one layer of nesting by refactoring the SurveyPage and ScrollableLayout?

packages/datatrak-web/src/features/Survey/SurveyToolbar.tsx
```diff
@@ -40,10 +42,11 @@ const SurveyTitle = styled(Typography).attrs({
 `;
 
 export const SurveyToolbar = () => {
-  const { surveyCode } = useParams();
-  const { numberOfScreens, screenNumber } = useSurveyForm();
+  const { surveyCode, screenNumber } = useParams();
+  const { data: surveyScreenComponents } = useSurveyScreenComponents(surveyCode);
```
tcaiger, line 46: What is the reason for using useSurveyScreenComponents(surveyCode) instead of the context? Seems like the context would be better to use to me?

packages/datatrak-web/src/features/Survey/SurveySuccessScreen.tsx
```diff
@@ -5,31 +5,86 @@
...
+  const params = useParams();
+  const navigate = useNavigate();
+  const { activeScreen, formData, setFormData } = useSurveyForm();
+  const { reset } = useForm({ defaultValues: formData });
+
+  const repeatSurvey = data => {
+    setFormData(reset({ ...formData, ...data }));
+    const path = generatePath(ROUTES.SURVEY_SCREEN, {
+      ...params,
+      screenNumber: String(activeScreen),
```
tcaiger, line 67:
```suggestion

```
Screen number should either be not set or set to 1 to take us back to the start.

packages/datatrak-web/src/features/Survey/SurveySuccessScreen.tsx
```diff
@@ -5,31 +5,86 @@
...
+`;
+
 export const SurveySuccessScreen = () => {
+  const params = useParams();
+  const navigate = useNavigate();
+  const { activeScreen, formData, setFormData } = useSurveyForm();
+  const { reset } = useForm({ defaultValues: formData });
+
+  const repeatSurvey = data => {
+    setFormData(reset({ ...formData, ...data }));
```
tcaiger, line 64: This doesn't work. We don't need reset since that won't change the global form context value and setFormData wont work with empty values because of the way we spread the updated values in the form context reducer. I think it would be nice to add a new method to the reducer for resetting the form state and then we could use that here.

tcaiger (changes requested):
Thanks for picking up these extra changes. Mostly looks good - the form reset just needs to be tweaked.

packages/datatrak-web/src/Routes.tsx
acdunham: I wondered that, and sort of did something similar in the review screen PR, so will do that as part of that one

tcaiger (approved):
Awesome thanks. Great work Alex and Chao!

PR #4970, opened 2023-09-07
no-issue: Set winston as the logger for pg-pubsub

rohan-bes:
Means that we don't get those annoying console logs during the tests.

Decided to bump to latest version while at it.


chris-pollard (approved):
LGTM!

PR #4972, opened 2023-09-08
WAITP-1221: allow point and poly overlays together

acdunham:
### Changes:
- Split out poly and point layers so that they can exist at the same time on the map


tcaiger:
The code all looks good but when I tested it locally I noticed a regression where the wrong icons are showing up. I can't figure out why but would you take a look?
image

packages/tupaia-web/src/features/Map/MapOverlays/DataVisualsLayer/DataVisualsLayer.tsx
```diff
@@ -0,0 +1,63 @@
...
+  hiddenValues,
+}: {
+  hiddenValues: LegendProps['hiddenValues'];
+}) => {
+  const navigateToEntity = useNavigateToEntity();
+  const { projectCode, entityCode } = useParams();
+  const { data: entity } = useEntity(projectCode, entityCode);
+  const { serieses, measureData, activeEntity, selectedOverlay } = useMapOverlayData(hiddenValues);
+
+  useEffect(() => {
```
tcaiger, line 27: I find these GA calls quite messy and distracting. Maybe we could add a useGAEffect util in the utils/ga to clean it up and then it could be just a one liner `useGAEffect(selectedOverlay?.name)`

Also I wonder if the hooks should be in the react-query queries and maybe they could even be an onSuccess handler?!

tcaiger (changes requested):
The code changes look good and I agree that splitting it into smaller chunks was a good extra refactor. We just need to figure out what's going on with those icons.

packages/tupaia-web/src/ModalRoutes.tsx
```diff
@@ -42,19 +42,14 @@ export const ModalRoutes = () => {
   const { hash, closeModal } = useModal();
 
   const modal = hash as typeof MODAL_ROUTES[keyof typeof MODAL_ROUTES];
-  useEffect(() => {
-    if (modal !== null) {
-      gaEvent('User', modal, 'Open Dialog');
-    }
-  }, [modal]);
+  useGAEffect('User', 'Open Dialog', modal);
```
tcaiger, line 45: aahhh much nicer

packages/tupaia-web/src/utils/useGAEffect.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import { useEffect } from 'react';
+import { gaEvent } from '.';
+
+export const useGAEffect = (eventCategory: string, eventAction: string, watchValue?: string) => {
```
tcaiger, line 8: this is nice

tcaiger (approved):
Everything looks good to me

PR #4973, opened 2023-09-08
Fixup permission skip issues

EMcQ-BES:
### Changes:

- Create a new AccessPolicyBuilder that strips `BES Admin` permission out of your list
- Allow boilerplate auth middleware to pick up authenticators attached to the `req` object
- Allow `central-server` to selectively use different policy builders
- Setup `tupaia-web-server` to disable admin permissions in central fetches


packages/tupaia-web-server/src/routes/CountryAccessListRoute.ts
```diff
@@ -17,6 +17,6 @@ export type CountryAccessListRequest = Request<
 export class CountryAccessListRoute extends Route<CountryAccessListRequest> {
   public async buildResponse() {
     const { ctx } = this.req;
-    return ctx.services.central.fetchResources('me/countries');
+    return ctx.services.central.fetchResources('me/countries', { disableAdmin: true });
```
EMcQ-BES, line 20: Extremely WET for no reason. But the current boilerplate code makes it difficult to override some these connections in a useful way. I've split out a ticket WAITP-1433 to make that easier and get rid of all these hard set disables.

rohan-bes:
I think based on the discussion we've had on slack we're going to leave this functionality as is with `BES Admin` users having full access to Tupaia.

Sorry that the work you've here won't be used. It's a really clean implementation!

PR #4974, opened 2023-09-08
No issue: Enable nginx on-the-fly compression

passcod:
(I don't know about Tupaia PR etiquette)


rohan-bes (approved):
Thanks for this @passcod 🙏 So speedy!!

Etiquette is to squash-n-merge PRs into `dev`, and to prefix commit messages with the issue number. Since this has no issue, we prefix with `no-issue: `

PR #4977, opened 2023-09-11
No issue: Fixup servers conf

EMcQ-BES:
### Changes:

- Fix `nging` typo
- Update h5bp to latest release
  - Why are we using a version from 2017??
  - I don't know if we want to upgrade to an inbetween version, but upgrading is necessary if we want to use `compression.conf`

### DIFF
--- a/packages/devops/scripts/deployment/setupGoldMaster.sh
+++ b/packages/devops/scripts/deployment/setupGoldMaster.sh
@@ -7,7 +7,7 @@
 sudo apt-get install -yqq nginx
 git clone https://github.com/h5bp/server-configs-nginx.git
 cd ./server-configs-nginx
-git checkout tags/2.0.0
+git checkout tags/5.0.1
 cd ..
 sudo cp -R ./server-configs-nginx/h5bp/ /etc/nginx/
 rm -rf server-configs-nginx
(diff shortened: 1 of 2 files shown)


packages/devops/scripts/deployment/setupGoldMaster.sh
```diff
@@ -7,7 +7,7 @@
 sudo apt-get install -yqq nginx
 git clone https://github.com/h5bp/server-configs-nginx.git
 cd ./server-configs-nginx
-git checkout tags/2.0.0
+git checkout tags/5.0.1
```
EMcQ-BES, line 10: According to git blame this line was added by @edmofro in late 2021. I'm curious if you have any memory why we picked `2.0.0`, that release was already 4 years old at that point. Is there a specific reason we want to use `2.0.0`? If so we'd need to undo these uses of `compression.conf`

edmofro: Some informed speculation on my reasoning: it's what we were using at the time, and to limit the scope of what was quite a big undertaking (moving us to blue/green deployments using a base AMI), I kept it as is so that I didn't have to review breaking changes.

edmofro (approved):
This lgtm, we'll need to manually deploy the change to the gold master script - let's do that together after merge

PR #4980, opened 2023-09-11
WAITP-1211: Filter child_codes when fetching entities in tupaia-web-server

EMcQ-BES:
### Changes:

- Prevents an issue where drop down arrows appear on entities without children in the entity search list
- We request two layers of entities at a time, and we only filter the first layer in this implementation
  - If we wanted to filter all `child_codes` fields we'd need an additional fetch from `entity-server`, pull the filtered descendants of the entity list and the filter all `child_codes` fields by that list
  - I implemented _that_ version of code in the first commit, then replaced it with this version, which uses less api requests
    - But we the data from that version is more consistent, so we might still prefer it. This might be a tech design sort of discussion

### DIFF
--- a/packages/tupaia-web-server/src/routes/EntitiesRoute.ts
+++ b/packages/tupaia-web-server/src/routes/EntitiesRoute.ts
@@ -6,7 +6,7 @@
 import { Request } from 'express';
 import { Route } from '@tupaia/server-boilerplate';
 import camelcaseKeys from 'camelcase-keys';
-import { TupaiaWebEntitiesRequest, Entity } from '@tupaia/types';
+import { TupaiaWebEntitiesRequest } from '@tupaia/types';
 import { generateFrontendExcludedFilter } from '../utils';
 
 export type EntitiesRequest = Request<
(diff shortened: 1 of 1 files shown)


packages/tupaia-web-server/src/routes/EntitiesRoute.ts
```diff
@@ -6,7 +6,7 @@
 import { Request } from 'express';
 import { Route } from '@tupaia/server-boilerplate';
 import camelcaseKeys from 'camelcase-keys';
-import { TupaiaWebEntitiesRequest } from '@tupaia/types';
+import { TupaiaWebEntitiesRequest, Entity } from '@tupaia/types';
```
rohan-bes, line 9: Are we using the `Entity` type here for anything?

rohan-bes (approved):
Looks good 👍

packages/tupaia-web-server/src/routes/EntitiesRoute.ts
EMcQ-BES: Oh good catch 👍

PR #4985, opened 2023-09-12
WAITP-1439: Survey review screen

acdunham:
### Changes:
- Refactored file structure to group related survey components together
- Created review screen
- Created a layout and made some components reusable so that the various form screens can use the same layout

Note: this doesn't include any logic around actually submitting the survey, as this is to come in a separate PR

---

### Screenshots:
Screenshot 2023-09-12 at 2 06 38 pm


packages/datatrak-web/src/components/TopProgressBar.tsx
```diff
@@ -31,6 +31,9 @@ export const TopProgressBar = ({
   totalNumberOfSurveyQuestions,
 }: ProgressPercentage) => {
   const fraction = (currentSurveyQuestion / totalNumberOfSurveyQuestions) * 100;
+  console.log('fraction', fraction);
+  console.log('currentSurveyQuestion', currentSurveyQuestion);
+  console.log('totalNumberOfSurveyQuestions', totalNumberOfSurveyQuestions);
```
tcaiger, line 36:
```suggestion
```

acdunham: Doh

packages/datatrak-web/src/features/Survey/Components/CancelSurveyModal.tsx
```diff
@@ -35,8 +42,9 @@ const ModalButton = styled(Button)`
 
 export const CancelSurveyModal = ({ open, onClose }: { open: boolean; onClose: () => void }) => {
   return (
-    <Modal open={open} onClose={onClose} title="Survey in progress">
+    <Modal open={open} onClose={onClose}>
```
tcaiger, line 45: Can we remove the title prop from the modal since it's no longer being used.

packages/datatrak-web/src/features/Survey/Components/SurveyQuestionGroup.tsx
```diff
@@ -0,0 +1,73 @@
...
+  width: 3.5rem;
+  text-transform: lowercase;
+`;
+
+/**
+ * This is the component that renders questions in a survey.
+ */
+export const SurveyQuestionGroup = ({ questions }: { questions: SurveyScreenComponent[] }) => {
+  return (
+    <>
```
tcaiger, line 30:
```suggestion

```
Unnecessary fragment

tcaiger: on second thoughts this might be necessary to keep typescript happy, in which case no worries

packages/datatrak-web/src/features/Survey/Screens/SurveyReviewScreen.tsx
```diff
@@ -0,0 +1,65 @@
...
+          Please review your survey answers below. To edit any answers, please navigate back using
+          the 'Back' button below. Once submitted, your survey answers will be uploaded to Tupaia.{' '}
+        </Typography>
+      </Header>
+      <ScrollableBody as="form" noValidate>
+        {questionSections.map(({ heading, questions }, index) => (
+          <Section key={index}>
+            <SectionHeader>{heading}</SectionHeader>
+            <SurveyQuestionGroup questions={questions} />
+          </Section>
```
tcaiger, line 60: Sweet refactoring - this is beautifully organised

packages/datatrak-web/src/features/Survey/SurveyLayout.tsx
```diff
@@ -0,0 +1,175 @@
...
+      });
+    }
+
+    handleStep(path, data);
+  });
+
+  const handleSubmitForm = data => {
+    // Placeholder for actual submission logic
+    console.log(data);
+    navigate(ROUTES.SURVEY_SUCCESS);
```
tcaiger, line 120:
```suggestion
    navigate(generatePath(ROUTES.SURVEY_SUCCESS, params))
```

packages/datatrak-web/src/features/Survey/SurveyLayout.tsx
```diff
@@ -0,0 +1,175 @@
...
+                  color="default"
+                >
+                  Back
+                </Button>
+                <ButtonGroup>
+                  <Button onClick={openCancelModal} variant="outlined">
+                    Cancel
+                  </Button>
+                  <Button type="submit" onClick={handleSubmitScreen}>
+                    {isReviewScreen ? 'Submit' : 'Next'}
```
tcaiger, line 164: If it's the last screen, it should say "Review & submit"

packages/datatrak-web/src/features/Survey/SurveyLayout.tsx
```diff
@@ -0,0 +1,175 @@
...
+                  variant="text"
+                  color="default"
+                >
+                  Back
+                </Button>
+                <ButtonGroup>
+                  <Button onClick={openCancelModal} variant="outlined">
+                    Cancel
+                  </Button>
+                  <Button type="submit" onClick={handleSubmitScreen}>
```
tcaiger, line 163: If it's not possible, then it's not a big deal but my preference is to use the form onSubmit handler to submit each screen as it's more semantic and accessible and ensures the enter button works. I think it should be easy enough to just add a form element around the `Outlet` or some thing similar.

packages/datatrak-web/src/features/Survey/Screens/SurveyReviewScreen.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <Typography>
+          Please review your survey answers below. To edit any answers, please navigate back using
+          the 'Back' button below. Once submitted, your survey answers will be uploaded to Tupaia.{' '}
+        </Typography>
+      </Header>
+      <ScrollableBody as="form" noValidate>
+        {questionSections.map(({ heading, questions }, index) => (
+          <Section key={index}>
+            <SectionHeader>{heading}</SectionHeader>
+            <SurveyQuestionGroup questions={questions} />
```
tcaiger, line 59:
```suggestion
 <fieldset disabled>
     <SurveyQuestionGroup questions={questions} />
  </fieldset>
```
This wasn't mentioned in the original ticket that you picked up but the Review Screen should be readonly.

I did a quick google of how to make an entire form non-editable and wrapping a section of the form in a disabled fieldset seems like a quick but semantic way of doing it. It would also need a couple of style tweaks to disable default fieldset styles and turn off all pointer-events but it would do the trick. But I'm open to other approaches if you have another idea.

tcaiger (approved):
Great refactoring - looking tight. We just need to make the screen non-editable and then it's good to go.

packages/datatrak-web/src/features/Survey/Components/SurveyQuestionGroup.tsx
acdunham: Yeah I think it has to be there for TS

packages/datatrak-web/src/features/Survey/SurveyLayout.tsx
acdunham: The main reason I did this was because I didn't want to have headings (eg an `h2` inside the `form` element, which would happen if this was the case.

tcaiger: oh true that makes sense - that is a bit of a conundrum but having the headings inside the form is probably the lesser of the 2 evils so thanks for the change.

tcaiger (approved):
Thanks for the changes!

PR #4986, opened 2023-09-12
Fix appcenter build fail due to mathjs

IgorNadj:
```
> Task :app:bundleReleaseJsAndAssets
error While trying to resolve module `fraction.js` from file `/Users/runner/work/1/s/packages/expression-parser/node_modules/mathjs/lib/cjs/type/fraction/Fraction.js`, the package `/Users/runner/work/1/s/packages/expression-parser/node_modules/fraction.js/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`/Users/runner/work/1/s/packages/expression-parser/node_modules/fraction.js/fraction.cjs`. Indeed, none of these files exist:

```

See a broken [build](https://appcenter.ms/orgs/Beyond-Essential/apps/Tupaia-MediTrak-Android/build/branches/dev/builds/1882).

The issue is math.js had an unlocked dependency on fraction.js which upgraded and [broke](https://github.com/josdejong/mathjs/issues/3024) math.js. 

Two options to resolve:
1. Upgrade math.js to latest
    - This would include 2 major version jumps, I tried this and it was causing issues
2. Fork math.js
    - There are also easier options like the yarn resolutions field but for some reason appcenter build uses npm 🤯 which doesn't understand these. 
 
So I had to go with a fork. We can unfork at a later date when we go to upgrade mathjs.

### DIFF
--- a/packages/expression-parser/package.json
+++ b/packages/expression-parser/package.json
@@ -22,9 +22,9 @@
     "test:coverage": "yarn test --coverage"
   },
   "dependencies": {
+    "@beyondessential/mathjs": "v9.4.400",
     "date-fns": "^2.16.1",
-    "lodash.startcase": "^4.4.0",
-    "mathjs": "^9.4.0"
+    "lodash.startcase": "^4.4.0"
   },
   "devDependencies": {
     "npm-run-all": "^4.1.5"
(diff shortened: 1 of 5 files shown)


IgorNadj:
Green appcenter build for this PR: https://appcenter.ms/orgs/Beyond-Essential/apps/Tupaia-MediTrak-Android/build/branches/fix-appcenter-build-fail/builds/1889

rohan-bes (approved):
Looks good! I'm happy to have a crack at bumping mathjs version sometime soon. I'll add to maintenance sprint items

PR #4987, opened 2023-09-12
WAITP-1418: Misc FE fixes

acdunham:
### Changes:
- Fixed issue where titles weren't showing for multi single values correctly
- Fixed value formatting in matrix visuals
- Updated map overlay measure level handling for arrays
- Added 'no search result' message for matrix search
- Don't show data visuals layer when loading map overlay data


tcaiger (approved):
LGTM

PR #4989, opened 2023-09-13
WAITP-1242 Fixup headers on survey response export

EMcQ-BES:
### Changes:

- The filter argument was formatted incorrectly, causing an unfiltered request which was causing incorrect (but valid) data to be passed to the next step


acdunham (approved):
🙏

PR #4993, opened 2023-09-14
WAITP-1231: Handle isExpanded params for matrix report

acdunham:
### Changes:
- Added `isExpanded` param to query for enlarged reports, so that reports get handled correctly. 

Note: BE piece of this is a separate PR


tcaiger (approved):
LGTM

PR #4994, opened 2023-09-15
RN-994 Auto cleanup old ec2s

IgorNadj:
RN-994


rohan-bes (approved):
Great change! Thanks 🙏

PR #4995, opened 2023-09-15
RN-664: Move calculation of change_time for sync queues into application logic

rohan-bes:
### Issue RN-664:

This change is primarily motivated by testing interests. Having the `change_time` being calculated by db trigger means it's really hard to accurately await for when the sync queues have been properly updated.

I guess the other motivation is to reduce complexity a little.

Having the database be responsible for this though is a more resilient approach, so happy to drop this if we feel the cons outweigh the pros.

@edmofro wouldn't mind getting your perspective on this 🙏


chris-pollard (approved):
Looks good to me, from a code point of view. Will defer to @edmofro for a strategic review.

edmofro (approved):
Most of the robustness of the db approach is thwarted by the fact that everything else about these sync queues appropriately updating requires the server to be online, so I'm really happy with this change. Thanks @rohan-bes

PR #4997, opened 2023-09-18
WAITP-1410: Request project access

acdunham:
### Changes:
- Added RequestProjectAccess view in both page and modal view
- Added logic to stop user being able to select a project when not logged in
- Added `EditUserForMe` route handler so that editing own user is standardised
- Added `pending` state to project list
- Updated projects endpoint handling to return with access state
- Added single project endpoint and associated types
- Refactor select list so the checked state gets handled 

---

### Screenshots:

Pending state
Screenshot 2023-09-18 at 2 13 03 pm

Request project access modal
Screenshot 2023-09-18 at 2 13 21 pm

Request project access screen
Screenshot 2023-09-18 at 2 14 18 pm

Success state
Screenshot 2023-09-18 at 2 14 38 pm


packages/datatrak-web-server/src/app/createApp.ts
```diff
@@ -39,6 +41,7 @@ export function createApp() {
     .get<SurveysRequest>('surveys', handleWith(SurveysRoute))
     .get<SurveyRequest>('surveys/:surveyCode', handleWith(SurveyRoute))
     .get<ProjectsRequest>('projects', handleWith(ProjectsRoute))
+    .get<ProjectRequest>('project/:projectCode', handleWith(ProjectRoute))
```
tcaiger, line 44: any reason not to just use the central server project route?

packages/datatrak-web-server/src/routes/ProjectRoute.ts
```diff
@@ -0,0 +1,32 @@
...
+>;
+
+export class ProjectRoute extends Route<ProjectRequest> {
+  public async buildResponse() {
+    const {
+      ctx,
+      params: { projectCode },
+    } = this.req;
+
+    const { projects } = await ctx.services.webConfig.fetchProjects();
```
tcaiger, line 24: The plan is to phase out web-config-server because it's a bit of a frankenstein legacy server (which is why Ethan has been re-writing backend stuff so much on tupaia-web) so we should get the projects from central server instead.

packages/datatrak-web-server/src/routes/ProjectsRoute.ts
```diff
@@ -15,26 +14,10 @@ export type ProjectsRequest = Request<
...
-
 export class ProjectsRoute extends Route<ProjectsRequest> {
   public async buildResponse() {
-    const { ctx, query = {} } = this.req;
-    const projects = await ctx.services.central.fetchResources('projects', {
-      columns,
-    });
-    return camelcaseKeys(projects, { deep: true });
+    const { ctx } = this.req;
+    const { projects } = await ctx.services.webConfig.fetchProjects();
```
tcaiger, line 20: as mentioned above, we are trying our best not to use web-config-server anymore

packages/datatrak-web-server/src/routes/UserRoute.ts
```diff
@@ -32,6 +32,12 @@ export class UserRoute extends Route<UserRequest> {
...
     } = await ctx.services.central.getUser();
 
-    return { userName: `${firstName} ${lastName}`, email, id, projectId };
+    let project = null;
+    if (projectId) {
+      const { projects } = await ctx.services.webConfig.fetchProjects();
+      project = projects.find((p: WebServerProjectRequest.ResBody) => p.id === projectId);
+    }
+
+    return { userName: `${firstName} ${lastName}`, email, id, projectId, project };
```
tcaiger, line 41: I like the idea of attaching the project here since it makes life much easier on the front end

packages/datatrak-web/src/features/ProjectSelectForm.tsx
```diff
@@ -31,99 +31,100 @@ const ListWrapper = styled.div`
...
+  const getProjectTooltip = (hasAccess: boolean, hasPendingAccess: boolean) => {
+    if (!hasAccess && !hasPendingAccess) return 'Request project access';
+    return '';
+  };
+
+  const getProjectContent = (name: string, hasPendingAccess: boolean) => {
+    // customise the tooltip to appear over the project name if there is pending access, instead of the default of over the whole list item
+    if (hasPendingAccess)
+      return (
+        <Tooltip title="Approval in progress">
```
tcaiger, line 89: Is it possible to consolidate the tooltips to either both use the ListItem tooltip prop or both be part of the content?

packages/datatrak-web/src/features/ProjectSelectForm.tsx
```diff
@@ -31,99 +31,100 @@ const ListWrapper = styled.div`
...
+      return {
+        content,
+        value: id,
+        code,
+        selected: id === selectedProjectId,
+        icon,
+        tooltip,
+        button: !hasPendingAccess,
+        appearsDisabled: hasAccess === false,
+        hasAccess
```
tcaiger, line 110: I find that this approach can get unwieldy as we end up with a prop to support every style difference and the list of props can get quite long if there are lots of different implementations of the components. It looks like there are only 3 use cases of the List in the design so this list of props should be the full extent so it should be all good and I do like how this approach is more convenient to use once the component is set up.

packages/datatrak-web/src/features/RequestProjectAccess/ProjectAccessForm.tsx
```diff
@@ -0,0 +1,200 @@
...
+  if (isLoading)
+    return (
+      <LoaderWrapper>
+        <SpinningLoader />
+      </LoaderWrapper>
+    );
+
+  return (
+    <Form onSubmit={submitForm as SubmitHandler<any>} formContext={formContext}>
+      {isError && <Error>{error.message}</Error>}
```
tcaiger, line 159: I'm not exactly sure how it will work but I think that we will try to use the error and success toasts to display errors like this throughout the UI. I'm happy if you want to leave this here for now so we don't forget to handle it though.

packages/types/src/types/requests/web-server/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+/**
+ * This is the index file for all web server requests. This folder is for request types that are common between multiple web servers, e.g. datatrak-web and tupaia-web.
+ */
+export * as WebServerProjectRequest from './ProjectRequest';
```
tcaiger, line 9: Good idea - thanks for adding this

packages/datatrak-web/src/Routes.tsx
```diff
@@ -48,9 +63,10 @@ const LoggedInRedirect = ({ children }) => {
 
 // Reusable wrapper to handle redirecting to login if user is not logged in and the route is private
 const PrivateRoute = ({ children }: { children?: ReactNode }): any => {
-  const { isLoggedIn, isLoading, isFetched } = useUser();
+  const { isLoggedIn, isLoading, isFetched, data } = useUser();
   if (isLoading || !isFetched) return <FullPageLoader />;
   if (!isLoggedIn) return <Navigate to="/login" replace={true} />;
+  if (!data?.projectId) return <Navigate to={ROUTES.PROJECT_SELECT} replace={true} />;
```
tcaiger, line 69:
```suggestion
if (location.pathname === ROUTES.PROJECT_SELECT && data?.projectId) 
    return <Navigate to={ROUTES.HOME} replace={true} />;
  
  if (!data?.projectId) return <Navigate to={ROUTES.PROJECT_SELECT} replace={true} />;
```
The ProjectSelectRedirect seems so similar to the PrivateRoute and so it would be nice if we could just extend the private route component. This would clean up things a bit and also makes more sense to me since the ProjectSelectRedirect is still doing the Private Route check. It is a bit messy to add project id logic to the PrivateRoute but since we are already doing that, we might as well just do it agian.

tcaiger (changes requested):
Hey, this is looking really good - I always forget how complex the requesting access is. It's probably going to make it more complex but unfortunately we are not meant to use the web-config-server anymore so we need to redo those data fetches to be from the central-server

tcaiger:
> Hey, this is looking really good - I always forget how complex the requesting access is. It's probably going to make it more complex but unfortunately we are not meant to use the web-config-server anymore so we need to redo those data fetches to be from the central-server

On further reflection, this PR is probably not the place where you should have to re-write the web-config-server endpoints, so can you please make another ticket for them and put it in the refinement queue and then we can either do it as a standalone task for datatrak or as a maintenance item. thanks

packages/datatrak-web-server/src/routes/ProjectRoute.ts
tcaiger: As I mentioned below, let's split out the work to refactor this into a new ticket.

acdunham: Sure - the reason for this is that getting it from the central server doesn't return things like `name`, `hasAccess` etc, which we need

packages/datatrak-web/src/features/ProjectSelectForm.tsx
acdunham: The reason here is that the pending access state has the tooltip over just the text and the others have over the full width list item

tcaiger (approved):
LGTM

PR #5003, opened 2023-09-19
RN-988: Switched PSSS survey response CRUD over to using /surveyResponse endpoint

rohan-bes:
### Issue RN-988:

Previously PSSS has been using the `POST /changes` endpoint in the `central-server` for survey response submission. This endpoint really ought to just be used by meditrak, and will shortly be moved over to the `meditrak-app-server`.

So this change is just to switch PSSS over to the more appropriate `/surveyResponse` endpoints

### DIFF
--- a/packages/central-server/src/apiV2/surveyResponse.js
+++ b/packages/central-server/src/apiV2/surveyResponse.js
@@ -3,6 +3,7 @@
  * Copyright (c) 2019 Beyond Essential Systems Pty Ltd
  */
 
+import { AnalyticsRefresher } from '@tupaia/database';
 import {
   ValidationError,
   MultiValidationError,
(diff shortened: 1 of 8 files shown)


packages/central-server/src/apiV2/surveyResponse.js
```diff
@@ -135,6 +136,11 @@ export async function surveyResponse(req, res) {
     );
 
     results = await saveResponsesToDatabase(transactingModels, userId, responses);
+
+    if (req.query.waitForAnalyticsRebuild) {
```
rohan-bes, line 140: I don't necessarily love adding these flags, however they're required for PSSS. And until we have a cleaner system for ingesting survey responses I think they'll stay.

chris-pollard (approved):
Code looks good!

PR #5004, opened 2023-09-19
WAITP-1421: Handle user request errors

tcaiger:

### DIFF
--- a/packages/tupaia-web/index.html
+++ b/packages/tupaia-web/index.html
@@ -112,12 +112,43 @@
       background: #262834;
       color: white;
     }
+    @keyframes spinner {
+       to {
+           transform: rotate(360deg);
+       }
+     }
+    #tupaia-spinner {
+        position: fixed;
+        top: 0;
+        left: 0;
+        height: 100vh;
+        width: 100vw;
+        z-index: 9999;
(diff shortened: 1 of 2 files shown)


packages/tupaia-web/index.html
```diff
@@ -112,12 +112,43 @@
       background: #262834;
       color: white;
     }
+    @keyframes spinner {
```
tcaiger, line 115: this isn't really related but I took the chance to add an initial loader to the page

packages/tupaia-web/src/Routes.tsx
```diff
@@ -53,12 +53,14 @@ const UserPageRedirect = ({ modal }: { modal: MODAL_ROUTES }) => {
  *
  * **/
 export const Routes = () => {
-  const { isLoading } = useUser();
+  const { isLoading, isFetched } = useUser();
+
+  const showLoader = isLoading && !isFetched;
```
tcaiger, line 58: When there was an error in the user endpoint, isLoading was switching back and forth from true to false and getting stuck in an infinite loop. Adding the isFetched check, breaks the circuit.

IgorNadj (approved):
LGTM

PR #5005, opened 2023-09-19
WAITP-1210: Tupaia-web frontend fixes

tcaiger:
### Changes:

- Redirect to home on logout
- Fix load more button on Entity Search


acdunham (approved):
Thanks for this!

PR #5007, opened 2023-09-19
feat(meditrak): RN-502: Switch meditrak-app over to the meditrak-app-server

rohan-bes:
### Issue RN-502:

This is the big switch over! Might be easiest to review commit by commit.

Main thing to be aware of here is that we're doing a switch over of `api.tupaia.org` to now point to the `meditrak-app-server`. That way existing meditrak-app clients can switch over without needing to upgrade.

We will now server `central-server` at `central-api.tupaia.org`.

### Merge tasks:
- [ ] Update all `CENTRAL_API_URL` .env vars to point to `*-central-api.tupaia.org`
- [ ] Update the `subdomains` ec2 instance tag to include `central-api` and create the route53 entry for production
- [ ] Remove `TUPAIA_APP_SERVER_URL` .env var from `web-config-server` .env vars


packages/meditrak-app-server/src/index.ts
```diff
@@ -10,7 +10,6 @@ import http from 'http';
 import winston from 'winston';
```
IgorNadj, line 10: (Unrelated to this file)
In package.json we have a script that calls `node dist/sync/createMeditrakSyncView.js`. The dist files for this may not be present. Can we use babel-node like we do in other packages for consistency? This also seems like a script so it should be outside the src dir.

scripts/bash/getDeployablePackages.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash -e
 
-echo "admin-panel" "lesmis" "psss" "datatrak-web" "tupaia-web" "web-frontend" "central-server" "data-table-server" "datatrak-web-server" "entity-server" "lesmis-server" "meditrak-app-server" "psss-server" "report-server" "tupaia-web-server" "web-config-server" "admin-panel-server" # admin-panel-server last as it depends on report-server
+# meditrak-app-server before central-server so that the meditrak-sync-queue is listening before we run the migrations
```
IgorNadj, line 3: Hmm, I don't love the way this implies servers need to start in a specific order. IMO servers should be smart enough to not have to be run in any specific order. The start order is also not guaranteed in general. 

They probably do start in this order the first time when these are started with `pm2 start x` but not subsequently when the machine starts again and loads the pm2 save file. And in the future if we move to a pure pm2 config file setup/or another similar infrastructure by declaration, then everything will start at the same time.

packages/central-server/src/apiV2/meditrakApp/getChanges.js
```diff
@@ -1,112 +0,0 @@
-/**
```
IgorNadj, line 1: Soooo good to see all this moved out! Thanks for doing this!

scripts/bash/getDeployablePackages.sh
IgorNadj: Discussed, will sort out in a separate PR.

packages/devops/scripts/deployment/startBackEnds.sh
```diff
@@ -10,6 +10,13 @@ PACKAGES=$(${TUPAIA_DIR}/scripts/bash/getDeployablePackages.sh)
 # Start back end server packages
 for PACKAGE in ${PACKAGES[@]}; do
     if [[ $PACKAGE == *server ]]; then
+        if [[ $PACKAGE == 'meditrak-app-server' ]]; then
+            # reset cwd back to `/tupaia`
+            cd ${TUPAIA_DIR}
+
+            # ensure that the latest permissions based meditrak sync queue has been built
+            yarn workspace @tupaia/meditrak-app-server create-meditrak-sync-view
```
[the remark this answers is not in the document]
IgorNadj: Discussed, could version and insert with migrations like we do the analytics table fns.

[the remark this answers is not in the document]
rohan-bes: @IgorNadj I've created a separate PR for this one: https://github.com/beyondessential/tupaia/pull/5063

packages/meditrak-app-server/src/index.ts
rohan-bes: Done in: https://github.com/beyondessential/tupaia/pull/5063

IgorNadj (changes requested):
Just the things we found when we went through it together, otherwise looks great!

PR #5012, opened 2023-09-21
WAITP-1316: Map overlay table fixes

acdunham:
### Changes:
- Fix issues in map flippa table display
- Fix issue where point and poly layers display at the same time sometimes unnecessarily


tcaiger (approved):
Looks good to me. I will copy across the change to the polygon layer to my refactor after this is merged.

PR #5015, opened 2023-09-22
WAITP-1437 toast messages

acdunham:
### Changes:
- Added toast messages to Datatrak web
- Apply toast message error handling by default, opting out where appropriate

---

### Screenshots:
Generic success message
Screenshot 2023-09-22 at 5 05 05 pm

Coconut message
Screenshot 2023-09-22 at 5 05 12 pm

Pig message
Screenshot 2023-09-22 at 5 05 17 pm

Error message
Screenshot 2023-09-22 at 5 05 21 pm


packages/datatrak-web/src/layout/MainPageLayout.tsx
```diff
@@ -6,18 +6,22 @@
...
   min-height: 100vh;
+  .notistack-SnackbarContainer {
+    top: calc(1rem + ${HEADER_HEIGHT});
+  }
 `;
 
 export const MainPageLayout = () => {
   return (
-    <PageWrapper>
+    <PageWrapper id="page">
```
tcaiger, line 24: What is the id for?

acdunham: It's so that we can tell the toast messages to sit within this container, instead of being on top of the header

tcaiger: ah yes I just saw that in the other file. Makes sense. Would you mind adding a comment in this file too. thanks

packages/datatrak-web/src/components/Toast/ToastProvider.tsx
```diff
@@ -0,0 +1,27 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import React from 'react';
+import { SnackbarProvider } from 'notistack';
+import { Toast } from './Toast';
+
+export const ToastProvider = ({ children }: { children: React.ReactNode }) => {
```
tcaiger, line 9: hmmm I'm not 100% sure but I wouldn't think to look in components for the provider. Maybe this should just go in the AppProviders file? Or in a new Providers directory if we want to split out this provider and the react query provider?

packages/datatrak-web/src/AppProviders.tsx
```diff
@@ -3,19 +3,32 @@
...
+    },
+  }),
   defaultOptions: {
     queries: {
       staleTime: 1000 * 60 * 5, // 5 minutes
       refetchOnWindowFocus: false,
       retry: 0,
       keepPreviousData: false,
+      // use the errorToast function to display errors by default. If you want to override this, apply an onError function to the query
+      onError: (error: any) => {
```
tcaiger, line 29: This is a nice api how we can just override it in the useQuery implementation 👍

packages/datatrak-web/src/components/Toast/utils/useToast.tsx
```diff
@@ -0,0 +1,40 @@
...
+ */
+import React from 'react';
+import { enqueueSnackbar } from 'notistack';
+import { Coconut, Pig } from '../../Icons';
+import { errorToast } from '.';
+
+/**
+ * Utility hook for displaying toast messages.
+ */
+export const useToast = () => {
```
tcaiger, line 13: This doesn't actually need to be a custom hook since there is no useState or useEffect in it. We could just export each of these as a util function or maybe we don't even need them since the pig and coconut will probably only be used in one place.

tcaiger (changes requested):
I love the design and implementation - I just left a couple of nitpicky change requests.

tcaiger (approved):
Looks good. Thanks for cleaning up that bad merge situation.

PR #5017, opened 2023-09-24
WAITP-1443 Better entity filtering

EMcQ-BES:
### Changes:

- Add a secondary filter to entity routes that checks for a users permissions against the specific project they're viewing
- Attach access policy to the `req` object directly (using session switching)


rohan-bes (changes requested):
Functionality looks great. I'm thinking that having this behaviour be standard across the `entity-server` would be nice.

I think that should work well across Admin Panel, LESMIS, and PSSS too.

packages/tupaia-web-server/src/utils/generateAccessibleCountryFilter.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import { AccessPolicy } from '@tupaia/access-policy';
+
+const BES_ADMIN_PERMISSION_GROUP = 'BES Admin';
+
+// Takes an access policy and list of permission groups, generates the list of
+// accessible countries
+// Used to filter results from the entity server
+export function generateAccessibleCountryFilter(
```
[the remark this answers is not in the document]
EMcQ-BES: I originally started doing this work against the entity server but the code wound up getting quite messy. Although I think that was because I was trying to make the whole thing switchable with a parameter. If we're happy having it applied to all users of the entity server I can take another shot at applying the logic in there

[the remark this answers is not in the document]
rohan-bes: Yep I reckon we're good to apply it generally across the entity-server if you're happy to give it another shot.

I can't think of a real scenario where we really want to be returning all countries the user has any access to regardless of project. I think we designed that way originally since the idea was that multiple projects share one entity hierarchy. But since we scrapped that a while back I think filtering by project permissions makes a lot of sense 👍

packages/tupaia-web-server/src/routes/EntityAncestorsRoute.ts
```diff
@@ -38,7 +38,9 @@ export class EntityAncestorsRoute extends Route<EntityAncestorsRequest> {
       projectCode,
       rootEntityCode,
       {
-        filter: generateFrontendExcludedFilter(config, typesExcludedFromWebFrontend),
+        filter: {
+          ...generateFrontendExcludedFilter(config, typesExcludedFromWebFrontend),
+        },
```
rohan-bes, line 43: We don't need these spreads for anything do we?

packages/entity-server/src/models/Project.ts
```diff
@@ -13,6 +13,7 @@ type ProjectFields = Readonly<{
   name: string;
   entity_id: string;
   entity_hierarchy_id: string;
+  permission_groups: string[];
 }>;
```
rohan-bes, line 17: Optional, but feel free to sub out this whole `ProjectFields` type with the `Project` type from `@tupaia/types`. We're steadily cleaning all these duplicate type defs up.

rohan-bes (approved):
Great stuff! So minimal. Couple of minor comments. Could I also request that at least one test gets added to the `permissions.test.ts` to check that we filter by the appropriate project permission? :pray: 

Pre-approving :+1:

EMcQ-BES:
@rohan-bes Rather than add a new unit test, I've updated the existing permissions unit tests to trigger the specific issue I've fixed:

```
- grantAccessToCountries(['JOHTO']);
+ grantAccess({ JOHTO: ['Public'], KANTO: ['Team Rocket'] });
```

This permission setup on `dev` causes all the `permission.test.ts` tests to fail since it respects _all_ permissions to an entity. On this branch they successfully filter and pass.

packages/entity-server/src/__tests__/__integration__/permissions.test.ts
```diff
@@ -4,15 +4,15 @@
...
+import { grantAccess, revokeCountryAccess, setupTestApp } from '../testUtilities';
 import { getEntitiesWithFields, getHierarchiesWithFields } from './fixtures';
 
 describe('permissions', () => {
   let app: TestableServer;
 
   beforeAll(async () => {
     app = await setupTestApp();
-    grantAccessToCountries(['JOHTO']);
+    grantAccess({ JOHTO: ['Public'], KANTO: ['Team Rocket'] });
```
rohan-bes, line 15: :heart: Meowth! That's right!

PR #5020, opened 2023-09-25
Use direct secrets rather than going through lastpass

IgorNadj:


IgorNadj:
Replaced by https://github.com/beyondessential/tupaia/pull/5021

PR #5023, opened 2023-09-25
WAITP-1407: Add support for user preferences to user accounts

tcaiger:
### Changes:

- Update user account schema with new preferences column (and remove old project_id column)
- Add support for saving country_id or project_id to edit user endpoint
- Update datatrak-web-server to get projectId from user preferences


packages/central-server/src/apiV2/userAccounts/EditUserAccounts.js
```diff
@@ -37,13 +39,39 @@ export class EditUserAccounts extends EditHandler {
...
+      const { preferences } = userRecord;
+
+      const updatedPreferenceFields = updatedUserPreferences.reduce((obj, [key, value]) => {
+        return { ...obj, [key]: value };
+      }, {});
+
+      updatedFields = {
+        preferences: { ...preferences, ...updatedPreferenceFields },
+        ...updatedFields,
+      };
```
EMcQ-BES, line 72: One note here. If someone sends a `preferences` field it will override all of this.

```
updatedFields = {
  preferences: { country_id: 'DL' },
  project_id: 'explore',
}
```
Something like this would only submit the `country_id`, do we want to combine with the passed in `preferences` field? Or maybe we should error if someone tries to send a `preferences` field? 🤔

EMcQ-BES (approved):
Code does the job, question about how we want to handle an edge case

packages/central-server/src/apiV2/userAccounts/EditUserAccounts.js
tcaiger: Good catch thanks. I think that the intention of my api is that it only allows setting specified fields (for now anyway) so I like the idea of throwing an error if a preferences field is sent.

packages/types/src/types/requests/datatrak-web-server/UserRequest.ts
```diff
@@ -10,8 +10,9 @@ export interface ResBody {
   id?: string;
   userName?: string;
   email?: string;
-  projectId?: string;
+  preferences?: any;
   project?: ProjectResponse | null;
+  projectId?: string;
```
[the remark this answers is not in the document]
tcaiger: Yes the user route still returns the projectId field for convenience on the front end since it's used in a bunch of places for conditional rendering. I'm not sure if that's bad practise but it seemed like the orchestration server making things convenient for the front end is part of their role

PR #5025, opened 2023-09-25
no-issue: Fix broken sync queues

rohan-bes:
### Issue https://beyondessential.slack.com/archives/C01MAA3NKM1/p1695611754430549:

In this release we are now calculating the sync queue `change_time` in the application. An issue was discovered in the release that there can be clashes with the way we calculate this `change_time`.

Fix this by:
1. Using the same function for determining the change time as we'd been using previously
2. Switch a couple instances of `Promise.all()` to explicit loops in order to preserve change order


IgorNadj (approved):
Over my head but looks ok

PR #5026, opened 2023-09-26
WAITP-1431: Select country

acdunham:
### Changes:
- Added country selector to select survey view
- Added logic to update user preferences with country id
- Added logic to return country to with user if set
- Added mobile styling to select survey view
- Added country name to survey title
- Added country to filters when getting entities for entity question
---

### Screenshots:

Screenshot 2023-09-26 at 1 15 52 pm

Screenshot 2023-09-26 at 1 16 07 pm

Screenshot 2023-09-26 at 1 16 24 pm


packages/datatrak-web/src/api/mutations/useEditUser.ts
```diff
@@ -6,22 +6,34 @@
...
         return;
       }
 
+      const updates = {} as {
+        project_id?: Project['id'];
+        country_id?: Entity['id'];
+      };
+
+      if (projectId) {
+        updates.project_id = projectId;
```
tcaiger, line 28: It's a shame that we have to convert these to snake case on the front end but there's nothing we can really do about it since we are not going to the orchestration server for this one

packages/datatrak-web/src/api/queries/useSurveys.ts
```diff
@@ -17,4 +18,9 @@ export const useSurveys = () => {
         },
       }),
   );
+  // Filter the surveys by the selected country name
+  return {
+    isLoading: isLoading || !isFetched,
+    surveys: surveys?.filter((survey: any) => survey.countryNames?.includes(selectedCountryName)),
```
tcaiger, line 24: This is an optional suggestion but I came across the concept of [react-query selectors](https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option) this morning and this would be a good candidate for it. The advantage is that it caches the transformation and also minimises re-renders.

packages/datatrak-web/src/features/Survey/Components/SurveyToolbar.tsx
```diff
@@ -42,18 +43,25 @@ const SurveyTitle = styled(Typography).attrs({
...
+const CountryName = styled.span`
+  padding-left: 0.3rem;
+  font-weight: ${({ theme }) => theme.typography.fontWeightRegular};
+`;
+
 export const SurveyToolbar = () => {
   const { surveyCode } = useParams();
   const { screenNumber, numberOfScreens } = useSurveyForm();
   const { data: survey } = useSurvey(surveyCode);
+  const { selectedCountry } = useUserCountries();
```
tcaiger, line 55:
```suggestion
  const { data: { country } } = useUser();
```
Can we use the useUser query instead of the useUserCountries util. There is a lot of complexity in useUserCountries and it would be good not to tie it here unnecessarily.

packages/datatrak-web/src/features/Questions/EntityQuestion/utils.ts
```diff
@@ -2,17 +2,15 @@
...
+import { useUserCountries } from '../../../utils';
 import { useSurveyForm } from '../../Survey/SurveyContext';
 
-// Todo: Remove this once we have a way to get the country id for the survey (WAITP-1431)
-const getCountryCode = () => 'DL';
-
 export const useEntityBaseFilters = config => {
   const { getAnswerForQuestion } = useSurveyForm();
-  const countryCode = getCountryCode();
+  const { selectedCountry } = useUserCountries();
```
tcaiger, line 10: Same here, can we use the useUser hook instead?

packages/datatrak-web/src/utils/useUserCountries.ts
```diff
@@ -0,0 +1,58 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import { useState } from 'react';
+import { useEntities, useUser } from '../api/queries';
+import { Entity } from '../types';
+
+export const useUserCountries = () => {
```
tcaiger, line 9: Great to see all these edge cases thought out and handled. I think it would be nice to re-organise where the code lives a bit to make it easier to maintain. I think that it could be nice to create a CountrySelect component that could live in `features` that has this and the JSX related to it either in the same file or a folder. I'm open to other organisation too, I'm just keen to move this code out of utils since it is pretty specific to one bit of ui.

packages/datatrak-web/src/views/SurveySelectPage.tsx
```diff
@@ -2,21 +2,36 @@
...
+import { Entity, Survey } from '../types';
+import { useUserCountries } from '../utils';
+import { useSurveys } from '../api/queries';
 
 const Container = styled(Paper).attrs({
   variant: 'outlined',
 })`
   width: 48rem;
   display: flex;
   flex-direction: column;
```
tcaiger, line 22: I'm not sure how to best fix it, but the paper moves around a bit when I select a country because the length of the list changes. Maybe this should be justified to the top and there could be a min height?

tcaiger (changes requested):
Looks good and I like how snappy it is when changing the country. I just have a couple of suggestions about organisation. Cheers

tcaiger (approved):
Looks awesome - thank for the changes!

PR #5027, opened 2023-09-26
WAITP-1423: Primary entity question

tcaiger:
### Changes:

- Move mocks outside src directory and add mock for testing form submit
- Setup form submit mutation and add handling of primary entity question
- Update entity question with attributes filter

---

### Screenshots:
image


packages/datatrak-web/src/features/Survey/Components/SurveyQuestionGroup.tsx
```diff
@@ -56,7 +56,7 @@ export const SurveyQuestionGroup = ({ questions }: { questions: SurveyScreenComp
                 detailLabel={detailLabel}
                 id={questionId}
                 code={questionCode}
-                name={questionId}
+                name={questionCode}
```
tcaiger, line 59: I'm not 100% sure if it's better to use the questionId, or the questionCode for the form data. It might make more sense one way or the other when we do the validation and form submit later on

acdunham (changes requested):
Looks good! Just one change requested from me - using `questionId` instead of `questionCode` for answers

packages/datatrak-web/src/features/Questions/TextQuestion.tsx
```diff
@@ -25,7 +25,7 @@ export const TextQuestion = ({
   label,
   name,
   type,
-  controllerProps: { onChange, value, ref },
+  controllerProps: { onChange, value = '', ref },
```
tcaiger, line 28: Otherwise there is the controlled vs uncontrolled errror

acdunham (approved):

PR #5028, opened 2023-09-26
hot-fix: skip migration validation on master

chris-pollard:
This hot-fix updates the validateMigration CI script to ignore the process when the current branch is the same as origin (master).


rohan-bes (approved):
Thanks!

PR #5032, opened 2023-09-27
WAITP-1349: Survey submit

acdunham:
### Changes:
- Add logic for survey submit, including processing of survey response into a shape the API will accept
- Add logic for showing 'repeat survey' button
- Add a loader when survey is submitting


packages/datatrak-web/src/features/Survey/SurveyContext.tsx
```diff
@@ -74,6 +77,11 @@ export const surveyReducer = (
         ...state,
         formData: {},
       };
+    case ACTION_TYPES.SET_SURVEY_START_TIME:
```
tcaiger, line 80: Nice update!

tcaiger (approved):
This is honestly some beautiful code. There is so much going on but it's so well organised.

PR #5034, opened 2023-09-28
No issue: add test runner to datatrak-web

tcaiger:
### Changes:

- Configure jest
- Setup a custom render method helper which includes app providers and router
- Add basic survey render test


packages/datatrak-web/package.json
```diff
@@ -55,7 +55,8 @@
     "start-web-config-server": "yarn workspace @tupaia/web-config-server start-dev",
     "start-entity-server": "yarn workspace @tupaia/entity-server start-dev",
     "start-servers": "npm-run-all -c -l -p start-central-server start-web-config-server start-datatrak-web-server start-entity-server",
-    "test": "yarn package:test"
+    "test": "yarn test:jest",
+    "test:jest": "yarn package:test --env=jsdom"
```
acdunham, line 59: Any reason we need both of these?

packages/datatrak-web/src/api/api.ts
```diff
@@ -6,7 +6,8 @@
 import axios from 'axios';
 import FetchError from './fetchError';
 
-export const API_URL = import.meta.env.REACT_APP_DATATRAK_WEB_API_URL || 'http://localhost:8110/v1';
+// Needs to use process.env instead of import.meta.env for compatibility with jest
+export const API_URL = process.env.REACT_APP_DATATRAK_WEB_API_URL || 'http://localhost:8110/v1';
```
acdunham, line 10: Will this still work when running the app ?

acdunham (approved):
Looks great! I have a couple of questions, but otherwise pre-approving

packages/datatrak-web/package.json
tcaiger: I just copied what was in the package.json for psss. I think that it is a convention because I've seen it in some npm modules around the place. Maybe so we organise scripts for other types of test such as cypress

packages/datatrak-web/src/api/api.ts
tcaiger: Yes I double checked. Also this is what we do in psss which is re-assuring. I think it was you who added capability for this, right? By adding `loadEnv` in the vite config.

PR #5035, opened 2023-09-28
WAITP-1235: matrix qa fixes

acdunham:
### Changes:
- fixed issue where drilldown crashed matrix
- fixed incorrect order of matrix columns on flippa table
- fixed extra entity showing in project level overlay table


tcaiger (approved):
Looks good to me

PR #5037, opened 2023-09-28
WAITP-1221: Map overlay fixes

acdunham:
### Changes:
- Fixed issue where duplicated entities appeared on the map
- Fixed issue where hidden measures still appeared
- Fixed points appearing when the selected overlay is not a point type overlay
- Handle array type measure levels (likely to be deprecated later though)


tcaiger (approved):
LGTM - thanks

PR #5038, opened 2023-09-29
WAITP-1287: Misc FE fixes

acdunham:
### Changes:
- Hide static text hover effect when enlargable
- Fix height of dashboard menu to allow for branch name banner
- Fix sort of dashboard items to use `name` instead of `code` after `sort_order`
- Don't display single value name if is not `multiSingleValue`
- Add disabled button state to download files viz
- Fix loading state of map overlay selector when there are no overlays for an entity
- Fix issue with missing table rows when only some table rows are grouped
- Fix issue with missing search result at end of list

### DIFF
--- a/packages/tupaia-web/src/features/Dashboard/DashboardMenu.tsx
+++ b/packages/tupaia-web/src/features/Dashboard/DashboardMenu.tsx
@@ -34,7 +34,7 @@ const ItemButton = styled(Menu)`
   }
   .MuiMenu-paper {
     max-height: calc(
-      100vh - ${TOP_BAR_HEIGHT} + ${TOP_BAR_HEIGHT}
+      100vh - (${TOP_BAR_HEIGHT} + ${TOP_BAR_HEIGHT})
     ); // 2x top bar height, to make up for any possibly extra in header, e.g. the branch name banner
   }
 
--- a/packages/tupaia-web/src/features/Dashboard/DashboardMenu.tsx
+++ b/packages/tupaia-web/src/features/Dashboard/DashboardMenu.tsx
@@ -34,7 +34,7 @@ const ItemButton = styled(Menu)`
   }
   .MuiMenu-paper {
     max-height: calc(
-      100vh - ${TOP_BAR_HEIGHT * 2}
+      100vh - ${TOP_BAR_HEIGHT} + ${TOP_BAR_HEIGHT}
     ); // 2x top bar height, to make up for any possibly extra in header, e.g. the branch name banner
   }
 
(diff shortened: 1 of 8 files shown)


packages/tupaia-web/src/features/Dashboard/DashboardMenu.tsx
```diff
@@ -33,7 +33,9 @@ const ItemButton = styled(Menu)`
     background: ${({ theme }) => theme.palette.background.default};
   }
   .MuiMenu-paper {
-    max-height: calc(100vh - ${TOP_BAR_HEIGHT});
+    max-height: calc(
+      100vh - ${TOP_BAR_HEIGHT} + ${TOP_BAR_HEIGHT}
```
tcaiger, line 37: I assume that you tested this and it works but I'm surprised calc works this way. Shouldn't it be either `100vh - ${TOP_BAR_HEIGHT} - ${TOP_BAR_HEIGHT}` or `100vh - (${TOP_BAR_HEIGHT} + ${TOP_BAR_HEIGHT})`?

packages/tupaia-web/src/features/EntitySearch/SearchResults.tsx
```diff
@@ -105,8 +105,8 @@ export const SearchResults = ({ searchValue, onClose }: SearchResultsProps) => {
   };
 
   const showLoadMoreButton = resultsCount > pageSize;
-  // Remove the extra result
-  const displayResults = searchResults.slice(0, -1);
+  // Remove the extra result only when the load more button is shown, otherwise the last result gets chopped off when there are no more results to load
+  const displayResults = showLoadMoreButton ? searchResults.slice(0, -1) : searchResults;
```
tcaiger, line 109: Woops that is totally my silly error. Thanks for fixing.

packages/tupaia-web/src/features/Map/utils/useMapOverlayTableData.ts
```diff
@@ -86,7 +86,7 @@ export const useMapOverlayTableData = ({
 
   return {
     ...data,
-    isLoading: isLoading || isFetching || !isFetched,
+    isLoading: isLoading || isFetching || (!isFetched && !isIdle),
```
tcaiger, line 89: I find these loading states are a bit frustrating. I would just expec isLoading to include all of these cases (I think it might in the latest version). I wonder if we should try and extend and modify the default loading logic somehow to do this. Maybe not a task for now though.

tcaiger (approved):
Thanks for all the fixes. Just the one question but I'm happy to pre-approve.

PR #5040, opened 2023-10-01
WAITP-1429: Autocomplete add and filtering of options

acdunham:
### Changes:
- Enabled autocomplete attribute filtering
- Enabled autocomplete option creation
- Added tests for `AutocompleteQuestion` and `useSubmitSurvey`


packages/datatrak-web/src/api/mutations/useSubmitSurvey.ts
```diff
@@ -55,35 +82,54 @@ export const processSurveyResponse = ({
...
+        surveyResponse.entity_id = answer as string;
         break;
       }
-      default:
+      case 'Autocomplete': {
+        // if the answer is a new option, add it to the options_created array to be added to the DB
+        const { isNew, value, label, optionSetId } = answer as AutocompleteAnswer;
+        if (isNew) {
+          surveyResponse.options_created.push({
+            option_set_id: optionSetId,
```
tcaiger, line 111: I noticed that on Meditrak the optionsCreated have an id and a sortOrder, does it need them here too?

tcaiger (approved):
Looks great. I just left one question just in case.

packages/datatrak-web/src/api/mutations/useSubmitSurvey.ts
acdunham: I don't think so - it automatically adds the sort order as the the next number (to make the option the last one in the list) and a unique id gets applied when it's created

tcaiger:
Oh yeah I just noticed that the base branch is still pointing at the old one. Ethan made a new one yesterday which is `datatrak-web-epic`

PR #5044, opened 2023-10-03
WAITP-1416: Visibility criteria

acdunham:
### Changes:
- Added filtering to questions and screens based on visibility criteria
- Added ability to save values to formData when questions depend on answers being updated
- Removed logic around changing question numbers based on screens with only instructions, because the question numbers don't match the side menu numbers in this case
- Tidied up context methods to try make more readable
- Changed survey components to be an array of arrays in the FE so is easier to work with visibility
- Added logic to redirect the user to the first survey question if they attempt to access a survey screen that is not available
- Added logic to reset the survey questions that have changed to being invisible


tcaiger (approved):
Looks really nice - nothing from me :)

PR #5045, opened 2023-10-04
WAITP-1424: Entity question upserts

tcaiger:
### Changes:

- Add SubmitSurveyRoute to datatrak-web-server
- Add SubmitSurveyRequest types to types package
- Handle preparing entities_upserted in the FE form submission handler
- Handle creating Entity records for upsert in the SubmitSurveyRequest

Note: I need to add a test for the new server endpoint but I will do that in a follow up PR since it's the first test in the package and might require a bit of setup.


packages/datatrak-web/src/__tests__/features/Questions/AutocompleteQuestion.test.tsx
```diff
@@ -66,21 +66,21 @@ describe('Autocomplete Question', () => {
...
 
-  it('renders all the options when there is no attribute filtering involved', () => {
+  it('renders all the options when there is no attribute filtering involved', async () => {
     renderComponent(<AutocompleteQuestion {...props} />);
 
     const openButton = screen.getByTitle('Open');
     openButton.click();
 
-    const displayOptions = screen.getAllByRole('option');
+    const displayOptions = await screen.findAllByRole('option');
```
tcaiger, line 75: I was getting inconsistent results on the test where sometimes it was loading results and sometimes not so adding this await seems to make it consistent.

packages/datatrak-web/src/api/mutations/submitSurvey/useSubmitSurvey.ts
```diff
@@ -0,0 +1,59 @@
+/*
+ * Tupaia
```
tcaiger, line 2: I split up this file as it was getting pretty big and doing a few different things

packages/datatrak-web/src/features/Questions/EntityQuestion/utils.ts
```diff
@@ -4,14 +4,22 @@
  */
 import { useUser } from '../../../api/queries';
 import { useSurveyForm } from '../../Survey/SurveyContext';
+import { DatatrakWebSubmitSurveyRequest as RequestT } from '@tupaia/types';
 
-export const useEntityBaseFilters = config => {
+export const useEntityBaseFilters = (config: RequestT.EntityQuestionConfig) => {
```
tcaiger, line 9: This type for the EntityQuestionConfig was initially in the front-end but I wanted to use it in the server to so I moved it to the request type. I'm still not sure if that's the correct place for it.

packages/types/src/types/requests/central-server/MeditrakSurveyResponseRequest.ts
```diff
@@ -41,7 +41,7 @@ type OptionCreated = {
 };
 
 export interface MeditrakSurveyResponseRequest {
-  id: Id;
+  id?: Id;
```
tcaiger, line 44: In the surveyResponses endpoint this is actually an optional param

packages/datatrak-web-server/src/routes/SubmitSurveyRoute.ts
```diff
@@ -0,0 +1,37 @@
...
+  RequestT.ResBody,
+  RequestT.ReqBody,
+  RequestT.ReqQuery
+>;
+
+export class SubmitSurveyRoute extends Route<SubmitSurveyRequest> {
+  public async buildResponse() {
+    const { central: centralApi } = this.req.ctx.services;
+    const surveyResponseData = this.req.body;
+    const { answers, entities_upserted, country_id } = surveyResponseData;
```
acdunham, line 21: Can you possibly check that `options_created` still gets handled for the autocomplete questions?

Also the other thing is, we do a lot of processing of the response on the FE, so potentially we could move that into here if you wanted to keep this endpoint, and do the converting of the formData object into the expected shape. Then we don't have half the processing in the FE and half in the BE

packages/datatrak-web-server/src/utils/createUpsertEntityObjects.ts
```diff
@@ -0,0 +1,78 @@
...
+  const entity = { id: entityId } as Entity;
+  const fields = config?.entity?.fields || {};
+
+  for (const [fieldName, value] of Object.entries(fields)) {
+    // Value is not defined, skip
+    if (value === undefined) {
+      return;
+    }
+
+    // const fieldValue = typeof value === 'string' ? value : answers[value.questionId];
```
acdunham, line 30:
```suggestion
```

packages/datatrak-web/src/features/Questions/EntityQuestion/utils.ts
acdunham: I think that's fine, and probably the most sensible approach

acdunham (changes requested):
This is great! I know we haven't really been doing tests in the orchestration server, but potentially we should add some for the entity handling util since it is quite encapsulated and is a good candidate for this. I also wonder about whether the processing of the survey response should all be moved to the BE since it's already doing half of it there.

packages/datatrak-web/src/api/mutations/useSubmitSurvey.ts
```diff
@@ -20,11 +30,11 @@ export const useSurveyResponseData = () => {
   const { surveyStartTime, surveyScreenComponents } = useSurveyForm();
   const { data: survey } = useSurvey(surveyCode);
   return {
-    surveyStartTime,
+    startTime: surveyStartTime,
     surveyId: survey?.id,
     questions: Object.values(surveyScreenComponents!).reduce((acc, val) => acc.concat(val), []), // flattened array of survey questions
     countryId: user?.country?.id,
-    useId: user?.id,
```
acdunham: Woops

acdunham (approved):
I think this makes way more sense, nice work! Thanks for doing this

packages/datatrak-web-server/src/utils/processSurveyResponse.ts
```diff
@@ -0,0 +1,125 @@
...
+    timezone,
+    entities_upserted: [],
+    options_created: [],
+    answers: [],
+  } as MeditrakSurveyResponseRequest;
+  // Process answers and save the response in the database
+  const answersToSubmit = [] as Record<string, unknown>[];
+
+  for (const question of questions) {
+    const { questionId, questionType, id } = question;
```
EMcQ-BES, line 57: I really dislike that `Question.id` isn't an id in the `question` table, but `Question.questionId` is. That feels like it's just inviting mistakes down the line.

packages/types/src/types/requests/datatrak-web-server/SubmitSurveyRequest.ts
```diff
@@ -0,0 +1,33 @@
...
+  optionSetId: string;
+  value: string;
+  label: string;
+};
+
+type Answer = string | number | boolean | null | undefined | AutocompleteAnswer;
+
+export type Answers = Record<string, Answer>;
+
+type Question = SurveyScreenComponent;
```
EMcQ-BES, line 19: See my other comment about `Question.questionId`

I think I'd generally prefer to keep the type name relative to the related db table, even if that means we have something that looks a little weird like `questions: SurveyScreenComponents[]`

Or possibly we could use a mixed name like `ScreenQuestion` just to make it a bit more clear that this isn't directly related to the db question type

EMcQ-BES (commented):
Core logic of the backend looks good 👍 Minor comment about the naming of the `Question` type, interested to hear if you have any counterpoints about that

packages/datatrak-web-server/src/utils/processSurveyResponse.ts
tcaiger: Yes that's a good point. There's actually a componentId for the id but I realised when I reviewed this that the id and the questionId are the same field and that the question if should not be getting sent with the answer so I have removed reference to it completely!

packages/types/src/types/requests/datatrak-web-server/SubmitSurveyRequest.ts
tcaiger: Yes fair point. From looking through the types definitions there is a convention to use SurveyScreenComponent as the type name so I've just updated it here to be that. I will think about adding a new naming convention in a later PR as I don't want this PR to drag on too much more.

PR #5046, opened 2023-10-05
WAITP-1366 Update dashboard permissions

EMcQ-BES:
### Changes:

- Allow fetching `dashboardRelations` from the db to skip permissions checking
  - This is because we use the difference between `relations` and `items` to decide whether we show the "No access" or "No data" default dashboard
- Attach access policy to req object in `tupaia-web-server`
- Filter dashboard items by `accessPolicy`
  - `central-server` checks you can access the item in _any_ of its countries, so we need a more specific check against the requested entity to make sure we're not showing you dashboard items that you can't access data for


tcaiger (approved):
Looks good to me. Thanks for the fix

PR #5047, opened 2023-10-05
WAITP-1347: Validation criteria

acdunham:
### Changes:
- Added validation logic to question inputs
- Added logic to send user back to first error screen if validation happens on review screen
- Added styling for error states for all question inputs
- Added error message when applicable

### Screenshots:
Radio error
Screenshot 2023-10-05 at 2 30 07 pm

Text error
Screenshot 2023-10-05 at 2 30 39 pm

Entity error
Screenshot 2023-10-05 at 2 30 54 pm

Date error
Screenshot 2023-10-05 at 2 31 27 pm

Max number error
Screenshot 2023-10-05 at 2 32 22 pm

Autocomplete error
Screenshot 2023-10-05 at 2 47 05 pm

Checkbox error
Screenshot 2023-10-05 at 2 50 45 pm

Geolocate error
Screenshot 2023-10-05 at 2 53 01 pm


packages/datatrak-web/src/features/Survey/SurveyLayout.tsx
```diff
@@ -145,10 +147,50 @@ export const SurveyLayout = () => {
     handleStep(path, data);
   };
 
-  const handleSubmitScreen = handleSubmit(data => {
+  const onError = errors => {
+    // If we're not on the review screen, we don't need to do anything, because the errors get focussed and handled by react-hook-form
```
tcaiger, line 151: great comment

packages/datatrak-web/src/features/Survey/Components/SurveyQuestionGroup.tsx
```diff
@@ -26,6 +28,32 @@ const QuestionNumber = styled(Typography)`
  * This is the component that renders questions in a survey.
  */
 export const SurveyQuestionGroup = ({ questions }: { questions: SurveyScreenComponent[] }) => {
+  const { setError, errors } = useFormContext();
+  const location = useLocation() as { state: { errors?: string } };
```
tcaiger, line 32: Good solution using the location state to send the errors between screens!

packages/datatrak-web/src/features/Survey/Components/SurveyQuestion.tsx
```diff
@@ -70,44 +78,100 @@ export const SurveyQuestion = ({
...
     return <QuestionPlaceholder>{name}</QuestionPlaceholder>;
   }
 
+  const controllerName = getNameForController(name, type);
+
   // If the question dictates the visibility of any other questions, we need to update the formData when the value changes, so the visibility of other questions can be updated in real time. This doesn't happen that often, so it shouldn't have too much of a performance impact, and we are only updating the formData for the question that is changing, not the entire formData object.
   const handleOnChange = e => {
     if (updateFormDataOnChange) {
-      setSingleAnswer(name, e.target.value);
+      setFormData({
```
tcaiger, line 97: What made you decide to take this out of the context reducer?

acdunham: I realised we didn't actually need it, because we the `setFormData` can handle an object with one key, and then working on the condition question type I ended up needing to do more updates to the data as anything changed, and didn't want to call the same thing in multiple places and potentially end up with bugs. And since I was removing it that branch I just did it here too so I didn't get conflicts

packages/datatrak-web/src/features/Survey/Components/SurveyQuestion.tsx
```diff
@@ -70,44 +78,100 @@ export const SurveyQuestion = ({
...
+        [name]: e.target ? e.target.value : e.value, // autocomplete event returns an object with value, others return an event with a target
+      });
     }
   };
 
+  const { mandatory: required, min, max } = validationCriteria;
+
+  const getRules = () => {
```
tcaiger, line 105: I think this point is pretty subjective so there no action required but I have a slight preference for defining these logic heavy functions outside of the react component because I find it easier to read. For example this function is just dependant on type but that is not very obvious when it is inline. 

**A couple of other minor pros to pulling them out**
It can also make testing these functions in isolation easy (not that I do that very often yet...).
Some people argue that it's more performant (although I don't give much weight to that)

There are definitely cases where I don't like extracting them out which is when there are lots of arguments (more than 1 or 2) that would need to be passed in. 

Like I said I don't feel too strongly either way but just wanted to mention it in case you agree.

PR #5049, opened 2023-10-05
WAITP-1430: condition question

acdunham:
### Changes:
- Added support for conditional question, including determining when to update questions on change, if the question determines the outcome of a condition question
- Added logic for calculating condition questions
- Added 'read only' question component for use with condition and arithmetic questions

---

### Screenshots:
Empty
Screenshot 2023-10-06 at 9 00 37 am

Filled in
Screenshot 2023-10-06 at 9 00 13 am


tcaiger (approved):
Everything looks good. I just have a general comment that it would be good to split up the SurveyContext file soon as it's getting big and intimidating so let's look at that next time we are adding something to that file.

PR #5050, opened 2023-10-06
No issue: Update survey screen fetching

acdunham:
### Changes:
- Updated survey endpoint in central-server to include relevant question fields
- Updated single survey endpoint in datatrak-web-server to format the data more appropriately
- Updated datatrak-web to use this endpoint instead of survey screen components endpoint
- Removed survey screen components endpoint from datatrak-web-server
- Updated types and tests


packages/types/src/types/requests/datatrak-web-server/SurveyRequest.ts
```diff
@@ -0,0 +1,83 @@
...
+export type AutocompleteConfig = {
+  createNew?: boolean;
+  attributes?: {
+    [key: string]: { questionId: Question['id'] };
+  };
+};
+
+type Config = {
+  codeGenerator?: CodeGeneratorConfig;
+  autocomplete?: AutocompleteConfig;
```
tcaiger, line 54: It would be great to add the entity config here too which I currently have in the survey submit types but I'm happy to do that after we have merged our two PRs together.

tcaiger (approved):
wow this ended up being a pretty big refactor but I think it makes more sense and I like that you handled transforming the survey data in the orchestration server

PR #5051, opened 2023-10-06
RN-1042 Fix media perm Meditrak App

IgorNadj:
### Issue #: RN-1042

Fix in fork: https://github.com/beyondessential/react-native-image-picker/commit/f789975563e09443d03ec6c50fb67bf0fd2522bd

### DIFF
--- a/packages/meditrak-app/app/main.js
+++ b/packages/meditrak-app/app/main.js
@@ -4,7 +4,7 @@
  */
 import 'react-native-gesture-handler';
 import React from 'react';
-import { AppRegistry } from 'react-native';
+import { AppRegistry, LogBox } from 'react-native';
 import AsyncStorage from '@react-native-async-storage/async-storage';
 import { Provider } from 'react-redux';
 import { createStore, compose } from 'redux';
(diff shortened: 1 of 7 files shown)


packages/meditrak-app/app/main.js
```diff
@@ -72,3 +72,6 @@ const App = () => (
 );
 
 AppRegistry.registerComponent('TupaiaMediTrak', () => App);
+
+LogBox.ignoreLogs(['Warning: ...']); // Ignore log notification by message
+LogBox.ignoreAllLogs(); // Ignore all log notifications
```
rohan-bes, line 77: Is the annoying yellow box a new thing?

rohan-bes (approved):
Looks good :+1: Thanks for solving this one. It seems it will, but just double checking this is backwards compatible?

packages/meditrak-app/app/main.js
IgorNadj: This thing at the bottom, just gives PropType warnings, and perfectly covers our survey screen navigation 😆 
image

IgorNadj:
> just double checking this is backwards compatible?

Yep, I have tested and we are doing normal testing on this as well

IgorNadj:
This has gone into master in a separate PR and then into dev with a master -> dev PR

PR #5052, opened 2023-10-08
WAITP-1453: UI updates

acdunham:
### Changes:
- Add tooltip delay of 1 second
- Update country select hover color, and outline
- Update drop shadow on all popover menus
- Update header sizes
- Make width of number fields smaller
- Make text fields become multiline when text overflows the width
- Make radio labels full width
- Make side menu instructions not have numbers, and screens match
- Clip side menu text
- Vertically centre instruction text if no other questions on screen
- Remove placeholder text when on review screen
- Update survey icons
- Handle when location access is denied on geolocate question
- Fix redirect issue for request project access view
---

### Screenshots:
Screenshot 2023-10-09 at 11 06 30 am

Screenshot 2023-10-09 at 11 06 38 am

Screenshot 2023-10-09 at 11 07 03 am

Screenshot 2023-10-09 at 11 07 14 am

Screenshot 2023-10-09 at 11 07 26 am

Screenshot 2023-10-09 at 11 07 34 am

Screenshot 2023-10-09 at 11 07 48 am

Screenshot 2023-10-09 at 11 08 09 am


packages/datatrak-web/src/views/SurveySelectPage/SurveyCountrySelector.tsx
```diff
@@ -51,6 +55,7 @@ export const SurveyCountrySelector = ({
   return (
     <CountrySelectWrapper>
       <Pin />
+      {/** @ts-ignore */}
```
tcaiger, line 58: Is this intentionally left here? If so we should add a comment?

packages/ui-components/src/components/Tooltip.tsx
```diff
@@ -13,14 +13,19 @@ const TOOLTIP_COLOR = '#002d47';
...
-export const Tooltip = styled((props: TooltipProps) => (
-  <MuiTooltip
-    {...props}
-    classes={{ popper: props.className }}
-    placement={props.placement || 'top'}
-    arrow
-  />
-))`
+export const Tooltip = styled(
+  ({ className, placement = 'top', enterDelay = 1000, ...props }: TooltipProps) => (
```
tcaiger, line 17: this will add a delay everywhere including tupaia-web which I'm not sure we want. The tooltip is only used in two places in datatrak-web so can we just set the delay as a constant in there and pass it into here?

packages/datatrak-web/src/components/Icons/SurveyFolderIcon.tsx
```diff
@@ -0,0 +1,34 @@
...
+      />
+      <rect y="1.84766" width="12" height="10.1538" rx="2.49969" fill="#D9E8F8" />
+      <line
+        x1="3.26917"
+        y1="7.57843"
+        x2="8.9616"
+        y2="7.57843"
+        stroke="#328DE5"
+        stroke-width="0.999878"
+        stroke-linecap="round"
```
tcaiger, line 30:
```suggestion
        strokeWidth="0.999878"
        strokeLinecap="round"
```

tcaiger:
This is not related to these changes but would you be able to look at this bug
http://localhost:5173/survey/AA/1
image

packages/datatrak-web/src/features/Survey/Components/SurveySideMenu/SurveySideMenu.tsx
```diff
@@ -62,6 +65,15 @@ const SurveyMenuItem = styled(ListItem).attrs({
   &:hover {
     background-color: ${({ theme }) => `${theme.palette.grey[400]}4d`};
   }
+  .MuiButton-label {
+    > span:not(.MuiTouchRipple-root) {
+      display: -webkit-box;
+      -webkit-box-orient: vertical;
+      line-clamp: ${({ $isInstructionOnly }) => ($isInstructionOnly ? 1 : 2)};
+      -webkit-line-clamp: ${({ $isInstructionOnly }) => ($isInstructionOnly ? 1 : 2)};
+      overflow: hidden;
```
tcaiger, line 74:
```suggestion
    text-overflow: ellipsis;
    overflow: hidden;
    white-space: nowrap;
    padding-right: 2rem;
```
-webkit-box will only work on chrome, right? Any reason why we can't use text-overflow?

tcaiger (changes requested):
Thanks for all the changes. Just a couple of minor things from me.

packages/datatrak-web/src/features/Survey/Components/SurveySideMenu/SurveySideMenu.tsx
acdunham: This actually works in all modern browsers (since ~2018 release date) according to `caniuse.com`, just not old IE which is deprecated now anyway.
The `line-clamp` approach is preferred in this case because you don't have to calculate height based on line-height and so on, it just automatically does the calculation. Your suggested approach would be okay for 1 line clamping, but as soon as it gets to be more than that it becomes more complicated to work out than just changing a number

tcaiger (approved):
Tino pai!

PR #5054, opened 2023-10-09
WAITP-1412: Arithmetic question

acdunham:
### Changes:
- Refactored survey context to be in multiple smaller files
- Added support for updating arithmetic question when related questions are updated
- Added display of arithmetic question the same as condition questions

---

### Screenshots:
Screenshot 2023-10-09 at 2 13 19 pm


tcaiger (approved):
looks great - thanks for splitting up that context

PR #5058, opened 2023-10-10
no-issue: Added 'unique' function to the viz-builder

rohan-bes:
Would be handy when you want all the unique values in a column (eg. when fetching data element names based on the data elements being used)


EMcQ-BES (approved):
LGTM

PR #5059, opened 2023-10-10
No Issue: Use handleError correctly with the new auth middleware

EMcQ-BES:
### Changes:

- Run the `handleError` middleware after the `auth` middleware
  - I believe because the existing `handleError` in within the scope of the `apiV2` express router, errors from this layer weren't hitting it correctly
  - The error response is used to signal to the `web-config-server` to refresh the access token, so without the errors returning correctly, your access token stayed expired for extended periods and broke normal behaviour


tcaiger (approved):
Nice

PR #5060, opened 2023-10-10
WAITP-1221: tupaia web front end fixes

tcaiger:
### Changes:
- Fix issue where shaded-spectrum map overlays were not displaying at facility level
- Fix dashboard scroll glitch
- Fix entity search dropdown z index

### DIFF
--- a/packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
+++ b/packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
@@ -53,9 +53,8 @@ const ScrollBody = styled.div`
   position: relative;
   height: 100%;
   @media screen and (min-width: ${MOBILE_BREAKPOINT}) {
-    overflow: auto;
-    // Fix issue where scroll bar appears and dis-appears repeatedly and flickers
-    margin-right: 3px;
+    // Need to be permanently set to scroll to fix issue where scroll bar appears and dis-appears repeatedly and flickers
+    overflow-y: scroll;
   }
 `;
 
--- a/packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
+++ b/packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
@@ -54,7 +54,7 @@ const ScrollBody = styled.div`
   height: 100%;
   @media screen and (min-width: ${MOBILE_BREAKPOINT}) {
     // Need to be permanently set to scroll to fix issue where scroll bar appears and dis-appears repeatedly and flickers
-    overflow-y: scroll;
+    overflow: auto;
   }
 `;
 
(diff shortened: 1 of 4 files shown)


packages/tupaia-web/src/features/Dashboard/Dashboard.tsx
```diff
@@ -53,9 +53,8 @@ const ScrollBody = styled.div`
   position: relative;
   height: 100%;
   @media screen and (min-width: ${MOBILE_BREAKPOINT}) {
-    overflow: auto;
-    // Fix issue where scroll bar appears and dis-appears repeatedly and flickers
-    margin-right: 3px;
+    // Need to be permanently set to scroll to fix issue where scroll bar appears and dis-appears repeatedly and flickers
```
acdunham, line 56: My only question around this would be, if there are cases where scroll isn't actually needed, e.g. 1 dashboard item only, then the scroll would still show right? Would be nicer if we didn't need to do that, I think

packages/tupaia-web/src/features/Map/MapOverlaysLayer/PolygonLayer.tsx
```diff
@@ -36,13 +35,10 @@ const ActivePolygon = styled(Polygon)<PolygonProps>`
...
 const ShadedPolygon = styled(BasePolygon)<PolygonProps>`
   fill: ${({ $shade }) => $shade};
-  stroke: ${({ $shade, $active }) => ($active ? POLYGON_HIGHLIGHT : $shade)};
+  stroke: ${({ $shade }) => $shade};
   fill-opacity: 0.5;
   &:hover {
-    fill-opacity: ${({ $active }) =>
-      $active
-        ? 0.5
-        : 0.8}; // don't add hover effect when it is the active polygon because it isn't clickable
```
acdunham, line 45: I actually think I remember why this was there - sometimes the active polygon would need to display as a shaded polygon, which it does, from the `getDisplayType` method. But in this case, the active polygon shouldn't be clickable, so removing the hover effect makes sense. Depends if you mind that an unclickable entity has a hover effect or not though?

packages/tupaia-web/src/layout/TopBar/TopBar.tsx
```diff
@@ -27,7 +27,8 @@ const Header = styled.header<{
   display: flex;
   justify-content: space-between;
   align-items: center;
-  z-index: 1000;
+  // Needs to be higher than 1000 as leaflet sets z-index of map controls to 1000
+  z-index: 1100;
```
acdunham, line 31: Classic z-index wars

acdunham (approved):
Thanks for doing these. I have a couple of questions, but pre-approving

packages/tupaia-web/src/features/Map/MapOverlaysLayer/PolygonLayer.tsx
tcaiger: oh yeah that's a good idea, I'll put the change back. I should have read the comment more carefully :)

PR #5061, opened 2023-10-10
RN-502: Remove $async field from schemas

rohan-bes:
### Issue RN-502 (task):

We have been adding an `$async` key to our model schemas if we want to support custom asynchronous validation. This was only being done on survey responses, to ensure that all questionIds in the `answers` object matched actual question ids. We decided to scrap this functionality as it's a bit niche and makes validation more complex than it needs to be.

Will add a note to testers to double check what happens if we try to submit a survey response for a question that doesn't exist (eg. it got deleted before the response is synced)


packages/data-table-server/src/middleware/attachDataTableFromPreviewToContext.ts
```diff
@@ -4,24 +4,18 @@
...
-  if (!validate && dataTableValidator.errors) {
-    throw new Error(dataTableValidator.errors[0].message);
-  }
-
-  return dataTableFields;
+const validateDataTableFields = (dataTableFields: unknown) => {
+  const { properties: schemaProperties } = DataTablePreviewRequestSchema;
+  const { id, ...propertiesWithoutId } = schemaProperties;
+  const schemaWithoutId = { ...DataTablePreviewRequestSchema, properties: propertiesWithoutId };
+  return ajvValidate<Omit<DataTablePreviewRequest, 'id'>>(schemaWithoutId, dataTableFields);
```
rohan-bes, line 18: Felt the previous dance around of adding a fake id to get past the validator was slightly more awkward than stripping the id from the schema...

packages/meditrak-app-server/src/__tests__/__integration__/sync/PushChangesRoute.test.ts
```diff
@@ -363,26 +363,6 @@ describe('changes (POST)', () => {
     });
 
     describe('Reject if invalid', () => {
-      it('rejects if answer provide non-exist question_id', async () => {
```
rohan-bes, line 366: This test is no longer enforced. At the meditrak-app-server level, if a survey response comes in with a question_id that doesn't exist we'll still pass it through to the central-server.

At create time this will fail due to foreign key constraint. All survey responses are created within a transaction so no responses will be saved.

I think I'm okay with this. But happy to add a piece of validation to check all the questions exist.

packages/data-table-server/src/middleware/attachDataTableFromPreviewToContext.ts
IgorNadj: I'm not sure this is possible. A json schema can represent A | B, with anyOf: https://json-schema.org/understanding-json-schema/reference/combining. The id property may not exist top level.

rohan-bes: > I was going to suggest in the general case where if want to CREATE/EDIT a model and we don't have an id, we should update our generation script to add WithoutId variants of each db model e.g. EntityWithoutId = Omit<Entity, 'id'>. So we would get that schema for free and across all models.

I like that, I can see those types coming in handy pretty often :+1: 

But you're also right that this request type is just wrong, I'll fix that now

IgorNadj (approved):
Nice one!

PR #5062, opened 2023-10-10
RN-502: Made timestamp and optional field of a survey response

rohan-bes:

### DIFF
--- a/packages/central-server/src/apiV2/surveyResponse.js
+++ b/packages/central-server/src/apiV2/surveyResponse.js
@@ -29,7 +29,6 @@ const createSurveyResponseValidator = models =>
     survey_id: [hasContent, constructRecordExistsWithId(models.survey)],
     start_time: [constructIsEmptyOr(takesDateForm)],
     end_time: [constructIsEmptyOr(takesDateForm)],
-    timestamp: [constructIsEmptyOr(takesDateForm)],
   });
 
 async function validateResponse(models, body) {
--- a/packages/central-server/src/apiV2/surveyResponse.js
+++ b/packages/central-server/src/apiV2/surveyResponse.js
@@ -29,6 +29,7 @@ const createSurveyResponseValidator = models =>
     survey_id: [hasContent, constructRecordExistsWithId(models.survey)],
     start_time: [constructIsEmptyOr(takesDateForm)],
     end_time: [constructIsEmptyOr(takesDateForm)],
+    timestamp: [constructIsEmptyOr(takesDateForm)],
   });
 
 async function validateResponse(models, body) {
--- a/packages/central-server/src/apiV2/surveyResponse.js
+++ b/packages/central-server/src/apiV2/surveyResponse.js
@@ -29,7 +29,6 @@ const createSurveyResponseValidator = models =>
     survey_id: [hasContent, constructRecordExistsWithId(models.survey)],
     start_time: [constructIsEmptyOr(takesDateForm)],
     end_time: [constructIsEmptyOr(takesDateForm)],
-    timestamp: [hasContent],
   });
 
 async function validateResponse(models, body) {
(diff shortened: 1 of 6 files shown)


packages/central-server/src/apiV2/surveyResponses/saveResponsesToDatabase.js
```diff
@@ -53,21 +53,29 @@ function buildResponseRecord(user, entitiesByCode, body) {
...
+  const defaultToTimestampOrThrow = (value, parameterName) => {
+    if (value) return new Date(value).toISOString();
+    if (timestamp) return new Date(timestamp).toISOString();
+
+    throw new ValidationError(`Must provide ${parameterName} or timestamp`);
+  };
+
+  const startTime = defaultToTimestampOrThrow(inputStartTime, 'start_time');
+  const endTime = defaultToTimestampOrThrow(inputEndTime, 'end_time');
+  const dataTime = stripTimezoneFromDate(defaultToTimestampOrThrow(inputDataTime, 'data_time'));
```
rohan-bes, line 78: Currently meditrak-app used the `/changes` route to submit data. That route does not require a `timestamp` field.
Once we switch to the meditrak-app-server, we will be calling the `/surveyResponse` route which does require a timestamp.

The timestamp is only needed if we are missing values for any of the following: `start_time`, `end_time`,`data_time`,`timezone`

This change will support a missing timestamp if all those values are present.

packages/central-server/src/apiV2/surveyResponse.js
```diff
@@ -29,7 +29,6 @@ const createSurveyResponseValidator = models =>
     survey_id: [hasContent, constructRecordExistsWithId(models.survey)],
     start_time: [constructIsEmptyOr(takesDateForm)],
     end_time: [constructIsEmptyOr(takesDateForm)],
-    timestamp: [hasContent],
```
IgorNadj, line 32: constructIsEmptyOr(takesDateForm) ?

packages/central-server/src/apiV2/surveyResponses/saveResponsesToDatabase.js
IgorNadj: Looks good, thanks for cleaning this up.

Happy as is, my only suggestion for making this a little bit clearer would be a big if-else, with the main logic first and then the fallback timestamp based logic after, with a comment that this is legacy.

IgorNadj (approved):

packages/central-server/src/apiV2/surveyResponse.js
rohan-bes: Good call, will do :+1:

packages/central-server/src/apiV2/surveyResponses/saveResponsesToDatabase.js
rohan-bes: I'm not certain this is legacy or not. I think it's a valid for each of these fields to either provide them or rely on the `timestamp` field. It does seem a little odd though, doesn't it.

IgorNadj: Hmm odd, yeah either way a big ol comment there would be helpful :)

packages/central-server/src/apiV2/surveyResponse.js
rohan-bes: Discovered that this value can actually be a iso-string or a number. Did a bit of a spike into working out if we could deprecate it entirely, but I don't think we can. It's used as part of the import logic as well as being called externally by msupply.

I don't think it's possible for our ObjectValidator to do a `isANumber OR takesDateForm` so I might just leave it with no validation

IgorNadj: All good!

PR #5067, opened 2023-10-11
WAITP-1418 Permission revoking fixup

EMcQ-BES:
### Changes:

- Add ability to expire sessions for a user from `central-server`
- Trigger an expiry after a `user_entity_permission` change

This should only expire your access token for `tupaia_web`, once the `access_token` expires, the web session will automatically refresh using the `refresh_token` which means you get the updated `access_policy`

### DIFF
--- a/packages/central-server/src/database/models/User.js
+++ b/packages/central-server/src/database/models/User.js
@@ -1,0 +1,32 @@
+/**
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { UserType as CommonUserType, UserModel as CommonUserModel } from '@tupaia/database';
+
+const SERVICES = {
+  // admin_panel: 'admin_panel_session',
+  // datatrak_web: 'datatrak_web_session',
+  tupaia_web: 'tupaia_web_session',
+};
+
+class UserType extends CommonUserType {
+  async expireSessionToken(service) {
+    if (!SERVICES[service]) {
+      throw new Error(`${service} is not a support service for session expiry`);
+    }
+    await this.database.executeSql(
+      `
+        UPDATE ?? SET access_token_expiry = 0 WHERE email = ?
+      `,
+      [SERVICES[service], this.email],
+    );
+  }
+}
+
+export class UserModel extends CommonUserModel {
+  get DatabaseTypeClass() {
+    return UserType;
+  }
+}
--- a/packages/central-server/src/database/models/User.js
+++ b/packages/central-server/src/database/models/User.js
@@ -5,6 +5,10 @@
 
 import { UserType as CommonUserType, UserModel as CommonUserModel } from '@tupaia/database';
 
+// Currently our pattern is that session tables don't have models
+// in the generic database package, this is a quick and dirty way to get
+// context for them into central-server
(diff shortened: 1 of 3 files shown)


rohan-bes (changes requested):
Oooh I really like this approach. Even in the existing web-frontend we get a bit of weird behaviour if a user's permissions change while they're logged in.

Forcing a user to logout and login again after their permissions change seems like the most sensible approach. I think we could apply this generally across all our platforms, not just `tupaia-web` :+1: 

Thoughts on changing this so that it expires the user's session on all platforms? (tupaia-web, lesmis, psss, admin-panel, datatrak-web)

Keeping a hard coded list of the different platforms and their respective session tables does feel a little brittle, but for now I think it's okay. We might actually want to consolidate all those session tables into one table going forward, as they all seem to be following the same schema.

EMcQ-BES:
@rohan-bes 

I might need to update that comment in code because a quirk of this system means you don't _actually_ get logged out of the frontend. The `access_token` expires but since the session has a valid `refresh_token` it will refresh the `access_token` and continue with the users new `accessPolicy`. This means you lose any cached access immediately on your next click, and gain any new access.

Initially when writing this I was expiring the users refresh token as well, which _does_ log them out completely. I found that to be _extremely_ annoying for the admin panel specifically since it's pretty common for devs and testers to adjust their own permissions in the admin panel it interrupts your workflow if you keep getting logged out. So _that_ specifically I'd like to avoid.

Admittedly the list of session tables is a bit smelly. Currently we have a pattern where we don't add session tables to the default database models, so this was just a quick way to get context into the `central-server`... I'm pretty on board with consolidating those session tables into one, but I think that's out of scope for this particular ticket.

rohan-bes:
@EMcQ-BES ah I see. Out of curiosity, how does refreshing the accessToken and accessPolicy play with react-query? If we've cached data on the browser side, does that persist even after we refresh the accessPolicy?

Good point about not wanting to force users out if they change their permissions in the admin-panel, that does seem annoying.

Yep, agreed that consolidating session tables is out of scope :+1: and probably would need a bit of time to investigate if it's really a good idea.

EMcQ-BES:
@rohan-bes Ah good call I'll need to double check the caching behaviour. I'll sort that on monday

EMcQ-BES:
@rohan-bes I can confirm the browser caching does persist after the refresh, but it starts clearing up relatively quickly as you make more requests. Let me know if that's a concern.

I know you asked about expiring the other sessions as well, but I think for right now I'd prefer to keep this restricted to `tupaia-web` since I can't entirely guarantee the behaviour of the other servers with this method 🤔

rohan-bes:
> I can confirm the browser caching does persist after the refresh, but it starts clearing up relatively quickly as you make more requests. Let me know if that's a concern.

Not really a concern. Though I'd also be perfectly happy with the revoking the refresh token and forcing the user out for non-Admin Panel platforms, that seems reasonable. But not too fussed with either.

I definitely would like for us to clean up the specific reference to `tupaia-web` and the session tables. Quite keen to keep the `central-server` as platform agnostic as possible. But agree that we can cover that at another stage, I'll put something in the maintenance sprint items about it. Thanks for all the comments btw, adds some really good context :+1:

rohan-bes (approved):

PR #5073, opened 2023-10-12
WAITP-1366 Fetch dashboards from models

EMcQ-BES:
### Changes:

- Fetch dashboards from models
  - We need to know that a dashboard exists that you don't have permission to in order to show you the `No access` visual. But central was filtering out the ones you don't have permissions to.


tcaiger (approved):
Makes sense to me

PR #5074, opened 2023-10-12
WAITP-1458 Detach Timor-Leste from entity hierarchies

EMcQ-BES:
### Changes:

- Migration to remove `Timor-Leste` from entity hierarchies

We don't display `Timor-Leste` in the frontend for historic reasons. Let's just remove it from the hierarchies for the future


tcaiger (approved):
Looks good to me - thanks

PR #5076, opened 2023-10-12
WAITP-1448 Remove unused projects

EMcQ-BES:
### Changes:

- Delete `laos_eoc` and `covidau` projects from db

I know we generally just do data migrations directly in the db these days, but this is pretty low priority and better kept as part of the rewrite branch I think.


tcaiger (approved):
yeya

PR #5082, opened 2023-10-15
Fix DeleteAfter

IgorNadj:
### Issue #: No Issue

Currently the DeleteAfter logic requires the check to run the same day as the DeleteAfter value.

If an instance is shut down overnight it can be missed by the deleter completely. This change just makes it consider all instances with DeleteAfter, not just those with a value of today.

Tested in `igor-test-deployment` lambda function.


rohan-bes (approved):
LGTM

PR #5086, opened 2023-10-16
WAITP-1449: QR code generator

acdunham:
### Changes:
- Updated survey submit endpoint to handle entity create questions when these are hidden
- Updated survey submit endpoint to return a list of created entities
- Added QR code panel and modal to success screen

---

### Screenshots:
Screenshot 2023-10-16 at 4 14 03 pm

Screenshot 2023-10-16 at 4 22 26 pm

Screenshot 2023-10-16 at 4 05 15 pm

Screenshot 2023-10-16 at 4 29 32 pm

Screenshot 2023-10-16 at 4 27 29 pm


packages/datatrak-web-server/src/utils/buildUpsertEntity.ts
```diff
@@ -33,7 +34,9 @@ export const buildUpsertEntity = async (
     const fieldValue = typeof value === 'string' ? value : answers[value.questionId];
 
     if (fieldName === 'parentId') {
+      console.log('fieldValue', fieldValue);
       const entityRecord = await getEntity(fieldValue);
+      console.log('entityRecord', entityRecord);
       entity.parent_id = entityRecord.id;
```
tcaiger, line 40:
```suggestion
    if (fieldName === 'parentId') {
      const entityRecord = await getEntity(fieldValue);
      entity.parent_id = entityRecord.id;
```

packages/datatrak-web-server/src/utils/processSurveyResponse.ts
```diff
@@ -76,31 +89,20 @@ export const processSurveyResponse = async (
...
-            countryId,
-            getEntity,
-          );
-          surveyResponse.entities_upserted.push(entityObj);
-        }
-        answersToSubmit.push(answerObject);
+
+      // add the entity id to the response if the question is a primary entity question
+      case QuestionType.PrimaryEntity: {
+        const entityId = answer?.hasOwnProperty('id') ? (answer as Entity).id : (answer as string);
```
tcaiger, line 101: What is the use case where the answer is an object?

packages/datatrak-web-server/src/utils/processSurveyResponse.ts
```diff
@@ -54,13 +57,23 @@ export const processSurveyResponse = async (
...
   const answersToSubmit = [] as Record<string, unknown>[];
 
   for (const question of questions) {
     const { questionId, type } = question;
-    const answer = answers[questionId];
+    let answer = answers[questionId] as AnswerT | Entity;
+    const config = question?.config as ConfigT;
+
+    // If the question is an entity question and an entity should be created by this question, build the entity object. We need to do this before we get to the check for the answer being empty, because most of the time these questions are hidden and therefore the answer will always be empty
+    if (['PrimaryEntity', 'Entity'].includes(type) && isUpsertEntityQuestion(config)) {
```
tcaiger, line 72:
```suggestion
    if ([QuestionType.PrimaryEntity, QuestionType.Entity] && isUpsertEntityQuestion(config)) {
```
Might as well use the enum here for consistency

packages/datatrak-web-server/src/utils/processSurveyResponse.ts
acdunham: This is when a new entity has been created. When it hasn't it will be a string, because the answer will just be the entity id

packages/datatrak-web-server/src/utils/processSurveyResponse.ts
```diff
@@ -54,13 +57,23 @@ export const processSurveyResponse = async (
...
+      const entityObj = await buildUpsertEntity(config, questionId, answers, countryId, getEntity);
+      if (entityObj) surveyResponse.entities_upserted.push(entityObj);
+      answer = entityObj;
```
tcaiger, line 75: Thanks for fixing up this logic - it's pretty dense stuff 😬

packages/datatrak-web/src/features/Survey/Components/SurveyQRCode/SurveyQRCode.tsx
```diff
@@ -0,0 +1,32 @@
+/*
+ * Tupaia
+ * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
+import { useLocation } from 'react-router';
+import { QRCodeList } from './QRCodeList';
+import { QRCodeModal } from './QRCodeModal';
+import { QRCodePanel } from './QRCodePanel';
```
tcaiger, line 10: I think these components are really nicely split up and organised. I'm keen to avoid nesting the folder structure too deeply and I'm keen to avoid using folders named `components` if possible as I think it's not that clear to code readers. So for this one could we just move the `SurveyQRCode` folder up one level to simply be `Survey/SurveyQRCode`?

packages/datatrak-web/src/api/mutations/useSubmitSurvey.ts
```diff
@@ -50,14 +50,19 @@ export const useSubmitSurvey = () => {
...
     },
     {
-      onSuccess: () => {
+      onSuccess: data => {
         successToast("Congratulations! You've earned a coconut", Coconut);
-        navigate(generatePath(ROUTES.SURVEY_SUCCESS, params));
+        // include the survey response data in the location state, so that we can use it to generate QR codes
+        navigate(generatePath(ROUTES.SURVEY_SUCCESS, params), {
+          state: {
+            surveyResponse: JSON.stringify(data),
```
tcaiger, line 63: This could just be an object and we could avoid stingify and parse right?

acdunham: No, with react router, you can't pass complex objects as location state, they have to be stringified

tcaiger: I reckon it does work as it's just using the browser [history object](https://developer.mozilla.org/en-US/docs/Web/API/History_API/Working_with_the_History_API) (plus I tested it :)) but I don't mind too much either way 👍

packages/datatrak-web-server/src/__tests__/utils/processSurveyResponse.test.ts
```diff
@@ -314,7 +319,10 @@ describe('processSurveyResponse', () => {
         {
           question_id: 'question1',
           type: QuestionType.Entity,
-          body: 'answer1',
```
tcaiger, line 317: I think that this should still be a string answer right? I don't think that central server handles body's that are an Entity object?

tcaiger (changes requested):
Looks great overall just one question about the Entity Question processing - I could be wrong but I think that we need to rejig to the saving of Entity Questions slightly to be strings.

tcaiger (approved):
Perfecto!

PR #5087, opened 2023-10-16
RN-1052: Improve Entity ID display in QR code image

chris-pollard:
### Changes:

- Change character wrapping limit from 12 to 13 in meditrak app and ui-components packages.


rohan-bes (approved):
EZPZ

PR #5089, opened 2023-10-16
WAITP-1454: UI changes

acdunham:
### Changes:
- Updated button disabled state opacity
- Reordered components so that survey side menu doesn't re-render on screen change and close the menu (on mobile)
- Re-ordered context so that cancel survey modal can be controlled by clicking the header logo button


tcaiger (approved):
Looks good to me - thanks for the updates :)

PR #5090, opened 2023-10-16
WAITP-1377 Forward signup route to web config server

EMcQ-BES:
### Changes:

- Fixes 404 error in datatrak web


acdunham (approved):
Thanks for doing this! I guess it's not ideal to have to use web-config-server since we want to deprecate it, but that can probably be a maintenance task when we actually do that.
I guess we need to make sure the web config api url is added to the lastpass env file too

EMcQ-BES:
@acdunham Yeah I double checked and we were already using `web-config-server` for the projects route in `datatrak-web-server` so I just did it the easy way

PR #5092, opened 2023-10-16
WAITP-1459: Remove brush from charts

acdunham:
### Changes:
- Removed Brush from charts


tcaiger (approved):
3 points in the bag :)

PR #5097, opened 2023-10-17
no-issue: Update MeditrakSurveyResponseRequest type

rohan-bes:
Ran into some issues with tests failing for `datatrak-web-server` [here](https://github.com/beyondessential/tupaia/actions/runs/6540366652/job/17760151802?pr=5007) so updated some of the types there.

### DIFF
--- a/packages/datatrak-web-server/src/types.ts
+++ b/packages/datatrak-web-server/src/types.ts
@@ -3,8 +3,12 @@
  *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
-import { ModelRegistry, EntityModel } from '@tupaia/database';
+import { ModelRegistry, EntityModel, EntityType as BaseEntityType } from '@tupaia/database';
+import { Model } from '@tupaia/server-boilerplate';
+import { Entity } from '@tupaia/types';
+
+export type EntityType = BaseEntityType & Entity;
 
 export interface DatatrakWebServerModelRegistry extends ModelRegistry {
-  readonly entity: EntityModel;
+  readonly entity: Model<EntityModel, Entity, EntityType>;
 }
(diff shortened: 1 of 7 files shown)


packages/datatrak-web-server/src/types.ts
```diff
@@ -3,8 +3,12 @@
...
-import { ModelRegistry, EntityModel } from '@tupaia/database';
+import { ModelRegistry, EntityModel, EntityType as BaseEntityType } from '@tupaia/database';
+import { Model } from '@tupaia/server-boilerplate';
+import { Entity } from '@tupaia/types';
+
+export type EntityType = BaseEntityType & Entity;
 
 export interface DatatrakWebServerModelRegistry extends ModelRegistry {
-  readonly entity: EntityModel;
+  readonly entity: Model<EntityModel, Entity, EntityType>;
```
rohan-bes, line 13: This is probably what we want to be doing when defining ModelRegistry types. By wrapping in the `Model` type, we get access to the various methods defined there (`findById` is the one we want in this context).

packages/types/src/types/requests/central-server/MeditrakSurveyResponseRequest.ts
```diff
@@ -59,8 +42,8 @@ export interface MeditrakSurveyResponseRequest {
    */
   timestamp?: string;
   approval_status?: string;
-  entities_created?: EntityCreated[];
-  options_created?: OptionCreated[];
+  entities_upserted?: Entity[];
+  options_created?: Omit<Option, 'id'>[];
```
rohan-bes, line 46: @IgorNadj I might take a look at how easy it'd be to do the auto-generated `<Model>WithoutId` types since they do crop up a lot...

packages/datatrak-web-server/src/types.ts
tcaiger: Oh I didn't know about that Generic, thanks for the info. @EMcQ-BES you probably already know about this but in case not take a look

tcaiger (approved):
Looks great - thanks for cleaning up all of these. It's good at least that we are using these types across meditrak and datatrak. After you merge I will pull this into the datatral-web epic

PR #5098, opened 2023-10-17
no-issue: Add <Model>WithoutId types for convenient typedefs when cre…

rohan-bes:
For convenient typedefs when creating new model type instances

### DIFF
--- a/packages/types/generate-models.ts
+++ b/packages/types/generate-models.ts
@@ -30,13 +30,29 @@ const run = async () => {
   const failOnChanges = process.argv[2] === '--failOnChanges';
 
   // @ts-ignore
-  const tsString = await sqlts.toTypeScript(config, db);
+  const definitions = await sqlts.toObject(config, db);
+  const { tables } = definitions;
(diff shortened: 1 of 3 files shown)


rohan-bes:
@IgorNadj turns out this one was pretty easy to do. TBH I'm still on the fence about it though, feels as though we could just define a generic `WithoutId<T>` which we could use instead? Would save a lot of (admittedly auto generated) code. But I think the con there is that we don't get easy-to-read type definitions.

Thoughts?

IgorNadj:
> @IgorNadj turns out this one was pretty easy to do. TBH I'm still on the fence about it though, feels as though we could just define a generic `WithoutId<T>` which we could use instead? Would save a lot of (admittedly auto generated) code. But I think the con there is that we don't get easy-to-read type definitions.
> 
> Thoughts?

Oh, very interesting approach, nicely done.

Yeah the benefit of this way is we get the schemas for these WithoutId types auto-generated.

IgorNadj (approved):

PR #5099, opened 2023-10-17
WAITP-1415: Photo upload question

acdunham:
### Changes:
- Added photo question to `datatrak-web`
- Updated `ImageUploadField` in `ui-components` to fix issue where clicking the 'delete' button triggered file explorer to open, and added a few new optional props
- Updated comments in backend

Note: This will not look like the design because creating a drag and drop file upload is out of scope of this ticket, so we are using the existing input from `ui-components`

---

### Screenshots:
Screenshot 2023-10-18 at 8 22 57 am


tcaiger (approved):
Very nice :)

PR #5103, opened 2023-10-19
WAITP-1366 Bugfix circular recursion in tupaia-web-server

EMcQ-BES:
### Changes:

- Use explicit type declarations
- Strip 'model' from Dashboards before returning


tcaiger (approved):
LGTM

PR #5106, opened 2023-10-23
no-issue: Added <Model>Create and <Model>Update types

rohan-bes:
Reworked this one a bit: https://github.com/beyondessential/tupaia/pull/5098

Decided to add 2 more types for each table, Create and Update.

Create omits the `id` column.
Update has all columns be optional.

I also tested with setting all columns on the base type to required, since they will be still have `null` as an option if they are nullable. But that was a bit unweildy and felt like it was tying these types a bit too closely to their database implementation. Conceptually a nullable column is an optional field in almost all cases.

Also filtered out all the mvrefresh columns (eg. `m_row$`) from the type defs.

### DIFF
--- a/packages/types/generate-models.ts
+++ b/packages/types/generate-models.ts
@@ -46,6 +46,12 @@ const removeIdColumn = (tables: Table[]) =>
     columns: columns.filter(({ name }) => name !== 'id'),
   }));
 
+const removeUnwantedColumns = (tables: Table[]) =>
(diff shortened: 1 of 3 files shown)


rohan-bes:
@IgorNadj could I request a review on this one? It's slightly changed from the other one.

IgorNadj (approved):
Beautiful!

PR #5108, opened 2023-10-23
Release 2023-43: Hotfix css issues

rohan-bes:


packages/ui-components/src/components/Inputs/Autocomplete.tsx
```diff
@@ -44,9 +44,6 @@ const StyledAutocomplete = styled(MuiAutocomplete)`
       border-color: ${props => props.theme.palette.error.main};
     }
   }
-  .MuiFormControl-root {
-    margin-bottom: 0;
-  }
```
rohan-bes, line 49: @tcaiger I removed this since it was causing the input fields to be too bunched up.

Not sure how this will affect datatrak, perhaps something to do with the error messaging?

packages/ui-components/package.json
```diff
@@ -28,7 +28,6 @@
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.47",
     "@material-ui/pickers": "^3.2.10",
-    "@material-ui/styles": "v4.2.0",
```
acdunham, line 31: This looks like it was added to make the data grid work - might be worth checking with @chris-bes about that?

packages/ui-components/src/components/Inputs/Autocomplete.tsx
acdunham: I think this is fine, we can just add in any extra styling to datatrak-web instead. This was my bad

acdunham (approved):
In general this looks good, but I would just check the removal of the mui styles package, since it appears to have been added for the data-grid work. Pre-approving, so if that's all good, then I'm happy with that, otherwise potentially need a work around here

packages/ui-components/package.json
rohan-bes: I was indeed! But we chatted and he said it was all good to remove :+1: (and I checked and the data-grid is still looking good)

PR #5111, opened 2023-10-23
no-issue: Consolidate all model type definitions into server-boilerplate 

rohan-bes:
We've been defining Model types (eg. `EntityModel`) in various packages over time, and since they get reuse between packages it'd be good to consolidate them.

Why server-boilerplate? Well the Model types depend on the types in `@tupaia/database`, so they can't go in the `@tupaia/types` package. Since they're only used by servers, it feels most appropriate.

Also made a few updates to make types with JSONB columns.


packages/database/src/DatabaseType.js
```diff
@@ -164,6 +164,10 @@ export class DatabaseType {
       }
     } else {
       const record = await this.model.create(data);
+
+      /**
+       * @type {string}
+       */
```
rohan-bes, line 170: Without this the type of `id` was `any` which was causing issues. Previously we had to explicitly omit this type from the base type.

packages/database/src/migrations/20231023053924-FixNullableColumnsOnEntity-modifies-schema.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`
+    ALTER TABLE entity ALTER COLUMN attributes SET NOT NULL;
```
rohan-bes, line 19: This JSONB column defaults to an empty object, so it should never be NULL

packages/datatrak-web-server/src/__tests__/utils/processSurveyResponse.test.ts
```diff
@@ -2,23 +2,18 @@
...
 import { getBrowserTimeZone } from '@tupaia/utils';
 import { processSurveyResponse } from '../../utils';
 
-// Mock out moment so that that toISOString returns a consistent value for our tests
-jest.mock('moment', () => {
-  const mMoment = {
-    toISOString: jest.fn(() => 'theISOString'),
-  };
-  return jest.fn(() => mMoment);
-});
```
rohan-bes, line 16: Switched out moment as:
a) It's being deprecated and we want to stop using it
b) It was breaking the tests

packages/datatrak-web/src/views/SurveySelectPage/SurveyCountrySelector.tsx
```diff
@@ -40,7 +40,7 @@ const CountrySelectWrapper = styled.div`
 
 interface SurveyCountrySelectorProps {
   countries: Entity[];
-  selectedCountry?: Entity | null;
+  selectedCountry?: Pick<Entity, 'code'> | null;
```
rohan-bes, line 43: Only `code` is needed here, and wanted a type that works both with `Country` and `Entity`

PR #5113, opened 2023-10-25
WAITP-1356: recent surveys

acdunham:
### Changes:
- Added route for recent survey fetching
- Added recent survey tiles in landing page

---

### Screenshots:

Mobile
Screenshot 2023-10-26 at 10 37 07 am

Desktop
Screenshot 2023-10-26 at 10 36 57 am

No data
Screenshot 2023-10-24 at 11 41 38 am


packages/datatrak-web/src/api/queries/useRecentSurveys.ts
```diff
@@ -0,0 +1,22 @@
...
+export const useRecentSurveys = (userId?: string) => {
+  return useQuery(
+    ['recentSurveys', userId],
+    (): Promise<DatatrakWebRecentSurveysRequest.ResBody> =>
+      get('recentSurveys', { params: { userId } }),
+    { enabled: !!userId },
+  );
+};
+
+export const useCurrentUserRecentSurveys = () => {
```
tcaiger, line 19: I'm liking this pattern of basic hook and then extended hook

packages/datatrak-web/src/components/Tile.tsx
```diff
@@ -57,19 +57,26 @@ const Text = styled(Typography)`
...
+  onClick?: () => void;
 }
 
-export const Tile = ({ title, text, children, to, tooltip }: TileProps) => {
+export const Tile = ({ title, text, children, to, tooltip, Icon, onClick }: TileProps) => {
   return (
-    <ButtonWrapper component={RouterLink} to={to} tooltip={tooltip}>
-      <SurveyTickIcon />
+    <ButtonWrapper
+      component={to ? RouterLink : undefined}
```
tcaiger, line 70:
```suggestion

```
I don't think that we need this logic here since it's happening in the button component right?

packages/datatrak-web-server/src/routes/UserRoute.ts
```diff
@@ -41,10 +41,11 @@ export class UserRoute extends Route<UserRequest> {
...
     }
     if (countryId) {
-      country =
-        (await ctx.services.central.fetchResources(`/entities/${countryId}`, {
-          columns: ['id', 'name', 'code'],
-        })) || null;
+      const countryResponse = await ctx.services.central.fetchResources(`/entities/${countryId}`, {
+        columns: ['id', 'name', 'code'],
+      });
+      console.log(countryResponse);
```
tcaiger, line 47:
```suggestion

```

packages/datatrak-web-server/src/routes/RecentSurveysRoute.ts
```diff
@@ -0,0 +1,40 @@
...
+>;
+
+export class RecentSurveysRoute extends Route<RecentSurveysRequest> {
+  public async buildResponse() {
+    const { query, models } = this.req;
+
+    const { userId } = query;
+
+    // The central server queries already built in don't support GROUP BY and this would be unnecessary to add for this one use case, so we use raw SQL in this instance
+    const surveyResponses = await models.database.executeSql(
```
tcaiger, line 25: I think it's fair enough to go around the central server if it doesn't support group by and I agree that would be out of scope to add support for. Apart from that we should use [parameter binding](https://knexjs.org/guide/raw.html#raw-parameter-binding) to pass in the userId since it is  a value coming over the network and could be a security risk.

Also I have a personal preference for using the model syntax over raw sql but I think that is just personal preference. @EMcQ-BES would you also mind casting your eye over this in case we have forgotten anything. For example, do we need to consider permissions in this context?

tcaiger (changes requested):
Looks good - nice full stacking. Just have one suggestion on the Surveys Route

packages/datatrak-web-server/src/routes/RecentSurveysRoute.ts
EMcQ-BES: @tcaiger We raised the permissions question with Georgia. Technically a user should _always_ have permissions for surveys they themselves submitted (or else they couldn't have submitted them)

The edge case of someone having their permissions revoked after submitting we've agreed the tile will still display, but upon clicking through the user will be rejected from the survey route.

EMcQ-BES: As far as the model syntax, we can't actually use a `group by` with it either. The `buildQuery` function in `TupaiaDatabase` doesn't have support for it. When we need a group by we always either use `RAW` or run the grouping in js

tcaiger: Thanks Ethan. So then the only change we need to make here is the param binding / replacements tweak and then it's good to go

tcaiger (approved):

PR #5115, opened 2023-10-26
no-issue: Move SurveyScreenComponent config types

rohan-bes:
@tcaiger and I paired on this and decided to move them into models-extra for now. Ideally we'd put them on the survey_screen_component model itself, but since the underlying config field is text for now, best to just define them separately.

### DIFF
--- a/packages/datatrak-web/src/features/Survey/SurveyContext/utils.ts
+++ b/packages/datatrak-web/src/features/Survey/SurveyContext/utils.ts
@@ -4,7 +4,12 @@
  */
 
 import { BooleanExpressionParser, ExpressionParser } from '@tupaia/expression-parser';
-import { ArithmeticQuestionConfig, ConditionQuestionConfig, QuestionType } from '@tupaia/types';
+import {
+  ArithmeticQuestionConfig,
+  CodeGeneratorQuestionConfig,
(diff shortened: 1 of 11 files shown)


packages/datatrak-web/src/features/Survey/SurveyContext/utils.ts
```diff
@@ -155,6 +152,16 @@ const resetInvisibleQuestions = (
   return updatedFormData;
 };
 
+const hasConditionConfig = (
+  ssc: SurveyScreenComponent,
+): ssc is SurveyScreenComponent & { config: { condition: ConditionQuestionConfig } } =>
+  ssc.type === QuestionType.Condition && ssc.config?.condition !== undefined;
```
rohan-bes, line 158: @tcaiger Decided to include these type guards in the end. Happy with either approach in this context really, but in general prefer to avoid `as <Type>` :+1:

tcaiger: Thanks @rohan-bes I totally agree and like this approach. @acdunham this is an example of the typeguards I mentioned the other day

tcaiger (approved):
Great refactor :)

PR #5118, opened 2023-10-27
MDEV-200: survey response data table

acdunham:
### Changes:
- Created a `survey_responses` data table
- Created migrations for `survey_responses` data table
- Added tests


packages/data-table-server/src/dataTableService/services/SurveyResponseDataTableService.ts
```diff
@@ -0,0 +1,126 @@
...
+import { getDefaultEndDate, getDefaultStartDate } from './utils';
+
+const requiredParamsSchema = yup.object().shape({
+  ids: yup.array().of(yup.string()).default([]),
+  assessorNames: yup.array().of(yup.string()).default([]),
+  countryCodes: yup.array().of(yup.string()).default([]),
+  surveyCodes: yup.array().of(yup.string()).default([]),
+  entityCodes: yup.array().of(yup.string()).default([]),
+  startDate: yup.date().default(getDefaultStartDate),
+  endDate: yup.date().default(getDefaultEndDate),
```
rohan-bes, line 19: I'd probably lean towards no default values for start/end date, feels a little unexpected that we'd be filtering by default

packages/data-table-server/src/dataTableService/services/SurveyResponseDataTableService.ts
```diff
@@ -0,0 +1,126 @@
...
+  [ResponseColumns.surveyName]: Survey['name'];
+  [ResponseColumns.surveyCode]: Survey['code'];
+  [ResponseColumns.surveyResponseId]: SurveyResponse['id'];
+  [ResponseColumns.surveyResponseAssessorName]: SurveyResponse['assessor_name'];
+  [ResponseColumns.surveyResponseDataTime]: SurveyResponse['data_time'];
+  [ResponseColumns.surveyResponseOutdated]: SurveyResponse['outdated'];
+};
+
+/**
+ * DataTableService for pulling data from aggregator's fetchDataGroup() endpoint
```
rohan-bes, line 74:
```suggestion
 * DataTableService for pulling data from central-api `/surveyResponses` endpoint 
```

packages/data-table-server/src/dataTableService/services/SurveyResponseDataTableService.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    if (params.startDate && params.endDate) {
+      // Set the start date to the beginning of the day and the end date to the end of the day
+      const startDate = new Date(params.startDate).setHours(0, 0, 0, 0);
+      const endDate = new Date(params.endDate).setHours(23, 59, 59, 999);
+      filter['survey_response.data_time'] = {
+        comparator: 'between',
+        comparisonValue: [new Date(startDate), new Date(endDate)],
+      };
+    }
```
rohan-bes, line 110: I think we could have a couple more conditions here for when either startDate or endDate are present. The 3 options should be:
```
startDate: data_time > startDate
endDate: data_time < endDate
startDate && endDate: startDate  < data_time < endDate
```

packages/data-table-server/src/dataTableService/services/SurveyResponseDataTableService.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      if (filterValue === undefined || filterValue === null) return acc;
+
+      if (Array.isArray(filterValue) && filterValue.length === 0) return acc;
+
+      return {
+        ...acc,
+        [filterField]: filterValue,
+      };
+    }, {} as Record<string, any>);
```
rohan-bes, line 100: Could we just use the params here if we removed the defaults for each param? I think the frontend should prevent empty arrays being passed in

packages/data-table-server/src/dataTableService/services/SurveyResponseDataTableService.ts
```diff
@@ -0,0 +1,126 @@
...
+    return filter;
+  }
+
+  protected async pullData(params: Params) {
+    const filter = this.getFilter(params);
+
+    const results = await this.ctx.apiClient.central.fetchResources('surveyResponses', {
+      filter,
+      columns: Object.values(ResponseColumns),
+      sort: [`${ResponseColumns.surveyResponseDataTime} DESC`],
```
rohan-bes, line 121: Good call to sort by data_time :+1:

packages/api-client/src/connections/mocks/MockCentralApi.ts
```diff
@@ -25,11 +72,18 @@ export class MockCentralApi implements CentralApiInterface {
...
+      this.mockData[endpoint]?.filter(resource => {
+        if (params?.filter) {
+          return Object.entries(params.filter)?.every(([field, filter]) => {
+            const resourceValue = resource[field];
+            return getValueMatchesFilter(resourceValue, filter);
+          });
+        }
+        return true;
+      }) ?? [],
+    );
```
rohan-bes, line 86:
```suggestion
public async fetchResources(endpoint: string, params?: Params): Promise<any> {
    const resourceData = this.mockData[endpoint];
    if (!resourceData) return [];

    const filter = params?.filter;
    if (!filter) return resourceData;

    return resourceData.filter(resource =>
      Object.entries(filter).every(([field, check]) =>
        getValueMatchesFilter(resource[field], check),
      ),
    );
  }
```

Bit of personal preference on this one but I took a stab at rewriting this to be a bit more succinct. I'm a bit caution with using the optional chaining operator, as I find it can be hard to parse. So I prefer to exit early where it's convenient.

rohan-bes (approved):
Great stuff @acdunham. Your first data-table :grin: 

A few changes requested here but nothing major so pre-approving :+1:

PR #5121, opened 2023-10-29
no-issue: Add @tupaia/server-utils package

rohan-bes:
We've discussed wanting this package for a while now, as some of the server-side utility functions add large dependencies to our common utils packages that may be unsupported in the browser.

I've added the package here and moved a couple of key utility functions over. There's more that we could and should move over, but they might take a bit of time to dis-entangle (`winston`, `node-fetch`, etc.) so happy to leave for now.

Doing this now since we're about to move the mailing logic into a utils package and I didn't want to add yet another server-only utility to `@tupaia/utils`


tcaiger (approved):
Everything looks good to me. It's great to see this change coming through.

PR #5125, opened 2023-10-30
RN-1060: (task) Added dashboards/:project/:entity/:dashboard/export route to tupaia-web-server

rohan-bes:
### Issue RN-1060:

Since we're adding custom functionality to this route as part of the email work, decided to port it over from web-config-server.

Also chose to rename the route from a generic `/pdf` to `/dashboards/:project/:entity/:dashboard/export`. Since this route is really just being used once, I feel it's nice if we capture the business case as part of the route interface.


EMcQ-BES (approved):
Clean, LGTM

PR #5127, opened 2023-10-31
RN-1060: (task) Add `allowSendToMailingListPermissionGroups` flag to the project.config

rohan-bes:
### Issue RN-1060 (task):

Added a new `allowSendToMailingListPermissionGroups` config to control which users can send out the dashboard export to the mailing list.

Ran into some issues with the `KeysToCamelCase` type which required changing how to generate the schema, so decided to break this out into a separate PR.

### DIFF
--- a/packages/database/src/migrations/20231031011521-RemoveNullableFromProjectConfig-modifies-schema.js
+++ b/packages/database/src/migrations/20231031011521-RemoveNullableFromProjectConfig-modifies-schema.js
@@ -1,0 +1,27 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`ALTER TABLE project ALTER COLUMN config SET NOT NULL`);
+};
+
+exports.down = function (db) {
+  return db.runSql(`ALTER TABLE project ALTER COLUMN config DROP NOT NULL`);
+};
+
+exports._meta = {
+  version: 1,
+};
--- a/packages/types/generate-schemas.ts
+++ b/packages/types/generate-schemas.ts
@@ -14,6 +14,18 @@ const settings: TJS.PartialArgs = {
   noExtraProps: true,
 };
 
+function getTsFiles(dir: string): string[] {
+  const dirContents = fs.readdirSync(dir);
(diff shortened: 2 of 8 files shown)


packages/types/src/utils/casing.ts
```diff
@@ -16,14 +16,10 @@ type CamelCasePart<S extends string> = S extends `${infer First}_${infer Rest}`
...
 
 // Converts type objects or arrays to camel case
-export type KeysToCamelCase<T> = {
-  [K in keyof T as CamelCase<string & K>]: T[K] extends Array<any>
-    ? KeysToCamelCase<T[K][number]>[]
-    : ObjectToCamel<T[K]>;
-};
+export type KeysToCamelCase<T> = T extends Array<infer ArrayElm>
+  ? KeysToCamelCase<ArrayElm>[] // If array, camel case items
+  : ObjectToCamel<T>; // If object, camel case keys;
```
rohan-bes, line 25: This was broken previously, such that if you passed in an object with a field that's an array, it'd cast the array as an object (with fields like `concat`,`map`,`forEach` etc.)

packages/types/generate-schemas.ts
```diff
@@ -26,10 +38,13 @@ const HEADER = `/*
...
 `;
 
+const rootDir = 'src/types';
+const filesToBuildSchemasFor = getTsFiles(rootDir);
+
 const { filename, typesPath } = config as any;
 
 const program = TJS.getProgramFromFiles([resolve(typesPath)]);
-const schemas = TJS.generateSchema(program, '*', settings);
+const schemas = TJS.generateSchema(program, '*', settings, filesToBuildSchemasFor);
```
rohan-bes, line 47: We `import` the `KeysToCamelCase` type into the `src/types/request` module, so it was always getting added to the schema.

Tried a few ways to get the generate schemas to just build against the `src/types` directory, but ultimately had to brute force this.

packages/database/src/migrations/20231031011521-RemoveNullableFromProjectConfig-modifies-schema.js
```diff
@@ -0,0 +1,27 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`ALTER TABLE project ALTER COLUMN config SET NOT NULL`);
```
rohan-bes, line 18: Our types package plays nicely when we set fields to be not null, and in this case we never really expect the config to be `NULL` ever (empty object represents empty state)

rohan-bes:
Closing this one out as we don't want to control permissions on mailing list this way

PR #5129, opened 2023-10-31
WAITP-1214: Mobile dashboard tabs fixes

tcaiger:
### Changes:

Fixes issues 6-12 that were raised in this ticket https://linear.app/bes/issue/WAITP-1214/mobile-dashboard-and-map-tabs

---

### Screenshots:

image

### DIFF
--- a/packages/tupaia-web/src/AppProviders.tsx
+++ b/packages/tupaia-web/src/AppProviders.tsx
@@ -12,7 +12,7 @@ import { theme } from './theme';
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
-      staleTime: 1000 * 60 * 5, // 5 minutes
+      staleTime: 1000 * 60 * 15, // 15 minutes
       refetchOnWindowFocus: false,
       retry: 0,
       keepPreviousData: true,
--- a/packages/tupaia-web/src/features/Map/Map.tsx
+++ b/packages/tupaia-web/src/features/Map/Map.tsx
@@ -43,6 +43,11 @@ const StyledMap = styled(LeafletMap)`
     z-index: 0;
   }
 
+  .leaflet-bottom {
+    // Set the z-index to 1 so it doesn't overlap dashboard controls on mobile
+    z-index: 1;
+  }
+
   // Overwrite default zoom control styles
   .leaflet-control-zoom {
     z-index: 1;
--- a/packages/tupaia-web/src/features/Map/Map.tsx
+++ b/packages/tupaia-web/src/features/Map/Map.tsx
@@ -150,7 +150,7 @@ export const Map = () => {
           zoom={zoom as LeafletMapProps['zoom']}
           shouldSnapToPosition
         >
-          <TileLayer tileSetUrl={activeTileSet.url} showAttribution />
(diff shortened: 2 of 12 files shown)


packages/tupaia-web/src/AppProviders.tsx
```diff
@@ -12,7 +12,7 @@ import { theme } from './theme';
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
-      staleTime: 1000 * 60 * 5, // 5 minutes
+      staleTime: 1000 * 60 * 15, // 15 minutes
```
tcaiger, line 15: Made the stale time longer as the data can take a long time to load and is not updated often and this came up in testing.

packages/tupaia-web/src/features/Map/MapOverlaySelector/MapOverlayList.tsx
```diff
@@ -167,7 +171,12 @@ export const MapOverlayList = () => {
     // when overlay changes, reset period to default
     urlSearchParams.set(URL_SEARCH_PARAMS.MAP_OVERLAY_PERIOD, DEFAULT_PERIOD_PARAM_STRING);
     setUrlParams(urlSearchParams);
+
+    if (toggleOverlayLibrary) {
+      toggleOverlayLibrary();
```
tcaiger, line 176: close overlay after a change

packages/tupaia-web/src/features/Map/Map.tsx
```diff
@@ -145,7 +150,7 @@ export const Map = () => {
           zoom={zoom as LeafletMapProps['zoom']}
           shouldSnapToPosition
         >
-          <TileLayer tileSetUrl={activeTileSet.url} showAttribution />
+          <TileLayer tileSetUrl={activeTileSet.url} showAttribution={false} />
```
acdunham, line 153: Just checking, was this something that just got changed to be true? If so, will setting it to false undo that fix?

packages/tupaia-web/src/features/Map/MapOverlaySelector/MobileMapOverlaySelector.tsx
```diff
@@ -96,19 +103,27 @@ export const MobileMapOverlaySelector = ({
...
+  const { hasMapOverlays } = useMapOverlays(projectCode, entityCode);
+
   return (
     <Wrapper>
       {!overlayLibraryOpen && (
         <ButtonWrapper>
           <MapOverlayDatePicker />
-          <ExpandButton onClick={toggleOverlayLibrary} aria-controls="overlay-selector">
+          <ExpandButton
+            onClick={hasMapOverlays ? toggleOverlayLibrary : undefined}
```
acdunham, line 115: To be really pedantic, we could just disable the button when `hasMapOverlays` is false, which from an accessibility perspective is preferable I think

acdunham (approved):
LGTM, thanks for doing these changes. I have added a couple of questions, but otherwise pre-approving.

packages/tupaia-web/src/features/Map/Map.tsx
tcaiger: It was but now I've upgraded the attribution to a custom one to match what was on tupaia before.

PR #5130, opened 2023-10-31
RN-1060 (task): Added dashboards/:project/:entity/:dashboard/email route to tupaia-web-server

rohan-bes:
### Issue RN-1060 (task):

Added endpoint to tupaia-web-server for emailing the dashboard export to all subscribed users in the mailing list.

There is a permissions check to ensure that the user requesting the email has appropriate permissions for that project.

### DIFF
--- a/packages/tupaia-web/src/api/mutations/useEmailDashboard.tsx
+++ b/packages/tupaia-web/src/api/mutations/useEmailDashboard.tsx
@@ -1,0 +1,45 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import { useMutation } from 'react-query';
+import { API_URL, post } from '../api';
+import { Dashboard, DashboardItem, EntityCode, ProjectCode } from '../../types';
+
+type EmailDashboardParams = {
+  projectCode?: ProjectCode;
+  entityCode?: EntityCode;
+  dashboardName?: Dashboard['name'];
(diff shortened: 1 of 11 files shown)


packages/tupaia-web/src/api/mutations/useEmailDashboard.tsx
```diff
@@ -0,0 +1,45 @@
...
+ */
+import { useMutation } from 'react-query';
+import { API_URL, post } from '../api';
+import { Dashboard, DashboardItem, EntityCode, ProjectCode } from '../../types';
+
+type EmailDashboardParams = {
+  projectCode?: ProjectCode;
+  entityCode?: EntityCode;
+  dashboardName?: Dashboard['name'];
+  dashboardId?: Dashboard['id'];
```
rohan-bes, line 13: We have to send through the dashboardId separately, since we use the `dashboardName` in the tupaia urls. I took a quick looking to refactoring things so that we use the unique `dashboard.code` in the url instead, but it looked a little awkward for when there's a multi-country project with a default dashboard. So left it for now.

packages/tupaia-web/src/api/mutations/useEmailDashboard.tsx
rohan-bes, line 1: Decided to add the mutation now, to provide an interface for the frontend to interact with this new route. Not actually adding this functionality to the front end right now as there's a couple of tickets that need to be completed first

chris-pollard (approved):
LG!

PR #5132, opened 2023-11-01
WAITP-1354: Leaderboard

acdunham:
### Changes:

- Added `central-server` `leaderboard` endpoint for use across `meditrak-app` and `datatrak-web`
- Updated `meditrak-app-server` to use `leaderboard` endpoint instead of duplicated logic
- Updated email blacklist to handle `bes.au` emails
- Added leaderboard to `datatrak-web`
 
---

### Screenshots:

Screenshot 2023-11-01 at 4 21 21 pm

Screenshot 2023-11-01 at 4 21 27 pm

Screenshot 2023-11-01 at 4 21 36 pm

Screenshot 2023-11-01 at 4 21 45 pm


packages/central-server/src/apiV2/meditrakApp/getLeaderboardList.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Tupaia MediTrak
+ * Copyright (c) 2017 Beyond Essential Systems Pty Ltd
+ */
+
+import { respond } from '@tupaia/utils';
+import { getLeaderboard } from '../../social';
+import { allowNoPermissions } from '../../permissions';
+
+export const getLeaderboardList = async (req, res) => {
```
acdunham, line 10: I created this because the `meditrak-app-server` had the same logic as `datatrak-web-server` was going to, and much of this was already in `central-server`. Are you okay with that @rohan-bes ? I didn't like the idea of duplicating the exact same logic

packages/datatrak-web/src/features/Leaderboard/Leaderboard.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import React from 'react';
+import styled from 'styled-components';
+import { useUserRewards } from '../../api/queries';
+import { UserRewardsSection } from './UserRewardsSection';
+import { LeaderboardTable } from './LeaderboardTable';
+
+const ScrollBody = styled.div`
+  overflow: auto;
+  background: white;
```
tcaiger, line 14: This is a good candidate for the paper variable from the theme

packages/datatrak-web/src/features/Leaderboard/LeaderboardTable.tsx
```diff
@@ -0,0 +1,123 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
+import styled from 'styled-components';
+import { LeaderboardItem } from '@tupaia/types';
```
tcaiger, line 8:
```suggestion

```

packages/central-server/src/apiV2/meditrakApp/getLeaderboardList.js
tcaiger: I won't speak for Rohan but From my point of view, this is a good call 👍

tcaiger (approved):
Great work - approved from me. But it would be good to wait for Rohan to do a quick review too

packages/central-server/src/social/getLeaderboard.js
```diff
@@ -33,19 +33,18 @@ const INTERNAL_EMAIL = '@beyondessential.com.au';
...
+      FROM (
+        SELECT user_id, COUNT(*) as coconuts, FLOOR(COUNT(*) / 100) as pigs
+        FROM survey_response
+        GROUP BY user_id
+      ) r
+      JOIN user_account on user_account.id = r.user_id
+      WHERE ${INTERNAL_EMAIL.map(email => `email NOT LIKE '%${email}'`).join(' AND ')}
+      AND email NOT IN (${EXCLUDED_USERS.join(',')})
+      ORDER BY coconuts DESC
+      LIMIT ?;
```
rohan-bes, line 46: Great call to try and keep things DRY here!

What I might recommend instead of moving this to a route in the central-server, is instead we move this function to the `SurveyResponseModel` in `@tupaia/database`. Since it's really a leader board of survey responses (though I think an argument could be made for the `UserModel` as well). I think that allows us to keep the api of central-server a little cleaner :+1:

rohan-bes (changes requested):
Great stuff @acdunham! Just a request to move the leaderboard calculation into the database model rather than into central-server.

packages/datatrak-web-server/src/models/SurveyResponse.ts
```diff
@@ -0,0 +1,30 @@
...
+  user_id: string;
+  start_time: string;
+  end_time: string;
+  metadata: Record<string, any>;
+  timezone: string | null;
+  entity_id: string;
+  data_time: string | null;
+  outdated: boolean | null;
+  approval_status: string;
+}>;
```
rohan-bes, line 23: The most up to date pattern here is to just use the type from `@tupaia/types` rather than defining all the fields here.

ie.
```
import {
  SurveyResponseModel as BaseSurveyResponseModel,
  SurveyResponseType as BaseSurveyResponseType,
} from '@tupaia/database';
import { SurveyResponse } from '@tupaia/types';

export interface SurveyResponseModelType
  extends SurveyResponse,
    Omit<BaseSurveyResponseType, 'id'> {} // Omit base `id: any` type as we explicity define as a string here

export interface SurveyResponseModel
  extends Model<BaseSurveyResponseModel, SurveyResponse, SurveyResponseModelType> {}
```

rohan-bes: Btw, I've got a [PR](https://github.com/beyondessential/tupaia/pull/5111) to centralise all these model definitions over to `@tupaia/server-boilerplate` so we don't need to keep defining them in the orchestration server. It just needs a bit of updating due to recent changes but I'll try and get it in soon :+1:

packages/database/src/modelClasses/SurveyResponse.js
```diff
@@ -7,6 +7,21 @@ import { MaterializedViewLogDatabaseModel } from '../analytics';
 import { DatabaseType } from '../DatabaseType';
 import { TYPES } from '../types';
 
+const EXCLUDED_USERS = [
```
rohan-bes, line 10:
```suggestion
const USERS_EXCLUDED_FROM_LEADER_BOARD = [
```
Perhaps a little clearer?

packages/central-server/src/social/getLeaderboard.js
rohan-bes, line 1: Just checking that this route is going to get removed?

rohan-bes (approved):
Few comments but nothing major so pre-approving. Pigs and Coconuts for all :tada: :pig: :coconut: :smile:

PR #5134, opened 2023-11-02
WAITP-1469: account delete

acdunham:
### Changes:
- Setup account settings page
- Add logic to save account delete requested boolean to the user preferences
- Add delete account section to account settings page

---

### Screenshots:
Screenshot 2023-11-02 at 10 16 14 am

Screenshot 2023-11-02 at 10 15 34 am

Screenshot 2023-11-02 at 3 19 52 pm
 
Screenshot 2023-11-02 at 3 42 08 pm

Screenshot 2023-11-02 at 3 41 52 pm


packages/datatrak-web/src/api/mutations/useEditUser.ts
```diff
@@ -8,20 +8,25 @@ import { Project } from '@tupaia/types';
...
 
-export const useEditUser = onSuccess => {
+export const useEditUser = (onSuccess?: () => void) => {
   const queryClient = useQueryClient();
 
   return useMutation<any, Error, UserPreferences, unknown>(
-    async ({ projectId, countryId }: UserPreferences) => {
-      if (!projectId && !countryId) {
+    async ({ projectId, countryId, deleteAccountRequested }: UserPreferences) => {
+      if (!projectId && !countryId && !deleteAccountRequested) {
```
tcaiger, line 22:
```suggestion
      if (!projectId && !countryId && deleteAccountRequested === undefined) {
```
Since it's a boolean, it could theoretically be passed in as false and the update wouldn't run as expected. There isn't a use case in UI for setting to false but it's possible there could be in the future and this could lead to a bug.

packages/datatrak-web/src/api/mutations/useEditUser.ts
```diff
@@ -32,14 +37,17 @@ export const useEditUser = onSuccess => {
         updates.country_id = countryId;
       }
 
+      if (deleteAccountRequested) {
```
tcaiger, line 40:
```suggestion
      if (deleteAccountRequested !== undefined) {
```

packages/datatrak-web/src/api/mutations/useRequestDeleteAccount.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import { useMutation } from 'react-query';
+import { post } from '../api';
+import { useEditUser } from '.';
+
+export const useRequestDeleteAccount = () => {
+  // After the user requests to delete their account, we need to update the user's record to make a note that the request has been made
```
tcaiger, line 11: Nice comment and good re-use of the updateUser hook :)

packages/datatrak-web/src/views/AccountSettingsPage/AccountSettingsPage.tsx
```diff
@@ -0,0 +1,52 @@
...
+import React from 'react';
+import styled from 'styled-components';
+import { Typography } from '@material-ui/core';
+import { Settings } from '@material-ui/icons';
+import { PersonalDetailsSection } from './PersonalDetailsSection';
+import { ChangePasswordSection } from './ChangePasswordSection';
+import { RequestCountryAccessSection } from './RequestCountryAccessSection';
+import { DeleteAccountSection } from './DeleteAccountSection';
+
+const Wrapper = styled.div`
```
tcaiger, line 15: Do you think that it would work out ok to use the PageContainer component here?

packages/datatrak-web/src/views/AccountSettingsPage/AccountSettingsPage.tsx
```diff
@@ -0,0 +1,52 @@
...
+`;
+
+export const AccountSettingsPage = () => {
+  return (
+    <Wrapper>
+      <PageTitle>
+        <Settings color="primary" />
+        Account Settings
+      </PageTitle>
+      <PersonalDetailsSection />
```
tcaiger, line 46: thanks for setting up these sections 👍

packages/datatrak-web/src/views/AccountSettingsPage/DeleteAccountSection/UserDetails.tsx
```diff
@@ -0,0 +1,81 @@
...
+    font-size: 1.25rem;
+  }
+  & + & {
+    margin-left: 1rem;
+  }
+`;
+
+export const UserDetails = ({ user }: { user?: DatatrakWebUserRequest.ResBody }) => {
+  const { deleteAccountRequested } = user || ({} as DatatrakWebUserRequest.ResBody);
+  const { data: userRewards } = useUserRewards();
```
tcaiger, line 64: This is not from your change but comes up due to it. There is an error in central server getRewardsForUser `Internal server error: Cannot read property 'coconuts' of undefined` which happens when there aren't any rewards for a user

packages/datatrak-web/src/views/AccountSettingsPage/AccountSettingsSection.tsx
```diff
@@ -0,0 +1,110 @@
...
+export const AccountSettingsSection = ({
+  title,
+  description,
+  leftColumn,
+  centerColumn,
+  rightColumn,
+  button,
+}: AccountSettingsSectionProps) => {
+  return (
+    <Wrapper>
```
tcaiger, line 88: I think this layout component is a bit too opinionated to cover all the sections because:
- the three other sections will be a form that goes across columns 2 and 3
- the button should be a form submit button in the other sections
- the title of the Request country access section includes a custom button

I'm not sure the best solution but perhaps the component could simply be:
- the main layout with 2 columns (one third and then two thirds)
- title prop which could either be text or a custom component
- description

But I don't mind if you have other ideas about how to solve it
- the personal details

tcaiger (changes requested):
Looking really good. Just a couple of small issues. Let me know if you want to discuss. Thanks

tcaiger (approved):
Looks great - thanks. ps. i just tweaked the user rewards to return 0s when there is none which I think makes more sense than returning undefiend.

PR #5136, opened 2023-11-02
No issue: Update branch name validation script

EMcQ-BES:
### Changes:

- `api` is one of our suffixes. In some cases like `data-table-api` and `tupaia-web-api` we don't have an equivalent non `api` suffix so we don't catch invalid branch names
  - e.g. `branch-data-table` would generate `branch-data-table-api` _and_ `branch-data-table-data-table-api` both of which match `*-data-table-api`


acdunham (approved):
LGTM - thank you!

PR #5145, opened 2023-11-06
WAITP-1366: tupaia web fe fixes

tcaiger:
### Changes:

- Fix matrix mobile screen warning
- Fix landing page regressions and integrate url params into useLandingPage hook
- Fix permissions entity error handling

### DIFF
--- a/packages/tupaia-web/src/api/queries/useLandingPage.ts
+++ b/packages/tupaia-web/src/api/queries/useLandingPage.ts
@@ -5,14 +5,13 @@
  */
 
 import { useQuery } from 'react-query';
-import { useLocation } from 'react-router';
+import { useParams } from 'react-router';
 import { get } from '../api';
 import { SingleLandingPage } from '../../types';
 
-export const useLandingPage = (urlSegment?: string) => {
-  const { pathname } = useLocation();
-  // occasionally the urlSegment is not passed in, so we need to get it from the pathname. This is in cases where this hook is accessed from a modal, for example, that doesn't have access to the urlSegment param
-  const landingPageUrlSegment = urlSegment || pathname.split('/')[1];
+export const useLandingPage = () => {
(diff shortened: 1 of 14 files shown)


packages/tupaia-web/src/api/queries/useLandingPage.ts
```diff
@@ -5,14 +5,13 @@
...
+import { useParams } from 'react-router';
 import { get } from '../api';
 import { SingleLandingPage } from '../../types';
 
-export const useLandingPage = (urlSegment?: string) => {
-  const { pathname } = useLocation();
-  // occasionally the urlSegment is not passed in, so we need to get it from the pathname. This is in cases where this hook is accessed from a modal, for example, that doesn't have access to the urlSegment param
-  const landingPageUrlSegment = urlSegment || pathname.split('/')[1];
+export const useLandingPage = () => {
+  const { landingPageUrlSegment } = useParams();
```
tcaiger, line 13: I ended up just combining the url param into the api hook since the usLanding page check is based on the url anyway

packages/tupaia-web/src/api/queries/useLandingPage.ts
```diff
@@ -5,14 +5,13 @@
  */
 
 import { useQuery } from 'react-query';
-import { useLocation } from 'react-router';
+import { useParams } from 'react-router';
 import { get } from '../api';
 import { SingleLandingPage } from '../../types';
 
-export const useLandingPage = (urlSegment?: string) => {
-  const { pathname } = useLocation();
```
acdunham, line 13: I think this part here might still be needed - in some views, (e.g. in a modal route) the `landingPageUrlSegment` param isn't accessible, so it has to be grabbed from the pathname

acdunham (approved):
LGTM, thanks! Just one question but other than that good to go

PR #5146, opened 2023-11-06
WAITP-1456: Save selected map overlay dates

tcaiger:
### Changes:

- Listen for urls changes to the map overlay and map overlay date params and save the selected date range in state
- On user change of the map overlay, check if there is a saved value and use it if so


acdunham (approved):
Nice work

PR #5149, opened 2023-11-07
WAITP-1355: activity feed

acdunham:
### Changes:
- Setup activity feed with infinite scroll
- Create `/activityFeed` endpoint in orchestration server that handles permissions

---

### Screenshots:
Screenshot 2023-11-07 at 3 05 03 pm
Screenshot 2023-11-07 at 3 05 16 pm


packages/types/src/types/models-extra/leaderboard.ts
```diff
@@ -3,12 +3,10 @@
  * Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
  */
 
-import { UserAccount } from '../models';
-
 export type LeaderboardItem = {
-  user_id: UserAccount['id'];
-  first_name: UserAccount['first_name'];
-  last_name: UserAccount['last_name'];
```
acdunham, line 11: This was causing a circular dependencies error

packages/datatrak-web-server/src/routes/ActivityFeedRoute.ts
```diff
@@ -0,0 +1,67 @@
...
+    const feedItems = await models.feedItem.find(conditions, {
+      limit,
+      offset: page * NUMBER_PER_PAGE,
+      sort: ['creation_date DESC'],
+    });
+    const hasMorePages = feedItems.length > NUMBER_PER_PAGE;
+
+    const items = (
+      await Promise.all(feedItems.slice(0, NUMBER_PER_PAGE).map(f => f.getData()))
+    ).map((item: any) => ({ ...item, creation_date: new Date(item.creation_date!).toJSON() }));
```
tcaiger, line 54:
```suggestion
      await Promise.all(feedItems.slice(0, NUMBER_PER_PAGE).map(feedItem => feedItem.getData()))
    ).map((feedItem: any) => ({ ...feedItem, creation_date: new Date(item.creation_date!).toJSON() }));
```
Very minor but might as well use the full name to make it easy to ready

packages/datatrak-web/src/api/queries/useActivityFeed.ts
```diff
@@ -0,0 +1,22 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+import { useInfiniteQuery } from 'react-query';
+import { get } from '../api';
+import { DatatrakWebActivityFeedRequest } from '@tupaia/types';
```
tcaiger, line 7: Just need to re-order the imports

packages/datatrak-web/src/views/LandingPage/ActivityFeedSection/ActivityFeedSurveyItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const getLocationName = () => {
+    if (templateVariables?.regionName)
+      return `${templateVariables.regionName}, ${templateVariables.countryName}`;
+
+    return templateVariables?.countryName;
+  };
+
+  const locationName = getLocationName();
+  return (
```
tcaiger, line 60: The styles don't quite match the design for me. It looks like there is a bit too much space and the image is a bit too big

packages/datatrak-web/src/views/LandingPage/ActivityFeedSection/ActivityFeedSurveyItem.tsx
```diff
@@ -0,0 +1,75 @@
+/*
+ * Tupaia
+ *  Copyright (c) 2017 - 2023 Beyond Essential Systems Pty Ltd
+ */
+
+import React from 'react';
+import styled from 'styled-components';
+import { format } from 'date-fns';
```
tcaiger, line 8: Good idea adding date-fns. We are in a bit of a transition time as we still have moment in ui-components but I think that it makes sense to draw a line and use date-fns in datatrak-web. Unfortunately we have already installed moment and used it in a couple of places in the app. Would you mind doing a follow up PR or creating a ticket to switch out the other uses of moment in datatrak-web to date-fns. Better sooner rather than later.

acdunham: I actually ended up using the built in JS date in the second part of this - it was better for locale date strings I found

packages/datatrak-web/src/views/LandingPage/ActivityFeedSection/InfiniteScroll.tsx
```diff
@@ -0,0 +1,75 @@
...
+    if (isVisible && !isFetchingNextPage) onScroll();
+  };
+
+  useEffect(() => {
+    // add scroll listener
+    container?.current?.addEventListener('scroll', handleScroll);
+
+    // remove scroll listener on unmount
+    return () => {
+      console.log('removing');
```
tcaiger, line 64:
```suggestion

```

packages/datatrak-web/src/views/LandingPage/ActivityFeedSection/InfiniteScroll.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+const Loading = styled(Typography).attrs({
+  variant: 'body2',
+  color: 'textSecondary',
+})`
+  text-align: center;
+  padding: 1rem;
+`;
+
+const getIsAtEndOfList = (container?: HTMLDivElement, loader?: HTMLDivElement) => {
```
tcaiger, line 25: This is a nice way of doing this of triggering the next fetch

packages/datatrak-web/src/views/LandingPage/ActivityFeedSection/InfiniteScroll.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+interface InfiniteScrollProps {
+  children: ReactNode;
+  onScroll: () => void;
+  hasNextPage?: boolean;
+  isFetchingNextPage?: boolean;
+}
+
+export const InfiniteScroll = ({
```
tcaiger, line 43: Very tidy implementation of infinite scroll

tcaiger (approved):
A couple of tiny things but I'm happy to approve

tcaiger:
this is not a blocker because it doesn't impact how it works in any way but when I tested it I noticed that there are duplicate requests going out. If you had any idea what that might be it could be good to look into.

image

packages/datatrak-web/src/views/LandingPage/ActivityFeedSection/ActivityFeedSurveyItem.tsx
tcaiger: ok that's a good point. I think that we should make a date display component but I just remembered that Georgia is going to make a ticket for it so we can figure that out later :)

PR #5153, opened 2023-11-08
WAITP-1478 Recent survey response endpoint

EMcQ-BES:
### Changes:

- Add route to fetch single survey response with answers


packages/datatrak-web-server/src/routes/RecentSurveyResponseRoute.ts
```diff
@@ -0,0 +1,36 @@
...
+  public async buildResponse() {
+    const { ctx, params } = this.req;
+    const { id: responseId } = params;
+
+    const surveyResponse = await ctx.services.central.fetchResources(`surveyResponses/${responseId}`);
+    const answerList = await ctx.services.central.fetchResources('answers', {
+      filter: { survey_response_id: surveyResponse.id },
+      columns: ANSWER_COLUMNS,
+    });
+    const answers = answerList.reduce((output: Record<string, string>, answer: any) => ({ ...output, [answer['question.code']]: answer.text }), {});
```
tcaiger, line 32: I like this data format, but could we switch out the question.code for question.id since that is what we use everywhere for tracking the questions.

packages/datatrak-web-server/src/routes/RecentSurveyResponseRoute.ts
```diff
@@ -0,0 +1,36 @@
...
+  DatatrakWebRecentSurveyResponseRequest.ResBody,
+  DatatrakWebRecentSurveyResponseRequest.ReqBody,
+  DatatrakWebRecentSurveyResponseRequest.ReqQuery
+>;
+
+const ANSWER_COLUMNS = [
+  'text', 'question.code'
+]
+
+export class RecentSurveyResponseRoute extends Route<RecentSurveyResponseRequest> {
```
tcaiger, line 22: This is pretty subjective so I don't mind if you want to keep it as is but I would prefer this to be named SurveyResponseRoute since it doesn't know (or need to know) that we are using it for recent survey responses.

tcaiger (approved):
Looks good. Just one actual requested change of the question.code to question.id but happy to pre-approve.

packages/datatrak-web-server/src/routes/RecentSurveyResponseRoute.ts
EMcQ-BES: I can definitely change it. The only issue is that it looks a bit messy/confusing when we have stuff like:

```
SurveyResponsesRoute,
SurveyResponsesRequest,
SurveyResponseRoute,
SurveyResponseRequest,
```

EMcQ-BES: How about something like `SurveyResponseAnswersRoute`? Captures the use case _and_ differentiates it

tcaiger: SingleSurveyResponseRoute sounds good to me 👍

PR #5156, opened 2023-11-10
WAITP-1452: Datatrak-web refactor survey context

tcaiger:
### Changes:

- Refactor Survey cancel modal to be generic
- Move SurveyContext down to be in the survey page


packages/datatrak-web/src/api/mutations/useLogin.ts
```diff
@@ -12,12 +12,19 @@ type LoginCredentials = {
...
+    return typeof state.from === 'string';
+  }
+  return false;
+}
 export const useLogin = () => {
   const queryClient = useQueryClient();
   const navigate = useNavigate();
   const location = useLocation();
 
-  const { from } = location.state as { from: string };
```
tcaiger, line 20: I was getting an error on first login where state was null so I added some more robust checking.

acdunham: We could probably simplify this to be just `location.state && location.state.from`

packages/datatrak-web/src/layout/Header.tsx
```diff
@@ -47,27 +46,30 @@ const LogoButton = styled(Button)`
...
 export const Header = () => {
-  const { surveyCode } = useParams();
-  const { openCancelConfirmation, isSuccessScreen } = useSurveyForm();
-  const isActiveSurvey = !!surveyCode && !isSuccessScreen;
-  const onClickLogo = () => {
-    if (surveyCode && !isSuccessScreen) {
-      openCancelConfirmation();
+  const [surveyCancelModalIsOpen, setIsOpen] = useState(false);
+  const isSurveyScreen = !!useMatch(ROUTES.SURVEY_SCREEN);
+  const isSuccessScreen = !!useMatch(ROUTES.SURVEY_SUCCESS);
```
acdunham, line 51: Nice, I like this

acdunham (approved):
Looks good, thanks for doing that. A couple small comments but nothing major

packages/datatrak-web/src/api/mutations/useLogin.ts
tcaiger: I tried that but I get a TS error. I could cast it using `as` but Rohan said that it's not good practise to cast and that I should use typeguards so I'm trying that out. It seems a bit verbose so maybe there is a more concise way of doing it.

PR #5159, opened 2023-11-10
WAITP-1474: Update success illustration on DataTrak

jaskfla:
### Changes:

- Replaced high-five graphic, shown upon survey submission success, with new version. (This is already used in activity feed.)
- Renamed asset file to be more generic/reusable, rather than coupling it to activity feed feature.
- Relabelled `Close` button to `Return to dashboard`.
- RAYG: Changed `margin-bottom` to `margin-block-end` where semantically appropriate (which in this file happened to be all instances). If unclear on rationale, see [MDN explanation of block/inline axes](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flow_layout/Block_and_inline_layout_in_normal_flow).

---

#### 10 Nov 2023

Screenshot 2023-11-10T150311

#### 13 Nov 2023

Screenshot 2023-11-13T111224


acdunham (changes requested):
Nice work. Just a comment from me about the width setting.

Also, it's likely not related to your change, but while you are working on this view, would you be able to look into this overflow issue on mobile sizes?
Screenshot 2023-11-13 at 9 59 44 am

packages/datatrak-web/src/features/Survey/Screens/SurveySuccessScreen.tsx
```diff
@@ -30,9 +30,11 @@ const Container = styled.div`
 `;
 
 const StyledImg = styled.img`
+  aspect-ratio: 1;
+  width: 369px;
```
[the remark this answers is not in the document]
jaskfla: Ah, great point—thanks for pointing this out! I’ll see what works better between `rems` and `vw`

jaskfla:
Absolutely, I’ll see how gracefully I can make it respond to smaller viewports

packages/datatrak-web/src/views/SurveyPage.tsx
```diff
@@ -36,7 +36,7 @@ const SurveyScreenContainer = styled.div<{
   overflow: ${({ $scrollable }) => ($scrollable ? 'auto' : 'hidden')};
   align-items: flex-start;
   height: calc(100vh - ${HEADER_HEIGHT} - ${SURVEY_TOOLBAR_HEIGHT});
-  width: 100vw;
+  width: 100%;
```
jaskfla, line 39: @acdunham Looks like you added the `100vw` last week to fix some layout things on mobile. Is it possible this change will have broken something else? I’m not sure if `SurveyScreenContainer` ever wraps something _other_ than `SurveySuccessScreen`.

The padding on this container was causing the overflow in narrow windows.

acdunham: It's possible - you might need to play around with the different views to make sure it all looks legit at all screen sizes 🙏

jaskfla: Had a squiz (in particular `SurveyLayout`) and looks like the `%` actually does fix the same issue for the different views

acdunham: I think the 100vw was there to fix this issue, as that layout is shared between multiple screens. Ideally there is a fix for both issues
Screenshot 2023-11-13 at 3 07 56 pm

jaskfla: Huh, I think I’ve accidentally made a mess of this branch. The behaviour from your screenshot doesn’t happen on the epic branch.

I think what’s happened is:

1. I originally branched of `dev` (incorrectly) and pushed that commit
2. I deleted my local branch, and created a new one, correctly branching off `datatrak-web-epic`
3. …but I gave the branch the same name, so when I pushed to GitHub it pushed to the existing remote branch

Which would also explain this 😅

Screenshot 2023-11-13T162957

Given the small scope of this change, I think tidiest for me to redo this on (yet) another branch **with a different name**

jaskfla: Looks like the clipping of the bottom toolbar (with Back, Cancel and Next) happens with both `100%` or `100vw`. I’ve just merged the latest code from the epic branch, and this is what it’s doing at 320 and 220px viewports:

Screenshot 2023-11-13T165817

Screenshot 2023-11-13T165834

The most straightforward workaround I can think of would be to use a single-column button group in the small size class, but it doesn’t quite jive with the current styling of these three buttons:

Frame 5

Versus:

Frame 6

…but this would mean also changing the non-narrow styling.

acdunham: Probably don't need to worry about 220px width - I don't know if there are any devices that narrow, and if there are it's not likely they will be commonly used. 320px wide is usually the limit people work with

acdunham (approved):
Looks good, thanks for doing those changes

PR #5160, opened 2023-11-12
Survey not found error handling

acdunham:
### Changes:

- Add error handling when an error is caught when fetching the survey

---

### Screenshots:
 

Screenshot 2023-11-13 at 9 50 47 am


packages/datatrak-web/src/AppProviders.tsx
```diff
@@ -21,16 +21,20 @@ const defaultQueryClient = new QueryClient({
...
+        errorToast(error.message);
+      }
+    },
+  }),
   defaultOptions: {
     queries: {
       staleTime: 1000 * 60 * 5, // 5 minutes
       refetchOnWindowFocus: false,
-      retry: 0,
+      retry: false,
```
tcaiger, line 36: ah great find - I assume that this will fix the other places such as login where the request errors show up multiple times

acdunham: I actually think the errors showing up multiple times is a side effect of various queries being present on different screens and possibly causing re-renders when we have redirects. So from what I could tell, for the most part when we handle surveys not being found before it redirects infinitely, then this issue is handled (90% of the time)

packages/datatrak-web/src/Routes.tsx
```diff
@@ -102,14 +102,34 @@ const SurveyStartRedirect = () => {
...
   if (visibleScreens && visibleScreens.length && visibleScreens.length < Number(screenNumber)) {
     return <SurveyStartRedirect />;
   }
   return children;
 };
 
+/**
+ * This is to redirect the user to the survey not found page if they try to access a survey that does not exist
+ */
+const SurveyNotFoundRedirect = ({ children }) => {
```
tcaiger, line 115: This is a good idea. 
On a side note this file is getting a bit big. Maybe next time we add something we should split it up a bit

packages/datatrak-web/src/components/ErrorDisplay.tsx
```diff
@@ -0,0 +1,51 @@
...
+  border: 1px solid ${({ theme }) => theme.palette.divider};
+  h1 {
+    margin-bottom: 1rem;
+  }
+  ${({ theme }) => theme.breakpoints.up('sm')} {
+    padding: 1.5rem 2.5rem;
+  }
+`;
+
+export const ErrorDisplay = ({
```
tcaiger, line 33: Also a great idea - it's handy to have this component

packages/datatrak-web/src/views/NotFoundPage.tsx
```diff
@@ -3,12 +3,22 @@
...
+import { DialogActions } from '@material-ui/core';
+import { Button, ErrorDisplay } from '../components';
 
-export const NotFoundPage = () => {
+interface NotFoundPageProps {
+  error?: Error;
+  title?: string;
+}
+
+export const NotFoundPage = ({ error, title = '404: Page not found' }: NotFoundPageProps) => {
```
tcaiger, line 14: Can I make a request to update the NotFoundPage to be called ErrorPage since it's now generic. For example we might use it for Permission Denied errors or something. Also would you be able to tweak the styles so the Paper is in the middle of the page 🙏

tcaiger (approved):
Thanks for picking up these extra changes, it's great to have these error handled gracefully! Left a request for a couple of tweaks but I'm happy to pre-approve.

PR #5163, opened 2023-11-12
WAITP-1224 Add an additional filter to MapOverlaysRoute

EMcQ-BES:
### Issue #: WAITP-1224

When fetching map overlays from `central-server` we don't fetch in the context of an entity. So if a map overlay is available in both `DL` and `VU` for example, we will return it if you have appropriate permissions to either country. When we fetch for `tupaia-web-server` we want to filter specifically for the entity we're looking at.

### Changes:

- Add an additional permission filter for map overlays

---


tcaiger (approved):
Tino Pai - gracias

PR #5164, opened 2023-11-13
WAITP-1490: project id for surveys

acdunham:
### Changes:
- Add migration for project_id column in surveys
- Update queries that relate to surveys to allow for optional projectId filter

### DIFF
--- a/packages/database/src/migrations/20231112212117-AddProjectIdToSurveys-modifies-schema.js
+++ b/packages/database/src/migrations/20231112212117-AddProjectIdToSurveys-modifies-schema.js
@@ -15,7 +15,11 @@ exports.setup = function (options, seedLink) {
 };
 
 exports.up = function (db) {
-  return db.runSql(`ALTER TABLE survey ADD project_id TEXT NULL`);
+  return db.runSql(`
+    ALTER TABLE survey
+    ADD COLUMN project_id TEXT,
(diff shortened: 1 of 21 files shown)


packages/database/src/migrations/20231112212117-AddProjectIdToSurveys-modifies-schema.js
```diff
@@ -0,0 +1,29 @@
...
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function (db) {
+  return db.runSql(`ALTER TABLE survey ADD project_id TEXT NULL`);
```
tcaiger, line 18: I'm not sure of the syntax but we should also set this as a foreign key. Something like `FOREIGN KEY (project_id) REFERENCES project(id)` 

Also I checked with the others and we generally add an index for any foreign keys. Again I'm not 100% sure on the syntax.

tcaiger (changes requested):
Just a couple of comments on the migration. Everything else looks good 👍

tcaiger (approved):
Nice - thanks for picking this up

PR #5165, opened 2023-11-13
No issue - fix undefined bug from rn-1059

chris-pollard:


acdunham (approved):
Thanks for doing this!

PR #5166, opened 2023-11-13
WAITP-1389: Add project column to admin panel

acdunham:
### Changes:
- Add project column to surveys on admin panel

---

### Screenshots:
Screenshot 2023-11-13 at 3 57 04 pm


tcaiger (approved):
I think we overestimates this one haha

